PR #1409, opened 2015-02-24
remove redundant per app agg for metrics

abaranau:
It is a premature optimization. After changing aggregations yesterday there's even less obvious reasons to keep it


gsps1:
LGTM

abaranau:
actually this is needed. And one test fails that indicates that :)

PR #1411, opened 2015-02-24
bugfix : Allow query metrics with wildcard

gsps1:
due to ImmutableMap, we were returning 500 when there were null in tag-value.

build running : http://builds.cask.co/browse/CDAP-DUT896-1


abaranau:
LGTM

PR #1412, opened 2015-02-24
Catch exceptions thrown in the notification feed handler

jguery:
It removes exceptions shown in the master logs when starting CDAP.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NotificationFeedHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NotificationFeedHttpHandler.java
@@ -96,7 +96,10 @@ public class NotificationFeedHttpHandler extends AuthenticatedHttpHandler {
       responder.sendString(HttpResponseStatus.BAD_REQUEST, "Invalid json object provided in request body.");
     } catch (IOException e) {
       LOG.error("Failed to read Notification feed request body.", e);
-      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
+    } catch (Throwable t) {
+      LOG.debug("Error in creating notification feed.", t);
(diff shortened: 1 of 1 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NotificationFeedHttpHandler.java
```diff
@@ -96,7 +96,10 @@ public void createFeed(HttpRequest request, HttpResponder responder,
       responder.sendString(HttpResponseStatus.BAD_REQUEST, "Invalid json object provided in request body.");
     } catch (IOException e) {
       LOG.error("Failed to read Notification feed request body.", e);
-      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
+    } catch (Throwable t) {
+      LOG.debug("Error in creating notification feed.", t);
```
ParthGandhi: Error log?

ParthGandhi:
I think these should all be error logs.

Otherwise LGTM.

jguery:
The thing is these exceptions are always thrown when cdap starts up, because the dataset service is not up and a runtime exception is thrown. It's not exactly an error, and if we log them as such, then people might think something is really wrong when it's not.

ParthGandhi:
But what happens when it actually *is* an error (say once everything comes up correctly)? Then you still log as debug..

Is it possible to distinguish between exceptions thrown when DS is down and other runtime errors?

jguery:
I could do that but then DS down could still be a valid error, when it happens when CDAP is not starting... I feel like passing the exception message to the responder is enough though, to understand what went wrong.

ParthGandhi:
Okay, LGTM as is. 

Just FYI, response messages are almost always useless. All post-failure debugging happens through logs. But in this case I think thats OK.

jguery:
I would have left it as is, as error, if it wasn't for that JIRA https://issues.cask.co/browse/CDAP-1578.

ParthGandhi:
Yeah its fine. Theres also a JIRA for making all our services more resilient to dependent services not being up.

PR #1413, opened 2015-02-24
fix for double namespacing metrics tables during upgrade

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-DUT880-3

abaranau:
tested on cluster

gsps1:
LGTM

PR #1416, opened 2015-02-25
fix metrics discovery for old dashboard

abaranau:
So old dashboard expects v2 metrics API to only return user metrics...


abaranau:
http://builds.cask.co/browse/CDAP-DUT903-1

gsps1:
LGTM

PR #1417, opened 2015-02-25
Remove remaining truncation logic in stream size aggregation, and always...

jguery:
... send an init notification when creating a stream/starting cdap

Build: http://builds.cask.co/browse/CDAP-DUT905-1

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/BasicStreamWriterSizeCollector.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/BasicStreamWriterSizeCollector.java
@@ -21,7 +21,6 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.AbstractIdleService;
-import com.google.inject.Inject;
 import org.apache.twill.common.Cancellable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -40,7 +39,6 @@ public class BasicStreamWriterSizeCollector extends AbstractIdleService implemen
   private final ConcurrentMap<Id.Stream, AtomicLong> streamSizes;
   private final List<Cancellable> truncationSubscriptions;
 
-  @Inject
   public BasicStreamWriterSizeCollector() {
     this.streamSizes = Maps.newConcurrentMap();
     this.truncationSubscriptions = Lists.newArrayList();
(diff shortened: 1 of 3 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/BasicStreamWriterSizeCollector.java
```diff
@@ -39,13 +37,11 @@
 public class BasicStreamWriterSizeCollector extends AbstractIdleService implements StreamWriterSizeCollector {
   private static final Logger LOG = LoggerFactory.getLogger(BasicStreamWriterSizeCollector.class);
 
-  private final StreamCoordinatorClient streamCoordinatorClient;
   private final ConcurrentMap<Id.Stream, AtomicLong> streamSizes;
   private final List<Cancellable> truncationSubscriptions;
 
   @Inject
```
albertshau: dont need inject without args

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/DistributedStreamService.java
```diff
@@ -555,13 +554,15 @@ protected StreamSizeAggregator(Id.Stream streamId, long baseCount, int streamThr
...
-      this.isInit = true;
       this.streamId = streamId;
       this.streamFeed = new Id.NotificationFeed.Builder()
         .setNamespaceId(streamId.getNamespaceId())
         .setCategory(Constants.Notification.Stream.STREAM_FEED_CATEGORY)
         .setName(String.format("%sSize", streamId.getName()))
         .build();
+
+      // Publish an initialization notification
+      publishNotification(baseCount);
```
albertshau: it's usually not a good idea to do any work in a constructor, especially if you're connecting to other systems. There is a hidden requirement for the other system to be up and running when you create the object. I think it's better just to call publish explicitly after you create it, especially if this is a private class.  What happens if the publish fails? Do we still want to do the init check in case it the initial notification failed or are we ok with dropping notifications?

albertshau:
small comment and a question

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/BasicStreamWriterSizeCollector.java
jguery: Ok will remove.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/DistributedStreamService.java
jguery: With notifications we are okay if a publish fail. I'll move the publish call, like you suggested.

jguery:
@albertshau I addressed your comments. Please have another look. Thank you

albertshau:
:+1:

PR #1418, opened 2015-02-25
Source backwards compatability

anwar6953:
Users should have to modify as little code as possible when they upgrade CDAP versions (marking it as deprecated first so they have plenty of time to modify the code to new APIs).

Brought back a copy of TxRunnable in cdap.api.service package, and changed AppWithServices to use it to test it. Simply delegates execution to the TxRunnable in cdap.api package.
Also removed the checked Exception from TestBase#clear method (it has been deprecated for 2.8, so we should support it as is in previous version, until we remove it).

Unsure how to handler RuntimeStats#clearStats(String). Perhaps we can't do anything but remove that method and require the user to modify their code.

http://builds.cask.co/browse/CDAP-DUT904-5

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/service/ServiceWorkerContext.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/service/ServiceWorkerContext.java
@@ -41,6 +41,12 @@ public interface ServiceWorkerContext extends RuntimeContext, ServiceDiscoverer 
   void execute(TxRunnable runnable);
 
   /**
+   * Execute a set of operations on datasets via a {@link TxRunnable} that are committed as a single transaction.
+   * @param runnable The runnable to be executed in the transaction
+   */
+  void execute(co.cask.cdap.api.service.TxRunnable runnable);
+
+  /**
    * @return Number of instances of this worker.
    */
   int getInstanceCount();
--- a/cdap-api/src/main/java/co/cask/cdap/api/service/TxRunnable.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/service/TxRunnable.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.service;
+
+import co.cask.cdap.api.data.DatasetContext;
+import co.cask.cdap.api.dataset.Dataset;
+
+/**
+ * A runnable that provides a {@link DatasetContext} to programs which may be used to get
+ * access to and use datasets.
+ * @deprecated As of version 2.8.0, replaced by {@link co.cask.cdap.api.TxRunnable}
+ */
+public interface TxRunnable {
+
+  /**
+   * Provides a {@link DatasetContext} to get instances of {@link Dataset}s.
+   *
+   * <p>
+   *   Operations executed on a dataset within the execution of this method are committed as a single transaction.
+   *   The transaction is started before this method is invoked and is committed upon successful execution.
+   *   Exceptions thrown while committing the transaction or thrown by user-code result in a rollback of the
+   *   transaction.
+   * </p>
+   * @param context to get datasets from.
+   * @throws Exception
+   */
+  void run(DatasetContext context) throws Exception;
+
+}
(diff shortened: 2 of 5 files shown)


cdap-unit-test/src/main/java/co/cask/cdap/test/TestBase.java
```diff
@@ -171,9 +171,14 @@ protected ApplicationManager deployApplication(Class<? extends Application> appl
...
-  protected void clear() throws Exception {
-    TestManager testManager = getTestManager();
-    testManager.clear();
+  protected void clear() {
+    try {
+      TestManager testManager = getTestManager();
+      testManager.clear();
+    } catch (Exception e) {
+      // Unchecked exception to maintain compatibility until we remove this method
+      Throwables.propagate(e);
```
ParthGandhi: `throw Throwables.propagate(e)`

ParthGandhi:
LGTM

cdap-unit-test/src/main/java/co/cask/cdap/test/TestBase.java
anwar6953: It seems like the propagate method already throws it:
throw new RuntimeException(throwable);

ParthGandhi: https://code.google.com/p/guava-libraries/wiki/ThrowablesExplained

"... but throwing the result -- e.g. throw Throwables.propagate(t) -- can be useful to prove to the compiler that an exception will be thrown."

cdap-api/src/main/java/co/cask/cdap/api/service/TxRunnable.java
```diff
@@ -0,0 +1,43 @@
...
+
+import co.cask.cdap.api.data.DatasetContext;
+import co.cask.cdap.api.dataset.Dataset;
+
+/**
+ * A runnable that provides a {@link DatasetContext} to programs which may be used to get
+ * access to and use datasets.
+ * @deprecated As of version 2.8.0, replaced by {@link co.cask.cdap.api.TxRunnable}
+ */
+public interface TxRunnable {
```
chtyim: You need to mark the interface with `@Deprecate` as well.

cdap-api/src/main/java/co/cask/cdap/api/service/ServiceWorkerContext.java
```diff
@@ -41,6 +41,12 @@
   void execute(TxRunnable runnable);
 
   /**
+   * Execute a set of operations on datasets via a {@link TxRunnable} that are committed as a single transaction.
+   * @param runnable The runnable to be executed in the transaction
+   */
+  void execute(co.cask.cdap.api.service.TxRunnable runnable);
```
nitinmotgi: Why whole package name ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/BasicServiceWorkerContext.java
```diff
@@ -178,6 +179,16 @@ public void execute(TxRunnable runnable) {
   }
 
   @Override
+  public void execute(final co.cask.cdap.api.service.TxRunnable runnable) {
```
nitinmotgi: Why whole namespace ?

anwar6953: We use TxRunnable from two separate packages in this class.

cdap-api/src/main/java/co/cask/cdap/api/service/ServiceWorkerContext.java
anwar6953: We use TxRunnable from two separate packages in this class.

ParthGandhi: This method should also be marked deprecated.

cdap-unit-test/src/main/java/co/cask/cdap/test/TestBase.java
```diff
@@ -157,6 +157,7 @@ protected static TestManager getTestManager() {
    * @param applicationClz The application class
    * @return An {@link co.cask.cdap.test.ApplicationManager} to manage the deployed application.
    */
+  @Deprecated
```
ParthGandhi: For the methods in this class marked deprecated, add a link to the method that should be used instead.

cdap-api/src/main/java/co/cask/cdap/api/service/ServiceWorkerContext.java
anwar6953: Isn't it sufficient to mark the interface as deprecated

cdap-unit-test/src/main/java/co/cask/cdap/test/TestBase.java
anwar6953: All of them already have it in the comments.
For instance:
`* @deprecated Use {@link TestManager#deployApplication(Class, java.io.File...)} from {@link #getTestManager()}.`

ParthGandhi: My bad. You're right.

cdap-api/src/main/java/co/cask/cdap/api/service/ServiceWorkerContext.java
chtyim: You have to mark this method as deprecate since this method will get removed in future version.

anwar6953:
So good to merge?

ParthGandhi:
Merge it.

PR #1419, opened 2015-02-25
Stop queue config table cache update thread when the table is gone.

chtyim:
- Since we have one config table per namespace, the cache update thread should stop if the table doesn’t exists.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/coprocessor/ConsumerConfigCache.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/coprocessor/ConsumerConfigCache.java
@@ -20,8 +20,10 @@ import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.data2.transaction.queue.QueueConstants;
 import co.cask.cdap.data2.transaction.queue.QueueEntryRow;
 import co.cask.cdap.data2.util.hbase.ConfigurationTable;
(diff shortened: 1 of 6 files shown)


chtyim:
Build pass: http://builds.cask.co/browse/CDAP-DUT906-1

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/coprocessor/ConsumerConfigCache.java
```diff
@@ -161,18 +169,27 @@ public synchronized void updateCache() {
...
           updateConfig();
           long now = System.currentTimeMillis();
           if (now > (lastUpdated + configCacheUpdateFrequency)) {
-            updateCache();
+            try {
+              updateCache();
+              tableNotFoundCount = 0;
+            } catch (TableNotFoundException e) {
+              tableNotFoundCount++;
+              LOG.error("Queue config table not found: {}. Retries remaining {}",
```
jwang47: Maybe not log single retry failure in ERROR, but instead at INFO and log as ERROR when we stop trying?

jwang47:
One minor comment, otherwise LGTM

chtyim:
@jwang47 Updated the PR. Logic changed a bit, please take a look again.

jwang47:
LGTM

chtyim:
build passed after the latest change http://builds.cask.co/browse/CDAP-DUT906-3

PR #1420, opened 2015-02-25
Fix Distributed mode instance update

gokulavasan:
Bamboo : http://builds.cask.co/browse/CDAP-DUT907

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
@@ -24,6 +24,7 @@ import co.cask.cdap.api.flow.FlowletConnection;
 import co.cask.cdap.api.flow.FlowletDefinition;
 import co.cask.cdap.api.mapreduce.MapReduceSpecification;
 import co.cask.cdap.api.procedure.ProcedureSpecification;
+import co.cask.cdap.api.worker.WorkerSpecification;
 import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.runtime.ProgramRuntimeService;
 import co.cask.cdap.app.services.Data;
@@ -528,6 +529,12 @@ public abstract class AbstractAppFabricHttpHandler extends AuthenticatedHttpHand
           for (MapReduceSpecification mrSpec : appSpec.getMapReduce().values()) {
             if (data == Data.DATASET && mrSpec.getDataSets().contains(name)) {
               result.add(makeProgramRecord(appSpec.getName(), mrSpec, ProgramType.MAPREDUCE));
+            }
+          }
+        } else if (type == ProgramType.WORKER) {
+          for (WorkerSpecification workerSpec : appSpec.getWorkers().values()) {
+            if (data == Data.DATASET && workerSpec.getDatasets().contains(name)) {
+              result.add(makeProgramRecord(appSpec.getName(), workerSpec, ProgramType.WORKER));
             }
           }
         }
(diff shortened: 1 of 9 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -530,6 +531,12 @@ protected final void programListByDataAccess(HttpRequest request, HttpResponder
               result.add(makeProgramRecord(appSpec.getName(), mrSpec, ProgramType.MAPREDUCE));
             }
           }
+        } else if (type == ProgramType.WORKER) {
+          for (WorkerSpecification workerSpec : appSpec.getWorkers().values()) {
```
sreevatsanraman: No Unit tests for `listProgramsByDataAccess` for workers. Can you add them?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/WorkerTwillProgramController.java
```diff
@@ -45,11 +44,11 @@ protected void doCommand(String name, Object value) throws Exception {
...
     }
 
-    Map<String, Integer> command = (Map<String, Integer>) value;
+    Map<String, String> command = (Map<String, String>) value;
     try {
-      for (Map.Entry<String, Integer> entry : command.entrySet()) {
+      for (Map.Entry<String, String> entry : command.entrySet()) {
         LOG.info("Changing worker instance count: {} new count is: {}", entry.getKey(), entry.getValue());
-        changeInstances(entry.getKey(), entry.getValue());
+        changeInstances(entry.getKey(), Integer.valueOf(entry.getValue()));
```
nitinmotgi: This might throw an exception right ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkerProgramRunner.java
```diff
@@ -65,7 +65,7 @@ protected ProgramController launch(Program program, ProgramOptions options, File
 
     TwillController controller = launcher.launch(new WorkerTwillApplication(program, workerSpec, hConfFile, cConfFile,
                                                                             eventHandler));
-    return new WorkerTwillProgramController(program, controller).startListen();
+    return new WorkerTwillProgramController(program.getName(), controller).startListen();
```
nitinmotgi: Why String ?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
gokulavasan: Added. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/WorkerTwillProgramController.java
gokulavasan: We always verify the string is derived from an integer before passing it on to the controller. We can change it to always pass in integer.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkerProgramRunner.java
gokulavasan: It needs just the program name to print log messages and for naming threads.

gokulavasan:
@ParthGandhi Please review. Thanks!

ParthGandhi:
:+1:

PR #1421, opened 2015-02-25
Refactored AppLifecycleHttpHandler.{deploy,getAllApps,getApp}

jwang47:
This is needed to make application authorization cleaner.

* Throw `ApplicationNotFoundException` or `NamespaceNotFoundException` instead of returning null
* Move `deploy`, `getAllApps`, `getApp` logic to `AppLifecycleService`
* Make `AppLifecycleHttpHandler` use `AppLifecycleService`


jwang47:
http://builds.cask.co/browse/CDAP-DUT901

PR #1422, opened 2015-02-25
Service Instances => Service Handler Instances; Remove reference to runnable

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-388

Bamboo : http://builds.cask.co/browse/CDAP-DUT911

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -1079,7 +1079,7 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
   }
 
   /**
-   * Return the number of instances of a service handler.
+   * Return the number of instances of a service.
    */
   @GET
   @Path("/apps/{app-id}/services/{service-id}/instances")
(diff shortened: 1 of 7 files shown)


cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
```diff
@@ -389,6 +391,27 @@ public void setProcedureInstances(String appId, String procedureId, int instance
   }
 
   /**
+   * Gets the number of instances of a service handler.
```
ParthGandhi: I think now that they're separated we should just refer to them as `Service` and not "service handler". This is a general comment for all javadocs/documentation.

The reason is that you can't set instances for individual handlers, but only for the service as a whole. Also, now that workers have been moved out we don't need to distinguish the types.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1079,6 +1079,18 @@ public void getWorkflowSchedules(HttpRequest request, HttpResponder responder,
   }
 
   /**
+   * Return the number of instances of a service handler.
```
ParthGandhi: Same here.

ParthGandhi:
Just one comment. Otherwise LGTM.

gokulavasan:
@ParthGandhi Addressed your comments. Will merge after build passes.

PR #1423, opened 2015-02-25
move EntityTable class to right place

abaranau:
that data package was old leftover. Removed it.


abaranau:
http://builds.cask.co/browse/CDAP-DUT913-1

abaranau:
GREEN

gsps1:
LGTM :+1:

PR #1424, opened 2015-02-25
(CDAP-1204) Refactor queue producer/consumer to abstract out the row key strategy

chtyim:
- Also includes addition of scan timeout logic to minimize chance to have transaction timeout due to long scan


cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueScanner.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  public ImmutablePair<byte[], Map<byte[], byte[]>> next() throws IOException {
+    while (true) {
+      if (cached.size() > 0) {
+        Result result = cached.removeFirst();
+        Map<byte[], byte[]> row = result.getFamilyMap(QueueEntryRow.COLUMN_FAMILY);
+        return ImmutablePair.of(rowKeyConverter.apply(result.getRow()), row);
+      }
+      Result[] results = scanner.next(numRows);
```
ghelmling: This may trigger multiple RPCs if scanner caching < numRows.  Is this what we want?  Would it be better to call setCaching(numRows) on the scan instance?

ghelmling: I realize this is just extracting an inner class to top level, but seems worth asking.

chtyim: Yeah, it just an extraction of inner class. The creation of the `Scan` object takes care of the cache size (which is inside on of those `HBase9xQueueConsumer`, not changed in this PR). See

https://github.com/caskdata/cdap/blob/develop/cdap-hbase-compat-0.98/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBase98QueueConsumer.java#L49

chtyim:
Build pass: http://builds.cask.co/browse/CDAP-DUT914-3

ghelmling:
lgtm

chtyim:
Thanks @ghelmling . Merging.

PR #1425, opened 2015-02-25
util to figure out cdap version of metric's hbase table

abaranau:
https://issues.cask.co/browse/CDAP-1522


abaranau:
http://builds.cask.co/browse/CDAP-DUT915-1

abaranau:
GREEN

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTableAdmin.java
```diff
@@ -81,6 +81,7 @@ public void create() throws IOException {
 
 
     final HTableDescriptor tableDescriptor = new HTableDescriptor(name);
+    setVersion(tableDescriptor);
```
abaranau: yes, we were not adding cdap version property to tables up until now.. we only did on table upgrade

ghelmling:
lgtm

PR #1426, opened 2015-02-25
Add exceptions to the Scheduler interface, and add the creation of schedules to the app deployment pipeline

jguery:
The deployment of an app has been tested manually, as well as the updating of a schedule.

Build: http://builds.cask.co/browse/CDAP-DUT916-1

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
@@ -20,9 +20,6 @@ import co.cask.cdap.api.ProgramSpecification;
 import co.cask.cdap.api.flow.FlowSpecification;
 import co.cask.cdap.api.flow.FlowletConnection;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
-import co.cask.cdap.api.schedule.Schedule;
-import co.cask.cdap.api.schedule.ScheduleSpecification;
-import co.cask.cdap.api.workflow.ScheduleProgramInfo;
 import co.cask.cdap.api.workflow.WorkflowSpecification;
 import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.deploy.Manager;
@@ -36,6 +33,7 @@ import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.discovery.RandomEndpointStrategy;
 import co.cask.cdap.common.exception.AdapterNotFoundException;
+import co.cask.cdap.common.exception.NotFoundException;
 import co.cask.cdap.common.http.AbstractBodyConsumer;
 import co.cask.cdap.common.utils.DirUtils;
 import co.cask.cdap.config.PreferencesStore;
@@ -55,6 +53,7 @@ import co.cask.cdap.internal.app.runtime.adapter.AdapterTypeInfo;
 import co.cask.cdap.internal.app.runtime.adapter.InvalidAdapterOperationException;
 import co.cask.cdap.internal.app.runtime.flow.FlowUtils;
 import co.cask.cdap.internal.app.runtime.schedule.Scheduler;
+import co.cask.cdap.internal.app.runtime.schedule.SchedulerException;
 import co.cask.cdap.proto.AdapterConfig;
 import co.cask.cdap.proto.AdapterSpecification;
 import co.cask.cdap.proto.ApplicationRecord;
@@ -341,10 +340,12 @@ public class AppLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
         return;
       }
       responder.sendStatus(HttpResponseStatus.OK);
-    } catch (AdapterNotFoundException e) {
+    } catch (NotFoundException e) {
       responder.sendString(HttpResponseStatus.NOT_FOUND, e.getMessage());
     } catch (InvalidAdapterOperationException e) {
       responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
+    } catch (SchedulerException e) {
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
     }
   }
 
@@ -374,8 +375,10 @@ public class AppLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     try {
       adapterService.removeAdapter(namespaceId, adapterName);
       responder.sendStatus(HttpResponseStatus.OK);
-    } catch (AdapterNotFoundException e) {
+    } catch (NotFoundException e) {
       responder.sendString(HttpResponseStatus.NOT_FOUND, e.getMessage());
+    } catch (SchedulerException e) {
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
     }
   }
 
@@ -509,10 +512,7 @@ public class AppLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
         }
       });
 
-      ApplicationWithPrograms applicationWithPrograms =
-        manager.deploy(id, appId, deploymentInfo).get();
-      ApplicationSpecification specification = applicationWithPrograms.getSpecification();
-      setupSchedules(namespaceId, specification);
+      manager.deploy(id, appId, deploymentInfo).get();
     } catch (Throwable e) {
       LOG.warn(e.getMessage(), e);
       throw new Exception(e.getMessage());
@@ -541,30 +541,8 @@ public class AppLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
       }
     } catch (InterruptedException e) {
       throw new ExecutionException(e);
-    }
-  }
-
-  private void deleteSchedules(String namespaceId, ApplicationSpecification specification) throws IOException {
-    // Delete the existing schedules.
-    for (Map.Entry<String, ScheduleSpecification> entry : specification.getSchedules().entrySet()) {
-      ScheduleProgramInfo programInfo = entry.getValue().getProgram();
-      Id.Program programId = Id.Program.from(namespaceId, specification.getName(), programInfo.getProgramName());
-      scheduler.deleteSchedules(programId, programInfo.getProgramType());
-    }
-  }
-
-  private void setupSchedules(String namespaceId, ApplicationSpecification specification) throws IOException {
-
-    deleteSchedules(namespaceId, specification);
-
-    // Add new schedules.
-    for (Map.Entry<String, ScheduleSpecification> entry : specification.getSchedules().entrySet()) {
-      ScheduleProgramInfo programInfo = entry.getValue().getProgram();
-      Id.Program programId = Id.Program.from(namespaceId, specification.getName(),
-                                             programInfo.getProgramName());
-      List<Schedule> scheduleList = Lists.newArrayList();
-      scheduleList.add(entry.getValue().getSchedule());
-      scheduler.schedule(programId, programInfo.getProgramType(), scheduleList);
+    } catch (SchedulerException e) {
+      throw new ExecutionException(e);
     }
   }
 
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
@@ -345,7 +345,12 @@ public class AppLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     } catch (InvalidAdapterOperationException e) {
       responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
     } catch (SchedulerException e) {
-      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
+      LOG.error("Scheduler error in namespace '{}' for adapter '{}' with action '{}'",
+                namespaceId, adapterId, action, e);
+      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+    } catch (Throwable t) {
+      LOG.error("Error in namespace '{}' for adapter '{}' with action '{}'", namespaceId, adapterId, action, t);
+      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
     }
   }
 
@@ -378,7 +383,13 @@ public class AppLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     } catch (NotFoundException e) {
       responder.sendString(HttpResponseStatus.NOT_FOUND, e.getMessage());
     } catch (SchedulerException e) {
-      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
+      LOG.error("Scheduler error in namespace '{}' for adapter '{}' with action '{}'",
+                namespaceId, adapterName, "delete", e);
+      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+    } catch (Throwable t) {
+      LOG.error("Error in namespace '{}' for adapter '{}' with action '{}'",
+                namespaceId, adapterName, "delete", t);
+      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
     }
   }
 
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -35,6 +35,7 @@ import co.cask.cdap.app.store.StoreFactory;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.discovery.RandomEndpointStrategy;
+import co.cask.cdap.common.exception.NotFoundException;
 import co.cask.cdap.config.PreferencesStore;
 import co.cask.cdap.data2.transaction.queue.QueueAdmin;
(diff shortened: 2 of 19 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -341,10 +340,12 @@ public void startStopAdapter(HttpRequest request, HttpResponder responder,
         return;
       }
       responder.sendStatus(HttpResponseStatus.OK);
-    } catch (AdapterNotFoundException e) {
+    } catch (NotFoundException e) {
       responder.sendString(HttpResponseStatus.NOT_FOUND, e.getMessage());
     } catch (InvalidAdapterOperationException e) {
       responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
+    } catch (SchedulerException e) {
```
anew: we have a conention not to include the exception in HTTP responses for internal errors - because they reveal internal implementation details. 

but we should always log internal errors. For not found, that is a user error and kind of expected to happen. Logging it as an error would be very verbose so we would at most log it at debug level. But internal errors are no expected to happen and should be logged as errors, along with the context in which they happen (the app name, for example, or the URL of the request, which gives a lot of context).

I know there is a lot of existing code that does not do this right. Let's do it right here.

anew: I am also wondering what we do with other exceptions? Runtime exceptions etc. Shouldn't they also be treated and logged as internal errors?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -359,6 +360,8 @@ private void suspendResumeSchedule(HttpResponder responder, String namespaceId,
       }
     } catch (SecurityException e) {
       responder.sendStatus(HttpResponseStatus.UNAUTHORIZED);
+    } catch (NotFoundException e) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND, e.getMessage());
```
anew: this is almost good. (below). For unexpected Throwables: Log as error, respond with internal error. The only thing missing is: what schedule was it and did we try to suspend it or resume it?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalManager.java
```diff
@@ -107,6 +111,7 @@ public LocalManager(CConfiguration configuration, PipelineFactory pipelineFactor
     pipeline.addLast(new DeployDatasetModulesStage(datasetFramework));
     pipeline.addLast(new CreateDatasetInstancesStage(datasetFramework));
     pipeline.addLast(new CreateStreamsStage(id, streamAdmin, exploreFacade, exploreEnabled));
+    pipeline.addLast(new CreateSchedulesStage(store, scheduler));
```
anew: shouldn't it happen after the app is registered? Or at least after all programs are.

anew: Another thing that is tricky is that a schedule might fire right away... before the app and its programs (that is, the schedulable actions) are registered. In that case, the action would fail. That is why I feel that this must be the last stage of deployment.

jguery: We need to still have the app specification of the old app in the store, so that we can compare which schedules are new, and which schedules are to be updated.

jguery: To answer your second comment, if I put the stage before the ``ApplicationRegistrationStage``, but after the ``ProgramGenerationStage``, then the programs are registered and things should work, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/VerificationStage.java
```diff
@@ -169,6 +170,19 @@ public void process(ApplicationDeployable input) throws Exception {
       }
     }
 
+    // TODO StreamSizeSchedules should be resilient to stream inexistence [CDAP-1446]
+    for (Map.Entry<String, ScheduleSpecification> entry : specification.getSchedules().entrySet()) {
+      Schedule schedule = entry.getValue().getSchedule();
+      if (schedule instanceof StreamSizeSchedule) {
+        StreamSizeSchedule streamSizeSchedule = (StreamSizeSchedule) schedule;
+        String streamName = streamSizeSchedule.getStreamName();
+        if (!specification.getStreams().containsKey(streamName)) {
```
anew: Is this accurate? I am not sure what is allowed... I see two valid cases:

1. this app creates the stream, then it is in the spec of this app
2. another app created the stream, and it already exists. In this case, it would not necessarily be in the app spec. You would have to check whether the stream already exists.

jguery: Ok, I thought we said this morning that we could go with the assumption that the stream had to be in the app specification for now, and that we would allow the stream to not exist, etc, later.
But I can make the extra check to the store to make sure that the stream exists.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
```diff
@@ -72,8 +72,8 @@ protected final void startScheduler() {
     try {
       timeScheduler.start();
       LOG.info("Started time scheduler");
-    } catch (SchedulerException e) {
-      LOG.error("Error starting time scheduler {}", e.getCause(), e);
+    } catch (org.quartz.SchedulerException e) {
```
anew: why was this changed? By the way, should we have cdap SchedulerException that can be thrown by all schedulers? That would mean the time scheduler would have to catch quartz exceptions and wrap them into our SchedulerException. But it would be nice because it decouples the abstract scheduler service from the quartz implementation of time scheduler.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
```diff
@@ -81,7 +81,7 @@ protected final void startScheduler() {
       streamSizeScheduler.start();
       LOG.info("Started stream size scheduler");
     } catch (Throwable t) {
-      LOG.error("Error starting stream size scheduler {}", t.getCause(), t);
+      LOG.error("Error starting stream size scheduler", t);
```
anew: Interesting here: you already catch all Throwables. Here could wrap it into a SchedulerException instead of a RuntimeException.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
```diff
@@ -92,23 +92,24 @@ protected final void startScheduler() {
...
+        LOG.error("Error stopping time scheduler", e);
         throw Throwables.propagate(e);
       }
     }
   }
 
   @Override
-  public void schedule(Id.Program programId, SchedulableProgramType programType, Schedule schedule) {
+  public void schedule(Id.Program programId, SchedulableProgramType programType, Schedule schedule)
+    throws SchedulerException {
```
anew: Oh we have a SchedulerException now. My bad, it is because of the order in which I read these changes.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
```diff
@@ -53,8 +58,10 @@
    * @param programType type of program.
    * @return list of Scheduled runtimes for the program. Empty list if there are no schedules
    *         or if the program is not found
+   * @throws SchedulerException on unforeseen error.
    */
-  public List<ScheduledRuntime> nextScheduledRuntime(Id.Program program, SchedulableProgramType programType);
+  public List<ScheduledRuntime> nextScheduledRuntime(Id.Program program, SchedulableProgramType programType)
+    throws SchedulerException;
```
anew: wouldn't this also throw a notfound exception if the program is not found?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
```diff
@@ -80,16 +93,37 @@
...
+  public void resumeSchedule(Id.Program program, SchedulableProgramType programType, String scheduleName)
+    throws NotFoundException, SchedulerException;
+
+  /**
+   * Update the given schedule. The schedule with the same name than the given {@code schedule} will be replaced.
+   *
+   * @param program the program for which schedule needs to be updated
+   * @param programType the type of the program
+   * @param schedule the new schedule. The schedule with the same name will be replaced
+   * @throws NotFoundException if the {@code schedule} does not exist, or if the application the {@code program}
```
anew: what if the program does not exist? Or is that validated in the deployment stage?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -67,25 +66,27 @@
     this.preferencesStore = preferencesStore;
   }
 
-  void start() throws SchedulerException {
+  void start() throws org.quartz.SchedulerException {
```
anew: would be better if  this throws a CDAP scheduler exception

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -147,35 +150,45 @@ public void schedule(Id.Program programId, SchedulableProgramType programType, I
...
-    } catch (SchedulerException e) {
-      throw Throwables.propagate(e);
+    } catch (org.quartz.SchedulerException e) {
+      throw new SchedulerException(e);
     }
   }
 
   @Override
-  public void deleteSchedule(Id.Program program, SchedulableProgramType programType, String scheduleName) {
+  public void updateSchedule(Id.Program program, SchedulableProgramType programType, Schedule schedule)
```
anew: you will change this to update the schedule without deleting, in the next PR(s)? To make sure we do not miss a scheduled run because deleting a schedule also deletes its last run statistics.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/deploy/LocalManagerTest.java
```diff
@@ -78,6 +78,7 @@ public void testImproperOrNoManifestFile() throws Exception {
    */
   @Test
   public void testGoodPipeline() throws Exception {
+    // TODO modify
```
anew: next PR?

anew:
@jguery I like the overall approach. Please see my detailed comments.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/deploy/LocalManagerTest.java
jguery: Oops, I forgot to remove that.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
jguery: Okay, will do.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
jguery: I thought the DefaultAppConfigurer validated that, but I checked and it doesn't, so I will make that extra validation in the deployment stage too.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
jguery: True, will make the change.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
jguery: I will do that for the StreamSizeSchedules, but for the TimeSchedules is should be necessary, right? I think the old logic, of deleting all schedules and recreating them, works for the time schedules.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
jguery: I will make those changes, and also log the other runtime exceptions and return internal errors.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
jguery: Ok my bad, the validation stage already checks if the program exists.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
jguery: Ok will add those things.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
jguery: We can say that in the API, but the implementation actually doesn't check if the program exists or not, we only need to make sure that the app exists, so that we can retrieve the app spec which contains the schedule. Do you think we should also check if the program exists? 

Note that it's also we the endpoint that hits the resume/suspend does, it only checks that the app exists and that the schedule exists in the store, not the program.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalManager.java
anew: I think this is a weakness in the way that the deploy pipeline is handled. Or perhaps not. Because every stage can pass to the next stage whatever it needs to. So one could have a stage that takes a snapshot of the existing state (app spec, schedules, whether they are active, ...) and pass it into the pipeline. Then every stage has to pass it on from there on. When you reach the schedule stage, that old information would be in the pipeline, whereas the new app spec is already in the store (or also in the pipeline). To be honest, I am not sure whether the pipeline is designed well. Feels like every stage should be able to add information to the state that is passed between stages, without having to change the next stage. So perhaps what is passed on should be a map (key->Object), and every stage only extracts the things from the map that it needs, and can add new things that downstream stages might need.

anew: what is the stage that deletes the programs that are not in the new app spec, but were in the old one? Before that point (or in the same stage) we also have to remove the schedules for those programs, right? And after that we can add the new programs and the schedules for those?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/VerificationStage.java
anew: Sorry I guess I was not clear. The third case (which we won't support for now) is where another app creates the stream in the future, and the schedule is inactive until then.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
anew: I looked it up in the verification stage, we actually ensure that the scheduled programs exist in the app. So I think we do not need to check it here. Perhaps add a comment somewhere that states this.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
anew: correct unless a time schedule would fire exactly between those two times. Unlikely but possible.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
jguery: Ok will do.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
jguery: Jira created for this issue: https://issues.cask.co/browse/CDAP-1618

anew:
This looks very good now. I have only one comment. The fix for that could be more work... I think to avoid all possible race conditions the deployment should to this or similar:

1. identify what programs were removed 
2. delete the schedules for those programs
3. delete those programs
3. suspend the schedules for all programs that changed
4. update the programs that changed
5. update any schedules that changed for existing programs
6. resume the schedules that were changed
7. add the programs that are new
8. add the schedules for the new programs

This for sure deserves its own Jira. So if you can just make sure that schedules removed from the app are deleted from the scheduler, then this PR is good to go

jguery:
@anew Addressed your comment and created Jira https://issues.cask.co/browse/CDAP-1620.

anew:
If you want, address my one comment. If not, LGTM.

anew:
LGTM. Nice work!

PR #1429, opened 2015-02-26
CDAP-1258: Allow arguments to be passed on startup to CLI

jwang47:
Instead of reading from environment variables, the CLI now accepts runtime options:

```
usage: CDAP CLI
usage: CDAP CLI
 -a,--autoconnect <arg>   If "true", try provided connection (from uri)
                          upon launch or try default connection if none
                          provided. Defaults to "true".
 -d,--debug               Print all exception stack traces.
 -h,--help                Prints the usage message.
 -s,--verify-ssl <arg>    If "true", verify SSL certificate when making
                          requests. Defaults to "true".
 -u,--uri <arg>           CDAP instance URI to interact with in the format
                          "[http[s]://]<hostname>[:<port>[/<namespace>]]".
                          Defaults to "http://127.0.0.1:10000/default".
```

* Includes some refactoring and bug fixes.
* Fixes the bug described in https://issues.cask.co/browse/CDAP-1517

TODO
- [ ] In debug mode, when there's an exception from a *Client call, log the request and response.
- [ ] Option to run a script containing CLI commands
- [ ] Documentation


jwang47:
http://builds.cask.co/browse/CDAP-DUT917

jwang47:
This should also fix the bug described in https://issues.cask.co/browse/CDAP-1517

awholegunch:
I think that with the usage, if you are going to have some options in quotes, then they should all be. What I mean is:

```
If "true", try provided...
```

The URI usage has an unfortunate line break. But are the < > around "http" required? So then it would be:
```
"[http[s]://]<hostname>[:<port>[/<namespace>]]".
```
The first line above it could be:
```
CDAP instance URI to interact with in the format
```

jwang47:
Thanks for the catch, updated the PR.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -308,21 +291,33 @@ public void addHostnameChangeListener(ConnectionChangeListener listener) {
...
    * Connection information to a CDAP instance.
    */
   public static final class ConnectionInfo {
+
     private final String hostname;
     private final int port;
     private final boolean sslEnabled;
+    private final Id.Namespace namespace;
 
-    public ConnectionInfo(String hostname, int port, boolean sslEnabled) {
```
bdmogal: Should probably also have a constructor here that does not accept ``Id.Namespace`` and delegates to this constructor using ``Constants.DEFAULT_NAMESPACE_ID``?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -49,25 +61,73 @@
...
+   * @param output output to print to
+   * @param namespace the CDAP namespace to use initially
+   * @param uri provided URI of CDAP instance
+   * @param autoconnect if true, try provided connection (or default from CConfiguration) before startup
+   * @param verbose if true, log all exception stack traces
+   * @throws URISyntaxException
+   * @throws IOException
+   */
+  @Inject
+  public CLIMain(Injector injector, PrintStream output,
```
bdmogal: Injecting an ``Injector`` doesn't seem like a common pattern, right? Was there a problem with the old method of using ``Guice.createInjector``?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -92,22 +151,40 @@ public boolean handleException(PrintStream output, Exception e, int timesRetried
           output.println("Error: " + e.getMessage());
         }
 
+        if (verbose) {
+          e.printStackTrace(output);
```
bdmogal: This doesn't look right. A user shouldn't see stack traces right? *verbose* should be for seeing verbose output messages. But stack traces is a more 'debug' thing?

cdap-cli/src/main/java/co/cask/cdap/cli/util/InstanceURIParser.java
```diff
@@ -0,0 +1,62 @@
...
+import co.cask.cdap.common.conf.CConfiguration;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.proto.Id;
+import com.google.inject.Inject;
+
+import java.net.URI;
+
+/**
+ */
+public class InstanceURIParser {
```
bdmogal: Does it seem like an overkill to have this class? Can the parse method just be ``CLIConfig.ConnectionInfo.from(String uriString)``?

bdmogal:
A few minor comments. Also, https://issues.cask.co/browse/CDAP-1517 is still not resolved, right? The issues mentioned in the comments, etc?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
jwang47: True, I'll change `verbose` to `debug`

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
jwang47: Don't really need to inject `Injector`, but I made `CLIMain` injected b/c it looked cleaner.

jwang47:
Right, this doesn't handle the user deleting the current namespace. Created issue for that: https://issues.cask.co/browse/CDAP-1629

cdap-cli/src/main/java/co/cask/cdap/cli/util/InstanceURIParser.java
jwang47: True, will fix.

jwang47: Actually it requires `CConfiguration`, so putting it into `ConnectionInfo` isn't convenient.

jwang47:
Going to merge this. Resolving the TODOs in another PR.

PR #1430, opened 2015-02-26
Fix CDAP-1619

anwar6953:
Using the toString of Id.Stream object in file path is not good. Also, streamId is not necessary in the path in this case as the directory it is passed is already namespaced for the stream it is writing to.

Alternative: namespace (for a particular stream) the eventFile/indexFile in this class (instead of namespacing the directory and then passing it in - from FileContentWriterFactory).

Resolves: https://issues.cask.co/browse/CDAP-1619
Build: http://builds.cask.co/browse/CDAP-RBT133-1

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/FileContentWriter.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/FileContentWriter.java
@@ -54,8 +54,8 @@ final class FileContentWriter implements ContentWriter {
     this.streamEvent = new MutableStreamEvent();
 
     directory.mkdirs();
-    this.eventFile = directory.append(String.format("upload.%s.dat", streamConfig.getStreamId()));
-    this.indexFile = directory.append(String.format("upload.%s.idx", streamConfig.getStreamId()));
+    this.eventFile = directory.append("upload.dat");
+    this.indexFile = directory.append("upload.idx");
 
     Map<String, String> properties = createStreamFileProperties(headers);
     properties.put(StreamDataFileConstants.Property.Key.UNI_TIMESTAMP,


nitinmotgi:
LGTM :+1:

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/FileContentWriter.java
```diff
@@ -54,8 +54,8 @@
     this.streamEvent = new MutableStreamEvent();
 
     directory.mkdirs();
-    this.eventFile = directory.append(String.format("upload.%s.dat", streamConfig.getStreamId()));
-    this.indexFile = directory.append(String.format("upload.%s.idx", streamConfig.getStreamId()));
+    this.eventFile = directory.append("upload.dat");
```
chtyim: I think it's better keeping the stream id in it. Can you use `streamConfig.getStreamId().toId()`?

chtyim: Or just the `getName()` part if the directory is already within a namepsaced directory.

chtyim: The reason for keep the name in the file name is because the upload directory is shared across all streams (within the same namespace?). Having the stream name helps debugging.

anwar6953: ok, the directory thats passed in already is in the streamLocation (which has namespace + streamName in it).

chtyim: That's not the same dir being passed in. It is the temp directory that is relative to the stream base location

chtyim: hum, looks like the tmp is relative to the stream base location based on the stream config. So that's ok to just use a fixed name for the tmp files.

anwar6953: so LGTY im assuming

chtyim: Lgtm

PR #1433, opened 2015-02-26
Fix/cdap ui postmerge

gaarf:
* nail d3 version so that bower doesn't ask on install
 * update package.json and README to reflect new setup

### DIFF
--- a/cdap-ui/package.json
+++ b/cdap-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cdap-ui",
-  "version": "0.2.8",
+  "version": "2.8.0",
   "description": "Front-end for CDAP",
   "scripts": {
     "start": "node ./server.js",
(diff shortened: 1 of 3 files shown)


cdap-ui/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "cdap-ui",
-  "version": "0.1.3",
+  "version": "0.2.8",
```
nistala: cdap-master is using 2.8.0, @sreevatsanraman should this be the same as cdap version?

nistala:
minor comment, otherwise LGTM

elfenheart:
On the README maybe we want to remove the **npm test** since we don't have a complete test suite yet?

cdap-ui/package.json
gaarf: as of now this version number in package.json does nothing, apart from being shown in the console when the app boots.

gaarf:
@elfenheart why remove an easy way to run the little tests that we do have? FWIW `npm test` will only run karma unit tests, there is no protractor end-to-end anymore.

elfenheart:
:+1: LGTM

ajainarayanan:
:+1: LGTM

PR #1435, opened 2015-02-26
[CDAP-1519] Namespaced Filesets

bdmogal:
This PR namespaces filesets under ``/cdap/[ns-id]/data`` directory.

This also suffers from the drawback mentioned in [CDAP-1593](https://issues.cask.co/browse/CDAP-1593)

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/DefaultDatasetNamespace.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/DefaultDatasetNamespace.java
@@ -61,16 +61,16 @@ public class DefaultDatasetNamespace implements DatasetNamespace {
     int rootIndex = rootPrefix.length();
     // namespaceIndex is the index of the first dot after the rootIndex
     int namespaceIndex = namespaced.indexOf(".", rootIndex);
-    String expectedFormatMsg = "Dataset name is expected to be in the format " + rootPrefix +
-      "<namespace>.<dataset-name>.";
     // This check implies also that namespace is non-empty
     // Also, '.' is not permitted in namespace name. So this should return the full namespace.
     Preconditions.checkArgument(namespaceIndex > rootIndex,
-                                expectedFormatMsg + " Found invalid format - " + namespaced);
+                                "Dataset name is expected to be in the format %s<namespace>.<dataset-name>. Found - %s",
+                                rootPrefix, namespaced);
     String namespace = namespaced.substring(rootIndex, namespaceIndex);
     String datasetName = namespaced.substring(namespaceIndex + 1);
     Preconditions.checkArgument(!datasetName.isEmpty(),
-                                expectedFormatMsg + " Found empty dataset name in " + namespaced);
+                                "Dataset name is expected to be in the format %s<namespace>.<dataset-name>. Found - %s",
+                                rootPrefix, namespaced);
     return Id.DatasetInstance.from(Id.Namespace.from(namespace), datasetName);
   }
 
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/DefaultDatasetNamespace.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/DefaultDatasetNamespace.java
@@ -55,12 +55,23 @@ public class DefaultDatasetNamespace implements DatasetNamespace {
   @Override
   public Id.DatasetInstance fromNamespaced(String namespaced) {
     Preconditions.checkArgument(namespaced != null, "Dataset name should not be null");
-    // Dataset name is of the format cdap.<namespace>.<dataset-name>
-    String invalidFormatError = String.format("Invalid format for dataset name '%s'. " +
-                                                "Expected - cdap.<namespace>.<dataset-name>", namespaced);
-    String [] parts = namespaced.split("\\.", 3);
-    Preconditions.checkArgument(parts.length == 3, invalidFormatError);
-    return Id.DatasetInstance.from(Id.Namespace.from(parts[1]), parts[2]);
+    // Dataset name is of the format <table-prefix>.<namespace>.<dataset-name>
+    Preconditions.checkArgument(namespaced.startsWith(rootPrefix), "Dataset name should start with " + rootPrefix);
+    // rootIndex is the index of the first character after the root prefix
+    int rootIndex = rootPrefix.length();
+    // namespaceIndex is the index of the first dot after the rootIndex
+    int namespaceIndex = namespaced.indexOf(".", rootIndex);
+    String expectedFormatMsg = "Dataset name is expected to be in the format " + rootPrefix +
+      "<namespace>.<dataset-name>.";
+    // This check implies also that namespace is non-empty
+    // Also, '.' is not permitted in namespace name. So this should return the full namespace.
+    Preconditions.checkArgument(namespaceIndex > rootIndex,
+                                expectedFormatMsg + " Found invalid format - " + namespaced);
+    String namespace = namespaced.substring(rootIndex, namespaceIndex);
+    String datasetName = namespaced.substring(namespaceIndex + 1);
+    Preconditions.checkArgument(!datasetName.isEmpty(),
+                                expectedFormatMsg + " Found empty dataset name in " + namespaced);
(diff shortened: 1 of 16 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/file/FileAdmin.java
```diff
@@ -32,9 +36,13 @@
 
   private final String basePath;
 
-  public FileAdmin(LocationFactory locationFactory, DatasetSpecification spec) {
+  public FileAdmin(CConfiguration cConf, LocationFactory locationFactory, DatasetSpecification spec) {
+    TableId tableId = TableId.from(spec.getName());
```
anew: TableId? Don't you have a DatasetId?

bdmogal: Due to the current drawback in the design mentioned in CDAP-1593, the ``DatasetAdmin`` and ``Dataset`` classes use the ``TableId`` class to find out the namespace that the table has to go in, and its table name (in HBase). It has the capability to handle backwards compatibility in table names (from 2.6, when there were no namespaces).

This will go away when ``FileAdmin`` can get an ``Id.DatasetInstance`` as another parameter, which is the work that I'm starting now.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/file/FileSetDataset.java
```diff
@@ -20,6 +20,9 @@
 import co.cask.cdap.api.dataset.lib.FileSet;
 import co.cask.cdap.api.dataset.lib.FileSetArguments;
 import co.cask.cdap.api.dataset.lib.FileSetProperties;
+import co.cask.cdap.common.conf.CConfiguration;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.data2.util.hbase.TableId;
```
anew: Hbase util? For file set?

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/FileSetTest.java
```diff
@@ -34,33 +40,56 @@
...
+    Assert.assertEquals(fileSet2NsDir.getName(), NAMESPACE2.getId());
+
+    OutputStream out = fileSet1.getOutputLocation().getOutputStream();
     out.write(42);
     out.close();
 
-    InputStream in = fileSet.getInputLocations().get(0).getInputStream();
+    InputStream in = fileSet1.getInputLocations().get(0).getInputStream();
     Assert.assertEquals(42, in.read());
     in.close();
```
anew: don't you want to assert that the two filesets don't see each other's files? Or that the locations are different?

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/FileSetTest.java
```diff
@@ -34,33 +40,56 @@
...
 public class FileSetTest extends AbstractDatasetTest {
 
-  static FileSet fileSet;
-  private static final Id.DatasetInstance testFileSetInstance = Id.DatasetInstance.from(NAMESPACE_ID, "testFileSet");
+  static FileSet fileSet1;
+  static FileSet fileSet2;
+  private static final Id.Namespace NAMESPACE2 = Id.Namespace.from("yourspace");
+  private static final DatasetNamespace dsNamespace = new DefaultDatasetNamespace(CConfiguration.create());
+  private static final Id.DatasetInstance testFileSetInstance1 =
+    dsNamespace.namespace(Id.DatasetInstance.from(NAMESPACE_ID, "testFileSet"));
```
anew: so this gets name-spaced twice?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/file/FileSetDataset.java
bdmogal: Yeah, that doesn't seem right. Let me think about it.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/FileSetTest.java
bdmogal: In a different namespace, though. I should probably call it ``NAMESPACE_ID2`` or ``OTHER_NAMESPACE_ID`` to make the distinction clearer.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/FileSetTest.java
bdmogal: Yes, I should. will add.

bdmogal: I will add tests to make sure that the resulting locations are different.

However, this test already verifies that the base location is a sub directory of each fileset's own namespace (implying that the two filesets are under 2 different parent directories).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/DefaultDatasetNamespace.java
```diff
@@ -52,9 +53,14 @@ public String namespace(Id.Namespace namespaceId, String suffix) {
   }
 
   @Override
-  @Nullable
-  public Id.DatasetInstance fromNamespaced(String datasetInstanceName) {
-    return fromNamespaced(Id.DatasetInstance.from(Constants.SYSTEM_NAMESPACE, datasetInstanceName));
+  public Id.DatasetInstance fromNamespaced(String namespaced) {
+    Preconditions.checkArgument(namespaced != null, "Dataset name should not be null");
+    // Dataset name is of the format cdap.<namespace>.<dataset-name>
+    String invalidFormatError = String.format("Invalid format for dataset name '%s'. " +
```
anew: Not a good idea to use String.format, that means you always pay for formatting the string, but you only need it in case of error.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/DefaultDatasetNamespace.java
```diff
@@ -52,9 +53,14 @@ public String namespace(Id.Namespace namespaceId, String suffix) {
...
+                                                "Expected - cdap.<namespace>.<dataset-name>", namespaced);
```
anew: You code does not match your error message. If this is what you expect, then you can 
- verify that it starts with "cdap."
- find the next "."
- extract the namespace as the sequence between

anew:
I am not a big fan of instantiating the DefaultDatasetNamespace inside a dataset. I think namespace should be apply _to_ a dataset, not _by_ a dataset. But some of the namespacing logic will change again, when we add namespace awareness to the dataset APIs, right?

Other than that, I only have comments about one of the utility methods.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/DefaultDatasetNamespace.java
bdmogal: Will change.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/DefaultDatasetNamespace.java
bdmogal: I think the error message is wrong. We allow users to change the prefix by setting *dataset.table.prefix* to something else, so its not always guaranteed to start with "cdap.". So the error message should instead be 'Expected - <table-prefix>.<namespace>.<dataset-name>'

bdmogal:
I agree. Like you said, with the changes in the ``DatasetDefinition`` APIs that we discussed yesterday, we should no longer need to extract the namespace from the dataset name. So that change will be temporary. I want to get this change in, so the work of using the namespace id is done. In a future PR, the namespace Id will be received separately from the ``DatasetSpecification``. So ``DefaultDatasetNamespace`` should not be needed and can (potentially, I'm yet to ascertain the impact) be removed.

anew:
LGTM

PR #1439, opened 2015-02-27
fix non-tx delete increments (metrics tables)

abaranau:
Two issues fixed:
* in level db case, in non-tx, we should not write empty byte array delete tombstone
* in hbase case, in non-txt, when deleting a column, Delete.deleteColumnS method must be called to delete all versions

Just in case also added unit-test for transactional table.


abaranau:
http://builds.cask.co/browse/CDAP-DUT926-1

abaranau:
GREEN

gsps1:
LGTM :+1:

PR #1442, opened 2015-02-27
handling dots in metric tag names and values in rest api

abaranau:
https://issues.cask.co/browse/CDAP-1645

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
@@ -59,6 +59,7 @@ public class MetricsHandler extends AuthenticatedHttpHandler {
 
   public static final String ANY_TAG_VALUE = "*";
   public static final String TAG_DELIM = ".";
+  public static final String DOT_ESCAPE_CHAR = "~";
 
   private final MetricStore metricStore;
 
@@ -187,10 +188,23 @@ public class MetricsHandler extends AuthenticatedHttpHandler {
       if (ANY_TAG_VALUE.equals(val)) {
         val = null;
       }
-      result.put(name, val);
+      if (val != null) {
+        val = decodeTag(val);
+      }
+      result.put(decodeTag(name), val);
     }
 
     return result;
+  }
+
+  private String decodeTag(String val) {
+    // we use "." as separator between tags, so if tag name or value has dot in it, we encode it with "~"
+    return val.replaceAll(DOT_ESCAPE_CHAR, ".");
+  }
+
+  private String encodeTag(String val) {
+    // we use "." as separator between tags, so if tag name or value has dot in it, we encode it with "~"
+    return val.replaceAll("\\.", DOT_ESCAPE_CHAR);
   }
 
   private void searchMetricAndRespond(HttpResponder responder, String context) {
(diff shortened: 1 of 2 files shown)


abaranau:
http://builds.cask.co/browse/CDAP-DUT918-2

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -187,12 +188,25 @@ public void query(HttpRequest request, HttpResponder responder,
       if (ANY_TAG_VALUE.equals(val)) {
         val = null;
       }
-      result.put(name, val);
+      if (val != null) {
+        val = unEncodeTag(val);
```
gsps1: decodeTag

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -187,12 +188,25 @@ public void query(HttpRequest request, HttpResponder responder,
...
     return result;
   }
 
+  private String unEncodeTag(String val) {
+    // we use "." as separator between tags, so if tag name or value has dot in it, we encode it with "#"
+    return val.replaceAll(DOT_ESCAPED, ".");
+  }
+
+  private String encodeTag(String val) {
+    // we use "." as separator between tags, so if tag name or value has dot in it, we encode it with "#"
```
gsps1: ecode with a '~'

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -187,12 +188,25 @@ public void query(HttpRequest request, HttpResponder responder,
...
+        val = unEncodeTag(val);
+      }
+      result.put(unEncodeTag(name), val);
     }
 
     return result;
   }
 
+  private String unEncodeTag(String val) {
+    // we use "." as separator between tags, so if tag name or value has dot in it, we encode it with "#"
```
gsps1: so if tag name or value has '~' in it, we decode it to "."

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -59,6 +59,7 @@
 
   public static final String ANY_TAG_VALUE = "*";
   public static final String TAG_DELIM = ".";
+  public static final String DOT_ESCAPED = "~";
```
gsps1: DOT_ESCAPE_CHARACTER ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -282,7 +296,8 @@ private String toCanonicalContext(List<TagValue> tagValues) {
         sb.append(TAG_DELIM);
       }
       first = false;
-      sb.append(tv.getTagName()).append(TAG_DELIM).append(tv.getValue() == null ? ANY_TAG_VALUE : tv.getValue());
+      sb.append(encodeTag(tv.getTagName())).append(TAG_DELIM)
+        .append(tv.getValue() == null ? ANY_TAG_VALUE : encodeTag(tv.getValue()));
```
gsps1: may be we can move this logic to encodeTag ? if null , return ANY_TAG_VALUE there.

gsps1:
would it make sense to have an example before each endpoint?

gsps1:
just couple of minor comments, looks good otherwise.

abaranau:
GREEN

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
abaranau: true

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
abaranau: yes

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
abaranau: yes

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
abaranau: DOT_ESCAPE_CHAR

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
abaranau: I would not - it may be different for different requests, some may not accept it, etc.

abaranau:
> would it make sense to have an example before each endpoint?

I think they are pretty self descriptive - I would not put "~" in there. We'll put these in docs

PR #1446, opened 2015-02-28
Feature/metrics data migration

gsps1:
hooking up with Main method for metrics upgrade tool and determining version of CDAP to migrate data from


cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -71,6 +73,7 @@
    */
   private enum Action {
     UPGRADE("Upgrade all tables."),
+    MIGRATE_METRICS_DATA("Migrate metrics data to CDAP-2.8"),
```
abaranau: "Migrate metrics data".

Also, we want to say why this is a separate action in a comment: "Metrics migration is not required for CDAP to work after upgrade, some may opt to not do it as it may take a while"

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +203,54 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
     }
   }
 
+  private void migrateMetricsData(Injector injector) throws Exception {
+    // find version to migrate
+    ProjectInfo.Version version = findMetricsTableVersion(injector.getInstance(CConfiguration.class),
```
abaranau: print out smth like "Starting metrics data migration"

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +203,54 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
...
+                                                  injector.getInstance(Configuration.class));
+    if (version != null) {
+      DatasetFramework framework = createRegisteredDatasetFramework(injector);
+      // migrate metrics data
+      DefaultMetricDatasetFactory.migrateMetricsData(injector.getInstance(CConfiguration.class), framework, version);
+    }
+  }
+
+  private ProjectInfo.Version findMetricsTableVersion(CConfiguration cConf, Configuration hConf) {
+    // Upgrade all datasets in system namespace
```
abaranau: remove comment

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +203,54 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
...
+    boolean metricsTable27Found = false;
+    boolean metricsTable26Found = false;
```
abaranau: I'd add a comment with the logic, how we figure out the version. And I think it should be like this:
* figure out what is the latest working version of CDAP (we just talked with @bdmogal about that, the upgrade tool framework will give you any of 2.8.x or "earlier" currently)
* if latest is 2.8.x - nothing to do, if "pre-2.8", proceed to next step
* find a most recent metrics table: start by looking for 2.7, then 2.6 (I'd reformat the code to not look for 2.6 if we found 2.7)
* if you found 2.7 - try upgrade from there, if not - try upgrade from 2.6

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +203,54 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
...
+    ProjectInfo.Version version = findMetricsTableVersion(injector.getInstance(CConfiguration.class),
+                                                  injector.getInstance(Configuration.class));
+    if (version != null) {
+      DatasetFramework framework = createRegisteredDatasetFramework(injector);
+      // migrate metrics data
+      DefaultMetricDatasetFactory.migrateMetricsData(injector.getInstance(CConfiguration.class), framework, version);
+    }
+  }
+
+  private ProjectInfo.Version findMetricsTableVersion(CConfiguration cConf, Configuration hConf) {
```
abaranau: add @ Nullable, if it can return null

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricDatasetFactory.java
```diff
@@ -135,7 +137,7 @@ private MetricsTable getOrCreateMetricsTable(String tableName, DatasetProperties
    *
    * @param datasetFramework framework to add types and datasets to
    */
-  public static void setupDatasets(CConfiguration conf, DatasetFramework datasetFramework)
+  public static MetricDatasetFactory setupDatasets(CConfiguration conf, DatasetFramework datasetFramework)
```
abaranau: why return MetricDatasetFactory?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricDatasetFactory.java
```diff
@@ -148,6 +150,23 @@ public static void setupDatasets(CConfiguration conf, DatasetFramework datasetFr
 
     // adding kafka consumer meta
     factory.createKafkaConsumerMeta();
+    return factory;
+  }
+
+  /**
+   * Migrates metrics data from version 2.7 and older to 2.8
+   * @param conf
```
abaranau: what is with the params? In every PR you miss to describe them :)

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricDatasetFactory.java
```diff
@@ -148,6 +150,23 @@ public static void setupDatasets(CConfiguration conf, DatasetFramework datasetFr
...
+   * @param datasetFramework
+   * @param version - version we migrate the data from
+   * @throws IOException
+   * @throws DatasetManagementException
+   */
+  public static void migrateMetricsData(CConfiguration conf, DatasetFramework datasetFramework,
```
abaranau: migrateData is enough - we know we are in metrics

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/upgrade/MetricsDataMigrator.java
```diff
@@ -134,13 +140,15 @@ public MetricsDataMigrator(final CConfiguration cConf, final DatasetFramework ds
     aggMetricStore = new DefaultMetricStore(factory, new int[]{Integer.MAX_VALUE});
   }
 
-  public void migrateMetricsTables(Version cdapVersion) {
-    if (cdapVersion.getVersion().equals("2.6")) {
-      migrateMetricsTableFromVersion26(cdapVersion);
+  public void migrateMetricsTables(ProjectInfo.Version cdapVersion) {
```
abaranau: this looks odd. You first convert 2.6_or_older or 2.7 into Version (in tool Main), then do other way around... Should you just pass that enum value from Main?

I was also thinking - does it make sense to run the tool I wrote before on the metrics table that you found in Main - to verify version? Then you'd reuse that enum

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
abaranau: also: should you do this in the metrics migrate tool? Or, in the Metrics version tool?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/upgrade/MetricsDataMigrator.java
```diff
@@ -186,6 +194,7 @@ private void migrateMetricsData(EntityTable entityTable, MetricsTable metricsTab
         rowCount++;
         printStatus(rowCount);
       }
+      System.out.println("Migrated " + rowCount + " records...");
```
abaranau: why "..." in end?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/upgrade/MetricsDataMigrator.java
```diff
@@ -299,7 +308,7 @@ private MetricsTable getOrCreateMetricsTable(String tableName, DatasetProperties
 
   // todo : use Enum instead of string comparison for the below methods
   private String getMetricNameBasedOnVersion(String metricName, Version version) {
-    if (version.getVersion().equals("2.6")) {
+    if (version == Version.VERSION_26_OR_OLDER) {
       return metricName;
     } else {
```
abaranau: it is usually a good idea to verify the other enum value, and in the last else - fail with "unsupported version" thing

abaranau:
See some comments :)

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricDatasetFactory.java
gsps1: i pass that to MetricsDataMigrator

gsps1:
@abaranau addressed the comments. please review again.

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -71,6 +76,9 @@
    */
   private enum Action {
     UPGRADE("Upgrade all tables."),
+    // NOTE : Metrics migration is not required for CDAP to work after upgrade,
+    // some may opt not to migrate data as it may take a while.
```
abaranau: + "Hence, it is extracted as separate command/step"

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +208,76 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
     }
   }
 
+  private void migrateMetricsData(Injector injector) throws Exception {
+    System.out.println("Starting metrics data migration");
+    // find version to migrate
+    MetricHBaseTableUtil metricHBaseTableUtil = new MetricHBaseTableUtil(injector.getInstance(HBaseTableUtil.class));
+    Version version = findMetricsTableVersion(injector.getInstance(CConfiguration.class),
+                                                  injector.getInstance(Configuration.class), metricHBaseTableUtil);
```
abaranau: indent

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +208,76 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
...
+    if (version != null) {
+      DatasetFramework framework = createRegisteredDatasetFramework(injector);
+      // migrate metrics data
+      DefaultMetricDatasetFactory.migrateData(injector.getInstance(CConfiguration.class), framework, version);
+    }
```
abaranau: what if it is null? should we at least tell smth?

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +208,76 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
...
   }
 
+  private void migrateMetricsData(Injector injector) throws Exception {
+    System.out.println("Starting metrics data migration");
+    // find version to migrate
+    MetricHBaseTableUtil metricHBaseTableUtil = new MetricHBaseTableUtil(injector.getInstance(HBaseTableUtil.class));
+    Version version = findMetricsTableVersion(injector.getInstance(CConfiguration.class),
+                                                  injector.getInstance(Configuration.class), metricHBaseTableUtil);
+    if (version != null) {
+      DatasetFramework framework = createRegisteredDatasetFramework(injector);
```
abaranau: do you want to print out which version was detected?

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +208,76 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
...
+      // migrate metrics data
+      DefaultMetricDatasetFactory.migrateData(injector.getInstance(CConfiguration.class), framework, version);
+    }
+  }
+
+  @Nullable
+  private Version findMetricsTableVersion(CConfiguration cConf, Configuration hConf,
```
abaranau: didn't we say this should go out of this class?

cdap-master/src/main/java/co/cask/cdap/data/tools/Main.java
```diff
@@ -197,6 +208,76 @@ private void upgradeSystemDatasets(Injector injector) throws Exception {
...
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+    return version;
+  }
+
+  private Version verifyVersion(Version expected, Version actual) {
+    if (expected != actual) {
+      System.out.println("Unable to migrate, Version mismatch, Expected Version " + expected +
```
abaranau: shouldn't you say: "version detected based on table name does not match table configuration"?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricDatasetFactory.java
abaranau: can we get it from injector in Main? it looks hacky this way...

abaranau:
couple more comments

gsps1:
http://builds.cask.co/browse/CDAP-DUT909 - Green

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -0,0 +1,250 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.hadoop.hbase.HTableDescriptor;
+import org.apache.hadoop.hbase.client.HBaseAdmin;
+
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/**
+ * Command line tool.
```
abaranau: Command line tool to migrate data between different versions of CDAP. Usually used along with upgrade tool (see ...))

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -0,0 +1,250 @@
...
+ */
+public class DataMigration {
+  /**
+   * Set of Action available in this tool.
+   */
+  private enum Action {
+    // NOTE : Metrics migration is not required for CDAP to work after upgrade,
```
abaranau: remove this comment - no longer needed. We will say that in the installation docs

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -0,0 +1,250 @@
...
+   */
+  public static DatasetFramework createRegisteredDatasetFramework(Injector injector)
+    throws DatasetManagementException, IOException {
+    DatasetDefinitionRegistryFactory registryFactory = injector.getInstance(DatasetDefinitionRegistryFactory.class);
+    DatasetFramework datasetFramework =
+      new InMemoryDatasetFramework(registryFactory, injector.getInstance(CConfiguration.class));
+    // TODO: this doesn't sound right. find out why its needed.
+    datasetFramework.addModule(Id.DatasetModule.from(Constants.SYSTEM_NAMESPACE, "table"),
+                               new HBaseTableModule());
+    datasetFramework.addModule(Id.DatasetModule.from(Constants.SYSTEM_NAMESPACE, "metricsTable"),
```
abaranau: seems like you only need metricsTable...

abaranau: but if hard to figure out why - let's keep for now all and leave the todo - as you did

abaranau:
Real minor comments and let's merge

PR #1448, opened 2015-03-02
Writing to Stream through WorkerContext

gokulavasan:
Allowing Workers to write to Streams through WorkerContext.

* StreamContext allows Workers to write to streams - single events and in batch.
* LengthBasedContentWriter that uses BufferedContentWriter and then switches to FileBasedContentWriter once batchThreshold is crossed.

Bamboo: http://builds.cask.co/browse/CDAP-DUT857

JIRA: https://issues.cask.co/browse/CDAP-1454


cdap-app-fabric/src/main/java/co/cask/cdap/app/services/DefaultStreamBatchWriter.java
```diff
@@ -0,0 +1,56 @@
...
+    ByteBuffers.writeToStream(data, connection.getOutputStream());
+  }
+
+  @Override
+  public void close() throws IOException, StreamWriteException {
+    connection.getOutputStream().close();
+    int responseCode = connection.getResponseCode();
+    if (responseCode == HttpResponseStatus.NOT_FOUND.code()) {
+      throw new StreamWriteException(String.format("Stream %s not found", stream));
+    } else if (responseCode < 200 && responseCode >= 300) {
```
anwar6953: you mean `||`

anwar6953: Or the other way around `responseCode >= 200 && responseCode < 300`

cdap-app-fabric/src/main/java/co/cask/cdap/app/services/DefaultStreamBatchWriter.java
```diff
@@ -0,0 +1,56 @@
...
+import java.net.HttpURLConnection;
+import java.nio.ByteBuffer;
+
+/**
+ * Implementation of {@link StreamBatchWriter}.
+ */
+public class DefaultStreamBatchWriter implements StreamBatchWriter {
+
+  private final HttpURLConnection connection;
+  private final String stream;
```
anwar6953: I think you should store Id.Stream here. Better for debugging, since it more uniquely identifies the stream (it includes namespace).

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamBatchWriter.java
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Each call to write will write data to the stream batch endpoint.
+   *
+   * @param data {@link ByteBuffer}
+   * @throws IOException
+   */
+  void write(ByteBuffer data) throws IOException;
+
+  /**
+   * Method is used to finish the current set of batch writes.
```
ParthGandhi: What does "finish" mean? Close the connection?

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
```diff
@@ -0,0 +1,110 @@
...
+
+import co.cask.cdap.api.stream.StreamEventData;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Map;
+
+/**
+ * Stream Context.
```
ParthGandhi: Needs better javadoc since its in API.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public interface StreamContext {
```
ParthGandhi: Isn't "Context" usually per instance? This seems to work across all streams.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Write a File to a stream in batch
+   * @param stream stream id
+   * @param file File
+   * @param contentType content type
+   *
+   * @throws IOException if a network error occurred
+   * @throws StreamWriteException exception caused during writing to stream
+   */
+  public void writeInBatch(String stream, File file, String contentType) throws IOException, StreamWriteException;
```
ParthGandhi: `getBatchWriter`?

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamWriteException.java
```diff
@@ -0,0 +1,35 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.data.stream;
+
+/**
+ * Exception to indicate that stream was not found.
+ */
+public class StreamWriteException extends Exception {
```
ParthGandhi: Theres already a `StreamNotFoundException` in `common`. Move that to `api` and remove this class.

cdap-app-fabric/src/main/java/co/cask/cdap/app/services/AbstractServiceDiscoverer.java
```diff
@@ -37,7 +53,7 @@
  * It provides definition for {@link ServiceDiscoverer#getServiceURL}  and expects the sub-classes to give definition
  * for {@link AbstractServiceDiscoverer#getDiscoveryServiceClient}.
  */
-public abstract class AbstractServiceDiscoverer implements ServiceDiscoverer {
+public abstract class AbstractServiceDiscoverer implements ServiceDiscoverer, StreamContext {
```
ParthGandhi: Name is incorrect now..

ParthGandhi: Actually... this gives the implementation to *all* program contexts (eg. `BasicFlowletContext`) but doesn't expose them through the interface. So technically I could do: `((StreamContext) getContext()).write("stream", "stream");` inside a flowlet process method and it would work. Do we want to allow that? If yes, then all the program contexts should extend `StreamContext`. If not, then the implementation should move into its own abstract class.

(Btw, why are worker's special when it comes to writing to streams? If we're exposing this capability shouldn't we expose it uniformly for all programs?)

cdap-app-fabric/src/main/java/co/cask/cdap/app/services/AbstractServiceDiscoverer.java
```diff
@@ -85,4 +101,85 @@ private URL createURL(@Nullable Discoverable discoverable, String applicationId,
...
+      } catch (MalformedURLException e) {
+        LOG.error("Got exception while creating StreamURL", e);
+      }
+    }
+    throw new StreamWriteException("Stream Service Endpoint not found");
+  }
+
+  private void writeToStream(String stream, HttpRequest request) throws IOException, StreamWriteException {
+    HttpResponse response = HttpRequests.execute(request);
+    if (response.getResponseCode() == HttpResponseStatus.NOT_FOUND.code()) {
```
ParthGandhi: Can this ever happen? Don't you already check for stream existence when getting the URL?

Also, maybe throw exception on all writes that aren't 200s? Otherwise you can have failing writes without knowing.

cdap-app-fabric/src/main/java/co/cask/cdap/app/services/AbstractServiceDiscoverer.java
```diff
@@ -85,4 +101,85 @@ private URL createURL(@Nullable Discoverable discoverable, String applicationId,
...
+      throw new StreamWriteException(String.format("Stream %s not found", stream));
+    }
+  }
+
+  @Override
+  public void write(String stream, String data) throws IOException, StreamWriteException {
+    write(stream, data, Maps.<String, String>newHashMap());
```
ParthGandhi: `ImmutableMap.<String, String>of()`

cdap-app-fabric/src/main/java/co/cask/cdap/app/services/AbstractServiceDiscoverer.java
```diff
@@ -85,4 +101,85 @@ private URL createURL(@Nullable Discoverable discoverable, String applicationId,
...
+    writeToStream(stream, request);
+  }
+
+  @Override
+  public StreamBatchWriter writeInBatch(String stream, String contentType) throws IOException, StreamWriteException {
+    URL url = getStreamURL(stream, true);
+    HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+    connection.setRequestMethod(HttpMethod.POST.name());
+    connection.setReadTimeout(15000);
+    connection.setConnectTimeout(15000);
```
ParthGandhi: So the user can use the `StreamBatchWriter` returned only for this time, right? We should document that.

cdap-client-tests/src/test/java/co/cask/cdap/client/app/AppWritingtoStream.java
```diff
@@ -0,0 +1,132 @@
...
+      setName(WORKER);
+    }
+
+    @Override
+    public void run() {
+      for (int i = 0; i < VALUE; i++) {
+        try {
+          getContext().write(STREAM, String.format("Event %d", i));
+        } catch (IOException e) {
+          e.printStackTrace();
```
ParthGandhi: `LOG.error()`

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
```diff
@@ -0,0 +1,101 @@
...
+    fileContentWriter = null;
+  }
+
+  @Override
+  public void append(ByteBuffer body, boolean immutable) throws IOException {
+    if (fileContentWriter != null) {
+      fileContentWriter.append(body, immutable);
+    } else {
+      bufferedContentWriter.append(body, immutable);
+      updateWriter(body.array().length);
```
ParthGandhi: `.array()` is an optional method: http://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#array()

ParthGandhi:
A few comments, mostly around API. Please address.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
anwar6953: it doesn't return a writer though.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamBatchWriter.java
```diff
@@ -0,0 +1,42 @@
...
+   *
+   * @throws IOException
+   * @throws StreamWriteException
+   */
+  void close() throws IOException, StreamWriteException;
```
chtyim: Should extends from `Closeable` interface instead. Also, why just `IOException` is not enough?

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
ParthGandhi: Sorry, meant to write that for the next method.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Write a string to a stream
+   *
+   * @param stream stream id
+   * @param data data to write
+   *
+   * @throws IOException if a network error occurred
+   * @throws StreamWriteException exception caused during writing to stream
+   */
+  public void write(String stream, String data) throws IOException, StreamWriteException;
```
chtyim: I don't think we need to have two different exceptions while basically meaning the same thing. If you want to let user to distinguish exception that is stream specific (e.g. `StreamNotFoundException`), you can through those.

cdap-app-fabric/src/main/java/co/cask/cdap/app/services/AbstractServiceDiscoverer.java
anwar6953: @ParthGandhi In getStreamURL, it only checks that the Stream Service is up. Not that a particular stream exists.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
anwar6953: Bytes.toBytes(ByteBuffer) should do it

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
```diff
@@ -0,0 +1,110 @@
...
+   * Write a {@link ByteBuffer} to a stream
+   *
+   * @param stream stream id
+   * @param data {@link ByteBuffer}
+   * @param headers headers for the data
+   *
+   * @throws IOException if a network error occurred
+   * @throws StreamWriteException exception caused during writing to stream
+   */
+  public void write(String stream, ByteBuffer data, Map<String, String> headers)
```
chtyim: No need to have this. It's the same as `new StreamEventData(headers, data)`.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
chtyim: `writeFile`.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamContext.java
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Write in batch using {@link StreamBatchWriter} to a stream
+   * @param stream stream id
+   * @param contentType content type
+   * @return {@link StreamBatchWriter}
+   *
+   * @throws IOException if a network error occurred
+   * @throws StreamWriteException exception caused during writing to stream
+   */
+  public StreamBatchWriter writeInBatch(String stream, String contentType) throws IOException, StreamWriteException;
```
chtyim: `createBatchWriter`.

cdap-app-fabric/src/main/java/co/cask/cdap/app/services/AbstractServiceDiscoverer.java
```diff
@@ -85,4 +101,85 @@ private URL createURL(@Nullable Discoverable discoverable, String applicationId,
...
   }
+
+  private URL getStreamURL(String stream) throws StreamWriteException {
+    return getStreamURL(stream, false);
+  }
+
+  private URL getStreamURL(String stream, boolean batch) throws StreamWriteException {
+    ServiceDiscovered discovered = getDiscoveryServiceClient().discover(Constants.Service.STREAMS);
+    Discoverable discoverable = new RandomEndpointStrategy(discovered).pick(1, TimeUnit.SECONDS);
+    if (discoverable != null) {
```
chtyim: invert the logic to give better control flow:

```java
// Exception/error logic first
if (discoverable == null) {
  throw ...
}
// Normal logic here.
```

cdap-app-fabric/src/main/java/co/cask/cdap/app/services/DefaultStreamBatchWriter.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Override
+  public void close() throws IOException, StreamWriteException {
+    connection.getOutputStream().close();
+    int responseCode = connection.getResponseCode();
+    if (responseCode == HttpResponseStatus.NOT_FOUND.code()) {
+      throw new StreamWriteException(String.format("Stream %s not found", stream));
+    } else if (responseCode < 200 || responseCode >= 300) {
+      throw new StreamWriteException(String.format("Writing to Stream %s did not succeed", stream));
```
chtyim: Shouldn't it be throwing `IOException`?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
chtyim: Shouldn't use `array().length`. Should should `remaining()`.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
```diff
@@ -0,0 +1,101 @@
...
+  @Override
+  public void appendAll(Iterator<ByteBuffer> bodies, boolean immutable) throws IOException {
+    if (fileContentWriter != null) {
+      fileContentWriter.appendAll(bodies, immutable);
+    } else {
+      long tempSize = 0;
+      while (bodies.hasNext()) {
+        ByteBuffer next = bodies.next();
+        bufferedContentWriter.append(next, immutable);
+        tempSize += next.array().length;
```
chtyim: Same here.

cdap-client-tests/src/test/java/co/cask/cdap/client/ProgramClientTestRun.java
```diff
@@ -116,4 +128,26 @@ public void testAll() throws Exception {
     assertProgramStopped(programClient, FakeApp.NAME, ProgramType.FLOW, FakeFlow.NAME);
 
   }
+
+  @Test
+  public void testStreamWrite() throws Exception {
```
anwar6953: It seems like the test cases in .*ClientTestRun should focus on testing the clients, and not to test new features in cdap.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
chtyim: Also, you should get the remaining before the `append` call, not after.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
chtyim: should get the `body.remaining` before the append call.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
```diff
@@ -0,0 +1,101 @@
...
+ * file based on the length of streaming data.
+ */
+final class LengthBasedContentWriter implements ContentWriter {
+
+  private final long bufferThreshold;
+  private final ContentWriter bufferedContentWriter;
+  private final FileContentWriterFactory fileContentWriterFactory;
+
+  private ContentWriter fileContentWriter;
+  private long dataSoFar;
```
chtyim: better just call it `bodySize`.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
```diff
@@ -0,0 +1,101 @@
...
+    } else {
+      bufferedContentWriter.close();
+    }
+  }
+
+  private void updateWriter(long length) throws IOException {
+    dataSoFar += length;
+    if (dataSoFar >= bufferThreshold) {
+      bufferedContentWriter.close();
+      fileContentWriter = fileContentWriterFactory.create(ImmutableMap.<String, String>of());
```
chtyim: You cannot close the buffered writer, since that will commit the buffer to stream writer, meaning the file content will not be appeared atomically to stream read, which break the batch endpoint contract. What you need is a way to get the buffered content writes to the newly create file content writer (probably a package only method to get and Iterator of the buffered content).

gokulavasan:
@ParthGandhi @anwar6953 @chtyim Thanks for your comments. Addressed them. Please take a look again. I will add more tests to StreamWriterTestRun to test batch writes.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/AbstractStreamWriter.java
```diff
@@ -0,0 +1,148 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.ByteBuffer;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Abstract implementation of {@link StreamWriter}
+ */
+public abstract class AbstractStreamWriter extends AbstractServiceDiscoverer implements StreamWriter {
```
ParthGandhi: Why extend `AbstractServiceDiscoverer`?

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/AbstractStreamWriter.java
```diff
@@ -0,0 +1,148 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractServiceDiscoverer.class);
```
ParthGandhi: Incorrect Logger.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractContext.java
```diff
@@ -45,7 +45,7 @@
 /**
  * Base class for program runtime context
  */
-public abstract class AbstractContext extends AbstractServiceDiscoverer implements DatasetContext, RuntimeContext {
+public abstract class AbstractContext extends AbstractStreamWriter implements DatasetContext, RuntimeContext {
```
ParthGandhi: This doesn't fix the problem I'd commented earlier since it exposes the stream writing capabilities to all program contexts (which, as per demo, we don't want to allow). 

This class should continue extending `AbstractServiceDiscoverer` and *only* `BasicWorkerContext` should extend `AbstractStreamWriter`.

gokulavasan:
@chtyim @ParthGandhi Addressed your comment. Please review.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamBatchWriter.java
```diff
@@ -0,0 +1,44 @@
...
+   * @param data {@link ByteBuffer}
+   * @throws IOException
+   */
+  void write(ByteBuffer data) throws IOException;
+
+  /**
+   * Method is used to complete the current set of batch writes (needs to be invoked after the batch writes
+   * are complete).
+   *
+   * @throws IOException
```
chtyim: No doc?

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamBatchWriter.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Used to write to Stream in Batch mode. Once closed, a new {@link StreamBatchWriter} needs to be created to do
+ * another set of batch writes.
+ */
+public interface StreamBatchWriter extends Closeable {
+
+  /**
+   * Each call to write will write data to the stream batch endpoint.
+   *
+   * @param data {@link ByteBuffer}
```
chtyim: No doc?

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/AbstractStreamWriter.java
```diff
@@ -0,0 +1,160 @@
...
+import java.net.URL;
+import java.nio.ByteBuffer;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Abstract implementation of {@link StreamWriter}
+ */
+public abstract class AbstractStreamWriter extends AbstractContext implements StreamWriter {
```
chtyim: Why extends AbstractContext? The `StreamWriter` is not a RuntimeContext.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/AbstractStreamWriter.java
```diff
@@ -0,0 +1,160 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractStreamWriter.class);
+
+  protected String namespaceId;
```
chtyim: final

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/AbstractStreamWriter.java
```diff
@@ -0,0 +1,160 @@
...
+  private URL getStreamURL(String stream) throws IOException {
+    return getStreamURL(stream, false);
+  }
+
+  private URL getStreamURL(String stream, boolean batch) throws IOException {
+    ServiceDiscovered discovered = getDiscoveryServiceClient().discover(Constants.Service.STREAMS);
+    Discoverable discoverable = new RandomEndpointStrategy(discovered).pick(1, TimeUnit.SECONDS);
+    if (discoverable == null) {
+      throw new IOException("Stream Service Endpoint not found");
+    } else {
```
chtyim: No need to have else.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/AbstractStreamWriter.java
```diff
@@ -0,0 +1,160 @@
...
+      if (batch) {
+        path = String.format("%s/batch", path);
+      }
+      try {
+        return new URL(path);
+      } catch (MalformedURLException e) {
+        LOG.error("Got exception while creating StreamURL", e);
+      }
+    }
+    return null;
```
chtyim: Why return null, why not throw exception if there is `MalformedURLException`?

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/AbstractStreamWriter.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  private void writeToStream(Id.Stream stream, HttpRequest request) throws IOException {
+    HttpResponse response = HttpRequests.execute(request);
+    int responseCode = response.getResponseCode();
+    if (responseCode == HttpResponseStatus.NOT_FOUND.code()) {
+      throw new IOException(String.format("Stream %s not found", stream));
+    } else if (responseCode < 200 || responseCode >= 300) {
```
chtyim: No need to have else.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/AbstractStreamWriter.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  @Override
+  public void write(String stream, String data) throws IOException {
+    write(stream, data, ImmutableMap.<String, String>of());
+  }
+
+  @Override
+  public void write(String stream, String data, Map<String, String> headers) throws IOException {
+    URL streamURL = getStreamURL(stream);
```
chtyim: Why not just call the `write(String, ByteBuffer, Map)` method?

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamBatchWriter.java
```diff
@@ -0,0 +1,56 @@
...
+    ByteBuffers.writeToStream(data, connection.getOutputStream());
+  }
+
+  @Override
+  public void close() throws IOException {
+    connection.getOutputStream().close();
+    int responseCode = connection.getResponseCode();
+    if (responseCode == HttpResponseStatus.NOT_FOUND.code()) {
+      throw new IOException(String.format("Stream %s not found", stream));
+    } else if (responseCode < 200 || responseCode >= 300) {
```
chtyim: Remove else.

anwar6953: or put else if (responseCode == 200)?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/BufferedContentWriter.java
```diff
@@ -74,6 +74,10 @@ public void cancel() {
     // No-op
   }
 
+  Iterator<ByteBuffer> getBufferedContent() {
```
chtyim: Should just have `BufferedContentWriter` implements `Iterable<ByteBuffer>`.

chtyim:
Done review.

gokulavasan:
@chtyim Thanks for the insightful comments! Addressed them. Please take a look again.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamBatchWriter.java
```diff
@@ -0,0 +1,58 @@
...
+  @Override
+  public void close() throws IOException {
+    connection.getOutputStream().close();
+    int responseCode = connection.getResponseCode();
+    if (responseCode == HttpResponseStatus.NOT_FOUND.code()) {
+      throw new IOException(String.format("Stream %s not found", stream));
+    }
+
+    if (responseCode < 200 || responseCode >= 300) {
+      throw new IOException(String.format("Writing to Stream %s did not succeed", stream));
```
anwar6953: Should we log the code, so this information is not lost (if we're not going to put the errorcode in the exception message)

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private void writeToStream(Id.Stream stream, HttpRequest request) throws IOException {
+    HttpResponse response = HttpRequests.execute(request);
+    int responseCode = response.getResponseCode();
+    if (responseCode == HttpResponseStatus.NOT_FOUND.code()) {
+      throw new IOException(String.format("Stream %s not found", stream));
+    }
+
+    if (responseCode < 200 || responseCode >= 300) {
```
anwar6953: same here.

cdap-client-tests/src/test/java/co/cask/cdap/client/ProgramClientTestRun.java
```diff
@@ -50,6 +51,7 @@ public void setUp() throws Throwable {
     appClient = new ApplicationClient(clientConfig);
     procedureClient = new ProcedureClient(clientConfig);
     programClient = new ProgramClient(clientConfig);
+    serviceClient = new ServiceClient(clientConfig);
```
anwar6953: why is this needed still?

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
gokulavasan: Added responseCode in the exception msg.

gokulavasan:
@anwar6953 @chtyim Please review.

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamBatchWriter.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/**
+ * Used to write to Stream in Batch mode. Once closed, a new {@link StreamBatchWriter} needs to be created to do
+ * another set of batch writes.
+ */
+public interface StreamBatchWriter extends Closeable {
```
chtyim: Actually, it'll be good for this interface extends from `WritableByteChannel` (which is almost the same as this interface right now). http://docs.oracle.com/javase/6/docs/api/java/nio/channels/WritableByteChannel.html

cdap-api/src/main/java/co/cask/cdap/api/data/stream/StreamWriter.java
```diff
@@ -0,0 +1,91 @@
...
+import java.nio.ByteBuffer;
+import java.util.Map;
+
+/**
+ * Enabling Programs to write to Streams through their Context.
+ */
+public interface StreamWriter {
+
+  /**
+   * Write a string to a stream
```
chtyim: I think for javadoc, all verbs are of third-person singular form (Writes, Sets, Reads).

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
```diff
@@ -0,0 +1,143 @@
...
+    this.discoveryServiceClient = discoveryServiceClient;
+  }
+
+  private URL getStreamURL(String stream) throws IOException {
+    return getStreamURL(stream, false);
+  }
+
+  private URL getStreamURL(String stream, boolean batch) throws IOException {
+    ServiceDiscovered discovered = discoveryServiceClient.discover(Constants.Service.STREAMS);
+    Discoverable discoverable = new RandomEndpointStrategy(discovered).pick(1, TimeUnit.SECONDS);
```
chtyim: You don't need to new an EndpointStrategy every time. You can create once in constructor and keep using it.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  @Override
+  public void write(String stream, String data) throws IOException {
+    write(stream, data, ImmutableMap.<String, String>of());
+  }
+
+  @Override
+  public void write(String stream, String data, Map<String, String> headers) throws IOException {
+    write(stream, ByteBuffer.wrap(Bytes.toBytes(data)), headers);
```
chtyim: Better do `Charsets.UTF_8.encode(data)`.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/upload/LengthBasedContentWriter.java
```diff
@@ -0,0 +1,103 @@
...
+    if (fileContentWriter != null) {
+      fileContentWriter.appendAll(bodies, immutable);
+    } else {
+      long tempSize = 0;
+      while (bodies.hasNext()) {
+        ByteBuffer next = bodies.next();
+        tempSize += next.remaining();
+        bufferedContentWriter.append(next, immutable);
+      }
+      updateWriter(tempSize);
```
chtyim: If you update at the end, it could already way over the threshold memory limit. It think it's better to update on every append call. Something like:

```java
while (bodies.hasNext()) {
  ByteBuffer body = bodies.next();
  int size = next.remaining();
  bufferedContentWriter.append(next, immutable);
  if (updateWriter(size)) {         // updateWriter returns true if the writer is switched
    appendAll(bodies, immutable);
    break;
  }
}
```

gokulavasan:
@chtyim Addressed comments. Please review. Thanks!

anwar6953:
@gokulavasan no further comments from me. lgtm.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamBatchWriter.java
```diff
@@ -0,0 +1,69 @@
...
+  @Override
+  public int write(ByteBuffer data) throws IOException {
+    int size = data.remaining();
+    ByteBuffers.writeToStream(data, connection.getOutputStream());
+    return size;
+  }
+
+  @Override
+  public void close() throws IOException {
+    connection.getOutputStream().close();
```
chtyim: Also call `connection.disconnect()` in a finally block.

chtyim:
Just one comment left. Otherwise LGTM

gokulavasan:
@chtyim Addressed your final comment and also set the chunkedStreamingMode. Please review.

chtyim:
LGTM

gokulavasan:
Will merge after build passes : http://builds.cask.co/browse/CDAP-DUT857

PR #1449, opened 2015-03-02
Rename Preferences endpoints

gokulavasan:
JIRA: https://issues.cask.co/browse/CDAP-1654

Bamboo: http://builds.cask.co/browse/CDAP-DUT934


sreevatsanraman:
This does not change the APIS as per JIRA. 
For instance: We wanted to rename
`v3/configuration/preferences/namespaces/{id}` to `v3/namespaces/{id}/preferences` in the JIRA>

gokulavasan:
@sreevatsanraman Updated PR. Please review.

gokulavasan:
@ajainarayanan @gaarf This PR will change the endpoints used for Console Settings (user specific) in the UI. Please refer to the JIRA for the REST API change.

gaarf:
@gokulavasan @sreevatsanraman does this mean per-user configuration is now also per-namespace?

gokulavasan:
@sreevatsanraman Build passed. Please review.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/PreferencesHttpHandlerTest.java
```diff
@@ -33,6 +33,7 @@
  * Tests for {@link PreferencesHttpHandler}
  */
 public class PreferencesHttpHandlerTest extends AppFabricTestBase {
+
```
sreevatsanraman: We don't have test for other preferences endpoints? 
* Namespace level
* App level
* Program level

Can you add them if they are not there currently?

gokulavasan: There are tests that test those endpoints. setProperty, getProperty, deleteProperty are helper methods that help construct the right URI.

gokulavasan:
@sreevatsanraman Can you please review? I have addressed your comments.

sreevatsanraman:
LGTM :+1:

PR #1452, opened 2015-03-02
CDAP-1663: Refactor Id.Program to include ProgramType

jwang47:
First part of several large refactoring PRs to enable a clean implementation of application authorization.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/program/ManifestFields.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/program/ManifestFields.java
@@ -25,7 +25,7 @@ import java.util.jar.Attributes;
 public final class ManifestFields {
   public static final Attributes.Name MAIN_CLASS = Attributes.Name.MAIN_CLASS;
   public static final Attributes.Name MANIFEST_VERSION = Attributes.Name.MANIFEST_VERSION;
-  public static final Attributes.Name PROCESSOR_TYPE = new Attributes.Name("Processor-Type");
+  public static final Attributes.Name PROGRAM_TYPE = new Attributes.Name("Program-Type");
   public static final Attributes.Name SPEC_FILE = new Attributes.Name("Spec-File");
 
   public static final Attributes.Name ACCOUNT_ID = new Attributes.Name("Account-Id");
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/program/ManifestFields.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/program/ManifestFields.java
@@ -25,7 +25,7 @@ import java.util.jar.Attributes;
 public final class ManifestFields {
   public static final Attributes.Name MAIN_CLASS = Attributes.Name.MAIN_CLASS;
   public static final Attributes.Name MANIFEST_VERSION = Attributes.Name.MANIFEST_VERSION;
-  public static final Attributes.Name PROGRAM_TYPE = new Attributes.Name("Program-Type");
+  public static final Attributes.Name PROGRAM_TYPE = new Attributes.Name("Processor-Type");
   public static final Attributes.Name SPEC_FILE = new Attributes.Name("Spec-File");
 
   public static final Attributes.Name ACCOUNT_ID = new Attributes.Name("Account-Id");
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -238,8 +238,8 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     }
 
     try {
-      Id.Program program = Id.Program.from(namespaceId, appId, id);
       ProgramType programType = ProgramType.valueOfCategoryName(type);
+      Id.Program program = Id.Program.from(namespaceId, appId, programType, id);
       StatusMap statusMap = getStatus(program, programType);
       // If status is null, then there was an error
       if (statusMap.getStatus() == null) {
@@ -271,7 +271,8 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
       }
 
       String programName = scheduleSpec.getProgram().getProgramName();
-      Id.Program programId = Id.Program.from(namespaceId, appId, programName);
+      ProgramType programType = ProgramType.valueOfSchedulableType(scheduleSpec.getProgram().getProgramType());
+      Id.Program programId = Id.Program.from(namespaceId, appId, programType, programName);
       JsonObject json = new JsonObject();
       json.addProperty("status", scheduler.scheduleState(programId, scheduleSpec.getProgram().getProgramType(),
                                                          scheduleName).toString());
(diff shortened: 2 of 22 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/program/ManifestFields.java
```diff
@@ -25,7 +25,7 @@
 public final class ManifestFields {
   public static final Attributes.Name MAIN_CLASS = Attributes.Name.MAIN_CLASS;
   public static final Attributes.Name MANIFEST_VERSION = Attributes.Name.MANIFEST_VERSION;
-  public static final Attributes.Name PROCESSOR_TYPE = new Attributes.Name("Processor-Type");
+  public static final Attributes.Name PROGRAM_TYPE = new Attributes.Name("Program-Type");
```
chtyim: So this will break jars that are already deployed?

jwang47: Will revert this change.

jwang47:
Build passed http://builds.cask.co/browse/CDAP-DUT941-5

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -955,8 +957,8 @@ public void liveInfo(HttpRequest request, HttpResponder responder, @PathParam("n
                        @PathParam("program-id") String programId) {
     ProgramType type = getProgramType(programCategory);
     if (type == null) {
-      responder.sendString(HttpResponseStatus.METHOD_NOT_ALLOWED, String.format("Live-info not supported for program" +
-                                                                                  " type '%s'", programCategory));
+      responder.sendString(HttpResponseStatus.BAD_REQUEST,
```
bdmogal: Should we send a review for this response code change?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -332,18 +332,27 @@ protected boolean respondIfElementNotFound(Throwable t, HttpResponder responder)
...
+                                Data type, String namespaceId, String name, String appId) {
     try {
       if ((name != null && name.isEmpty()) || (appId != null && appId.isEmpty())) {
         responder.sendString(HttpResponseStatus.BAD_REQUEST, "Empty name provided");
         return;
       }
 
-      Id.Program program = Id.Program.from(namespace, appId == null ? "" : appId, "");
-      String json = name != null ? getDataEntity(store, dsFramework, program, type, name) :
-        appId != null ? listDataEntitiesByApp(store, dsFramework, program, type)
```
bdmogal: huh? was there a bug here that you fixed?

bdmogal: looks like you did the right thing, just wondering how we didn't catch this bug so far. am I missing something?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
```diff
@@ -200,7 +200,7 @@ public synchronized RuntimeInfo lookup(final RunId runId) {
           continue;
         }
 
-        Id.Program programId = Id.Program.from(matcher.group(2), matcher.group(3), matcher.group(4));
+        Id.Program programId = Id.Program.from(matcher.group(2), matcher.group(3), appType, matcher.group(4));
```
bdmogal: appType?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
jwang47: I don't think I fixed any bugs here, just refactored it to be easier to read and also not pass unused information. For example `listDataEntitiesByApp(store, dsFramework, program, type)` doesn't need `Id.Program program`, it only needs `Id.Application`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
jwang47: Not sure why the local variable was previously named `appType`. Anyway, the local variable is not needed, so I can remove it.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
jwang47: I think it's straightforward, but I probably used the wrong HTTP code here.

`Method Not Allowed` is for when you make a request to a path that doesn't support the HTTP method in your request. In this case, `Not Found` would probably make the most sense since conceptually there isn't a live-info endpoint for program types that don't support it.

jwang47: But actually this isn't really in the scope of this PR, so I'll revert that change.

cdap-proto/src/main/java/co/cask/cdap/proto/ProgramType.java
```diff
@@ -61,28 +106,41 @@
...
-    return categoryName;
+    return parameters.getCategoryName();
   }
 
   public String getPrettyName() {
-    return prettyName;
+    return parameters.getPrettyName();
+  }
+
+  public SchedulableProgramType getSchedulableType() {
```
bdmogal: ``@Nullable``?

cdap-proto/src/main/java/co/cask/cdap/proto/ProgramType.java
```diff
@@ -99,7 +157,82 @@ public static ProgramType valueOfCategoryName(String categoryName) {
...
+      Preconditions.checkArgument(prettyName != null, "prettyName cannot be null");
+      Preconditions.checkArgument(listable != null, "listable cannot be null");
+      Preconditions.checkArgument(categoryName != null, "categoryName cannot be null");
+      this.prettyName = prettyName;
+      this.listable = listable;
+      this.categoryName = categoryName;
+      this.schedulableType = schedulableType;
+    }
+
+    public SchedulableProgramType getSchedulableType() {
```
bdmogal: ``@Nullable`` here too? Although where is this used?

bdmogal:
Overall, just minor comments. However, I am not too sure about the manifest/jar part of the code that this PR modifies. Has @chtyim reviewed it already?

jwang47:
Yes, @chtyim did. I reverted the change to the manifest field name, so it's only changing the variable name now.

bdmogal:
Ok. The other stuff LGTM.

bdmogal:
Apart from the two minor ``@Nullable`` comments :)

jwang47:
Build passed http://builds.cask.co/browse/CDAP-DUT941-7

PR #1455, opened 2015-03-02
Feature/ui unit tests

elfenheart:
Unit test for Filters:
- ellipsis
- bytes
- title

### DIFF
--- a/cdap-ui/test/karma-conf.js
+++ b/cdap-ui/test/karma-conf.js
@@ -11,7 +11,8 @@ module.exports = function(config){
       'bower_components/jquery/dist/jquery.js',
       'bower_components/angular-mocks/angular-mocks.js',
       'app/**/*-unit-test.js',
-      'test/unit/**/*.js'
+      'test/unit/**/*.js',
+      'test/unit/**/**/*.js'
     ],
 
     autoWatch : true,
--- a/cdap-ui/test/karma-conf.js
+++ b/cdap-ui/test/karma-conf.js
@@ -11,8 +11,7 @@ module.exports = function(config){
       'bower_components/jquery/dist/jquery.js',
       'bower_components/angular-mocks/angular-mocks.js',
       'app/**/*-unit-test.js',
-      'test/unit/**/*.js',
-      'test/unit/**/**/*.js'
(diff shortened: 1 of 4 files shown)


cdap-ui/test/karma-conf.js
```diff
@@ -11,7 +11,8 @@ module.exports = function(config){
       'bower_components/jquery/dist/jquery.js',
       'bower_components/angular-mocks/angular-mocks.js',
       'app/**/*-unit-test.js',
-      'test/unit/**/*.js'
+      'test/unit/**/*.js',
+      'test/unit/**/**/*.js'
```
ajainarayanan: I suppose this is redundant right? Line 14 does this for us..

cdap-ui/test/unit/filters/bytes-test.js
```diff
@@ -0,0 +1,66 @@
...
+
+  var bytes,
+      $filter;
+
+  beforeEach(inject(function(_$filter_) {
+    $filter = _$filter_;
+  }));
+
+  beforeEach(function() {
+    bytes = $filter('bytes');
```
ajainarayanan: Would it be better if we combined the two ```beforeEach``` calls to one?
Something like this,

```
  var bytes;

  beforeEach(inject(function(_$filter_) {
    bytes = _$filter_('bytes');
  }));
```

One less variable and one less ```beforeEach``` invocation :smile:

cdap-ui/test/unit/filters/ellipsis-test.js
```diff
@@ -0,0 +1,64 @@
...
+  beforeEach(function() {
+    ellipsis = $filter('myEllipsis');
+  });
+
+  it('should truncate string longer that limit to limit plus ellipsis', function() {
+
+    var input = 'string longer than ten characters';
+    var limit = 10;
+
+    expect(ellipsis(input, limit).length).toBe(limit + 1);
```
ajainarayanan: May be if have ```ellipsis(input, limit)``` in a variable then again we can reduce the function to once instead of twice.

ajainarayanan:
:+1: LGTM

PR #1457, opened 2015-03-02
Feature/cli packaging fixes

dereklwood:
This PR fixes https://issues.cask.co/browse/CDAP-1397

- [x] removes ``cdap-cli.{sh,bat}`` from non-cli packages by moving the ``copy-cli-scripts`` plugin execution from the parent pom.xml dist profile, to that of cdap-cli/pom.xml.  
   - [x] I also had to split it into two separate executions, one for each rpm/deb, since some ordering of resources changed.  Previously ``copy-cli-scripts`` would run and substitute tokens in ``stage-package`` before it is copied to ``stage-package-deb``.  Now, ``copy-cli-scripts`` gets run after ``stage-package-deb`` is already created, so the token substitution needs to be run for each.
- [x] prevents ``/etc/init.d/cdap-cli`` from getting created in the ``cdap-cli`` packages by overriding the relevant parent plugin executions into no-ops.

### DIFF
--- a/cdap-cli/pom.xml
+++ b/cdap-cli/pom.xml
@@ -155,39 +155,12 @@
               <!-- Copy CLI scripts -->
               <execution>
                 <id>copy-cli-scripts</id>
-                <phase>prepare-package</phase>
+                <phase>process-resources</phase>
                 <goals>
                   <goal>copy-resources</goal>
                 </goals>
                 <configuration combine.self="override">
                   <outputDirectory>${stage.opt.dir}</outputDirectory>
-                  <useDefaultDelimiters>false</useDefaultDelimiters>
-                  <delimiters>
-                    <delimiter>@@</delimiter>
-                  </delimiters>
-                  <resources>
-                    <resource>
-                      <directory>${project.parent.basedir}/cdap-cli/bin</directory>
-                      <targetPath>bin</targetPath>
-                      <includes>
-                        <include>cdap-cli.sh</include>
-                        <include>cdap-cli.bat</include>
-                      </includes>
-                      <filtering>true</filtering>
-                    </resource>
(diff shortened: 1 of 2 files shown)


cdap-cli/pom.xml
```diff
@@ -147,6 +151,61 @@
...
+                      <includes>
+                        <include>cdap-cli.sh</include>
+                        <include>cdap-cli.bat</include>
+                      </includes>
+                      <filtering>true</filtering>
+                    </resource>
+                  </resources>
+                </configuration>
+              </execution>
+              <!-- Copy CLI scripts for .deb -->
```
chtyim: Why need a separated one for deb? The current pom structure is, in the deb-prepare profile, it will copies everything done by the rpm-prepare (meaning when building deb package only, you still need to includes the rpm-prepare profile, although we usually build both, hence specifying "dist,rpm-prepare,rpm,deb-prepare,deb"

chtyim:
Just one comment.

cdap-cli/pom.xml
dereklwood: It's what I tried to explain in the PR description.  Without the separate one for deb, the generated ``cdap-cli.sh`` and ``cdap-cli.bat`` in the debian package do not have the ``@@project.version@@`` tokens substituted.  I believe this is because the token substitution is now done only after the "copy rpm to deb" is done

dereklwood: after some discussion, will try changing the phase to "process-resources"

dereklwood:
updated to run the cli-scripts copy/token substitution at an earlier phase, instead of requiring two separate executions for rpm/deb

chtyim:
LGTM

PR #1459, opened 2015-03-03
Feature/2.8.0 namespace example changes

awholegunch:
These changes update the examples *in the documentation* (**not** the actual examples) to use *v3* RESTful APIs.

This is staged at: http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-2.8.0_namespace_example_changes/en/examples-manual/examples/index.html

All examples (with the exception of CountRandom) have been altered.


cdap-docs/admin-manual/source/operations/index.rst
```diff
@@ -91,6 +91,24 @@ On success, the calls return a valid HTTP response with a 200 code.
 - To retrieve the **health check of all the services running in YARN**, make a GET request
   to the URI::
   
-    http://<host>:10000/v2/system/services
+    http://<host>:10000/v3/namespaces/system/services
```
bdmogal: there is no 'namespaces' component in the system/services endpoint.

bdmogal:
Should we expose the ability to use custom namespaces in example docs (rather than always using default)? Perhaps with a variable?

cdap-docs/admin-manual/source/operations/index.rst
awholegunch: Fixed.

awholegunch:
Comment on the last commit should be "Add a note about namespaces..."; which refers to adding text at the top of all examples that explains that all examples use the "default" namespace. See the [developers' manual](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-2.8.0_namespace_example_changes/en/developers-manual/getting-started/building-apps.html) and the [individual examples](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-2.8.0_namespace_example_changes/en/examples-manual/examples/index.html).

awholegunch:
@bdmogal Can you take another look, and see if you think I've addressed the examples adequately?

bdmogal:
Only a couple of minor changes that we discussed offline to explain how to use a non-default namespace, otherwise LGTM

awholegunch:
See sample in [Hello World Example](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-2.8.0_namespace_example_changes/en/examples-manual/examples/hello-world.html#building-and-starting).

bdmogal:
:+1:

PR #1460, opened 2015-03-03
remove the {runnable} from our APIs where appropriate

anew:
Part of /v3 REST API cleanup

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/ApplicationSpecification.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/ApplicationSpecification.java
@@ -35,7 +35,7 @@ import co.cask.cdap.data.dataset.DatasetCreationSpec;
 import java.util.Map;
 
 /**
- * Application Specification used in core code.
+ * Application Specification used in core code.P
  */
 public interface ApplicationSpecification {
 
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/ApplicationSpecification.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/ApplicationSpecification.java
@@ -35,7 +35,7 @@ import co.cask.cdap.data.dataset.DatasetCreationSpec;
 import java.util.Map;
 
 /**
- * Application Specification used in core code.P
+ * Application Specification used in core code.
  */
 public interface ApplicationSpecification {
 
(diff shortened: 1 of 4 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/ApplicationSpecification.java
```diff
@@ -35,7 +35,7 @@
 import java.util.Map;
 
 /**
- * Application Specification used in core code.
+ * Application Specification used in core code.P
```
sreevatsanraman: Un-intended character 'P'

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
```diff
@@ -301,44 +301,44 @@ public void stopProgram(HttpRequest request, HttpResponder responder,
...
    */
   @GET
-  @Path("/apps/{app-id}/{runnable-type}/{runnable-id}/runs")
+  @Path("/apps/{app-id}/{program-type}/{program-id}/runs")
   public void runnableHistory(HttpRequest request, HttpResponder responder,
                               @PathParam("app-id") String appId,
-                              @PathParam("runnable-type") String runnableType,
-                              @PathParam("runnable-id") String runnableId,
+                              @PathParam("program-type") String runnableType,
+                              @PathParam("program-id") String runnableId,
```
sreevatsanraman: We still have runnableType and runnableId in the variable.

sreevatsanraman:
There are a few places (variable names) where we use runnableType and runnableId. Can we fix those as well? LGTM other than those comments.

awholegunch:
We also have numerous references in the docs. Did you want to fix this as part of this or another PR?

cdap-app-fabric/src/main/java/co/cask/cdap/app/ApplicationSpecification.java
anew: Oops - and unnoticed. Will fix

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
anew: will fix.

anew:
@awholegunch where in the docs? Can you send me a reference?

anew:
@sreevatsanraman I fixed the remaining ones - except where a method takes a programId and a runnable id (runnable being a flowlet or a service runnable, which will get fixed in a separate PR I hope).

anew:
@awholegunch I searched through the docs, I could only find it in the context where it means flowlet or handler of a service. For these cases we are keeping this notation for now. But we are removing the "runnable" notation from services, and then we should make the doc change, too.

PR #1458, opened 2015-03-03
Upgrade tool: Refactoring around existing code

rohitsinha54:
- This PR refactors around the code in the Main class to give a basic architecture for the framework.

PR #1461, opened 2015-03-03
Upgrade tool: Archive/file upgrade

rohitsinha54:
- Renames the stream files for namespace 
- Renames the log files for namespace

This PR is created against feature/upg-tool branch which has received LGTM here: https://github.com/caskdata/cdap/pull/1458
But has not been merged as the build are being hung: 
https://github.com/caskdata/cdap/pull/1458#issuecomment-76904895

Opening it out for review and will get it in only after the earlier one goes in.


cdap-master/src/main/java/co/cask/cdap/data/tools/AbstractUpgrader.java
```diff
@@ -16,15 +16,60 @@
...
+  protected Location renameLocation(Location oldLocation, Location newLocation) throws IOException {
+    // if the newLocation does not exists or the oldLocation does we try to rename. If either one of them is false then
+    // the underlying call to renameTo will throw IOException which we propagate.
+    if (!newLocation.exists() || oldLocation.exists()) {
+      Locations.getParent(newLocation).mkdirs();
+      try {
+        return oldLocation.renameTo(newLocation);
+      } catch (IOException ioe) {
+        newLocation.delete();
+        LOG.warn("Failed to rename {} with {}", oldLocation, newLocation);
```
albertshau: nit: "rename {} to {}"

cdap-master/src/main/java/co/cask/cdap/data/tools/ArchiveUpgrader.java
```diff
@@ -0,0 +1,93 @@
...
+  public void upgrade() throws Exception {
+    upgradeStream();
+    upgradeLogs();
+  }
+
+  /**
+   * Upgrades the streams archive path
+   */
+  private void upgradeStream() throws IOException {
+    LOG.info("Upgrading stream files ...");
```
albertshau: is there any stream metadata that needs to be upgraded?  Or is that coming in the future?

cdap-master/src/main/java/co/cask/cdap/data/tools/ArchiveUpgrader.java
```diff
@@ -0,0 +1,93 @@
...
+    if (renameLocation(oldLocation, newLocation) != null) {
+      LOG.info("Upgraded streams archives from {} to {}", oldLocation, newLocation);
+    }
+  }
+
+  /**
+   * Upgraded the logs path
+   */
+  private void upgradeLogs() throws IOException {
+    LOG.info("Upgrading log files ...");
```
albertshau: similar question, is there log metadata that needs upgrading?

cdap-master/src/main/java/co/cask/cdap/data/tools/ArchiveUpgrader.java
```diff
@@ -0,0 +1,93 @@
...
+    String logBaseDir = cConf.get(LoggingConfiguration.LOG_BASE_DIR);
+    upgradeLogsHelper(locationFactory.create(logBaseDir).append(Constants.Logging.SYSTEM_NAME),
+                      Constants.SYSTEM_NAMESPACE);
+    upgradeLogsHelper(locationFactory.create(logBaseDir).append(DEVELOPER_ACCOUNT), Constants.DEFAULT_NAMESPACE);
+  }
+
+  /**
+   * Upgrades the log path through a temp location. This is required as the source and final destination are under same
+   * directory.
```
albertshau: I'm confused about this.  Isn't it moving /logs/avro/cdap to /system/logs/avro and /logs/avro/developer to /default/logs/avro?

cdap-master/src/main/java/co/cask/cdap/data/tools/ArchiveUpgrader.java
rohitsinha54: Yes it is coming as an independent PR.

cdap-master/src/main/java/co/cask/cdap/data/tools/ArchiveUpgrader.java
rohitsinha54: Yes.

cdap-master/src/main/java/co/cask/cdap/data/tools/ArchiveUpgrader.java
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * @param oldLogLocation the old {@link Location} of the log
+   * @param namespace      the namespace under which this log should be stored
+   * @throws IOException
+   */
+  private void upgradeLogsHelper(Location oldLogLocation, String namespace) throws IOException {
```
albertshau: nit: it's better for a method name to be a verb instead of a noun.  I'm still confused why there needs to be a temp dir though.  If it's really required, you need to be able to handle the case where it renames to temp and then fails.

albertshau:
small comments, and one question about renaming.

cdap-master/src/main/java/co/cask/cdap/data/tools/ArchiveUpgrader.java
rohitsinha54: Yes its a part of MDSUpgrade as the data in app.meta table. It will come in a independent PR

cdap-master/src/main/java/co/cask/cdap/data/tools/ArchiveUpgrader.java
rohitsinha54: In the case when it moves to temp and then fails when the upgrade tools runs again after the failure it will move it from the temp to newlocation

rohitsinha54: Both new and old location are under /cdap in starting its hostname/cdap so we have to go through the temp dir.

albertshau: I don't think that is a problem right?  It's only a problem if you're doing something like moving /logs/avro/cdap to /logs/avro/cdap/xyz

rohitsinha54: Yes my bad. Even the temp is under that.
Let me have a look.

rohitsinha54:
Closing this. Opened against develop here: https://github.com/caskdata/cdap/pull/1475

PR #1463, opened 2015-03-03
Feature/namespaced dataset definition api

bdmogal:
Summary:

1. Adds ``co.cask.cdap.api.DatasetContext`` to only contain ``namespaceId`` for now, but potentially more fields later.
2. ``DatasetDefinition.getDataset`` and ``DatasetDefinition.getAdmin`` accept an additional ``DatasetContext`` parameter.
3. Does not use the ``DatasetContext`` in this PR, that will happen in a future PR once the API is finalized.

Questions:
1. Should we mark ``co.cask.cdap.api.data.DatasetContext`` deprecated as part of this change?
2. Thought about using an object for the namespace field rather than a String, however, what we need is the ``Id.Namespace``, which is currently in ``cdap-proto``. 

Build:
http://builds.cask.co/browse/CDAP-DUT948-3


cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetDefinition.java
```diff
@@ -71,24 +74,28 @@
...
-   * @param spec specification of the dataset instance.
+   *
+   * @param datasetContext context for the dataset
    * @param classLoader classloader to use when executing admin operations
+   * @param spec specification of the dataset instance.
    * @return dataset admin to perform administrative operations
    * @throws IOException
    */
-  A getAdmin(DatasetSpecification spec, ClassLoader classLoader) throws IOException;
+  A getAdmin(DatasetContext datasetContext, ClassLoader classLoader, DatasetSpecification spec) throws IOException;
```
ghelmling: Why switch ClassLoader and DatasetSpecification parameters?

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetDefinition.java
```diff
@@ -71,24 +74,28 @@
...
    * @param arguments arguments for this instance of the dataset. Should not be null - provide an empty map for no
    *                  arguments.
    * @param classLoader classloader to use when executing dataset operations
+   * @param spec specification of the dataset instance.
    * @return dataset to perform object operations
    * @throws IOException
    */
-  D getDataset(DatasetSpecification spec, Map<String, String> arguments, ClassLoader classLoader)
-      throws IOException;
+  D getDataset(DatasetContext datasetContext, Map<String, String> arguments, ClassLoader classLoader,
```
ghelmling: Same here, why move DatasetSpecification to the end?

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetDefinition.java
bdmogal: Must have happened during refactoring in the IDE. I'll move it to be the second parameter.

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetDefinition.java
bdmogal: Will fix to make it the second parameter.

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetDefinition.java
abaranau: What will happen if I create dataset in namespace A and in this method provide the context with namespace B?
With this change DatasetSpecification no longer identifies dataset instance, but rather dataset instance is identified by a pair: context and spec. How do we make sure these two are "in sync", meaning that we don't provide wrong context with specific spec and the other way around?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/executor/DatasetAdminOpHTTPHandler.java
```diff
@@ -104,7 +105,7 @@ public void create(HttpRequest request, HttpResponder responder,
     }
 
     DatasetSpecification spec = type.configure(name, props);
-    DatasetAdmin admin = type.getAdmin(spec);
+    DatasetAdmin admin = type.getAdmin(new DatasetContext.Builder().setNamespaceId(namespaceId).build(), spec);
```
abaranau: Will it be more readable to have DatasetContext.fromNamespace()?

abaranau: In general, why do we use builder? It is extremely simple object to build with single param...

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetContext.java
```diff
@@ -0,0 +1,93 @@
...
+   * Returns the namespace for a dataset
+   *
+   * @return the dataset's namespace
+   */
+  public String getNamespaceId() {
+    return namespaceId;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
abaranau: why do you need equals and hashcode? do you use it in map as key?

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetContext.java
```diff
@@ -0,0 +1,93 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.dataset;
+
+import co.cask.cdap.api.annotation.Beta;
+
+/**
+ * Provides access to the context for a dataset including its environment and configuration
```
abaranau: I know we discussed it and you guys decided to go the other way, but it still looks confusing to mix at least two things here:
* dataset instance identification (namespace, name - identifies dataset instance)
* environment/context where the dataset object is being instantiated and used

Oh, and per javadoc, dataset configuration is added to the mix,,,

abaranau:
please see couple comments and questions

abaranau:
"We can’t automatically merge this pull request"

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetContext.java
```diff
@@ -0,0 +1,93 @@
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    DatasetContext that = (DatasetContext) o;
+
+    if (!namespaceId.equals(that.namespaceId)) {
```
albertshau: ide autogenerated right?  simpler to do ```return namespaceId.equals(that.namespaceId)```

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/executor/DatasetAdminOpHTTPHandler.java
bdmogal: No other reason but future-proofing. There may be other fields in ``DatasetContext`` in future - other things that we may want to pass to storage providers.

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetContext.java
bdmogal: yeah :( Changed.

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetDefinition.java
bdmogal: Correct me if I am wrong. The ``DatasetAdminOpHttpHandler`` is the one that will call these methods, right? It will get a request for a dataset with a namespace-id in the request. It uses that namespace-id to create the ``DatasetContext`` and calls this method. If the namespace id is wrong (or you use a namespace that you do not have access to), the admin operation will fail.

Anything that I am missing?

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetContext.java
bdmogal: Not currently, so not needed. It was just a case of methods generated together. Will remove.

albertshau: did you mean to remove equals as well?  Should either have both or none.

cdap-api/src/main/java/co/cask/cdap/api/dataset/DatasetDefinition.java
```diff
@@ -71,24 +74,28 @@
...
+   * @param classLoader classloader to use when executing dataset operations
    * @param arguments arguments for this instance of the dataset. Should not be null - provide an empty map for no
    *                  arguments.
-   * @param classLoader classloader to use when executing dataset operations
    * @return dataset to perform object operations
    * @throws IOException
    */
-  D getDataset(DatasetSpecification spec, Map<String, String> arguments, ClassLoader classLoader)
-      throws IOException;
+  D getDataset(DatasetContext datasetContext, DatasetSpecification spec, ClassLoader classLoader,
```
albertshau: nit: seems more consistent if arguments is before classLoader

albertshau:

bdmogal:
@albertshau addressed all comments. Will merge after I'm satisfied that end-to-end works.

albertshau:
lgtm

bdmogal:
Build passed - http://builds.cask.co/browse/CDAP-DUT948-7. Merging.

PR #1465, opened 2015-03-03
no ttl for aggregates

gsps1:
for aggregate resolution , we skip adding TTL
https://issues.cask.co/browse/CDAP-1626

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricDatasetFactory.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricDatasetFactory.java
@@ -80,7 +80,8 @@ public class DefaultMetricDatasetFactory implements MetricDatasetFactory {
     int ttl =  cConf.getInt(MetricsConstants.ConfigKeys.RETENTION_SECONDS + "." + resolution + ".seconds", -1);
 
     DatasetProperties.Builder props = DatasetProperties.builder();
-    if (ttl > 0) {
+    // don't add TTL for MAX_RESOLUTION table. CDAP-1626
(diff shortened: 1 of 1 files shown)


cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricDatasetFactory.java
```diff
@@ -80,7 +80,8 @@ public FactTable get(int resolution) {
     int ttl =  cConf.getInt(MetricsConstants.ConfigKeys.RETENTION_SECONDS + "." + resolution + ".seconds", -1);
 
     DatasetProperties.Builder props = DatasetProperties.builder();
-    if (ttl > 0) {
+    // don't add TTL for MAX_RESOLUTION table.
+    if (ttl > 0 && resolution != Integer.MAX_VALUE) {
```
abaranau: Is there an easy way to make that configurable? I see above lines try to get it from config. Can we default it in config to unlim or is it toodangerous?

gsps1: by default if TTL is not set , the TTL would be forever (INT_MAX) , if someone sets a lower TTL for the totals table, we want to ignore it and use TTL forever ( according to CDAP-1626). if we want to allow that to be configurable, then the issue and PR is not relevant right ?

abaranau: alright - let's override it for now. Can you add in that comment mention about jira issue "CDAP-1626" please.

abaranau:
adjust a comment and merge please!

PR #1470, opened 2015-03-03
Restrict 'root.prefix' values.

anwar6953:
StandaloneMain and MasterServiceMain do not start if the 'root.prefix' configured in CConfiguration is not strictly alphanumeric.

This is not a comprehensive list of all the checks we need to perform against the cConf, but it is a start - one that is necessary due to limitations on the characters that can be in hbase namespace, and assumptions we make about what values can be in the hbase namespace for cdap tables.
Note that our allowed values (strictly alphanumeric) are stricter than hbase's.

Build: http://builds.cask.co/browse/CDAP-DUT961-3

[Here is what happens in sdk when invalid root.prefix is given.](https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/ZJstMN1Qken9iDD/Screen%20Shot%202015-03-03%20at%203.03.57%20PM.png)

Things to consider:
I am not restricting the zookeeper value in cConf or a few other related ones that use 'root.prefix' by default, but can be overridden.

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
@@ -16,6 +16,8 @@
 
 package co.cask.cdap.common.conf;
 
+import com.google.common.base.Preconditions;
+
 import java.util.Properties;
 
 /**
@@ -32,5 +34,19 @@ public class CConfigurationUtil extends Configuration {
         destination.set(property, cConf.get(property));
       }
     }
+  }
+
+  public static void checkCConfValidity(CConfiguration cConf) {
+    // Checks to ensure that certain keys (e.g. "root.prefix") are valid as expected by CDAP.
+    checkAlphaNumberic(cConf, Constants.CFG_ROOT_NAMESPACE);
+    checkAlphaNumberic(cConf, Constants.Dataset.TABLE_PREFIX);
+  }
+
+  private static void checkAlphaNumberic(CConfiguration cConf, String key) {
+    String value = cConf.get(key);
+    Preconditions.checkNotNull(value, String.format("Entry of CConf with key: %s is null", key));
+    Preconditions.checkArgument(value.matches("[a-zA-Z0-9]+"),
+                                String.format("CConf entry with key: %s must consist " +
+                                                "of only alphanumeric characters; it is: %s", key, value));
   }
 }
--- a/cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
@@ -36,17 +36,22 @@ public class CConfigurationUtil extends Configuration {
     }
   }
 
-  public static void checkCConfValidity(CConfiguration cConf) {
+  /**
+   * Asserts that the given CConfiguration has valid properties.
+   * @param cConf the CConfiguration object to check
+   * @throws IllegalArgumentException if the given cConf is invalid.
+   */
+  public static void verify(CConfiguration cConf) {
     // Checks to ensure that certain keys (e.g. "root.prefix") are valid as expected by CDAP.
-    checkAlphaNumberic(cConf, Constants.CFG_ROOT_NAMESPACE);
-    checkAlphaNumberic(cConf, Constants.Dataset.TABLE_PREFIX);
+    assertAlphanumeric(cConf, Constants.ROOT_NAMESPACE);
+    assertAlphanumeric(cConf, Constants.Dataset.TABLE_PREFIX);
   }
 
-  private static void checkAlphaNumberic(CConfiguration cConf, String key) {
+  private static void assertAlphanumeric(CConfiguration cConf, String key) {
     String value = cConf.get(key);
-    Preconditions.checkNotNull(value, String.format("Entry of CConf with key: %s is null", key));
+    Preconditions.checkNotNull(value, "Entry of CConf with key: {} is null", key);
     Preconditions.checkArgument(value.matches("[a-zA-Z0-9]+"),
-                                String.format("CConf entry with key: %s must consist " +
-                                                "of only alphanumeric characters; it is: %s", key, value));
+                                "CConf entry with key: {} must consist " +
+                                  "of only alphanumeric characters; it is: {}", key, value);
   }
 }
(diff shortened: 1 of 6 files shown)


cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
```diff
@@ -33,4 +35,18 @@ public static void copyTxProperties(CConfiguration cConf, org.apache.hadoop.conf
       }
     }
   }
+
+  public static void checkCConfValidity(CConfiguration cConf) {
+    // Checks to ensure that certain keys (e.g. "root.prefix") are valid as expected by CDAP.
+    checkAlphaNumberic(cConf, Constants.CFG_ROOT_NAMESPACE);
+    checkAlphaNumberic(cConf, Constants.Dataset.TABLE_PREFIX);
```
jwang47: Just checking, is this duplicating any existing logic?

jwang47:
LGTM if not duplicating any existing logic.

cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
```diff
@@ -33,4 +35,18 @@ public static void copyTxProperties(CConfiguration cConf, org.apache.hadoop.conf
       }
     }
   }
+
+  public static void checkCConfValidity(CConfiguration cConf) {
```
chtyim: Call it `isValid`

chtyim: So this method will throw runtime exception and is not documented?

cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
```diff
@@ -33,4 +35,18 @@ public static void copyTxProperties(CConfiguration cConf, org.apache.hadoop.conf
...
+    // Checks to ensure that certain keys (e.g. "root.prefix") are valid as expected by CDAP.
+    checkAlphaNumberic(cConf, Constants.CFG_ROOT_NAMESPACE);
+    checkAlphaNumberic(cConf, Constants.Dataset.TABLE_PREFIX);
+  }
+
+  private static void checkAlphaNumberic(CConfiguration cConf, String key) {
+    String value = cConf.get(key);
+    Preconditions.checkNotNull(value, String.format("Entry of CConf with key: %s is null", key));
```
chtyim: Don't use `String.format` for Preconditions. Preconditions supports formatting already. Read the javadoc of Precondtions.

```
public static <T> T checkNotNull(T reference,
                                 @Nullable
                                 String errorMessageTemplate,
                                 @Nullable
                                 Object... errorMessageArgs)
```

cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
```diff
@@ -33,4 +35,18 @@ public static void copyTxProperties(CConfiguration cConf, org.apache.hadoop.conf
...
+    Preconditions.checkArgument(value.matches("[a-zA-Z0-9]+"),
```
chtyim: Same, don't use String.format.

cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
```diff
@@ -26,6 +26,7 @@
 public final class Constants {
 
   public static final String ARCHIVE_DIR = "archive";
+  public static final String CFG_ROOT_NAMESPACE = "root.namespace";
```
chtyim: Remove `CFG_`.

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
```diff
@@ -142,6 +143,7 @@ public static void main(final String[] args) throws Exception {
 
   @Override
   public void init(String[] args) {
+    CConfigurationUtil.checkCConfValidity(cConf);
```
chtyim: Shouldn't be call here. Call it inside the ConfigModule.

cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
```diff
@@ -148,6 +149,7 @@ public void run() {
    * Start the service.
    */
   public void startUp() throws Exception {
+    CConfigurationUtil.checkCConfValidity(configuration);
```
chtyim: Shouldn't be called here. Call it inside `ConfigModule`.

cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
anwar6953: The reason I had it like that, is because the value of the variable is actually the key in the CConfiguration (as opposed to the value in the CConfiguration).

Same is used in other variables in Constants.java (prefixed by CFG if its the key in the CConfiguration).

cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
anwar6953: No where in the code had direct access to the root namespace before (only through things that inherited it, such as Dataset Table Prefix). I didn't find validations on table prefix.
So I think this validation on the two properties are not duplicating logic.

jwang47:
LGTM

cdap-common/src/main/java/co/cask/cdap/common/conf/CConfigurationUtil.java
```diff
@@ -33,4 +35,23 @@ public static void copyTxProperties(CConfiguration cConf, org.apache.hadoop.conf
       }
     }
   }
+
+  /**
+   * Asserts that the given CConfiguration has valid properties.
+   * @param cConf the CConfiguration object to check
+   * @throws IllegalArgumentException if the given cConf is invalid.
+   */
+  public static void assertCConfValidity(CConfiguration cConf) {
```
chtyim: Please just call it `verify`. The method is already inside CConfigurationUtil and is taking a CConfiguration as argument. No need to be that super verbose.

anwar6953: done

chtyim:
LGTM

PR #1472, opened 2015-03-04
Fix javadocs build.

awholegunch:
Fixes broken Javadoc build.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/ObjectMappedTableProperties.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/ObjectMappedTableProperties.java
@@ -137,8 +137,8 @@ public class ObjectMappedTableProperties {
     /**
      * Sets the column type for the row key column in the corresponding Hive table for an ObjectMappedTable.
      * By default, the type of the row key in your Hive table will be binary. You can set the type using this
-     * method. Only {@link co.cask.cdap.api.data.schema.Schema.Type#BYTES Schema.Type#BYTES} and 
-     * {@link co.cask.cdap.api.data.schema.Schema.Type#STRING Schema.Type#STRING} are allowed.
+     * method. Only {@link co.cask.cdap.api.data.schema.Schema.Type#BYTES Schema.Type.BYTES} and
+     * {@link co.cask.cdap.api.data.schema.Schema.Type#STRING Schema.Type.STRING} are allowed.
      *
      * For example, if you are storing an Object with a single string field named "id", the corresponding
      * Hive table will have a schema of (rowkey binary, id string). If you set the type to a string using this method,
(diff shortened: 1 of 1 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/ObjectMappedTableProperties.java
```diff
@@ -137,7 +137,8 @@ public Builder setRowKeyExploreName(String name) {
     /**
      * Sets the column type for the row key column in the corresponding Hive table for an ObjectMappedTable.
      * By default, the type of the row key in your Hive table will be binary. You can set the type using this
-     * method. Only {@link Schema.Type#BYTES} and {@link Schema.Type#STRING} are allowed.
+     * method. Only {@link co.cask.cdap.api.data.schema.Schema.Type#BYTES Schema.Type#BYTES} and
```
chtyim: Why need the fully qualified name?

awholegunch: Because it's Javadocs? I don't know why, but Javadocs just won't build and find it without it. It's not a good as javac.

awholegunch: though I'm thinking I don't like the labels I proposed. Fixing...

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/ObjectMappedTableProperties.java
```diff
@@ -137,7 +137,8 @@ public Builder setRowKeyExploreName(String name) {
     /**
      * Sets the column type for the row key column in the corresponding Hive table for an ObjectMappedTable.
      * By default, the type of the row key in your Hive table will be binary. You can set the type using this
-     * method. Only {@link Schema.Type#BYTES} and {@link Schema.Type#STRING} are allowed.
+     * method. Only {@link co.cask.cdap.api.data.schema.Schema.Type#BYTES Schema.Type.BYTES} and
```
chtyim: So, why need the full name?

awholegunch: I don't know. It just doesn't work without it. Perhaps because it is part of an imported class. Would the complier find it?

chtyim: It's a bug in javadoc long time back and not getting fixed.

http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4464323

chtyim:
LGTM

PR #1473, opened 2015-03-04
fix metric search using star

abaranau:


gsps1:
LGTM :+1:

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/timeseries/FactTable.java
```diff
@@ -300,12 +299,18 @@ public int compare(TagValue t1, TagValue t2) {
...
+    if (actualTags.size() < expectedTags.size()) {
       return false;
     }
-    for (int i = 0; i < expected.size(); i++) {
-      if (expected.get(i).getValue() != null && !expected.get(i).equals(actual.get(i))) {
+    for (int i = 0; i < expectedTags.size(); i++) {
+      TagValue expected = expectedTags.get(i);
+      TagValue actual = actualTags.get(i);
+      // compare tag names
+      boolean matches = expected.getTagName().equals(actual.getTagName());
```
gsps1: so its safe to assume we won't support '*.*' ?

abaranau: we don't support that yet

PR #1474, opened 2015-03-04
MetricsTable test case fix

gsps1:
Fix for failing test in develop, Build Running here
http://builds.cask.co/browse/CDAP-DUT969-1


anwar6953:
LGTM

PR #1477, opened 2015-03-04
Update usage of Trademarks

awholegunch:
Adds info on Trademarks, adds trademark symbols, moves the trademark symbol replace directives to the common file, so they are available in all files.

Corrects some usages of trademarks. Makes the copyright string available as a replace directive, and uses it on the License page.

Staged at: [2.8.0-SNAPSHOT-2.8.0_trademarks](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-2.8.0_trademarks/en/index.html)

### DIFF
--- a/cdap-docs/reference-manual/source/licenses/index.rst
+++ b/cdap-docs/reference-manual/source/licenses/index.rst
@@ -55,12 +55,14 @@ Other product and company names mentioned herein, within our documentation or we
 be the trademarks of their respective owners. Where the ownership or registration is known,
 the addition of appropriate symbols has been made.
 
-Apache®, Apache Hadoop, and Hadoop® are either registered trademarks or trademarks of the
-Apache Software Foundation in the United States and/or other countries.
+Apache®, `Apache Hadoop <http://hadoop.apache.org>`__, and `Hadoop®
+<http://hadoop.apache.org>`__ are either registered trademarks or trademarks of the
+`Apache Software Foundation <http://www.apache.org>`__ in the United States and/or other
+countries.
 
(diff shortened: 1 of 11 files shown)


cdap-docs/reference-manual/source/licenses/index.rst
```diff
@@ -13,12 +13,61 @@ Cask Data Application Platform Licenses and Dependencies
...
+
+
+Other Products and Trademarks
+-----------------------------
+
+Other product and company names mentioned herein, within our documentation or webpages, may
+be the trademarks of their respective owners. Where the ownership or registration is known,
+the addition of appropriate symbols has been made.
+
+Apache®, Apache Hadoop, and Hadoop® are either registered trademarks or trademarks of the
```
albertshau: do we need a (r) after Apache Hadoop here?

albertshau:
one question, lgtm

cdap-docs/reference-manual/source/licenses/index.rst
awholegunch: Apparently not. This was copied directly from the Apache Hadoop website: 
http://www.apache.org/foundation/marks/faq/#attribution

awholegunch: Though now that I re-read it, they suggest adding hyperlinks. Which I will do...

awholegunch: Done.

awholegunch:
Thanks. Addressed comment, merging.

PR #1483, opened 2015-03-04
CDAP-1692: Documentation for CLI launch options

jwang47:


cdap-docs/reference-manual/source/cli-api.rst
```diff
@@ -67,6 +67,23 @@ you for the required credentials to acquire an access token from the CDAP instan
 the CLI will save it to ~/.cdap.accesstoken.<hostname> for later use and use it for the rest of
 the current CLI session.
 
+Options
+-------
+
+The CLI may be started with command-line options, as detailed below::
+
+  usage: CDAP CLI
```
awholegunch: Is this correct? Should this not be:
```
usage: cdap-cli.sh [-option]
```

cdap-docs/reference-manual/source/cli-api.rst
```diff
@@ -67,6 +67,23 @@ you for the required credentials to acquire an access token from the CDAP instan
...
+                            upon launch or try default connection if none
+                            provided. Defaults to "true".
+   -d,--debug               Print exception stack traces.
+   -h,--help                Print the usage message.
+   -s,--verify-ssl <arg>    If "true", verify SSL certificate when making
+                            requests. Defaults to "true".
+   -u,--uri <arg>           CDAP instance URI to interact with in the format
+                            "[http[s]://]<hostname>[:<port>[/<namespace>]]".
+                            Defaults to "http://127.0.0.1:10000/default".
+
```
awholegunch: Can we add the generation of this to the other script that generates the rest of the doc? (It could either be another class or a command line option, or, since it's not for general usage, just dump both out together, perhaps with a divider between them. The divider could be an rst comment line, or just a bunch of dashes...)

awholegunch:
Two comments, otherwise LGTM.

cdap-docs/reference-manual/source/cli-api.rst
jwang47: Is it alright if you get this by running `CLIMain` with the `-h` option?

PR #1484, opened 2015-03-04
CDAP-1693: Use netty-http ExceptionHandler

jwang47:
http://builds.cask.co/browse/CDAP-DUT980


jwang47:
http://builds.cask.co/browse/CDAP-DUT980

pom.xml
```diff
@@ -119,7 +119,7 @@
     <mockito.version>1.9.5</mockito.version>
     <mysql.version>5.1.21</mysql.version>
     <netty.version>3.6.6.Final</netty.version>
-    <netty.http.version>0.10.0</netty.http.version>
+    <netty.http.version>0.11.0-SNAPSHOT</netty.http.version>
```
anew: we should not have snapshot dependencies. Instead we should release anew version of netty-http.

nitinmotgi: Agree.

cdap-common/src/main/java/co/cask/cdap/common/exception/ConflictException.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright © 2014 Cask Data, Inc.
```
bdmogal: update copyright date

cdap-common/src/main/java/co/cask/cdap/common/exception/NotImplementedException.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright © 2014 Cask Data, Inc.
```
bdmogal: update copyright

cdap-common/src/main/java/co/cask/cdap/common/exception/NotImplementedException.java
```diff
@@ -0,0 +1,23 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.common.exception;
+
+/**
+ * Thrown when some operation is not implemented.
+ */
+public class NotImplementedException extends Exception {
```
bdmogal: Any reason to not use ``UnsupportedOperationException`` instead of creating a custom exception, especially since we do not seem to have any specific logic here.

cdap-common/src/main/java/co/cask/cdap/common/exception/HttpExceptionHandler.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Common logic to handle exceptions in handler methods.
+ */
+public class HttpExceptionHandler extends ExceptionHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(HttpExceptionHandler.class);
+
+  @Override
+  public void handle(Throwable t, HttpRequest request, HttpResponder responder) {
+    if (t instanceof IllegalArgumentException) {
```
bdmogal: I can guess that we mean to handle the exceptions thrown from ``Preconditions.checkArgument`` here. It seems a little too generic, but unsure how we can get around to changing the ``IllegalArgumentException`` to a ``BadRequestException`` without adding a fair bit of boiler-plate code.

jwang47: Currently we're catching `IllegalArgumentException` in our handler methods. See https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java#L421-L422 for an example. Also, this `HttpExceptionHandler` is not meant to handle everything, just provide some default ways to handle stuff so we don't have to repeat code -- so we can migrate to using `BadRequestException` anytime without having to change this class.

cdap-common/src/main/java/co/cask/cdap/common/exception/NotImplementedException.java
jwang47: True, I don't think it's even being used right now. Will remove and handle `UnsupportedOperationException` for now. But in the long run, the idea is to start using our own exceptions only.

cdap-common/src/main/java/co/cask/cdap/common/exception/HttpExceptionHandler.java
bdmogal: I see. Then we should probably not provide this convenience for ``IllegalArgumentException``? Have handlers methods catch it and handle if they know what they are doing?

I am just unsure if ``IllegalArgumentException`` is always a result of bad input from users, although I agree that that is probably the case most of the time.

cdap-common/src/main/java/co/cask/cdap/common/exception/ConflictException.java
```diff
@@ -0,0 +1,35 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.common.exception;
+
+/**
+ * Thrown when there was a conflict.
+ */
+public class ConflictException extends Exception {
```
bdmogal: We don't yet use this, right? Is the idea that wherever we check for ``CONFLICT``, we will instead throw this exception in later PRs?

cdap-common/src/main/java/co/cask/cdap/common/exception/HttpExceptionHandler.java
```diff
@@ -0,0 +1,52 @@
...
+  @Override
+  public void handle(Throwable t, HttpRequest request, HttpResponder responder) {
+    if (t instanceof BadRequestException) {
+      responder.sendString(HttpResponseStatus.BAD_REQUEST, t.getMessage());
+    } else if (t instanceof ConflictException) {
+      responder.sendString(HttpResponseStatus.CONFLICT, t.getMessage());
+    } else if (t instanceof NotFoundException) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND, t.getMessage());
+    } else if (t instanceof UnauthorizedException) {
+      responder.sendStatus(HttpResponseStatus.UNAUTHORIZED);
```
bdmogal: better to ``sendString()`` with ``t.getMessage()`` here as well?

cdap-common/src/main/java/co/cask/cdap/common/exception/HttpExceptionHandler.java
```diff
@@ -0,0 +1,52 @@
...
+    } else {
+      LOG.error("Unexpected error: request={} {} user={}:",
+                request.getMethod().getName(), request.getUri(),
+                SecurityRequestContext.getUserId().or("<null>"), t);
+      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
```
bdmogal: same here

bdmogal:
Minor comments, LGTM otherwise

cdap-common/src/main/java/co/cask/cdap/common/exception/HttpExceptionHandler.java
jwang47: Didn't want to send string here, since it could mistakenly reveal some protected information.

bdmogal: I see. ok.

PR #1487, opened 2015-03-04
remove confusing info log

rohitsinha54:
The stopping message can be confusing in ideal case when upgrade runs well as its basically completing and not stopping. Just getting rid of it.


albertshau:
:+1:

PR #1488, opened 2015-03-05
CDAP-1650 refactoring in memory dataset framework in preparation

albertshau:
for upcoming changes to v3 dataset apis. Also fixing bugs around
namespaces, where remove all and get all operations were across
all namespaces instead of local to a namespace.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
@@ -110,7 +110,7 @@ public class InMemoryDatasetFramework implements DatasetFramework {
       // keep track of default module classes. These are used when creating registries for other namespaces,
       // which need to register system classes too.
       String moduleClassName = DatasetModules.getDatasetModuleClass(module).getName();
-      Id.DatasetModule moduleId = Id.DatasetModule.from(Constants.SYSTEM_NAMESPACE, moduleName);
+      Id.DatasetModule moduleId = Id.DatasetModule.from(Constants.SYSTEM_NAMESPACE_ID, moduleName);
       moduleClasses.put(Constants.SYSTEM_NAMESPACE_ID, moduleId, moduleClassName);
     }
     registries.put(Constants.SYSTEM_NAMESPACE_ID, systemRegistry);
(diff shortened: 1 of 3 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT983

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
```diff
@@ -78,24 +85,61 @@ public InMemoryDatasetFramework(DatasetDefinitionRegistryFactory registryFactory
...
+    DatasetDefinitionRegistry systemRegistry = registryFactory.create();
+    for (Map.Entry<String, ? extends DatasetModule> entry : defaultModules.entrySet()) {
+      LOG.info("Adding Default module {} to system namespace", entry.getKey());
+      String moduleName = entry.getKey();
+      DatasetModule module = entry.getValue();
+      entry.getValue().register(systemRegistry);
+      // keep track of default module classes. These are used when creating registries for other namespaces,
+      // which need to register system classes too.
+      String moduleClassName = DatasetModules.getDatasetModuleClass(module).getName();
+      Id.DatasetModule moduleId = Id.DatasetModule.from(Constants.SYSTEM_NAMESPACE, moduleName);
```
bdmogal: nit: Perhaps use ``Constants.SYSTEM_NAMESPACE_ID``, so we don't create another ``Id.Namespace``

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/AbstractDatasetFrameworkTest.java
```diff
@@ -292,7 +294,144 @@ public void testNamespaces() throws DatasetManagementException {
...
+    Assert.assertEquals(1, specs.size());
+    Assert.assertEquals("table2", specs.iterator().next().getProperty("tag"));
+
+    // delete one instance and make sure the other still exists
+    framework.deleteInstance(table1ID);
+    Assert.assertFalse(framework.hasInstance(table1ID));
+    Assert.assertTrue(framework.hasInstance(table2ID));
+
+    // delete all instances in one namespace and make sure the other still exists
+    framework.deleteAllInstances(namespace1);
```
bdmogal: nit: This will be a no-op right? since there are no instances in namespace1? maybe add back the deleted instance before this call?

bdmogal:
@albertshau just a couple of minor nits. Thanks for the fix! :+1:

PR #1489, opened 2015-03-05
Fix/dashboard 1690

gaarf:
* fixes a bug when single user accesses dashboards in multiple namespaces
 * change metric-picker behavior to only display child context
 * support metric-picker validation via standard angular implements
 * improve widget footer CSS

cf https://issues.cask.co/browse/CDAP-1690


ajainarayanan:
:+1: LGTM

PR #1491, opened 2015-03-05
Remove root prefix from TableId class. Instead, HBaseTableUtil has a CCo...

anwar6953:
...nfiguration from which it derives the table prefix.

Build: http://builds.cask.co/browse/CDAP-DUT990-12

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTable.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTable.java
@@ -21,13 +21,15 @@ import co.cask.cdap.api.dataset.DataSetException;
 import co.cask.cdap.api.dataset.table.ConflictDetection;
 import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Scanner;
+import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.data2.dataset2.lib.table.BufferingTable;
 import co.cask.cdap.data2.dataset2.lib.table.IncrementValue;
 import co.cask.cdap.data2.dataset2.lib.table.PutValue;
 import co.cask.cdap.data2.dataset2.lib.table.Update;
+import co.cask.cdap.data2.util.TableId;
 import co.cask.cdap.data2.util.hbase.HBaseTableUtil;
 import co.cask.cdap.data2.util.hbase.HBaseTableUtilFactory;
-import co.cask.cdap.data2.util.hbase.TableId;
+import co.cask.cdap.data2.util.hbase.HTableNameConverter;
 import co.cask.tephra.Transaction;
 import co.cask.tephra.TransactionCodec;
 import com.google.common.base.Function;
@@ -71,13 +73,14 @@ public class HBaseTable extends BufferingTable {
 
   private final TransactionCodec txCodec;
 
-  public HBaseTable(String name, ConflictDetection level, Configuration hConf, boolean enableReadlessIncrements)
+  public HBaseTable(String name, ConflictDetection level, Configuration hConf, CConfiguration cConf,
+                    boolean enableReadlessIncrements)
     throws IOException {
     super(name, level, enableReadlessIncrements);
-    hTableName = HBaseTableUtil.getHBaseTableName(name);
-    HBaseTableUtil tableUtil = new HBaseTableUtilFactory().get();
+    hTableName = HTableNameConverter.getHBaseTableName(name);
+    HBaseTableUtil tableUtil = new HBaseTableUtilFactory().get(cConf);
     TableId tableId = TableId.from(name);
-    HTable hTable = tableUtil.getHTable(hConf, tableId);
+    HTable hTable = tableUtil.createHTable(hConf, tableId);
     // todo: make configurable
     hTable.setWriteBufferSize(HBaseTableUtil.DEFAULT_WRITE_BUFFER_SIZE);
     hTable.setAutoFlush(false);
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTable.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTable.java
@@ -21,14 +21,12 @@ import co.cask.cdap.api.dataset.DataSetException;
 import co.cask.cdap.api.dataset.table.ConflictDetection;
 import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Scanner;
-import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.data2.dataset2.lib.table.BufferingTable;
 import co.cask.cdap.data2.dataset2.lib.table.IncrementValue;
 import co.cask.cdap.data2.dataset2.lib.table.PutValue;
 import co.cask.cdap.data2.dataset2.lib.table.Update;
 import co.cask.cdap.data2.util.TableId;
 import co.cask.cdap.data2.util.hbase.HBaseTableUtil;
-import co.cask.cdap.data2.util.hbase.HBaseTableUtilFactory;
 import co.cask.cdap.data2.util.hbase.HTableNameConverter;
 import co.cask.tephra.Transaction;
 import co.cask.tephra.TransactionCodec;
@@ -73,12 +71,10 @@ public class HBaseTable extends BufferingTable {
 
   private final TransactionCodec txCodec;
 
-  public HBaseTable(String name, ConflictDetection level, Configuration hConf, CConfiguration cConf,
-                    boolean enableReadlessIncrements)
-    throws IOException {
+  public HBaseTable(String name, ConflictDetection level, Configuration hConf,
+                    HBaseTableUtil tableUtil, boolean enableReadlessIncrements) throws IOException {
     super(name, level, enableReadlessIncrements);
     hTableName = HTableNameConverter.getHBaseTableName(name);
-    HBaseTableUtil tableUtil = new HBaseTableUtilFactory().get(cConf);
     TableId tableId = TableId.from(name);
     HTable hTable = tableUtil.createHTable(hConf, tableId);
     // todo: make configurable
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTable.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTable.java
@@ -27,7 +27,6 @@ import co.cask.cdap.data2.dataset2.lib.table.PutValue;
 import co.cask.cdap.data2.dataset2.lib.table.Update;
 import co.cask.cdap.data2.util.TableId;
 import co.cask.cdap.data2.util.hbase.HBaseTableUtil;
-import co.cask.cdap.data2.util.hbase.HTableNameConverter;
 import co.cask.tephra.Transaction;
 import co.cask.tephra.TransactionCodec;
 import com.google.common.base.Function;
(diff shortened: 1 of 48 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -16,18 +16,113 @@
...
+import com.google.common.base.Preconditions;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+
 /**
  * Common utility methods for dealing with HBase table name conversions.
  */
 public abstract class HTableNameConverter {
+  protected static final String HBASE_NAMESPACE_PREFIX = "cdap_";
```
bdmogal: This should not be needed anymore right?

cdap-data-fabric/src/test/java/co/cask/cdap/data/hbase/HBaseTestBase.java
```diff
@@ -51,20 +49,18 @@ public HBaseAdmin getHBaseAdmin() throws IOException {
     return new HBaseAdmin(getConfiguration());
   }
 
-  // TODO: This method should be removed in favor of HBaseTableUtil#getHTable. Currently only used in Queue Tests
-  public HTable getHTable(byte [] tableName) throws IOException {
-    return new HTable(getConfiguration(), tableName);
-  }
+  // TODO: This method should be removed in favor of HBaseTableUtil#createHTable. Currently only used in Queue Tests
+    public HTable createHTable(byte[] tableName) throws IOException {
```
bdmogal: indentation?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
anwar6953: So this PR doesn't use the CConf for hbase namespace prefix, but only in the tablename (for backwards compatability).
It should be only few changes to do that though.

bdmogal: Worth doing it in this PR?

anwar6953: I found that getting PR merged takes quite a bit longer, the more things I try to fit in one PR.
I'll make a follow up PR to do this there.

bdmogal: sounds good.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtilFactory.java
```diff
@@ -16,10 +16,34 @@
...
+  }
+
+  @Override
+  protected HBaseTableUtil createInstance(String className) throws ClassNotFoundException {
+    HBaseTableUtil hBaseTableUtil = super.createInstance(className);
+    hBaseTableUtil.setCConf(cConf);
+    return hBaseTableUtil;
+  }
+
+  public static Class<? extends HBaseTableUtil> getHBaseTableUtilClass() {
```
chtyim: Why need this static method?

anwar6953: There's 4-5 places in the code that only need the class of the tableUtil, not the actual instance.
And some of those, the cConf is not available.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -16,18 +16,113 @@
...
+
+  public static String getHBaseTableName(String tableName) {
```
chtyim: Why this method is public?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -16,18 +16,113 @@
...
+    return encodeTableName(tableName);
+  }
+
+  public static String encodeTableName(String tableName) {
```
chtyim: Why this is public?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -16,18 +16,113 @@
...
+    try {
+      return URLEncoder.encode(tableName, "ASCII");
+    } catch (UnsupportedEncodingException e) {
+      // this can never happen - we know that ASCII is a supported character set!
+      throw new RuntimeException(e);
+    }
+  }
+
+  public static String getBackwardCompatibleTableName(String tablePrefix, TableId tableId) {
```
chtyim: Why public?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -16,18 +16,113 @@
...
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+
 /**
  * Common utility methods for dealing with HBase table name conversions.
  */
 public abstract class HTableNameConverter {
```
chtyim: Feels like many methods of this class should be package private? Ideally they are not used outside of the HBaseTableUtil and CoProcessor (some method may need to be public because of that, but we should document it clearly that methods of this class shouldn't be used outside).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
anwar6953: made private.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
anwar6953: made private

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
anwar6953: Ok, I removed the usages of it outside the package (this method will be applied when they call the publicly exposed methods), and made it private.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
anwar6953: Updated.
The remaining public method is:
`  public abstract String getSysConfigTablePrefix(String hTableName);`

Used from code in coprocessors.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtilFactory.java
chtyim: Better be a default constructor. In general, try to avoid static as much as possible.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTable.java
```diff
@@ -65,19 +64,18 @@
...
-    super(name, level, enableReadlessIncrements);
-    hTableName = HBaseTableUtil.getHBaseTableName(name);
-    HBaseTableUtil tableUtil = new HBaseTableUtilFactory().get();
-    TableId tableId = TableId.from(name);
-    HTable hTable = tableUtil.getHTable(hConf, tableId);
+  public HBaseTable(String tableName, ConflictDetection level, Configuration hConf,
+                    HBaseTableUtil tableUtil, boolean enableReadlessIncrements) throws IOException {
+    super(tableName, level, enableReadlessIncrements);
+    this.tableName = tableName;
+    TableId tableId = TableId.from(tableName);
```
chtyim: This will go away as well, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtil.java
```diff
@@ -92,20 +92,14 @@
...
-  private static String encodeTableName(String tableName) {
-    try {
-      return URLEncoder.encode(tableName, "ASCII");
-    } catch (UnsupportedEncodingException e) {
-      // this can never happen - we know that ASCII is a supported character set!
-      LOG.error("Error encoding table name '" + tableName + "'", e);
-      throw new RuntimeException(e);
-    }
+  protected CConfiguration cConf;
+  private String tableNamePrefix;
```
chtyim: I don't see it's used.

chtyim:
Mostly good. Just couple comments left.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTable.java
anwar6953: Yes. Bhooshan has that in his PR.
https://github.com/caskdata/cdap/pull/1495/files#diff-3940118a7982e1cceeaa023c99b1b35f

cdap-hbase-compat-0.96/src/main/java/co/cask/cdap/data2/util/hbase/HBase96TableUtil.java
```diff
@@ -138,7 +139,19 @@ public void modifyTable(HBaseAdmin admin, HTableDescriptor tableDescriptor) thro
...
     Preconditions.checkArgument(tableId != null, "Table Id should not be null.");
-    return admin.getTableRegions(toTableName(tableId));
+    return admin.getTableRegions(HTable96NameConverter.toTableName(cConf, tableId));
+  }
+
+  @Override
+  public void deleteAllInNamespace(HBaseAdmin admin, Id.Namespace namespaceId, String tablePrefix) throws IOException {
+    TableName[] tableNames = admin.listTableNamesByNamespace(HTableNameConverter.toHBaseNamespace(namespaceId));
+    for (TableName tableName : tableNames) {
+      String name = HTable96NameConverter.fromTableName(tableName).getTableName();
```
chtyim: Shouldn't do this. Why not use the `tableNamespace()`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -16,18 +16,113 @@
...
+   * @param hTableName Full HBase table name.
    * @return System configuration table prefix (full table name minus the table qualifier).
    * Example input: "cdap_ns.table.name"  -->  output: "cdap_system."   (hbase 94)
    * Example input: "cdap.table.name"     -->  output: "cdap_system."   (hbase 94. input table is in default namespace)
    * Example input: "cdap_ns:table.name"  -->  output: "cdap_system:"   (hbase 96, 98)
    */
-  public abstract String getSysConfigTablePrefix(String tableName);
+  public abstract String getSysConfigTablePrefix(String hTableName);
+
+  protected abstract TableId fromTableName(String hTableName);
```
chtyim: Why have this? I don't think we should have this. The signature is different for different HBase version, right? And it is only used within the HBase compat module, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtilFactory.java
anwar6953: I did not want the code using this to think that the factory could be instantiated without a parameter (because in that case, it is missing a CConf, and so operations on it won't be valid).

cdap-hbase-compat-0.96/src/main/java/co/cask/cdap/data2/util/hbase/HBase96TableUtil.java
anwar6953: the tableName that the admin returns is the hbase table name.
So I convert it to the corresponding cdap table name, and then check for prefix match.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
anwar6953: Ok, I removed it for now. But the idea was that this could be used from non-hbase-specific code, via the TableNameConverterFactory.

chtyim:
LGTM

PR #1492, opened 2015-03-05
[CDAP-1488] Stream Size Schedule Persistent Store

jguery:
Persistent store for stream size schedules, so that they don't disappear whenever CDAP restarts.

Build: http://builds.cask.co/browse/CDAP-DUT989


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
```diff
@@ -122,6 +125,19 @@ public void start() {
       Threads.createDaemonThreadFactory("polling-booking-executor"));
 
     store = storeFactory.create();
+
+    scheduleStore.initialize();
+
+    // Reschedule the persisted schedules
+    synchronized (this) {
```
anew: why is this synchronized? To prevent new schedules being added or existing ones being updated/deleted while the service is starting up?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
```diff
@@ -749,6 +766,14 @@ public void startSchedule(long basePollSize, long basePollTs, long lastRunSize,
       this.lastRunSize = lastRunSize;
       this.lastRunTs = lastRunTs;
       this.active.set(active);
+      if (persist) {
+        scheduleStore.persist(programId, programType, streamSizeSchedule, basePollSize, basePollTs,
```
anew: I am not sure why you persist here. If basePollSize etc. are -1, then there is nothing to persist, and if they are not -1, then that is because you just read from the store. If I understand correctly.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
jguery: Yes, that was the point, although now that you mention it I think this can't happen, since we have to wait till the scheduler services starts up before doing any operation.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
jguery: There are two cases here, yes: either I just read the schedule info from the store, and then ``persist`` will be false and I won't store anything, or we create a new schedule. In that last case, basePollSize and Ts will be initialized with the values obtained from a polling - see ``createScheduleTask()``. This schedule needs to be persisted right at its creation, hence doing it here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
```diff
@@ -122,6 +125,42 @@ public void start() {
...
+   */
+  private void initializeScheduleStore() throws SchedulerException {
+    try {
+      scheduleStore.initialize();
+    } catch (Throwable t) {
+      throw new SchedulerException("Error when initializing store for stream size schedules", t);
+    }
+
+    // Reschedule the persisted schedules
+    List<DatasetBasedStreamSizeScheduleStore.StreamSizeScheduleState> scheduleStates;
```
anew: why is StreamSizeScheduleState an inner class of the store? Isn't it more closely tied to a ScheduleTask?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
```diff
@@ -493,11 +532,11 @@ public void createScheduleTask(Id.Program programId, SchedulableProgramType prog
                                    StreamSizeSchedule streamSizeSchedule, boolean active,
                                    long basePollSize, long basePollTs, long lastRunSize, long lastRunTs,
                                    boolean persist) throws SchedulerException {
```
anew: perhaps you should rename the parameter "persist" to something else. It does not indicate whether a schedule is persisted (all schedules are persisted, right?). It indicates that this is the first time the schedule is created. Or if false, it indicates that we're in the startup sequence and we are reinstating the schedules that were previously persisted.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
```diff
@@ -506,32 +545,44 @@ public void createScheduleTask(Id.Program programId, SchedulableProgramType prog
...
+          newTask.startSchedule(basePollSize, basePollTs, lastRunSize, lastRunTs, active, false);
         }
 
+        // We only modify the scheduleTasks if the persistence in startSchedule() did not throw any exception
+        scheduleTasks.put(scheduleId, newTask);
+
         if (active) {
           activeTasks.incrementAndGet();
         }
       }
```
anew: I am not sure whether I like using the same method for creating a new schedule and for restoring the state after a restart of the scheduler. In case of a restart, you can simply read all schedules from the store once, the pollOnce and send to all active tasks. If you separate this out into separate methods, it becomes much easier to understand.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
```diff
@@ -741,7 +799,7 @@ private StreamSizeScheduleTask(Id.Program programId, SchedulableProgramType prog
      * @param persist {@code true} if this information should be persisted, {@code false} otherwise
      */
     public void startSchedule(long basePollSize, long basePollTs, long lastRunSize, long lastRunTs,
-                              boolean active, boolean persist) {
+                              boolean active, boolean persist) throws SchedulerException {
```
anew: similar to my above comment, it is much clearer if you separate the code for restoring an existing schedule task vs. creating a new one

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
```diff
@@ -749,6 +807,29 @@ public void startSchedule(long basePollSize, long basePollTs, long lastRunSize,
...
+        }
+      }
+    }
+
+    /**
+     * Remove the schedule task information from the persistent store.
+     *
+     * @throws SchedulerException when the task could not be deleted from the store
+     */
+    public void deleteSchedule() throws SchedulerException {
```
anew: Can you call this deleteFromStore?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
```diff
@@ -784,43 +865,103 @@ public void receivedPollingInformation(@Nonnull StreamSize pollingInfo) {
...
+          scheduleStore.updateLastRun(programId, programType, streamSizeSchedule.getName(),
+                                      pollingInfo.getSize(), pollingInfo.getTimestamp());
+          lastRunSize = pollingInfo.getSize();
+          lastRunTs = pollingInfo.getTimestamp();
+        } catch (Throwable t) {
+          LOG.error("Error when persisting last run information for schedule {} in store",
+                    streamSizeSchedule.getName(), t);
+          return;
+        }
+        // The taskRunner.run() call will wait for the program to be run entirely, and can't be wrapped in
```
anew: really? That seems wrong to me. It should only submit the program for execution, not wait for it. What is the point of waiting?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedStreamSizeScheduleStore.java
```diff
@@ -0,0 +1,380 @@
...
+        public void apply() throws Exception {
+          byte[] rowKey = Bytes.toBytes(String.format("%s:%s", KEY_PREFIX,
+                                                      getScheduleId(programId, programType, scheduleName)));
+          table.put(rowKey, columns, values);
+          LOG.debug("Updated schedule {} with columns {}, values {}", scheduleName, columns, values);
+        }
+      });
+  }
+
+  private String getScheduleId(Id.Program program, SchedulableProgramType programType, String scheduleName) {
```
anew: Is this a duplication of AbstractSchedulerService.scheduleIdFor()?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedStreamSizeScheduleStore.java
```diff
@@ -0,0 +1,380 @@
...
+    return String.format("%s:%s", getProgramScheduleId(program, programType), scheduleName);
+  }
+
+  private String getProgramScheduleId(Id.Program program, SchedulableProgramType programType) {
```
anew: And this one, too?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedStreamSizeScheduleStore.java
```diff
@@ -0,0 +1,380 @@
...
+    return String.format("%s:%s:%s:%s", program.getNamespaceId(), program.getApplicationId(),
+                         programType.name(), program.getId());
+  }
+
+  /**
+   * POJO containing a {@link StreamSizeSchedule} and its state.
```
anew: This should be a separate class, not an inner class.

anew:
@jguery please see my comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedStreamSizeScheduleStore.java
jguery: Yes, it is, will remove.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedStreamSizeScheduleStore.java
jguery: Yes, will remove.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
jguery: Ok.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
jguery: Alright, I will create two paths for those two scenarios.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
jguery: Will do.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedStreamSizeScheduleStore.java
jguery: Ok.

jguery:
@anew I addressed your comments, please have another look. Thank you

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/StreamSizeScheduler.java
jguery: ``TaskExecutionException`` is thrown when the program can't be found in the store, or if it is already running and the option to set concurrent runs is not on, or if waiting for the program to finish has been interrupted.

jguery: After talking to @chtyim, it seems like the ``taskRunner.run`` method was originally blocking to satisfy the Quartz scheduler requirements. But I will change it to return a ``ListenableFuture``, that the time scheduler will wait for, but not the data scheduler.

anew:
Only one comment and LGTM

PR #1493, opened 2015-03-05
Refactor usage of TransactionExecutorFactory and Log saver tests

chtyim:
- We already have a Guice binding for TransactionExecutorFactory, no need to create one every time
- We should use Guice injection as much as possible when they are present. Log saver related tests are not doing it right.


chtyim:
Build running in http://builds.cask.co/browse/CDAP-DUT985-1

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/MDSStreamMetaStore.java
```diff
@@ -61,33 +58,28 @@
   private Transactional<StreamMds, MetadataStoreDataset> txnl;
 
   @Inject
-  public MDSStreamMetaStore(CConfiguration conf, final TransactionSystemClient txClient, DatasetFramework framework) {
+  public MDSStreamMetaStore(CConfiguration conf,
```
albertshau: while we're changing these constructors, can we check if we can change them to private to prevent direct creation?

albertshau:
one question, otherwise lgtm

chtyim:
I have diverted view myself on whether we should make `@Inject` constructor private or not (flexibility vs encapsulation. For some classes, there are cases that needs direct instantiation, but maybe we can make it protected/package private). We probably need to go through all Guice modules/bindings in one pass to fix the general Guice usage in CDAP.

chtyim:
Created https://issues.cask.co/browse/CDAP-1698 for tracking the general Guice issue.

chtyim:
Build passed. Merging.

PR #1497, opened 2015-03-05
Added ability to get the information about the current Workflow run base...

sagarkapare:
...d on run id


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1355,57 +1356,7 @@ private StatusMap getStatus(final Id.Program id, final ProgramType type) {
...
-
-            // This would make all changes in the result statusMap available to the other thread that doing
-            // the take() call.
-            statusFuture.set(result);
-          }
-        }
-      );
-      // wait for status to come back in case we are polling mapreduce status in workflow
-      // status map contains either a status or an error
-      return Futures.getUnchecked(statusFuture);
```
sagarkapare: Deleted this code since it assumes the MAPREDUCE program running inside WORKFLOW is named as WORKFLOW_MAPREDUCE and then returns the status of the WORKFLOW instead of the MAPREDUCE program. In 2.7 we moved MAPREDUCE to application level, so their names no longer contain the WORKFLOW name. This will be fixed with jira - https://issues.cask.co/browse/CDAP-1293

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -997,13 +997,14 @@ public void deleteFlowQueues(HttpRequest request, HttpResponder responder,
 
   /**************************** Workflow/schedule APIs *****************************************************/
   @GET
-  @Path("/apps/{app-id}/workflows/{workflow-name}/current")
+  @Path("/apps/{app-id}/workflows/{workflow-name}/{run-id}/current")
```
jwang47: Should this be `../{run-id}/status`?

sagarkapare: The purpose of the API is to get the information about what programs are currently in execution for the run specified by the run-id. This is different from the `status` which only returns whether the current program is `RUNNING` or `STOPPED`

jwang47:
LGTM

sagarkapare:
Thank you @jwang47 for the review! Build is running here - http://builds.cask.co/browse/CDAP-DUT993-2. Will merge once it passes.

sreevatsanraman:
LGTM :+1:

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT993-4 Merging.

PR #1498, opened 2015-03-05
Feature/ui data modelling

ajainarayanan:
Initial shot at abstracting the socket usage in cdap.

- Adds ```MySocketProvider``` that behaves exactly the same as ```$resource```. 
- Modifies ```WorkFlows``` detailed view to see if it actually works.
- Modifies ```Namespace Creation``` to use MySocketProvider to check if ```PUT/POST``` would work with a proper body and params.
- Adds ```MyWorkflowsApi & MyNamespaceApi``` factories that could be used across different controllers (the same way we use $resource).

PROS:
  - If in production we hit a point where sockets are not working out for us then changing the current implementation to use ```$resource``` would be easier.

Current MySocketProvider implementation: 

EDIT: NOT VALID ANYMORE:
```
angular.module(PKG.name + '.services')
  .factory('myNamespaceApi', function(MySocketProvider, $state, myCdapUrl) {
    var url = myCdapUrl.constructUrl;
    
    return MySocketProvider(
      url({
        _cdapPath: '/namespaces/:namespaceId'
      }),
      {
        namespaceId: '@namespaceId'
      },
      {
        create: {
          method: 'PUT'
        }
      }
    )
  });
```

~~What it can be modified to:~~

EDIT: This will be the actual implementation.  For reference: https://github.com/caskdata/cdap/blob/feature/ui-data-modelling/cdap-ui/app/services/workflows/myworkflowapi.js
```
angular.module(PKG.name + '.services')
  .factory('myNamespaceApi', function($resource, $state, myCdapUrl) {
    var url = myCdapUrl.constructUrl;
    
    return $resource(
      url({
        _cdapPath: '/namespaces/:namespaceId'
      }),
      {
        namespaceId: '@namespaceId'
      },
      {
        create: {
          method: 'PUT'
        }
      }

    )
  });
```

Switching from MySocketProvider to $resource should easier.

- We can add polling to the ```$resource``` setup by a minor fix that would mimic something like this,

Existing MySocketProvider polling:
```
    myWorkFlowApi.status(params, function(res) {
      $scope.status = res.status;
    });
```
$resource polling:
```
  (function pollStatus () {
      myWorkFlowApi.status(params, function(res) {
        $scope.status = res.status;
        $timeout(pollStatus);
    });
  })()
```

CONS:
 - ```MySocketProvider``` is nothing but $resource without ```$http```. So any upgrades that goes into ```$resource``` would not be easily ported to MySocketProvider


cdap-ui/app/services/data/socket-provider.js
```diff
@@ -0,0 +1,304 @@
...
+              action.url);
+
+
+            httpConfig.url = buildUrl(httpConfig.url, httpConfig.params);
+
+            if (!requestType || requestType === 'request') {
+              dataSrc.request(httpConfig, success);
+            } else if (requestType === 'poll') {
+              dataSrc.poll(httpConfig, success);
+            }
```
gaarf: can we make this a `switch` statement ?

cdap-ui/app/services/data/socket-provider.js
```diff
@@ -0,0 +1,304 @@
...
+var MEMBER_NAME_REGEX = /^(\.[a-zA-Z_$][0-9a-zA-Z_$]*)+$/;
+
+function isValidDottedPath(path) {
+  return (path != null && path !== '' && path !== 'hasOwnProperty' &&
+      MEMBER_NAME_REGEX.test('.' + path));
+}
+
+function lookupDottedPath(obj, path) {
+  if (!isValidDottedPath(path)) {
+    throw $resourceMinErr('badmember', 'Dotted member path "@{0}" is invalid.', path);
```
gaarf: where is our `$resourceMinErr` defined ?

gaarf:
@ajai awesome work :trophy:

elfenheart:
@ajainarayanan this is awesome dude! :+1:

ajainarayanan:
Alright. Finally an update.

Changes:
  - No change to $resource library. We should not directly modify a library. We have a decorator pattern which allows us to modify any external library we want by adding/removing some functionality from it in a more cleaner way.
  - Implemented ```MyPromise``` 
     - To use ```$resource``` as ```$resource``` primarily use promises.
     - To give a promise type api which is also an observable.
  - Fixed MyDataSource appropriately to factor in the above changes.


PROS:
  - Enables a more cleaner use of ```$resource``` in a socket based environment 
  - Enables easier transition to ```$resource``` if sockets doesn't work out.

CONS: 
  - MyPromise is a minimal implementation of any Promise pattern. If we need more utilities (like ```when``` or ```all```) then we need to spend more time on it. This is on the assumption that if ```$resource``` uses any of these in later versions.
  - Does some processing of response data. We should try avoid it. This is mainly to be compatible with ```$resource```


TODO:
- Not sure exactly why but if we poll for the same resource from parent and a child state then each callback (both in parent and child) gets called back twice. The reason I believe is that we are not able to differentiate between callbacks as we invoke callbacks based on resource(url + json + method). 
- Cleanup MyDataSource to be more readable.
- MUST HAVE TEST CASES FOR ```MyPromise``` AND ```MyDataSource``` ASAP.

cdap-ui/app/services/mypromise.js
```diff
@@ -0,0 +1,183 @@
...
+    @param {function} which gets a 'resolve' and a 'reject' methods
+    @param {boolean} is observable or not.
+
+    PS: Inspired from
+      - https://www.promisejs.org/implementing/
+      - https://github.com/kriskowal/q/blob/v1/design/README.js
+*/
+angular.module(PKG.name + '.services')
+  .provider('MyPromise', function() {
+    var PENDING = 0;
```
nistala: how about using an object:
var states = {
  PENDING: 0...
}

cdap-ui/app/services/data/datasource.js
```diff
@@ -115,54 +164,129 @@ angular.module(PKG.name+'.services')
...
+        mySocket.send({
+          action: 'request',
+          resource: generatedResource
+        });
+
+      }, false);
+      if (!resource.$isResource) {
+        prom = prom.then(function(res) {
+          res = res.data;
+          return res;
```
nistala: why return here?

cdap-ui/app/services/mypromise.js
ajainarayanan: Not sure if I see the advantage here. Any specific reason it needs to be inside an object?

cdap-ui/app/services/data/datasource.js
ajainarayanan: This is so that it is compatible with the $resource object. It needs to return the value to be used by the subsequent ```then``` handler functions.

elfenheart:
:+1: LGTM

Let's merge so we can start playing with it

ajainarayanan:
Closing this as we have ##$!@ load of things to do. This will be taken up later once feature request and styling requests stop to some extent.

PR #1500, opened 2015-03-05
CDAP-1629: Use default namespace when current namespace is deleted in CLI

jwang47:
https://issues.cask.co/browse/CDAP-1629

This is what happens when you delete the current namespace now:

```
cdap (http://127.0.0.1:10000/default)> create namespace sdf
Namespace 'sdf' created successfully.

cdap (http://127.0.0.1:10000/default)> use namespace sdf
Now using namespace 'sdf'

cdap (http://127.0.0.1:10000/sdf)> delete namespace sdf
Namespace 'sdf' deleted successfully.
Now using namespace 'default'

cdap (http://127.0.0.1:10000/default)> 
```

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/DeleteNamespaceCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/DeleteNamespaceCommand.java
@@ -17,8 +17,10 @@
 package co.cask.cdap.cli.command;
 
 import co.cask.cdap.cli.ArgumentName;
+import co.cask.cdap.cli.CLIConfig;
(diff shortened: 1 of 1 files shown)


jwang47:
http://builds.cask.co/browse/CDAP-DUT995

anwar6953:
`Switched back to default namespace` seems a bit imperfect.
How about `Now using Default namespace`?

Not a big deal, but the message seems unusual for some reason.

jwang47:
Yeah, it does sound awkward. Will change.

anwar6953:
LGTM

PR #1502, opened 2015-03-05
Require angular motion

gaarf:
Explicitly import `angular-motion` CSS via bower. It used to be required by `angular-strap`, but no more.


ajainarayanan:
:+1: LGTM

PR #1504, opened 2015-03-05
CDAP-1707: Fix AltStyleTableRenderer when there are long headers

jwang47:
https://issues.cask.co/browse/CDAP-1707

### DIFF
--- a/cdap-cli/src/test/java/co/cask/cdap/cli/util/table/TableRendererTest.java
+++ b/cdap-cli/src/test/java/co/cask/cdap/cli/util/table/TableRendererTest.java
@@ -78,4 +78,17 @@ public abstract class TableRendererTest {
       .build();
     getRenderer().render(System.out, table);
   }
+
+  @Test
+  public void testLongHeader() {
+    Table table = Table.builder()
+      .setHeader("c1", "c2", "c3333veryveryveryveryveryveryveryveryveryveryvery long header")
+      .setRows(Table.rows()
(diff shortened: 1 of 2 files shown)


jwang47:
http://builds.cask.co/browse/CDAP-DUT999

cdap-cli/src/test/java/co/cask/cdap/cli/util/table/TableRendererTest.java
```diff
@@ -78,4 +78,17 @@ public void testTwoLineCell2() {
       .build();
     getRenderer().render(System.out, table);
   }
+
+  @Test
+  public void testLongHeader() {
```
jguery: How do we actually make sure that the header stays in the table? I don't see any assertions? Is it just a visual test?

jguery:
One comment, otherwise LGTM.

cdap-cli/src/test/java/co/cask/cdap/cli/util/table/TableRendererTest.java
jwang47: Yeah, this test is not meant to be a real test.

jguery: Ok, that's fine then, if all the tests of that class are like that.

jwang47:
All tests passed except for tests in one unrelated module.

PR #1505, opened 2015-03-05
adds flow graph visualization to app workflows

astewes:


cdap-ui/app/services/helpers.js
```diff
@@ -97,12 +97,62 @@ angular.module(PKG.name+'.services')
...
+    for(var i=0; i<nodes.length; i++) {
+      if (nodes[i].nodeType === 'ACTION') {
+        expandedNodes.push(nodes[i]);
+      } else if (nodes[i].nodeType === 'FORK') {
+        for (var j=0; j<nodes[i].branches.length; j++) {
+          expandForks(nodes[i].branches[j], expandedNodes);
+        }
+      }
+    }
+  }
```
gaarf: these 3 functions are very specific to the API being used... IMHO they should not live in `myHelpers`. For now I would put them in the controller file that uses them, or if they are going to be re-used, in a different service. Later this belongs in a `myWorkflowsApi`.

cdap-ui/app/services/helpers.js
```diff
@@ -97,12 +97,62 @@ angular.module(PKG.name+'.services')
...
+
+  function flatten(source, fork, target, connections) {
+    // Flatten a source-fork-target combo to [list of connections]
+    var branches = fork.branches,
+        temp = [];
+
+    for (var i =0; i<branches.length; i++) {
+      temp = branches[i];
+      temp.unshift(source);
+      temp.push(target);
```
gaarf: use `Array.splice` to do both operations in one statement.

cdap-ui/app/features/workflows/controllers/tabs/runs/flow-ctrl.js
```diff
@@ -1,5 +1,5 @@
 angular.module(PKG.name + '.feature.workflows')
```
nistala: should call this file workflow-ctrl

cdap-ui/app/services/helpers.js
ajainarayanan: As in like this? ```a.splice(0, 0, 0).splice(a.length-1, 0 , 2)```

gaarf: you're right, splice is useless here. how about:
`temp = [source].concat(branches[i], [target]);`

cdap-ui/app/features/workflows/controllers/tabs/runs/status-ctrl.js
```diff
@@ -0,0 +1,192 @@
+angular.module(PKG.name + '.feature.workflows')
+  .controller('WorkflowsDetailRunStatusController', function($state, $scope, MyDataSource, amMoment, $filter) {
+    var dataSrc = new MyDataSource($scope),
+        filterFilter = $filter('filter'),
+        basePath = '/apps/' + $state.params.appId + '/workflows/' + $state.params.programId;
+    $scope.moment = amMoment;
+    $scope.moment.changeLocale('en');
```
gaarf: I'm assuming that would change the locale for all amMoment instances throughout the app. What is it for, why is it here?

gaarf:
is `$scope.moment` used anywhere ? Also the utility functions can go outside of the controller's closure, I can show you how.

gaarf:
LGTM :+1:

PR #1507, opened 2015-03-05
Feature/ui integration tests

elfenheart:
Initial integration test

Management screen: testing sidebars


gaarf:
shouldn't this PR include adding `protractor` do the package.json ? else how would one run the integration suite?

gaarf:
LGTM :+1:

PR #1509, opened 2015-03-06
Id.Stream to store Id.Namespace instead of String

anwar6953:
https://issues.cask.co/browse/CDAP-1486
http://builds.cask.co/browse/CDAP-DUT1003-2

### DIFF
--- a/cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
+++ b/cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
@@ -79,7 +79,7 @@ public final class Id  {
 
     @Override
     public int hashCode() {
-      return Objects.hashCode(id);
+      return id.hashCode();
     }
 
     public static Namespace from(String namespace) {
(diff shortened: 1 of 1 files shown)


bdmogal:
LGTM pending build.

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
```diff
@@ -383,18 +383,17 @@ public int hashCode() {
    * Id.Stream uniquely identifies a stream.
    */
   public static final class Stream {
-    private final String namespace;
+    private final Namespace namespace;
```
chtyim: this affect the efficiency of the computation of hashCode for `Id.Stream`. You'd like to change the `Id.Namespace` hashCode computation to be efficient as well.

anwar6953: We cache the hashcode for StreamId, so the hashcode will only be computed once for the object.

That should be fine, right

chtyim: no. The computation itself need to be fast, as you still computing it on every event (from the StreamHandler, you construct a new Id.Stream object and that will be used to lookup a Map, and that's when the hashCode is computed on the newly created Id.Stream object. The cache only saves the one in the Map already).

anwar6953: ok, made Namespace's hashcode just the hashcode of the string it consists of.

chtyim:
LGTM

PR #1510, opened 2015-03-06
CDAP-1140: Enable metrics support for integration test framework

jwang47:
https://issues.cask.co/browse/CDAP-1140
http://builds.cask.co/browse/CDAP-DUT1005

Very hacky, but I couldn't find any clean alternatives that didn't require a complete refactoring of all our test code.


cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsConstants.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package co.cask.cdap.common.metrics;
+
+/**
+ * Constants for metrics.
+ */
+public interface MetricsConstants {
+
+  public static final String MAPREDUCE_INPUT = "system.process.tuples.read";
```
gsps1: same as flowlet , looks incorrect ?

cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsConstants.java
```diff
@@ -0,0 +1,39 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.common.metrics;
+
+/**
+ * Constants for metrics.
+ */
+public interface MetricsConstants {
```
gsps1: why is this an interface ? , haven't seen such usage before. is it normal ?

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -50,21 +61,142 @@ public MetricsClient(ClientConfig config) {
...
+  public MetricQueryResult query(Map<String, String> context, String metric, @Nullable String groupBy)
+    throws IOException, UnauthorizedException {
+
+    URL url = config.resolveURLV3(String.format("metrics/query?context=%s&metric=%s%s",
+                                                contextToPathParam(context), metric,
+                                                groupBy == null ? "" : "&groupBy=" + groupBy));
+    HttpResponse response = restClient.execute(HttpMethod.POST, url, config.getAccessToken());
+    return ObjectResponse.fromJsonBody(response, MetricQueryResult.class).getResponseObject();
+  }
+
```
gsps1: should we add a method for metrics search as well ?

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -50,21 +61,142 @@ public MetricsClient(ClientConfig config) {
...
+  public RuntimeMetrics getMapReduceMetrics(Id.Program id) {
+    return getMetrics(MetricsContext.forMapReduce(id), MAPREDUCE_INPUT, MAPREDUCE_PROCESSED, MAPREDUCE_EXCEPTIONS);
```
gsps1: MAPREDUCE_EXCEPTIONS - not declared anywhere.

cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsConstants.java
jwang47: Just to simplify using it. I think there's probably a better place to put this though, but I don't know it.

gsps1: Typically we have it as final class, 
https://github.com/caskdata/cdap/blob/develop/cdapwatchdog/src/main/java/co/cask/cdap/metrics/MetricsConstants.java

jwang47: Should we move cdap-watchdog's MetricsConstants to cdap-common, so we can combine these two?

gsps1: no, they are different, that file contains config keys and default values used while creating metrics tables, while this is more like standard metric-name constants .

cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
```diff
@@ -31,16 +41,34 @@
...
+    // TODO: remove arbitrary sleep
+    TimeUnit.SECONDS.sleep(5);
+
+    Id.Application appId = Id.Application.from(Constants.DEFAULT_NAMESPACE_ID, FakeApp.NAME);
+    Id.Program programId = Id.Program.from(appId, ProgramType.FLOW, FakeFlow.NAME);
+    String flowlet = FakeFlow.FLOWLET_NAME;
+
+    MetricQueryResult result = metricsClient.query(MetricsContext.forFlowlet(programId, flowlet),
+                                                   MetricsConstants.FLOWLET_INPUT, null);
+    Assert.assertEquals(1, result.getSeries()[0].getData()[0].getValue());
```
gsps1: very less coverage. just add a todo to increase tests here.

gsps1:
took a pass, as you mentioned there are duplication between RuntimeStats and MetricsClient class, apart from that we should add searchContext and searchMetrics in MetricsClient, otherwise add todo and bring back resolveURL for now, as we will need to use MetricsClient to test search api's.  also it would be good to get @abaranau's suggestions about reducing duplication.

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
jwang47: I'd prefer to add that in another PR since it isn't necessary for https://github.com/caskdata/cdap/pull/1567

gsps1: ok, but you will add that for 2.8 right ?

jwang47: Yes, we'll have to.

gsps1:
LGTM

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -16,22 +16,33 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 
 /**
  * Provides ways to interact with CDAP Metrics.
  */
-public class MetricsClient {
+public class MetricsClient implements MetricsConstants {
```
abaranau: hah?

abaranau: typo? What does it mean to implement constants?

cdap-api/src/main/java/co/cask/cdap/api/metrics/RuntimeMetrics.java
```diff
@@ -14,7 +14,7 @@
  * the License.
  */
 
-package co.cask.cdap.test;
+package co.cask.cdap.api.metrics;
```
abaranau: In general I hate the idea of using this ill-designed RuntimeMetrics class as an API for metrics. It raised questions like: how do I access different (or custom) metric in test? How do I wait on it? Etc. We want to revisit the design.

We have multiple jiras for better access to metrics in unit-tests and such - It is fine to revisit this later.

abaranau: can we mark it as @ Beta though - since we are going to change it

abaranau: https://issues.cask.co/browse/CDAP-765 add a note with todo

cdap-app-fabric/src/main/java/co/cask/cdap/test/RuntimeStats.java
```diff
@@ -34,11 +40,13 @@
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import javax.annotation.Nullable;
 
 /**
  *
  */
-public final class RuntimeStats {
+public final class RuntimeStats implements MetricsConstants {
```
chtyim: Anti-pattern. Shouldn't define constants in interface. `Constants` class should be final class with private constructor.

cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsConstants.java
abaranau: Let's move it into watchdog-api (see also https://issues.cask.co/browse/CDAP-1794). Let's name it MetricNames or smth: we already have MetricsConstancts class.. Let's make is final class

cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsContext.java
```diff
@@ -0,0 +1,59 @@
...
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.proto.Id;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+
+/**
+ * Provides ways to obtain the context for certain metrics.
+ */
+public class MetricsContext {
```
abaranau: move to watchdog-api, make final, call MetricContexts

abaranau:
couple comments. Thank you for adding this whole thing

cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsContext.java
jwang47: couldn't move to watchdog-api b/c this requires `Constants.Metrics.Tag` which is in cdap-common

cdap-api/src/main/java/co/cask/cdap/api/metrics/RuntimeMetrics.java
abaranau: todo + link?

cdap-app-fabric/src/main/java/co/cask/cdap/test/RuntimeStats.java
```diff
@@ -50,34 +58,31 @@ public static void resetAll() throws Exception {
     metricStore.deleteBefore(System.currentTimeMillis() / 1000);
   }
 
+  public static RuntimeMetrics getMapReduceMetrics(String applicationId, String mapReduceId) {
```
abaranau: does wait for works for mapreduce? does any unit-test or any test uses this method? Do you know if it works?

jwang47: Currently untested.. unfortunately. Should I remove this change?

abaranau:
one comment l3eft

abaranau:
please add JIRA (if haven't already) for mr metrics and merge

albertshau:
build is basically good, flaky tests failing. going to go ahead and merge.

PR #1511, opened 2015-03-06
inject on streams, better flow diagrams

nistala:
- inject into streams
- use variables in flow diagram / calculate positions dynamically / center align 

https://issues.cask.co/browse/CDAP-1679

### DIFF
--- a/cdap-ui/app/directives/flow-graph/flow.js
+++ b/cdap-ui/app/directives/flow-graph/flow.js
@@ -80,7 +80,7 @@ module.directive('myFlowGraph', function (d3, dagreD3, $state, $filter) {
         // Draw the flowlet shape.
         renderer.shapes().flowlet = function(parent, bbox, node) {
           var instances = getInstances(node.elem.__data__); // No other way to get name from node.
-
+          var instaceCircleScaled = getInstancesScaledRadius(instances, instanceCircleRadius);
           var shapeSvg = parent.insert('circle', ':first-child')
             .attr('x', -bbox.width / 2)
             .attr('y', -bbox.height / 2)
@@ -89,13 +89,13 @@ module.directive('myFlowGraph', function (d3, dagreD3, $state, $filter) {
 
           // Elements are positioned with respect to shapeSvg.
           parent.insert('circle')
-            .attr('cx', flowletCircleRadius - instanceCircleRadius)
+            .attr('cx', flowletCircleRadius - instaceCircleScaled)
             .attr('cy', -instanceBufferHeight)
-            .attr('r', instanceCircleRadius)
+            .attr('r', instaceCircleScaled)
             .attr('class', 'flow-shapes flowlet-instances');
 
           parent.insert('text')
-            .attr('x', flowletCircleRadius - instanceCircleRadius)
+            .attr('x', flowletCircleRadius - instaceCircleScaled)
             .attr('y', -instanceBufferHeight + metricCountPadding)
             .text(instances)
             .attr('class', 'flow-shapes flowlet-instance-count');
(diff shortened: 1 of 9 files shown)


cdap-ui/app/directives/flow-graph/flow.js
```diff
@@ -30,16 +30,26 @@ module.directive('myFlowGraph', function (d3, dagreD3, $state, $filter) {
...
 
         var instanceMap = {};
+        var labelMap = {};
 
         var bytesFilter = $filter('bytes');
 
         var renderer = new dagreD3.render();
         var g = new dagreD3.graphlib.Graph();
+        var metricCircleRadius = 25;
+        var instanceCircleRadius = 10;
```
elfenheart: This might create a problem when the number of instance exceeds 2 digits?

cdap-ui/app/filters/bytes.js
```diff
@@ -3,13 +3,13 @@
...
-      return '0 bytes';
+      return '0b';
     }
     if (typeof precision === 'undefined') {
       precision = bytes>1023 ? 1 : 0;
     }
     var number = Math.floor(Math.log(bytes) / Math.log(1024));
     return (bytes / Math.pow(1024, Math.floor(number))).toFixed(precision) +
-      ' ' + ['bytes', 'kB', 'MB', 'GB', 'TB', 'PB'][number];
+      '' + ['b', 'kB', 'MB', 'GB', 'TB', 'PB'][number];
```
elfenheart: Need to modify the unit tests for bytes filter to work with this change

cdap-ui/app/directives/flow-graph/flow.js
```diff
@@ -184,7 +199,14 @@ module.directive('myFlowGraph', function (d3, dagreD3, $state, $filter) {
          * Handles node click and sends to flowlet page.
          */
         function handleNodeClick(nodeId) {
-          $state.go('flows.detail.runs.detail.flowlets.detail', {flowletId: nodeId});
+          handleHideTip(nodeId);
+          var instance = instanceMap[nodeId];
+          if (instance.type === 'STREAM') {
+            $state.go('flows.detail.runs.tabs.status.streamsDetail', {streamId: nodeId});
+          } else {
+            $state.go('flows.detail.runs.tabs.status.flowletsDetail', {flowletId: nodeId});
```
ajainarayanan: This was causing issues with workflow diagrams. This directive has elements specific for flows. It has to handle this state change a little bit more dynamic so that we can use it in workflows too. (On node click on a workflow diagram should probably take it to a workflows child state)

cdap-ui/app/directives/flow-graph/flow.js
```diff
@@ -54,6 +64,7 @@ module.directive('myFlowGraph', function (d3, dagreD3, $state, $filter) {
         angular.forEach(nodes, function (node) {
           var nodeLabel = node.name.length > 8 ? node.name.substr(0, 5) + '...' : node.name;
           instanceMap[node.name] = node;
+          labelMap[nodeLabel] = node;
           if (node.type === 'STREAM') {
```
ajainarayanan: Same as the other comment. In case of workflows we have the node type as 'ACTION' or 'FORK'..

ajainarayanan:
- Not sure if we should have the shapes in the same directive (flows have one type of shapes and workflows have different ones).
- Not sure how to make it more generic so that it could be used for both flow diagram & workflows diagram. In the context of navigating to child states on individual node clicks. (Should we create separate directives for both?)

cdap-ui/app/directives/flow-graph/flow.js
nistala: This won't draw a workflow diagram, i thought those were different. How could I change this to work for workflows?

cdap-ui/app/filters/bytes.js
nistala: done

cdap-ui/app/directives/flow-graph/flow.js
nistala: scales correctly now

nistala:
Based on the mocks, we have different shapes for flows and workflows so we should probably have different directives if we don't want to end up with too many forks (if/else, case-switch)

ajainarayanan:
:+1: LGTM

elfenheart:
:+1: LGTM too

PR #1512, opened 2015-03-06
[CDAP-1711] add support for text format to (partitioned) file sets

anew:
title says it all.


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/FileSetProperties.java
```diff
@@ -128,6 +139,31 @@ public static boolean isExploreEnabled(Map<String, String> properties) {
...
+
+  /**
+   * @return the schema of the explore table.
+   */
+  public static String getExploreSchema(Map<String, String> properties) {
+    return properties.get(PROPERTY_EXPLORE_SCHEMA);
+  }
+
+  /**
+   * @return the schema of the explore table.
```
albertshau: update javadoc

anew: oh yeah, sorry.

albertshau:
minor comment, otherwise lgtm

PR #1514, opened 2015-03-06
Feature/hbase namespaces integration queues

anwar6953:
Use HBase namespaces for queues/stream tables.

Resolves JIRAs:
https://issues.cask.co/browse/CDAP-1594
https://issues.cask.co/browse/CDAP-1670
Build: http://builds.cask.co/browse/CDAP-DUT908-49


cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
```diff
@@ -425,12 +427,24 @@ public Long getResult() {
...
   /**
-   * Gets a table name for stream consumer state stores within a given namespace.
+   * Gets a TableId for stream consumer state stores within a given namespace.
+   * @param namespace the namespace for which the table is for.
+   * @return constructed TableId
+   */
+  public static TableId getStateStoreTableId(Id.Namespace namespace) {
+    String tableName = String.format("%s.%s.state.store",
+                                     Constants.SYSTEM_NAMESPACE, QueueConstants.QueueType.STREAM.toString());
+    return TableId.from(namespace.getId(), tableName);
```
chtyim: Why the `TableId.from` doesn't take `Id.Namespace` (or at least have a overloaded method).

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
```diff
@@ -425,12 +427,24 @@ public Long getResult() {
...
+  }
+
+  /**
+   * Gets the table name for stream consumer state stores within a given namespace.
    * @param namespace the namespace for which the table is for.
    * @return constructed table name
    */
   public static String getStateStoreTableName(Id.Namespace namespace) {
```
chtyim: Should get rid of this method and use the `TableId` one instead.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueAdmin.java
```diff
@@ -16,46 +16,25 @@
...
  * Common implementation of table-based QueueAdmin
  */
 public abstract class AbstractQueueAdmin implements QueueAdmin {
-  private final String unqualifiedTableNamePrefix;
-  private final String unqualifiedConfigTableNameSuffix;
-  protected final DefaultDatasetNamespace namespace;
+  protected final QueueConstants.QueueType type;
+  protected final String unqualifiedTableNamePrefix;
+  // system.queue.config'
+  public static final String CONFIG_TABLE_NAME =
```
chtyim: Why public?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
anwar6953: this is used in levelDB and inmemory.
it delegates to TableId one.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
anwar6953: Will add one.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueAdmin.java
anwar6953: its used to construct the queue config table in QueueRegionObserver.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
chtyim: They should not be using it. Just make them call the TableId one.

chtyim: Each of them should do their own logic to translate TableId into something it needs (e.g. for leveldb, it can use namespace as directory and name as sub-directory, while for in-memory, it just concat ns-name and use it as a String to lookup a map).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueAdmin.java
```diff
@@ -78,38 +57,32 @@ public static String getFlowName(String queueTableName) {
     return parts[parts.length - 1];
   }
 
+  public static  TableId getConfigTableId(QueueName queueName) {
```
chtyim: I think only HBase implementation has config table. Why it's needed in AbstractQueueAdmin?

anwar6953: Right, I just put it here, since getConfigTableName was here.
Moved to HBaseQueueAdmin.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
anwar6953: Since the LevelDBTableService is used for more than just streams/queues (And its a larger change, I prefer to do that in a separate PR).
Will file a JIRA.

chtyim: What I mean is, it's up to particular implementation to decide what to do with the TableId. We should get rid of this method.

anwar6953: Removed it:
https://github.com/caskdata/cdap/commit/9ade94aee90e68d395472a4990e160ff7fff93d8

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/QueueAdmin.java
```diff
@@ -102,11 +102,4 @@
    */
   void drop(String name) throws Exception;
 
-  /**
-   * Performs update of entity.
-   *
-   * @param name Name of the entity to update
-   * @throws Exception if update fails
-   */
-  void upgrade(String name, Properties properties) throws Exception;
```
bdmogal: upgrade goes away now?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -119,6 +124,14 @@ protected HBaseQueueAdmin(Configuration hConf,
     this.locationFactory = locationFactory;
   }
 
+  public static TableId getConfigTableId(QueueName queueName) {
```
bdmogal: why do these methods have to be static

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -192,21 +205,17 @@ public void truncate(String name) throws Exception {
...
-  private void truncate(byte[] tableNameBytes) throws IOException {
+  private void truncate(TableId tableId) throws IOException {
     HBaseAdmin admin = getHBaseAdmin();
-    if (admin.tableExists(tableNameBytes)) {
-      HTableDescriptor tableDescriptor = admin.getTableDescriptor(tableNameBytes);
-      admin.disableTable(tableNameBytes);
-      admin.deleteTable(tableNameBytes);
-      admin.createTable(tableDescriptor);
+    if (tableUtil.tableExists(admin, tableId)) {
+      tableUtil.truncateTable(getHBaseAdmin(), tableId);
```
bdmogal: inconsistency - either use ``admin`` or ``getHBaseAdmin``, not both

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -234,27 +241,15 @@ public void drop(String name) throws Exception {
...
-      dsAdmin.close();
-    }
-  }
-
-  private void drop(byte[] tableName) throws IOException {
-    HBaseAdmin admin = getHBaseAdmin();
-    if (admin.tableExists(tableName)) {
-      admin.disableTable(tableName);
-      admin.deleteTable(tableName);
+  private void drop(HBaseAdmin admin, TableId tableId) throws IOException {
```
bdmogal: this method shouldn't need admin, no? ``getHBaseAdmin()`` should be available here?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -367,8 +362,10 @@ private void createConfigTable(String configTableName) throws IOException {
   @Override
   public void dropAllInNamespace(String namespaceId) throws Exception {
     // Note: The trailing "." is crucial, since otherwise nsId could match nsId1, nsIdx etc
-    dropTablesWithPrefix(String.format("%s.", getTableNamePrefix(namespaceId)));
-    drop(Bytes.toBytes(getConfigTableName(namespaceId)));
+    String trueTableNamePrefix = String.format("%s.", unqualifiedTableNamePrefix);
```
bdmogal: ``queueTableNamePrefix``?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -482,60 +479,51 @@ public void configureGroups(QueueName queueName, Map<Long, Integer> groupInfo) t
...
         }
-        upgrade(tableName, properties);
-        LOG.info(String.format("Upgraded queue hbase table: %s", tableName));
+        upgrade(tableId, properties);
+        LOG.info(String.format("Upgraded queue table: %s", tableId));
       }
     }
   }
 
+  private void upgrade(TableId tableId, Properties properties) throws Exception {
```
bdmogal: why does this need a separate method?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -482,60 +479,51 @@ public void configureGroups(QueueName queueName, Map<Long, Integer> groupInfo) t
...
    * or tables for things other than queues).
    */
-  private boolean isDataTable(String tableName) {
-    // checks if table is constructed by getActualTableName(String)
+  private boolean isDataTable(TableId tableId) {
+    // checks if table is constructed by getDataTableId
+    String tableName = tableId.getTableName();
     String[] parts = tableName.split("\\.");
-    if (parts.length != 6) {
+    if (parts.length != 4) {
```
bdmogal: hmmm, can a flow id have a '.'? That should have broken even earlier, but just making sure

bdmogal: is it better to do a check for config table instead? since its name cannot be influenced by users? not sure

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
```diff
@@ -42,35 +44,37 @@
...
     this.queueAdmin = (HBaseQueueAdmin) queueAdmin;
     this.streamAdmin = streamAdmin;
     this.queueUtil = new HBaseQueueUtilFactory().get();
+    this.hBaseTableUtil = hBaseTableUtil;
   }
 
   // for testing only
-  String getTableName(QueueName queueName) {
-    return (queueName.isStream() ? streamAdmin : queueAdmin).getActualTableName(queueName);
+  TableId getTableId(QueueName queueName) {
```
bdmogal: if its indeed only for testing, remove comment and add ``@VisibleForTesting``

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
```diff
@@ -42,35 +44,37 @@
...
+    return (queueName.isStream() ? streamAdmin : queueAdmin).getDataTableId(queueName);
   }
 
   // for testing only
-  String getConfigTableName(QueueName queueName) {
-    return (queueName.isStream() ? streamAdmin : queueAdmin).getConfigTableName(queueName);
+  TableId getConfigTableId(QueueName queueName) {
```
bdmogal: same here

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
```diff
@@ -42,35 +44,37 @@
...
+    return HBaseQueueAdmin.getConfigTableId(queueName);
   }
 
   @Override
   public QueueConsumer createConsumer(QueueName queueName,
                                       ConsumerConfig consumerConfig, int numGroups) throws IOException {
     HBaseQueueAdmin admin = ensureTableExists(queueName);
-    HTable configTable = createHTable(admin.getConfigTableName(queueName));
+    HTable configTable = createHTable(HBaseQueueAdmin.getConfigTableId(queueName));
```
bdmogal: this is why it needs to be static? maybe this can be refactored so we don't need this static method. Create a jira and add a TODO?

bdmogal: actually, what do you feel about having this be a non-static method in HBaseQueueAdmin. All it needs is the namespace, which is available in the ``queueName``.

Although it may mean that ``HBaseQueueAdmin`` - the admin for a data table knows about the config table as well? Not sure.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
```diff
@@ -42,35 +44,37 @@
...
-  String getConfigTableName(QueueName queueName) {
-    return (queueName.isStream() ? streamAdmin : queueAdmin).getConfigTableName(queueName);
+  TableId getConfigTableId(QueueName queueName) {
+    return HBaseQueueAdmin.getConfigTableId(queueName);
   }
 
   @Override
   public QueueConsumer createConsumer(QueueName queueName,
                                       ConsumerConfig consumerConfig, int numGroups) throws IOException {
     HBaseQueueAdmin admin = ensureTableExists(queueName);
```
bdmogal: nit: maybe this should be called ``createIfNotExists``

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/leveldb/LevelDBQueueAdmin.java
```diff
@@ -136,13 +137,8 @@ public void drop(String name) throws Exception {
...
 
   @Override
-  public void upgrade(String name, Properties properties) throws Exception {
-    // No-op
-  }
-
-  @Override
   public void dropAllInNamespace(String namespaceId) throws Exception {
-    dropAllTablesWithPrefix(String.format("%s.", getTableNamePrefix(namespaceId)));
+    dropAllTablesWithPrefix(String.format("%s.%s", namespaceId, unqualifiedTableNamePrefix));
```
bdmogal: missing '.' in the end?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/leveldb/LevelDBStreamConsumerStateStoreFactory.java
```diff
@@ -30,26 +39,38 @@
...
-    String tableName = StreamUtils.getStateStoreTableName(namespace);
-    tableService.ensureTableExists(tableName);
-    LevelDBTableCore coreTable = new LevelDBTableCore(tableName, tableService);
+    TableId tableId = StreamUtils.getStateStoreTableId(namespace);
+
+    getLevelDBTableAdmin(tableId).create();
+    String levelDBTableName =
+      PrefixedNamespaces.namespace(cConf, tableId.getNamespace().getId(), tableId.getTableName());
+    LevelDBTableCore coreTable = new LevelDBTableCore(levelDBTableName, tableService);
     return new LevelDBStreamConsumerStateStore(streamConfig, coreTable);
```
bdmogal: I feel that ``LevelDBStreamConsumerStateStore`` should take a ``LevelDBTable`` and not a ``LevelDBCoreTable``. That way you don't need to create the namespaced name here. ``InMemoryStreamConsumerStateStore`` and ``HBaseStreamConsumerStateStore`` follow a similar pattern.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/leveldb/LevelDBStreamConsumerStateStoreFactory.java
```diff
@@ -30,26 +39,38 @@
  * Factory for creating {@link StreamConsumerStateStore} in level db.
  */
 public final class LevelDBStreamConsumerStateStoreFactory implements StreamConsumerStateStoreFactory {
-
   private final LevelDBTableService tableService;
+  private final CConfiguration cConf;
 
   @Inject
-  LevelDBStreamConsumerStateStoreFactory(LevelDBTableService tableService) {
+  LevelDBStreamConsumerStateStoreFactory(LevelDBTableService tableService, CConfiguration cConf) {
```
bdmogal: nit: conf should be the first param, consistent with ``HBaseStreamConsumerStateStore``

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtil.java
```diff
@@ -93,10 +94,14 @@
...
 
-  protected CConfiguration cConf;
+  protected String tablePrefix;
 
   public void setCConf(CConfiguration cConf) {
-    this.cConf = cConf;
+    this.tablePrefix = cConf.get(Constants.Dataset.TABLE_PREFIX);
+  }
+
+  protected boolean isCDAPTable(String hTableName) {
```
bdmogal: so this is an operation on the full table string? hmmm, better to delegate to version specific classes with ``TableId``? Not sure.

bdmogal: Any operations on the full table name string make me a little wary.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -87,40 +83,62 @@ protected static String getHBaseTableName(CConfiguration cConf, TableId tableId)
...
 
     // Id.Namespace already checks for non-null namespace
-    return TableId.from(namespace, hTableName);
+    return PrefixedTableId.from(prefix, namespace, hTableName);
+  }
+
+  /**
+   * Used internal to HTableNameConverter, so that one parsing method can extract both the prefix and TableId.
+   */
+  protected static final class PrefixedTableId {
```
bdmogal: why protected if its internal?

cdap-data-fabric/src/test/java/co/cask/cdap/data/hbase/HBaseTestBase.java
```diff
@@ -49,11 +47,6 @@ public HBaseAdmin getHBaseAdmin() throws IOException {
     return new HBaseAdmin(getConfiguration());
   }
 
-  // TODO: This method should be removed in favor of HBaseTableUtil#createHTable. Currently only used in Queue Tests
```
bdmogal: Thanks :)

cdap-data-fabric/src/test/java/co/cask/cdap/data2/increment/hbase/AbstractIncrementHandlerTest.java
```diff
@@ -73,10 +79,10 @@ public static void tearDown() throws Exception {
 
   @Test
   public void testIncrements() throws Exception {
-    String tableName = "incrementTest";
-    createTable("incrementTest");
+    TableId tableId = TableId.from(Constants.DEFAULT_NAMESPACE, "incrementTest");
```
bdmogal: failing to understand why this class had to change. can you link me to the failure?

cdap-data-fabric/src/test/java/co/cask/cdap/data2/util/hbase/AbstractHBaseTableUtilTest.java
```diff
@@ -167,28 +169,29 @@ public void testTableSizeMetrics() throws Exception {
...
-    Assert.assertEquals("default", HTableNameConverter.toHBaseNamespace(tableId.getNamespace()));
+    Assert.assertEquals("default", HTableNameConverter.toHBaseNamespace(tablePrefix, tableId.getNamespace()));
     Assert.assertEquals("cdap.user.my.dataset",
-                        HTableNameConverter.getHBaseTableName(cConf, tableId));
+                        HTableNameConverter.getHBaseTableName(tablePrefix, tableId));
     Assert.assertEquals(getTableNameAsString(tableId),
                         Bytes.toString(tableUtil.createHTable(testHBase.getConfiguration(), tableId).getTableName()));
     drop(tableId);
-    tableId = TableId.from("cdap.default.system.queue.config");
+    tableId = TableId.from("default", "system.queue.config");
```
bdmogal: Thanks :)

bdmogal:
Some comments. Also wondering why the Increment Handler tests had to change.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtil.java
anwar6953: Regardless of version, the full table name either begins with cdap. or cdap_
(no point in copying the same implementation to version specific methods).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
anwar6953: version-specific subclasses utilize it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtil.java
bdmogal: It should only be moved to subclasses if we can operate on ``TableId``. 

I'm only afraid that we may end up using this for non-fully qualified table names (without namespace, perhaps?)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
anwar6953: used by QueueRegionObserver.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtil.java
bdmogal: Actually, I have another suggestion. It maybe a good idea to accept an ``HTable`` here. That should solve the problem of users calling it for non-fully qualified names. You can do htable.getDescriptor.getName then.

anwar6953:
Moved Increment Handler test changes to a separate PR. Reason is there:
https://github.com/caskdata/cdap/pull/1528

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/QueueAdmin.java
anwar6953: There's an no parameter `upgrade()`upgrade method, which upgrades all tables that the queueAdmin is responsible for.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
anwar6953: I just figured it'd be good to separate the logic of determining which tables need upgrading and then the actual table upgrade (even though its only 5 lines).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
anwar6953: You mean `if (!isConfigTable(tableName)`?

I don't want to upgrade all other tables, simply because its not a config table (could be *any* other table created by cdap).

anwar6953: And no, flow id can not have '.'
FlowVerification.java, line 63.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
anwar6953: It needs to be static because its used in QueueRegionObserver.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
bdmogal: Ok. ignore this comment then.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/increment/hbase/AbstractIncrementHandlerTest.java
anwar6953: https://github.com/caskdata/cdap/pull/1528

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
bdmogal: I see. I ignored the usage in QueueRegionObserver. So this is a substitute for QueueUtils.determineConfigTableName(), right. If so, again, ignore my comment.

anwar6953: correct; removed QueueUtils.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
anwar6953: Isn't there a difference - its not just non-private for the sake of testing, it exists only for testing?
`@VisibleForTesting` gives the impression that its used elsewhere, but non-private for testing reasons.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HBaseTableUtil.java
anwar6953: Ok, I made it accept HTableDescriptor instead. Thats what admin.listTables returns.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
anwar6953: both method names seem to have the same meaning to me.
I'll leave it as it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
bdmogal: ignore for now, I feel if it is only required for testing, and is not even used anywhere in the main code base, we should probably move it to a base class. 

Not something you introduced though, maybe we can fix later.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/leveldb/LevelDBStreamConsumerStateStoreFactory.java
bdmogal: Ignore this comment right now, but please create a jira to make this change and add a TODO.

bdmogal:
LGTM, pending build and testing.

PR #1517, opened 2015-03-07
CDAP-1715: Upgrade tool now reads the old metatable again

rohitsinha54:
Fixes the CDAP-1715 and enables the upgrade tool to read the old meta data tables again 

Build : http://builds.cask.co/browse/CDAP-DUT1019-1

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/config/DefaultConfigStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/config/DefaultConfigStore.java
@@ -31,6 +31,7 @@ import co.cask.cdap.data2.dataset2.tx.Transactional;
 import co.cask.cdap.proto.Id;
 import co.cask.tephra.TransactionExecutor;
 import co.cask.tephra.TransactionExecutorFactory;
+import com.google.common.base.Joiner;
 import com.google.common.base.Supplier;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Iterators;
@@ -80,7 +81,10 @@ public class DefaultConfigStore implements ConfigStore {
   }
 
   public static void setupDatasets(DatasetFramework dsFramework) throws DatasetManagementException, IOException {
-    dsFramework.addInstance(Table.class.getName(), configStoreDatasetInstanceId, DatasetProperties.EMPTY);
+    dsFramework.addInstance(Table.class.getName(), Id.DatasetInstance.from(
+                              Constants.DEFAULT_NAMESPACE_ID, Joiner.on(".").join(Constants.SYSTEM_NAMESPACE,
+                                                                                  Constants.ConfigStore.CONFIG_TABLE)),
+                            DatasetProperties.EMPTY);
   }
 
   @Override
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/ScheduleStoreTableUtil.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/ScheduleStoreTableUtil.java
@@ -19,10 +19,12 @@ package co.cask.cdap.internal.app.runtime.schedule.store;
 import co.cask.cdap.api.dataset.DatasetProperties;
 import co.cask.cdap.api.dataset.table.Table;
 import co.cask.cdap.common.conf.CConfiguration;
+import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.data2.dataset2.DatasetFramework;
 import co.cask.cdap.data2.dataset2.DatasetManagementException;
 import co.cask.cdap.data2.dataset2.lib.table.MetaTableUtil;
 import co.cask.cdap.proto.Id;
+import com.google.common.base.Joiner;
 import com.google.inject.Inject;
 
 import java.io.IOException;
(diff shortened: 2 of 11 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/ScheduleStoreTableUtil.java
```diff
@@ -46,11 +48,13 @@ public String getMetaTableName() {
 
   /**
    * Adds datasets and types to the given {@link DatasetFramework} used by schedule mds.
+   *
    * @param datasetFramework framework to add types and datasets to
    */
   public static void setupDatasets(DatasetFramework datasetFramework) throws IOException, DatasetManagementException {
     Id.DatasetInstance scheduleStoreDatasetInstance =
-      Id.DatasetInstance.from(SYSTEM_NAMESPACE, SCHEDULE_STORE_DATASET_NAME);
+      Id.DatasetInstance.from(Constants.DEFAULT_NAMESPACE_ID, (Joiner.on(".").join(Constants.SYSTEM_NAMESPACE,
```
albertshau: is this used anywhere besides the upgrade tool?  The javadoc indicates it is, which would make this problematic

cdap-app-fabric/src/main/java/co/cask/cdap/config/DefaultConfigStore.java
```diff
@@ -80,7 +81,10 @@ public ConfigTable get() {
   }
 
   public static void setupDatasets(DatasetFramework dsFramework) throws DatasetManagementException, IOException {
-    dsFramework.addInstance(Table.class.getName(), configStoreDatasetInstanceId, DatasetProperties.EMPTY);
+    dsFramework.addInstance(Table.class.getName(), Id.DatasetInstance.from(
```
albertshau: if this is all we're doing, do we need this method at all?  Nothing here is specific to DefaultConfigStore right?

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -52,6 +54,7 @@
   private final QueueAdmin queueAdmin;
   private final HBaseTableUtil hBaseTableUtil;
   private final DatasetFramework namespacedFramework;
+  private static final Pattern USER_TABLE_PREFIX = Pattern.compile("cdap\\.user\\..*");
```
albertshau: is it always guaranteed to match this pattern?

albertshau:
actually looking at it, all of those public static methods could be removed right?

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -93,45 +96,40 @@ private void upgradeSystemDatasets(DatasetFramework framework) throws Exception
...
-          @Override
-          public void create() throws IOException {
-            // no-op
-            throw new UnsupportedOperationException("This DatasetAdmin is only used for upgrade() operation");
-          }
-        };
-        admin.upgrade();
-        LOG.info("Upgraded hbase table: {}", tableName);
-      }
+      HTableNameConverter hTableNameConverter = new HTableNameConverterFactory().get();
```
albertshau: im a bit confused here. Won't this expect the table name to conform to 2.8 conventions whereas we expect these to be older names?

cdap-hbase-compat-0.94/src/main/java/co/cask/cdap/data2/util/hbase/HTable94NameConverter.java
```diff
@@ -31,6 +31,11 @@ public String getSysConfigTablePrefix(String hTableName) {
     return HBASE_NAMESPACE_PREFIX + Constants.SYSTEM_NAMESPACE + ".";
   }
 
+  @Override
+  public TableId from(String hTableName) {
```
albertshau: check with @anwar6953 to see if this is how we want to do it.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
bdmogal: ``HTableNameConverterFactory#get()`` returns a version-specific ``HTableNameConverter``, which has parsing logic for its specific version, so this should work.

rohitsinha54:
@albertshau : Yes we can do that. But I guess the idea was to have them in the Store which they add. Do we want to take them out ?

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
bdmogal: Actually mis-understood your question. This should be ok, since ``HTableNameConverter`` has logic for maintaining backward-compatibility, etc specific to an HBase version.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
bdmogal: Yes, user tables until 2.7 always had the ``cdap.user.`` prefix. However, I think if there is a way to to indicate starts with (^ maybe), we should do that, to be doubly sure.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -86,6 +86,14 @@ protected static String getHBaseTableName(CConfiguration cConf, TableId tableId)
    */
   public abstract String getSysConfigTablePrefix(String hTableName);
 
+  /**
+   * Returns the {@link TableId} for the given hbase tablename
+   * Note: Sub-classes should override this method for their version specific implementation.
```
anwar6953: I think its fine to remove this Note.

cdap-hbase-compat-0.94/src/main/java/co/cask/cdap/data2/util/hbase/HTable94NameConverter.java
anwar6953: Yea, this looks fine.

rohitsinha54: Thanks @anwar6953

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
rohitsinha54: @bdmogal ?

anwar6953: I simply don't think its needed because it should be obvious - the method is marked `abstract`.

albertshau:
lgtm, as this is just fixing the bug. we can deal with general framework issues separately

rohitsinha54:
Build :green_heart:  here: http://builds.cask.co/browse/CDAP-DUT1019-3
Merging.

PR #1518, opened 2015-03-07
CDAP-1628 add cdap version as a property to all hive tables in

albertshau:
preparation for upgrade work. Also did some refactoring to move
logic around building the create statement for hive tables into
one class.

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/executor/ExploreExecutorHttpHandler.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/executor/ExploreExecutorHttpHandler.java
@@ -116,7 +116,7 @@ public class ExploreExecutorHttpHandler extends AbstractHttpHandler {
         return;
       }
 
-      LOG.debug("Enabling explore for stream {} at location {}", streamName, streamLocation);
+      LOG.debug("Enabling explore for stream {} at location {}", streamName, streamLocation.toURI());
       String createStatement;
       try {
         createStatement = generateStreamCreateStatement(streamId, streamLocation,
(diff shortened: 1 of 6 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT1021

cdap-explore/src/main/java/co/cask/cdap/explore/executor/ExploreExecutorHttpHandler.java
```diff
@@ -105,26 +101,25 @@ public void enableStream(HttpRequest request, HttpResponder responder,
...
         }
-        streamLocationURI = streamLocation.toURI().toString();
       } catch (IOException e) {
         LOG.info("Could not find stream {} to enable explore on.", streamName, e);
         responder.sendString(HttpResponseStatus.NOT_FOUND, "Could not find stream " + streamName);
         return;
       }
 
-      LOG.debug("Enabling explore for stream {} at location {}", streamName, streamLocationURI);
+      LOG.debug("Enabling explore for stream {} at location {}", streamName, streamLocation);
```
bdmogal: I remember seeing ``Location.toString()`` returning the object reference as output. Can you please check?

bdmogal:
Should we also change the table names for streams and datasets or is that in a different PR?

bdmogal:
Just 1 minor comment and one question. LGTM otherwise.

cdap-explore/src/main/java/co/cask/cdap/explore/executor/ExploreExecutorHttpHandler.java
albertshau: o yes changed the type without changing this

albertshau:
names will be done in a separate PR

PR #1521, opened 2015-03-07
CDAP-1722 Fix/namespace name

elfenheart:
angular initialize $scope.model.name with empty string and it's causing the backend to process name to empty string.


sreevatsanraman:
LGTM :+1:

PR #1522, opened 2015-03-07
(CDAP-1648) Remove deprecated instances endpoint for services

ParthGandhi:
Addresses: https://issues.cask.co/browse/CDAP-1648

Steps 2 and 3 of the JIRA have already been done during the separation of workers and services by @gokulavasan. The only problem was that the old endpoints also got transferred to v3.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -66,8 +66,6 @@ import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.Lists;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.SettableFuture;
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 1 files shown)


ParthGandhi:
http://builds.cask.co/browse/CDAP-DUT1022-1/

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1097,16 +1095,8 @@ public void getServiceInstances(HttpRequest request, HttpResponder responder,
...
-   * Return the number of instances for the given runnable of a service.
-   */
-  @GET
-  @Path("/apps/{app-id}/services/{service-id}/runnables/{runnable-name}/instances")
-  public void getServiceInstances(HttpRequest request, HttpResponder responder,
-                                  @PathParam("namespace-id") String namespaceId,
-                                  @PathParam("app-id") String appId,
-                                  @PathParam("service-id") String serviceId,
-                                  @PathParam("runnable-name") String runnableName) {
+  void getServiceInstances(HttpRequest request, HttpResponder responder,
```
bdmogal: why are these methods here if the APIs are not required? maybe move to the v2 handlers?

bdmogal:
Just one question, LGTM otherwise. 

Btw, this API wasn't working as of a couple of days ago. Had a task on me to investigate. Not needed anymore?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anwar6953: yea it seems like the body of this method should be in the v2 handler.

ParthGandhi: Both the methods are used in endpoints in the current class - to support the new endpoint for services.

I didn't think it made sense to refactor/break-down further since it's going to change in 3.0 anyway. Thoughts?

ParthGandhi:
I wasn't aware of these endpoints not working - the unit and integration tests have been passing. Link me to the JIRA?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anwar6953: Oh ok - so the service handler is sticking around in v3, but simply the workers are being removed.
And so this method is still used for the handler runnable.

bdmogal:
Was planning on addressing it as part of [CDAP-1091](https://issues.cask.co/browse/CDAP-1091)

bdmogal:
Lets merge this. Fix the services endpoint as another jira. :+1:

PR #1523, opened 2015-03-07
[cdap 1718] Fix transacting of dataset operations in MR runner

anew:


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -396,26 +394,17 @@ public void apply(MapReduceContext mapReduceContext) throws Exception {
   // since the user code may create new TransactionAwares, we need to do this ourselves instead
   // of going through tephra's TransactionExecutor, since the executor requires that you know all the txAwares
   // before executing.
```
albertshau: this comment should be removed

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -396,26 +394,17 @@ public void apply(MapReduceContext mapReduceContext) throws Exception {
...
-  private void runUserCodeInTx(TransactionExecutor.Procedure<MapReduceContext> userCode, String methodName) {
+  private void runUserCodeInTx(TransactionExecutor.Subroutine userCode, String methodName) {
     // add datasets given in the application specification to the transaction context
```
albertshau: this comment should be removed

albertshau:
minor comment on a comment. lgtm

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
anew: oh, right, it does not apply any more...

anew:
Thanks for the review, Albert. 

I also added a unit test. It uses a "pedantic" dataset that checks that beforeSubmit() and getSplits() are called inside the same transaction. That fails on previous develop but succeeds in this branch. 

Will merge as soon as the new build passes.

anew:
build is running here: http://builds.cask.co/browse/CDAP-DUT1016-3

PR #1525, opened 2015-03-07
AppMeta table upgrade

rohitsinha54:
- Upgrades app.meta table 
- Also upgrades stream meta for streams which does not belong to any app.

Issues: 
- CDAP-1510
- CDAP-1579

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
@@ -59,7 +59,7 @@ public class DatasetUpgrader extends AbstractUpgrader {
   @Inject
   private DatasetUpgrader(CConfiguration cConf, Configuration hConf, LocationFactory locationFactory,
                           QueueAdmin queueAdmin, HBaseTableUtil hBaseTableUtil,
-                          @Named("namespacedDSFramework") DatasetFramework namespacedFramework) {
+                          @Named("dsFramework") DatasetFramework dsFramework) {
 
     super(locationFactory);
     this.cConf = cConf;
@@ -67,7 +67,7 @@ public class DatasetUpgrader extends AbstractUpgrader {
     this.locationFactory = locationFactory;
     this.queueAdmin = queueAdmin;
     this.hBaseTableUtil = hBaseTableUtil;
-    this.namespacedFramework = namespacedFramework;
+    this.namespacedFramework = dsFramework;
   }
 
   @Override
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
@@ -233,7 +233,7 @@ public class MDSUpgrader extends AbstractUpgrader {
                                   curCompletedRunRecord.getStartTs());
           store.setStop(Id.Program.from(Id.Application.from(Constants.DEFAULT_NAMESPACE, appId), programType,
                                         programId), curCompletedRunRecord.getPid(), curCompletedRunRecord.getStopTs(),
-                        getControllerStateByStatus(curCompletedRunRecord.getStatus()));
+                        curCompletedRunRecord.getStatus());
         }
         return null;
       }
@@ -282,27 +282,6 @@ public class MDSUpgrader extends AbstractUpgrader {
 
     return locationFactory.create(Constants.DEFAULT_NAMESPACE).append(
       cConf.get(Constants.AppFabric.OUTPUT_DIR)).append(appId).append(Constants.ARCHIVE_DIR).append(archiveFilename);
-  }
-
-  /**
-   * Gives the {@link ProgramController.State} for a given {@link ProgramRunStatus}
-   * Note: This will given the state as {@link ProgramController.State#STARTING} for {@link ProgramRunStatus#RUNNING}
-   * even though running has multiple mapping but that is fine in our case as we use this
-   * {@link ProgramController.State} to write the runRecordCompleted through {@link DefaultStore#setStop} which converts
-   * it back to {@link ProgramRunStatus#RUNNING}. So we don't really care about the temporary intermediate
-   * {@link ProgramController.State}
-   *
-   * @param status the status
-   * @return the state for the status or null if there is no defined state for the given status
-   */
-  @Nullable
-  private static ProgramController.State getControllerStateByStatus(ProgramRunStatus status) {
-    for (ProgramController.State state : ProgramController.State.values()) {
-      if (state.getRunStatus() == status) {
-        return state;
-      }
-    }
-    return null;
   }
 
   private static final class AppMDS implements Iterable<MetadataStoreDataset> {
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
@@ -1,0 +1,320 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data.tools;
+
+import co.cask.cdap.api.data.stream.StreamSpecification;
+import co.cask.cdap.api.dataset.DatasetDefinition;
+import co.cask.cdap.api.dataset.DatasetProperties;
+import co.cask.cdap.api.dataset.table.Table;
+import co.cask.cdap.app.ApplicationSpecification;
+import co.cask.cdap.app.runtime.ProgramController;
+import co.cask.cdap.app.store.Store;
+import co.cask.cdap.common.conf.CConfiguration;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.data2.datafabric.dataset.DatasetsUtil;
+import co.cask.cdap.data2.dataset2.DatasetFramework;
+import co.cask.cdap.data2.dataset2.lib.table.MDSKey;
+import co.cask.cdap.data2.dataset2.lib.table.MetadataStoreDataset;
+import co.cask.cdap.data2.dataset2.tx.Transactional;
+import co.cask.cdap.internal.app.store.AppMetadataStore;
+import co.cask.cdap.internal.app.store.ApplicationMeta;
+import co.cask.cdap.internal.app.store.DefaultStore;
+import co.cask.cdap.internal.app.store.ProgramArgs;
+import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.ProgramRunStatus;
+import co.cask.cdap.proto.ProgramType;
+import co.cask.cdap.proto.RunRecord;
+import co.cask.tephra.TransactionExecutor;
+import co.cask.tephra.TransactionExecutorFactory;
+import com.google.common.base.Joiner;
+import com.google.common.base.Supplier;
+import com.google.common.base.Throwables;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.Sets;
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
+import org.apache.twill.filesystem.Location;
+import org.apache.twill.filesystem.LocationFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/**
+ * Upgraded the Meta Data for applications
+ */
+public class MDSUpgrader extends AbstractUpgrader {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MDSUpgrader.class);
+  private final Transactional<AppMDS, MetadataStoreDataset> appMDS;
+  private final CConfiguration cConf;
+  private final Store store;
+  private final Set<String> appStreams;
+
+  @Inject
+  private MDSUpgrader(LocationFactory locationFactory, TransactionExecutorFactory executorFactory,
+                      @Named("dsFramework") final DatasetFramework dsFramework, CConfiguration cConf,
+                      @Named("defaultStore") final Store store) {
+    super(locationFactory);
+    this.cConf = cConf;
+    this.store = store;
+    this.appMDS = Transactional.of(executorFactory, new Supplier<AppMDS>() {
+      @Override
+      public AppMDS get() {
+        try {
+          Table table = DatasetsUtil.getOrCreateDataset(dsFramework, Id.DatasetInstance.from
+                                                          (Constants.DEFAULT_NAMESPACE_ID, Joiner.on(".").join(
+                                                            Constants.SYSTEM_NAMESPACE, DefaultStore.APP_META_TABLE)),
+                                                        "table", DatasetProperties.EMPTY,
+                                                        DatasetDefinition.NO_ARGUMENTS, null);
+          return new AppMDS(new MetadataStoreDataset(table));
+        } catch (Exception e) {
+          LOG.error("Failed to access {} table", Joiner.on(".").join(Constants.SYSTEM_NAMESPACE,
+                                                                     DefaultStore.APP_META_TABLE), e);
+          throw Throwables.propagate(e);
+        }
+      }
+    });
+    appStreams = Sets.newHashSet();
+  }
+
+  @Override
+  public void upgrade() throws Exception {
+    appMDS.executeUnchecked(new TransactionExecutor.
+      Function<AppMDS, Void>() {
+      @Override
+      public Void apply(AppMDS appMetaStore) throws Exception {
+        MDSKey appMetaRecordPrefix = new MDSKey.Builder().add(AppMetadataStore.TYPE_APP_META).build();
+        List<ApplicationMeta> appsMeta = appMetaStore.mds.list(appMetaRecordPrefix, ApplicationMeta.class);
+        for (ApplicationMeta curAppMeta : appsMeta) {
+          handleAppSpec(curAppMeta.getId(), curAppMeta.getSpec());
+          upgradeAppMeta(curAppMeta);
+        }
+        return null;
+      }
+    });
+
+    // Scans for all the streams in the meta table
+    appMDS.executeUnchecked(new TransactionExecutor.
+      Function<AppMDS, Void>() {
+      @Override
+      public Void apply(AppMDS appMetaStore) throws Exception {
+        MDSKey streamRecordPrefix = new MDSKey.Builder().add(AppMetadataStore.TYPE_STREAM).build();
+        List<StreamSpecification> streamsSpec = appMetaStore.mds.list(streamRecordPrefix, StreamSpecification.class);
+        for (StreamSpecification curStreamSpec : streamsSpec) {
+          // Streams which are a part of an application gets upgraded when we upgrade the application meta
+          // if this stream was not a part of any application then this stream meta was not upgraded so upgrade it
+          // separately
+          if (!appStreams.contains(curStreamSpec.getName())) {
+            store.addStream(Id.Namespace.from(Constants.DEFAULT_NAMESPACE), curStreamSpec);
+          }
+        }
+        return null;
+      }
+    });
+  }
+
+  /**
+   * For an application calls {@link MDSUpgrader#handlePrograms} for the different program types whose meta data
+   * might exist in the meta table and will need upgrade
+   *
+   * @param appId the application id
+   * @param appSpec the {@link ApplicationSpecification} of the application
+   */
+  private void handleAppSpec(String appId, ApplicationSpecification appSpec) {
+    handlePrograms(appId, appSpec.getFlows().keySet(), ProgramType.FLOW);
+    handlePrograms(appId, appSpec.getMapReduce().keySet(), ProgramType.MAPREDUCE);
+    handlePrograms(appId, appSpec.getSpark().keySet(), ProgramType.SPARK);
+    handlePrograms(appId, appSpec.getWorkflows().keySet(), ProgramType.WORKFLOW);
+    handlePrograms(appId, appSpec.getServices().keySet(), ProgramType.SERVICE);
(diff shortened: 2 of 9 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+  @Inject
+  private MDSUpgrader(LocationFactory locationFactory, TransactionExecutorFactory executorFactory,
+                      @Named("dsFramework") final DatasetFramework dsFramework, CConfiguration cConf,
+                      @Named("defaultStore") final Store store) {
+    super(locationFactory);
+    this.cConf = cConf;
+    this.store = store;
+    this.appMDS = Transactional.of(executorFactory, new Supplier<DatasetContext<Table>>() {
+      @Override
+      public DatasetContext<Table> get() {
```
anwar6953: The following part is hard to read.
Maybe construct the DatasetInstance in a separate statement above the call to `getOrCreateDataset`, to make the entire thing easier to read.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+      Function<DatasetContext<Table>, Void>() {
+      @Override
+      public Void apply(DatasetContext<Table> configTable) throws Exception {
+        byte[] appMetaRecordPrefix = new MDSKey.Builder().add(AppMetadataStore.TYPE_STREAM).build().getKey();
+        Scanner rows = configTable.get().scan(appMetaRecordPrefix, Bytes.stopKeyForPrefix(appMetaRecordPrefix));
+        Row row;
+        while ((row = rows.next()) != null) {
+          StreamSpecification streamSpec = GSON.fromJson(Bytes.toString(row.get(COLUMN)), StreamSpecification.class);
+          // if this stream was not a part of any application
+          if (!appStreams.contains(streamSpec.getName())) {
```
anwar6953: I don't think you should be concerned with whether a stream was part of an app or not. just upgrade all of their metas, right?

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+          MDSKey.Splitter keyParts = new MDSKey(row.getRow()).split();
+          // skip runRecordStarted
+          keyParts.getString();
+          // skip default
+          keyParts.skipString();
+          // skip appId
+          keyParts.skipString();
+          // skip programId
+          keyParts.skipString();
+          String pId = keyParts.getString();
```
anwar6953: what is `pId`?

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+   * Writes the updated application metadata through the {@link DefaultStore}
+   *
+   * @param row the {@link Row} containing the application metadata
+   * @throws URISyntaxException if failed to create {@link URI} from the archive location in metadata
+   */
+  private void appMetaUpgrader(Row row) throws URISyntaxException, IOException {
+    ApplicationMeta appMeta = GSON.fromJson(Bytes.toString(row.get(COLUMN)), ApplicationMeta.class);
+    store.addApplication(Id.Application.from(Constants.DEFAULT_NAMESPACE, appMeta.getId()), appMeta.getSpec(),
+                         updateAppArchiveLocation(appMeta.getId(), new URI(appMeta.getArchiveLocation())));
+    // store the stream names which belongs to this app as we want to filter out streams which do not belong to any app
```
anwar6953: I don't think this filter is necessary.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+  }
+
+  /**
+   * Creates the new archive location for application with namespace
+   *
+   * @param appId           : the application id
+   * @param archiveLocation : the application archive location
+   * @return {@link Location} for the new archive location with namespace
+   */
+  private Location updateAppArchiveLocation(String appId, URI archiveLocation) throws IOException {
```
anwar6953: Is this necessary - since we're requiring app redeployment?

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+   * Note: This will given the state as {@link ProgramController.State#STARTING} for {@link ProgramRunStatus#RUNNING}
+   * even though running has multiple mapping but that is fine in our case as we use this
+   * {@link ProgramController.State} to write the runRecordCompleted through {@link DefaultStore#setStop} which converts
+   * it back to {@link ProgramRunStatus#RUNNING}. So we don't really care about the temporary intermediate
+   * {@link ProgramController.State}
+   *
+   * @param status : the status
+   * @return the state for the status or null if there is no defined state for the given status
+   */
+  private static ProgramController.State getControllerStateByStatus(ProgramRunStatus status) {
```
anwar6953: Mark the method with `@nullable` ?

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: We are already getting all the appSpec and all we have to do is store the stream name which is just a string. 
I think its cleaner since we anyways going through the appspec and it will save us a lot of overwrite since generally we will have streams which are part of apps not the other way.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: the process id

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: re-writing unnecessarily when we can avoid it.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: In in half states about this. We can remove this or keep it
@bdmogal : Views ?

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
anwar6953: sorry, but what is process id, in regards to run records?

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
anwar6953: But I think the upgrade tool logic can be simpler if it simply upgrades upgrades all stream meta (without maintaining logic to upgrade this set of streams, but not *that* set of streams). Simplicity means fewer chance for bugs.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: You can look up how the key is made up.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: Thats debatable I guess and this doesn't look complex too me at all.
Lets wait for few other pair of eyes and we will do which look more right to all.

anwar6953: Ok, my main concern was that in the future, streams in an app in one namespace would stop all streams in other namespaces from being upgraded, since you're only storing String to identify the stream.
Should use Id.Stream, to also work in the future.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
anwar6953: Ok, from the usages, it looks like its actually `runId`. Not process id.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
albertshau: should move most of this class into a new upgrade method for Store.  Then you don't have to do ugly stuff like this as appMDS is already in that class, as well as transaction stuff.  Also keep internal implementation internal that way.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+    appStreams = Sets.newHashSet();
+  }
+
+  @Override
+  public void upgrade() throws Exception {
+    appMDS.executeUnchecked(new TransactionExecutor.
+      Function<DatasetContext<Table>, Void>() {
+      @Override
+      public Void apply(DatasetContext<Table> configTable) throws Exception {
+        byte[] appMetaRecordPrefix = new MDSKey.Builder().add(AppMetadataStore.TYPE_APP_META).build().getKey();
```
albertshau: for example, here you could just re-use the ```list``` method MetaStoreDataset instead of scanning yourself

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
albertshau: this should be working with an AppMetadataStore instead of the table that it uses, similar to what Store does

albertshau: could make sense to move most of this into a new upgrade method in Store

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+        Scanner rows = configTable.get().scan(appMetaRecordPrefix, Bytes.stopKeyForPrefix(appMetaRecordPrefix));
+        Row row;
+        while ((row = rows.next()) != null) {
+          ApplicationMeta appMeta = GSON.fromJson(Bytes.toString(row.get(COLUMN)), ApplicationMeta.class);
```
albertshau: and you wouldn't have to set up GSON yourself

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+        return null;
+      }
+    });
+
+    // Scans for all the streams in the meta table
+    appMDS.executeUnchecked(new TransactionExecutor.
+      Function<DatasetContext<Table>, Void>() {
+      @Override
+      public Void apply(DatasetContext<Table> configTable) throws Exception {
+        byte[] appMetaRecordPrefix = new MDSKey.Builder().add(AppMetadataStore.TYPE_STREAM).build().getKey();
```
albertshau: similarly, you can re-use method to list all streams

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+      }
+    });
+  }
+
+  /**
+   * Upgrades the stream record in the meta table
+   *
+   * @param streamSpec the {@link StreamSpecification} of the stream
+   */
+  private void streamHandler(StreamSpecification streamSpec) {
```
albertshau: methods should be verbs, not nouns.  Also, is there a reason we need a separate method for this if it's just one line?

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
albertshau: i don't understand this either, why are we not updating all stream specs?

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgraderMain.java
```diff
@@ -162,16 +162,8 @@ public DatasetFramework getNamespacedDSFramework(CConfiguration cConf,
...
-          throws DatasetManagementException {
-          return createNonNamespaceDSFramework(registryFactory);
-        }
-
-        @Provides
-        @Singleton
-        @Named("nonNamespacedStore")
-        public Store getNonNamespacedStore(@Named("nonNamespacedDSFramework") DatasetFramework nonNamespacedFramework,
+        @Named("defaultStore")
+        public Store getNonNamespacedStore(@Named("dsFramework") DatasetFramework nonNamespacedFramework,
```
albertshau: also change the name of ```nonNamespacedFramework``` to avoid confusion

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
anwar6953: I think the idea is that if the user deploys an app that uses a stream, its streams' meta will be upgraded.

But that's not the right approach because:
What if someone creates a stream outside of an app first, and then upgrades and decides not to deploy an app that happens to use it. They should still be able to use the stream, since its usage is not restricted to an app.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+  }
+
+  /**
+   * For an application calls {@link MDSUpgrader#programHandler} for the different program types whose meta data
+   * might exist in the meta table and will need upgrade
+   *
+   * @param appId   the application id
+   * @param appSpec the {@link ApplicationSpecification} of the application
+   */
+  private void appSpecHandler(String appId, ApplicationSpecification appSpec) {
```
albertshau: same here, should be a verb.  handleAppSpec is better than appSpecHandler

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,347 @@
...
+   * @param programId   the program id of the program
+   * @param programType the {@link ProgramType} of the program
+   */
+  private void programArgsHandler(final String appId, final String programId, final ProgramType programType) {
+    final byte[] partialKey = new MDSKey.Builder().add(AppMetadataStore.TYPE_RUN_RECORD_STARTED, DEVELOPER_ACCOUNT,
+                                                       appId, programId).build().getKey();
+    appMDS.executeUnchecked(new TransactionExecutor.Function<DatasetContext<Table>, Void>() {
+      @Override
+      public Void apply(DatasetContext<Table> input) throws Exception {
+        Scanner rows = input.get().scan(partialKey, Bytes.stopKeyForPrefix(partialKey));
```
albertshau: similarly, could re-use method to list runs if this logic was moved

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
albertshau: don't think there is a need to do this either as it makes things more complicated than needed. just have one method for updating streams that gets all the specs and updates them

albertshau:
overall comment is that we are duplicating a lot of logic that already exists inside the app meta store, and that we should move upgrade logic to the guys that know how the data is stored.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: No we don't need to upgrade the streams which belongs too apps because when we add an app the code in AppMetadataStore goes through all the streams which belongs to it and adds it. So for streams which belong to an app we don't have to move them as it gets moved when we move the app spec

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: Since the streams which belongs to an app gets upgraded when the app meta is upgraded do we really need to do it all over again for those streams ? It just looks like unnecessary writes to me.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: Lets keep this pId here to be consistent with how the key was constructed else it will get too confusing.
If we want to rename it we will rename it both places in a different PR.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
rohitsinha54: Just the old table getting part or the whole MDS upgrade logic ? Last time when we discussed this we planned that only for log and for MDS since it just makes some constants public we decided its fine to keep it out.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
albertshau: i see.. can you add a comment explaining then

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
albertshau: the main problem is that this code works at a lower abstraction (Table) than it should, which means it ends up re-implementing a lot of logic that already exists and which should be contained only in the meta store.  So now we have the same logic in 2 places which is bad.

rohitsinha54: I did your suggested modification to use MetaDataStore. Will be updating it now. Have a look

rohitsinha54:
@albertshau : Addressed comments and changed to use MetaDataStore directly. 
Please review. Thanks.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -59,7 +59,7 @@
   @Inject
   private DatasetUpgrader(CConfiguration cConf, Configuration hConf, LocationFactory locationFactory,
                           QueueAdmin queueAdmin, HBaseTableUtil hBaseTableUtil,
-                          @Named("namespacedDSFramework") DatasetFramework namespacedFramework) {
+                          @Named("dsFramework") DatasetFramework namespacedFramework) {
```
albertshau: change the name of the variable too

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  /**
+   * Handles the {@link AppMetadataStore#TYPE_RUN_RECORD_STARTED} meta data and writes it back with namespace
+   *
+   * @param appId the application id to which this program belongs to
+   * @param programId the program id of the program
+   * @param programType the {@link ProgramType} of the program
+   */
+  private void runRecordStartedHandler(final String appId, final String programId, final ProgramType programType) {
```
albertshau: nit: rename to handleRunRecordStarted

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  /**
+   * Handles the {@link AppMetadataStore#TYPE_RUN_RECORD_COMPLETED} meta data and writes it back with namespace
+   *
+   * @param appId the application id to which this program belongs to
+   * @param programId the program id of the program
+   * @param programType the {@link ProgramType} of the program
+   */
+  private void runRecordCompletedHandler(final String appId, final String programId, final ProgramType programType) {
```
albertshau: nit: rename to handleRunRecordCompleted

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
rohitsinha54: Damn! I keep missing something. Apologies :(

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,320 @@
...
+
+  /**
+   * Gives the {@link ProgramController.State} for a given {@link ProgramRunStatus}
+   * Note: This will given the state as {@link ProgramController.State#STARTING} for {@link ProgramRunStatus#RUNNING}
+   * even though running has multiple mapping but that is fine in our case as we use this
+   * {@link ProgramController.State} to write the runRecordCompleted through {@link DefaultStore#setStop} which converts
+   * it back to {@link ProgramRunStatus#RUNNING}. So we don't really care about the temporary intermediate
+   * {@link ProgramController.State}
+   *
+   * @param status : the status
```
albertshau: no colon

rohitsinha54:
@albertshau : Addressed comments. Please have a final look.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -0,0 +1,320 @@
...
+      cConf.get(Constants.AppFabric.OUTPUT_DIR)).append(appId).append(Constants.ARCHIVE_DIR).append(archiveFilename);
+  }
+
+  /**
+   * Gives the {@link ProgramController.State} for a given {@link ProgramRunStatus}
+   * Note: This will given the state as {@link ProgramController.State#STARTING} for {@link ProgramRunStatus#RUNNING}
+   * even though running has multiple mapping but that is fine in our case as we use this
+   * {@link ProgramController.State} to write the runRecordCompleted through {@link DefaultStore#setStop} which converts
+   * it back to {@link ProgramRunStatus#RUNNING}. So we don't really care about the temporary intermediate
+   * {@link ProgramController.State}
```
albertshau: i see... this is bad though because we only know this by looking at the implementation of AppMetadataStore.  I think instead of this we can refactor the api to take a ProgramRunStatus, since that is what is being used anyway.

albertshau:
its still not great that the row key parts have been made public, but not as bad as replicating the logic.  Maybe it makes sense to move this into Store, but in the interest of time, let's see if we can improve that in another PR.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
bdmogal: I'm not completely sure either. @albertshau any thoughts?

rohitsinha54:
@albertshau : Addressed comments. Please have a final look. 
Build running here: http://builds.cask.co/browse/CDAP-DUT1023-5

albertshau:
lgtm assuming build passes

albertshau:
:+1:

PR #1527, opened 2015-03-07
[CDAP-1369] Namespaced log dir

bdmogal:
Summary:
- [x] Moved all CDAP logs under ``/<root>/[namespace]/${log.base.dir}`` directory
- [x] System services logs use **system** namespace rather than **cdap** namespace.

Impacts a large number of files only because of the change from "cdap" to "system" for system services.

Most of the functional changes are in ``AvroFileWriter`` and ``LogCleanup``.

Issue: https://issues.cask.co/browse/CDAP-1369
Build: http://builds.cask.co/browse/CDAP-DUT1025-1/

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
@@ -483,7 +483,6 @@ public final class Constants {
    * Logging constants.
    */
   public static final class Logging {
-    public static final String SYSTEM_NAME = "cdap";
     public static final String COMPONENT_NAME = "services";
   }
 
@@ -726,6 +725,11 @@ public final class Constants {
    */
   public static final String SYSTEM_NAMESPACE = "system";
   public static final Id.Namespace SYSTEM_NAMESPACE_ID = Id.Namespace.from(SYSTEM_NAMESPACE);
+
+  /**
+   * 'cdap' reserved namespace name. Unused right now (other than in logging. Reserved in case we need it in future.
+   */
+  public static final String CDAP_NAMESPACE = "cdap";
 
   /**
    * Constants related to external systems.
--- a/cdap-common/src/main/java/co/cask/cdap/common/logging/NamespaceLoggingContext.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/logging/NamespaceLoggingContext.java
@@ -23,11 +23,18 @@ public abstract class NamespaceLoggingContext extends AbstractLoggingContext {
   public static final String TAG_NAMESPACE_ID = ".namespaceId";
 
   /**
-   * Constructs AccountLoggingContext.
+   * Constructs NamespaceLoggingContext.
    * @param namespaceId namespace id
    */
   public NamespaceLoggingContext(final String namespaceId) {
     setSystemTag(TAG_NAMESPACE_ID, namespaceId);
+  }
+
+  @Override
+  public String getNamespacedLogBaseDir(String logBaseDir) {
+    // cannot just call #getLogPathFragment here, because if called on a subclass' object, that would return
+    // a wrong value
+    return String.format("%s/%s", getSystemTag(TAG_NAMESPACE_ID), logBaseDir);
   }
 
   @Override
@@ -36,7 +43,7 @@ public abstract class NamespaceLoggingContext extends AbstractLoggingContext {
   }
 
   @Override
-  public String getLogPathFragment() {
-    return String.format("%s", getSystemTag(TAG_NAMESPACE_ID));
+  public String getLogPathFragment(String logBaseDir) {
+    return String.format("%s/%s", getSystemTag(TAG_NAMESPACE_ID), logBaseDir);
   }
 }
--- a/cdap-watchdog/src/main/java/co/cask/cdap/logging/write/AvroFileWriter.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/logging/write/AvroFileWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2014 Cask Data, Inc.
+ * Copyright © 2014-2015 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -45,7 +45,8 @@ public final class AvroFileWriter implements Closeable, Flushable {
   private static final Logger LOG = LoggerFactory.getLogger(AvroFileWriter.class);
 
   private final FileMetaDataManager fileMetaDataManager;
-  private final Location baseDir;
+  private final Location rootDir;
+  private final String logBaseDir;
   private final Schema schema;
   private final int syncIntervalBytes;
   private final Map<String, AvroFile> fileMap;
@@ -57,17 +58,18 @@ public final class AvroFileWriter implements Closeable, Flushable {
   /**
    * Constructs an AvroFileWriter object.
    * @param fileMetaDataManager used to store file meta data.
-   * @param baseDir base dir.
+   * @param rootDir the CDAP root dir on the filesystem
+   * @param logBaseDir the basedirectory for logs as defined in configuration
    * @param schema schema of the Avro data to be written.
    * @param maxFileSize Avro files greater than maxFileSize will get rotated.
    * @param syncIntervalBytes the approximate number of uncompressed bytes to write in each block.
    * @param inactiveIntervalMs files that have no data written for more than inactiveIntervalMs will be closed.
    */
-  public AvroFileWriter(FileMetaDataManager fileMetaDataManager,
-                        Location baseDir, Schema schema,
+  public AvroFileWriter(FileMetaDataManager fileMetaDataManager, Location rootDir, String logBaseDir, Schema schema,
                         long maxFileSize, int syncIntervalBytes, long inactiveIntervalMs) {
     this.fileMetaDataManager = fileMetaDataManager;
-    this.baseDir = baseDir;
+    this.rootDir = rootDir;
+    this.logBaseDir = logBaseDir;
     this.schema = schema;
     this.syncIntervalBytes = syncIntervalBytes;
     this.fileMap = Maps.newHashMap();
@@ -91,7 +93,8 @@ public final class AvroFileWriter implements Closeable, Flushable {
     LoggingContext loggingContext = event.getLoggingContext();
 
     if (LOG.isTraceEnabled()) {
-      LOG.trace("Appending {} messages for logging context {}", events.size(), loggingContext.getLogPathFragment());
+      LOG.trace("Appending {} messages for logging context {}", events.size(),
+                loggingContext.getLogPathFragment(logBaseDir));
     }
 
     long timestamp = event.getLogEvent().getTimeStamp();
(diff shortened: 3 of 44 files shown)


cdap-watchdog/src/main/java/co/cask/cdap/logging/write/AvroFileWriter.java
```diff
@@ -174,9 +182,25 @@ private AvroFile createAvroFile(LoggingContext loggingContext, long timestamp) t
...
-  private Location createLocation(String pathFragment, long timestamp) throws IOException {
+  private Location createLocation(LoggingContext loggingContext, long timestamp) throws IOException {
+    Map<String, LoggingContext.SystemTag> systemTags = loggingContext.getSystemTagsMap();
+    LoggingContext.SystemTag topLevelTag;
+    if (systemTags.containsKey(NamespaceLoggingContext.TAG_NAMESPACE_ID)) {
+      topLevelTag = systemTags.get(NamespaceLoggingContext.TAG_NAMESPACE_ID);
+    } else if (systemTags.containsKey(SystemLoggingContext.TAG_SYSTEM_ID)) {
+      topLevelTag = systemTags.get(SystemLoggingContext.TAG_SYSTEM_ID);
+    } else {
+      throw new IllegalArgumentException(
```
poornachandra: Not a good idea to throw exception from here. A stray log event will block the whole pipeline. Instead we should validate the log event when it is received, and drop it if it does not have right tags.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/AvroFileWriter.java
```diff
@@ -174,9 +182,25 @@ private AvroFile createAvroFile(LoggingContext loggingContext, long timestamp) t
...
+        String.format("Expected top level tag in logging context to be either %s or %s. Path fragment = %s.",
+                      NamespaceLoggingContext.TAG_NAMESPACE_ID, SystemLoggingContext.TAG_SYSTEM_ID,
+                      loggingContext.getLogPathFragment()));
+    }
+    String namespace = topLevelTag.getValue();
+    String logBaseDir = cConf.get(LoggingConfiguration.LOG_BASE_DIR);
+    String pathFragment = loggingContext.getLogPathFragment().replaceAll(String.format("%s/", namespace), "");
```
poornachandra: Should we instead change `getLogPathFragment()` to take in a base dir and then give the right path? Or, do away with `getLogPathFragment()`, as it does not seem to be adding any value.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/AvroFileWriter.java
```diff
@@ -174,9 +182,25 @@ private AvroFile createAvroFile(LoggingContext loggingContext, long timestamp) t
     return avroFile;
   }
 
-  private Location createLocation(String pathFragment, long timestamp) throws IOException {
+  private Location createLocation(LoggingContext loggingContext, long timestamp) throws IOException {
+    Map<String, LoggingContext.SystemTag> systemTags = loggingContext.getSystemTagsMap();
+    LoggingContext.SystemTag topLevelTag;
+    if (systemTags.containsKey(NamespaceLoggingContext.TAG_NAMESPACE_ID)) {
+      topLevelTag = systemTags.get(NamespaceLoggingContext.TAG_NAMESPACE_ID);
```
poornachandra: This logic is very specific to `LoggingContext`, I don't think it is a good idea to have it here. Also, see my comment about `getLogPathFragment()`

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/LogCleanup.java
```diff
@@ -86,33 +90,49 @@ Location getParent(Location location) {
...
-   * @param dir dir to be deleted.
+   * For the specified directory to be deleted, finds its namespaced log location, then deletes
+   * @param dir
+   * @throws IOException
    */
-  void deleteEmptyDir(Location dir) {
+  void deleteEmptyDir(Location dir) throws IOException {
     LOG.debug("Got path {}", dir.toURI());
 
+    List<Location> namespaceDirs = rootDir.list();
```
poornachandra: Instead of trying to construct base dir by listing all namespaces, it is possible to figure out the base dir from the dir to be deleted since it is of format `/{root}/ns/logBaseDir`, right? This will avoid listing root dir for every file deletion.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/LogCleanup.java
```diff
@@ -86,33 +90,49 @@ Location getParent(Location location) {
...
+    for (Location namespaceDir : namespaceDirs) {
+      Location namespacedLogBaseDir = namespaceDir.append(logBaseDir);
+      if (dir.toURI().getRawPath().startsWith(namespacedLogBaseDir.toURI().getRawPath())) {
+        deleteEmptyDirsInNamespace(namespacedLogBaseDir, dir);
+      }
```
poornachandra: In the else block, we should log a message saying dir of unknown format passed in for deletion, since we are not taking any action in that case.

poornachandra:
@bdmogal Done with review, please address comments.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/LogCleanup.java
bdmogal: Agree its a little dirty, but the problem is that ``LogCleanup`` does not know about the namespace. Let me think. Maybe we could pass it in.

poornachandra: Does it need to know the namespace, isn't just the format enough?

bdmogal: Will have to do some sort of regex matching then, right? Or something with ``getParent(location)``. Explored both, seemed a little too hacky for comfort to me for a delete operation.

bdmogal: Any other ideas?

bdmogal:
@poornachandra can you please take a look at it again?

cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
```diff
@@ -725,6 +724,11 @@
   public static final Id.Namespace SYSTEM_NAMESPACE_ID = Id.Namespace.from(SYSTEM_NAMESPACE);
 
   /**
+   * 'cdap' reserved namespace name
```
albertshau: this is only here for upgrade purposes right? maybe add a comment saying nobody should use this.

cdap-common/src/main/java/co/cask/cdap/common/logging/NamespaceLoggingContext.java
```diff
@@ -31,12 +31,19 @@ public NamespaceLoggingContext(final String namespaceId) {
   }
 
   @Override
+  public String getNamespacedPathFragment(String logBaseDir) {
```
albertshau: should rename this so it is clear it's the namespaced base dir. sounds like it's a namespaced version of getLogPathFragment()

cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
bdmogal: No, we're actually reserving the namespace name 'cdap'. Just in case we need it in future too.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/FileMetaDataManager.java
```diff
@@ -194,6 +199,10 @@ public Integer apply(DatasetContext<Table> ctx) throws Exception {
     });
   }
 
+  private String getLogPartition(byte[] rowKey) {
+    return Bytes.toString(Arrays.copyOfRange(rowKey, ROW_KEY_PREFIX_END.length, rowKey.length));
```
albertshau: Bytes has a version with offset and length. or is there some other reason you're copying?

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/LogCleanup.java
```diff
@@ -67,14 +71,18 @@ public void handle(Location location) {
                                                               location.toURI()), e);
                                               throw Throwables.propagate(e);
                                             }
+
+                                            // Delete any empty parent dirs
+                                            for (Location dir : parentDirs) {
```
albertshau: how come this was moved into the callback? This runs in a transaction right?  If it should stay here we can move parentDirs into here instead of making it final outside.

albertshau:

cdap-common/src/main/java/co/cask/cdap/common/logging/NamespaceLoggingContext.java
bdmogal: Fixed - ``getNamespacedLogBaseDir``

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/FileMetaDataManager.java
bdmogal: Used ``Bytes`` version with offset and length

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/LogCleanup.java
bdmogal: Moving it out. Please review again.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/LogCleanup.java
```diff
@@ -50,17 +51,20 @@ public void run() {
     LOG.info("Running log cleanup...");
     try {
       long tillTime = System.currentTimeMillis() - retentionDurationMs;
-      final Set<Location> parentDirs = Sets.newHashSet();
+      final Map<String, Set<Location>> parentDirs = Maps.newHashMap();
```
poornachandra: Guava Multimap can be used here to simplify usage.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/AvroFileWriter.java
```diff
@@ -56,18 +59,19 @@
...
    * @param maxFileSize Avro files greater than maxFileSize will get rotated.
    * @param syncIntervalBytes the approximate number of uncompressed bytes to write in each block.
    * @param inactiveIntervalMs files that have no data written for more than inactiveIntervalMs will be closed.
    */
-  public AvroFileWriter(FileMetaDataManager fileMetaDataManager,
-                        Location baseDir, Schema schema,
-                        long maxFileSize, int syncIntervalBytes, long inactiveIntervalMs) {
+  public AvroFileWriter(CConfiguration cConf, FileMetaDataManager fileMetaDataManager, Location rootDir,
+                        Schema schema, long maxFileSize, int syncIntervalBytes, long inactiveIntervalMs) {
+    this.cConf = cConf;
```
poornachandra: The only reason of having `cConf` here is to get log base dir. Should we just pass log base dir directly here?

poornachandra:
@bdmogal Changes look good. Couple of minor comments, please address them.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/AvroFileWriter.java
bdmogal: Absolutely. Done.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/LogCleanup.java
bdmogal: Yup, fixed.

bdmogal:
@poornachandra addressed comments. Please review again.

bdmogal:
Build is green - http://builds.cask.co/browse/CDAP-DUT1025-8

poornachandra:
LGTM :+1:

bdmogal:
Squashed commits. Merging.

PR #1528, opened 2015-03-07
Use TableId for IncrementHandlerTest / IncrementSummingScannerTest

anwar6953:
Need for this change arises due to the fact that the IncrementHandler coprocessor needs to read from the ConfigurationTable.
In order to do so, it needs to construct the configuration table name based off of the table the coprocessor instance is operating on, and so it needs the table prefix (which is 'cdap' by default).

In 2.7, this test case was fine, because of the following logic:
[release/2.7.0](https://github.com/caskdata/cdap/blob/release/2.7.0/cdap-data-fabric/src/main/java/co/cask/cdap/data2/increment/hbase/IncrementHandlerState.java#L78-L82)
So if it couldn't parse the root prefix, it improperly constructed the configuration table name, and so failed to talk to the ConfigurationTable.

Instead, the test cases should use a table name from which a root prefix can be computed (which this PR changes).

**Note**: The alternative to the changes in this PR is to simply use a table name such as "cdap.X", instead of just "X".

Build: http://builds.cask.co/browse/CDAP-DUT1026-4

### DIFF
--- a/cdap-data-fabric/src/test/java/co/cask/cdap/data2/increment/hbase/AbstractIncrementHandlerTest.java
+++ b/cdap-data-fabric/src/test/java/co/cask/cdap/data2/increment/hbase/AbstractIncrementHandlerTest.java
@@ -16,9 +16,13 @@
 
 package co.cask.cdap.data2.increment.hbase;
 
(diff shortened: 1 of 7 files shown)


ghelmling:
lgtm

PR #1529, opened 2015-03-08
Capabilities to submit programs in non-default yarn Queues

sreevatsanraman:
* Supports two configurations at cdap level
  * `cdap.services.scheduler.queue` which specifies the queue name for cdap master twill app
  * `app.scheduler.queue` which is used to specify queue name for programs (flows, mapreduce, etc)
* The setting for program can be specified at a namespace level which will take precedence over cdap level setting

Build: http://builds.cask.co/browse/CDAP-DUT885-2


sreevatsanraman:
Opened a JIRA to write tests in distributed setup - https://issues.cask.co/browse/CDAP-1730

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -99,8 +99,14 @@ public final ProgramController run(final Program program, final ProgramOptions o
     final File cConfFile;
     final Program copiedProgram;
     final File programDir;    // Temp directory for unpacking the program
+    final String schedulerQueueName = options.getArguments().getOption(Constants.AppFabric.APP_SCHEDULER_QUEUE);
 
     try {
+
+      if (schedulerQueueName != null) {
+        hConf.set(Constants.MapReduce.MAP_REDUCE_JOB_QUEUE_NAME, schedulerQueueName);
```
poornachandra: I would be better if this is set in `DistributedMapReduceProgramRunner`, right?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1980,4 +1982,45 @@ private void deleteProcessMetricsForFlow(String application, String flow) throws
...
+    if (schedulerQueue != null) {
+      configs.put(Constants.AppFabric.APP_SCHEDULER_QUEUE, schedulerQueue);
+    }
+
+    return configs;
+  }
+
+  private Map<String, String> getNamespaceResolvedSystemArguments(String namespaceId, Map<String, String> configs) {
+    Map<String, String> resolvedConfigs = Maps.newHashMap();
+    for (Map.Entry<String, String> entry : configs.entrySet()) {
```
poornachandra: The for loop can be replaced by `Map<String, String> resolvedConfigs = Maps.newHashMap(configs)`

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1980,4 +1982,45 @@ private void deleteProcessMetricsForFlow(String application, String flow) throws
...
+      resolvedConfigs.put(entry.getKey(), entry.getValue());
+    }
+
+    NamespaceMeta meta = store.getNamespace(Id.Namespace.from(namespaceId));
+    Preconditions.checkNotNull(meta);
```
poornachandra: Can you add a message to the pre-condition check?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1980,4 +1982,45 @@ private void deleteProcessMetricsForFlow(String application, String flow) throws
...
+
+    NamespaceConfig config = meta.getConfig();
+    // The only config currently as system arguments is Scheduler queue.
+    String schedulerQueue = config.getSchedulerQueueName();
+    if (schedulerQueue != null && !schedulerQueue.isEmpty()) {
```
poornachandra: It would be a good idea to perform this check in the beginning, and return `configs` if `schedulerQueue` is not set at namespace level. We won't have to copy maps in that case, right?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sreevatsanraman: Yes. Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sreevatsanraman: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sreevatsanraman: I think this is easy to extend to more config type incase you add more params. otherwise we will rewrite this part of code. I will leave it as is unless there is a strong reason not to do this.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
sreevatsanraman: I thought about this, I did not want to override the run method and replicate the logic there. This way all logic for run is in AbstractDistributedProgramRunner

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -158,6 +158,12 @@ protected BaseHiveExploreService(TransactionSystemClient txClient, DatasetFramew
     this.cConf = cConf;
     this.hConf = hConf;
     this.hiveConf = hiveConf;
+    String schedulerQueue = cConf.get(Constants.AppFabric.APP_SCHEDULER_QUEUE);
+    if (schedulerQueue != null && !schedulerQueue.isEmpty()) {
+      hConf.set(Constants.MapReduce.MAP_REDUCE_JOB_QUEUE_NAME, schedulerQueue);
```
poornachandra: Every Explore execute call comes with a default namespace. Instead of determining namespace here, if we do it in `execute` method then we can figure out which namespace to use for each call.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
poornachandra: What about setting it in `launch` method?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
poornachandra: Sure, let's leave it as-is for now. We can revisit this when we add more params.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -218,7 +227,7 @@ protected CLIService getCliService() {
   private IMetaStoreClient getMetaStoreClient() throws ExploreException {
     if (metastoreClientLocal.get() == null) {
       try {
-        IMetaStoreClient client = new HiveMetaStoreClient(getHiveConf());
+          IMetaStoreClient client = new HiveMetaStoreClient(getHiveConf());
```
poornachandra: Extra space?

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -193,7 +202,7 @@ protected BaseHiveExploreService(TransactionSystemClient txClient, DatasetFramew
   }
 
   protected HiveConf getHiveConf() {
-    HiveConf conf = new HiveConf();
+    HiveConf conf = new HiveConf(hiveConf);
```
poornachandra: This line is no longer required, right?

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -667,11 +676,18 @@ public QueryHandle execute(Id.Namespace namespace, String statement) throws Expl
...
       // It looks like the username and password below is not used when security is disabled in Hive Server2.
       SessionHandle sessionHandle = cliService.openSession("", "", sessionConf);
       try {
         String database = getHiveDatabase(namespace.getId());
+        String schedulerQueue = getNamespaceResolvedSchedulerQueue(namespace);
+        if (schedulerQueue != null) {
+          sessionConf.put(Constants.MapReduce.MAP_REDUCE_JOB_QUEUE_NAME, schedulerQueue);
+          LOG.info("Setting scheduler queue to {}", schedulerQueue);
+        } else {
+          LOG.info("scheduler queue not set");
```
poornachandra: Info log can be removed now, right?

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -1002,7 +1018,26 @@ private String getHiveDatabase(String namespace) {
...
+
+    Transaction tx = startTransaction();
+    ConfigurationUtil.set(sessionConf, Constants.Explore.TX_QUERY_KEY, TxnCodec.INSTANCE, tx);
+    ConfigurationUtil.set(sessionConf, Constants.Explore.CCONF_KEY, CConfCodec.INSTANCE, cConf);
+    ConfigurationUtil.set(sessionConf, Constants.Explore.HCONF_KEY, HConfCodec.INSTANCE, hConf);
+
+    return sessionConf;
+  }
+
+  protected Map<String, String> startSession(Id.Namespace namespace) throws IOException {
```
poornachandra: As we discussed, if `startSession()` calls `startSession(null)` we can eliminate duplicate code.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -1002,7 +1018,26 @@ private String getHiveDatabase(String namespace) {
...
+    Map<String, String> sessionConf = Maps.newHashMap();
+
+    QueryHandle queryHandle = QueryHandle.generate();
+    sessionConf.put(Constants.Explore.QUERY_ID, queryHandle.getHandle());
+
+    String schedulerQueue = getNamespaceResolvedSchedulerQueue(namespace);
+    if (schedulerQueue != null) {
+      sessionConf.put(Constants.MapReduce.MAP_REDUCE_JOB_QUEUE_NAME, schedulerQueue);
```
poornachandra: No need to add job queue name to session conf again, as `startSession` already does it, right?

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -667,11 +676,18 @@ public QueryHandle execute(Id.Namespace namespace, String statement) throws Expl
...
     try {
-      Map<String, String> sessionConf = startSession();
+      Map<String, String> sessionConf = startSession(namespace);
       // It looks like the username and password below is not used when security is disabled in Hive Server2.
       SessionHandle sessionHandle = cliService.openSession("", "", sessionConf);
       try {
         String database = getHiveDatabase(namespace.getId());
+        String schedulerQueue = getNamespaceResolvedSchedulerQueue(namespace);
+        if (schedulerQueue != null) {
+          sessionConf.put(Constants.MapReduce.MAP_REDUCE_JOB_QUEUE_NAME, schedulerQueue);
```
poornachandra: No need to add job queue name to session conf again, as startSession already does it, right?

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -1149,6 +1186,16 @@ protected Object tColumnToObject(TColumnValue tColumnValue) throws ExploreExcept
     throw new ExploreException("Unknown column value encountered: " + tColumnValue);
   }
 
+  private String getNamespaceResolvedSchedulerQueue(Id.Namespace namespaceId) {
```
poornachandra: This method is duplicated in 3 places. We should probably re-factor this logic out into a class.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -99,8 +99,14 @@ public final ProgramController run(final Program program, final ProgramOptions o
     final File cConfFile;
     final Program copiedProgram;
     final File programDir;    // Temp directory for unpacking the program
+    final String schedulerQueueName = options.getArguments().getOption(Constants.AppFabric.APP_SCHEDULER_QUEUE);
 
     try {
+      if (schedulerQueueName != null) {
+        hConf.set(Constants.MapReduce.MAP_REDUCE_JOB_QUEUE_NAME, schedulerQueueName);
```
poornachandra: Looks like my comment about setting this in `DistributedMapReduceProgramRunner.launch()` somehow got deleted. What do you think about it?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/DefaultSchedulerService.java
```diff
@@ -60,24 +69,42 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
...
-        ProgramOptionConstants.SCHEDULE_NAME, scheduleName
-      ));
+      ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+
+      builder.put(ProgramOptionConstants.LOGICAL_START_TIME, Long.toString(context.getScheduledFireTime().getTime()));
+      builder.put(ProgramOptionConstants.RETRY_COUNT, Integer.toString(context.getRefireCount()));
+      builder.put(ProgramOptionConstants.SCHEDULE_NAME, scheduleName);
+      String schedulerQueue = getNamespaceResolvedScheduleQueue(namespaceId);
+      if (schedulerQueue != null) {
+        builder.put(Constants.AppFabric.APP_SCHEDULER_QUEUE, schedulerQueue);
```
poornachandra: Would we have to add queue name conf here even if we set it in `DistributedMapReduceProgramRunner.launch()`?

poornachandra:
@sreevatsanraman Few more comments to address.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
sreevatsanraman: Fixed.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
sreevatsanraman: Fixed.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
sreevatsanraman: Fixed.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
sreevatsanraman: Fixed.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
sreevatsanraman: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
sreevatsanraman: I replied that if I had a `launch` method then I would need to replicate the code in launch. I thought this was simpler.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/DefaultSchedulerService.java
sreevatsanraman: Yes. I noticed that the triggered jobs we submitted to default queues. that's when I found this missing piece.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
sreevatsanraman: ok. will think about the right place and fix.

sreevatsanraman: Fixed.

sreevatsanraman:
@poornachandra addressed the comments.

sreevatsanraman:
> ScheduleQueueResolver => Schedule[r]QueueResolver?
Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/scheduler/SchedulerQueueResolver.java
```diff
@@ -0,0 +1,53 @@
...
+public class SchedulerQueueResolver {
+  private final String schedulerQueueFromCConf;
+  private final Store store;
+
+  public SchedulerQueueResolver(CConfiguration cConf, Store store) {
+    this.schedulerQueueFromCConf = cConf.get(Constants.AppFabric.APP_SCHEDULER_QUEUE);
+    this.store = store;
+  }
+
+  public String getSchedulerQueueFromCConf() {
```
chtyim: call it `getDefaultSchedulerQueue` because it doesn't have to be come from cConf.

chtyim: Or simply, getDefaultQueue.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/scheduler/SchedulerQueueResolver.java
```diff
@@ -0,0 +1,53 @@
...
+    return schedulerQueueFromCConf;
+  }
+
+  public String getNamespaceResolvedSchedulerQueue(Id.Namespace namespaceId) {
```
chtyim: Can simply be `getSchedulerQueue` or `getQueue`. The argument tells what it act on. Also, missing javadoc for public methods in public class.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/scheduler/SchedulerQueueResolver.java
```diff
@@ -0,0 +1,53 @@
...
+    NamespaceMeta meta = store.getNamespace(namespaceId);
+    Preconditions.checkNotNull(meta, "Namespace meta cannot be null");
+
+    NamespaceConfig config = meta.getConfig();
+    return config.getSchedulerQueueName() != null ? config.getSchedulerQueueName() : schedulerQueueFromCConf;
```
chtyim: better call `getDefault` for the null case.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1971,4 +1976,34 @@ private void deleteProcessMetricsForFlow(String application, String flow) throws
...
   }
+
+  private BasicArguments getSystemArguments(String namespaceId) {
+    // Get Configs from Cconf
+    Map<String, String> systemConfigsFromCDAP = getSystemArgumentsFromCConf();
+    // Override the Configs from configs at namespace level.
+    return new BasicArguments(getNamespaceResolvedSystemArguments(namespaceId, systemConfigsFromCDAP));
+  }
+
+  private Map<String, String> getSystemArgumentsFromCConf() {
```
chtyim: Same here. I prefer not to have `cConf` in the method name.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
chtyim: Why define the constants ourselves? Can we use `JobContext.QUEUE_NAME` from MapReduce?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/ScheduleTaskRunner.java
```diff
@@ -23,6 +23,7 @@
 import co.cask.cdap.app.runtime.ProgramOptions;
 import co.cask.cdap.app.runtime.ProgramRuntimeService;
 import co.cask.cdap.app.store.Store;
+import co.cask.cdap.common.conf.CConfiguration;
```
chtyim: Why new import but no usage?

chtyim:
Just couple comments from me.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
sreevatsanraman: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/scheduler/SchedulerQueueResolver.java
sreevatsanraman: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/scheduler/SchedulerQueueResolver.java
sreevatsanraman: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/scheduler/SchedulerQueueResolver.java
sreevatsanraman: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/ScheduleTaskRunner.java
sreevatsanraman: Fixed.

sreevatsanraman:
@chtyim addressed the comments. please take a look.

chtyim:
LGTM

PR #1532, opened 2015-03-08
[CDAP-870] Namespace Delete

bdmogal:
Summary
- [x] Added ``DELETE /v3/unrecoverable/namespaces/{namespace-id}`` API to delete a namespace. This is the equivalent of today's unrecoverable reset API, for a particular namespace.
- [x] Added ``DELETE /v3/unrecoverable/namespaces/{namespace-id}/datasets`` API to delete all datasets in a namespace. This is the equivalent of today's unrecoverable reset datasets API, for a particular namespace
- [x] Both the above APIs are governed by the ``UNRECOVERABLE_RESET_ENABLED`` option.
- [x] Removed ``DELETE /v3/namespaces/{namespace-id}`` API, so as to make deletion that much harder (you need to add the unrecoverable path param and set the unrecoverable configuration).
- [x] Moved the ``checkAnyRunning`` method from ``AppLifecycleHttpHandler`` to ``ProgramRuntimeService``, so it can be used from ``NamespaceAdmin`` and any other class that needs it.
- [x] Added tests in ``NamespaceHttpHandlerTest`` to illustrate how delete is supposed to work.
- [x] The v2 unrecoverable reset API calls namespace delete (on *default* namespace) apart from a bunch of other operations that are not namespace specific. This operation clears all the CDAP data in the default namespace, however, does not percolate the delete to Hive DB or HBase namespace. It also does not delete the default namespace from the MDS.

Issues Addressed:
* [CDAP-870](https://issues.cask.co/browse/CDAP-870)
* [CDAP-1501](https://issues.cask.co/browse/CDAP-1501)

Build:
* http://builds.cask.co/browse/CDAP-DUT1027-JOB1-1/

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
@@ -16,20 +16,19 @@
 
 package co.cask.cdap.gateway.handlers;
 
+import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.exception.AlreadyExistsException;
 import co.cask.cdap.common.exception.NotFoundException;
 import co.cask.cdap.gateway.auth.Authenticator;
 import co.cask.cdap.gateway.handlers.util.AbstractAppFabricHttpHandler;
 import co.cask.cdap.internal.app.namespace.NamespaceAdmin;
+import co.cask.cdap.internal.app.namespace.NamespaceCannotBeDeletedException;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.NamespaceMeta;
 import co.cask.http.HttpHandler;
 import co.cask.http.HttpResponder;
 import com.google.common.base.CharMatcher;
-import com.google.common.base.Charsets;
-import com.google.common.reflect.TypeToken;
-import com.google.gson.Gson;
 import com.google.gson.JsonSyntaxException;
 import com.google.inject.Inject;
 import org.jboss.netty.handler.codec.http.HttpRequest;
@@ -38,8 +37,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.lang.reflect.Type;
-import java.util.Map;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.PUT;
@@ -52,14 +49,14 @@ import javax.ws.rs.PathParam;
 @Path(Constants.Gateway.API_VERSION_3)
 public class NamespaceHttpHandler extends AbstractAppFabricHttpHandler {
   private static final Logger LOG = LoggerFactory.getLogger(NamespaceHttpHandler.class);
-  private static final Gson GSON = new Gson();
-  private static final Type STRING_MAP_TYPE = new TypeToken<Map<String, String>>() { }.getType();
 
+  private final CConfiguration cConf;
   private final NamespaceAdmin namespaceAdmin;
 
   @Inject
-  public NamespaceHttpHandler(Authenticator authenticator, NamespaceAdmin namespaceAdmin) {
+  public NamespaceHttpHandler(Authenticator authenticator, CConfiguration cConf, NamespaceAdmin namespaceAdmin) {
     super(authenticator);
+    this.cConf = cConf;
     this.namespaceAdmin = namespaceAdmin;
   }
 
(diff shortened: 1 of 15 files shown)


bdmogal:
Build is GREEN

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
```diff
@@ -164,9 +161,17 @@ public void create(HttpRequest request, HttpResponder responder, @PathParam("nam
...
   @DELETE
-  @Path("/namespaces/{namespace-id}")
+  @Path("/unrecoverable/namespaces/{namespace-id}")
   public void delete(HttpRequest request, HttpResponder responder, @PathParam("namespace-id") String namespace) {
+    // NOTE: DO NOT DOCUMENT
+    if (!cConf.getBoolean(Constants.Dangerous.UNRECOVERABLE_RESET, Constants.Dangerous.DEFAULT_UNRECOVERABLE_RESET)) {
+      responder.sendStatus(HttpResponseStatus.FORBIDDEN);
+      return;
+    }
     if (isReserved(namespace)) {
```
sreevatsanraman: So we cannot delete all apps, streams etc in default namespace? I thought the behavior should be to delete everything and not the namespace/database in hbase and explore

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
```diff
@@ -179,6 +184,39 @@ public void delete(HttpRequest request, HttpResponder responder, @PathParam("nam
...
+   */
+  @DELETE
+  @Path("/unrecoverable/namespaces/{namespace-id}/datasets")
+  public void deleteDatasets(HttpRequest request, HttpResponder responder,
+                             @PathParam("namespace-id") String namespace) {
+    if (!cConf.getBoolean(Constants.Dangerous.UNRECOVERABLE_RESET, Constants.Dangerous.DEFAULT_UNRECOVERABLE_RESET)) {
+      responder.sendStatus(HttpResponseStatus.FORBIDDEN);
+      return;
+    }
+    if (isReserved(namespace)) {
```
sreevatsanraman: Same here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -155,32 +167,85 @@ public void createNamespace(NamespaceMeta metadata) throws NamespaceCannotBeCrea
...
+      // Delete datasets and modules
       dsFramework.deleteAllInstances(namespaceId);
       dsFramework.deleteAllModules(namespaceId);
-      dsFramework.deleteNamespace(namespaceId);
+      // Delete queues and streams data
+      queueAdmin.dropAllInNamespace(namespaceId.getId());
+      streamAdmin.dropAllInNamespace(namespaceId);
+      // Delete all meta data
+      store.removeAll(namespaceId);
+      // TODO: CDAP-1729 - Delete/Expire Metrics. API unavailable right now.
```
sreevatsanraman: What happens to explore and Hbase namespaces?

sreevatsanraman:
Couple of questions on default and system namespaces. Rest look good.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
bdmogal: Updated to remove this check from here. Contents of default namespace can be deleted now.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
bdmogal: Updated to remove this check from here. Contents of default namespace can be deleted now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
bdmogal: The condition right below will ensure that the delete does not percolate to default namespaces in Hive, HBase, etc.

bdmogal:
@sreevatsanraman addressed comments. Updated a test in ``NamespaceHttpHandlerTest`` (testReserved) to reflect how it is supposed to work.

sreevatsanraman:
LGTM :+1:

bdmogal:
Unrelated test failure that passes consistently locally. Will merge instead of hogging resources on build slave for 2 hours again. Reported the failure in https://issues.cask.co/browse/CDAP-1683

PR #1536, opened 2015-03-09
CDAP-1368: Change internal handler tests to use v3 APIs

jwang47:
https://issues.cask.co/browse/CDAP-1368
http://builds.cask.co/browse/CDAP-DUT1018


jwang47:
Closing since this is probably not needed.

PR #1537, opened 2015-03-09
Bug/ui apps program api fix

ajainarayanan:
- Fixes /data/datasets endpoints
- Fixes /instances endpoint in services feature
- Fixes /apps/program-type endpoint. The response is now consolidated in GET /apps/appId


elfenheart:
:+1: LGTM

PR #1538, opened 2015-03-09
console log cleanup

elfenheart:


cdap-ui/server/express.js
```diff
@@ -148,7 +148,7 @@ function makeApp (authAddress, cdapConfig) {
 
   // any other path, serve index.html
   app.all('*', [
-    httpIndexLogger,
+    //httpIndexLogger,
```
gaarf: I think this one should remain uncommented, so we get at least a little bit of positive logging.

elfenheart:
@gaarf Comment addressed

gaarf:
:+1:

PR #1540, opened 2015-03-09
Fix/ui flows event count

elfenheart:
change the usage of bytes filter to number filter in flow graph. The metrics does not return bytes, but number of events.

Add preliminary unit test for it.


ajainarayanan:
:+1: LGTM

PR #1544, opened 2015-03-09
Feature/cleanup old metrics tables

gsps1:
Truncate destination tables before migrating data and option to delete old metrics data tables.

https://issues.cask.co/browse/CDAP-1661
https://issues.cask.co/browse/CDAP-1528


gsps1:
http://builds.cask.co/browse/CDAP-DUT956-2

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
```diff
@@ -181,7 +181,6 @@ public synchronized void addInstance(String datasetType, Id.DatasetInstance data
     }
     DatasetDefinition def = registry.get(datasetType);
     DatasetSpecification spec = def.configure(datasetInstanceId.getId(), props);
-    LOG.info("Dataset Specification name {} and properties {}", spec.getName(), spec.getProperties());
```
abaranau: why remove? I thought you added it for a reason. I'd though say "Creating dataset with spec name {} and properties {}..."

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -52,18 +55,23 @@
...
  */
 public class DataMigration {
+  private static final String KEEP_OLD_METRICS_TABLES = "--keep-old-metrics-tables";
+
   /**
    * Set of Action available in this tool.
    */
   private enum Action {
-    METRICS("Migrate metrics data"),
+    METRICS("Migrate metrics data, to preserve old table data use option --keep-old-metrics-tables"),
```
abaranau: is this description consistent with other descriptions in out tools like that?

abaranau: use KEEP_OLD_METRICS_TABLES constant in here ;)

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -113,7 +121,11 @@ protected void configure() {
 
       switch (action) {
         case METRICS:
-          migrateMetricsData(injector);
+          boolean keepOldTable = false;
```
abaranau: Before switch - remove action from args. So that at this point you have only args for specific action. Then, here:
* if > 1 argument - tell that invalid arguments, and print help
* if 1 argument - check that is it the KEEP_OLD_METRICS_TABLES, otherwise - tell invalid args, print help
* if 0 args - use default KEEP_OLD_METRICS_TABLES

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -52,18 +55,23 @@
...
+import java.util.Set;
 import javax.annotation.Nullable;
 
+
 /**
  * Command line tool to migrate data between different versions of CDAP.
  * Usually used along with upgrade tool{@link UpgraderMain}
  */
 public class DataMigration {
+  private static final String KEEP_OLD_METRICS_TABLES = "--keep-old-metrics-tables";
```
abaranau: keep old metrics data

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -150,23 +169,125 @@ private Action parseAction(String action) {
...
     // find version to migrate
     MetricHBaseTableUtil metricHBaseTableUtil = new MetricHBaseTableUtil(injector.getInstance(HBaseTableUtil.class));
     Version version = findMetricsTableVersion(injector.getInstance(CConfiguration.class),
                                               injector.getInstance(Configuration.class), metricHBaseTableUtil);
+
     if (version != null) {
       System.out.println("Migrating Metrics Data from " + version.name());
       DatasetFramework framework = createRegisteredDatasetFramework(injector);
+      // perform sanity check - truncate 2.8 tables data
+      truncateV2Tables(cConf, hConf);
```
abaranau: truncateDestinationTables - in 2.9 these are no longer V2, right? ;)

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -150,23 +169,125 @@ private Action parseAction(String action) {
...
     System.out.println("Starting metrics data migration");
     // find version to migrate
     MetricHBaseTableUtil metricHBaseTableUtil = new MetricHBaseTableUtil(injector.getInstance(HBaseTableUtil.class));
     Version version = findMetricsTableVersion(injector.getInstance(CConfiguration.class),
                                               injector.getInstance(Configuration.class), metricHBaseTableUtil);
+
     if (version != null) {
       System.out.println("Migrating Metrics Data from " + version.name());
       DatasetFramework framework = createRegisteredDatasetFramework(injector);
+      // perform sanity check - truncate 2.8 tables data
```
abaranau: remove 2.8 from comment too ;)

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -150,23 +169,125 @@ private Action parseAction(String action) {
...
+      truncateV2Tables(cConf, hConf);
       // migrate metrics data
       DefaultMetricDatasetFactory.migrateData(injector.getInstance(CConfiguration.class), framework, version);
+      // delete old-metrics table identified by version, unless keepOldTables flag is true
+      if (!keepOldTables) {
+        cleanUpOldTables(cConf, hConf, version);
```
abaranau: print out message to tell what is the current step - also do that for other steps if applicable: cleanup destination, migrate, cleanup old

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
abaranau: One thing (feel free to create separate JIRA): we want to fail by default if destination tables are not clean, unless user specifies argument --cleanup-destination or smth. I.e. if someone runs migration by mistake - it'll tell that destination tables exist and not empty, are you sure you want to delete them and migrate old data over? use --foo option to skip this check in that case. Smth like that

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -150,23 +169,125 @@ private Action parseAction(String action) {
...
+      }
       System.out.println("Successfully Migrated Metrics Data from " + version.name());
     } else {
       System.out.println("Did not find compatible CDAP Version to migrate Metrics data from");
     }
   }
 
+  private void truncateV2Tables(CConfiguration cConf, Configuration hConf) throws Exception {
+    String rootPrefix = cConf.get(Constants.Dataset.TABLE_PREFIX) + "_";
+    String v2EntityTableName =  cConf.get(MetricsConstants.ConfigKeys.ENTITY_TABLE_NAME,
```
abaranau: no v2, dest

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -150,23 +169,125 @@ private Action parseAction(String action) {
...
+      if (!keepOldTables) {
+        cleanUpOldTables(cConf, hConf, version);
+      }
       System.out.println("Successfully Migrated Metrics Data from " + version.name());
     } else {
       System.out.println("Did not find compatible CDAP Version to migrate Metrics data from");
     }
   }
 
+  private void truncateV2Tables(CConfiguration cConf, Configuration hConf) throws Exception {
```
abaranau: should all this be part of MetricsDataMigrator? You just pass arguments to it: delete old data, cleanup destination, etc.

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -150,23 +169,125 @@ private Action parseAction(String action) {
...
+    String rootPrefix = cConf.get(Constants.Dataset.TABLE_PREFIX) + "_";
+    String v2EntityTableName =  cConf.get(MetricsConstants.ConfigKeys.ENTITY_TABLE_NAME,
+                                          MetricsConstants.DEFAULT_ENTITY_TABLE_NAME);
+    v2EntityTableName = getTableName(rootPrefix, Id.DatasetInstance.from(
```
abaranau: I think in migrator you already compute those names - if you move there, extract methods to avoid code duplication

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
gsps1: oh, i added it for debugging purpose.

abaranau:
couple comments. Also, you may want to unit-test parsing args and cleanup of old tables. If that is easy - I know it is stretch for code freeze. So you can add it later

abaranau:
some more!

abaranau:
Big +1 for adding extra unit-testing

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
abaranau: let's keep - with the message I suggested

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -77,53 +83,68 @@ private String getDescription() {
...
-          break;
-        case HELP:
-          printHelp();
-          break;
+    if (action.equals(Action.METRICS)) {
+      if (args.length > 2) {
+        System.out.println("invalid number of arguments");
+        printHelp(true);
+        return false;
+      } else if (args.length == 2) {
```
abaranau: missing else after it? what if the length is 1?

aslo - extract into a method - migrate metric step

gsps1: that's cool right? if length is 1, it would be action -> metrics and we by default assume to delete old tables

abaranau: oh - sorry - I see this is parsing args only...

abaranau: real confusing. So if you add more actions, more arguments - you'll have all those variables here and just use some of them in the second switch? confusing.

Can we have 
``` java
private static interface MigrationAction {
  void perform(Injector injector);
}

private static final class MetricsMigration implements MigrationAction {
    public MetricsMigration(keepOldMetricsData) {
        this. keepOldMetricsData = keepOldMetricsData;
    }

    void perform (injector) {
          // whatever in migrateMetricsData(injector, keepOldMetricsData) method
    }
}
```

then as result of parsing arguments you'll have MigrationAction instance that you will just run after constructing injector.

Makes sense?

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -77,53 +83,68 @@ private String getDescription() {
...
+          return false;
+        }
       }
-    } catch (Exception e) {
-      System.out.println(String.format("Failed to perform action '%s'. Reason: '%s'.", action, e.getMessage()));
-      e.printStackTrace(System.out);
-      throw e;
     }
+
+    if (isTest) {
```
abaranau: with doing as in the above comment - you can extract method that will take in args and return MigrationAction instance. Then you'll test that method - no need to hack with isTest field, etc.

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -150,17 +171,29 @@ private Action parseAction(String action) {
...
+  private void migrateMetricsData(Injector injector, boolean keepOldTables) {
+    CConfiguration cConf = injector.getInstance(CConfiguration.class);
+    Configuration hConf = injector.getInstance(Configuration.class);
+
     System.out.println("Starting metrics data migration");
     // find version to migrate
     MetricHBaseTableUtil metricHBaseTableUtil = new MetricHBaseTableUtil(injector.getInstance(HBaseTableUtil.class));
-    Version version = findMetricsTableVersion(injector.getInstance(CConfiguration.class),
-                                              injector.getInstance(Configuration.class), metricHBaseTableUtil);
+    Version version = findMetricsTableVersion(cConf, hConf, metricHBaseTableUtil);
```
abaranau: didn't we say figuring out version, etc. should go into the metrics migrator?

abaranau:
couple more

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
gsps1: this is InMemoryDatasetFramework, the logging we discussed are in data migration. are you sure we need this log?

abaranau: nah

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -81,51 +85,68 @@ public void doMain(String[] args) throws Exception {
     System.out.println(String.format("%s - version %s.", getClass().getSimpleName(), ProjectInfo.getVersion()));
     System.out.println();
 
-    if (args.length < 1) {
-      printHelp();
+    MigrationAction action = getMigrationActionFromArguments(args);
```
abaranau: getAction()

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
```diff
@@ -81,51 +85,68 @@ public void doMain(String[] args) throws Exception {
...
+    if (action == null) {
       return;
```
abaranau: you wanna log smth?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/upgrade/DataMigrationException.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.cdap.metrics.store.upgrade;
+
+/**
+ * thrown during failure while migrating data
```
abaranau: Thrown

abaranau:
very minor comments. Merge after you fix

cdap-master/src/main/java/co/cask/cdap/data/tools/DataMigration.java
gsps1: no, we are already printing in getAction

gsps1:
http://builds.cask.co/browse/CDAP-DUT956-4/test - GREEN

PR #1545, opened 2015-03-09
CDAP-1735: Deduplicate logic between UnitTestManager and IntegrationTestManager

jwang47:
Depends on https://github.com/caskdata/cdap/pull/1510 and https://github.com/caskdata/cdap/pull/1546.

https://issues.cask.co/browse/CDAP-1735
http://builds.cask.co/browse/CDAP-DUT1035


jwang47:
Working on getting tests to pass right now.

PR #1549, opened 2015-03-09
CDAP-1726: Support sending content of a file to a service

jwang47:
https://issues.cask.co/browse/CDAP-1726
http://builds.cask.co/browse/CDAP-DUT1038


cdap-cli/src/main/java/co/cask/cdap/cli/command/CallServiceCommand.java
```diff
@@ -114,9 +127,10 @@ public void perform(Arguments arguments, PrintStream output) throws Exception {
 
   @Override
   public String getPattern() {
-    return String.format("call service <%s> <%s> <%s> [headers <%s>] [body <%s>]",
+    return String.format("call service <%s> <%s> <%s> [headers <%s>] [body <%s>] [body:file <%s>]",
```
ParthGandhi: Will this allow me to specify both a string and a file? Or will the CLI auto-complete to only one of them?

cdap-cli/src/main/java/co/cask/cdap/cli/command/CallServiceCommand.java
```diff
@@ -86,16 +87,28 @@ public void perform(Arguments arguments, PrintStream output) throws Exception {
...
+    }
 
     Map<String, String> headerMap = GSON.fromJson(headers, new TypeToken<Map<String, String>>() { }.getType());
     URL url = new URL(serviceClient.getServiceURL(appId, serviceId), path);
 
     HttpMethod httpMethod = HttpMethod.valueOf(method);
     HttpRequest.Builder builder = HttpRequest.builder(httpMethod, url).addHeaders(headerMap);
-    if (!body.isEmpty() && httpMethod != HttpMethod.GET) {
-      builder.withBody(body);
+    if (httpMethod != HttpMethod.GET) {
```
ParthGandhi: I don't think we should restrict this. You're right that usually it is never meaningful to have a body with a `GET` call, but I believe theres nothing in the HTTP spec that actually says this.

ParthGandhi:
A couple of minor comments.

cdap-cli/src/main/java/co/cask/cdap/cli/command/CallServiceCommand.java
anew: See https://groups.yahoo.com/neo/groups/rest-discuss/conversations/messages/9962. It makes no sense to include a body with a GET.

jwang47: Unfortunately Java's `HttpURLConnection` throws an exception if you have a request body along with a `GET` call, so I'm just ignoring the body in that case.. but probably I should just give a user-friendly error.

cdap-cli/src/main/java/co/cask/cdap/cli/command/CallServiceCommand.java
jwang47: CLI will auto-complete to either, but if you send a `body` and `body:file`, then the command will throw an exception. This isn't very good since nothing in the pattern indicates this, but I can add a note to the description.

cdap-cli/src/main/java/co/cask/cdap/cli/command/CallServiceCommand.java
ParthGandhi: Yes, of course it makes no sense. I was just pointing out that the spec itself doesn't forbid it.

ParthGandhi:
Answered all my questions. Try and make better errors. Otherwise LGTM.

cdap-cli/src/main/java/co/cask/cdap/cli/command/CallServiceCommand.java
anew: You should also update the getDescription to mention body:file

anew:
I have one more comment - update the description of the command in the help text. Other than that LGTM

anew:
LGTM

PR #1552, opened 2015-03-09
[CDAP-1728] Prevent stream size scheduler from crashing with NPE when stopped

jguery:
JIRA: https://issues.cask.co/browse/CDAP-1728


anwar6953:
lgtm

PR #1546, opened 2015-03-09
CDAP-1300: ScheduleClient for listing, suspending, and resuming schedules

jwang47:
https://issues.cask.co/browse/CDAP-1300
http://builds.cask.co/browse/CDAP-DUT1037

PR #1553, opened 2015-03-09
CDAP-1738: Schedule CLI: list, suspend, resume, get status

jwang47:
Depends on https://github.com/caskdata/cdap/pull/1546

https://issues.cask.co/browse/CDAP-1738
http://builds.cask.co/browse/CDAP-DUT1040


cdap-cli/src/main/java/co/cask/cdap/cli/commandset/ScheduleCommands.java
```diff
@@ -0,0 +1,228 @@
...
+      String[] programIdParts = arguments.get(ElementType.WORKFLOW.getArgumentName().toString()).split("\\.");
+      if (programIdParts.length < 2) {
+        throw new CommandInputError(this);
+      }
+
+      final String appId = programIdParts[0];
+      String scheduleId = programIdParts[1];
+      List<ScheduleSpecification> list = scheduleClient.list(appId, scheduleId);
+      Table table = Table.builder()
+        .setHeader("application", "program", "program type", "name", "description", "properties")
```
jguery: I think you'd also want to print the type of the schedule, ie Time, or StreamSize - we only have those two as of now. You can figure it out by doing the following:

```
if (object.getSchedule() instanceof TimeSchedule) {
  // type is time
} else if (object.getSchedule() instanceof StreamSizeSchedule) {
  // type is stream size
}
```

cdap-cli/src/main/java/co/cask/cdap/cli/commandset/ScheduleCommands.java
```diff
@@ -0,0 +1,228 @@
...
+import com.google.inject.Inject;
+import com.google.inject.Injector;
+
+import java.io.PrintStream;
+import java.util.List;
+
+/**
+ * Schedule commands.
+ */
+public class ScheduleCommands extends CommandSet<Command> implements Categorized {
```
jguery: Should there also be a command to describe one schedule in more depth? Because, when listing the schedules, you only print their names and their types - time or data - but not the cron entry, for a time schedule, or the stream name and the data trigger for a stream size schedule.

jguery: I guess this would require an additional API in the ScheduleClient, something like ``ScheduleClient.getSchedule(String appId, String scheduleId)``.

jguery: Actually I see that there is no endpoint in ``ProgramLifecycleHttpHandler`` to get the specification of one schedule. But I still think that more info should be printed for each schedule. Maybe have in the list command, another column that contains the arguments of each specific schedule. I'm not sure how that should look like.

cdap-client-tests/src/main/java/co/cask/cdap/client/app/FakeApp.java
```diff
@@ -55,7 +58,10 @@ public void configure() {
     createDataset(DS_NAME, FakeDataset.class.getName());
     addProcedure(new FakeProcedure());
     addFlow(new FakeFlow());
-    addSpark(new FakeSpark());
+    addWorkflow(new FakeWorkflow());
+    scheduleWorkflow(new TimeSchedule(SCHEDULE_NAME, "", "0 0 1 1 *"), FakeWorkflow.NAME);
```
jguery: In that line, and the next, use ``Schedules.create...`` instead of instantiating the ``TimeSchedule`` and ``StreamSizeSchedule`` classes directly. They are in the ``api.internal`` package, and are not meant to be used directly by app developers.

albertshau:
lgtm

jwang47:
Now just waiting for some flaky tests to pass..

PR #1554, opened 2015-03-10
Fixes workflow diagram click handler for individual nodes in workflows

ajainarayanan:
Fixes individual nodes in workflow diagram be un-clickable. (for 2.8.0 alone)


cdap-ui/app/features/workflows/templates/tabs/runs.html
```diff
@@ -1,4 +1,4 @@
-<div ng-show="runs.length > 0">
+<div>
```
elfenheart: do we need this div?

elfenheart:
:+1: LGTM

PR #1533, opened 2015-03-09
Feature/cdap 1628 upgrade explore

albertshau:

PR #1556, opened 2015-03-10
(CDAP-1685) Additional namespace explore tests

ParthGandhi:
Mostly adding some test cases for Namespaced explore.

Also fixes a couple of small bugs introduced by a recent change (hive databases beginning with "cdap_*").

Build: http://builds.cask.co/browse/CDAP-DUT1029-4/


cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -634,11 +634,12 @@ public QueryHandle createNamespace(Id.Namespace namespace) throws ExploreExcepti
       Map<String, String> sessionConf = startSession();
       SessionHandle sessionHandle = cliService.openSession("", "", sessionConf);
 
+      String database = getHiveDatabase(namespace.getId());
       // "IF NOT EXISTS" so that this operation is idempotent.
-      String statement = String.format("CREATE DATABASE IF NOT EXISTS %s", getHiveDatabase(namespace.getId()));
+      String statement = String.format("CREATE DATABASE IF NOT EXISTS %s", database);
```
bdmogal: unrelated to this PR, but shouldn't it fail if the db exists? You should only be able to create the DB once - during namespace creation, no?

cdap-explore/src/test/java/co/cask/cdap/explore/service/BaseHiveExploreServiceTest.java
```diff
@@ -99,10 +99,15 @@
     .registerTypeAdapter(Schema.class, new SchemaTypeAdapter())
     .create();
 
-  // TODO: When datasets are namespaced, we should add tests for multiple, non-default namespaces.
   protected static final Id.Namespace NAMESPACE_ID = Id.Namespace.from("default");
+  protected static final Id.Namespace OTHER_NAMESPACE_ID = Id.Namespace.from("other");
```
bdmogal: can we use 2 non-default namespaces in tests? default namespace can really cause indeterministic behavior in tests, since you do not control its lifecycle as a user.

bdmogal: eventually, we should probably get to a state where we only use default namespace in tests that:
1. test v2 apis
2. test backward compatibility

bdmogal:
@ParthGandhi took a cursory look right now. Will do a more detailed review tomorrow morning. However, have a couple of comments about default namespace, let me know what you think about them.

bdmogal:
Btw, you're going to have a fair amount of conflicts with #1533

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
albertshau: i agree with this being idempotent, especially since explore operations are often done as a step of some larger operation

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -657,10 +658,10 @@ public QueryHandle deleteNamespace(Id.Namespace namespace) throws ExploreExcepti
       SessionHandle sessionHandle = openSession(sessionConf);
 
       String database = getHiveDatabase(namespace.getId());
-      String statement = String.format("DROP DATABASE %s", database);
+      String statement = String.format("DROP DATABASE %s CASCADE", database);
```
albertshau: hm... this seems dangerous. What if there is table in the database that is not owned by cdap?

albertshau: also, what happens if this is the default database?

albertshau: actually, regarding the "cdap_" prefix, we should be using the root prefix from CConf instead of hardcoding to cdap

ParthGandhi: Hm good point. Let me think about it.

cdap-explore/src/test/java/co/cask/cdap/explore/service/ExploreMetadataTestRun.java
```diff
@@ -74,9 +89,13 @@ public void testGetTables() throws Exception {
                           ),
                           Lists.newArrayList(
                             new QueryResult(Lists.<Object>newArrayList(
-                              "", "default", "my_table", "TABLE", "CDAP Dataset")),
+                              "", OTHER_NAMESPACE_DATABASE, "my_table", "TABLE", "CDAP Dataset")),
```
albertshau: use variables for the name instead of a hardcoded string so that it is easier to update if table names change. (incoming PR prepends "dataset_" or "stream_" to table names)

ParthGandhi:
@albertshau @bdmogal Addressed comments. Please review.

ParthGandhi:
Green build: https://builds.cask.co/browse/CDAP-DUT1029-8

cdap-explore/src/test/java/co/cask/cdap/explore/service/ExploreMetadataTestRun.java
```diff
@@ -91,7 +112,28 @@ public void testGetTables() throws Exception {
...
+                             "", OTHER_NAMESPACE_DATABASE, namespacedOtherTableName, "TABLE", "CDAP Dataset")))
+
+    );
+
+    // Pattern on database
+    future = getExploreClient().tables(null, OTHER_NAMESPACE_ID.getId(), "%", null);
+    assertStatementResult(future, true,
+                          Lists.newArrayList(
+                            new ColumnDesc("TABLE_CAT", "STRING", 1, "Catalog name. NULL if not applicable."),
+                            new ColumnDesc("TABLE_SCHEM", "STRING", 2, "Schema name."),
```
albertshau: since tests pass, this isn't a typo?

albertshau:
lgtm. lets merge and i can resolve conflicts on the other PR

cdap-explore/src/test/java/co/cask/cdap/explore/service/ExploreMetadataTestRun.java
ParthGandhi: Nope. Not a typo. I've not actually understood how Hive abbreviates properties though... It seems arbitrary.

ParthGandhi:
Thanks @albertshau . Merging.

PR #1559, opened 2015-03-10
Modify Purchase app to include a stream size schedule

jguery:


sagarkapare:
Code looks good to me. Can you please also make the corresponding changes in the `PurchaseApp` documentation?

jguery:
Yes, will do.

jguery:
@sagarkapare I modified the docs to mention the new schedule. Yet, there was no section to explain in detail what the already existing schedule does, so I didn't create it.

cdap-examples/Purchase/src/main/java/co/cask/cdap/examples/purchase/PurchaseApp.java
```diff
@@ -65,7 +65,16 @@ public void configure() {
     addService(new CatalogLookupService());
 
     // Schedule the workflow
-    scheduleWorkflow(Schedules.createTimeSchedule("DailySchedule", "", "0 4 * * *"), "PurchaseHistoryWorkflow");
+    scheduleWorkflow(Schedules.createTimeSchedule("DailySchedule", "Schedule execution every day", "0 4 * * *"),
```
awholegunch: Should this be called "TimeSchedule" instead of "DailySchedule" to be consistent?

cdap-examples/Purchase/src/main/java/co/cask/cdap/examples/purchase/PurchaseApp.java
```diff
@@ -65,7 +65,16 @@ public void configure() {
...
+                     "PurchaseHistoryWorkflow");
+
+    // Schedule the workflow based on the data coming in the purchaseStream stream
+    scheduleWorkflow(
+      Schedules.createDataSchedule("DataSchedule",
+                                   "Schedule execution when 1MB of data is ingested in the purchaseStream",
```
awholegunch: 1 MB or more of data ?

cdap-examples/Purchase/src/main/java/co/cask/cdap/examples/purchase/PurchaseApp.java
jguery: I think "daily schedule" already contain the information that it is a time schedule. And it says more: than it runs exactly once a day.

cdap-examples/Purchase/src/test/resources/cdap-site.xml
```diff
@@ -0,0 +1,27 @@
...
+  ~ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  ~ License for the specific language governing permissions and limitations under
+  ~ the License.
+  -->
+<configuration>
+
+  <property>
+    <name>stream.notification.threshold</name>
+    <value>1</value>
+    <description>Default size of data, in MB, to be ingested by a Stream before a notification is published</description>
```
awholegunch: This is not a default, but is instead the setting being used, correct?
```
Amount of data, in MB, to be ingested by a Stream before a notification is published
```

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -231,8 +231,8 @@
 
     <property>
         <name>stream.notification.threshold</name>
-        <value>1000</value>
-        <description>Size of data, in MB, to be ingested by a Stream before a notification is published</description>
+        <value>1024</value>
+        <description>Default size of data, in MB, to be ingested by a Stream before a notification is published</description>
```
awholegunch: ~/Source/cdap/cdap-docs/admin-manual/source/installation/cdap-site.rst
needs to be updated as well.

awholegunch:
three comments; a commit is needed from cdap-site.rst. Otherwise looking good.

cdap-examples/Purchase/src/main/java/co/cask/cdap/examples/purchase/PurchaseApp.java
jguery: Correct, will change.

cdap-examples/Purchase/src/test/resources/cdap-site.xml
jguery: No it is actually still a default, I just change the default value here in the context of the tests of the purchase app.

cdap-common/src/main/resources/cdap-default.xml
jguery: Right, will do.

jguery:
@awholegunch I addressed your comments, please have another look. Thank you.

cdap-examples/Purchase/src/test/resources/cdap-site.xml
awholegunch: Sounds good then.

awholegunch:
LGTM.

PR #1560, opened 2015-03-10
Fixes minor form validation in namespace create form

ajainarayanan:


cdap-ui/app/features/admin/templates/namespace/create.html
```diff
@@ -24,6 +27,7 @@
              ng-model="model.name"
              class="form-control"
              placeholder="Name"
+             required
```
elfenheart: this is not actually required. Do we want to force user to specify name?

ajainarayanan: We use name to display in the UI. I think that would be necessary for us to display it in the UI.

elfenheart:
:+1: LGTM

PR #1561, opened 2015-03-10
Ignore NamespaceClientTestRun till its fixed

bdmogal:
Ignoring NamespaceClientTestRun until its fixed correctly. Currently state from other tests seems to be impacting it.


jwang47:
LGTM

PR #1562, opened 2015-03-10
Fix broken Javadoc build.

awholegunch:
Two small changes: to fix the broken Javadoc build, and fix a few spelling mistakes I stumbled on upon the way.


anew:
LGTM

PR #1563, opened 2015-03-10
add documentation for the sports results example

anew:


cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+     aggregates total counts such as games won and lost, or points scored and conceded, across all
+     seasons, and writes these totals to the partitioned file set ``totals`` that is partitioned by league.
+   - Both the original game results and the aggregated totals can be explored using ad-hoc SQL queries.
+
+Let's look at some of these components, and then run the Application and see the results.
+
+The SportResults Application
+----------------------------
+
+As in the other :ref:`examples,<examples-index>` the components
```
awholegunch: I think it looks better if punctuation isn't part of link:
```
:ref:`examples <examples-index>`,
```

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+=============
+
+A Cask Data Application Platform (CDAP) Example demonstrating partitioned file sets through sport results analytics.
+
+Overview
+========
+
+This Application demonstrates the use of the PartitionedFileSet dataset:
+
+   - Game results are stored in the PartitionedFileSet ``results``. It is partitioned by league and season,
```
awholegunch: These bulleted points should not be indented but instead flush left. (The lines after the first are indented, but the entire block should be moved to the left margin.) Common.

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+   - The first dataset (``results``) is partitioned by league and season. Each record represents
+     a single game with a date, a winning and a losing team, and the winner's and the loser's points, for example::
+
+       2011/9/5,Dallas Cowboys,New York Giants,24,17
+       2011/9/9,Philadelphia Eagles,Cleveland Browns,17,16
+       2011/9/9,New England Patriots,Tennessee Titans,34,13
+
+     We have included some sample data in the ``resources`` directory.
+
+   - The ``totals`` dataset stores aggregates across all seasons and therefore has the league as its single
```
awholegunch: perhaps:

and thus has the league ....

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+     partitioning field. Each record represents the total number of games won and lost, and the total number
+     of points scored and conceded, by a team.
```
awholegunch: Each record has, for an individual team, the total number of games won and lost and the total number
of points scored and conceded.

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+
+We will use the ``UploadService`` to upload files into the ``results`` dataset, then compute the ``totals``
```
awholegunch: to upload the sample data files

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+    :lines: 56-59
+
+Let's take a closer look at the upload method:
+
+   - It first creates a partition key and a relative file path from the league and season
+     received as path parameters in the request URL.
+   - It then uses the ``getLocation()`` of the embedded file set of the ``results`` dataset to obtain the location
+     for writing the file, and opens an output stream for that location to write the file contents.
+     ``Location`` is a file system abstraction from `Apache™ Twill® <http://twill.incubator.apache.org>`__;
+     you can read more about its interface in the Apache Twill
```
awholegunch: Start the link earlier, before "Apache":
```
`Apache
```

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+==============================
+
+``ScoreCounter`` is a simple MapReduce that reads from the ``results`` PartitionedFileSet and writes to
+the ``totals`` PartitionedFileSet. The ``beforeSubmit()`` method prepares the MapReduce program for this:
+
+   - It reads the league that it is supposed to process from the runtime arguments.
+   - It constructs a partition filter for the input using the league as the only condition, and instantiates
+     the ``results`` dataset with arguments that contain this filter.
+   - It constructs an output partition key for the new partition, and instantiates the ``totals`` dataset
+     with arguments that specify this partition key.
```
awholegunch: with arguments specifying that partition key.

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+
+.. literalinclude:: /../../../cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/ScoreCounter.java
+    :language: java
+    :lines: 60-86
+
+After the MapReduce has completed, the ``onFinish()`` method finalizes the addition of the new partition.
+Note that this is only needed due to a current limitation in the ``PartitionedFileSet`` implementation: itself
+should add the partition in the output committer of its output format (this will be addressed in a future CDAP release):
```
awholegunch: I'm not sure how to correct this: but "itself" doesn't sound correct. What did you mean? 
The MapReduce itself should add the partition in the output committer of its output format, but as it currently doesn't, this must be done in the ``onFinish`` method ?

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+  <#building-an-example-application>`__) or use the pre-built JAR file included in the CDAP SDK.
+- Start CDAP, deploy and start the application and its component as described below in 
+  `Running CDAP Applications`_\ .
+  Make sure you start the Service as described below.
+- Once the application has been deployed and started, you can `run the example. <#running-the-example>`__
+
+Running CDAP Applications
+=========================
+
+.. |example| replace:: FileSetExample
```
awholegunch: replace:: SportResults

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+
+    * - On Linux:
+      - ``$ ./bin/cdap-cli.sh start service SportResults.UploadService``
+    * - On Windows:
+      - ``> bin\cdap-cli.bat start service SportResults.UploadService``
+
+Uploading Game Results
+----------------------
+
+First we will upload some CSV files into the ``results`` dataset. For example, to upload the results
```
awholegunch: First, we will upload some CSV files...

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+for the 2012 season of the NFL::
+
+  cdap-cli.sh call service SportResults.UploadService PUT leagues/nfl/seasons/2012 body:file resources/nfl-2012.csv
```
awholegunch: Is this path going to work? I'm thinking it should be based on the SDK home (though awfully long):
```
 ``$ ./bin/cdap-cli.sh call service SportResults.UploadService PUT leagues/nfl/seasons/2012 body:file examples/SportResults/resources/nfl-2012.csv``
```

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Feel free to add more seasons and sport leagues:
+
+  cdap-cli.sh call service SportResults.UploadService PUT leagues/nfl/seasons/2013 body:file resources/nfl-2013.csv
+  cdap-cli.sh call service SportResults.UploadService PUT leagues/nba/seasons/2012 body:file resources/nba-2012.csv
```
awholegunch: ibid., ibid...

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+
+To run the ``ScoreCounter`` over all seasons of the NFL::
+
+  cdap-cli.sh start mapreduce SportResults.ScoreCounter \"league=nfl\"
+
+Exploring with Ad-hoc SQL
+-------------------------
+
+Both of the partitioned file sets are registered as external tables in Hive and can be explored with SQL. To
+see the existing partitions of a dataset, use the ``show partitions`` query:
```
awholegunch: query:: (double colon; note that the next line is not colored blue)

cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
```diff
@@ -37,11 +37,11 @@ public void configure() {
     createDataset("results", PartitionedFileSet.class, PartitionedFileSetProperties.builder()
       // properties for partitioning
       .setPartitioning(Partitioning.builder().addStringField("league").addIntField("season").build())
-        // properties for file set
+      // properties for file set
```
awholegunch: Properties

cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
```diff
@@ -51,11 +51,11 @@ public void configure() {
     createDataset("totals", PartitionedFileSet.class, PartitionedFileSetProperties.builder()
       // properties for partitioning
       .setPartitioning(Partitioning.builder().addStringField("league").build())
-        // properties for file set
+      // properties for file set
```
awholegunch: Properties

cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
```diff
@@ -51,11 +51,11 @@ public void configure() {
...
       .setInputFormat(TextInputFormat.class)
       .setOutputFormat(TextOutputFormat.class)
       .setInputProperty(TextOutputFormat.SEPERATOR, ",")
-        // properties for Explore (to create a partitioned Hive table)
+      // properties for Explore (to create a partitioned Hive table)
```
awholegunch: Properties

cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
```diff
@@ -37,11 +37,11 @@ public void configure() {
     createDataset("results", PartitionedFileSet.class, PartitionedFileSetProperties.builder()
       // properties for partitioning
       .setPartitioning(Partitioning.builder().addStringField("league").addIntField("season").build())
-        // properties for file set
+      // properties for file set
       .setInputFormat(TextInputFormat.class)
       .setOutputFormat(TextOutputFormat.class)
       .setInputProperty(TextOutputFormat.SEPERATOR, ",")
```
awholegunch: Shouldn't this be spelt "SEPARATOR" ??

cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
```diff
@@ -51,11 +51,11 @@ public void configure() {
     createDataset("totals", PartitionedFileSet.class, PartitionedFileSetProperties.builder()
       // properties for partitioning
       .setPartitioning(Partitioning.builder().addStringField("league").build())
-        // properties for file set
+      // properties for file set
       .setInputFormat(TextInputFormat.class)
       .setOutputFormat(TextOutputFormat.class)
       .setInputProperty(TextOutputFormat.SEPERATOR, ",")
```
awholegunch: SEPARATOR

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+
+  cdap-cli.sh call service SportResults.UploadService PUT leagues/nfl/seasons/2013 body:file resources/nfl-2013.csv
+  cdap-cli.sh call service SportResults.UploadService PUT leagues/nba/seasons/2012 body:file resources/nba-2012.csv
+
+Starting the MapReduce
+----------------------
+
+To run the ``ScoreCounter`` over all seasons of the NFL::
+
+  cdap-cli.sh start mapreduce SportResults.ScoreCounter \"league=nfl\"
```
awholegunch: Can the MapReduce be started from the GUI?
In the other examples, we have shown both Linux and Windows command calls. Are you thinking that this is unnecessary?

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,212 @@
...
+-------------------------
+
+Both of the partitioned file sets are registered as external tables in Hive and can be explored with SQL. To
+see the existing partitions of a dataset, use the ``show partitions`` query:
+
+  cdap-cli.sh execute \"show partitions results\"
+
+
+For example, to find the three games with the highest point difference in the 2012 NFL season, over all
+FL seasons (that have been uploaded), and for all seasons of all sport leagues::
```
awholegunch: NFL ??

cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
awholegunch: Guess this deserves an issue...https://issues.cask.co/browse/CDAP-1757 but I'm not sure how we should handle it. Just document it everywhere we use it?

awholegunch: Looks like from this (https://hadoop.apache.org/docs/r2.3.0/api/src-html/org/apache/hadoop/mapreduce/lib/output/TextOutputFormat.html#line.45), that this is the source of the mis-spelling.

awholegunch:
Let me know when you'd like me to look at it again.

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: This was copied from an existing example... better check everywhere else, too. Found 4 more and fixed.

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: That was also copied from another example. I will change it here, but won't go through other examples.

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: ok

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: ok

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: This was also copied... fixing.

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: ok

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: this is your wording. You wrote this as a correction for what I wrote for the Stream Conversion example. It means, the dataset itself should do it. We could say instead: 

Note that this is only needed due to a current limitation in the ``PartitionedFileSet`` implementation: the partition should be added by the output committer of the dataset's output format (this will be addressed in a future CDAP release):

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: ok

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: ok

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: Again copied from other examples. We should do it the same way everywhere. What do you think?

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: oh right

cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
anew: John, developers typically do not start comments in upper case. If you want to change that, we should do a pass over all the examples. I can change it here, but how pedantic do we need to be here?

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: It is extremely annoying to repeat every cli command twice, once with cdap-cli.sh and once with cdap-cli.bat. I would rather throw away all the windows specific ones, everywhere, and add a note somewhere, that on Windows, you need to use cdap-cli.bat instead.

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: oops.

anew:
@awholegunch please take another look

cdap-docs/examples-manual/source/examples/sport-results.rst
awholegunch: Sounds good.

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,211 @@
...
+
+  .. list-table::
+    :widths: 20 80
+    :stub-columns: 1
+
+    * - On Linux:
+      - ``$ ./bin/cdap-cli.sh start service SportResults.UploadService``
+    * - On Windows:
+      - ``> bin\cdap-cli.bat start service SportResults.UploadService``
+
```
awholegunch: This might be a good place to add a note about using the windows cdap-cli.bat:
```
In the examples that follow, Windows users should substitute ``bin\cdap-cli.bat`` for ``cdap-cli.sh`` as appropriate.
```

cdap-docs/examples-manual/source/examples/sport-results.rst
awholegunch: I am fine with that. In a similar fashion, we explain that curl.exe is included for Windows users, and expect them to handle that themselves. See note above.

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,211 @@
...
+Uploading Game Results
+----------------------
+
+First, we will upload some CSV files into the ``results`` dataset. For example, to upload the results
+for the 2012 season of the NFL::
+
+  cdap-cli.sh call service SportResults.UploadService PUT leagues/nfl/seasons/2012 body:file resources/nfl-2012.csv
+
+Feel free to add more seasons and sport leagues:
```
awholegunch: Feel free to add more seasons and sport leagues::

cdap-docs/examples-manual/source/examples/sport-results.rst
awholegunch: The only problem with these calls is that above we say "From the Standalone CDAP SDK directory", but these calls won't work from that directory. We could:
- suggest that they add the ``bin`` directory to their path, then change to the example directory, and run the command from there; or
- modify the paths (but they get really long); or
- could/can the cdap-cli change its current directory, when run interactively? You could start it, and then cd into the example, and then run the commands.

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: done, and also in stream-conversion.rst

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: done.

cdap-docs/examples-manual/source/examples/sport-results.rst
anew: oops, that was one more. Hope we got all of them now.

cdap-docs/examples-manual/source/examples/sport-results.rst
```diff
@@ -0,0 +1,204 @@
...
+.. include:: /../../developers-manual/source/getting-started/building-apps.rst
+   :start-line: 11
+
+Running the Example
+===================
+
+In the examples that follow, for brevity we will simply use ``cdap-cli.sh`` for the Command Line Interface.
+Substitute the actual path for ``bin/cdap-cli.sh``, or ``bin\cdap-cli.bat`` on Windows, as appropriate. We
+assume that you will run all commands from the example's base directory (``examples/SportResults`` under the
+Standalone CDAP SDK directory).
```
awholegunch: +1

awholegunch:
LGTM.

PR #1541, opened 2015-03-09
Feature/ui v3 explore streams

gaarf:
adds a form to set start/end/limit of event search for streams/explore

PR #1564, opened 2015-03-10
Feature/ui timestamp picker

gaarf:
this PR depends on #1541 , review that first.

demo


elfenheart:
This PR is going to go to #1585

PR #1565, opened 2015-03-10
Refactor HTableNameConverter

chtyim:
- Get rid of unnecessary static
  - Static is really evil
- Cleanup code, remove unneeded methods
- Simplify the HTableNameConverter and sub-classes

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
@@ -19,9 +19,9 @@ package co.cask.cdap.data2.util.hbase;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.data2.util.TableId;
 import co.cask.cdap.proto.Id;
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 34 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
```diff
@@ -67,78 +67,72 @@ private static String getBackwardCompatibleTableName(String tablePrefix, TableId
...
       String prefix = parts[0];
-      hTableName = parts[1];
-      return PrefixedTableId.from(prefix, namespace, hTableName);
+      qualifier = parts[1];
+      return new PrefixedTableId(prefix, namespace, qualifier);
     }
 
-
-    String[] parts = hBaseNamespace.split("_");
+    // If HBase namespace is used, namespace is something like 'cdap_userNS'
```
anwar6953: This should probably say "If non-default namespace is used..."

anwar6953:
remainder lgtm

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/hbase/HTableNameConverter.java
chtyim: ok

chtyim:
Build running in http://builds.cask.co/browse/CDAP-DUT1050-1

chtyim:
Build passed. http://builds.cask.co/browse/CDAP-DUT1050-5

PR #1566, opened 2015-03-10
Upgrade/Migrate dataset instances mds.

anwar6953:
https://issues.cask.co/browse/CDAP-1511
http://builds.cask.co/browse/CDAP-DUT1048-6

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDS.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDS.java
@@ -39,7 +39,7 @@ public final class DatasetInstanceMDS extends MetadataStoreDataset {
    * NOTE: even though we don't have to have it now we may want to store different type of data in one table, so
    *       the prefix may help us in future
    */
-  public static final String INSTANCE_PREFIX = "i_";
+  static final String INSTANCE_PREFIX = "i_";
 
   public DatasetInstanceMDS(DatasetSpecification spec, @EmbeddedDataset("") Table table) {
     super(table);
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDS.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDS.java
@@ -39,7 +39,7 @@ public final class DatasetInstanceMDS extends MetadataStoreDataset {
    * NOTE: even though we don't have to have it now we may want to store different type of data in one table, so
    *       the prefix may help us in future
    */
-  private static final String INSTANCE_PREFIX = "i_";
+  public static final String INSTANCE_PREFIX = "i_";
 
   public DatasetInstanceMDS(DatasetSpecification spec, @EmbeddedDataset("") Table table) {
     super(table);
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
@@ -59,16 +59,23 @@ public final class DatasetInstanceMDSUpgrader {
          public UpgradeMdsStores<DatasetInstanceMDS> get() {
            String dsName = Joiner.on(".").join(Constants.SYSTEM_NAMESPACE, DatasetMetaTableUtil.INSTANCE_TABLE_NAME);
            Id.DatasetInstance datasetId = Id.DatasetInstance.from(Constants.DEFAULT_NAMESPACE_ID, dsName);
+           DatasetInstanceMDS oldMds;
+           DatasetInstanceMDS newMds;
            try {
-             DatasetInstanceMDS oldMds =
+             oldMds =
                DatasetsUtil.getOrCreateDataset(dsFramework, datasetId, DatasetInstanceMDS.class.getName(),
                                                DatasetProperties.EMPTY, DatasetDefinition.NO_ARGUMENTS, null);
-             DatasetInstanceMDS newMds = new DatasetMetaTableUtil(dsFramework).getInstanceMetaTable();
-             return new UpgradeMdsStores<DatasetInstanceMDS>(oldMds, newMds);
            } catch (Exception e) {
-             LOG.error("Failed to access table: {}", datasetId, e);
+             LOG.error("Failed to access table of Dataset: {}", datasetId, e);
              throw Throwables.propagate(e);
            }
+           try {
+             newMds = new DatasetMetaTableUtil(dsFramework).getInstanceMetaTable();
+           } catch (Exception e) {
+             LOG.error("Failed to access Datasets instances meta table.");
+             throw Throwables.propagate(e);
+           }
+           return new UpgradeMdsStores<DatasetInstanceMDS>(oldMds, newMds);
          }
        });
   }
@@ -76,18 +83,19 @@ public final class DatasetInstanceMDSUpgrader {
   public void upgrade() throws Exception {
     // Moves dataset instance meta entries into new table (in system namespace)
     // Also updates the spec's name ('cdap.user.foo' -> 'foo')
-    LOG.info("Upgrading dataset instance mds.");
+    LOG.info("Upgrading Dataset instance mds.");
     datasetInstanceMds.execute(new TransactionExecutor.Function<UpgradeMdsStores<DatasetInstanceMDS>, Void>() {
       @Override
       public Void apply(UpgradeMdsStores<DatasetInstanceMDS> ctx) throws Exception {
         MDSKey key = new MDSKey(Bytes.toBytes(DatasetInstanceMDS.INSTANCE_PREFIX));
         DatasetInstanceMDS newMds = ctx.getNewMds();
         List<DatasetSpecification> dsSpecs = ctx.getOldMds().list(key, DatasetSpecification.class);
+        LOG.info("Upgrading {} Dataset Specifications", dsSpecs.size());
         for (DatasetSpecification dsSpec: dsSpecs) {
-          LOG.info("Migrating dataset Spec: {}", dsSpec);
+          LOG.info("Migrating Dataset Spec: {}", dsSpec);
           Id.Namespace namespace = namespaceFromDatasetName(dsSpec.getName());
           DatasetSpecification migratedDsSpec = migrateDatasetSpec(dsSpec);
-          LOG.info("Writing new dataset Spec: {}", migratedDsSpec);
+          LOG.info("Writing new Dataset Spec: {}", migratedDsSpec);
           newMds.write(namespace, migratedDsSpec);
         }
         return null;
@@ -127,7 +135,7 @@ public final class DatasetInstanceMDSUpgrader {
     } else if ("user".equals(namespace)) {
       return Constants.DEFAULT_NAMESPACE_ID;
     } else {
-      throw new IllegalArgumentException(String.format("Expected dataset namespace to be either 'system' or 'user': %s",
+      throw new IllegalArgumentException(String.format("Expected Dataset namespace to be either 'system' or 'user': %s",
                                                        dsId));
     }
   }
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data2.datafabric.dataset.service.mds;
+
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.dataset.DatasetDefinition;
+import co.cask.cdap.api.dataset.DatasetProperties;
+import co.cask.cdap.api.dataset.DatasetSpecification;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.data2.datafabric.dataset.DatasetMetaTableUtil;
+import co.cask.cdap.data2.datafabric.dataset.DatasetsUtil;
+import co.cask.cdap.data2.dataset2.DatasetFramework;
+import co.cask.cdap.data2.dataset2.lib.table.MDSKey;
+import co.cask.cdap.data2.dataset2.tx.Transactional;
+import co.cask.cdap.proto.Id;
+import co.cask.tephra.TransactionExecutor;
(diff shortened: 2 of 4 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -132,4 +179,88 @@ public void create() throws IOException {
...
+  }
+
+  /**
+   * Construct a {@link TableId} from a dataset name
+   * TODO: CDAP-1593 - This is bad and should be removed, since it makes assumptions about the dataset name format.
+   *
+   * @param name the dataset/table name to construct the {@link TableId} from
+   * @return the {@link TableId} object for the specified dataset/table name
+   */
+  private static TableId from(String name) {
```
bdmogal: Is it necessary to use ``TableId``? Its meant for a different purpose right? maybe just return what you want from the dataset name (the non namespaced name)?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/TableId.java
```diff
@@ -51,26 +51,6 @@ public static TableId from(Id.Namespace namespaceId, String tableName) {
...
-  /**
-   * Construct a {@link TableId} from a dataset name
-   * TODO: CDAP-1593 - This is bad and should be removed, since it makes assumptions about the dataset name format.
-   * Will need a bigger change to have DatasetAdmin classes accept a namespace id in some other form to pass to 
-   * #from(tablePrefix, namespace, tableName) as opposed to getting it from the {@link DatasetSpecification#getName}
-   *
-   * @param name the dataset/table name to construct the {@link TableId} from
-   * @return the {@link TableId} object for the specified dataset/table name
-   */
-  public static TableId from(String name) {
```
bdmogal: thanks! :)

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -132,4 +179,88 @@ public void create() throws IOException {
...
+  // Also updates the spec's name ('cdap.user.foo' -> 'foo')
+  private void upgradeDatasetInstanceMDS() throws IOException, DatasetManagementException, InterruptedException,
+    TransactionFailureException {
+    LOG.info("Upgrading dataset instance mds.");
+    datasetInstanceMds.execute(new TransactionExecutor.Function<UpgradeMdsStores<DatasetInstanceMDS>, Void>() {
+      @Override
+      public Void apply(UpgradeMdsStores<DatasetInstanceMDS> ctx) throws Exception {
+        MDSKey key = new MDSKey(Bytes.toBytes(DatasetInstanceMDS.INSTANCE_PREFIX));
+        DatasetInstanceMDS newMds = ctx.getNewMds();
+        Map<MDSKey, DatasetSpecification> dsSpecs = ctx.getOldMds().listKV(key, DatasetSpecification.class);
```
bdmogal: is this correct? should you be reading ``MDSKey`` from the old MDS? or am I missing something?

bdmogal:
just a couple of questions

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
anwar6953: I need two things: table name and namespace.
Thats what TableId provides :)

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
anwar6953: Yea, the list method requires a MDSKey.
So as long as I construct it correctly, it'll work.
`new MDSKey(Bytes.toBytes(DatasetInstanceMDS.INSTANCE_PREFIX));`

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -16,12 +16,22 @@
 
 package co.cask.cdap.data.tools;
```
abaranau: Shouldn't this class/logic be in data-fabric (i.e. in datasets module). It would help to contain internals (we even use "knowledge" about key format and prefixes in here) to the place that manages them. Otherwise it will be extremely hard to maintain the code...

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
bdmogal: My concern again - We're using a *table* abstraction, that is really meant for use in the storage providers for an operation on a *dataset*. It will create a problem in future, if our dataset name has absolutely no relation to table name.  Unlikely, but the assumption that ``dataset name == table name`` has caused problems in the last couple of weeks.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -132,4 +179,88 @@ public void create() throws IOException {
...
+    private UpgradeMdsStores(T oldMds, T newMds) {
+      this.stores = ImmutableList.of(oldMds, newMds);
+    }
+
+    private T getOldMds() {
+      return stores.get(0);
+    }
+
+    private T getNewMds() {
+      return stores.get(1);
```
rohitsinha54: Nice,

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDS.java
```diff
@@ -39,7 +39,7 @@
    * NOTE: even though we don't have to have it now we may want to store different type of data in one table, so
    *       the prefix may help us in future
    */
-  private static final String INSTANCE_PREFIX = "i_";
+  public static final String INSTANCE_PREFIX = "i_";
```
abaranau: ouch

abaranau: see below

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
bdmogal: Oh! missed line 192. ignore.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -132,4 +179,88 @@ public void create() throws IOException {
...
+      public Void apply(UpgradeMdsStores<DatasetInstanceMDS> ctx) throws Exception {
+        MDSKey key = new MDSKey(Bytes.toBytes(DatasetInstanceMDS.INSTANCE_PREFIX));
+        DatasetInstanceMDS newMds = ctx.getNewMds();
+        Map<MDSKey, DatasetSpecification> dsSpecs = ctx.getOldMds().listKV(key, DatasetSpecification.class);
+        for (Map.Entry<MDSKey, DatasetSpecification> dsSpecEntry : dsSpecs.entrySet()) {
+          DatasetSpecification dsSpec = dsSpecEntry.getValue();
+          LOG.info("Migrating dataset Spec: {}", dsSpec);
+          Id.Namespace namespace = namespaceFromDatasetName(dsSpec.getName());
+          DatasetSpecification migratedDsSpec = migrateDatasetSpec(dsSpec);
+          LOG.info("Writing new dataset Spec: {}", migratedDsSpec);
```
rohitsinha54: Can we just print the name rather than whole spec. Its too much info.

anwar6953: Most of the time, its just one line of output, right

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Dataset instances metadata store
```
abaranau: fix

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+                                     @Named("dsFramework") final DatasetFramework dsFramework) {
+
+    this.datasetInstanceMds = Transactional.of(executorFactory,
+       new Supplier<UpgradeMdsStores<DatasetInstanceMDS>>() {
+         @Override
+         public UpgradeMdsStores<DatasetInstanceMDS> get() {
+           String dsName = Joiner.on(".").join(Constants.SYSTEM_NAMESPACE, DatasetMetaTableUtil.INSTANCE_TABLE_NAME);
+           Id.DatasetInstance datasetId = Id.DatasetInstance.from(Constants.DEFAULT_NAMESPACE_ID, dsName);
+           try {
+             DatasetInstanceMDS oldMds =
```
abaranau: so what happens if there's no data to migrate? Should we skip migration and tell user about it - may be smth is wrong? Eg. user might change some configuration that we used to build old table name (dataset.table.prefix), etc.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+               DatasetsUtil.getOrCreateDataset(dsFramework, datasetId, DatasetInstanceMDS.class.getName(),
+                                               DatasetProperties.EMPTY, DatasetDefinition.NO_ARGUMENTS, null);
+             DatasetInstanceMDS newMds = new DatasetMetaTableUtil(dsFramework).getInstanceMetaTable();
+             return new UpgradeMdsStores<DatasetInstanceMDS>(oldMds, newMds);
+           } catch (Exception e) {
+             LOG.error("Failed to access table: {}", datasetId, e);
+             throw Throwables.propagate(e);
```
abaranau: ouch. Can we have some class for UpgradeFailedException or smth - smth "checked", so that upgrade tool itself could act on it appropriately? catch - revert - tell user smth - whatever

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+           String dsName = Joiner.on(".").join(Constants.SYSTEM_NAMESPACE, DatasetMetaTableUtil.INSTANCE_TABLE_NAME);
+           Id.DatasetInstance datasetId = Id.DatasetInstance.from(Constants.DEFAULT_NAMESPACE_ID, dsName);
+           try {
+             DatasetInstanceMDS oldMds =
+               DatasetsUtil.getOrCreateDataset(dsFramework, datasetId, DatasetInstanceMDS.class.getName(),
+                                               DatasetProperties.EMPTY, DatasetDefinition.NO_ARGUMENTS, null);
+             DatasetInstanceMDS newMds = new DatasetMetaTableUtil(dsFramework).getInstanceMetaTable();
+             return new UpgradeMdsStores<DatasetInstanceMDS>(oldMds, newMds);
+           } catch (Exception e) {
+             LOG.error("Failed to access table: {}", datasetId, e);
```
abaranau: Accessing table of "datasetId" is not the only thing you are doing here, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
anwar6953: Even if user changes the table prefix, this code will construct the old table name using that prefix (unless they change the cdap-site, in between versions, which is a problem).
I'll put a log saying how many specs there are to migrate.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
anwar6953: Well, upgrade tool catches all exceptions, and tell users.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -53,13 +54,15 @@
   private final LocationFactory locationFactory;
   private final QueueAdmin queueAdmin;
   private final HBaseTableUtil hBaseTableUtil;
-  private final DatasetFramework namespacedFramework;
+  private final DatasetFramework dsFramework;
   private static final Pattern USER_TABLE_PREFIX = Pattern.compile("^cdap\\.user\\..*");
```
abaranau: ouch. That is in develop already - but looks wrong. "cdap" is actually just a default value of "dataset.table.prefix" cconf property...

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/TableId.java
```diff
@@ -51,26 +50,6 @@ public static TableId from(Id.Namespace namespaceId, String tableName) {
...
-    Preconditions.checkArgument(name != null, "Dataset name should not be null");
-    // Dataset/Table name is expected to be in the format <table-prefix>.<namespace>.<name>
-    String invalidFormatError = String.format("Invalid format for dataset/table name '%s'. " +
-                                                "Expected - <table-prefix>.<namespace>.<dataset/table-name>", name);
-    String [] parts = name.split("\\.", 3);
-    Preconditions.checkArgument(parts.length == 3, invalidFormatError);
```
abaranau: so this will throw illegal argument exception and that one will be surfaced up to the user?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
abaranau: > unless they change the cdap-site, in between versions, which is a problem

cdap-data-fabric/src/main/java/co/cask/cdap/data2/util/TableId.java
anwar6953: Yea, it will be caught by the upgrade tool and logged.
Alternative is: if one part of upgrade tool fails, continue to other parts, but I think the current design is to stop if something is wrong.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
abaranau: > I'll put a log saying how many specs there are to migrate.

abaranau:
one comment left

abaranau:
@anwar6953  is fast

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
anwar6953: Opened a JIRA for this and another thing I noticed while making these updates.
https://issues.cask.co/browse/CDAP-1761

abaranau: add todo please. As long as it is fixed in 2.8, it is fine. Mark it as blocker please

anwar6953: Marked as blocker.

abaranau:
good to go

PR #1568, opened 2015-03-10
Feature/2.8.0 namespace doc changes

awholegunch:


cdap-docs/developers-manual/source/building-blocks/namespaces.rst
```diff
@@ -37,12 +38,16 @@ Namespace IDs are composed from a limited set of characters; they are restricted
...
 on the length of a namespace ID nor on the number of namespaces.
 
-The namespace IDs ``default`` and ``system`` are reserved. The ``default``
+The namespace IDs ``cdap``, ``default``, and ``system`` are reserved. The ``default``
 namespace, however, can be used by anyone, though like all reserved namespaces, it cannot
 be deleted.
 
-Namespaces: Independent and Non-hierarchal
-------------------------------------------
+Once created, these namespace properties (ID, display name, description) cannot be altered.
```
bdmogal: Forgot to point this out - you can change the display name and description now. It was recently pushed. We should probably just mention that the id cannot be modified once created.

bdmogal: Begs another point though - we need to add documentation for a new API. A namespace can have properties now. I guess we could do that in a later PR.

bdmogal:
Is this staged somewhere?

awholegunch:
No, it's not. I will though.

awholegunch:
Staged at http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-2.8.0_namespace_doc_changes/en/index.html
In particular:
- [Building Blocks](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-2.8.0_namespace_doc_changes/en/developers-manual/building-blocks/namespaces.html)
- [RESTful API](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-2.8.0_namespace_doc_changes/en/reference-manual/http-restful-api/http-restful-api-v3/namespace.html)

I am wondering if I have the order of these correct: look at the order of the Building Blocks and then compare it to the order of the RESTful API. Should they match, or be closer? Perhaps namespaces should be moved up higher in the Building Blocks.

cdap-docs/developers-manual/source/building-blocks/namespaces.rst
```diff
@@ -28,8 +28,9 @@ namespaces include:
 The first version of namespaces was introduced in :ref:`CDAP v2.8.0 <release-notes>`, and
```
bdmogal: Not part of this PR, but in lines 24-26 above, should we call it 'Partitioning a single CDAP cluster'

bdmogal: Also another use case would be to support multiple sub-organizations within an/your organization

cdap-docs/developers-manual/source/building-blocks/namespaces.rst
```diff
@@ -54,28 +56,23 @@ Similarly, moving applications or data from one namespace to another is not poss
 Quota management based on namespaces is also not possible.
```
bdmogal: ...in this release, but could be a feature in a future release?

cdap-docs/developers-manual/source/building-blocks/namespaces.rst
```diff
@@ -28,8 +28,9 @@ namespaces include:
 The first version of namespaces was introduced in :ref:`CDAP v2.8.0 <release-notes>`, and
 is part of the :ref:`HTTP RESTful API v3 <http-restful-api-v3>`.
 
-Namespace Rules
-===============
+
+Namespace Components
+====================
 
 A Namespace consists of a namespace identifier (the namespace ID), a display name, and a description.
```
bdmogal: I think this should be 'A namespace has a ...'

Primarily because when 'consists' could be misleading right - it could mean 'containership' in which case a namespace would consist of apps, data, etc.

bdmogal: Thoughts?

bdmogal:
@awholegunch just a couple of minor comments. I think we may have to do one more pass at this, especially around
1. There's an effort to standardize all CDAP entities (including namespaces) to have only a *name* and a *description* (display name will likely go away).
2. We will have to add documentation about the namespace properties.

Ofcourse, both of the above can be in later PRs.

awholegunch:
Updated. @bdmogal, please take another pass, thanks!

bdmogal:
:+1:

PR #1571, opened 2015-03-10
Exposing transaction manager APIs to truncate invalid list through Router

poornachandra:
https://issues.cask.co/browse/CDAP-1540

Also updating tephra version to 0.3.5-SNAPSHOT for now. Will replace the version when tephra release is done.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
@@ -43,6 +43,7 @@ import co.cask.http.BodyConsumer;
 import co.cask.http.HttpResponder;
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonSyntaxException;
 import com.google.inject.Inject;
 import org.jboss.netty.handler.codec.http.HttpRequest;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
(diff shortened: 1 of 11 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
```diff
@@ -172,9 +172,27 @@ public void invalidateTx(HttpRequest request, HttpResponder responder,
                                         responder, txId);
   }
 
-  /**
-   * Reset the state of the transaction manager.
-   */
+  @Path("/transactions/invalid/remove/until")
```
anew: are these needed in the v2/ APIs?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/TransactionHttpHandler.java
```diff
@@ -110,6 +118,70 @@ public void invalidateTx(HttpRequest request, HttpResponder responder,
...
+  
+  @Path("/transactions/invalid/remove/until")
+  @POST
+  public void truncateInvalidTxBefore(HttpRequest request, HttpResponder responder) {
+    try {
+      Map<String, Long> body;
+      try {
+        body = parseBody(request, STRING_LONG_MAP_TYPE);
+      } catch (IOException e) {
+        responder.sendStatus(HttpResponseStatus.BAD_REQUEST);
```
anew: How about adding a message to the response?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -144,11 +144,15 @@ public AbstractAppFabricHttpHandler(Authenticator authenticator) {
   }
 
   protected int getInstances(HttpRequest request) throws IOException, NumberFormatException {
-    return parseBody(request, Instances.class).getInstances();
+    Instances instances = parseBody(request, Instances.class);
```
anew: this looks like an unrelated change. I have noticed this warning myself and was tempted to fix it also...
If you want to fix this, you should throw an exception that can be recognized by all callers, so that they can return BAD_REQUEST. And make that change in all callers, too :)

anew:
I have only minor comments.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
poornachandra: Instances can be null when `request.getContent()` is not readable. It is may not be due to user error. We can catch `NumberFormatException` or `JsonSyntaxException` in callers and send BAD_REQUEST. I will add that too.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
poornachandra: We need these for 2.6.2.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/TransactionHttpHandler.java
poornachandra: Good point.

poornachandra:
@anew Addressed comments, please take a look.

anew:
LGTM

poornachandra:
Build is green http://builds.cask.co/browse/CDAP-DUT1056-8

PR #1572, opened 2015-03-10
CDAP 1512 Upgrade DatasetTypeMDS table

rohitsinha54:
Upgrades the datasets.type meta table

Issue : https://issues.cask.co/browse/CDAP-1512
Build : http://builds.cask.co/browse/CDAP-DUT1059-1


cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetTypeMDSUpgrader.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/**
+ * Upgrades {@link DatasetTypeMDS}
+ */
+public final class DatasetTypeMDSUpgrader {
+  private static final Logger LOG = LoggerFactory.getLogger(DatasetTypeMDSUpgrader.class);
+  // lists of datasets type modules which existed earlier but does not anymore
```
anwar6953: I think you should mention in the comment at least - which version did they last exist in (which version are they first removed from).

albertshau:
We should drop the old table once we're done migrating it, and make sure upgrade is just a no-op if there is no old table (in case a later stage of the full upgrade fails)

anwar6953:
Why not keep the old table around? (next time upgrade runs, it can just rewrite the rows).

albertshau:
upgrade today is a commitment because there is no rollback. Don't we destructively upgrade in other places? Why would you want old data lying around after upgrade finishes?

anwar6953:
Hmm... I'm just concerned with the possibility of a later stage failing. (so an upgrade not being complete, but the old data also missing).

Does it make sense then, to do the destructiveness at the end of the all the upgrade steps (sort of like a 'commit' phase).

albertshau:
That is why upgrade must be idempotent. You should be able to run it again to complete it.  We do not want to keep copies of all data. 

You could do this at the end if you want, but not all upgrades can follow that pattern. For example, files get moved, they do not get copied.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetTypeMDSUpgrader.java
```diff
@@ -0,0 +1,212 @@
...
+   * @throws InterruptedException
+   * @throws IOException
+   */
+  public void upgrade() throws Exception {
+    final MDSKey dsModulePrefix = new MDSKey(Bytes.toBytes(DatasetTypeMDS.MODULES_PREFIX));
+    try {
+      datasetTypeMDS.execute(new TransactionExecutor.Function<UpgradeMDSStores<DatasetTypeMDS>, Void>() {
+        @Override
+        public Void apply(UpgradeMDSStores<DatasetTypeMDS> ctx) throws Exception {
+          Map<MDSKey, DatasetModuleMeta> mdsKeyDatasetModuleMetaMap = ctx.getOldMds().listKV(dsModulePrefix,
```
anwar6953: `listKv` isn't necessary, if you're not using the key.
Simple use `list`

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetTypeMDSUpgrader.java
```diff
@@ -0,0 +1,212 @@
...
+  private Location updateUserDatasetModuleJarLocation(Location location, String datasetClassname,
+                                                      String namespace) throws IOException {
+    String jarFilename = location.getName();
+    Location parentLocation = Locations.getParent(location);  // strip jarFilename
+    parentLocation = Locations.getParent(parentLocation); // strip account_placeholder
+    String archive = parentLocation != null ? parentLocation.getName() : null;
+    parentLocation = Locations.getParent(parentLocation); // strip archive
+    String datasets = parentLocation != null ? parentLocation.getName() : null;
+
+    return locationFactory.create(namespace).append(datasets).append(datasetClassname).append(archive)
```
anwar6953: what happens if you're appending `null`?

albertshau:
:+1:

rohitsinha54:
Build running after develop was fixed here: http://builds.cask.co/browse/CDAP-DUT1059-4

Should be merged when build passes.

rohitsinha54:
Build http://builds.cask.co/browse/CDAP-DUT1059-6 :green_heart: 
Merging.

PR #1575, opened 2015-03-11
[CDAP-1515] Added queue config upgrade

bdmogal:
Adds the 2.8 ``QueueName`` to '``cdap.system.queue.config`` table.

Requires more rigorous testing. However, seeking feedback on the approach.

Note: This has to delete old rows since we share the same table in 2.8.

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data.tools;
+
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.queue.QueueName;
+import co.cask.cdap.data2.transaction.queue.QueueConstants;
+import co.cask.cdap.data2.transaction.queue.QueueEntryRow;
+import co.cask.cdap.data2.util.TableId;
+import co.cask.cdap.data2.util.hbase.HBaseTableUtil;
+import co.cask.cdap.proto.Id;
+import com.google.common.base.Joiner;
+import com.google.common.base.Throwables;
+import com.google.common.collect.Lists;
+import com.google.inject.Inject;
+import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.client.Delete;
+import org.apache.hadoop.hbase.client.HTable;
+import org.apache.hadoop.hbase.client.Mutation;
+import org.apache.hadoop.hbase.client.Put;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.twill.filesystem.LocationFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.NavigableMap;
+import javax.annotation.Nullable;
+
+/**
+ * Upgrades queue.config table
+ */
+public class QueueConfigUpgrader extends AbstractUpgrader {
+  private static final Logger LOG = LoggerFactory.getLogger(QueueConfigUpgrader.class);
+  private final HBaseTableUtil tableUtil;
+
+  @Inject
+  public QueueConfigUpgrader(LocationFactory locationFactory, HBaseTableUtil tableUtil) {
+    super(locationFactory);
+    this.tableUtil = tableUtil;
+  }
+
+  @Override
+  void upgrade() throws Exception {
+    String dsName = Joiner.on(".").join(Constants.SYSTEM_NAMESPACE, QueueConstants.QUEUE_CONFIG_TABLE_NAME);
+    Id.DatasetInstance datasetId = Id.DatasetInstance.from(Constants.DEFAULT_NAMESPACE_ID, dsName);
+    HTable queueConfigTable = tableUtil.createHTable(HBaseConfiguration.create(),
+                                                     TableId.from(datasetId.getNamespaceId(), dsName));
+    LOG.info("Starting upgrade for queue config table {}", Bytes.toString(queueConfigTable.getTableName()));
+    try {
+      Scan scan = new Scan();
+      scan.setTimeRange(0, HConstants.LATEST_TIMESTAMP);
+      scan.addFamily(QueueEntryRow.COLUMN_FAMILY);
+      scan.setMaxVersions(1); // we only need to see one version of each row
+      List<Mutation> mutations = Lists.newArrayList();
+      Result result;
+      ResultScanner resultScanner = queueConfigTable.getScanner(scan);
+      try {
+        while ((result = resultScanner.next()) != null) {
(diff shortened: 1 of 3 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
```diff
@@ -0,0 +1,140 @@
...
+      parts[i] = iter.next();
+    }
+
+    if (parts.length == 4) {
+      LOG.info("Found old queue config {}. Migrating.", oldQueueConfigRowKey);
+      return QueueName.fromFlowlet(Constants.DEFAULT_NAMESPACE, parts[0], parts[1], parts[2], parts[3]);
+    } else if (parts.length == 5) {
+      LOG.info("Queue config for {} seems to already have been migrated. Skipping.", oldQueueConfigRowKey);
+    } else {
+      LOG.info("Unknown format for queue config {}. Skipping.", oldQueueConfigRowKey);
```
anwar6953: this should probably be warn/error, rather than info.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
```diff
@@ -0,0 +1,140 @@
...
+  public QueueConfigUpgrader(LocationFactory locationFactory, HBaseTableUtil tableUtil) {
+    super(locationFactory);
+    this.tableUtil = tableUtil;
+  }
+
+  @Override
+  void upgrade() throws Exception {
+    String dsName = Joiner.on(".").join(Constants.SYSTEM_NAMESPACE, QueueConstants.QUEUE_CONFIG_TABLE_NAME);
+    Id.DatasetInstance datasetId = Id.DatasetInstance.from(Constants.DEFAULT_NAMESPACE_ID, dsName);
+    HTable queueConfigTable =  tableUtil.createHTable(HBaseConfiguration.create(),
```
anwar6953: extra space after `=`

bdmogal:
@anwar6953 addressed comments

chtyim:
This is going to change again as the queue config becomes a Dataset.

bdmogal:
@chtyim addressed your offline comment about reusing ``QueueName`` for parsing. Does this look ok for merge? I can fix it again when Queue config table becomes a dataset

anwar6953:
Perhaps we should merge this, and make updates to the destination of the data migration once @chtyim has his changes merged?

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+            LOG.info("Marking old key {} for deletion", rowKey);
+            deletes.add(new Delete(row));
+          }
+          LOG.info("Finished processing queue config for {}", rowKey);
+        }
+      } finally {
+        resultScanner.close();
+      }
+
+      queueConfigTable.put(puts);
```
chtyim: It's better to use `HTable.batch(List)`. You can represents both put and delete as `Mutation`.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+    Id.DatasetInstance datasetId = Id.DatasetInstance.from(Constants.DEFAULT_NAMESPACE_ID, dsName);
+    HTable queueConfigTable = tableUtil.createHTable(HBaseConfiguration.create(),
+                                                     TableId.from(datasetId.getNamespaceId(), dsName));
+    LOG.info("Starting upgrade for queue config table {}", Bytes.toString(queueConfigTable.getTableName()));
+    try {
+      Scan scan = new Scan();
+      scan.setTimeRange(0, HConstants.LATEST_TIMESTAMP);
+      scan.addFamily(QueueEntryRow.COLUMN_FAMILY);
+      scan.setMaxVersions(1); // we only need to see one version of each row
+      ResultScanner resultScanner = queueConfigTable.getScanner(scan);
```
chtyim: As a convention, the try-finally should start right after the resource you need to free-up (meaning the `try` should be here, not after new array list).

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+      List<Put> puts = Lists.newArrayList();
+      List<Delete> deletes = Lists.newArrayList();
+      Result result;
+      try {
+        while ((result = resultScanner.next()) != null) {
+          byte[] row = result.getRow();
+          String rowKey = Bytes.toString(row);
+          LOG.info("Processing queue config for  {}", rowKey);
```
chtyim: You decided to print every row key? Shouldn't it be debug?

chtyim: Also, should use `toStringBinary`.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+          NavigableMap<byte[], byte[]> columnsMap = result.getFamilyMap(QueueEntryRow.COLUMN_FAMILY);
+          QueueName queueName = fromRowKey(row);
+          if (queueName != null) {
+            Put put = new Put(queueName.toBytes());
+            LOG.info("Upgraded queue name: {}", queueName);
+            for (NavigableMap.Entry<byte[], byte[]> entry : columnsMap.entrySet()) {
+              LOG.info("Adding entry {} -> {} for upgrade",
```
chtyim: should log as debug as well.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
chtyim: I mean both the puts and deletes can be done in one `batch` call.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
anwar6953: I think this already does toStringBinary, effectively.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+
+    if (queueName.getNumComponents() == 4) {
+      LOG.info("Found old queue config {}. Upgrading.", queueName.toString());
+      return QueueName.fromFlowlet(Constants.DEFAULT_NAMESPACE,
+                                   queueName.getFirstComponent(),
+                                   queueName.getSecondComponent(),
+                                   queueName.getThirdComponent(),
+                                   queueName.getFourthComponent());
+    } else if (queueName.getNumComponents() == 5) {
+      LOG.info("Queue config for {} seems to already have been upgraded. Skipping.", queueName.toString());
```
chtyim: When logging with `{}`, no need to have `.toString`. Actually it's better not to have `.toString()` so that it only get invoked when the log level requires the log output.

chtyim:
Couple comments.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
bdmogal: Ok. Will move this line right above the try

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
bdmogal: Fixed.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
bdmogal: Fixed. The original intention was to keep the logs verbose until we finish testing, but I think we should rather change the log level to debug while testing.

cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
bdmogal: Fixed.

bdmogal:
@chtyim Thanks! Addressed comments. Please review.

chtyim:
LGTM

PR #1579, opened 2015-03-11
CDAP-1763 fileset upgrade

rohitsinha54:
Upgraded Filesets base paths:
Issue: https://issues.cask.co/browse/CDAP-1763

This PR was reviewed here: #1576

Opening this against develop and waiting for final approval.

Build : https://builds.cask.co/browse/CDAP-DUT1067-1

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
@@ -17,9 +17,11 @@
 package co.cask.cdap.data2.datafabric.dataset.service.mds;
 
 import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.dataset.Dataset;
 import co.cask.cdap.api.dataset.DatasetDefinition;
 import co.cask.cdap.api.dataset.DatasetProperties;
 import co.cask.cdap.api.dataset.DatasetSpecification;
+import co.cask.cdap.api.dataset.lib.FileSet;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.data2.datafabric.dataset.DatasetMetaTableUtil;
 import co.cask.cdap.data2.datafabric.dataset.DatasetsUtil;
@@ -29,15 +31,20 @@ import co.cask.cdap.data2.dataset2.tx.Transactional;
 import co.cask.cdap.proto.Id;
 import co.cask.tephra.TransactionExecutor;
 import co.cask.tephra.TransactionExecutorFactory;
+import co.cask.tephra.TransactionFailureException;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.google.inject.Inject;
 import com.google.inject.name.Named;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -102,13 +109,66 @@ public final class DatasetInstanceMDSUpgrader {
   }
 
   /**
+   * Gets the {@link DatasetSpecification} of all the {@link FileSet} from the {@link DatasetInstanceMDS} table
+   *
+   * @return A list of {@link DatasetSpecification} of all the {@link FileSet}
+   * @throws TransactionFailureException
+   * @throws InterruptedException
+   * @throws IOException
+   */
+  public List<DatasetSpecification> getFileSetsSpecs() throws TransactionFailureException,
+    InterruptedException, IOException {
+    return datasetInstanceMds.execute(new TransactionExecutor.Function<UpgradeMDSStores<DatasetInstanceMDS>,
+      List<DatasetSpecification>>() {
+      @Override
+      public List<DatasetSpecification> apply(UpgradeMDSStores<DatasetInstanceMDS> ctx) throws Exception {
+        MDSKey key = new MDSKey(Bytes.toBytes(DatasetInstanceMDS.INSTANCE_PREFIX));
+        List<DatasetSpecification> dsSpecs = ctx.getNewMds().list(key, DatasetSpecification.class);
+        List<DatasetSpecification> fileSetSpecs = Lists.newArrayList();
+        for (DatasetSpecification dsSpec : dsSpecs) {
+          fileSetSpecs.addAll(findFileSetsInEmbeddedDS(dsSpec));
+        }
+        return fileSetSpecs;
+      }
+    });
+  }
+
+  /**
+   * Recursively search the embedded datasets of a {@link Dataset} which is not a {@link FileSet} for {@link FileSet}
+   *
+   * @param dsSpec the {@link DatasetSpecification} of a {@link Dataset} which should be searched for {@link FileSet}
+   * @return A list of {@link DatasetSpecification} of {@link FileSet}
+   */
+  private List<DatasetSpecification> findFileSetsInEmbeddedDS(DatasetSpecification dsSpec) {
+    if (isFileSet(dsSpec)) {
+      return ImmutableList.of(dsSpec);
+    }
+    ArrayList<DatasetSpecification> dsSpecs = Lists.newArrayList();
+    for (DatasetSpecification embeddedDsSpec : dsSpec.getSpecifications().values()) {
+      dsSpecs.addAll(findFileSetsInEmbeddedDS(embeddedDsSpec));
+    }
+    return dsSpecs;
+  }
+
+  /**
+   * Checks if the the given dataset spec is of a {@link FileSet}
+   *
+   * @param dsSpec the {@link DatasetSpecification} of the dataset
+   * @return a boolean which is true if its a {@link FileSet} else false
+   */
+  public boolean isFileSet(DatasetSpecification dsSpec) {
+    String dsType = dsSpec.getType();
+    return (FileSet.class.getName().equals(dsType) || "fileSet".equals(dsType));
+  }
+
+  /**
    * Construct a {@link Id.DatasetInstance} from a pre-2.8.0 CDAP Dataset name
    *
    * @param datasetName the dataset/table name to construct the {@link Id.DatasetInstance} from
    * @return the {@link Id.DatasetInstance} object for the specified dataset/table name
    */
   private static Id.DatasetInstance from(String datasetName) {
-    Preconditions.checkArgument(datasetName != null, "Dataset name should not be null");
+    Preconditions.checkNotNull(datasetName, "Dataset name should not be null");
     // Dataset/Table name is expected to be in the format <table-prefix>.<namespace>.<name>
     String invalidFormatError = String.format("Invalid format for dataset '%s'. " +
                                                 "Expected - <table-prefix>.<namespace>.<dataset-name>", datasetName);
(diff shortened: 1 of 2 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
```diff
@@ -102,6 +106,43 @@ public Void apply(UpgradeMDSStores<DatasetInstanceMDS> ctx) throws Exception {
...
+                                                              InterruptedException, IOException {
+    return datasetInstanceMds.execute(new TransactionExecutor.Function<UpgradeMDSStores<DatasetInstanceMDS>,
+                                                                                        List<DatasetSpecification>>() {
+      @Override
+      public List<DatasetSpecification> apply(UpgradeMDSStores<DatasetInstanceMDS> ctx) throws Exception {
+        MDSKey key = new MDSKey.Builder().add(DatasetInstanceMDS.INSTANCE_PREFIX).build();
+        List<DatasetSpecification> dsSpecs = ctx.getNewMds().list(key, DatasetSpecification.class);
+        List<DatasetSpecification> fileSetSpecs = Lists.newArrayList();
+        for (DatasetSpecification dsSpec : dsSpecs) {
+          if (isFileSet(dsSpec)) {
```
anew: this finds the FileSets that are registered as instances. What about FileSets that are embedded in other datasets? You would have to recurse through the embedded specs of each dataset to find them.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -96,6 +100,31 @@ public void upgrade() throws Exception {
...
+
+  /**
+   * Upgrades the {@link FileSet} and also its embedded filesets if any by moving the base path under
+   * namespaced directory
+   *
+   * @param dsSpec the {@link DatasetSpecification} of the {@link FileSet} to be upgraded
+   * @throws IOException
+   */
+  private void upgradeFileSet(DatasetSpecification dsSpec) throws IOException {
+    for (DatasetSpecification embeddedDsSpec : dsSpec.getSpecifications().values()) {
```
anew: this only gets called for FileSets. They never have embedded datasets.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -96,6 +100,31 @@ public void upgrade() throws Exception {
...
+      upgradeFileSet(embeddedDsSpec);
```
anew: And you cannot assume that every embedded dataset is a FileSet. Does the recursive call fail if you call this on a dataset that is not a FileSet?

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -96,6 +100,31 @@ public void upgrade() throws Exception {
...
+    }
+    String basePath = FileSetProperties.getBasePath(dsSpec.getProperties());
```
anew: Better check that the basePath is not null.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -96,6 +100,31 @@ public void upgrade() throws Exception {
...
+    Location oldLocation = locationFactory.create(basePath);
+    Location newlocation = locationFactory.create(Constants.DEFAULT_NAMESPACE)
+      .append(cConf.get(Constants.Dataset.DATA_DIR)).append(basePath);
+    LOG.info("Upgrading base path for dataset {} from {} to {}", dsSpec.getName(), oldLocation, newlocation);
+    renameLocation(oldLocation, newlocation);
```
anew: I am not sure where this method went. In the last PR it was there and there was quite some discussion around it. We need to make sure that:
- calling this for a dataset that has already been upgraded does not fail
- calling this for multiple datasets that have the same base path does not cause failure

anew:
@rohitsinha54 this still need some fixing. Please see my comments.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
rohitsinha54: Yes. In the same way we upgrade their instance meta.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
rohitsinha54: Please see below.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
rohitsinha54: I was not aware of this case. Thanks. Will fix it.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
rohitsinha54: Since I moved this class that method was not needed anymore. Its in upgrade tool and it can use the common shared method which I have already changed according to your suggestions. 
Here is the PR which got merged: https://github.com/caskdata/cdap/pull/1580

rohitsinha54: No we use rename here: 
https://github.com/caskdata/cdap/blob/fa44c400e335b13e00a495f1f7ee91af3c3e091c/cdap-master/src/main/java/co/cask/cdap/data/tools/AbstractUpgrader.java#L57

So, if the basePath got moved initially then for other datasets it will say that it might have been moved and will not throw does FileNotFoundException as the renameTo does as we skip rename if it doesn't. So, I think that should be fine.


But I haven't tested the FileSet upgrade for FileSets which have same basepath. That is the only thing which I have on my plate for today. I have tested it with FileSet example.

rohitsinha54:
@anew : Addressed comment. Please have a look. Thanks

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
anew: Consider where I have a dataset type WithFiles that itself is not a FileSet but has an embedded FileSet. For that dataset, isFileSet will return false, and so you will exclude it from the list of specs returned by this method. That means the embedded FileSet will never be inspected. You need to fix this.

rohitsinha54: Oh I didn't know that was possible too. Okay doing it now.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/mds/DatasetInstanceMDSUpgrader.java
```diff
@@ -102,6 +106,43 @@ public Void apply(UpgradeMDSStores<DatasetInstanceMDS> ctx) throws Exception {
...
+  public List<DatasetSpecification> getFileSetsSpecs() throws TransactionFailureException,
+                                                              InterruptedException, IOException {
+    return datasetInstanceMds.execute(new TransactionExecutor.Function<UpgradeMDSStores<DatasetInstanceMDS>,
+                                                                                        List<DatasetSpecification>>() {
+      @Override
+      public List<DatasetSpecification> apply(UpgradeMDSStores<DatasetInstanceMDS> ctx) throws Exception {
+        MDSKey key = new MDSKey(Bytes.toBytes(DatasetInstanceMDS.INSTANCE_PREFIX));
+        List<DatasetSpecification> dsSpecs = ctx.getNewMds().list(key, DatasetSpecification.class);
+        List<DatasetSpecification> fileSetSpecs = Lists.newArrayList();
+        for (DatasetSpecification dsSpec : dsSpecs) {
```
anew: In theory, an embedded dataset can embed another - it is recursive. I don't think it is common, but you shuold do this recursively, too.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
anew: Here you do not need to iterate over the embedded specs.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -96,6 +100,33 @@ public void upgrade() throws Exception {
...
+   * @param dsSpec the {@link DatasetSpecification} of the {@link FileSet} to be upgraded
+   * @throws IOException
+   */
+  private void upgradeFileSet(DatasetSpecification dsSpec) throws IOException {
+    for (DatasetSpecification embeddedDsSpec : dsSpec.getSpecifications().values()) {
+      if (datasetInstanceMDSUpgrader.isFileSet(embeddedDsSpec)) {
+        upgradeFileSet(embeddedDsSpec);
+      }
+    }
+    String basePath = FileSetProperties.getBasePath(dsSpec.getProperties());
```
anew: Make sure that basePath is not null

anew:
Please address my comments. I feel like I am repeating the same comments.

rohitsinha54:
@anew : Sorry about that. I thought only a dataset which is fileset can have embedded fileset. Didn't know any dataset can have embedded filesets in it. Doing it now.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
rohitsinha54: Yes it was before I knew that a dataset which is not a fileset can have embedded filesets

rohitsinha54:
@anew : Addressed all the comments and also went through all previous comments to make sure I addressed everything as a double check :)
My initial understanding for where all can I find filesets was wrong which cause a bit of problem. Thanks for making it clear to me. I have fixed it. Please have a look.
Thanks.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -96,6 +101,28 @@ public void upgrade() throws Exception {
...
+  /**
+   * Upgrades the {@link FileSet} and also its embedded filesets if any by moving the base path under
+   * namespaced directory
+   *
+   * @param dsSpec the {@link DatasetSpecification} of the {@link FileSet} to be upgraded
+   * @throws IOException
+   */
+  private void upgradeFileSet(DatasetSpecification dsSpec) throws IOException {
+    String basePath = FileSetProperties.getBasePath(dsSpec.getProperties());
+    Preconditions.checkNotNull(basePath, "The basepath for fileset {} is null", dsSpec.getName());
```
anew: What I meant was that the upgrade tool would not fail. The Precondition only makes it throw the NPE one line earlier. Can you just ignore the fileset if it does not have a base path?

anew:
Only one comment and you are good to merge

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
rohitsinha54: Will do. But quick question. Is it possible to have a fileset which does not have a basepath ?

rohitsinha54:
Build :green_heart: http://builds.cask.co/browse/CDAP-DUT1067-6
Merging.

PR #1581, opened 2015-03-11
Feature/cdap 1644 app fabric api cleanup

anew:
This removes some old REST APIs that the new UI does not need any longer. 
Also cleans up unused code in the HTTP handlers.
Also attempts to streamline our approach to naming: All objects now have a name, and the (always identical) id is deprecated and will be removed in the future. Only exception is NamespaceMeta, which I will address separately. 

The builds have been failing, unfortunately, for the last day, due to instabilities in develop. I am sending this for review anyway, because I am certain that the failures are unrelated to my changes.


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
```diff
@@ -550,9 +546,32 @@ public void changeFlowletStreamConnection(HttpRequest request, HttpResponder res
                                             @PathParam("flow-id") String flowId,
                                             @PathParam("flowlet-id") String flowletId,
                                             @PathParam("stream-id") String streamId) throws IOException {
-    programLifecycleHttpHandler.changeFlowletStreamConnection(RESTMigrationUtils.rewriteV2RequestToV3(request),
-                                                              responder, Constants.DEFAULT_NAMESPACE,
-                                                              appId, flowId, flowletId, streamId);
```
anew: this code was moved from ProgramLifecycleHttpHandler, with the default namespace instead of receiving a path parameter. Other than that no change.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
```diff
@@ -910,8 +925,17 @@ public void getProceduresByApp(HttpRequest request, HttpResponder responder,
...
                                 @PathParam("app-id") String appId) {
-    programLifecycleHttpHandler.getMapreduceByApp(RESTMigrationUtils.rewriteV2RequestToV3(request), responder,
-                                                 Constants.DEFAULT_NAMESPACE, appId);
+    getProgramsByApp(responder, Constants.DEFAULT_NAMESPACE, appId, ProgramType.MAPREDUCE.getCategoryName());
+  }
+
+  /**
+   * Returns a list of spark jobs associated with account & application.
+   */
+  @GET
```
anew: moved this here from ServiceHandler

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
```diff
@@ -1090,4 +1110,16 @@ public void deleteDatasets(HttpRequest request, HttpResponder responder) {
                            String.format(UserMessages.getMessage(UserErrors.DATASETS_DELETE_FAIL), e.getMessage()));
     }
   }
+
+  protected void getProgramsByApp(HttpResponder responder, String namespaceId, String appId, String programCategory) {
```
anew: Moved here from ProgramLifecycleHandler

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -695,85 +680,6 @@ public void getAllWorkers(HttpRequest request, HttpResponder responder,
     programList(responder, namespaceId, ProgramType.WORKER, null, store);
   }
 
-  /********************** Programs by app APIs **********************************************************/
```
anew: these APIs are removed from v3/ and moved to the AppFabricHttpHandler

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1422,22 +1289,6 @@ protected ProgramStatus getProgramStatus(Id.Program id, ProgramType type) {
     return null;
   }
 
-  /**
```
anew: method that was not used anywhere

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -280,6 +279,15 @@ protected final void programList(HttpResponder responder, String namespaceId, Pr
     }
   }
 
+  @Nullable
```
anew: moved here from ProgramLifecycleHandler, because it is needed by the app fabric handler, too.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -680,39 +678,6 @@ public void testFlows() throws Exception {
     Assert.assertEquals(200, deleteQueues(TEST_NAMESPACE1, WORDCOUNT_APP_NAME, WORDCOUNT_FLOW_NAME));
   }
 
-  @Test
```
anew: test removed because API was removed from v3/. A similar test exists for the v2/ version of this.

cdap-proto/src/main/java/co/cask/cdap/proto/ProgramDetail.java
```diff
@@ -1,44 +0,0 @@
-/*
```
anew: removed this - using ProgramRecord everywhere now

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -620,16 +628,7 @@ private static boolean usesStream(FlowSpecification flowSpec, String stream) {
 
   /* -----------------  helpers to return Json consistently -------------- */
 
-  protected static final DatasetRecord makeDataSetRecord(String name, String classname) {
-    return new DatasetRecord("Dataset", name, name, classname);
-  }
-
-  protected static final StreamRecord makeStreamRecord(String name, StreamSpecification specification) {
-    return new StreamRecord("Stream", name, name, GSON.toJson(specification));
+  protected static StreamRecord makeStreamRecord(String name, StreamSpecification specification) {
```
albertshau: should we just have the constructor take a StreamSpecification then? or is the string version used somewhere else

albertshau: actually, StreamSpecification is just a name... so the api returns name 3 times? seems wrong though its existing behavior

cdap-client/src/main/java/co/cask/cdap/client/ApplicationClient.java
```diff
@@ -242,45 +241,65 @@ public void deploy(File jarFile) throws IOException, UnauthorizedException {
    * @throws IOException if a network error occurred
    * @throws UnauthorizedException if the request is not authorized successfully in the gateway server
    */
-  public List<ProgramRecord> listPrograms(String appId, ProgramType programType)
+  public List<ProgramRecord> listPrograms(String appId, final ProgramType programType)
```
albertshau: this doesn't need to be final right?

albertshau:
question about the stream api, otherwise lgtm

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
anew: It is strange. There is one place where this is called with specification=null, so I can't use the spec to get the name. I tried not to change existing logic, that's why I kept it this way.

cdap-client/src/main/java/co/cask/cdap/client/ApplicationClient.java
anew: no it does not. Not sure how that got here...

PR #1583, opened 2015-03-11
(CDAP-1204) Adds ShardedHBaseQueueStrategy

chtyim:
- Also includes refactoring on HBaseQueueProducer for future changes that includes ConsumerConfigGroup information


chtyim:
This PR just add the new classes. Not hookup with the queue system yet. Unit test (QueueTest) will be provided in future PR after the wiring up is completed.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueProducer.java
```diff
@@ -29,37 +34,36 @@
...
     super(queueMetrics, queueName);
     this.queueStrategy = queueStrategy;
-    // Base row key = queue_name + writePointer + counter
+    // Make sure only one config per consumer group
+    this.consumerGroupConfigs = ImmutableList.copyOf(
+      Iterables.filter(consumerGroupConfigs, new Predicate<ConsumerGroupConfig>() {
+        private final Set<Long> seenGroups = Sets.newHashSet();
+          @Override
+          public boolean apply(ConsumerGroupConfig config) {
+            return seenGroups.add(config.getGroupId());
```
abaranau: would it indicated an error if there's more than one?

abaranau: if not, I'd expect at least a note - why this may happen and why to keep one

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * An implementation of {@link HBaseQueueStrategy} with sharded keys.
+ *
+ * The row key has structure of:
+ *
+ * <pre>
+ * {@code
+ *
+ * row_key = <shard> <queue_prefix> <write_pointer> <counter>
+ * shard = <salt> <consumer_group_id> <consumer_instance_id>
```
abaranau: so you will never mix data of different strategies in same table for now, alright

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -0,0 +1,161 @@
...
+ * salt = First byte of MD5 of <consumer_group_id>, <consumer_instance_id> and <queue_name>
```
abaranau: add indents for sub-breakdown for easier readability

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueProducer.java
chtyim: This is a placeholder for now and it might change a bit. The original idea was just to have all consumer configs being passed in and only get the unique groups in here (individual consumer config has the same group config if they are in the same group). Since we introduced ConsumerGroupConfig class, it probably better to do it at the queue client factory level. Will have that change when actually hooking up with shared queue

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -0,0 +1,161 @@
...
+ * consumer_group_id = 8 bytes long value of the target consumer group or 0 if it is FIFO
```
abaranau: > or 0 if it is FIFO

how is that related to group_id? why you don't need to know group_id in case of fifo? Or is consumer group has diferrent meaning now?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -0,0 +1,161 @@
...
+ * consumer_instance_id = 4 bytes int value of target consumer instance id or -1 if FIFO
+ * queue_prefix = <name_hash> <queue_name>
```
abaranau: why do you need name hash?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
abaranau: oh, I get it now. Please add a comment - why :)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
chtyim: Right. Sharded queue is on a different table (not in this PR).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
abaranau: I guess because in fifo it doesn't matter how many consumer instances there are, so we just share same row set between all consumer groups that use fifo?

chtyim: right.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
chtyim: hum.. actually we can get rid of it. It's a carry over from existing queue schema.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * An implementation of {@link HBaseQueueStrategy} with sharded keys.
+ *
+ * The row key has structure of:
+ *
+ * <pre>
+ * {@code
+ *
+ * row_key = <shard> <queue_prefix> <write_pointer> <counter>
```
abaranau: any chance you want to reserve a byte for format version? E.g. as we discussed this PR we talked that wwe may change format a bit.. - i.e. already ;)

abaranau: add a "hint todo" (and may be jira) for potential perf improvement: removing queue part of row key and encoding it all with consumer_group_id

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
chtyim: hum. Maybe not in this version. It's a bit complicated and involves quite a lot of code change to do that.

abaranau: yes - let's do outside of this PR

abaranau:
alright - basically, just couple of Qs

abaranau:
otherwise LGTM

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
chtyim: Hum, I believe reducing the row key size has very minimal effect on the sharded queue when considering real data of size couple hunder to thousand bytes. Better leave the hint for now to give more room/possibility on the perf improvement

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
chtyim: It's just BNF-like grammer, which usually doesn't have indentation.

PR #1587, opened 2015-03-11
fix classname in upgrade docs

rohitsinha54:


cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -591,7 +591,7 @@ and then restart CDAP.
 
 #. Run the upgrade tool::
 
-     /opt/cdap/master/bin/svc-master run co.cask.cdap.data.tools.Main upgrade
+     /opt/cdap/master/bin/svc-master run co.cask.cdap.data.tools.UpgraderMain upgrade
```
awholegunch: Name seems a little...odd. What about "MainUpgrader"?

awholegunch:
LGTM...though I can imagine the code needs changing...

PR #1592, opened 2015-03-12
(CDAP-1204) Converting HBaseConsumerStateStore to be Dataset

chtyim:


cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/Transactions.java
```diff
@@ -44,6 +50,21 @@ public static void invalidateQuietly(TransactionSystemClient txClient, Transacti
     }
   }
 
+  /**
+   * Handy method to create {@link TransactionExecutor} (See TEPHRA-71).
+   */
+  public static TransactionExecutor createTransactionExecutor(TransactionExecutorFactory factory,
+                                                              Iterable<? extends TransactionAware> txAwares) {
+    return factory.createExecutor(Iterables.transform(txAwares, Functions.<TransactionAware>identity()));
```
anew: What is the point of the identify transformation? Is it because createExecutor() wants a TransactionAware, not ? extends TransactionAware? If so, we should fix that in Tephra - can you open a Jira?

chtyim: I opened the JIRA already and it's mentioned in the javadoc... (See TEPHRA-71)

chtyim: In line 54.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -15,44 +15,260 @@
...
-  public void saveState(HBaseConsumerState state) throws IOException {
-    // Writes latest startRow to queue config.
-    hTable.put(state.updatePut(new Put(queueName.toBytes())));
-    hTable.flushCommits();
+  /**
+   * Updates the start row state of the given consumer.
+   */
+  void updateState(long groupId, int instanceId, byte[] startRow) {
+    byte[] stateColumn = getConsumerStateColumn(groupId, instanceId);
+    table.put(queueName.toBytes(), stateColumn, startRow);
```
anew: does it make sense to convert the queue name to bytes once in the constructor, instead of doing it over and over again? (I assume this gets called many, many times)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -15,44 +15,260 @@
...
+    Row row = scanner.next();
+    while (row != null) {
+      table.delete(row.getRow());
+      row = scanner.next();
+    }
+  }
+
+  /**
+   * Changes the number of consumer instances of the given consumer group. The consumer group configuration needs to be
+   * existed already.
```
anew: must exist already

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
chtyim: `QueueName` is doing that already, hence the `toByte()` call is just retuning a `byte[]`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -15,44 +15,260 @@
...
+        table.delete(rowKey, getConsumerStateColumn(removeGroupId, instanceId));
+      }
+    }
+
+    // For each group that changed (either a new group or number of instances change), update the startRow
+    for (Map.Entry<Long, Integer> entry : groupInfo.entrySet()) {
+      long groupId = entry.getKey();
+      int instances = entry.getValue();
+      if (!oldGroupInfo.containsKey(groupId)) {
+        // For new group, simply put with smallest rowKey from other group or an empty byte array if none exists.
```
anew: So this happens when a flow is updated and a new flowlet is added? Or how else can that happen?
In that case, where should the new flowlet start reading? Perhaps the current time? 
It seems arbitrary to just use some other group's minStartRow...

chtyim: This is ported from existing logic and is going to be different for the sharded queue (in future PR). 
I think the current assumption was that the eviction is doing it's job so that it's won't be too far away. However, I do agree that it's better be the current TX (which is what the sharded queue logic be).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -15,44 +15,260 @@
...
+   */
+  private void updateGroupInstances(long groupId, int instances, Map<Integer, byte[]> startRows) {
+    // Find smallest startRow among existing instances
+    byte[] smallest = null;
+    for (byte[] startRow : startRows.values()) {
+      if (smallest == null || Bytes.BYTES_COMPARATOR.compare(startRow, smallest) < 0) {
+        smallest = startRow;
+      }
+    }
+    Preconditions.checkArgument(smallest != null, "No startRow found for consumer group %s", groupId);
```
anew: would this happen if one deploys a flow, then updates the instances before starting the flow for the first time? If so, then we should support that case.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -15,44 +15,260 @@
...
 
   private final QueueName queueName;
-  private final ConsumerConfig consumerConfig;
-  private final HTable hTable;
+  private final Table table;
+  private final byte[] minStartRow;
 
-  HBaseConsumerStateStore(QueueName queueName, ConsumerConfig consumerConfig, HTable hTable) {
+  HBaseConsumerStateStore(QueueName queueName, Table table) {
+    super(Constants.SYSTEM_NAMESPACE + "." + QueueConstants.QueueType.QUEUE, table);
```
anew: Use HBaseQueueAdmin.STATE_STORE_NAME?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
chtyim: The configure methods only get called when the Flow get started or when changing number of instances while the Flow is running, hence this condition shouldn't happen.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
chtyim: haha, right. That constant was added after this. Will fix it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -188,7 +179,16 @@ public void create(QueueName queueName, Properties properties) throws IOExceptio
     // not accessible by the queue region coprocessor for doing eviction.
 
     // Create the config table first so that in case the queue table coprocessor runs, it can access the config table.
-    createConfigTable(queueName);
+    try {
+      DatasetProperties configProperties = DatasetProperties.builder()
+        .add(Table.PROPERTY_COLUMN_FAMILY, Bytes.toString(QueueEntryRow.COLUMN_FAMILY))
+        .build();
+      DatasetsUtil.createIfNotExists(datasetFramework,
```
anew: could this be done when the dataset framework is bootstrapped? This does not seem to be the right place to do it.

chtyim: We have one table per namespace and only have this table if there is Flow running.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueConsumer.java
```diff
@@ -40,14 +39,6 @@
  */
 abstract class HBaseQueueConsumer extends AbstractQueueConsumer {
 
-  private static final Logger LOG = LoggerFactory.getLogger(HBaseQueueConsumer.class);
```
anew: bookmark for Andreas.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueConsumer.java
```diff
@@ -133,26 +119,37 @@ public void close() throws IOException {
...
+
+  @Override
+  public void startTx(Transaction tx) {
+    super.startTx(tx);
+    stateStore.startTx(tx);
+  }
+
+  @Override
+  public boolean rollbackTx() throws Exception {
+    return super.rollbackTx() && stateStore.rollbackTx();
```
anew: why && ? Don't you want to rollback as much as possible even if the super class fails to rollback?

chtyim: that's true. Should rollback both first.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
anew: Never mind, I misread this, it creates an instance, not a type.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/Transactions.java
anew: oh, good. If it had been a comment on this line, saying, we have to do this identity transform until TEPHRA-71 is fixed, it would have be more obvious.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
anew: Oh, that is not obvious. Perhaps that method should be called asBytes() instead... But ok.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
anew: If this goes away, then let's not worry about it more.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
anew: ok, got it.

anew:
LGTM

chtyim:
Closing this and reopen against release branch.

PR #1594, opened 2015-03-12
Feature/cadp 1644 namespace meta

anew:
Removes id from the namespace meta data API. Changes UI, too, to rely only on the name only. 
This change is because for all other objects, we use  the name as the unique key.
All methods that use the id field are deprecated as of now. 
In the future, we will remove the id field completely.


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -76,9 +75,9 @@ public DefaultNamespaceAdmin(StoreFactory storeFactory, PreferencesStore prefere
    */
   private void createDefaultNamespace() {
     NamespaceMeta.Builder builder = new NamespaceMeta.Builder();
-    NamespaceMeta defaultNamespace = builder.setId(Constants.DEFAULT_NAMESPACE)
-      .setName("Default namespace")
-      .setDescription("The default namespace, a reserved system namespace.")
+    NamespaceMeta defaultNamespace = builder
+      .setName(Constants.DEFAULT_NAMESPACE)
+      .setDescription("Default Namespace")
```
bdmogal: Should we keep the description? @awholegunch had requested it to be that, so this particular namespace is more descriptive.

cdap-proto/src/main/java/co/cask/cdap/proto/NamespaceMeta.java
```diff
@@ -23,22 +23,19 @@
  * Represents metadata for namespaces
  */
 public final class NamespaceMeta {
+  // id is always identical to name. We keep this field here because the UI may depend on it.
```
anwar6953: since it seems like you're making updates to UI in this PR, can we not keep this field here then?

cdap-cli/src/main/java/co/cask/cdap/cli/ArgumentName.java
```diff
@@ -68,7 +68,7 @@
   SETTINGS("settings"),
   FREQUENCY("frequency"),
 
-  NAMESPACE_ID("namespace-id"),
+  NAMESPACE_NAME("namespace-id"),
```
bdmogal: should we call the argument *namespace-name*?

cdap-cli/src/main/java/co/cask/cdap/cli/ArgumentName.java
```diff
@@ -68,7 +68,7 @@
...
   NAMESPACE_DISPLAY_NAME("namespace-display-name"),
```
bdmogal: I guess we can remove this now.

cdap-cli/src/main/java/co/cask/cdap/cli/DefaultCompleters.java
```diff
@@ -55,7 +55,7 @@ public DefaultCompleters(Injector injector) {
         .put(ArgumentName.DATASET_MODULE_JAR_FILE.getName(), new FileNameCompleter())
         .put(ArgumentName.HTTP_METHOD.getName(), new EndpointCompleter())
         .put(ArgumentName.ENDPOINT.getName(), new EndpointCompleter())
-        .put(ArgumentName.NAMESPACE_ID.getName(), injector.getInstance(NamespaceIdCompleter.class))
+        .put(ArgumentName.NAMESPACE_NAME.getName(), injector.getInstance(NamespaceIdCompleter.class))
```
bdmogal: Should we rename ``NamespaceIdCompleter`` to ``NamespaceNameCompleter``?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamFetchHandler.java
```diff
@@ -212,6 +212,7 @@ private boolean verifyGetEventsRequest(Id.Stream streamId, long startTime, long
     }
     if (count <= 0) {
       responder.sendString(HttpResponseStatus.BAD_REQUEST, "Cannot request for <=0 events");
+      return false;
```
bdmogal: nice catch!

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgraderMain.java
```diff
@@ -372,7 +372,7 @@ private void createNamespaces() {
       Throwables.propagate(e);
     }
     LOG.info("Creating and registering {} namespace", Constants.DEFAULT_NAMESPACE);
-    getStore().createNamespace(new NamespaceMeta.Builder().setId(Constants.DEFAULT_NAMESPACE)
+    getStore().createNamespace(new NamespaceMeta.Builder()
```
bdmogal: This class got renamed to ``UpgradeTool``, but hopefully ``git`` should do its job.

cdap-unit-test-standalone/src/main/java/co/cask/cdap/test/standalone/StandaloneTestBase.java
```diff
@@ -89,6 +92,7 @@ public static void tearDownClass() throws Exception {
     if (standaloneMain != null && testStackIndex == 0) {
       standaloneMain.shutDown();
       standaloneMain = null;
+      LevelDBTableService.getInstance().clearTables();
```
bdmogal: Thanks for this fix!

cdap-unit-test-standalone/src/main/java/co/cask/cdap/test/standalone/StandaloneTestBase.java
```diff
@@ -65,6 +66,8 @@ public static void setUpClass() throws Exception {
         // Start without UI
         standaloneMain = StandaloneMain.create(null, configuration, new Configuration());
         standaloneMain.startUp();
+
```
bdmogal: Extra newlines?

bdmogal:
@anew Thanks a lot for the cleanup. The backend part looks good, just a few minor comments. 

Is someone reviewing the UI part?

Also, the merge conflict may be due to the renaming of ``UpgraderMain`` to ``UpgradeTool``.

cdap-ui/app/features/admin/controllers/NamespaceCreateController.js
```diff
@@ -8,10 +8,10 @@ angular.module(PKG.name + '.feature.admin')
     $scope.submitHandler = function() {
       myDataSrc.request({
         method: 'PUT',
-        _cdapPath: '/namespaces/' + $scope.model.id,
+        _cdapPath: '/namespaces/' + $scope.model.name,
         body: {
-          id: $scope.model.id,
-          name: $scope.model.name || null,
+          id: $scope.model.name,
```
bdmogal: @ajainarayanan would it be ok to not send ``id`` and ``name`` in the body? It shouldn't cause an issue, since the API will ignore it, but since we're cleaning up, just checking if we can clean this too?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
anew: The thing is, it is incorrect, It is not a system namespace. It is the default user namespace.

cdap-proto/src/main/java/co/cask/cdap/proto/NamespaceMeta.java
anew: I guess I could. The backend is not using it anywhere. Just wondering whether it can cause Gson problems if the UI adds it?

cdap-cli/src/main/java/co/cask/cdap/cli/ArgumentName.java
anew: yes

cdap-cli/src/main/java/co/cask/cdap/cli/ArgumentName.java
anew: done

cdap-cli/src/main/java/co/cask/cdap/cli/DefaultCompleters.java
anew: sure.

cdap-unit-test-standalone/src/main/java/co/cask/cdap/test/standalone/StandaloneTestBase.java
anew: removed

cdap-proto/src/main/java/co/cask/cdap/proto/NamespaceMeta.java
anew: OK I removed it

cdap-ui/app/features/admin/controllers/NamespaceCreateController.js
anew: I took the freedom to change it. Since it has no effect in the backend...

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
bdmogal: I see your point. Maybe we can call it that? However, we don't really expose the concept of user vs system namespace yet. Maybe best to keep it as "Default Namespace" then to avoid creating any confusion.

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgraderMain.java
anew: I did fine when I pulled develop into the branch.

anew:
@bdmogal addressed all comments. Please take another look.

bdmogal:
:+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
anew: Discussed this in person with @awholegunch and we agreed to change it to "Default Namespace". But we will explain in the docs that it is created by the system, and it is reserved in the sense that you can't delete or create it yourself.

bdmogal: Sounds good!

albertshau:
build green, merging

PR #1600, opened 2015-03-12
Breakdown the Dataset module to smaller modules to allow reusability in different scenario.

chtyim:
Main purpose is to simplify unit-test module construction.


cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DefaultDatasetRuntimeModule.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Map;
+
+/**
+ * Provides guice bindings for {@link DatasetModule} that are by default available in the system.
+ * The guice modules provided by this class are not intended to be used directly, but rather for providing
+ * injections to {@link DatasetFramework}, hence for installing from other guice module.
+ * This class is separated out so that combining different {@link DatasetFramework} and {@link DatasetModule} is
+ * easier, especially for unit-test.
+ */
+public class DefaultDatasetRuntimeModule extends RuntimeModule {
```
abaranau: don't like the name, but couldn't figure out better one...

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DefaultDatasetRuntimeModule.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @Override
+  public Module getInMemoryModules() {
+    return new AbstractModule() {
+      @Override
+      protected void configure() {
+        MapBinder<String, DatasetModule> mapBinder = MapBinder.newMapBinder(binder(), String.class, DatasetModule.class,
+                                                                            Names.named("defaultDatasetModules"));
+        // NOTE: order is important due to dependencies between modules
+        mapBinder.addBinding("orderedTable-memory").toInstance(new InMemoryTableModule());
```
ghelmling: Since only the first two entries vary for each of these, could we move the common bindings to another method that is invoked here, like addCommon(MapBinder)?  That might make it a little bit easier to maintain the common bindings in the future.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DefaultDatasetRuntimeModule.java
chtyim: I am open to any naming suggestions. How about `SystemDatasetRuntimeModule`??

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DefaultDatasetRuntimeModule.java
abaranau: +1

ghelmling:
Very nice clean up.  Only one minor comment on the duplication of some of the common module bindings.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DefaultDatasetRuntimeModule.java
chtyim: ha.. didn't notice that it has so many common modules (copied from old code). Will unify them through a private method.

abaranau:
The changes make sense to me. I assume you tested them in 3 modes. LGTM in that case - after fixing gary's comment

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DefaultDatasetRuntimeModule.java
ghelmling: Makes sense, since these are "system datasets".

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/SystemDatasetRuntimeModule.java
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Add bindings for Dataset modules that are available by default
+   */
+  private void bindDefaultModules(MapBinder<String, DatasetModule> mapBinder) {
+    mapBinder.addBinding("core").toInstance(new CoreDatasetsModule());
+    mapBinder.addBinding("fileSet").toInstance(new FileSetModule());
+    mapBinder.addBinding("timePartitionedFileSet").toInstance(new TimePartitionedFileSetModule());
+    mapBinder.addBinding("partitionedFileSet").toInstance(new PartitionedFileSetModule());
+    mapBinder.addBinding("objectMappedTable").toInstance(new ObjectMappedTableModule());
+
```
ghelmling: minor nit: drop blank line

ghelmling:
One very minor nit.  Lgtm to merge, assuming tests in all 3 modes pass.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/SystemDatasetRuntimeModule.java
chtyim: Fixed.

chtyim:
Test running in http://builds.cask.co/browse/CDAP-DUT1088-3

chtyim:
Test passed. Merging.

PR #1602, opened 2015-03-12
CDAP-1759: Namespace TestManager

jwang47:
https://issues.cask.co/browse/CDAP-1759
http://builds.cask.co/browse/CDAP-RBT135


jwang47:
@bdmogal Regarding your comment "At least in the program lifecycle APIs, I know there is a jira to drop support for procedures with v3 apis. we may want to remove this/ always use default namespace", it's difficult to do this at the moment since most of the tests in `TestFrameworkTest` are using procedures to check results. I suggest we create another issue to remind us to port `TestFrameworkTest` to use services instead later.

bdmogal:
That's fine then. Let's do it as a separate jira/PR

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
```diff
@@ -240,6 +244,67 @@ public String toString() {
   }
 
   /**
+   * Uniquely identifies a Schedule.
+   */
+  public static class Schedule {
```
anwar6953: unused?

cdap-test/src/main/java/co/cask/cdap/test/ApplicationManager.java
```diff
@@ -101,8 +101,11 @@
    * @param dataSetName Name of the dataset to retrieve.
    * @param <T> Type of the dataset.
    * @return A {@link DataSetManager} instance.
+   * @deprecated As of version 2.8.0, replaced by
+   *             {@link TestManager#getDataset(co.cask.cdap.proto.Id.Namespace, String)}
```
anwar6953: I think this is wrong.
It's replaced by TestBase#getDataset, right?

anwar6953:
remainder lgtm

anwar6953:
:+1:

PR #1603, opened 2015-03-12
CDAP-1368: Use v3 APIs in TestFrameworkTest, AppFabricTestHelper for lifecycle management

jwang47:
Depends on https://github.com/caskdata/cdap/pull/1602 (merged) and https://github.com/caskdata/cdap/pull/1725.

https://issues.cask.co/browse/CDAP-1368
http://builds.cask.co/browse/CDAP-RBT137


jwang47:
Waiting on some unit test framework issues which are causing one of the tests to fail.

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -69,60 +77,88 @@
...
     String uri = String.format("/v2/unrecoverable/reset");
     HttpRequest request = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.PUT, uri);
     httpHandler.resetCDAP(request, responder);
     verifyResponse(HttpResponseStatus.OK, responder.getStatus(), "Reset application failed");
-  }
 
-  public void startProgram(String appId, String flowId, ProgramType type, Map<String, String> args) {
+    // delete all namespaces
+    for (NamespaceMeta namespaceMeta : namespaceAdmin.listNamespaces()) {
+      if (!Constants.DEFAULT_NAMESPACE.equals(namespaceMeta.getName()) &&
```
anwar6953: I don't think this check needs to be here.
I believe that the namespacehttpHandler is capable of deletes to default namespace.
And also, I don't think system namespace is exposed in listNamespaces.

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -69,60 +77,88 @@
...
+        !Constants.SYSTEM_NAMESPACE.equals(namespaceMeta.getName())) {
+        Id.Namespace namespace = Id.Namespace.from(namespaceMeta.getName());
+        streamAdmin.dropAllInNamespace(namespace);
+        namespaceHttpHandler.deleteDatasets(null, new MockResponder(), namespaceMeta.getName());
```
anwar6953: instead of passing null for the request, does it make sense to be consistent with other methods in this class?
`new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.POST, uri)`

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -69,60 +77,88 @@
...
+    this.namespaceHttpHandler = namespaceHttpHandler;
+    this.namespaceAdmin = namespaceAdmin;
+    this.streamAdmin = streamAdmin;
+    this.streamMetaStore = streamMetaStore;
   }
 
-  public void reset() {
+  public void reset() throws Exception {
     MockResponder responder = new MockResponder();
     String uri = String.format("/v2/unrecoverable/reset");
```
anwar6953: if we remove the condition on line 114, we can also remove this call to unrecoverable/reset.

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -69,60 +77,88 @@
...
 
-  public void startProgram(String appId, String flowId, ProgramType type, Map<String, String> args) {
+    // delete all namespaces
+    for (NamespaceMeta namespaceMeta : namespaceAdmin.listNamespaces()) {
+      if (!Constants.DEFAULT_NAMESPACE.equals(namespaceMeta.getName()) &&
+        !Constants.SYSTEM_NAMESPACE.equals(namespaceMeta.getName())) {
+        Id.Namespace namespace = Id.Namespace.from(namespaceMeta.getName());
+        streamAdmin.dropAllInNamespace(namespace);
+        namespaceHttpHandler.deleteDatasets(null, new MockResponder(), namespaceMeta.getName());
+        namespaceHttpHandler.delete(null, new MockResponder(), namespaceMeta.getName());
```
anwar6953: I think we should be consistent with other methods here.
`verifyResponse(HttpResponseStatus.OK, responder.getStatus(), "namespace delete failed" + responder.getMessage);` or something like that

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -325,6 +327,7 @@ public void truncate(Id.Stream streamId) throws Exception {
   public void drop(Id.Stream streamId) throws Exception {
     // Same as truncate
     truncate(streamId);
+    alterExploreStream(streamId, false);
```
anwar6953: nice catch :+1:

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -110,7 +110,9 @@ public void dropAllInNamespace(Id.Namespace namespace) throws Exception {
     }
 
     for (final Location streamLocation : locations) {
+      Id.Stream stream = StreamUtils.getStreamIdFromLocation(streamLocation);
       doTruncate(streamLocation);
```
anwar6953: hmm, can we just call drop(stream) here then?

anwar6953:
Another concern: a lot of the methods in AppFabricClient are also used for procedures.
Wouldn't that make all requests go to v3, even procedure requests?

jwang47:
Good point. I'll have to change that too.

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -131,47 +173,48 @@ public void setWorkerInstances(String appId, String workerId, int instances) {
     verifyResponse(HttpResponseStatus.OK, responder.getStatus(), "Set worker instances failed");
   }
 
-  public void setRunnableInstances(String applicationId, String serviceName, String runnableName, int instances) {
+  public void setRunnableInstances(String namespaceId, String applicationId, String serviceName, int instances) {
```
anwar6953: rename the method, if its not for runnable then?

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -131,47 +173,48 @@ public void setWorkerInstances(String appId, String workerId, int instances) {
...
     HttpRequest request = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.PUT, uri);
     JsonObject json = new JsonObject();
     json.addProperty("instances", instances);
     request.setContent(ChannelBuffers.wrappedBuffer(json.toString().getBytes()));
-    serviceHttpHandler.setInstances(request, responder, applicationId, serviceName, runnableName);
+    programLifecycleHttpHandler.setServiceInstances(request, responder, namespaceId, applicationId, serviceName);
     verifyResponse(HttpResponseStatus.OK, responder.getStatus(), "Set runnable instances failed");
   }
 
-  public ServiceInstances getRunnableInstances(String applicationId, String serviceName, String runnableName) {
```
anwar6953: rename the method, if its not for runnable then?

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -69,109 +78,184 @@
...
-    HttpRequest request = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.POST, uri);
-    String argString = GSON.toJson(args);
-    if (argString != null) {
-      request.setContent(ChannelBuffers.wrappedBuffer(argString.getBytes(Charsets.UTF_8)));
+    if (ProgramType.PROCEDURE.equals(type)) {
+      Preconditions.checkArgument(Constants.DEFAULT_NAMESPACE.equals(namespaceId));
+      String uri = String.format("/v2/apps/%s/%s/%s/start",
+                                 appId, type.getCategoryName(), flowId);
+      HttpRequest request = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.POST, uri);
+      String argString = GSON.toJson(args);
```
anwar6953: this statement doesn't need to be duplicated in both if cases, right? just pull one instance to before the if-else.

anwar6953:
remainder lgtm

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
jwang47: Yeah, but not duplicating it doesn't seem like much of an improvement, and we're removing procedure support anyway, so I just did the quickest thing.

anwar6953: seems fine to me.
lgtm

sreevatsanraman:
LGTM :+1:

PR #1606, opened 2015-03-12
Deleting schedules on deleting namespace.

sagarkapare:
Jira - https://issues.cask.co/browse/CDAP-1758


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
```diff
@@ -277,6 +279,12 @@ public void testDeleteAll() throws Exception {
     Assert.assertTrue(streamAdmin.exists(myStream));
     assertResponseCode(200, deleteNamespace(OTHER_NAME));
     Assert.assertFalse(streamAdmin.exists(myStream));
+
+    assertResponseCode(200, createNamespace(OTHER_NAME));
```
bdmogal: why create namespace after deleting ? Maybe we can test this before deleting the namespace, so we don;t have to create -> delete -> create -> delete?

bdmogal:
@sagarkapare should you be able to delete a namespace when a schedule is active? If not, we may want to test that in the same unit test you updated as well.

bdmogal:
Just a couple of comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -221,6 +235,22 @@ public void deleteNamespace(final Id.Namespace namespaceId)
     LOG.info("All data for namespace '{}' deleted.", namespaceId);
   }
 
+  private void deleteAllSchedules(Id.Namespace namespaceId) throws NotFoundException, SchedulerException {
+    List<ApplicationSpecification> allAppSpec =
+      new ArrayList<ApplicationSpecification>(store.getAllApplications(namespaceId));
+
+    for (ApplicationSpecification appSpec : allAppSpec) {
+      for (Map.Entry<String, ScheduleSpecification> entry : appSpec.getSchedules().entrySet()) {
```
rohitsinha54: If you don't use the keys just do values.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
sagarkapare: We need to delete the namespace which will delete the schedules and create the namespace with same name again and deploy the app, which should succeed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
sagarkapare: Fixed.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
bdmogal: Ok, I probably missed line 258. Anyway, maybe we should add a comment on why we need to drop and recreate the namespace then.

cdap-app-fabric/src/test/java/co/cask/cdap/AppForUnrecoverableResetTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import co.cask.cdap.api.app.AbstractApplication;
+import co.cask.cdap.api.mapreduce.AbstractMapReduce;
+import co.cask.cdap.api.schedule.Schedules;
+import co.cask.cdap.api.workflow.AbstractWorkflow;
+
+/**
+ *
+ */
+public class AppForUnrecoverableResetTest extends AbstractApplication {
```
bdmogal: empty javadoc

cdap-app-fabric/src/test/java/co/cask/cdap/AppForUnrecoverableResetTest.java
```diff
@@ -0,0 +1,52 @@
...
+    protected void configure() {
+      setName("DummyWorkflow");
+      addMapReduce("DummyMR");
+    }
+  }
+
+  /**
+   *
+   */
+  public static class DummyMR extends AbstractMapReduce {
```
bdmogal: although does this class need to be public?

bdmogal:
A couple of minor comments. Otherwise LGTM

cdap-app-fabric/src/test/java/co/cask/cdap/AppForUnrecoverableResetTest.java
sagarkapare: Fixed.

cdap-app-fabric/src/test/java/co/cask/cdap/AppForUnrecoverableResetTest.java
sagarkapare: Removed public

sagarkapare:
Build is running - http://builds.cask.co/browse/CDAP-RBT139-1

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -208,6 +219,9 @@ public void deleteNamespace(final Id.Namespace namespaceId)
         // Finally delete namespace from MDS
         store.deleteNamespace(namespaceId);
       }
+    } catch (SchedulerException e) {
+      LOG.warn("Error while deleting namespace {}", namespaceId, e);
+      throw new NamespaceCannotBeDeletedException(namespaceId.getId(), e);
```
gsps1: we are using the same warning message and we throw "NamespaceCannotBeDeletedException" for all the cases, should we just have the exception case then, if we don't have to handle them specially ? @bdmogal

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -221,6 +235,21 @@ public void deleteNamespace(final Id.Namespace namespaceId)
     LOG.info("All data for namespace '{}' deleted.", namespaceId);
   }
 
+  private void deleteAllSchedules(Id.Namespace namespaceId) throws NotFoundException, SchedulerException {
+    List<ApplicationSpecification> allAppSpec =
+      new ArrayList<ApplicationSpecification>(store.getAllApplications(namespaceId));
```
gsps1: we generally use Lists.newArrayList

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -221,6 +235,21 @@ public void deleteNamespace(final Id.Namespace namespaceId)
...
+
+    for (ApplicationSpecification appSpec : allAppSpec) {
+      for (ScheduleSpecification scheduleSpec : appSpec.getSchedules().values()) {
```
gsps1: do you think its better to have a method ```deleteAllSchedules(Id.Namespace namespaceId, ApplicaitonSpecification appSpec)``` , so you can delete schedules in an application, and you can use it here.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
```diff
@@ -277,6 +279,14 @@ public void testDeleteAll() throws Exception {
...
     assertResponseCode(200, deleteNamespace(OTHER_NAME));
     Assert.assertFalse(streamAdmin.exists(myStream));
+
+    // Create the namespace again and deploy the application containing schedules.
+    // Application deployment should succeed.
+    assertResponseCode(200, createNamespace(OTHER_NAME));
+    HttpResponse response = deploy(AppForUnrecoverableResetTest.class, Constants.Gateway.API_VERSION_3_TOKEN,
+                                   OTHER_NAME);
+    Assert.assertEquals(200, response.getStatusLine().getStatusCode());
+    assertResponseCode(200, deleteNamespace(OTHER_NAME));
```
gsps1: do we delete the schedules when we delete an app? or only during unrecoverable reset ? it may be better to delete the app with schedules and test re-deployment of that app, rather than adding the test here in deleteAll, right ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -221,6 +235,21 @@ public void deleteNamespace(final Id.Namespace namespaceId)
     LOG.info("All data for namespace '{}' deleted.", namespaceId);
   }
 
+  private void deleteAllSchedules(Id.Namespace namespaceId) throws NotFoundException, SchedulerException {
```
gsps1: not sure if this method belongs here

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
sagarkapare: Good point! Using Exception only.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
sagarkapare: Updated.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
sagarkapare: Updated the code.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
sagarkapare: The change is only related to the unrecoverable reset. We are already deleting the schedules when we delete the application.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
sagarkapare: Moved the method to the `Schedule` interface and added its implementation in the classes that inherits it.

gsps1:
please open a jira about the duplication of methods between TimeScheduler and StreamScheduler and refactoring to use the Abstract class. otherwise just one minor comment, please fix and merge! :+1:

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
```diff
@@ -277,6 +279,14 @@ public void testDeleteAll() throws Exception {
     Assert.assertTrue(streamAdmin.exists(myStream));
     assertResponseCode(200, deleteNamespace(OTHER_NAME));
     Assert.assertFalse(streamAdmin.exists(myStream));
+
+    // Create the namespace again and deploy the application containing schedules.
```
anwar6953: Is this depending on the idea that deploying an app with a schedule would fail, if the app doesn't already exist, but the schedule does?

sagarkapare: Yes. Thats correct. Unrecoverable reset deletes the app but not schedule. So deploying the app again fails with SchedulerException.

anwar6953: got it.

gsps1:
LGTM, just make sure a new build passes before merging.

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-RBT139-10

sagarkapare:
Build passed. Merging.

PR #1593, opened 2015-03-12
CDAP-1417 Port to 2.6 - Capabilities to submit programs in non-defaut yarn Queues

poornachandra:
https://issues.cask.co/browse/CDAP-1417

Porting from https://github.com/caskdata/cdap/pull/1529

PR #1611, opened 2015-03-12
CDAP-1417 Port to 2.6 - Capabilities to submit programs in non-default yarn Queues

poornachandra:
Reopening https://github.com/caskdata/cdap/pull/1593


ParthGandhi:
#1593  was already reviewed. This is the same changes. LGTM.

PR #1612, opened 2015-03-12
Failure of the program should move it to KILLED instead of COMPLETED.

sagarkapare:
Jira: https://issues.cask.co/browse/CDAP-1744
Failure of a program was incorrectly attempting to mark them as COMPLETED. It should be marked KILLED.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/ProgramControllerServiceAdapter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/ProgramControllerServiceAdapter.java
@@ -64,7 +64,7 @@ public class ProgramControllerServiceAdapter extends AbstractProgramController {
   /**
    * Returns {@code true} if service error is propagated as error for this controller state.
    * If this method returns {@code false}, service error is just getting logged and state will just change
-   * to {@link State#COMPLETED}. This method returns {@code true} by default.
+   * to {@link State#KILLED}. This method returns {@code true} by default.
    */
   protected boolean propagateServiceError() {
     return true;
@@ -83,8 +83,6 @@ public class ProgramControllerServiceAdapter extends AbstractProgramController {
         if (propagateServiceError()) {
           error(failure);
         } else if (getState() != State.STOPPING) {
-          complete();
-        } else {
           stop();
         }
       }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/ProgramControllerServiceAdapter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/ProgramControllerServiceAdapter.java
@@ -61,15 +61,6 @@ public class ProgramControllerServiceAdapter extends AbstractProgramController {
 
   }
 
-  /**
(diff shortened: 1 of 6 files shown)


jwang47:
LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/ProgramControllerServiceAdapter.java
```diff
@@ -80,13 +71,7 @@ public void running() {
...
       public void failed(Service.State from, Throwable failure) {
         LOG.error("Program terminated with exception", failure);
-        if (propagateServiceError()) {
-          error(failure);
-        } else if (getState() != State.STOPPING) {
-          complete();
-        } else {
-          stop();
-        }
+        error(failure);
```
awholegunch: Line 79: should it be "Service.State.STOPPING"?

sagarkapare: Not sure which line you mean. Please explain.

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-RBT160-1

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/ProgramControllerServiceAdapter.java
awholegunch: If you expand below these comments (using the icon in the left-sdie, where the line numbers are), you can display the code below where you made your changes. On line 94/79 (old/new line numbers), it says:
```
if (getState() != State.STOPPING) {
```
Should that not be ``Service.State.STOPPING``?

sagarkapare: Ohh ok. Got it.
No. It should not be `Service.State.STOPPING`. We are comparing the `ProgramController` states here and not the `Service` states.

awholegunch: But the object "State" in line 79 is never defined. What is it?

sagarkapare: Its `enum` in the `ProgramController` interface. The `getState()` method returns the state of the current ProgramController. We compare it with the state `State.STOPPING`. Since we are already in the `ProgramControllerServiceAdapter` which implements in the `ProgramController` interface, we don't need to qualify it as `ProgramController.State.STOPPING`

awholegunch: OK, I looked up ProgramController, and that's where State lives.

sagarkapare:
@chtyim @jwang47 Added one more commit to disable the error propagation from the WorkflowTwillRunnable. Please review.

jwang47:
Change LGTM

sagarkapare:
Build passed here http://builds.cask.co/browse/CDAP-RBT160-3. Merging.

PR #1616, opened 2015-03-12
fix upgrade tool filset key prefix

rohitsinha54:
We read the new meta table to get the filesets which is written with new key format. So read it in that way.
I had this fixed during the review process but looks like it got lost some where :/


albertshau:
:+1:

anwar6953:
LGTM also

PR #1617, opened 2015-03-12
Feature/ui split flow graph directive

nistala:
- workflow diagrams


ajainarayanan:
@nistala @astewes Sorry guys I think I might have mis-communicated when I said it will be nice to have generic functionality separated out. I have made two commits that communicates that idea. Please review when you have some time.. 
We still need to do it for flowgraphs.

ajainarayanan:
:+1: LGTM

PR #1621, opened 2015-03-13
CDAP-1743: Use unique path for preferences

jwang47:
https://issues.cask.co/browse/CDAP-1743
http://builds.cask.co/browse/CDAP-RBT143


gokulavasan:
LGTM :+1:

PR #1624, opened 2015-03-13
Bugfix/default ns

bdmogal:
- [x] Ensures that the default namespace exists asynchronously during ``AppFabricServer`` startup.
- [x] Creates default namespace in tests that do not start ``AppFabricServer``

Jira: [CDAP-1213](https://issues.cask.co/browse/CDAP-1213)
Build: http://builds.cask.co/browse/CDAP-RBT144-2

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerTestBase.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerTestBase.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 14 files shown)


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerTestBase.java
```diff
@@ -0,0 +1,47 @@
...
+import co.cask.cdap.proto.ProgramType;
+import com.google.common.base.Predicate;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Base class for scheduler tests
+ */
+public class SchedulerTestBase {
+  protected void waitUntilFinished(ProgramRuntimeService runtimeService,
```
anwar6953: why not put common fields (such as Store, NamespaceAdmin) in here?
And also make it an abstract class?

bdmogal: A lot more than that needs to be moved here - https://issues.cask.co/browse/CDAP-1807.

Wanted to get a review for these changes, then do it in a separate PR.

anwar6953: Ah right; ok.

bdmogal:
Build is green - http://builds.cask.co/browse/CDAP-RBT144-4

anwar6953:
lgtm :+1:

bdmogal:
Thanks @anwar6953. Merging.

PR #1627, opened 2015-03-13
Move Remote*Manager from cdap-test back to cdap-integration-test

jwang47:
http://builds.cask.co/browse/CDAP-RBT147


anwar6953:
LGTM

PR #1635, opened 2015-03-14
bug fix-v2 API : use tag-value "any" for program-type when type name is missing 

gsps1:
https://issues.cask.co/browse/CDAP-1591


gsps1:
http://builds.cask.co/browse/CDAP-RBT150-1

albertshau:
:+1:

PR #1636, opened 2015-03-14
Fix how embedded datasets are migrated (builder namespaces them with par...

anwar6953:
...ent ds names).

Since the key in the datasetSpecifications are not prefixed by parent's ds spec names, we simply use that as the name. And let the build() method take care of propogating the new parent name down to the children ds specs.

Resolves: https://issues.cask.co/browse/CDAP-1818
Build: http://builds.cask.co/browse/CDAP-RBT148-2


bdmogal:
:+1:

PR #1638, opened 2015-03-14
remove duplicate information (appId) from every row. also use programTyp...

anwar6953:
...e#getPrettyName rather than getCategoryName

Build: http://builds.cask.co/browse/CDAP-RBT152-1


jwang47:
LGTM

PR #1639, opened 2015-03-15
Bugfix/cdap 1589 disallow streams in nonexistent namespace

anwar6953:
JIRA: https://issues.cask.co/browse/CDAP-1589
Build: 
http://builds.cask.co/browse/CDAP-RBT141-7


cdap-app-fabric/src/test/java/co/cask/cdap/runtime/FlowTest.java
```diff
@@ -92,6 +98,13 @@ public File get() {
 
   private static final Logger LOG = LoggerFactory.getLogger(FlowTest.class);
 
+  @BeforeClass
+  public static void init() throws AlreadyExistsException, NamespaceCannotBeCreatedException {
+    AppFabricTestHelper.getInjector().getInstance(LocationFactory.class);
```
bdmogal: this line can be removed?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/deploy/LocalManagerTest.java
```diff
@@ -53,6 +55,10 @@
   public static void before() throws Exception {
     lf = new LocalLocationFactory();
     temp = TMP_FOLDER.newFolder("pipeline");
+
+    AppFabricTestHelper.getInjector().getInstance(LocationFactory.class);
```
bdmogal: hmmm, unnecessary?

cdap-data-fabric-tests/src/test/java/co/cask/cdap/data/stream/StreamCoordinatorTestBase.java
```diff
@@ -48,6 +50,13 @@
 
   protected abstract StreamAdmin getStreamAdmin();
 
+  protected static void setupNamespaces(LocationFactory locationFactory) throws IOException {
+    // FileStreamAdmin expects namespace directory to exist.
+    // Simulate namespace create
+    locationFactory.create(Constants.DEFAULT_NAMESPACE).mkdirs();
+  }
+
+
```
bdmogal: extra newline

cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/transaction/stream/StreamConsumerStateTestBase.java
```diff
@@ -41,6 +42,12 @@
   protected static final Id.Namespace TEST_NAMESPACE = Id.Namespace.from("streamConsumerStateTestNamespace");
   protected static final Id.Namespace OTHER_NAMESPACE = Id.Namespace.from("otherNamespace");
 
+  protected static void setupNamespaces(LocationFactory locationFactory) throws IOException {
+    locationFactory.create(TEST_NAMESPACE.getId()).mkdirs();
+    locationFactory.create(OTHER_NAMESPACE.getId()).mkdirs();
+  }
+
+
```
bdmogal: extra newline

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamHandler.java
```diff
@@ -200,8 +208,11 @@ public void create(HttpRequest request, HttpResponder responder,
       // TODO: For create successful, 201 Created should be returned instead of 200.
       responder.sendStatus(HttpResponseStatus.OK);
     } catch (IllegalArgumentException e) {
-      responder.sendString(HttpResponseStatus.BAD_REQUEST,
-                           e.getMessage());
+      LOG.error("Failed to create stream {}", stream, e);
+      responder.sendString(HttpResponseStatus.BAD_REQUEST, e.getMessage());
+    } catch (NotFoundException e) {
+      LOG.error("Failed to create stream {}", stream, e);
+      responder.sendString(HttpResponseStatus.BAD_REQUEST, e.getMessage());
```
bdmogal: what happens to other exceptions here? get propagated as ``INTERNAL_SERVER_ERROR``? Maybe add a catch block for ``Exception`` to do that explicitly?


Come to think of it, is it better to catch these 'specific' exceptions where they are thrown, log, respond and return? Meaning ``IllegalArgumentException`` would move to line 200, ``NotFoundException`` to line 203. Then you would catch a generic ``Exception`` here to send ``INTERNAL_SERVER_ERROR``?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamServiceRuntimeModule.java
```diff
@@ -73,6 +74,7 @@ protected void configure() {
         bind(StreamFileJanitorService.class).to(DistributedStreamFileJanitorService.class).in(Scopes.SINGLETON);
         bind(StreamWriterSizeCollector.class).to(BasicStreamWriterSizeCollector.class).in(Scopes.SINGLETON);
         bind(StreamService.class).to(DistributedStreamService.class).in(Scopes.SINGLETON);
+        bind(AbstractNamespaceClient.class).to(DiscoveryNamespaceClient.class).in(Scopes.SINGLETON);
```
bdmogal: no way to share bindings between the three modes, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -266,6 +266,7 @@ public void create(Id.Stream streamId) throws Exception {
 
   @Override
   public void create(final Id.Stream streamId, @Nullable final Properties props) throws Exception {
+    assertNamespaceHomeExists(streamId.getNamespace());
```
bdmogal: Is this needed? I mean can you access ``FileStreamAdmin`` outside of the HTTP Handler, which would make sure that the namespace exists.

It may be good to double check the directory like you are doing, but just wondering.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -296,6 +297,13 @@ public StreamConfig call() throws Exception {
     });
   }
 
+  private void assertNamespaceHomeExists(Id.Namespace namespaceId) throws IOException {
+    Location namespaceHomeLocation = Locations.getParent(getStreamsHomeLocation(namespaceId));
+    Preconditions.checkArgument(namespaceHomeLocation.exists(),
+                                String.format("Expected home directory %s for namespace %s to exist.",
```
bdmogal: Should we standardize this message between the app deploy, dataset module deploy and this methods? They essentially do the same thing, even if they are defined separately, it may make sense to at lease define the same message for all three.

cdap-explore/src/test/java/co/cask/cdap/explore/service/BaseHiveExploreServiceTest.java
```diff
@@ -276,11 +284,10 @@ protected static void assertStatementResult(ListenableFuture<ExploreExecutionRes
     return newResults;
   }
 
-  protected static void createStream(String namespace, String streamName) throws IOException {
-    HttpURLConnection urlConn = openStreamConnection(namespace, streamName);
-    urlConn.setRequestMethod(HttpMethod.PUT);
-    Assert.assertEquals(HttpResponseStatus.OK.getCode(), urlConn.getResponseCode());
-    urlConn.disconnect();
+  protected static void createStream(String namespace, String streamName) throws Exception {
```
bdmogal: why not accept ``Id.Stream`` here?

bdmogal:
Overall looks good. few minor comments.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
anwar6953: Do you mean to check the namespace dir existence from the handler?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamHandler.java
anwar6953: I added a catch(Exception) at the end, but having the other exceptions higher up means many more try catch blocks, which makes the code messy without too much benefit, imo.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamServiceRuntimeModule.java
anwar6953: If there is, I'm not sure how. We can put that one binding in a separate module, and use that module. I'm not sure thats necessary though - it'd be a very tiny module, and would still need to be repeated across the three modes.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
anwar6953: Suggestion on how/where to do this?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
bdmogal: No, if we have to check, its fine here. Just wondering if the check is needed at all, since the Handler ensures that the namespace exits.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
bdmogal: I would say just copy the message from ``AppLifecycleHttpHandler`` or ``DatasetTypeHandler``

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamServiceRuntimeModule.java
bdmogal: Yes I think eventually need s ``NamespaceRuntimeModule``.Even that module would need to be added in all 3 modes here, right. Fine for now. Maybe open a TechDebt jira?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamHandler.java
bdmogal: Ok.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
anwar6953: Yea, I think its worth it to do the check here. Incase FileStreamAdmin is used outside the handler.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamServiceRuntimeModule.java
anwar6953: I don't think its worth it to have a NamespaceRuntimeModule at this point. What class bindings would be defined in it besides this?

bdmogal: when I added various deletions to namespace delete, I had to add 3 bindings at multiple places (2-3). that's when the thought crossed my mind. nothing you can do in this PR, though.

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/InMemoryNamespaceClient.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.inject.Inject;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.List;
+
+/**
+ *
+ */
+public class InMemoryNamespaceClient extends AbstractNamespaceClient {
```
bdmogal: This isn't really in-memory, no. You're bypassing the remote call through other namespace clients, so maybe LocalNamespaceClient. Also, empty javadoc.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
bdmogal: SGTM

bdmogal:
Apart from one last comment, LGTM. Ok to fix and merge, pending build.

PR #1640, opened 2015-03-15
CDAP-1819: Fix how CLI handles --uri and changing namespace

jwang47:
https://issues.cask.co/browse/CDAP-1819
http://builds.cask.co/browse/CDAP-RBT153


cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -243,6 +241,13 @@ public InputStream getInput() throws IOException {
     }
   }
 
+  public void setNamespace(Id.Namespace namespace) {
+    ConnectionConfig connectionConfig = ConnectionConfig.builder(clientConfig.getConnectionConfig())
+      .setNamespace(namespace)
+      .build();
+    this.setConnectionConfig(connectionConfig);
```
anwar6953: if we're just setting the namespace, do we need to trigger a call to `tryAutoconnect`?

anwar6953:
lgtm

nitinmotgi:
lgtm :+1:

PR #1641, opened 2015-03-15
CDAP-1820: Treat ! as plain char in CLI commands

jwang47:
https://issues.cask.co/browse/CDAP-1820
http://builds.cask.co/browse/CDAP-RBT154


nitinmotgi:
LGMT :+1: 
http://code.metager.de/source/xref/scala/src/jline/src/main/java/scala/tools/jline/console/ConsoleReader.java#350

PR #1642, opened 2015-03-15
v2 Workflow spec api returns the specifications in pre 2.8 format

sagarkapare:
Jira - https://issues.cask.co/browse/CDAP-1669

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
@@ -17,6 +17,14 @@
 package co.cask.cdap.gateway.handlers;
 
 import co.cask.cdap.api.data.stream.StreamSpecification;
+import co.cask.cdap.api.schedule.SchedulableProgramType;
+import co.cask.cdap.api.workflow.ScheduleProgramInfo;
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
```diff
@@ -756,9 +773,62 @@ public void sparkSpecification(HttpRequest request, HttpResponder responder,
   public void workflowSpecification(HttpRequest request, HttpResponder responder,
                                     @PathParam("app-id") final String appId,
                                     @PathParam("workflow-id")final String workflowId) {
-    programLifecycleHttpHandler.programSpecification(RESTMigrationUtils.rewriteV2RequestToV3(request), responder,
```
nitinmotgi: Why are we diverging from centralizing this into a programSpecification ?

sagarkapare: We changed WorkflowSpecifications in 2.8. However we still want V2 WorkflowSpecification api to return the data in the format which is compatible to 2.7, which is why special handling is needed here.

sreevatsanraman:
LGTM :+1:

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-RBT170

PR #1643, opened 2015-03-16
Removing NUX and Checking for updates. Also, makes it work without inter...

nitinmotgi:
[x] Removed NUX
[x] Removed check_for_updates
[x] Fixed issue when CDAP wouldn't start because of reverse lookup of hostname when not connected to internet.


sreevatsanraman:
The logic to get host name is missing in bat script

sreevatsanraman:
When you start from IDE the host name will be missing. Perhaps you can fall back to the older logic of determining host name if the host name is empty - to fix the two cases.

sreevatsanraman:
LGTM :+1:

PR #1644, opened 2015-03-16
[CDAP-1365] Suspend all schedules when deploying an app in unit tests

jguery:
Build: http://builds.cask.co/browse/CDAP-RBT157
Jira: https://issues.cask.co/browse/CDAP-1365


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -211,13 +215,25 @@ public BodyConsumer deploy(HttpRequest request, HttpResponder responder,
...
       responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Deploy failed: " + ex.getMessage());
       return null;
     }
   }
 
+  private boolean suspendScheduleOnDeploy(HttpRequest request) throws URISyntaxException {
+    // Optional argument to suspend schedules upon application deployment
+    Map<String, List<String>> queryParams = new QueryStringDecoder(new URI(request.getUri())).getParameters();
+    boolean suspendSchedules = false;
+    List<String> values = queryParams.get(Constants.Scheduler.SUSPEND_SCHEDULES_DEPLOY_OPTION);
```
jwang47: I think we can use `@QueryParam` instead of manually parsing the query param here.

jwang47:
If this is purely for unit tests, then a more simple alternative would be to suspend all schedules in `UnitTestManager.deployApplication`: https://github.com/caskdata/cdap/blob/develop/cdap-unit-test/src/main/java/co/cask/cdap/test/UnitTestManager.java#L83

anew:
Ideally, the schedules never get enabled. If we enable them first and then disable them, there is a chance that a schedule fires before the it can be disabled. In that case, we get intermittent unit test failures.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
jguery: You're right, will do.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
```diff
@@ -608,9 +608,11 @@ public void changeFlowletStreamConnection(HttpRequest request, HttpResponder res
   @PUT
   @Path("/apps/{app-id}")
   public BodyConsumer deploy(HttpRequest request, HttpResponder responder, @PathParam("app-id") final String appId,
+                             @QueryParam(Constants.Scheduler.SUSPEND_SCHEDULES_DEPLOY_OPTION)
```
anew: we normally don't use constants for query parameters. It is easier to see what an actual request looks like if you put the literal string here. See other places in this class.

anew:
I know we discussed this before... seeing how much change this requires, I am wondering whether it is the right approach. Alvin's suggestion may work in almost all cases. 

Or, seeing that you already have a property for the scheduler itself, (isLazyStart) can't you just use that to determine whether a new schedule should active or suspended? We can simply inject isLazyStart=true in  the test base and all the API changes are not needed, right?

jguery:
@anew yes I can try that approach, it should be less invasive.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
jguery: Ok, it will be removed with the other approach anyway.

jguery:
@anew I implemented the approach you talked about. Please have a look. Thank you.

anew:
change looks good. One thing I am not sure of is, if a test actually needs schedules, and calls resumeSchedule for one of its schedules, will that start the scheduler? I may have missed it, but I think that requires that the resume method calls lazyStart(), right?

jguery:
Yes @anew, AbstractSchedulerService.resume does call the lazyStart() method for the right scheduler: time scheduler or stream size scheduler.

anew:
Oh yes, I see it now. 
LGTM

PR #1646, opened 2015-03-16
CDAP-1812: Fix column string truncation when string is only 1 longer than column width

jwang47:
https://issues.cask.co/browse/CDAP-1812
http://builds.cask.co/browse/CDAP-RBT158


cdap-cli/src/main/java/co/cask/cdap/cli/util/table/AltStyleTableRenderer.java
```diff
@@ -69,7 +69,7 @@
  */
 public class AltStyleTableRenderer implements TableRenderer {
 
-  private static final int DEFAULT_WIDTH = 80;
+  public static final int DEFAULT_WIDTH = 80;
```
anwar6953: why public? seems like you're inlining 80 in tests.

anew: Actually you should use this constants in the tests. And it can be package visible for the test with @VisibleForTesting

jwang47: I opened another PR for making line width a property of `CLIConfig`, which addresses this. https://github.com/caskdata/cdap/pull/1656

sreevatsanraman:
LGTM :+1:

jwang47:
Going to get https://github.com/caskdata/cdap/pull/1656 merged first.

PR #1647, opened 2015-03-16
CDAP-1750: Use different port in CLIMainTest

jwang47:
https://issues.cask.co/browse/CDAP-1750
http://builds.cask.co/browse/CDAP-RBT159


sreevatsanraman:
LGTM :+1:

PR #1650, opened 2015-03-16
(CDAP-1604) Don't enable explore on system datasets

ParthGandhi:
https://issues.cask.co/browse/CDAP-1604

### DIFF
--- a/cdap-explore-client/src/main/java/co/cask/cdap/explore/client/ExploreFacade.java
+++ b/cdap-explore-client/src/main/java/co/cask/cdap/explore/client/ExploreFacade.java
@@ -85,7 +85,7 @@ public class ExploreFacade {
    * @param datasetInstance dataset instance id.
    */
   public void enableExploreDataset(Id.DatasetInstance datasetInstance) throws ExploreException, SQLException {
-    if (!exploreEnabled) {
+    if (!(exploreEnabled && isDatasetExplorable(datasetInstance))) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


ParthGandhi:
http://builds.cask.co/browse/CDAP-DUT1129-1/

cdap-explore-client/src/main/java/co/cask/cdap/explore/client/ExploreFacade.java
```diff
@@ -144,6 +144,10 @@ public void removeNamespace(Id.Namespace namespace) throws ExploreException, SQL
     handleExploreFuture(futureSuccess, "remove", "namespace", namespace.getId());
   }
 
+  private boolean isDatasetExplorable(Id.DatasetInstance datasetInstance) {
```
bdmogal: nit: The contract of this method seems to indicate that it operates on a dataset and returns if it is explorable. To check if explore is enabled in CDAP as a whole seems to be outside of the purview of this method. 

Only a nit though.

ParthGandhi: Good point. Fixed.

ParthGandhi:
New build: http://builds.cask.co/browse/CDAP-DUT1129-2/commit

bdmogal:
:+1: Wish there was a way to add a unit test for this, I can't think of anything right now, maybe we can add something later. Let's just keep an eye open for the stack trace again.

PR #1652, opened 2015-03-16
Changes to support backwards compatability for a flow's groupId (for str...

anwar6953:
...eams/queues).

In FlowUtils#generateConsumerGroupId, use 'developer' instead of the program's namespace if the namespace is 'default'.

Data migration is less feasible because the groupId is used in the queue data tables, which can be quite large.

Helps to resolve: https://issues.cask.co/browse/CDAP-1823
Build: http://builds.cask.co/browse/CDAP-RBT161-2

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowUtils.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowUtils.java
@@ -21,6 +21,7 @@ import co.cask.cdap.api.flow.FlowletDefinition;
 import co.cask.cdap.app.program.Program;
 import co.cask.cdap.app.queue.QueueSpecification;
 import co.cask.cdap.app.queue.QueueSpecificationGenerator;
+import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.queue.QueueName;
 import co.cask.cdap.data2.transaction.ForwardingTransactionAware;
 import co.cask.cdap.data2.transaction.Transactions;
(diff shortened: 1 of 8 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowUtils.java
```diff
@@ -68,8 +69,13 @@ public static long generateConsumerGroupId(Program program, String flowletId) {
...
    */
   public static long generateConsumerGroupId(Id.Program program, String flowletId) {
+    // Use 'developer' in place of a program's namespace for programs in the 'default' namespace
+    // to support backwards compatibility for queues and streams.
+    String namespace = program.getNamespaceId();
+    String backwardsCompatibleNamespace =
+      Constants.DEFAULT_NAMESPACE.equals(namespace) ? Constants.DEVELOPER_ACCOUNT : namespace;
     return Hashing.md5().newHasher()
-                  .putString(program.getNamespaceId())
+                  .putString(backwardsCompatibleNamespace)
```
jwang47: So if someone creates a "developer" namespace and deploys an app with a flow, then that app could be reading from a stream already created in the "default" namespace?

anwar6953: namespace is not required in the groupId hash, since the queue/stream tables are created in separate namespaces. (The table name includes the namespace).

The hash only includes namespace ('developer' for default namespace) for backwards compatibility).

jwang47:
LGTM

PR #1653, opened 2015-03-16
Feature/cdap 1823 stream state store upgrade

anwar6953:
Depends on: https://github.com/caskdata/cdap/pull/1652

This follows a similar pattern as queue config table upgrade (row key migration), but I have kept this separate/independent, because queue config table migration may need to be changed to go through dataset APIs instead of HBase Table APIs.

JIRA: https://issues.cask.co/browse/CDAP-1823
Build: http://builds.cask.co/browse/CDAP-RBT162-3

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/StreamStateStoreUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/StreamStateStoreUpgrader.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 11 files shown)


cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
```diff
@@ -427,7 +427,8 @@ private Stream(final Namespace namespace, final String streamName) {
       Preconditions.checkNotNull(streamName, "Stream name cannot be null.");
 
       Preconditions.checkArgument(isId(streamName),
-                                  "Stream name can only contains alphanumeric, '-' and '_' characters only.");
+                                  String.format("Stream name can only contains alphanumeric, " +
```
bdmogal: Not introduced in your PR, but s/contains/contain/, also there is an extra 'only' in the end.

cdap-master/src/main/java/co/cask/cdap/data/tools/StreamStateStoreUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.NavigableMap;
+import javax.annotation.Nullable;
+
+/**
+ * Upgrades stream state store table
+ */
+public class StreamStateStoreUpgrader extends AbstractUpgrader {
```
bdmogal: Is there no way of extracting at least parts of the code from here (and ``QueueConfigUpgrader``) to an abstract base class? 

Seems like a fair bit of repetition?

bdmogal:
:+1:

PR #1654, opened 2015-03-16
removed unwanted binding and added test case for testing injector

gsps1:
http://builds.cask.co/browse/CDAP-RBT163-1


rohitsinha54:
LGTM assuming the build passes

gsps1:
green, merging

PR #1655, opened 2015-03-16
[CDAP 1700] forced upgrade datasets by default

anew:


cdap-unit-test/src/main/java/co/cask/cdap/test/ConfigurableTestBase.java
```diff
@@ -0,0 +1,534 @@
+/*
```
anew: This is identical with the previous TestBase, except that init() is now called initTestBase(), has a parameter to pass in additional configuration, and is not a @BeforeClass. This will now be called from @BeforeClass TestBase.init(), such that TestBase behaves exactly as before. But we can have other test cases extend ConfigurableTestBase and pass in additional properties.

cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetUpgradeEnabledTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ *
+ */
+@Category(SlowTests.class)
+public class DatasetUpgradeEnabledTest extends ConfigurableTestBase {
+
+  @BeforeClass
+  public static void init() throws Exception {
+    initTestBase(ImmutableMap.of(Constants.Dataset.DATASET_UNCHECKED_UPGRADE, Boolean.TRUE.toString()));
```
anew: this replaces the previously needed cdap-site.xml in the test resources.

cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetUpgradeEnabledTest.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Category(XSlowTests.class)
+  @Test
+  public void testDatasetUncheckedUpgrade() throws Exception {
```
anew: This is the original test moved from TestBaseTestRun

cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetUpgradeDisabledTest.java
```diff
@@ -0,0 +1,65 @@
...
+public class DatasetUpgradeDisabledTest extends ConfigurableTestBase {
+
+  @BeforeClass
+  public static void init() throws Exception {
+    initTestBase(ImmutableMap.of(Constants.Dataset.DATASET_UNCHECKED_UPGRADE, Boolean.FALSE.toString()));
+  }
+
+  @Category(XSlowTests.class)
+  @Test
+  public void testDatasetUncheckedUpgrade() throws Exception {
```
anew: new test to verify that upgrade can be turned off.

anew:
asking @ghelmling because @poornachandra has urgent customer support to do.

anew:
build succeeded http://builds.cask.co/browse/CDAP-RBT164-2

anew:
Closing this and reopening against develop.

PR #1656, opened 2015-03-16
CDAP-1830: Use terminal width as max allowable line width in CLI

jwang47:
`AltStyleTableRenderer` was not rendering tables with correct outer width, so this PR fixes that bug. Also, this PR makes `AltStyleTableRenderer` and the output from `help` use up the entire terminal width.

https://issues.cask.co/browse/CDAP-1830
http://builds.cask.co/browse/CDAP-RBT165


sreevatsanraman:
changes LGTM :+1:  - merge once the test passes.

PR #1661, opened 2015-03-17
[CDAP-1831] Fix explore on hive 0.10 by using a 'use db' statement rathe...

bdmogal:
...r than SessionState.setCurrentDatabase which is unavailable in hive 0.10

Jira: [CDAP-1831](https://issues.cask.co/browse/CDAP-1831)
Build: http://builds.cask.co/browse/CDAP-RBT167-1

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/HiveCDH4ExploreService.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/HiveCDH4ExploreService.java
@@ -19,6 +19,7 @@ package co.cask.cdap.explore.service.hive;
 import co.cask.cdap.app.store.StoreFactory;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.utils.Tasks;
 import co.cask.cdap.data2.dataset2.DatasetFramework;
 import co.cask.cdap.data2.transaction.stream.StreamAdmin;
 import co.cask.cdap.explore.service.ExploreException;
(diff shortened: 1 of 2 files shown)


anew:
Is there a unit test?
Other than that LGTM

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/HiveCDH4ExploreService.java
```diff
@@ -105,6 +109,22 @@ protected OperationHandle doExecute(SessionHandle sessionHandle, String statemen
   }
 
   @Override
+  protected void setCurrentDatabase(String dbName) throws IOException, HiveSQLException, ExploreException {
```
ParthGandhi: Just make this `throws Exception`

ParthGandhi:
One nit for uniformity. Otherwise LGTM.

@anew Yes, there are unit tests for this.

bdmogal:
We don't seem to have tests that use hive 0.10. There are tests that would call the ``setCurrentDatabaseMethod``. But none  that use hive 0.10. We should probably have a way of testing hive 0.10, since we support it.

bdmogal:
Maybe a much larger effort though?

bdmogal:
However, the way this problem was discovered was we ran audi on a CDH 4.3 cluster that has hive 0.10. Maybe we should make that part of our regular audi runs as well.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/HiveCDH4ExploreService.java
ParthGandhi: Ignore ^

PR #1663, opened 2015-03-17
support delete metrics in V3 api

gsps1:
https://issues.cask.co/browse/CDAP-1729

Build running : http://builds.cask.co/browse/CDAP-RBT169-1


cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -168,6 +170,31 @@ public void query(HttpRequest request, HttpResponder responder,
     }
   }
 
+  @POST
+  @Path("/delete")
+  public void delete(HttpRequest request, HttpResponder responder,
+                    @QueryParam("context") String context,
+                    @QueryParam("metric") String metric,
+                    @QueryParam("start") @DefaultValue("0") final long startTs,
+                    @QueryParam("end") @DefaultValue("2147483647") final long endTs) throws Exception {
```
sreevatsanraman: Why is default endtime arbitrarily ending on 19th Jan 2038? Can it be Long.MAX_VALUE ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -168,6 +170,31 @@ public void query(HttpRequest request, HttpResponder responder,
     }
   }
 
+  @POST
```
bdmogal: Shouldn't this be a ``DELETE``? Also, perhaps, we should send out an API review?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -168,6 +170,31 @@ public void query(HttpRequest request, HttpResponder responder,
...
+    try {
+      if (context == null) {
+        responder.sendString(HttpResponseStatus.BAD_REQUEST, "Please Specify context to delete");
+        return;
+      }
+      Map<String, String> tagsSliceBy = humanToTagNames(parseTagValuesAsMap(context));
+      MetricDeleteQuery query = new MetricDeleteQuery(startTs, endTs, metric, tagsSliceBy);
+      metricStore.delete(query);
+      responder.sendStatus(HttpResponseStatus.OK);
+    } catch (IllegalArgumentException e) {
```
bdmogal: Where does this get thrown? Better to respond with ``BAD_REQUEST`` only for known ``IllegalArgumentException``s and not all?

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricDeleteQuery.java
```diff
@@ -37,7 +38,7 @@ public MetricDeleteQuery(long startTs, long endTs, String metricName,
     this.startTs = startTs;
     this.endTs = endTs;
     this.metricName = metricName;
-    this.sliceByTagValues = ImmutableMap.copyOf(sliceByTagValues);
+    this.sliceByTagValues = Maps.newHashMap(sliceByTagValues);
```
bdmogal: why did this have to change?

bdmogal:
Just a few minor comments. Also, this is not hooked up to namespace delete yet, right? Will that be a separate PR?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
gsps1: this is Integer.MAX_VALUE, have to supply default value as constant. we cannot use Long.MAX_VALUE as we use only 4 bytes for time stamp right now.

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricDeleteQuery.java
gsps1: when you want to delete all apps in a namespace, you would specify , /v3/metrics/delete?context=namespace.somespace.app.* , the app tag will map to null value, and ImmutableMap doesn't allow null in key or value, that's why you need this change.

gsps1: this also answers your question, this is not hooked up to namespace delete yet, right?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
abaranau: add a constant to a class. Or do not provide default value with this annotation - set default in the method code

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
abaranau: Can we use MetricStore interface to delete metrics on delete of namespace? We don't want to expose REST API just yet

abaranau:
please see comments above

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -193,6 +199,15 @@ public void deleteNamespace(final Id.Namespace namespaceId)
   }
 
   @Override
+  public void deleteMetrics(Id.Namespace namespaceId) throws Exception {
+    long endTs = System.currentTimeMillis() / 1000;
+    Map<String, String> tags = Maps.newHashMap();
+    tags.put(Constants.Metrics.Tag.NAMESPACE, namespaceId.getId());
+    MetricDeleteQuery deleteQuery = new MetricDeleteQuery(0, endTs, null, tags);
```
abaranau: we should delete only from "totals", see https://issues.cask.co/browse/CDAP-1857

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/NamespaceAdmin.java
```diff
@@ -72,6 +72,14 @@
   public void deleteNamespace(Id.Namespace namespaceId) throws NotFoundException, NamespaceCannotBeDeletedException;
 
   /**
+   * Deletes the metrics in specified namespace
+   *
+   * @param namespaceId the {@link Id.Namespace} of the specified namespace
+   * @throws Exception if there are issues with metrics deletion
+   */
+  public void deleteMetrics(Id.Namespace namespaceId) throws Exception;
```
abaranau: should be called inside deleteNamespace method. No need to expose it here

abaranau:
one comment on NameSpaceAdmin API. The other one (delete only totals) - feel free to do in separate PR

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricHttpHandler.java
```diff
@@ -1171,8 +1171,7 @@ public void resetCDAP(HttpRequest request, HttpResponder responder) {
       // remove all data in console settings
       consoleSettingsStore.delete();
 
-      // todo: do efficiently and also remove timeseries metrics as well: CDAP-1125
-      appLifecycleHttpHandler.deleteMetrics(Constants.DEFAULT_NAMESPACE, null);
+      namespaceAdmin.deleteMetrics(Constants.DEFAULT_NAMESPACE_ID);
```
bdmogal: like @abaranau's comment below, this should go away. deleteMetrics should be part of ``namespaceAdmin.deleteNamespace(Constants.DEFAULT_NAMESPACE_ID)``

abaranau:
very minor (no need to rerun bamboo test after fix) - fix and merge

gsps1:
build - green and fixed comments. merging.

PR #1665, opened 2015-03-17
Feature/cdap 1700 restructure tests for unchecked dataset upgrade

anew:
this is identical to https://github.com/caskdata/cdap/pull/1655, please see my comments there.


anew:
Reopened this because it adds a test for en existing feature. Most of the changes are refactoring of the test base to allow injection of configuration by individual tests. Plus one new test case that tests the unchecked dataset upgrade introduced in 2.6.1.

cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetUpgradeDisabledTest.java
```diff
@@ -0,0 +1,65 @@
...
+      getDataset(DatasetUncheckedUpgradeApp.DATASET_NAME);
+    DatasetUncheckedUpgradeApp.Record expectedRecord = new DatasetUncheckedUpgradeApp.Record("0AXB", "john", "doe");
+    datasetManager.get().writeRecord("key", expectedRecord);
+    datasetManager.flush();
+
+    DatasetUncheckedUpgradeApp.Record actualRecord =
+      (DatasetUncheckedUpgradeApp.Record) datasetManager.get().getRecord("key");
+    Assert.assertEquals(expectedRecord, actualRecord);
+
+    // Test compatible upgrade
```
poornachandra: Comment should be "Test in-compatible upgrade"

poornachandra:
Just one minor comment, otherwise LGTM.

sreevatsanraman:
Build running here: http://builds.cask.co/browse/CDAP-DUT1376

anew:
Looks like the build failed. One failure seems unrelated (a scheduler error), but one is a class loading exception that appears related. Strange enough it succeeded back when I ran this against a 2.8 branch... @sree, thanks for taking it on, but let me figure out what is wrong with it, it seems that something else changed since then that has to be accounted for.

anew:
This was the old build that succeeded: http://builds.cask.co/browse/CDAP-RBT164-2

anew:
Interesting. Tow issues:
1. The original PR added a way to configure the TestBase by moving all initialization code to a ConfigurableTestBase and calling its initTestBase() from TestBase's @BeforeClass. The TestBase had quite a few changes since 2.8, and so the code that was moved to ConfigurableTestBase has diverged. 
2. The original PR moved one test (the one that tests unchecked dataset upgrade) to a new test class that enables dataset upgrade by calling ConfigurableTestBase with a configuration parameter. Somehow that method was copied back into TestBaseTestRun, where it had to fail... must be the git conflict resolution. 

To fix this PR, I rebased ConfigurableTestBase on current develop's TestBase, and removed the extra test. It passes on my laptop now. I will run a bamboo build. In the meantime, please have another look, @sreevatsanraman  and @poornachandra

anew:
New build running here: http://builds.cask.co/browse/CDAP-DUT1376-2

poornachandra:
LGTM

anew:
I will wait for the build to succeed before I merge

PR #1666, opened 2015-03-17
Fix documentation build.

awholegunch:
Fix the Build to use curl to download example documentation and set the version based on the release.

It doesn't depend on the example documentation to be locally available; instead, it downloads it from GitHub.


duddun:
LGTM

PR #1669, opened 2015-03-17
CDAP-1841 Dockerfile fixes

wolf31o2:
These two directories no longer exist. Tested on a `docker-base` single node spun up via Coopr in GCE.


sreevatsanraman:
LGTM :+1:

PR #1674, opened 2015-03-17
Hotfix/remove use datasets

gokulavasan:
Remove/deprecate the use of useDatasets method in MR and Workers

JIRA: https://issues.cask.co/browse/CDAP-1845

Bamboo : http://builds.cask.co/browse/CDAP-DUT1116

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
@@ -91,7 +91,7 @@ public abstract class AbstractMapReduce implements MapReduce {
 
   /**
    * Specifies set of Dataset names that are used by the {@link MapReduce}.
-   * @deprecated datasets used in runtime need not be specified during configuration
+   * @deprecated datasets used in runtime need not be specified in {@link MapReduce#configure}
    */
   @Deprecated
   protected final void useDatasets(String dataset, String...moreDatasets) {
@@ -103,7 +103,7 @@ public abstract class AbstractMapReduce implements MapReduce {
 
   /**
    * Specifies set of Dataset names that are used by the {@link MapReduce}.
-   * @deprecated datasets used in runtime need not be specified during configuration
+   * @deprecated datasets used in runtime need not be specified in {@link MapReduce#configure}
    */
   @Deprecated
   protected final void useDatasets(Iterable<String> datasets) {
--- a/cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
@@ -91,7 +91,9 @@ public abstract class AbstractMapReduce implements MapReduce {
 
   /**
    * Specifies set of Dataset names that are used by the {@link MapReduce}.
+   * @deprecated datasets used in runtime need not be specified during configuration
    */
+  @Deprecated
   protected final void useDatasets(String dataset, String...moreDatasets) {
     List<String> datasets = new ArrayList<String>();
     datasets.add(dataset);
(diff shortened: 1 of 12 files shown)


cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
```diff
@@ -91,7 +91,9 @@ protected final void setReducerResources(Resources resources) {
 
   /**
    * Specifies set of Dataset names that are used by the {@link MapReduce}.
+   * @deprecated datasets used in runtime need not be specified during configuration
```
ParthGandhi: `... specified during MapReduce#configure()` i.e. link to method.

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
```diff
@@ -91,7 +91,9 @@ protected final void setReducerResources(Resources resources) {
...
    */
+  @Deprecated
   protected final void useDatasets(String dataset, String...moreDatasets) {
     List<String> datasets = new ArrayList<String>();
     datasets.add(dataset);
```
ParthGandhi: Should we make this operation a no-op then? Why maintain a list?

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
```diff
@@ -101,7 +103,9 @@ protected final void useDatasets(String dataset, String...moreDatasets) {
 
   /**
    * Specifies set of Dataset names that are used by the {@link MapReduce}.
+   * @deprecated datasets used in runtime need not be specified during configuration
```
ParthGandhi: Ditto.

cdap-api/src/main/java/co/cask/cdap/api/worker/WorkerConfigurer.java
```diff
@@ -58,6 +58,8 @@
   /**
    * Adds the names of {@link Dataset Datasets} used by the worker.
    * @param datasets dataset names
+   * @deprecated datasets being used during runtime need not be specified during configuration
    */
+  @Deprecated
```
ParthGandhi: Ditto for above 2 comments.

ParthGandhi:
A couple of comments. Rest LGTM.

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
gokulavasan: We have marked it as deprecated this release. So if the user is still using useDatasets, he can get the info about MR, Workers associated with the dataset. We can remove this logic and the method completely in 3.0.

gokulavasan:
@ParthGandhi Addressed your comments. Please take a look. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/AbstractMapReduce.java
ParthGandhi: Yes, makes sense.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricDataHttpHandler.java
```diff
@@ -160,4 +148,16 @@ public void getMapReduceByDataset(HttpRequest request, HttpResponder responder,
     programListByDataAccess(responder, store, dsFramework, ProgramType.MAPREDUCE, Data.DATASET,
                             namespaceId, datasetId);
   }
+
```
ParthGandhi: Can you move this back to where it used to be? It avoids an unnecessary diff.

ParthGandhi:
Just one comment. The rest looks good.

gokulavasan:
@ParthGandhi  Addressed your comment. Merging.

PR #1675, opened 2015-03-17
CDAP-1811: Fix ScheduleClient in deserializing Schedule

jwang47:
* Added scheduleType to ScheduleSpecification
* Custom GSON deserializer for ScheduleSpecification in ScheduleClient

https://issues.cask.co/browse/CDAP-1811
http://builds.cask.co/browse/CDAP-RBT174


cdap-api/src/main/java/co/cask/cdap/api/schedule/ScheduleSpecification.java
```diff
@@ -27,16 +27,22 @@
  */
 public final class ScheduleSpecification {
   private final Schedule schedule;
+  private final String scheduleType;
```
jguery: Why would you need that extra argument? If you actually look at how ScheduleSpecification is serialized in JSON from CDAP - in ``ScheduleSpecificationCodec`` - a scheduleType param is already added to the json. You could actually reuse this SerDe, instead of creating your own ``ScheduleSpecificationDeserializer``.
Also if you add an attribute here, isn't it going to mess with existing app specifications present in CDAP?

cdap-client/src/main/java/co/cask/cdap/client/ScheduleClient.java
```diff
@@ -100,4 +115,32 @@ public String getStatus(String appId, String scheduleId)
...
+      JsonObject object = json.getAsJsonObject();
+      JsonObject scheduleJson = object.getAsJsonObject("schedule");
+      String scheduleType = object.get("scheduleType").getAsString();
+
+      Schedule schedule;
+      try {
+        Class<? extends Schedule> scheduleClass = (Class<? extends Schedule>) Class.forName(scheduleType);
+        schedule = context.deserialize(scheduleJson, scheduleClass);
+      } catch (Exception e) {
+        throw new JsonParseException("Unknown scheduleType: " + scheduleType, e);
```
anwar6953: this is the proper error message for line 133, but not 134, right?

cdap-api/src/main/java/co/cask/cdap/api/schedule/ScheduleSpecification.java
jwang47: Oops. Didn't know there was `ScheduleSpecificationCodec` already. Will use that.

anwar6953:
didn't remove: old file: AbstractSpecificationCodec.java?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -968,7 +972,7 @@ public void getWorkflowSchedules(HttpRequest request, HttpResponder responder,
         specList.add(entry.getValue());
       }
     }
-    responder.sendJson(HttpResponseStatus.OK, specList);
+    responder.sendJson(HttpResponseStatus.OK, GSON.toJsonTree(specList));
```
jguery: Do we need that change? Won't responder.sendJson() use the right codec by itself?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/ApplicationSpecificationAdapter.java
```diff
@@ -34,6 +34,7 @@
 import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.internal.io.SchemaGenerator;
 import co.cask.cdap.internal.io.SchemaTypeAdapter;
+import co.cask.cdap.proto.codec.ScheduleSpecificationCodec;
```
jguery: What for?

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -23,7 +23,7 @@
 import co.cask.cdap.gateway.handlers.AppFabricHttpHandler;
 import co.cask.cdap.gateway.handlers.ServiceHttpHandler;
 import co.cask.cdap.internal.app.BufferFileInputStream;
-import co.cask.cdap.internal.app.ScheduleSpecificationCodec;
+import co.cask.cdap.proto.codec.ScheduleSpecificationCodec;
```
jguery: Do you use it anywhere?

jguery: Nevermind, I got it.

cdap-proto/src/main/java/co/cask/cdap/proto/codec/AbstractSpecificationCodec.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright © 2014 Cask Data, Inc.
```
jguery: 2015?

cdap-proto/src/main/java/co/cask/cdap/proto/codec/ScheduleSpecificationCodec.java
```diff
@@ -14,7 +14,7 @@
  * the License.
  */
 
-package co.cask.cdap.internal.app;
+package co.cask.cdap.proto.codec;
```
jguery: If we move one codec there, do we want to move all ``*Codec`` classes there? Please check with @chtyim or @anew

cdap-proto/src/main/java/co/cask/cdap/proto/codec/AbstractSpecificationCodec.java
jwang47: Copied from cdap-app-fabric. Couldn't simply move b/c it's also used for other codecs, but I didn't want to move all the codecs over since I don't know what the impact is.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
jwang47: Nope. netty-http's `HttpResponder` uses its own gson instance (created using `new Gson()`) which doesn't register our codecs, so we need to use our own gson instance.

anwar6953: You can simply pass in a GSON instance, right (as the fourth parameter)

jwang47: Oh, yeah that's true. Didn't see it for some reason.

jguery:
So you basically moved the codecs to proto and added a test in cdap-client to make sure we get the schedules' properties, is that right? If that's all, then LGTM

jwang47:
That's right.

PR #1678, opened 2015-03-17
CDAP-1843 fix NaN issue on start time after upgrade

rohitsinha54:
The start time is just a part of the key in runRecordCompleted so get it from there 

Issue: https://issues.cask.co/browse/CDAP-1843
Build: http://builds.cask.co/browse/CDAP-RBT172-1

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
@@ -263,7 +263,7 @@ public class MDSUpgrader extends AbstractUpgrader {
   private void writeTempRunRecordStart(String appId, ProgramType programType, String programId, String pId,
                                        long startTs) {
     store.setStart(Id.Program.from(Id.Application.from(Constants.DEFAULT_NAMESPACE, appId), programType, programId),
-                   pId, Long.MAX_VALUE - startTs);
+                   pId, startTs);
   }
 
   /**


cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -235,15 +236,16 @@ private void handleRunRecordCompleted(final String appId, final String programId
...
       @Override
       public Void apply(AppMDS appMetaStore) throws Exception {
-        List<RunRecord> completedRunRecords = appMetaStore.mds.list(partialKey, RunRecord.class);
-        for (RunRecord curCompletedRunRecord : completedRunRecords) {
+        Map<MDSKey, RunRecord> completedRunRecords = appMetaStore.mds.listKV(partialKey, RunRecord.class);
+        for (Map.Entry<MDSKey, RunRecord> curCompletedRunRecord : completedRunRecords.entrySet()) {
           // we cannot add a runRecordCompleted if there is no runRecordStarted for the run so write it first
           // the next setStop call deletes the runRecordStarted for that run so we don't have to delete it here.
-          writeTempRunRecordStart(appId, programType, programId, curCompletedRunRecord.getPid(),
-                                  curCompletedRunRecord.getStartTs());
```
anwar6953: what did .getStartTs() return?

rohitsinha54: We don't put anything when we write stop record. The function is there to get start time for RunRecord of started.

cdap-master/src/main/java/co/cask/cdap/data/tools/MDSUpgrader.java
```diff
@@ -263,7 +263,7 @@ public Void apply(AppMDS appMetaStore) throws Exception {
   private void writeTempRunRecordStart(String appId, ProgramType programType, String programId, String pId,
                                        long startTs) {
     store.setStart(Id.Program.from(Id.Application.from(Constants.DEFAULT_NAMESPACE, appId), programType, programId),
-                   pId, Long.MAX_VALUE - startTs);
```
rohitsinha54: Do not invert  the start time as the one store in completed run record is not inverted in the first place

bdmogal:
:+1:

PR #1681, opened 2015-03-17
Feature/etl template api

gokulavasan:
Initial checkin of ETL App Template API.

JIRAs : 

https://issues.cask.co/browse/CDAP-1774
https://issues.cask.co/browse/CDAP-1775

Bamboo : http://builds.cask.co/browse/CDAP-DT0


cdap-app-templates/.gitignore
```diff
@@ -0,0 +1,21 @@
+*.class
```
chtyim: Why need a separate .gitignore?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/AbstractTransform.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package co.cask.cdap.templates.etl.api;
+
+/**
+ * Transform Stage
+ * @param <I> input
+ * @param <O> output
+ */
+public abstract class AbstractTransform<I, O> implements StageLifecycle {
```
chtyim: Just call it `Transform` (or `Transformer` ?)

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/SinkContext.java
```diff
@@ -0,0 +1,27 @@
...
+
+package co.cask.cdap.templates.etl.api;
+
+import co.cask.cdap.api.data.DatasetContext;
+import co.cask.cdap.api.data.stream.StreamWriter;
+
+/**
+ * Context passed to the Sink stages.
+ */
+public interface SinkContext extends SourceContext, StreamWriter, DatasetContext {
```
chtyim: Why `SinkContext` extends `SourceContext`??

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/SourceContext.java
```diff
@@ -0,0 +1,24 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.api;
+
+/**
+ * Context passed to the Source stages.
+ */
+public interface SourceContext extends TransformContext {
```
chtyim: Why extends?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/StageConfigurer.java
```diff
@@ -0,0 +1,43 @@
...
+   * Set the description of the Stage.
+   * @param description description
+   */
+  void setDescription(String description);
+
+  /**
+   * Set the properties this stage requires during runtime.
+   * @param properties {@link Set} of properties
+   */
+  void setReqdProperties(Set<String> properties);
```
chtyim: The method name is awkward. Don't use uncommon shorthand.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/StageLifecycle.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Provides lifecycle methods for ETL Pipeline Stages.
+ */
+public interface StageLifecycle {
+
+  /**
+   * Destroy the Pipeline Stage.
+   */
+  void destroy();
```
chtyim: How lifecycle only have destroy? Shouldn't it at least should have initialize as well?

gokulavasan:
@chtyim Thanks for the review. Addressed your comments. Please take a look again.

cdap-app-templates/cdap-etl/cdap-etl-api/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    <version>3.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>cdap-etl-api</artifactId>
+  <name>CDAP ETL App Template API</name>
+  <version>3.0.0-SNAPSHOT</version>
+  <packaging>jar</packaging>
+
+  <dependencies>
+    <dependency>
```
chtyim: So the etl-api module doesn't have dependency on cdap-api??

gokulavasan: Changed. Had the dependency in the cdap-etl parent pom.

gokulavasan:
@chtyim Take a look again. Thanks.

gokulavasan:
@chtyim Removed MR related dependency from etl api. Please review. Thanks.

cdap-app-templates/cdap-etl/cdap-etl-api/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-api</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
```
albertshau: should have newline at end of file

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
```diff
@@ -0,0 +1,59 @@
...
+ * Transform Stage
+ * @param <I> input
+ * @param <O> output
+ */
+public abstract class Transform<I, O> implements StageLifecycle {
+
+  private TransformContext context;
+
+  /**
+   * Configure the Transform stage.
```
albertshau: we should add more javadocs describing when this is called.  For somebody extending this class, the difference between configure and initialize should be clear.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/TransformContext.java
```diff
@@ -0,0 +1,37 @@
...
+
+import co.cask.cdap.api.RuntimeContext;
+
+/**
+ * Context passed to the Pipeline Stages.
+ */
+public interface TransformContext extends RuntimeContext {
+
+  /**
+   * Get Instance Id.
```
albertshau: how were you thinking of using these methods?  Seems weird to me to do instance specific things in a transform

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Configure the Sink.
+   * @param configurer {@link StageConfigurer}
+   */
+  public void configure(StageConfigurer configurer) {
+    configurer.setName(this.getClass().getSimpleName());
+  }
+
+  /**
+   * Prepare the Batch Job.
```
albertshau: the difference between prepareJob and initialize is not clear to me.  I assume this gets called when we're preparing the mapreduce job? What would you do in initialize that would not be done here?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSource.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Setup configuration related to the Batch Source.
+   * @param context {@link BatchContext}
+   */
+  public abstract void prepareJob(BatchContext context);
+
+  /**
+   * Initialize the Batch Source.
```
albertshau: similar question about initialize and prepareJob

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
```diff
@@ -0,0 +1,105 @@
...
+   * job is supposed to start if this job is started by the scheduler. Otherwise it would be the current time when the
+   * job runs.
+   *
+   * @return Time in milliseconds since epoch time (00:00:00 January 1, 1970 UTC).
+   */
+  long getLogicalStartTime();
+
+  /**
+   */
+  <T> T getHadoopJob();
```
albertshau: i think you can have this return the actual Job. I believe the reason MapReduceContext does this is to avoid bringing in mapreduce as a dependency into api.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Write the given object.
+   * @param object object to be written
+   * @param writer Writer to persist data to Batch Output
+   * @throws IOException
+   * @throws InterruptedException
+   */
+  public abstract void write(I object, BatchWriter<KEY, VALUE> writer) throws IOException, InterruptedException;
```
albertshau: how will the etl app know what writer to pass in here? For example, say I'm want to write an HBase sink and a FileSink.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSource.java
```diff
@@ -0,0 +1,79 @@
...
+    this.context = context;
+  }
+
+  /**
+   * Process data.
+   * @param key Key class from Input
+   * @param value Value class from Input
+   * @param data Emit data
+   */
+  public abstract void combine(KEY key, VALUE value, Emitter<O> data);
```
albertshau: a combiner has a specific meaning in mapreduce, so we should rename this.  Emit? Extract?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/realtime/AbstractRealtimeSource.java
```diff
@@ -0,0 +1,78 @@
...
+   * Invoked when source is suspended.
+   */
+  public abstract void onSuspend();
+
+  /**
+   * Resume/reconfigure from the state of suspension.
+   * @param oldInstance old instance count
+   * @param newInstance new instance count
+   */
+  public abstract void onResume(int oldInstance, int newInstance);
```
albertshau: oldInstanceCount may be a better name, as oldInstance could be interpreted as the old instance id.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
albertshau: or does this have the same issue?

cdap-app-templates/cdap-etl/cdap-etl-core/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-etl-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
+
+</project>
```
albertshau: newline

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+      <groupId>org.twitter4j</groupId>
+      <artifactId>twitter4j-stream</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
```
albertshau: newline

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+      <groupId>org.twitter4j</groupId>
+      <artifactId>twitter4j-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.twitter4j</groupId>
+      <artifactId>twitter4j-stream</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
```
albertshau: how come we need this?  Seems like we should just work with jdbc rather than mysql specifically.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
gokulavasan: prepareJob is called before the MR job is submitted. And so primarily setting properties in the hadoop conf. Initialize is called in the MR job in the setup. I will add this to the javadocs.

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
albertshau: in any case, probably best to leave these out and add them as needed

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
gokulavasan: Same reason applies here as well. We don't want hadoop dependencies in etl-api.

cdap-app-templates/cdap-etl/cdap-etl-test/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+      <artifactId>cdap-etl-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-unit-test</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
```
albertshau: newline

cdap-app-templates/cdap-etl/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+      </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>2.3.7</version>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
```
albertshau: newline

cdap-app-templates/cdap-etl/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+      </dependency>
+
+      <dependency>
+        <groupId>javax.ws.rs</groupId>
+        <artifactId>javax.ws.rs-api</artifactId>
+        <version>${rs-api.version}</version>
+      </dependency>
+
+      <!-- Twitter4j dependencies -->
+      <dependency>
```
albertshau: lets leave out until implemented

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
gokulavasan: This is used to just collect the key, value objects. The etl mr driver will finally use the mapper context to write the <key, value> through Mapper.Conext#write method. 

In case of say KeyValueTableSink, they key, value are expected to be byte[], byte[]. Its up to the BatchSink to decide what these are classes are.

cdap-api/src/main/java/co/cask/cdap/api/templates/AbstractAppTemplate.java
```diff
@@ -0,0 +1,45 @@
...
+
+import co.cask.cdap.api.app.AbstractApplication;
+import co.cask.cdap.api.service.http.HttpServiceHandler;
+
+import javax.annotation.Nullable;
+
+/**
+ * Abstract App Template that provides additional functionality required for App Templates.
+ */
+public abstract class AbstractAppTemplate extends AbstractApplication {
```
chtyim: Just call it `ApplicationTemplate` as there is no interface for that.

cdap-api/src/main/java/co/cask/cdap/api/templates/AbstractAppTemplate.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public abstract ManifestSpecification configureManifest(String configuration) throws Exception;
+
+  /**
+   * Provide a Service Handler class that provides HTTP endpoints
+   *
+   * @return {@link HttpServiceHandler}
+   */
+  @Nullable
+  public abstract Class<? extends HttpServiceHandler> getServiceHandler();
```
chtyim: should have a default implementation of it that returns `null`.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
albertshau: ok i see.

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
gokulavasan: I think we need the Driver class which is specific to a rdbms. Anyways, I will remove this.

cdap-api/src/main/java/co/cask/cdap/api/templates/AbstractAppTemplate.java
gokulavasan: Should AbstractBatchSink, AbstractBatchSource (and their Realtime counterparts) be renamed to BatchSink, BatchSource as well?

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestSpecification.java
```diff
@@ -0,0 +1,89 @@
...
+    this.instances = instances;
+    this.programArgs = programArgs;
+  }
+
+  /**
+   * Type of the Program to be started.
+   *
+   * @return program type
+   */
+  public String getProgramType() {
```
chtyim: What is program type? Why is it needed?

cdap-api/src/main/java/co/cask/cdap/api/templates/AbstractAppTemplate.java
```diff
@@ -0,0 +1,45 @@
...
+public abstract class AbstractAppTemplate extends AbstractApplication {
+
+  /**
+   * Given the manifest configuration, provides the {@link ManifestSpecification}
+   *
+   * @param configuration manifest configuration
+   * @return {@link ManifestSpecification}
+   * @throws Exception if the configuration is not valid
+   */
+  public abstract ManifestSpecification configureManifest(String configuration) throws Exception;
```
chtyim: Should follow the configurer style.

cdap-api/src/main/java/co/cask/cdap/api/templates/AbstractAppTemplate.java
chtyim: Mark this class with `@Beta` as I believe it will changed

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestSpecification.java
```diff
@@ -0,0 +1,89 @@
...
+  public Integer getInstances() {
+    return instances;
+  }
+
+  /**
+   * Arguments to be passed to the program.
+   *
+   * @return program arguments
+   */
+  public Map<String, String> getProgramArgs() {
```
chtyim: Why needed? How is it being used?

cdap-app-templates/cdap-etl/cdap-etl-api/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+
+  <parent>
+    <groupId>co.cask.cdap</groupId>
+    <artifactId>cdap-etl</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>cdap-etl-api</artifactId>
+  <name>CDAP ETL App Template API</name>
+  <version>3.0.0-SNAPSHOT</version>
```
chtyim: Shouldn't declare version in children pom

cdap-app-templates/cdap-etl/cdap-etl-core/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+
+  <parent>
+    <groupId>co.cask.cdap</groupId>
+    <artifactId>cdap-etl</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>cdap-etl-core</artifactId>
+  <name>CDAP ETL App Template Core</name>
+  <version>3.0.0-SNAPSHOT</version>
```
chtyim: Remove version

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+
+  <parent>
+    <groupId>co.cask.cdap</groupId>
+    <artifactId>cdap-etl</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>cdap-etl-lib</artifactId>
+  <name>CDAP ETL App Template Library</name>
+  <version>3.0.0-SNAPSHOT</version>
```
chtyim: Remove version

cdap-app-templates/cdap-etl/cdap-etl-test/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+
+  <parent>
+    <groupId>co.cask.cdap</groupId>
+    <artifactId>cdap-etl</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>cdap-etl-test</artifactId>
+  <name>CDAP ETL App Template Test</name>
+  <version>3.0.0-SNAPSHOT</version>
```
chtyim: Remove

cdap-app-templates/cdap-etl/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+  <modules>
+    <module>cdap-etl-api</module>
+    <module>cdap-etl-core</module>
+    <module>cdap-etl-lib</module>
+    <module>cdap-etl-test</module>
+  </modules>
+
+  <repositories>
+    <repository>
+      <id>continuuity-public</id>
```
chtyim: What is it for? Should remove it.

cdap-app-templates/cdap-etl/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+      <url>https://repository.continuuity.com/content/groups/public</url>
+    </repository>
+  </repositories>
+
+  <properties>
+    <twitter4j.version>4.0.1</twitter4j.version>
+    <mysql.jdbc.version>5.1.6</mysql.jdbc.version>
+  </properties>
+
+  <dependencies>
```
chtyim: Parent pom should only have dependency management, not dependency. Otherwise all these will be dependency of "cdap-etl-api", which is not what we wanted.

cdap-app-templates/cdap-etl/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.rat</groupId>
+          <artifactId>apache-rat-plugin</artifactId>
```
chtyim: Why have specialize RAT?

cdap-app-templates/cdap-etl/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.14.1</version>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
```
chtyim: Should use dependency management for bundle jar and only certain module should have this executed (e.g. cdap-etl-api shouldn't be a bundle jar).

cdap-app-templates/cdap-etl/suppressions.xml
```diff
@@ -0,0 +1,42 @@
...
+  the License.
+  -->
+
+<!DOCTYPE suppressions PUBLIC
+        "-//Puppy Crawl//DTD Suppressions 1.1//EN"
+        "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+
+  <suppress checks="Javadoc.*" files=".*/src/test/java/.*" />
```
chtyim: Why use a different checkstyle?

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -0,0 +1,47 @@
...
+public abstract class ApplicationTemplate extends AbstractApplication {
+
+  /**
+   * Given the manifest configuration, provides the {@link ManifestSpecification}
+   *
+   * @param configuration manifest configuration
+   * @return {@link ManifestSpecification}
+   * @throws Exception if the configuration is not valid
+   */
+  public abstract ManifestSpecification configureManifest(String configuration) throws Exception;
```
chtyim: Use the configurer style like the other API.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Emitter.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.api;
+
+/**
+ * Emit data.
```
chtyim: Some more javadoc.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Emitter.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @param <O> Object that Emitter emits
```
chtyim: Type of object that the Emitter can emit.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/StageConfigurer.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Set the description of the Stage.
+   *
+   * @param description description
+   */
+  void setDescription(String description);
+
+  /**
+   * Set the properties this stage requires during runtime.
```
chtyim: How about optional properties? Is there a way to specify them as well?

chtyim: Also, I believe it's for the ELT service handler to get information about a Stage, right? So, probably we need something more than just property name. Maybe something like a `List<Option>`, which `Option` has name, optional, description inside.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/TransformContext.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package co.cask.cdap.templates.etl.api;
+
+import co.cask.cdap.api.RuntimeContext;
+
+/**
+ * Context passed to the Pipeline Stages.
+ */
+public interface TransformContext extends RuntimeContext {
```
chtyim: So there is nothing on this interface?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
chtyim: Why have this method? Should the BatchSink is just setting up the job?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSource.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Process key, value and emit an object for subsequent stages to process.
+   *
+   * @param key Key class from Input
+   * @param value Value class from Input
+   * @param data Emit data
+   */
+  public abstract void process(KEY key, VALUE value, Emitter<O> data);
```
chtyim: Shouldn't have this. The execution sequence should be:

1. MR framework creates RecordReader from the InputFormat as being setup by this class.
2. ETLMapper.map() method get called with the Key and Value provided by the MR framework
3. Inside the ETLMapper.map() method, invoke a series of transform method with the given Key/Value.
4. What ever emitted by the latest stage of the transform is being written through the MR context object inside the ETLMapper.map() method.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSource.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Operation to be performed at the end of the Batch job.
+   *
+   * @param succeeded true if Batch operation succeeded, false otherwise
+   * @param context {@link BatchContext}
+   * @throws Exception
+   */
+  public abstract void onFinish(boolean succeeded, BatchContext context) throws Exception;
```
chtyim: Do you have a use case for this?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSource.java
```diff
@@ -0,0 +1,85 @@
...
+   * @param context {@link BatchContext}
+   */
+  public abstract void prepareJob(BatchContext context);
+
+  /**
+   * Initialize the Batch Source. Invoked at the start of the Batch Job.
+   *
+   * @param context {@link BatchContext}
+   */
+  public void initialize(BatchContext context) {
```
chtyim: What's the different between this and the prepareJob?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Overrides the output configuration of this Batch job to write to the specified dataset instance.
+   * Currently, the dataset passed in must either be an {@link OutputFormatProvider}.
+   * You may want to use this method instead of {@link #setOutput(String)} if your output dataset uses runtime
+   * arguments set in your own program logic.
+   *
+   * @param datasetName the name of the output dataset
+   * @param dataset the output dataset
+   */
+  void setOutput(String datasetName, Dataset dataset);
```
chtyim: So the BatchContext allows someone to set both input and output?? Then what's the point of having BatchSource and BatchSink??

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Emitter.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public interface Emitter<O> {
```
chtyim: I think the `Emitter` should have key and value.

```java
interface Emitter <K, V> {
  void emit(K key, V value);
}

interface ValueEmitter<V> extends Emitter<Void, V> {
  void emit(V value);
}
```

You can do similar thing in Transform. See how the Kafak flowlet is being done in the kafka-pack repo.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSource.java
gokulavasan: In the StreamConversionAdapter, we create a Partition in the TPFS dataset, inside the onFinish method. Hence I thought it might be useful to have.

chtyim: I think that's a hack we want to get rid of. Need to check with @anew on that.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/TransformContext.java
gokulavasan: Yes transforms don't have anymore info other than their config properties which is captured in RuntimeContext. I had instanceId before but realized it won't make any sense in Batch case and transforms don't need to depend on the instance id as @albertshau pointed out.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/StageConfigurer.java
gokulavasan: Added a class for the same.

gokulavasan:
@chtyim @albertshau Addressed your comments. Please review. Thanks!

awholegunch:
The README.md should be a .rst file instead of Markdown.

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestConfigurer.java
```diff
@@ -0,0 +1,62 @@
...
+package co.cask.cdap.api.templates;
+
+import co.cask.cdap.api.schedule.Schedule;
+
+import java.util.Map;
+
+/**
+ * Configurer used to configure program used in the execution of the Manifest.
+ */
+public interface ManifestConfigurer {
```
chtyim: @Beta as well.

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -0,0 +1,49 @@
...
+   */
+  public abstract void configureManifest(String configuration, ManifestConfigurer configurer) throws Exception;
+
+  /**
+   * Provide a Service Handler class that provides HTTP endpoints
+   *
+   * @return {@link HttpServiceHandler}
+   */
+  @Nullable
+  public Class<? extends HttpServiceHandler> getServiceHandler() {
```
chtyim: getServiceHandlerClass

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -0,0 +1,49 @@
...
+   * Given the manifest configuration, configures the Manifest.
+   *
+   * @param configuration manifest configuration
+   * @param configurer {@link ManifestConfigurer}
+   * @throws Exception if the configuration is not valid
+   */
+  public abstract void configureManifest(String configuration, ManifestConfigurer configurer) throws Exception;
+
+  /**
+   * Provide a Service Handler class that provides HTTP endpoints
```
chtyim: Full-stop missing.

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @return {@link HttpServiceHandler}
```
chtyim: Need to say what happen when it returns `null`, since the return is `@Nullable`.

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -0,0 +1,49 @@
...
+public abstract class ApplicationTemplate extends AbstractApplication {
+
+  /**
+   * Given the manifest configuration, configures the Manifest.
+   *
+   * @param configuration manifest configuration
+   * @param configurer {@link ManifestConfigurer}
+   * @throws Exception if the configuration is not valid
+   */
+  public abstract void configureManifest(String configuration, ManifestConfigurer configurer) throws Exception;
```
chtyim: Should it be an abstract method or just a regular method with no-op implementation? Can an app template have nothing to configure per manifest?

gokulavasan: If it doesn't have anything to configure, we won't know what program to start. By having all this info in the configureManifest, we have removed the need to have any info in the MANIFEST file of etl jar.

chtyim: hum.. I thought the program to be started is tied to the template-id, which is part of the configuration that embedded inside the app template jar?

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestConfigurer.java
```diff
@@ -0,0 +1,62 @@
...
+   * @param instances number of instances
+   */
+  public void setInstances(int instances);
+
+  /**
+   * Set the arguments to be passed to the program as runtime arguments.
+   *
+   * @param programArgs
+   */
+  public void setProgramArgs(Map<String, String> programArgs);
```
chtyim: Better have it as "addRuntimeArguments" (so it's a addition, not replacement). Also a method for adding entry one by one `addRuntimeArgument(String key, String value)`.

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
albertshau: Is configuration the request body provided when creating a manifest?  I think it should either be bytes if we are allowing anything, or a Map<String, Object> if we are requiring a json object.

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestConfigurer.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Set the name of the program.
+   *
+   * @param name program name
+   */
+  public void setProgramName(String name);
+
+  /**
+   * Set the schedule for the program.
```
albertshau: add a sentence about which program types can be scheduled

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestConfigurer.java
```diff
@@ -0,0 +1,62 @@
...
+ * Configurer used to configure program used in the execution of the Manifest.
+ */
+public interface ManifestConfigurer {
+
+  /**
+   * Set the program type.
+   *
+   * @param type program type
+   */
+  public void setProgramType(String type);
```
albertshau: seems like it should use the ProgramType enum

gokulavasan: @albertshau Removed setting of programType, programName and we could have it as part of the template id. So template-id will be (appid -> name of apptemplate) : appid.programid.

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestConfigurer.java
gokulavasan: Hmm, true. Will add a comment.

gokulavasan:
@chtyim @albertshau Thanks for the comments. Addressed them. Please take a look again.

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestConfigurer.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Configurer used to configure program used in the execution of the Manifest.
+ * Currently, only Worker or Workflow can be used in the execution.
+ */
+@Beta
+public interface ManifestConfigurer {
+
+  /**
+   * Set the schedule for the program.
```
albertshau: only workflow can have a schedule right?  Lets add a sentence that this is invalid for Workers.

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestConfigurer.java
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * @param schedule {@link Schedule}
+   */
+  public void setSchedule(Schedule schedule);
+
+  /**
+   * Set the number of instances of the program.
```
albertshau: Similarly, this is only valid for workers and defaults to 1, and is invalid for workflows.

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -0,0 +1,51 @@
...
+public abstract class ApplicationTemplate extends AbstractApplication {
+
+  /**
+   * Given the manifest configuration, configures the Manifest.
+   *
+   * @param configuration manifest configuration
+   * @param configurer {@link ManifestConfigurer}
+   * @throws Exception if the configuration is not valid
+   */
+  public void configureManifest(String configuration, ManifestConfigurer configurer) throws Exception {
```
albertshau: what is this string?  Should it be ```Map<String, Object>``` for some arbitrary json object? Or maybe some ManifestConfiguration object?

gokulavasan: There is no restriction that the configuration should be a json blob. For example, the user could have just a plain string as the configuration which he passes in when creating a 'manifest'. so whatever he provides in the body of the post call, we pass it to the configureManifest method.

albertshau: if it can be anything it shouldn't be a String, but bytes

albertshau:
minor comments, looks good from my end. won't be perfect but we'll iterate as needed.

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
gokulavasan: @chtyim Any comments on this? And on the overall PR in general?

chtyim: The fanciest way is to have `ApplicationTemplate` a generic type, with the configuration object of type `<T>` and CDAP can do the json deserialization of the create manifest body into that type. That would require the REST API spec to have the body as a valid Json.

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -0,0 +1,51 @@
...
+package co.cask.cdap.api.templates;
+
+import co.cask.cdap.api.annotation.Beta;
+import co.cask.cdap.api.app.AbstractApplication;
+import co.cask.cdap.api.service.http.HttpServiceHandler;
+
+import javax.annotation.Nullable;
+
+/**
+ * Abstract App Template that provides additional functionality required for App Templates.
```
chtyim: It should be describing what an app template is (anyway we can fix the doc later)

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -0,0 +1,51 @@
...
+   * @throws Exception if the configuration is not valid
+   */
+  public void configureManifest(String configuration, ManifestConfigurer configurer) throws Exception {
+    // no-op
+  }
+
+  /**
+   * Provide a Service Handler class that provides HTTP endpoints.
+   *
+   * @return {@link HttpServiceHandler} or null if no custom HTTP endpoints are required.
```
chtyim: no HTTP endpoint is supported.

cdap-api/src/main/java/co/cask/cdap/api/templates/ManifestConfigurer.java
```diff
@@ -0,0 +1,59 @@
...
+
+package co.cask.cdap.api.templates;
+
+import co.cask.cdap.api.annotation.Beta;
+import co.cask.cdap.api.schedule.Schedule;
+
+import java.util.Map;
+
+/**
+ * Configurer used to configure program used in the execution of the Manifest.
```
chtyim: execution of Manifest? What does it mean?

chtyim: Anyway I think we have to walkthrough all the javadoc again once the naming and classes are settled

gokulavasan: Ok, I will leave this specific javadoc as it is for now. Will polish them in another PR.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/StageConfigurer.java
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @param properties {@link List} of {@link Property}
+   */
+  void setExpectedProperties(List<Property> properties);
```
chtyim: Why "expected" properties? Is there unexpected properties? Also, make it as `addProperties` and `addProperty` (single `Property`).

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/StageLifecycle.java
chtyim: This interface is exactly the same as `ProgramLifecycle` in cdap-api. Why not just use that one?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Transform Stage.
+ *
+ * @param <KI> Type of KeyInput object
+ * @param <VI> Type of ValueInput object
+ * @param <KO> Type of KeyOutput object
+ * @param <VO> Type of ValueOutput object
+ */
+public abstract class Transform<KI, VI, KO, VO> implements StageLifecycle {
```
chtyim: Missing parameterized type for the lifecycle interface.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private TransformContext context;
+
+  /**
+   * Configure the Transform stage. Used to provide information about the Transform.
+   *
+   * @param configurer {@link StageConfigurer}
+   */
+  public void configure(StageConfigurer configurer) {
+    configurer.setName(this.getClass().getSimpleName());
```
chtyim: This is unnecessary. The configurer provided should have that set already (see how configurer is done in cdap).

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Process input Key and Value and emit output using {@link Emitter}.
+   *
+   * @param inputKey input key
+   * @param inputValue input value
+   * @param emitter {@link Emitter} to emit data to the next stage
+   * @throws Exception
+   */
+  public void transform(KI inputKey, VI inputValue, Emitter<KO, VO> emitter) throws Exception {
+    throw new Exception();
```
chtyim: Should be throwing `UnsupportedOperationException`

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
chtyim: Also, expand the type name to be more meaningful, `KEY_IN, VALUE_IN, KEY_OUT, VALUE_OUT`.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Process input and emit output using {@link ValueEmitter}
+   *
+   * @param input input
+   * @param emitter {@link ValueEmitter} to emit data to the next stage
+   */
+  public void transform(VI input, ValueEmitter<VO> emitter) throws Exception {
+    throw new Exception();
```
chtyim: Same as above.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
chtyim: hum, I think by default this should be just calling 

```java
transform(inputValue, new ValueEmitter<VO>() { 
  void emit(VO value) { 
    emitter.emit(inputKey, value); 
  }
});
```

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
chtyim: I mean throwing the `UnsupportedOperationException`.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/TransformContext.java
chtyim: Don't you at least have a `getName` for custom logging purpose?

chtyim: or `getManifestName`?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public abstract class AbstractBatchSink<KEY, VALUE> {
+
+  /**
+   * Configure the Sink.
+   *
+   * @param configurer {@link StageConfigurer}
+   */
+  public void configure(StageConfigurer configurer) {
+    configurer.setName(this.getClass().getSimpleName());
```
chtyim: shouldn't needed.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  /**
+   * Prepare the Batch Job. Used to configure the Hadoop Job before starting the Batch Job.
+   *
+   * @param context {@link BatchSinkContext}
+   */
+  public abstract void prepareJob(BatchSinkContext context);
```
chtyim: Is it always needed?

chtyim: I think the ETL batch program can have default configs such that for simple source/sink, one doesn't implements a class. E.g.

```
input.format : InputFormat class name
output.format : OutputFormat class name
```

chtyim: Hum, I started to wonder, maybe there is no BatchSource and BatchSink, because basically what they do is to setup the Batch job?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
```diff
@@ -0,0 +1,47 @@
...
+import co.cask.cdap.api.RuntimeContext;
+import co.cask.cdap.api.data.DatasetContext;
+
+/**
+ * Context passed to Batch Sources and Sink.
+ */
+public interface BatchContext extends RuntimeContext, DatasetContext {
+
+  /**
+   * Returns the logical start time of the Batch Job.  Logical start time is the time when this MapReduce
```
chtyim: Shouldn't mention MapReduce.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
chtyim: Is it possible that we don't have this but enforce everything done through configuration?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   */
+  <T> T getHadoopJob();
+
+  /**
+   * Overrides the resources, such as memory and virtual cores, to use for the execution of this Batch job.
+   *
+   * @param resources Resources to be used for execution of this job
+   */
+  void setResources(Resources resources);
```
chtyim: Why there is no such equivalent for the real time case?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/realtime/AbstractRealtimeSink.java
```diff
@@ -0,0 +1,64 @@
...
+
+import co.cask.cdap.templates.etl.api.StageConfigurer;
+import co.cask.cdap.templates.etl.api.StageLifecycle;
+
+/**
+ * Realtime Sink.
+ *
+ * @param <I> Object sink operates on
+ */
+public abstract class AbstractRealtimeSink<I> implements StageLifecycle {
```
chtyim: Not need to have `Abstract` in name.

chtyim: Shouldn't the Sink have both key and value?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/realtime/AbstractRealtimeSource.java
```diff
@@ -0,0 +1,84 @@
...
+   */
+  public abstract void onSuspend();
+
+  /**
+   * Resume/reconfigure from the state of suspension.
+   *
+   * @param oldInstanceCount old instance count
+   * @param newInstanceCount new instance count
+   */
+  public abstract void onResume(int oldInstanceCount, int newInstanceCount);
```
chtyim: The instance count should be provided through the context.

chtyim: Also onSuspend and onResume should have a no-op default implementation.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/realtime/AbstractRealtimeSource.java
```diff
@@ -0,0 +1,84 @@
...
+
+  private SourceContext context;
+
+  /**
+   * Configure the Source.
+   *
+   * @param configurer {@link SourceConfigurer}
+   */
+  public void configure(SourceConfigurer configurer) {
+    configurer.setName(this.getClass().getSimpleName());
```
chtyim: Not needed.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/realtime/AbstractRealtimeSource.java
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Poll for new data.
+   *
+   * @param writer {@link Emitter}
+   * @param currentState {@link SourceState} current state of the source
+   * @return {@link SourceState} state of the source after poll, will be persisted when all data from poll are processed
+   */
+  @Nullable
+  public abstract SourceState poll(ValueEmitter<O> writer, @Nullable SourceState currentState);
```
chtyim: Why `@Nullable`? Why not just have an empty state?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/realtime/AbstractRealtimeSink.java
chtyim: Shouldn't Sink has onSuspend and onResume as well?

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/AbstractBatchSink.java
gokulavasan: But don't we want separation of source and sink in batch? For example, I configure to read from RDBMS in the source and writing to TableDataset in the sink. So that I can mix and match.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
gokulavasan: True. Will add that.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchContext.java
gokulavasan: I think it would easier if we do require prepareJob. This way the source can set its own properties and also set other configuration if required. Found this to be useful when I tried DBInputFormat. Also when we are using CDAP sources it might be easy to prepareJob to set input/outputformats through the BatchSourceContext/BatchSinkContext (though I guess we will provide all possible Batch CDAP Sources/Sinks out of the box).

gokulavasan:
Will merge this to develop once build passes.
http://builds.cask.co/browse/CDAP-DT0-7
 Thanks!

albertshau:
lgtm

PR #1684, opened 2015-03-17
Added retry logic for starting the scheduler.

sagarkapare:
https://issues.cask.co/browse/CDAP-1636

Scheduler during initialization tries to create the table `schedulestore` in the hbase. However if the dataset service is not reachable, the table creation fails and `scheduler` instance is set to null. Fix is to keep retrying the creation of the scheduler till the dataset service comes up and can create the `schedulestore` successfully.


sagarkapare:
I am doing some more testing around the fix. Opening the PR to get early comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/DistributedSchedulerService.java
```diff
@@ -65,11 +66,22 @@ public void onChange(ServiceDiscovered serviceDiscovered) {
...
+            int retries = 0;
+            while (true) {
+              try {
+                startScheduler();
+                schedulerStarted.set(true);
+                LOG.info("Scheduler service started successfully.");
+                break;
+              } catch (Throwable t) {
+                retries++;
+                LOG.warn("Error during retry# {} - {}", retries, t);
```
bdmogal: should it be ``t.getMessage`` or something? Otherwise I'm not sure how this would work. We shouldn't log the entire stack trace for sure.

bdmogal:
Apart from one minor comment, LGTM pending testing.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/DistributedSchedulerService.java
```diff
@@ -65,11 +66,22 @@ public void onChange(ServiceDiscovered serviceDiscovered) {
           if (!Iterables.isEmpty(serviceDiscovered) && !schedulerStarted.get()) {
             LOG.info("Starting scheduler, Discovered {} dataset service(s)",
                      Iterables.size(serviceDiscovered));
-            try {
-              startScheduler();
-              schedulerStarted.set(true);
-            } catch (Throwable t) {
-              LOG.error("Exception when starting scheduler.", t);
+            int retries = 0;
+            while (true) {
```
ParthGandhi: Not a good idea to retry indefinitely. Possible to have a hard limit?

ParthGandhi: This is fine for now. But we may want to re-evaluate later.

sagarkapare:
Could not reproduce the issue. However by throwing an exception explicitly from DatasetHandler, I was able to verify the retry logic. 
Build passed here - http://builds.cask.co/browse/CDAP-RBT185-3

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/DistributedSchedulerService.java
sagarkapare: This seems to be only printing the exception message.

PR #1687, opened 2015-03-18
fix for runtime stats reset

gsps1:
fix for issue describer here : https://issues.cask.co/browse/CDAP-1846

build running : http://builds.cask.co/browse/CDAP-RBT177-1

### DIFF
--- a/cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
+++ b/cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
@@ -631,6 +631,18 @@ public class TestFrameworkTestRun extends TestFrameworkTestBase {
     Assert.assertEquals("value1", myTableManager3.get().get(new Get("key1", "column1")).getString("column1"));
   }
 
+  @Test(timeout = 60000L)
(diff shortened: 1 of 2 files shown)


cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
```diff
@@ -631,6 +631,18 @@ public void testAppRedeployKeepsData() throws Exception {
...
   }
 
+  @Test(timeout = 60000L)
+  public void testFlowletMetricsReset() throws Exception {
+    ApplicationManager appManager = deployApplication(DataSetInitApp.class);
+    FlowManager flowManager = appManager.startFlow("DataSetFlow");
+    RuntimeMetrics flowletMetrics = RuntimeStats.getFlowletMetrics("DataSetInitApp", "DataSetFlow", "Consumer");
+    flowletMetrics.waitForProcessed(1, 5, TimeUnit.SECONDS);
+    flowManager.stop();
+    RuntimeStats.resetAll();
```
abaranau: use RuntimeMetrics.getProcessed before and after - before it should be > 0, after it should be 0. Using waitfor is hacky

abaranau:
Small comment. Good to merge after fix

PR #1697, opened 2015-03-18
Adds a "v3" to the endpoint

awholegunch:
And fixes a bug where it was return for "hbase" and JSON, the CDAP config.

Build: http://builds.cask.co/browse/CDAP-DUT1167-1

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/gateway/handlers/ConfigHandler.java
+++ b/cdap-common/src/main/java/co/cask/cdap/gateway/handlers/ConfigHandler.java
@@ -17,6 +17,7 @@
 package co.cask.cdap.gateway.handlers;
 
 import co.cask.cdap.common.conf.CConfiguration;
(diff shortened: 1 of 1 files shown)


jwang47:
LGTM

cdap-common/src/main/java/co/cask/cdap/gateway/handlers/ConfigHandler.java
```diff
@@ -42,6 +43,7 @@
 /**
  * Handles requests to view {@link CConfiguration}.
  */
+@Path(Constants.Gateway.API_VERSION_3)
```
anwar6953: do we want a v2 also? (was this feature also in 2.7?)

awholegunch: No, this was added in 2.8. It's not in 2.7.

PR #1698, opened 2015-03-18
Feature/r2.8 config rest api

awholegunch:
Adds a "Configuration" API to the v3 RESTful APIs. 
Fix for https://issues.cask.co/browse/CDAP-1561?filter=10226

Staged at:

[Configuration API](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-r2.8_config_rest_API/en/reference-manual/http-restful-api/http-restful-api-v3/configuration.html)

### DIFF
--- a/cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/configuration.rst
+++ b/cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/configuration.rst
@@ -1,0 +1,61 @@
+.. meta::
+    :author: Cask Data, Inc.
+    :description: HTTP RESTful Interface to the Cask Data Application Platform
+    :copyright: Copyright © 2015 Cask Data, Inc.
+
+.. highlight:: console
+
+.. _http-restful-api-configuration:
+.. _http-restful-api-v3-configuration:
+
+======================================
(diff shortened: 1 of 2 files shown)


cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/configuration.rst
```diff
@@ -0,0 +1,107 @@
...
+    :description: HTTP RESTful Interface to the Cask Data Application Platform
+    :copyright: Copyright © 2015 Cask Data, Inc.
+
+.. highlight:: console
+
+.. _http-restful-api-configuration:
+.. _http-restful-api-v3-configuration:
+
+======================================
+Configuration Service HTTP RESTful API
```
bdmogal: Is this a service? It may give a wrong notion - should we just call it "Configuration HTTP RESTful API"

bdmogal:
Minor comment, rest LGTM

bdmogal:
:+1: it wasn't staged after your last commit right?

awholegunch:
Correct. I don't have time to restage all these.

PR #1700, opened 2015-03-18
CDAP-1000: Fix binding to 0.0.0.0 and announce routable interface when binding to 0.0.0.0

jwang47:
https://issues.cask.co/browse/CDAP-1000
http://builds.cask.co/browse/CDAP-RBT184


cdap-common/src/main/java/co/cask/cdap/common/discovery/ResolvingDiscoverable.java
```diff
@@ -0,0 +1,60 @@
...
+    return resolve(discoverable.getSocketAddress());
+  }
+
+  private InetSocketAddress resolve(InetSocketAddress bindAddress) {
+    try {
+      if (Inet6Address.getByName("0:0:0:0:0:0:0:0").equals(bindAddress.getAddress())) {
+        return new InetSocketAddress(InetAddress.getLocalHost().getHostName(), bindAddress.getPort());
+      }
+    } catch (UnknownHostException e) {
+      e.printStackTrace();
```
gokulavasan: Fix Logging

cdap-common/src/main/java/co/cask/cdap/common/discovery/ResolvingDiscoverable.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public InetSocketAddress getSocketAddress() {
+    return resolve(discoverable.getSocketAddress());
+  }
+
+  private InetSocketAddress resolve(InetSocketAddress bindAddress) {
+    try {
+      if (Inet6Address.getByName("0:0:0:0:0:0:0:0").equals(bindAddress.getAddress())) {
```
nitinmotgi: Why IPv6 binding ?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -262,7 +262,7 @@
 
     <property>
       <name>stream.bind.address</name>
-      <value>127.0.0.1</value>
```
nitinmotgi: Is every service in production going to be binding on all interfaces ?

jwang47: I think binding on all interfaces is nice as a default, but I think in production, developers should be configuring these properties. At least, that's how I interpreted this issue: https://issues.cask.co/browse/CDAP-1000

cdap-common/src/main/java/co/cask/cdap/common/discovery/ResolvingDiscoverable.java
jwang47: That's what I happened to get when I tried "0.0.0.0" as the bind address for the `NettyHttpService`s. Probably should also do a check for `Inet4Address`?

wolf31o2: Start your services with `-Djava.net.preferIPv4Stack=true` or you can disable IPv6 on the host and reboot so you won't get any IPv6 addresses. However, you should check for both, and resolve IPv6 before IPv4, since that's what all modern OS (Linux, OSX, Windows even) do when both are supported.

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -983,7 +983,7 @@
 
     <property>
         <name>router.server.address</name>
-        <value>localhost</value>
+        <value>0.0.0.0</value>
```
wolf31o2: This should be the address used to connect to the router. I think the default of localhost is fine, here. Setting it to `0.0.0.0` means anything that uses this file, like the UI, won't know what to connect to.

cdap-common/src/main/java/co/cask/cdap/common/discovery/ResolvingDiscoverable.java
chtyim: Why not use `InetAddress.isAnyLocalAddress()`?? We shouldn't deal with IPv4/IPv6 directly. Should use Java library for that.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/AppFabricServer.java
```diff
@@ -169,9 +170,9 @@ public String getName() {
 
             @Override
             public InetSocketAddress getSocketAddress() {
-              return socketAddress;
+              return httpService.getBindAddress();
```
chtyim: Why the change?

cdap-common/src/main/java/co/cask/cdap/common/discovery/ResolvingDiscoverable.java
```diff
@@ -0,0 +1,60 @@
...
+    return resolve(discoverable.getSocketAddress());
+  }
+
+  private InetSocketAddress resolve(InetSocketAddress bindAddress) {
+    try {
+      if (bindAddress.getAddress().isAnyLocalAddress()) {
+        return new InetSocketAddress(InetAddress.getLocalHost().getHostName(), bindAddress.getPort());
+      }
+    } catch (UnknownHostException e) {
+      // IGNORE
```
chtyim: Is it safe to ignore? Or at least need to log something??

cdap-common/src/main/resources/cdap-default.xml
dereklwood: :+1: for defaults being 0.0.0.0.  This will reduce the numerous error cases when a service cant be accessed in a distributed cluster because its listening on localhost only, and its consistent with hadoop (http://hadoop.apache.org/docs/r2.6.0/hadoop-project-dist/hadoop-hdfs/hdfs-default.xml).  Primary downside I can think of would be security related, but as jwang47 mentioned, production environments should be configured anyway.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/AppFabricServer.java
jwang47: The other services were using this sort of pattern, so I also applied it here.

cdap-common/src/main/java/co/cask/cdap/common/discovery/ResolvingDiscoverable.java
jwang47: Yeah, probably log a WARN

sreevatsanraman:
LGTM :-1:

PR #996, opened 2015-01-17
CDAP-1171 Remove use of k5start in init scripts and docs

ghelmling:
This change updates the init scripts and docs to account for CDAP-1134, which uses a programmatic keytab login:
* removes use of k5start from init scripts, since it is no longer required
* adds kerberos properties to cdap-default.xml
* updates installation doc to remove references to k5start and reflect new configuration
* fixes process limit configuration for cdap user

PR #1701, opened 2015-03-18
CDAP-1817 Run kinit on secure clusters so "hive -e" works

wolf31o2:
This is a semi-reversal of #996 for CDAP-1171 which allows Explore to run on secure clusters by detecting the Hive CLASSPATH properly.


albertshau:
:+1:

PR #1702, opened 2015-03-18
Adding Direct Parcel Access to the documentation.

awholegunch:
Staged at: [Installation](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-SNAPSHOT-r2.8_doc_parcel_access/en/admin-manual/installation/installation.html#installation): see the section "Direct Parcel Access"
Fix for https://issues.cask.co/browse/CDAP-1597


cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -309,6 +309,24 @@ recommended installation is a minimum of two boxes.
 
 This will download and install the latest version of CDAP with all of its dependencies. 
 
+Direct Parcel Access
```
dereklwood: This may be a little out of context on the installation page... should this just be on the cloudera manager integration page?  Parcels are only useful to Cloudera Manager users.  We should at least say something like "If you are using Cloudera Manager..."

cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -309,6 +309,24 @@ recommended installation is a minimum of two boxes.
...
+....................
+
+If you need to download and install the parcels directly (perhaps for a cluster that does
+not have direct network access), the parcels are available by their full URLs. As they are
+stored in a directory that does not offer browsing, they are listed here:
+
+.. parsed-literal::
+  |http:|//repository.cask.co/parcels/cdap/latest/CDAP-|version|-1-el6.parcel
+  |http:|//repository.cask.co/parcels/cdap/latest/CDAP-|version|-1-precise.parcel
+  |http:|//repository.cask.co/parcels/cdap/latest/CDAP-|version|-1-trusty.parcel
```
dereklwood: there is also wheezy now:  http://repository.cask.co/parcels/cdap/latest/CDAP-2.7.1-1-wheezy.parcel

dereklwood: we might also consider making this more future-proof by just giving the ``-el6`` example, and telling them to refer to the manifest.json for available parcels:  http://repository.cask.co/parcels/cdap/latest/manifest.json

dereklwood:
I made a few comments and possible suggestions... will leave it up to your expertise if you disagree

sreevatsanraman:
LGTM :+1:

dereklwood:
I would suggest moving it to the bottom of the page (maybe above the Troubleshooting section), and then referring to it in this line: https://github.com/caskdata/cdap/blob/feature/r2.8_doc_parcel_access/cdap-docs/integrations/source/partners/cloudera/configuring.rst#L107

maybe something like:  ``#. Download and distribute the CDAP-|version| parcel.  If the Cask parcel repo is inaccessible to your cluster, please see [here]``

Reasoning is that this is more of an edge case and its current location kinda disrupts the documentation flow.

Otherwise, LGTM

dereklwood:
:+1:

PR #1703, opened 2015-03-18
Replace clearspring imports with Guava

ParthGandhi:


gokulavasan:
LGTM :+1:

chtyim:
Do you know what brings in the clearspring dep? Can we exclude it?

anwar6953:
I think its spark

ParthGandhi:
Yeah, its `spark-core`. I don't know Spark well enough and didn't spend time trying to figure out if it actually gets used and whether excluding it is safe or not.

ParthGandhi:
Build: http://builds.cask.co/browse/CDAP-DUT1174-1/

Merging this for now. Evaluating whether that dependency can be excluded is a bit more work and can be done separately.

PR #1708, opened 2015-03-18
Increase worker dataset cache timeout to 2 hours

ParthGandhi:
This is a temporary fix till we can handle cache timeouts more elegantly in the future.


poornachandra:
LGTM

ParthGandhi:
Build: http://builds.cask.co/browse/CDAP-DUT1179-1/

gokulavasan:
Should we do this for Workers too?

ParthGandhi:
@gokulavasan Good catch. Can you take a look at the last commit?

gokulavasan:
LGTM :+1:

PR #1709, opened 2015-03-18
CDAP-1842 fix log filters for namespaces

rohitsinha54:
- This fixes the issues with filter and enables us to read log files.
- But there is another bigger problem with logs from Kafka due to namespace changes for which I will open another Jira.


rohitsinha54:
Build passed here: https://builds.cask.co/browse/CDAP-RBT186-1

poornachandra:
LGTM

PR #1711, opened 2015-03-18
CDAP-1592: delete only aggregated metrics on removal of program, ...

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-RBT188-1

gsps1:
so we don't want to change MetricDeleteQuery as you mentioned in Jira ?

abaranau:
you mean remove start/stop ts?

abaranau:
we keep it, yes, for now - would make to many changes it seems

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
```diff
@@ -220,8 +226,18 @@ public void delete(MetricDeleteQuery query) throws Exception {
...
+    // this will delete all aggregates metrics data
+    delete(new MetricDeleteQuery(0, System.currentTimeMillis() / 1000, null,
+                                             Maps.<String, String>newHashMap()));
+    // this will delete all timeseries data
+    deleteBefore(System.currentTimeMillis() / 1000);
+  }
+
   private CubeDeleteQuery buildCubeDeleteQuery(MetricDeleteQuery query) {
-    return new CubeDeleteQuery(query.getStartTs(), query.getEndTs(),
+    // note: delete query currently usually executed in sync way, so we only attempt to delete totals, to avoid timeout
```
gsps1: not executed asynchronously - might be better to understand.

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsSuiteTestBase.java
```diff
@@ -128,6 +131,9 @@ public static void beforeClass() throws Exception {
     StoreFactory storeFactory = injector.getInstance(StoreFactory.class);
     store = storeFactory.create();
     locationFactory = injector.getInstance(LocationFactory.class);
+
+    metricStore = injector.getInstance(MetricStore.class);
+
```
gsps1: remove unwanted new lines

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricStore.java
```diff
@@ -51,6 +51,12 @@
   void delete(MetricDeleteQuery query) throws Exception;
 
   /**
+   * Deletes all metrics data.
+   * @throws Exception
```
gsps1: does it make sense to add a note ,should only  be used in tests ?

abaranau:
http://builds.cask.co/browse/CDAP-RBT188-3 GREEN

PR #1714, opened 2015-03-19
(CDAP-1204) HBase sharded queue implementation

chtyim:


chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT166-19

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerState.java
```diff
@@ -15,30 +15,55 @@
...
 
-  HBaseConsumerState(byte[] startRow, long groupId, int instanceId) {
+  HBaseConsumerState(ConsumerConfig consumerConfig, byte[] startRow,
+                     @Nullable byte[] barrierStartRow, @Nullable byte[] barrierEndRow) {
+    this.consumerConfig = consumerConfig;
     this.startRow = startRow;
-    this.groupId = groupId;
-    this.instanceId = instanceId;
+    this.barrierStartRow = barrierStartRow;
+    this.barrierEndRow = barrierEndRow;
```
anew: hmm. Naming - although I think the word barrier is appropriate - is a little strange. If I understand correctly, barrierEndRow is where the config changed, and the consumer waits for all other consumers to each this row. So I would call this simply "barrier". But what is barrierStartRow? Isn't that the barrier of the previous config change? So maybe call it previousBarrier?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -17,49 +17,122 @@
...
+  private final byte[] barrierScanStartRow;
+  private final byte[] barrierScanEndRow;
+  private Transaction transaction;
 
   HBaseConsumerStateStore(String datasetName, QueueName queueName, Table table) {
     super(datasetName, table);
     this.queueName = queueName;
     this.table = table;
-    this.minStartRow = QueueEntryRow.getQueueEntryRowKey(queueName, 0L, 0);
+    this.barrierScanStartRow = Bytes.add(queueName.toBytes(), QueueEntryRow.getQueueEntryRowKey(queueName, 0L, 0));
```
anew: so this contains the queue name twice?

chtyim: the row key for the barrier is actually "queueName + queueEntryRow".. which is agnostic to what the actual format of the entry row key is (the actual entry row key is "flowletId + name + writeTx + counter").

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -17,49 +17,122 @@
...
+    ConsumerState consumerState = getConsumerState(groupId, instanceId);
+    QueueBarrier startBarrier = consumerState.getStartBarrier();
+    QueueBarrier endBarrier = consumerState.getEndBarrier();
+    if (startBarrier == null && endBarrier == null) {
+      throw new IllegalStateException(
+        String.format("Unable to find barrier information for consumer. Queue: %s, GroupId: %d, InstanceId:%d",
+                      queueName, groupId, instanceId));
+    }
+
+    ConsumerGroupConfig groupConfig = startBarrier != null ? startBarrier.getGroupConfig()
```
anew: I am trying to understand... Is it correct that:
- all consumers now have a state with a startBarrier (starting at 0) 
- so if the startBarrier is null, then we are in compatibility mode, consuming entries that were produced with the old strategy
- every time the config changes, a new barrier is added. That means if the state has an end barrier, then the consumer must wait for all other consumers to catch up when he reaches the barrier.

Sometimes it is useful to explain that in Javadocs or comments. The rest of the code gets much easier to understand

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -17,49 +17,122 @@
...
+                                                             : endBarrier.getGroupConfig();
+    ConsumerConfig consumerConfig = new ConsumerConfig(groupConfig, instanceId);
+    return new HBaseConsumerState(consumerConfig,
+                                  consumerState.getConsumerStartRow(),
+                                  startBarrier == null ? null : startBarrier.getStartRow(),
+                                  endBarrier == null ? null : endBarrier.getStartRow());
+  }
+
+  /**
+   * Checks if consumers in the given group has consumed all entries up to the given start row.
```
anew: all consumers have consumed?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
chtyim: Yes it is correct. I'll add comments for that.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -240,22 +223,44 @@ public void dropAllForFlow(String namespaceId, String app, String flow) throws E
...
+      } else if (isStateStoreTable(tableId)) {
+        stateStoreTableIds.add(tableId);
+      }
+    }
+
+    // Upgrade of state store table
+    for (TableId tableId : stateStoreTableIds) {
+      LOG.info("Upgrading queue state store: {}", tableId);
+      Id.DatasetInstance stateStoreId = createStateStoreDataset(tableId.getNamespace().getId());
+      co.cask.cdap.api.dataset.DatasetAdmin admin = datasetFramework.getAdmin(stateStoreId, null);
```
anew: why fully qualified?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
```diff
@@ -120,4 +228,87 @@ private HTable createHTable(TableId tableId) throws IOException {
...
+      }
+      super.startTx(tx);
+    }
+
+    @Override
+    public void postTxCommit() {
+      super.postTxCommit();
+      HBaseQueueConsumer consumer = consumers.poll();
+      if (!consumer.isClosed()) {
+        consumers.add(consumer);
```
anew: So, a single consumer does not have multiple threads if it is catching up for an instance id that was removed. Instead this SmartConsumer round-robins through multiple instance ids. I am sure this simplifies the code... in terms of performance, though, catch-up to the barrier will take longer, right?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -424,7 +424,7 @@
 
     <property>
         <name>data.tx.snapshot.interval</name>
-        <value>300</value>
+        <value>60</value>
```
anew: unrelated change? why?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedFlowletInstanceUpdater.java
```diff
@@ -56,16 +63,34 @@
...
+      twillController.changeInstances(flowletId, newInstanceCount).get();
+    } finally {
+      for (String id : flowlets) {
+        // Need to resume one by one due to TWILL-123.
+        // Also ignore any error as resuming a non-suspended flowlet is ok,
+        // except getting from exception log from the container.
+        try {
+          twillController.sendCommand(id, ProgramCommands.RESUME).get();
+        } catch (Exception e) {
+          LOG.warn("Failed to resume flowlet {}", id, e);
```
anew: maybe you should error out here? Because it means one flowlet is not running. So the flow is dysfunctional.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
chtyim: That's is right. It's a future improvement for using multiple-threads to catchup. The change is not that trivial as it involves changes in the Flow system.

cdap-common/src/main/resources/cdap-default.xml
chtyim: It actually should belongs to previous PR that changes state store to use dataset. A shorter interval for taking tx snapshot is more favorable since the queue eviction coprocessor now needs to read the state store transactionally, meaning a more up-to-date tx snapshot helps.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
anew: so because queueName already contains the flowlet id and name, there is some duplication. I guess that is ok.

anew:
This is quite complex a change, but overall the code looks very good. I have only a few comments.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerState.java
chtyim: Renamed them to "previousBarrier" and "nextBarrier".

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
chtyim: fixed

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
chtyim: Because there is an inner class called `DatasetAdmin` as well.

anew:
LGTM. Nice work :100:

PR #1717, opened 2015-03-19
Improve guice bindings in upgrade tool

bdmogal:
Guice fixes required to enable queue config upgrade in upgrade tool.
Build: http://builds.cask.co/browse/CDAP-RBT190-1

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
@@ -16,6 +16,7 @@
 package co.cask.cdap.data.tools;
 
 import co.cask.cdap.api.dataset.module.DatasetDefinitionRegistry;
+import co.cask.cdap.api.dataset.module.DatasetModule;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
 import co.cask.cdap.api.schedule.Schedule;
 import co.cask.cdap.app.guice.ProgramRunnerRuntimeModule;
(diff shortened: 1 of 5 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -386,10 +384,10 @@ public static void main(String[] args) throws Exception {
...
   private DatasetFramework createRegisteredDatasetFramework(CConfiguration cConf,
-                                                            DatasetDefinitionRegistryFactory registryFactory)
+                                                            DatasetDefinitionRegistryFactory registryFactory,
+                                                            Map<String, DatasetModule> defaultModules)
     throws DatasetManagementException, IOException {
-    DatasetFramework datasetFramework = new InMemoryDatasetFramework(registryFactory, cConf);
-    addModules(datasetFramework);
+    DatasetFramework datasetFramework = new InMemoryDatasetFramework(registryFactory, defaultModules, cConf);
     // dataset service
     DatasetMetaTableUtil.setupDatasets(datasetFramework);
```
chtyim: Actually we can remove all these together since the defaultModules injection is done through a multibinder. Anyway, we can do it separate from this PR.

chtyim:
LGTM

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
bdmogal: Do you mean the ``setupDatasets`` method? It seems to actually add the ``datasets.instance`` and ``datasets.type`` datasets. They are different from the default modules?

PR #1719, opened 2015-03-19
Bugfix/drop old tables

bdmogal:
Jira: [CDAP-1802](https://issues.cask.co/browse/CDAP-1802)
Build: http://builds.cask.co/browse/CDAP-RBT191-1

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
@@ -57,7 +57,6 @@ import co.cask.cdap.logging.write.FileMetaDataManager;
 import co.cask.cdap.metrics.store.DefaultMetricDatasetFactory;
 import co.cask.cdap.notifications.feeds.client.NotificationFeedClientModule;
 import co.cask.cdap.proto.Id;
-import co.cask.cdap.proto.NamespaceMeta;
 import co.cask.tephra.TransactionExecutorFactory;
 import co.cask.tephra.distributed.TransactionService;
 import com.google.common.base.Throwables;
@@ -97,10 +96,11 @@ public class UpgradeTool {
   private final Configuration hConf;
   private final TransactionService txService;
   private final ZKClientService zkClientService;
-  private Store store;
-  private FileMetaDataManager fileMetaDataManager;
   private final Injector injector;
   private final HBaseTableUtil hBaseTableUtil;
+
+  private Store store;
+  private FileMetaDataManager fileMetaDataManager;
 
   /**
    * Set of Action available in this tool.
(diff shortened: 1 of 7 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -345,12 +344,16 @@ private Action parseAction(String action) {
 
   private void performUpgrade() throws Exception {
     LOG.info("Upgrading System and User Datasets ...");
+    HBaseAdmin hBaseAdmin = new HBaseAdmin(hConf);
     DatasetUpgrader dsUpgrade = injector.getInstance(DatasetUpgrader.class);
     dsUpgrade.upgrade();
+    hBaseTableUtil.dropTable(hBaseAdmin, dsUpgrade.getDatasetInstanceMDSUpgrader().getOldDatasetInstanceTableId());
```
anwar6953: if dropping tables right after each upgrade method, why not drop them within the Upgrader class?

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -423,10 +427,7 @@ private void createNamespaces() {
       Throwables.propagate(e);
     }
     LOG.info("Creating and registering {} namespace", Constants.DEFAULT_NAMESPACE);
-    getStore().createNamespace(new NamespaceMeta.Builder()
-                                 .setName(Constants.DEFAULT_NAMESPACE)
-                                 .setDescription(Constants.DEFAULT_NAMESPACE)
-                                 .build());
+    getStore().createNamespace(Constants.DEFAULT_NAMESPACE_META);
```
anwar6953: i know this is from before, but did you come across why the metastore needs this in it?

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
bdmogal: You need ``HBaseTableUtil``, ``HBaseAdmin``, ``Configuration`` objects to drop a table. I feel very strongly that classes such as ``FileMetadataManager`` should not know about these classes at all. 

If the Upgrader classes were subclasses of the ``AbstractUpgrader``, then I could have done by having these objects as protected members in ``AbstractUpgrader``. However, seeing that that is not possible under the current design, I found that this approach was a decent compromise for 2.8.0. 

We need to refactor this code to have the Upgrader classes obey a hierarchy eventually. Thoughts?

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
bdmogal: Didn't investigate further after we discussed. @rohitsinha54 do you know why this is needed?

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
anwar6953: Ok sounds good.

anwar6953:
LGTM.

PR #1722, opened 2015-03-19
Feature/ui create streams

ajainarayanan:
- Adds ui-bootstrap modal
- Adds the markup for create streams 
- Fixes the compatibility issues between ng-strap $modal and ui-bootstrap $modal. Makes both usable now in the application but under a different name for ui-bootstrap modal.

Note: Once this is merged into develop we need to run bower install to get the dependencies.

TODO:
  - Integrate with the backend, making one POST call to create a stream. Right now its hidden under /ns/default/streams .


nistala:
:+1:

PR #1725, opened 2015-03-19
Fix dropping namespace issue

chtyim:
- We should disable the stream table in Hive when dropping a stream
- Also fixed a annoying exception when StreamClient is used to fetch events.

### DIFF
--- a/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data/stream/StreamFileJanitorTestBase.java
+++ b/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data/stream/StreamFileJanitorTestBase.java
@@ -74,7 +74,7 @@ public abstract class StreamFileJanitorTestBase {
       writer.close();
 
       // Call cleanup before truncate. The current generation should stand.
-      janitor.clean(streamConfig, System.currentTimeMillis());
+      janitor.clean(streamConfig.getLocation(), streamConfig.getTTL(), System.currentTimeMillis());
       verifyGeneration(streamConfig, i);
 
       streamAdmin.truncate(streamId);
@@ -83,7 +83,7 @@ public abstract class StreamFileJanitorTestBase {
     int generation = StreamUtils.getGeneration(streamConfig);
     Assert.assertEquals(5, generation);
 
-    janitor.clean(streamConfig, System.currentTimeMillis());
+    janitor.clean(streamConfig.getLocation(), streamConfig.getTTL(), System.currentTimeMillis());
 
     // Verify the stream directory should only contains the generation directory
     for (Location location : streamConfig.getLocation().list()) {
@@ -126,13 +126,37 @@ public abstract class StreamFileJanitorTestBase {
 
     // Perform clean with current time = 10000 (10 seconds since epoch).
     // Since TTL = 5 seconds, 2 partitions will be remove (Ends at 2000 and ends at 4000).
-    janitor.clean(config, 10000);
+    janitor.clean(config.getLocation(), config.getTTL(), 10000);
 
     Assert.assertEquals(3, generationLocation.list().size());
 
     // Cleanup again with current time = 16000, all partitions should be deleted.
-    janitor.clean(config, 16000);
+    janitor.clean(config.getLocation(), config.getTTL(), 16000);
     Assert.assertTrue(generationLocation.list().isEmpty());
+  }
+
+  @Test
+  public void testCleanupDeletedStream() throws Exception {
+    Id.Stream streamId = Id.Stream.from(Constants.DEFAULT_NAMESPACE, "cleanupDelete");
+    StreamAdmin streamAdmin = getStreamAdmin();
+    StreamFileJanitor janitor = new StreamFileJanitor(getCConfiguration(), streamAdmin, getLocationFactory());
+    streamAdmin.create(streamId);
+
+    // Write some data
+    FileWriter<StreamEvent> writer = createWriter(streamId);
+    try {
+      for (int i = 0; i < 10; i++) {
+        writer.append(StreamFileTestUtils.createEvent(i * 1000, "Testing " + i));
+      }
+    } finally {
+      writer.close();
+    }
+
+    // Delete the stream
+    streamAdmin.drop(streamId);
+
+    // Run janitor. Should be running fine without exception.
+    janitor.cleanAll();
   }
 
   private void verifyGeneration(StreamConfig config, int generation) throws IOException {
--- a/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/transaction/stream/StreamAdminTest.java
+++ b/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/transaction/stream/StreamAdminTest.java
@@ -19,9 +19,11 @@ package co.cask.cdap.data2.transaction.stream;
 import co.cask.cdap.api.flow.flowlet.StreamEvent;
 import co.cask.cdap.data.file.FileWriter;
 import co.cask.cdap.data.stream.StreamFileWriterFactory;
+import co.cask.cdap.data.stream.StreamUtils;
 import co.cask.cdap.proto.Id;
 import com.google.common.base.Charsets;
 import com.google.common.collect.Lists;
+import org.apache.twill.filesystem.Location;
 import org.apache.twill.filesystem.LocationFactory;
 import org.junit.Assert;
 import org.junit.Test;
@@ -87,9 +89,9 @@ public abstract class StreamAdminTest {
 
     streamAdmin.dropAllInNamespace(Id.Namespace.from(FOO_NAMESPACE));
 
-    // All of the streams within the default namespace should have no data in them
+    // All of the streams within the default namespace should no longer exist
     for (Id.Stream defaultStream : fooStreams) {
-      Assert.assertEquals(0, getStreamSize(defaultStream));
+      Assert.assertFalse(streamAdmin.exists(defaultStream));
     }
     // otherStream isn't in the foo namespace so its data is not deleted in the above call to dropAllInNamespace.
     Assert.assertNotEquals(0, getStreamSize(otherStream));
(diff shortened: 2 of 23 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -111,6 +111,7 @@ public void dropAllInNamespace(Id.Namespace namespace) throws Exception {
 
     for (final Location streamLocation : locations) {
       doTruncate(streamLocation);
+      alterExploreStream(StreamUtils.getStreamIdFromLocation(streamLocation), false);
```
jwang47: Should this also be added to drop()? and then in here just call drop() instead of truncate and alterExploreStream?

chtyim: hum.. I thought it was done in drop...

chtyim: but apparently it is not

bdmogal:
LGTM

jwang47:
Can you add the following two test cases?

1. Test create/drop/create stream
  * After create stream: explore for stream should exist
2. Test create stream, delete ns, create ns, create stream
  * After first and list create stream, explore for stream should exist
  * After delete ns, explore for stream should not exist

I don't think (2) will pass right now..

cdap-data-fabric-tests/src/test/java/co/cask/cdap/data/stream/StreamFileJanitorTestBase.java
```diff
@@ -126,15 +126,39 @@ public void testCleanupTTL() throws Exception {
...
+      }
+    } finally {
+      writer.close();
+    }
+
+    // Delete the stream
+    streamAdmin.drop(streamId);
+
+    // Run janitor. Should be running fine without exception.
+    janitor.cleanAll();
```
jwang47: Check that stream no longer exists?

cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/transaction/stream/StreamAdminTest.java
```diff
@@ -89,7 +91,7 @@ public void testDropAllInNamespace() throws Exception {
 
     // All of the streams within the default namespace should have no data in them
```
jwang47: Update comment

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -204,14 +204,26 @@ public void upgrade() throws Exception {
...
-    return loadConfig(streamLocation);
+    Location configLocation = getConfigLocation(streamId);
+    Preconditions.checkArgument(configLocation.exists(), "Stream '%s' does not exist.", streamId);
+
+    StreamConfig config = GSON.fromJson(
+      CharStreams.toString(CharStreams.newReaderSupplier(Locations.newInputSupplier(configLocation), Charsets.UTF_8)),
+      StreamConfig.class);
+
+    int threshold = config.getNotificationThresholdMB();
+    if (threshold <= 0) {
```
jwang47: So the minimum notification threshold is 0? Is this a constant for this?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -298,10 +304,9 @@ public StreamConfig call() throws Exception {
   }
 
   private void assertNamespaceHomeExists(Id.Namespace namespaceId) throws IOException {
-    Location namespaceHomeLocation = Locations.getParent(getStreamsHomeLocation(namespaceId));
-    Preconditions.checkArgument(namespaceHomeLocation.exists(),
-                                String.format("Home directory %s for namespace %s not found",
-                                              namespaceHomeLocation.toURI().getPath(), namespaceId));
+    Location namespaceHomeLocation = Locations.getParent(getNamespaceLocation(namespaceId));
+    Preconditions.checkArgument(namespaceHomeLocation == null || namespaceHomeLocation.exists(),
```
anwar6953: If namespaceHomeLocation == null, the exception won't be thrown.

cdap-data-fabric-tests/src/test/java/co/cask/cdap/data/stream/StreamFileJanitorTestBase.java
chtyim: There are other tests doing that already. In this Janitor test, it's mainly to verify the Janitor can handle the case correctly.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
chtyim: It's existing code. It is the default value of long field when deserializing using Gson and yet the field is missing.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
chtyim: good catch. I was thinking about it as a `if`. Precondition should flip the condition.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT193-10

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/AbstractStreamCoordinatorClient.java
```diff
@@ -81,6 +82,11 @@
    */
   protected abstract void streamCreated(Id.Stream streamId);
 
+  /**
+   * GEts invoked when a stream is deleted.
```
bdmogal: typo

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamFileJanitor.java
```diff
@@ -115,4 +120,12 @@ void clean(StreamConfig config, long currentTime) throws IOException {
   private boolean isPartitionDirector(Location location) throws IOException {
     return (location.isDirectory() && location.getName().indexOf('.') > 0);
   }
+
+  private boolean isStreamExists(Id.Stream streamId) throws IOException {
```
bdmogal: Only for converting exceptions to IOException?

chtyim: Yes, to simplify the try catch logic in the `clean` method

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -326,42 +331,61 @@ private void createStreamFeeds(StreamConfig config) {
...
+   */
+  private Location getStreamLocation(Id.Stream streamId) throws IOException {
+    return getNamespaceLocation(streamId.getNamespace()).append(streamId.getName());
   }
 
-  private Location getStreamsHomeLocation(Id.Namespace namespace) throws IOException {
+  /**
+   * Returns the location for the given namespace that contains all streams belong to that namespace.
+   */
+  private Location getNamespaceLocation(Id.Namespace namespace) throws IOException {
```
bdmogal: This might give a wrong idea, no? Its actually returning the streams directory under the namespace, and not the namespace location?

bdmogal: Maybe ``getStreamsLocationInNamespace``?

jwang47:
LGTM

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
chtyim: It's a private method and it's inside the StreamAdmin. We don't need to name everything in a private method name as the class already gives the context that everything inside this class should be just operating on stream. Also, the javadoc says what it returns.

chtyim: I rename it to `getStreamBaseLocation`.

anwar6953: the method that takes in a streamId and returns: /streams/fooStream dir is also named similarly.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamFileJanitor.java
bdmogal: I see. Do we care about other exceptions that ``StreamAdmin.exists`` may throw in other parts of the code? If not maybe we can have this cleanup in ``StreamAdmin.exists`` itself? May be in a different PR though.

bdmogal:
@chtyim only minor comments from me. LGTM.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamFileJanitor.java
chtyim: `cleanAll` itself throws `IOException`. The caller needs to handle that. In production code, the caller is actually a executor thread that schedule to do cleanup and it just log the exception and schedule it again (hence if it's interim failure, it'll get retried when the janitor runs again in next scheduled time).

bdmogal: Right. 

I was only wondering why ``StreamAdmin.exists`` should throw anything other than an ``IOException``. I just checked that ``HBaseQueueAdmin.exists`` throws a ``DatasetManagementException``, so that's probably why.

chtyim: It's a legacy issue. The `StreamAdmin` used to extend from `DatasetAdmin`, which throws `Exception` in all methods. When we detach `StreamAdmin` from `DatasetAdmin`, those method signature get retained.

bdmogal: I see. ok.

PR #1726, opened 2015-03-19
show modal when backend stops running

elfenheart:
Showing modal to disable UI when backend is not running. 
Ping backend every minute


ajainarayanan:
:+1: LGTM

PR #1734, opened 2015-03-20
upgrade metrics kafka tables

gsps1:
https://issues.cask.co/browse/CDAP-1858

Build: http://builds.cask.co/browse/CDAP-RBT194-1


cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+                                              MetricsTable.class.getName(), props, null, null);
+    } catch (Exception e) {
+      LOG.error("Exception while creating table {}.", tableName, e);
+    }
+    return table;
+  }
+
+  @Override
+  public void upgrade() throws Exception {
+    // copy kafka off set from old table to new kafka metrics table
```
abaranau: offset

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+
+    String kafkaTableNameOld = Joiner.on(".").join(Constants.SYSTEM_NAMESPACE, kafkaTableName);
+    byte[] columnFamily = getColumnFamily(kafkaTableNameOld);
+    HTable hTable = getHTable(kafkaTableNameOld);
```
abaranau: can it return null?

abaranau: or what happens when there's no table to migrate

abaranau: should not fail

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+
+    LOG.info("Starting upgrade for table {}", Bytes.toString(hTable.getTableName()));
+    // iterate old table and add those rows to the new kafkaConsumerMetaTable
```
abaranau: copy *all*

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
abaranau: also: add a comment: assuming we are migrating from 2.6 or 2.7

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+    try {
+      Scan scan = getScan(columnFamily);
+      ResultScanner resultScanner = hTable.getScanner(scan);
+      Result result;
+      try {
+        while ((result = resultScanner.next()) != null) {
+          TopicPartition topicPartition =  getTopicPartition(result.getRow());
```
abaranau: extra space

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+          if (topicPartition != null) {
+            long value  = Bytes.toLong(result.getFamilyMap(columnFamily).get(OFFSET_COLUMN));
+            kafkaMetaTableDestination.save(ImmutableMap.of(topicPartition, value));
+          } else {
+            LOG.warn("Invalid topic partition found");
```
abaranau: print which one

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+          }
+        }
+        LOG.info("Successfully completed upgrade for table {}", Bytes.toString(hTable.getTableName()));
+      } finally {
+        resultScanner.close();
+      }
+    } catch (Exception e) {
+      LOG.info("Exception during upgrading metrics-kafka table {}", e);
```
abaranau: should we fail the upgrade job

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
abaranau: try right after creating hTable

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+    return table;
+  }
+
+  @Override
+  public void upgrade() throws Exception {
+    // copy kafka off set from old table to new kafka metrics table
+
+    String kafkaTableNameOld = Joiner.on(".").join(Constants.SYSTEM_NAMESPACE, kafkaTableName);
+    byte[] columnFamily = getColumnFamily(kafkaTableNameOld);
```
abaranau: do getHTable first - and go out if it is not there. Otherwise getColumnFamily will create a table if it is not there... We don't want that

abaranau:
basically one comment - do check if htable is ther first

abaranau:
and then merege

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private final Configuration hConf;
+  private final HBaseTableUtil hBaseTableUtil;
+  private final DatasetFramework dsFramework;
+  private final String kafkaTableName;
+  private final KafkaConsumerMetaTable kafkaMetaTableDestination;
+
+  @Inject
+  public MetricsKafkaUpgrader(CConfiguration cConf, Configuration hConf, LocationFactory locationFactory,
+                              HBaseTableUtil hBaseTableUtil, @Named("dsFramework") final DatasetFramework dsFramework) {
```
chtyim: Remove the `@Named`. I think @bdmogal fixed the Guice injection so that the `dsFramework` is no longer needed.

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,155 @@
...
+    super(locationFactory);
+    this.hConf = hConf;
+    this.hBaseTableUtil = hBaseTableUtil;
+    this.dsFramework = dsFramework;
+    this.kafkaTableName =  cConf.get(MetricsConstants.ConfigKeys.KAFKA_META_TABLE,
+                                     MetricsConstants.DEFAULT_KAFKA_META_TABLE);
+    this.kafkaMetaTableDestination = new DefaultMetricDatasetFactory(cConf, dsFramework).createKafkaConsumerMeta();
```
chtyim: Will this acquire resources (e.g. open HTable connection?) Does it requires closing (it's a Closeable, right?)

Ideally resources shouldn't be acquired in constructor. Since you only use it in the upgrade method, new, user and close it there.

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Upgrade kafka metrics meta table data
+ */
+public class MetricsKafkaUpgrader extends AbstractUpgrader {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DatasetUpgrader.class);
+  private static final byte[] OFFSET_COLUMN = Bytes.toBytes("o");
```
rohitsinha54: What is this ?

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,167 @@
...
+      LOG.error("Exception while creating table {}.", tableName, e);
+    }
+    return table;
+  }
+
+  @Override
+  public void upgrade() throws Exception {
+    KafkaConsumerMetaTable kafkaMetaTableDestination =
+      new DefaultMetricDatasetFactory(cConf, dsFramework).createKafkaConsumerMeta();
+    // copy kafka offset from old table to new kafka metrics tabl
```
rohitsinha54: table

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,167 @@
...
+    String kafkaTableNameOld = Joiner.on(".").join(Constants.SYSTEM_NAMESPACE, "default", kafkaTableName);
```
rohitsinha54: oldkafkaTableName or just oldTableName

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,167 @@
...
+          }
+          LOG.info("Successfully completed upgrade for table {}", Bytes.toString(hTable.getTableName()));
+        } finally {
+          resultScanner.close();
+        }
+      } catch (Exception e) {
+        LOG.info("Exception during upgrading metrics-kafka table {}", e);
+        throw Throwables.propagate(e);
+      } finally {
+        hTable.close();
```
rohitsinha54: We are also dropping all the old meta tables. That needs to be added appropriately.

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,167 @@
...
+    super(locationFactory);
+    this.cConf = cConf;
+    this.hConf = hConf;
+    this.hBaseTableUtil = hBaseTableUtil;
+    this.dsFramework = dsFramework;
+    this.kafkaTableName =  cConf.get(MetricsConstants.ConfigKeys.KAFKA_META_TABLE,
+                                     MetricsConstants.DEFAULT_KAFKA_META_TABLE);
+  }
+
+  private MetricsTable getOrCreateKafkaTable(String tableName, DatasetProperties props) {
```
rohitsinha54: If its a private method and the props is always DatasetProperties.EMPTY then why its a parameter to the function ?

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,167 @@
...
+        hTable.close();
+      }
+    } catch (IOException e) {
+      LOG.info("Unable to find table {}", kafkaTableNameOld);
+    }
+  }
+
+  private byte[] getColumnFamily(String tableName) throws DatasetManagementException {
+    // add old table for getting dataset spec
+    getOrCreateKafkaTable(tableName, DatasetProperties.EMPTY);
```
rohitsinha54: You don't use the return value and since the method is just private and only called from here you can make it return void

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
gsps1: this is the column key used to write the kafka offsets

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,167 @@
...
+    } catch (Exception e) {
+      LOG.error("Exception while creating table {}.", tableName, e);
+    }
+    return table;
+  }
+
+  @Override
+  public void upgrade() throws Exception {
+    KafkaConsumerMetaTable kafkaMetaTableDestination =
+      new DefaultMetricDatasetFactory(cConf, dsFramework).createKafkaConsumerMeta();
```
chtyim: Should be a try-finally to close this dataset.

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,167 @@
...
+      HTable hTable = getHTable(kafkaTableNameOld);
+      byte[] columnFamily = getColumnFamily(kafkaTableNameOld);
+
+      // iterate old table and copy all rows to the new kafkaConsumerMetaTable
+      try {
+        LOG.info("Starting upgrade for table {}", Bytes.toString(hTable.getTableName()));
+        Scan scan = getScan(columnFamily);
+        ResultScanner resultScanner = hTable.getScanner(scan);
+        Result result;
+        try {
```
chtyim: Swap this and the line before. It's generally a good practice you have the 'try' begin immediately after the resource creation that you need to free in finally.

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
gsps1: good point, we are not implementing closeable on the KafkaConsumerMetaTable right now, we should, i will open a separate PR for that, as there are similar cases in FactTable where we don't close the metrics table. have added todo here now.

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
gsps1: done

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
gsps1: done

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
```diff
@@ -0,0 +1,175 @@
...
+
+  @Override
+  public void upgrade() throws Exception {
+    // todo : close the KafkaConsumerMetaTable after it implements closeable
+    KafkaConsumerMetaTable kafkaMetaTableDestination =
+      new DefaultMetricDatasetFactory(cConf, dsFramework).createKafkaConsumerMeta();
+    // copy kafka offset from old table to new kafka metrics table
+
+    try {
+      // assuming we are migrating from 2.6 or 2.7
```
abaranau: from 2.6 only, no? based on table name - you one try one format, no?

abaranau:
LGTM

cdap-master/src/main/java/co/cask/cdap/data/tools/MetricsKafkaUpgrader.java
gsps1: ouch. meant to type 2.6 to 2.8. will fix

PR #1735, opened 2015-03-20
Fix bullet points et al

awholegunch:
Removes mention of packaging info that never completed; adds mention of the custom logbook documentation.


poornachandra:
LGTM

PR #1736, opened 2015-03-20
create default dir also

rohitsinha54:


cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -433,6 +435,7 @@ private void createNamespaces() {
     }
     LOG.info("Creating and registering {} namespace", Constants.DEFAULT_NAMESPACE);
     getStore().createNamespace(Constants.DEFAULT_NAMESPACE_META);
+    locationFactory.create(Constants.DEFAULT_NAMESPACE);
```
bdmogal: This will not create the directory. Should call ``mkdirs`` on the object returned by ``create``

rohitsinha54:
There's other stuff which needs to be done. Will re-open again

PR #1738, opened 2015-03-20
Bugfix/namespace dir

bdmogal:
Jira: [CDAP-1875](https://issues.cask.co/browse/CDAP-1875)
Build: http://builds.cask.co/browse/CDAP-RBT196-6

Thorough testing pending, but want to get early feedback.

### DIFF
--- a/cdap-watchdog/src/test/java/co/cask/cdap/logging/write/LogCleanupTest.java
+++ b/cdap-watchdog/src/test/java/co/cask/cdap/logging/write/LogCleanupTest.java
@@ -66,12 +66,15 @@ public class LogCleanupTest {
   private static Injector injector;
   private static TransactionManager txManager;
   private static String logBaseDir;
+  private static String namespacesDir;
 
   @BeforeClass
   public static void init() throws Exception {
     Configuration hConf = HBaseConfiguration.create();
     final CConfiguration cConf = CConfiguration.create();
     cConf.set(Constants.CFG_LOCAL_DATA_DIR, TEMP_FOLDER.newFolder().getAbsolutePath());
+    logBaseDir = cConf.get(LoggingConfiguration.LOG_BASE_DIR);
+    namespacesDir = cConf.get(Constants.Namespace.NAMESPACES_DIR);
     injector = Guice.createInjector(
       new ConfigModule(cConf, hConf),
       new LocationRuntimeModule().getInMemoryModules(),
@@ -82,7 +85,6 @@ public class LogCleanupTest {
 
     txManager = injector.getInstance(TransactionManager.class);
     txManager.startAndWait();
-    logBaseDir = injector.getInstance(CConfiguration.class).get(LoggingConfiguration.LOG_BASE_DIR);
   }
 
   @AfterClass
@@ -105,7 +107,7 @@ public class LogCleanupTest {
     // Setup directories
     LoggingContext dummyContext = new FlowletLoggingContext("ns", "app", "flw", "flwt");
 
-    Location namespacedLogsDir = baseDir.append("ns").append(logBaseDir);
+    Location namespacedLogsDir = baseDir.append(namespacesDir).append("ns").append(logBaseDir);
     Location contextDir = namespacedLogsDir.append("app").append("flw");
     List<Location> toDelete = Lists.newArrayList();
     for (int i = 0; i < 5; ++i) {
@@ -145,7 +147,7 @@ public class LogCleanupTest {
     Assert.assertEquals(locationListsToString(toDelete, notDelete),
       toDelete.size() + notDelete.size(), fileMetaDataManager.listFiles(dummyContext).size());
 
-    LogCleanup logCleanup = new LogCleanup(fileMetaDataManager, baseDir, RETENTION_DURATION_MS);
+    LogCleanup logCleanup = new LogCleanup(fileMetaDataManager, baseDir, namespacesDir, RETENTION_DURATION_MS);
     logCleanup.run();
     logCleanup.run();
 
@@ -168,8 +170,8 @@ public class LogCleanupTest {
     // Create base dir
     Location baseDir = injector.getInstance(LocationFactory.class).create(TEMP_FOLDER.newFolder().toURI());
     // Create namespaced logs dirs
-    Location namespacedLogsDir1 = baseDir.append("ns1").append(logBaseDir);
-    Location namespacedLogsDir2 = baseDir.append("ns2").append(logBaseDir);
+    Location namespacedLogsDir1 = baseDir.append(namespacesDir).append("ns1").append(logBaseDir);
+    Location namespacedLogsDir2 = baseDir.append(namespacesDir).append("ns2").append(logBaseDir);
 
     // Create dirs with files
     Set<Location> files = Sets.newHashSet();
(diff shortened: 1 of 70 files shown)


cdap-watchdog/src/test/java/co/cask/cdap/logging/write/LogCleanupTest.java
```diff
@@ -235,18 +235,18 @@ public void testDeleteEmptyDir2() throws Exception {
...
-    logCleanup.deleteEmptyDir("ns/" + logBaseDir, baseDir);
+    logCleanup.deleteEmptyDir("namespaces" + "/ns/" + logBaseDir, baseDir);
     // Assert base dir exists
     Assert.assertTrue(baseDir.exists());
 
     baseDir.mkdirs();
     // Assert root exists
     Assert.assertTrue(baseDir.exists());
-    logCleanup.deleteEmptyDir("ns/" + logBaseDir, baseDir);
+    logCleanup.deleteEmptyDir("namespaces" + "/ns/" + logBaseDir, baseDir);
```
anwar6953: don't hardcode this here. can you use Namespaces util method? at the very minimum, get it from con?

bdmogal: Agreed, missed it in the last commit. Fixed.

cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.twill.filesystem.Location;
+import org.apache.twill.filesystem.LocationFactory;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Utilities for namespaces
+ */
+public class Namespaces {
```
chtyim: final class

cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java
```diff
@@ -0,0 +1,49 @@
...
+  private Namespaces() {
+  }
+
+  public static Location getNamespaceHomeLocation(LocationFactory locationFactory, CConfiguration cConf,
```
chtyim: Javadoc is needed for public methods.

cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java
```diff
@@ -0,0 +1,49 @@
...
+                                                  Id.Namespace namespaceId) throws IOException {
+    return locationFactory.create(getNamespacesDir(cConf)).append(namespaceId.getId());
+  }
+
+  public static String getNamespacesDir(CConfiguration cConf) {
```
chtyim: Better not to have method like this, as it relies on caller to construct the `Location` correctly. You should have it take a `LocationFactory` and returns a `Location`.

cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java
```diff
@@ -0,0 +1,49 @@
...
+    return cConf.get(Constants.Namespace.NAMESPACES_DIR);
+  }
+
+  public static File getLocalTemporaryDirectory(CConfiguration cConf, Id.Namespace namespaceId) {
+    String tempBase = String.format("%s/%s/%s", cConf.get(Constants.CFG_LOCAL_DATA_DIR),
```
chtyim: Don't use string format to form path as it'll break in Windows. Always use `File` to do so:

```java
File localDataDir = new File(cConf.get(Constants.CFG_LOCAL_DATA_DIR));
File namespaceBase = new File(localDataDir, cConf.get(Constants.Namespace.NAMESPACES_DIR));
return new File(new File(namespaceBase, namespaceId.toString()),
```

cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java
chtyim: Instead of having everyone to call this method, you can provide a method that creates a `LocationFactory` that operates inside a given namespace:

```java
static LocationFactory createNamespacedLocationFactory(LocationFactory locationFactory,
                                                       CConfiguration cConf, Id.Namespace namespace) {
  LocationFactory namespaceBaseFactory = LocationFactories.namespace(locationFactory,
                                                cConf.get(Constants.Namespace.NAMESPACES_DIR));
  return LocationFactories.namespace(namespaceBaseFactory, namespace.getId());
}

File: cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java (hunk 5 above)
Comment (bdmogal) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: Awesome. Thanks. Yes, needs to be fixed in ``FilesetAdmin`` as well.

File: cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java (hunk 3 above)
Comment (chtyim) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: I believe it'll simplify caller code as it doesn't need to keep passing in cConf and namespace for every call

Comment (anwar6953) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: A lot of the usages of this method check exists on the Location that was being returned.
This way, the caller would have to do:
`createNamespacedLocationFactory(...).create("").exists()`, which I think is messier.

Comment (chtyim) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: You can have both.

Comment (chtyim) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: Also, you don't need to modify `Programs.programLocation`.

Comment (anwar6953) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: You mean I don't need to modify that method, compared to whats on release/2.8, or whats already on this branch?

Comment (chtyim) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: I mean, if you have the `createNamespacedLocationFactory`, then there is no need to modify the Programs class in this PR.

Comment (anwar6953) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: the Programs class  (even before this PR) expects a locationFactory and does `lf.create(namespace`.

Comment (chtyim) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: I know, but it doesn't take `cConf`

Comment (anwar6953) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: ok, so there should be a method that namespaces with the "namespaces" dir, and expects caller to  `lf.create(namespace)`?
(thats different than `createNamespacedLocationFactory` method you wrote above.

Comment (anwar6953) [inline cdap-common/src/main/java/co/cask/cdap/common/namespace/Namespaces.java]: So a method:
```
static LocationFactory createNamespacedLocationFactory(LocationFactory locationFactory,
                                                       CConfiguration cConf, Id.Namespace namespace) {
return LocationFactories.namespace(locationFactory, cConf.get(Constants.Namespace.NAMESPACES_DIR));
}
```

chtyim: That can do it.

anwar6953: Ok, I tried it. 
The callers of the method in Programs class would need updating to pass in a different locationFactory.
and I think its better to just modify Programs class, instead of the 4 places that use that method.

anew:
Overall, changes look good to me. My only comment is that the logic of how to build a certain path is often replicated to different code locations. (That was the case in existing code, I know).

bdmogal:
@anew Tried to fix up and centralize all that logic in the last commit per @chtyim's suggestions

chtyim:
Probably we can merge it for now and do the proper refactoring to reduce code duplication and have better separation of concern (the NamespacedLocationFactory should have more logic rather than just constructing paths).

bdmogal:
@chtyim meaning keep the last commit as well for now right? with ``NamespacedLocationFactory``?

Also, what logic are you referring to?

chtyim:
Yes.

bdmogal:
Green - http://builds.cask.co/browse/CDAP-RBT196-21

anwar6953:
:+1:

PR #1739, opened 2015-03-20
Setting number of partitions for log topic in producer

poornachandra:


cdap-watchdog/src/main/java/co/cask/cdap/logging/LoggingConfiguration.java
```diff
@@ -54,7 +54,7 @@
   public static final String DEFAULT_LOG_PATTERN = "%d{ISO8601} - %-5p [%t:%C{1}@%L] - %m%n";
   public static final String DEFAULT_KAFKA_PRODUCER_TYPE = "async";
   public static final long DEFAULT_KAFKA_PROCUDER_BUFFER_MS = 1000;
-  public static final String DEFAULT_NUM_PARTITIONS = "10";
+  public static final String DEFAULT_NUM_PARTITIONS = "50";
```
chtyim: Remove default defined in class. Defaults are defined in `cdap-default.xml`. (You don't have to move all the defaults now, but since you are changing this one, it's good to start moving).

See https://wiki.continuuity.com/display/ENG/Configuration+Standards

rohitsinha54:
Closed because we think this will not solve this issue.

PR #1741, opened 2015-03-20
CDAP-1883 upgrade tool also creates default dir on hdfs

rohitsinha54:
Issue: https://issues.cask.co/browse/CDAP-1883
Build: http://builds.cask.co/browse/CDAP-RBT195-4


anwar6953:
LGTM

PR #1744, opened 2015-03-20
Hotfix/revert comma

gokulavasan:
* Preferences/RuntimeArguments CLI will use ' ' (space) as the separator between key-value pairs instead of ','.
* Enabling Namespace CLI Test.

Bamboo : http://builds.cask.co/browse/CDAP-DUT1204-1

Opened JIRA to address this in 3.0 : https://issues.cask.co/browse/CDAP-1886


jwang47:
I thought we decided to use `,` instead of ` ` as the separator?

sreevatsanraman:
@jwang47 - we need to fix it cleanly. The plan is to have key="value" so that we can support commas and spaces for the values. Will be fixed in next release.

jwang47:
LGTM

PR #1751, opened 2015-03-20
Use Text as map output for the SportResults app

jguery:
Workaround for JIRA: https://issues.cask.co/browse/CDAP-1887


albertshau:
:+1:

PR #1752, opened 2015-03-20
Use log base dir while upgrading log metadata

rohitsinha54:


rohitsinha54:
@bdmogal : Closing this as you said you will include this in another PR.

PR #1753, opened 2015-03-20
(CDAP-1893) Fix a bug introduced when converting HBase queue state store...

chtyim:
... to dataset

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueConsumer.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueConsumer.java
@@ -137,19 +137,11 @@ public abstract class AbstractQueueConsumer implements QueueConsumer, Transactio
   @Override
   public DequeueResult<byte[]> dequeue(int maxBatchSize) throws IOException {
     DequeueResult<byte[]> result = performDequeue(maxBatchSize);
-    if (scanStartRow != null) {
-      if (!consumingEntries.isEmpty()) {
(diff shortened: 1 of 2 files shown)


chtyim:
Working on adding a test case for that. But probably won't get in for 2.8.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueConsumer.java
```diff
@@ -137,19 +137,11 @@ public ConsumerConfig getConfig() {
...
-          updateStartRow(floorKey);
-        }
-      } else {
-        // If the dequeue has empty result, startRow can advance to scanStartRow
-        updateStartRow(scanStartRow);
-      }
-    }
+    // Start row can be updated to the largest rowKey in the consumingEntries
+    // that is smaller than or equal to scanStartRow. If no such key exists, update start row to scanStartRow
+    byte[] floorKey = consumingEntries.floorKey(scanStartRow);
```
albertshau: old code checks that scanStartRow isn't null but this doesn't.  Has it changed to never be null?

albertshau:
one question about possible NPE, otherwise lgtm

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueConsumer.java
chtyim: It never null now. It get initialized in constructor.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT201-1

PR #1758, opened 2015-03-21
Ask user for confirmation before upgrade 

rohitsinha54:
List what all will be upgraded 
and ask for confirmation.


chtyim:
LGTM

PR #1762, opened 2015-03-21
Fixed the program controller state transition.

sagarkapare:
Jira - https://issues.cask.co/browse/CDAP-1898

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractTwillProgramController.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractTwillProgramController.java
@@ -33,6 +33,7 @@ abstract class AbstractTwillProgramController extends AbstractProgramController 
 
   protected final String programName;
   protected final TwillController twillController;
+  private volatile boolean stopRequested = false;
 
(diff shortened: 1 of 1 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractTwillProgramController.java
```diff
@@ -33,6 +33,7 @@
 
   protected final String programName;
   protected final TwillController twillController;
+  private volatile boolean stopRequested = false;
```
chtyim: No need to init to `false`.

chtyim:
Just one comment,.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractTwillProgramController.java
sagarkapare: Fixed.

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-RBT206-2

sagarkapare:
Build passed. Merging.

PR #1765, opened 2015-03-21
CDAP-1902 enabling explore on a dataset that is not explorable

albertshau:
should be a no-op.


albertshau:
http://builds.cask.co/browse/CDAP-RBT207

sreevatsanraman:
LGTM :+1:

PR #1768, opened 2015-03-21
check table exists before upgrade/delete kafka metrics table

gsps1:
if table does not exist, we just return, previously we would throw exception which is not in contract with UpgradeTool.

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
@@ -122,6 +122,7 @@ public class UpgradeTool {
               "  6. Logs Metadata\n" +
               "  7. Stream state store table\n" +
               "  8. Queue config table\n" +
+              "  9. Metrics Kafka table\n" +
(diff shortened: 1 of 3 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -401,8 +402,10 @@ private void performUpgrade() throws Exception {
 
     LOG.info("Upgrading metrics.kafka.meta table ...");
     MetricsKafkaUpgrader metricsKafkaUpgrader = injector.getInstance(MetricsKafkaUpgrader.class);
-    metricsKafkaUpgrader.upgrade();
-    hBaseTableUtil.dropTable(hBaseAdmin, metricsKafkaUpgrader.getOldKafkaMetricsTableId());
+    if (metricsKafkaUpgrader.tableExists()) {
```
bdmogal: this should be tableUtil.tableExists, right?
-ali

bdmogal:
:+1:

PR #1769, opened 2015-03-21
Move version endpoint to v3

jwang47:
http://builds.cask.co/browse/CDAP-RBT208


albertshau:
should deprecate v2 instead of removing right? unless its not used anywhere?

albertshau:
after speaking to @sreevatsanraman and @jwang47, we're just going to move this out

PR #1770, opened 2015-03-21
revert notes about secure explore since it is only well tested

albertshau:
on a single distribution version and known to be broken on many
others.


sreevatsanraman:
LGTM :+1:

PR #1771, opened 2015-03-22
[CDAP-1906] Ask for confirmation before startup. Also don't startup upgr...

bdmogal:
...ade tool when calling help.

Jira: [CDAP-1906](https://issues.cask.co/browse/CDAP-1906)
Build: http://builds.cask.co/browse/CDAP-RBT209-1

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
@@ -117,7 +117,7 @@ public class UpgradeTool {
               "  The upgrade tool upgrades the following: \n" +
               "  1. User Datasets (Upgrades only the coprocessor jars)\n" +
               "  2. System Datasets\n" +
-              "  3. Dataset Type and Instance MetadataStore\n" +
+              "  3. Dataset Type and Instance Metadata\n" +
               "  4. Application Metadata\n" +
               "  5. Archives and Files\n" +
               "  6. Logs Metadata\n" +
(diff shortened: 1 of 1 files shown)


sreevatsanraman:
LGTM :+1: Please merge once the build passes.

rohitsinha54:
This just delays the startup process which prints stuff. But we are still creating the Upgrade tool object and doing all the guice binding. I don't think we should be doing guice binding and stuff before user confirmation. But if we are fine with this then no issues.

ex: 
2015-03-22 02:45:03,826 - INFO  [main:c.c.c.c.g.LocationRuntimeModule$HDFSLocationModule@88] - HDFS namespace is /cdap
2015-03-22 02:45:03,994 - WARN  [main:o.a.h.u.NativeCodeLoader@62] - Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
UpgradeTool - version 2.8.0-SNAPSHOT-1426991840180.

anwar6953:
@rohitsinha54 
this seems ok to me:
https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/QvWJzYWz0IzCcRu/Screen%20Shot%202015-03-21%20at%208.26.06%20PM.png

Just loading classes should be fine before asking.

bdmogal:
@rohitsinha54 Point taken. However, we could probably take it a step further too then, show all this in the shell script and only launch the JVM if needed? If we're loading a class, then loading it fully is ok IMO. I think we can fix this later to maybe not do the Guice binding in the constructor (have it in the startup method perhaps?), so it becomes even cleaner. Thoughts?

bdmogal:
Build passed. Merging.

PR #1772, opened 2015-03-22
Bugfix/queue config upgrader idempotency

anwar6953:
The Queue config table's rowkey format is different in 2.8.
Running the same logic to parse 2.6 rowkeys on the new rowkey caused it to fail.
Now ignoring those rows (that are already potentially upgraded).

http://builds.cask.co/browse/CDAP-RBT210-4

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/AbstractQueueUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/AbstractQueueUpgrader.java
@@ -18,6 +18,8 @@ package co.cask.cdap.data.tools;
 
 import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.common.namespace.NamespacedLocationFactory;
+import co.cask.cdap.common.utils.ProjectInfo;
+import co.cask.cdap.data2.dataset2.lib.hbase.AbstractHBaseDataSetAdmin;
 import co.cask.cdap.data2.transaction.queue.QueueEntryRow;
 import co.cask.cdap.data2.util.TableId;
 import co.cask.cdap.data2.util.hbase.HBaseTableUtil;
@@ -76,6 +78,13 @@ public abstract class AbstractQueueUpgrader extends AbstractUpgrader {
       return;
     }
     HTable hTable = tableUtil.createHTable(conf, tableId);
+    ProjectInfo.Version tableVersion = AbstractHBaseDataSetAdmin.getVersion(hTable.getTableDescriptor());
+    // Only upgrade if Upgrader's version is greater than table's version.
+    if (ProjectInfo.getVersion().compareTo(tableVersion) <= 0) {
+      LOG.info("Table {} has already been upgraded. Its version is: {}", tableId, tableVersion);
+      return;
+    }
+
     LOG.info("Starting upgrade for table {}", Bytes.toString(hTable.getTableName()));
     try {
       Scan scan = new Scan();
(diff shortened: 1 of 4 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/QueueConfigUpgrader.java
```diff
@@ -61,7 +61,13 @@ protected TableId getTableId() {
 
   @Nullable
   private QueueName fromRowKey(byte[] oldRowKey) {
-    QueueName queueName = QueueName.from(oldRowKey);
+    QueueName queueName;
+    try {
+      queueName = QueueName.from(oldRowKey);
```
chtyim: That's not a fool proof check. Actually for 2.6->2.8 upgrade, you can check if the version exists in the table attribute. Only 2.8 table has that attribute set.

anwar6953: This wouldn't solve the case where the upgrade tool crashes (or is cancelled) mid-upgrade, right (half the rows converted, other half not).

chtyim: The rewrite of the data happens before or after the table get upgrade to dataset table (i.e. before or after QueueAdmin.upgrade() is called)? If it is before, then it should be sufficient to just check the version.

chtyim: Ideally the actual rewriting of data (scan and copy) happens in one place (QueueAdmin.upgrade()) so that all knowledge of the upgrade is in one place.

anwar6953: Right, but if half the rows get converted, and then the tool is killed, the next time it runs, it would error, right?

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -407,6 +409,12 @@ private void performUpgrade() throws Exception {
     QueueConfigUpgrader queueConfigUpgrader = injector.getInstance(QueueConfigUpgrader.class);
     queueConfigUpgrader.upgrade();
 
+    // Upgrade all stream and queue tables.
+    StreamAdmin streamAdmin = injector.getInstance(StreamAdmin.class);
```
bdmogal: should we do this in ``QueueConfigUpgrader.upgrade`` and ``StreamStateStore.upgrade``?

bdmogal:
1 question, otherwise LGTM

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
anwar6953: done

cdap-master/src/main/java/co/cask/cdap/data/tools/AbstractQueueUpgrader.java
```diff
@@ -76,6 +78,13 @@ void upgrade() throws Exception {
       return;
     }
     HTable hTable = tableUtil.createHTable(conf, tableId);
+    ProjectInfo.Version version = AbstractHBaseDataSetAdmin.getVersion(hTable.getTableDescriptor());
+    if (version.getMajor() != 0) {
```
chtyim: Better compare with the current project version

bdmogal:
just confirming  this scenario again:
1. Upgrade runs, upgrades 2 records then fails. Does not call queueAdmin.upgrade so version is not updated.
2. Upgrade runs again, finds two records that are updated - I think we already ignore them.

Now what happens if we started the flow between 1 and 2? Should we document that we shouldn't do that?


Other than that, :+1:

anwar6953:
Yea, (1) and (2) are fine.
The last case you mentioned is valid (I don't think its possible to avoid an error in that case).

PR #1775, opened 2015-03-22
Feature/remove service worker

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-1544
Bamboo : http://builds.cask.co/browse/CDAP-DUT1230

This includes changes from Gokul and Andreas.
- removes service workers
- removes "runnable" from code where possible. Removes some deprecated internal methods on the fly.
- In logging and metrics, renames "runnable" to "handler". We have already been using the handler names but tagged them as "runnables".


gokulavasan:
Build is green.

gokulavasan:
@poornachandra Please review. Build is green after merging develop and fixing a test.

anwar6953:
ProgramClient#getServiceRunnableInstances and ProgramClient#setServiceRunnableInstances methods can also be removed.

anwar6953:
Is this comment still accurate?
https://github.com/caskdata/cdap/blob/983815f8dc7512d87cfbd7a15618f8aa78850ca9/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/ServiceProgramRunner.java#L42

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1004,21 +1002,10 @@ void getServiceInstances(HttpResponder responder,
       }
 
       // If the runnable name is the same as the service name, then uses the service spec, otherwise use the worker spec
```
anwar6953: Also remove the comment, since the code for the comment is removed.

anwar6953:
Updates comments:
https://github.com/caskdata/cdap/blob/983815f8dc7512d87cfbd7a15618f8aa78850ca9/cdap-api/src/main/java/co/cask/cdap/api/service/Service.java#L23
and
https://github.com/caskdata/cdap/blob/983815f8dc7512d87cfbd7a15618f8aa78850ca9/cdap-api/src/main/java/co/cask/cdap/api/service/Service.java#L29

anwar6953:
I recommend getting confirmation from UI guys that the old UI is not sticking around for 3.0.
Otherwise, if that UI is going to be provided as an alternative, service workers also need to be removed the the UI (file a JIRA, if so).

gokulavasan:
Closing this PR. Will reopen with full cleanup of Service Worker.

anew:
@poorna please review the logging changes. Per our discussion yesterday, the rename should not have an effect on querying/retrieving logs, because we don't partition of filter by "userrunnable". The new tag "userhandler" will only show as a tag in the log messages.

anew:
@abaranau please review metrics changes. I am not sure whether it is ok to overload "handlers" both for user services handlers and system handlers (which should, IMO, never have been called handlers, they are system services). Perhaps we need to discuss this approach in person.

awholegunch:
Will you be updating the documentation to reflect this? If not, please remind me or open issues to track the changes...we show "runnables" in the RESTful API docs, so they would need to be updated...See
the services scaling example at http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT/en/reference-manual/http-restful-api/lifecycle.html#scaling

anew:
Good point, @jj. I have updated the corresponding doc. Also added workers to that.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BasicHttpServiceContext.java
```diff
@@ -106,13 +106,13 @@ public TransactionContext getTransactionContext() {
...
   private static MetricsCollector getMetricCollector(MetricsCollectionService service,
-                                                     Program program, String runnableName,
+                                                     Program program, String handlerName,
                                                      String runId, int instanceId) {
     if (service == null) {
       return null;
     }
     Map<String, String> tags = Maps.newHashMap(getMetricsContext(program, runId));
-    tags.put(Constants.Metrics.Tag.SERVICE_RUNNABLE, runnableName);
+    tags.put(Constants.Metrics.Tag.HANDLER, handlerName);
```
abaranau: Looks good

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -111,7 +111,7 @@
       .put(Constants.Metrics.Tag.ADAPTER, "adapter")
 
       .put(Constants.Metrics.Tag.SERVICE, "service")
-      .put(Constants.Metrics.Tag.SERVICE_RUNNABLE, "runnable")
+      // SERVICE_HANDLER is the same HANDLER
```
abaranau: do we need this comment - I was confused by it

abaranau:
Metrics changes looks good

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
anew: It was specifically for you :) will remove it now.

anew:
thanks for the review, @abaranau ... but I am still wondering whether it is correct...

So in the past, Service Workers used a tag of SERVICE_RUNNABLE which is "srn". That now changes to HANDLER which is "hnd". 

Does that mean we will not be able to query existing metrics from before the change? And will the metrics query system deal with the now unknown tag "srn" if it encounters it in old metrics? Or is the only think we lose the ability to filter/slice by that old tag?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BasicHttpServiceContext.java
gsps1: question : do we care about the old data that's aggregated under SERVICE_RUNNABLE ?

abaranau:
discussed

anew:
@gsps1 right now we do not. That is an improvement we can make in the future

anew:
Metrics and Logging changes have been reviewed.
If I don't get more comments by noon, I will merge this.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/InMemoryServiceProgramRunner.java
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright © 2014-2015 Cask Data, Inc.
```
sreevatsanraman: 2015. Since this is a new file

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/InMemoryServiceProgramRunner.java
```diff
@@ -0,0 +1,133 @@
...
+                                         ProgramRunnerFactory.Type.SERVICE_COMPONENT);
+  }
+
+  /**
+   * Creates and starts all components of the given Service program.
+   *
+   * @return A {@link Table} with component name in the row, instance id in the column and the {@link ProgramController}
+   *         of the component runner as the cell value.
+   */
+  // TODO: Move this method to AbstractInMemoryProgramRunner to reduce duplication of code.
```
sreevatsanraman: This comment is not longer applicable because of the refactor. This TODO can be removed.

sreevatsanraman:
Two very minor nits. Please fix and merge. 

LGTM :+1:

PR #1778, opened 2015-03-23
adding zero filing for metrics

abaranau:
Needed for Ops Dashboard. Is not very clean (zero filling treated as special case in DefaultCube). Depending how critical to fix Ops Dashboard, we can get it in.

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
@@ -16,6 +16,7 @@
 
 package co.cask.cdap.metrics.query;
 
+import co.cask.cdap.api.metrics.Interpolator;
 import co.cask.cdap.api.metrics.MetricDataQuery;
 import co.cask.cdap.api.metrics.MetricSearchQuery;
 import co.cask.cdap.api.metrics.MetricStore;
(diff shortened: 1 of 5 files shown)


abaranau:
http://builds.cask.co/browse/CDAP-RBT213-1

abaranau:
against develop: http://builds.cask.co/browse/CDAP-DUT1232-1 GREEN

abaranau:
http://builds.cask.co/browse/CDAP-RBT213-1 GREEN

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -62,6 +63,9 @@
   public static final String ANY_TAG_VALUE = "*";
   public static final String TAG_DELIM = ".";
   public static final String DOT_ESCAPE_CHAR = "~";
+  // Had limit on returned metric values to avoid bad queries affect on cluster
```
gsps1: hard limit

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/cube/DefaultCube.java
```diff
@@ -299,9 +311,25 @@ private Aggregation findAggregation(CubeQuery query) {
...
         new TimeSeriesInterpolator(timeValues, query.getInterpolator(), query.getResolution()).iterator());
       List<TimeValue> resultTimeValues = Lists.newArrayList();
-      while (timeValueItor.hasNext()) {
-        TimeValue timeValue = timeValueItor.next();
-        resultTimeValues.add(new TimeValue(timeValue.getTimestamp(), timeValue.getValue()));
+
+      if (zeroInterpolator) {
+        long resultTimeStamp = (query.getStartTs() / query.getResolution()) * query.getResolution();
+        // return all data points if not more than limit
+        int limit = (int) Math.min(query.getLimit(), (query.getEndTs() - query.getStartTs()) / query.getResolution());
```
gsps1: should we do, ```(query.getEndTs() - query.getStartTs()) / query.getResolution() + 1``` as the end timestamp is inclusive?

gsps1:
couple of minor comments, would be good to have a test case.

albertshau:
should be against develop. closing

PR #1779, opened 2015-03-23
GrokRecordFormat: "grok" and "syslog" formats

nitinmotgi:
http://builds.cask.co/browse/CDAP-DUT1233

See [logstash docs](http://logstash.net/docs/1.4.2/filters/grok) for more info about what grok is.

For now, users can't specify additional grok patterns for later use.


cdap-data-fabric/src/main/java/co/cask/cdap/data/format/SyslogParser.java
```diff
@@ -0,0 +1,313 @@
...
+import java.nio.charset.Charset;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * SyslogParser.
+ */
+public class SyslogParser {
+  private static final Logger logger = LoggerFactory.getLogger(SyslogParser.class);
+  public static final String SYSLOG_FACILITY = "facility";
```
chtyim: Why all these are public?

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/SyslogParser.java
```diff
@@ -0,0 +1,313 @@
...
+ * the License.
+ */
+package co.cask.cdap.data.format;
+
+import com.google.common.base.Preconditions;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.collect.Maps;
+import org.apache.flume.source.SyslogUtils;
```
chtyim: Are you sure want to bring in flume as CDAP core dependency?

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/SyslogParser.java
```diff
@@ -0,0 +1,313 @@
...
+
+  /**
+   * Parses a Flume Event out of a syslog message string.
+   * @param msg Syslog message, not including the newline character
+   * @return Parsed Flume Event
+   * @throws IllegalArgumentException if unable to successfully parse message
+   */
+  public Map<String, String> parseMessage(String msg, Charset charset) {
+    Map<String, String> headers = Maps.newHashMap();
+
```
chtyim: Remove unnecessary empty lines.

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/SyslogParser.java
```diff
@@ -0,0 +1,313 @@
...
+  private static final int RFC3164_LEN = 15;
+  private static final int RFC5424_PREFIX_LEN = 19;
+  private final DateTimeFormatter timeParser;
+
+  public SyslogParser() {
+    timeParser = DateTimeFormat.forPattern(timePat).withZoneUTC();
+  }
+
+  /**
+   * Parses a Flume Event out of a syslog message string.
```
chtyim: Parsing flume event?? So it's not for parsing syslog file??

nitinmotgi:
lgtm

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/SyslogRecordFormat.java
```diff
@@ -0,0 +1,49 @@
...
+      Schema.Field.of("logsource", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+      Schema.Field.of("program", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+      Schema.Field.of("message", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+      Schema.Field.of("pid", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+  }
+
+  @Override
+  protected void addPatterns(Grok grok) {
+    addPattern(grok, "cdap/grok/patterns/grok-patterns");
+    addPattern(grok, "cdap/grok/patterns/linux-syslog");
```
chtyim: Why need this class?? Seems like the parent already supports the syslog pattern.

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
```diff
@@ -0,0 +1,179 @@
...
+    URL url = this.getClass().getClassLoader().getResource(resource);
+    if (url == null) {
+      LOG.error("Resource '{}' for grok pattern was not found", resource);
+      return;
+    }
+
+    try {
+      InputStream input = url.openStream();
+      try {
+        grok.addPatternFromReader(new InputStreamReader(input));
```
chtyim: Always specify charset when creating `Reader` from `InputStream`. Also, I suppose the pattern is relatively small, you can use `Resources.toString()` to read the content represented by the resource url and call `grok.addPattern`

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * GrokRecordFormat
+ */
+public class GrokRecordFormat extends StreamEventRecordFormat<StructuredRecord> {
+
+  public static final String PATTERN_SETTING = "pattern";
+  private static final Logger LOG = LoggerFactory.getLogger(GrokRecordFormat.class);
+  private static final String DEFAULT_PATTERN = "%{GREEDYDATA:body}";
+
+  private final Charset charset = Charsets.UTF_8;
```
chtyim: Why need a field for the constant `Charsets.UTF_8`?

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
```diff
@@ -0,0 +1,179 @@
...
+import java.nio.charset.Charset;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * GrokRecordFormat
+ */
+public class GrokRecordFormat extends StreamEventRecordFormat<StructuredRecord> {
+
+  public static final String PATTERN_SETTING = "pattern";
```
chtyim: Why public?

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
```diff
@@ -0,0 +1,179 @@
...
+
+    Match gm = grok.match(bodyAsStr);
+    gm.captures();
+    Map<String, Object> x = gm.toMap();
+
+    for (Schema.Field field : schema.getFields()) {
+      String fieldName = field.getName();
+      Object value = x.get(fieldName);
+      if (value != null) {
+        builder.convertAndSet(fieldName, value.toString());
```
chtyim: So conversion is always done in `String` form?

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
```diff
@@ -0,0 +1,179 @@
...
+  @Override
+  protected Schema getDefaultSchema() {
+    // default is a String[]
+    return Schema.recordOf("streamEvent", Schema.Field.of("body", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+  }
+
+  @Override
+  protected void validateSchema(Schema desiredSchema) throws UnsupportedTypeException {
+    // a valid schema is a record of simple types. In other words, no maps, arrays, records, unions, or enums allowed.
+    // the exception is the very last field, which is allowed to be an array of simple types.
```
chtyim: As far as I know, the Grok library is simply a regex parser. So where does this restriction and exception come from? Maybe worth document it in the javadoc of this class?

cdap-data-fabric/src/main/resources/cdap/grok/patterns/firewalls
```diff
@@ -0,0 +1,60 @@
+# NetScreen firewall logs
```
chtyim: Do we need license header for all these pattern files? If not, why? As far as I know, it is distributed as part of the software.

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/SyslogRecordFormat.java
jwang47: It does, but this provides a simpler and quicker way to set stream format to syslog format since user doesn't have to specify the pattern (as there is currently no way to list all the patterns, so the user would have to read the source code).

cdap-data-fabric/src/main/resources/cdap/grok/patterns/firewalls
jwang47: Wasn't sure how to handle the license here, as logstash stored these patterns under a directory that had the LICENSE in it, which is what I replicated under `cdap-data-fabric/src/main/resources/cdap/grok`. Having license headers here makes more sense though.

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
jwang47: Grok library gives you a map of string -> string after parsing, which results in this restriction.

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
jwang47: Yeah, apparently grok gives you a `Map<String, Object>` but the values are always `String`.

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/SyslogRecordFormat.java
chtyim: Seems like this duplication is unnecessary, especially this code is in the core data fabric module. Have you explore if there is any refactoring in the data fabric module to have it better supported for cases like this?

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
chtyim: But how about the exception? Also better have this documentation in the class javadoc

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/SyslogRecordFormat.java
jwang47: Going to remove it for now. Maybe introduce SyslogRecordFormat in a later PR.

cdap-data-fabric/src/main/java/co/cask/cdap/data/format/GrokRecordFormat.java
jwang47: Added this note to class javadoc. Don't know about the exception mentioned. @nitinmotgi, can you clarify?

chtyim: So, @nitinmotgi Nothing on this but merged?

PR #1786, opened 2015-03-23
(CDAP-1195) Basic implementation that uses Hive Predicate Pushdown to se...

chtyim:
...t start/end time when querying stream events.


cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
```diff
@@ -24,8 +24,21 @@
 import co.cask.cdap.data2.transaction.stream.StreamConfig;
 import co.cask.cdap.hive.context.ContextManager;
 import co.cask.cdap.proto.Id;
+import com.clearspring.analytics.util.Lists;
```
albertshau: change to guava

cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
```diff
@@ -94,12 +106,99 @@ public InputSplit createSplit(Path path, Path indexPath, long startTime, long en
...
+    }
+
+    // Stream can only be indexed by timestamp
+    analyzer.clearAllowedColumnNames();
+    analyzer.allowColumnName("ts");
+
+    ExprNodeGenericFuncDesc expr = Utilities.deserializeExpression(serializedExpr);
+
+    List<IndexSearchCondition> conditions = Lists.newArrayList();
+    analyzer.analyzePredicate(expr, conditions);
```
albertshau: javadoc says this only supports "pure conjunctions over binary expressions". What happens when you analyze something that isn't? like "ts < X or ts > Y"?

albertshau: can we add a test case to make sure it doesn't throw and exception and fail the query

albertshau:
one minor comment plus a question.  Have we tried this on multiple Hive versions? We sometimes run into problems on different versions.

cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
chtyim: I should add this class to abandon list in my IDE

cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
chtyim: It doesn't handle "or" case as far as I tested. We can enhance the support later in this class.

chtyim: For expression that the `IndexAnalyzer" not able to match anything, the list will be empty.

albertshau:
lgtm assuming it's been tested on a cluster

albertshau:
:+1:

chtyim:
Build passed and tested on cluster: http://builds.cask.co/browse/CDAP-DUT1241-4

PR #1787, opened 2015-03-23
Fix the new cdap-ui build to not require node installed, but always use ...

chtyim:
...the one downloaded during the build process.


sreevatsanraman:
LGTM :+1:

PR #1793, opened 2015-03-24
change modal message and add cdap bump gif

elfenheart:
change waiting message when backend service is not running

screen shot 2015-03-23 at 5 54 20 pm


astewes:
LGTM

PR #1796, opened 2015-03-24
[CDAP-1699] Improved exception messages and logging in InMemoryDatasetFr...

bdmogal:
...amework. Also cleaned up some dirt that is no longer required, since namespaces are not encoded in dataset names anymore

Build: http://builds.cask.co/browse/CDAP-DUT1245-1

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
@@ -128,7 +128,7 @@ public class InMemoryDatasetFramework implements DatasetFramework {
   public synchronized void addModule(Id.DatasetModule moduleId,
                                      DatasetModule module) throws ModuleConflictException {
     if (moduleClasses.contains(moduleId.getNamespace(), moduleId)) {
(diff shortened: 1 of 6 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryDatasetFramework.java
```diff
@@ -128,7 +128,7 @@ public InMemoryDatasetFramework(DatasetDefinitionRegistryFactory registryFactory
   public synchronized void addModule(Id.DatasetModule moduleId,
                                      DatasetModule module) throws ModuleConflictException {
     if (moduleClasses.contains(moduleId.getNamespace(), moduleId)) {
-      throw new ModuleConflictException("Cannot add module " + moduleId + ": it already exists.");
+      throw new ModuleConflictException(String.format("Cannot add module '%s' it already exists.", moduleId));
```
anwar6953: put some punctuation after the `'%s'`?

anwar6953: Remainder LGTM

bdmogal:
Build passed. Merging.

PR #1806, opened 2015-03-24
create stream

elfenheart:
Create Stream.

Setting property will be on another PR

### DIFF
--- a/cdap-ui/app/index.html
+++ b/cdap-ui/app/index.html
@@ -51,7 +51,7 @@
     <script type="text/javascript" src="/assets/bundle/tpl.js"></script>
 
     <script type="text/javascript">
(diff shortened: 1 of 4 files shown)


ajainarayanan:
:+1: LGTM

cdap-ui/app/index.html
```diff
@@ -51,7 +51,7 @@
     <script type="text/javascript" src="/assets/bundle/tpl.js"></script>
 
     <script type="text/javascript">
-      if(location.host==='localhost:8080') {
+      if(location.host==='localhost:9999') {
```
anwar6953: Shouldn't we get the '9999' from config["dashboard.bind.port"] ?

PR #1807, opened 2015-03-24
Fix metrics docs

awholegunch:
Fixes error in the v3 Metrics RESTful API

Staged at: [Metrics](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_fix_metrics_docs/en/reference-manual/http-restful-api/http-restful-api-v3/metrics.html)

In particular:
- [Search for Contexts](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_fix_metrics_docs/en/reference-manual/http-restful-api/http-restful-api-v3/metrics.html#search-for-contexts)
- [Search for Metrics](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_fix_metrics_docs/en/reference-manual/http-restful-api/http-restful-api-v3/metrics.html#search-for-metrics)
et al.

### DIFF
--- a/cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
+++ b/cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
@@ -73,12 +73,8 @@ enclosing context. These are the available Application contexts of CDAP:
      - ``namespace.<namespace>.app.<app-id>.flow.<flow-id>``
    * - All Flowlets of all app of an Application
      - ``namespace.<namespace>.app.<app-id>.flow.*``
-   * - One Procedure
-     - ``namespace.<namespace>.app.<app-id>.procedure.<procedure-id>``
    * - One Worker
      - ``namespace.<namespace>.app.<app-id>.worker.<worker-id>``
-   * - All Procedures of an Application
-     - ``namespace.<namespace>.app.<app-id>.procedure.*``
    * - All Workers of an Application
      - ``namespace.<namespace>.app.<app-id>.workers.*``
    * - All Mappers of a MapReduce
@@ -149,9 +145,6 @@ These metrics are available in a Flowlet context:
 
    * - Flowlet Metric
      - Description
-   * - ``system.process.busyness``
-     - A number from 0 to 100 indicating how “busy” the Flowlet is;
-       note that you cannot aggregate over this metric
    * - ``system.process.errors``
      - Number of errors while processing
    * - ``system.process.events.processed``
@@ -262,7 +255,7 @@ Searches and Queries
 The process of retrieving a metric involves these steps:
 
 1. Obtain (usually through a search) the correct context for a metric;
-#. Obtain (usually through a search of the context) the available metrics;
+#. Obtain (usually through a search within the context) the available metrics;
 #. Querying for a specific metric, supplying the context and any parameters.
 
 
@@ -395,9 +388,9 @@ To query a metric within a given context, perform an HTTP GET request::
 
    * - HTTP Method
      - ``POST '<base-url>/metrics/query?context=namespace.default.app.HelloWorld.flow.``
-       ``WhoFlow.flowlet.saver&metric=system.process.busyness?aggregate=true'``
+       ``WhoFlow.flowlet.saver&metric=system.process.events.processed?aggregate=true'``
    * - Description
-     - Using a *System* metric, *system.process.busyness*
+     - Using a *System* metric, *system.process.events.processed*
    * - 
      - 
    * - HTTP Method
--- a/cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
+++ b/cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
@@ -145,7 +145,7 @@ These metrics are available in a Flowlet context:
 
 .. list-table::
    :header-rows: 1
-   :widths: 40 60
+   :widths: 60 40
 
    * - Flowlet Metric
      - Description
@@ -173,7 +173,7 @@ These metrics are available in a Mappers and Reducers context:
 
 .. list-table::
    :header-rows: 1
-   :widths: 40 60
+   :widths: 60 40
 
    * - Mappers and Reducers Metric
      - Description
@@ -189,7 +189,7 @@ depends on the Spark program being queried:
 
 .. list-table::
    :header-rows: 1
-   :widths: 40 60
+   :widths: 60 40
 
    * - Spark Metric
      - Description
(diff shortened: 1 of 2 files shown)


cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
```diff
@@ -298,27 +298,28 @@ examples below for its use.
...
+     - ``["namespace.default.app.HelloWorld", "namespace.default.app.PurchaseHistory", "namespace.default.dataset.purchases", 
+       "namespace.default.dataset.whom", "namespace.default.stream.purchaseStream", ..., "namespace.default.stream.who"]``
    * - Description
      - Returns all child contexts of the given parent context; in this case, all entities in the default namespace.
    * - 
      - 
    * - HTTP Method
-     - ``POST <base-url>/metrics/search?target=childContext&context=namespace.default.app.PurchaseHistory.flow.PurchaseFlow.dataset.*.run.*``
+     - ``POST '<base-url>/metrics/search?target=childContext&context=``
+       ``namespace.default.app.PurchaseHistory.flow.PurchaseFlow.dataset.*.run.*'``
```
abaranau: no need for "dataset.*" - only confuses. It will work without it

cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
```diff
@@ -298,27 +298,28 @@ examples below for its use.
...
    * - Returns
-     - ``[ "namespace.default.app.PurchaseHistory.flow.PurchaseFlow.dataset.*.run.*.flowlet.collector", 
-       "namespace.default.app.PurchaseHistory.flow.PurchaseFlow.dataset.*.run.*.flowlet.reader" ]``
+     - ``["namespace.default.app.PurchaseHistory.flow.PurchaseFlow.dataset.*.run.*.flowlet.collector",
```
abaranau: then also remove here

cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
```diff
@@ -560,20 +590,28 @@ difference calculated between the start and end times:
      - From ``Thu, 28 Nov 2013 08:00:00 GMT`` to ``Thu, 28 Nov 2013 10:00:00 GMT``,
        with 1 hour resolution, will return 3 data points with metrics aggregated for each hour.
 
+Example::
+
+  POST '<base-url>/metrics/query?context=namespace.default.app.CountRandom&
+    metric=system.process.events.processed&start=now-1h&end=now&resolution=1m'
+
+This will return the value of the metric *system.process.events.processed* for the last
+hour at one-second intervals; potentially 61 entries may be returned, though only entries
```
abaranau: do we need this: "potentially 61 entries may be returned, though only entries
+where there is data will be returned."?

abaranau: confuses

abaranau:
Can we also add an example of the groupby query and its result? as in the API doc that we provided

cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
awholegunch: fixed.

cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
awholegunch: fixed.

cdap-docs/reference-manual/source/http-restful-api/http-restful-api-v3/metrics.rst
awholegunch: fixed.

awholegunch:
@abaranau Can you take a look again? Fixed the comments and added another example.

abaranau:
"For CDAP metrics (system metrics), the available metrics depend on the context. User-defined metrics are available in the context that they are emitted from."

abaranau:
"system.process.busyness"  - it is very unclear metric. Let's better remove its description

abaranau:
"system.<spark-id>.BlockManager.disk.diskSpaceUsed_MB" is it same spark-id as spark program name? I think it is not. @rohitsinha54 ?

abaranau:
we describe procedure metrics, but we should describe service metrics instead

abaranau:
"Obtain (usually through a search of the context) the available metrics;"

through a search WITHIN context?

abaranau:
"Retrieving multiple metrics at once, by issuing an HTTP POST request with a JSON list as the request body that enumerates the name and attributes for each metric, is currently not supported in this API. Instead, use the v2 API. It will be supported in a future release."

no need to say that. Also: do not recommend using v2 API - it'll be removed

abaranau:
Will be good to merge after fixing the above. Thank you!

awholegunch:
**You wrote:**
_Retrieving multiple metrics at once, by issuing an HTTP POST request with a JSON list as the request body that enumerates the name and attributes for each metric, is currently not supported in this API. Instead, use the v2 API. It will be supported in a future release."_

_no need to say that. Also: do not recommend using v2 API - it'll be removed_

**My reply:**
This is just for this release. Since anyone who used that feature can't do it in v3, we should tell them that and give them a workaround. When you restore that feature in 3.0 and we describe it here, we can remove the text then. We could change it to:

"Instead, use the v2 API until it is supported in a future release."

**You wrote:**
_"For CDAP metrics (system metrics), the available metrics depend on the context. User-defined metrics are available in the context that they are emitted from."_

Did you have a question about this? I have made other changes. I am following up with Rohit on Spark.

abaranau:
""Instead, use the v2 API until it is supported in a future release.""

+1

abaranau:
> Did you have a question about this?

Yes:

> not sure what did you want to say with this. Is it leftover from old docs? should we remove?

awholegunch:
**Sasha said:**
*"For CDAP metrics (system metrics), the available metrics depend on the context. User-defined metrics are available in the context that they are emitted from."*

*not sure what did you want to say with this. Is it leftover from old docs? should we remove?*

It was in the original. What is means is that the available system metrics change as you shift contexts. The same metric is sometimes available in more than one context. User metrics are only available in the single context that they are emitted. **Is that a better description?**

abaranau:
I don't think that description is needed, let's remove it

abaranau:
"Dataset metrics are available at the Dataset level, but they can also be queried down to the Flowlet, Worker, Procedure, Mapper, or Reducer level:"

Service

abaranau:
"we describe procedure metrics, but we should describe service metrics instead" - not addressed, right?

awholegunch:
I've made these changes per your comments: please look again. In particular, look at the change for the Service contexts, which I pulled from the MetricsConstants.java.

abaranau:
LGTM

PR #1808, opened 2015-03-25
Fix/ui non secure login screen

ajainarayanan:
- Fixes the cdap UI loading behavior to not show login form when security is disabled.
- Adds an app level loading icon directive + service to be used across features in cdap ui

### DIFF
--- a/cdap-ui/app/features/overview/controllers/overview-ctrl.js
+++ b/cdap-ui/app/features/overview/controllers/overview-ctrl.js
@@ -31,6 +31,17 @@ function ($scope, MyDataSource, $state, myLocalStorage, MY_CONFIG) {
   };
 
   $scope.isEnterprise = MY_CONFIG.isEnterprise;
+  // Testing if myLoadingService really works async.
+  // $timeout(function() {
+  //   myLoadingService
+  //     .showLoadingIcon()
+  //     .then(function() {
+  //       myLoadingService
+  //         .hideLoadingIcon();
+  //     });
+  //
+  // });
+
 
   dataSrc.request({
     _cdapNsPath: '/apps'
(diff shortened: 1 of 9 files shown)


ajainarayanan:
homepage

cdap-ui/app/features/overview/controllers/overview-ctrl.js
```diff
@@ -31,6 +31,17 @@ function ($scope, MyDataSource, $state, myLocalStorage, MY_CONFIG) {
...
+  // $timeout(function() {
+  //   myLoadingService
+  //     .showLoadingIcon()
+  //     .then(function() {
+  //       myLoadingService
+  //         .hideLoadingIcon();
+  //     });
+  //
+  // });
+
```
elfenheart: Do we still need this?

cdap-ui/app/services/app-level-loading-service.js
```diff
@@ -0,0 +1,25 @@
+angular.module(PKG.name + '.services')
+  .service('myLoadingService', function($q, $rootScope) {
+    var deferred;
+    this.showLoadingIcon = function() {
+      if (deferred) {
+        return deferred.promise;
+      } else {
+        deferred = $q.defer();
+        $rootScope.$broadcast('showLoadingIcon');
```
elfenheart: Should we create an EventPipe service? We don't want to clutter up $rootScope

cdap-ui/app/features/overview/controllers/overview-ctrl.js
ajainarayanan: Will remove it.

cdap-ui/app/services/app-level-loading-service.js
ajainarayanan: Since this is an app-level loading icon modal I am broadcasting the event on the rootscope. Since all scopes in angular inherit from rootscope this would work. Not sure what EventPipe here means. We need to communicate to all the scopes saying that the loading icon is shown/hidden in this case.

elfenheart:
:+1: LGTM

PR #1811, opened 2015-03-25
fix RPM and Debian pkg version identification logic for most cases

duddun:


cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -83,22 +83,35 @@ function sync_build_artifacts_to_server () {
...
+
+      *_all.deb )                             ## debian standard
+        _version_stub=`echo ${_package} | awk -F _ '{ print $2 }' | awk -F - '{ print $1 }'`
+        ;;
+
+      *noarch.rpm )                           ## rpm standard
+        _version_stub=`echo ${_package} | awk -F - '{ print $(NF-1) }'`
+        ;;
+
+      *_* )                                   ## default debian case
```
dereklwood: if you want this to catch only debians then add a .deb to the end.  otherwise this is a catchall for anything with a _, including rpms

dereklwood:
as a general comment, parsing versions is hard.  If possible, i'd recommend either getting the version from maven (there is a mvn one-liner that can do it), or looking at $COMPONENT/target/stage-packaging/opt/cdap/$component/VERSION

duddun:
Is this good to go?

cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -83,22 +83,35 @@ function sync_build_artifacts_to_server () {
...
     decho "PACKAGE=${i}"
     _package=`echo ${i} | awk -F / '{ print $(NF) }'`
-    if [[ "${_package}" == *rpm ]]
-    then
-      _version_stub=`echo ${_package} | awk -F - '{ print $(NF-1) }'`
-    else
-      if [[ "${_package}" == *_* ]]
-      then
+
+    case $_package in
```
wolf31o2: Braces...

cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -83,22 +83,35 @@ function sync_build_artifacts_to_server () {
...
+
+      cdap.*_[0-9]\.[0-9]\.[0-9]\.[0-9]+.*deb | cdap.*-[0-9]\.[0-9]\.[0-9]\.[0-9]+.*rpm )  ## snapshots with timestamp
```
wolf31o2: Why is there a product name in this line? Version parsing is hard. Let's simplify this logic a bit.

cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -83,22 +83,35 @@ function sync_build_artifacts_to_server () {
...
+      *noarch.rpm )                           ## rpm standard
+        _version_stub=`echo ${_package} | awk -F - '{ print $(NF-1) }'`
+        ;;
+
+      *_*deb )                                   ## default debian case
         _version_stub=`echo ${_package} | awk -F _ '{ print $2 }'`
-      else
+        ;;
+
+      *rpm | * )                              ## default RPM case and catchall
```
wolf31o2: Don't assume a versioning scheme in a catchall. Make the catchall fail, since if we hit this condition, something went horribly wrong, anyway.

cdap-distributions/bin/build_artifact_upload.sh
wolf31o2:
```bash
$ echo cdap-3.0.0.1427323964118-1.noarch.rpm | awk -F _ '{print $2}'

$
```
Is this the desired behavior?

cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -83,22 +83,35 @@ function sync_build_artifacts_to_server () {
   do
     decho "PACKAGE=${i}"
     _package=`echo ${i} | awk -F / '{ print $(NF) }'`
```
wolf31o2: Don't `echo | awk`... use `basename ${i}`

wolf31o2:
This will be a **lot** easier if you "ignore" (I use that lightly) what you think is the version, and rather remove what you know **isn't** the version. For example, **all** DEB packages have the format `packagename_version_arch.deb`, even though packagename could include underscores. So, work from the end to the beginning. Everything after the first underscore can be dropped when finding version and you know, by the way dpkg/apt parse versions, that everything form the next underscore to the new end of string **must** be the version, even if the package name includes underscores.

cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -82,23 +81,47 @@ function sync_build_artifacts_to_server () {
...
     decho "PACKAGE=${i}"
-    _package=`echo ${i} | awk -F / '{ print $(NF) }'`
-    if [[ "${_package}" == *rpm ]]
+    _package=`basename ${i}`
+
+    ## 
+    # most CDAP Debian packages will look like this: (typical, not just in CDAP)
+    #      cdap-blah-blah-blah_2.8.0-1_all.deb
+    #      cdap-blah-blah-blah_2.8.0.1427234000548-1_all.deb
+    #    Pattern: <package name>_<version>all.deb
```
wolf31o2: Fix this. `<package_name>_<version>-<build_revision>_all.deb`

cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -82,23 +81,47 @@ function sync_build_artifacts_to_server () {
...
+    # most CDAP RPM packages will look like this: 
+    #      cdap-blah-blah-blah-2.8.0-1.noarch.rpm
+    #      cdap-blah-blah-blah-2.8.0.1427234000548-1.noarch.rpm
+    #    Pattern: <package name>-<version>.noarch.rpm 
+    #
+    # Where version = <Major>.<Minor>.<Revision>(.optional time stamp)(-optional subrevision)
+    ##
+
+    # grab version stub
+    if [[ $_package == *_all.deb ]]
```
wolf31o2: Put braces around variables.

cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -82,23 +81,47 @@ function sync_build_artifacts_to_server () {
...
+    #   keep just the first 3 digits in case name includes timestamp
+    decho "version stub=${_version_stub}"
+    _version=`echo ${_version_stub} | awk -F - '{ print $1 }' | awk -F . '{ print $1"."$2"."$3 }'`
     decho "version = ${_version}"
 
+    # identify and create remote incoming directory
     OUTGOING_DIR=${BUILD_PACKAGE}/${_version}
-    mkdir -p ${OUTGOING_DIR}
+    echo "Creating remote directory"
+    ssh -l ${REMOTE_USER} ${REMOTE_HOST} "sudo mkdir -p ${REMOTE_INCOMING_DIR}/${OUTGOING_DIR}" || die "could not create remote directory"
```
wolf31o2: Let's assume the REMOTE_USER owns this directory structure. Don't use sudo. Otherwise, you need to chown/chmod after creation. Plus, this adds the requirement that the user has sudo on the remote machine.

wolf31o2:
:+1: LGTM

PR #1812, opened 2015-03-25
Avoid duplicate (and unnecessary) RPCs for reading each counter value, a...

anwar6953:
...s well as for reading the status.

Build: http://builds.cask.co/browse/CDAP-DUT1256-1 (GREEN)
JIRA: https://issues.cask.co/browse/CDAP-85

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceMetricsWriter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceMetricsWriter.java
@@ -23,6 +23,7 @@ import co.cask.cdap.metrics.collect.MapReduceCounterCollectionService;
 import org.apache.hadoop.mapreduce.Counter;
 import org.apache.hadoop.mapreduce.Counters;
 import org.apache.hadoop.mapreduce.Job;
+import org.apache.hadoop.mapreduce.JobStatus;
 import org.apache.hadoop.mapreduce.TaskCounter;
 import org.apache.hadoop.mapreduce.TaskReport;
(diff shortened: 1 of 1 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceMetricsWriter.java
```diff
@@ -59,14 +60,16 @@ public MapReduceMetricsWriter(Job jobConf, BasicMapReduceContext context) {
...
-    reportSystemStats();
+    Counters jobCounters = jobConf.getCounters();
+    reportMapredStats(jobCounters);
+    reportSystemStats(jobCounters);
   }
 
   // job level stats from counters built in to mapreduce
-  private void reportMapredStats() throws IOException, InterruptedException {
+  private void reportMapredStats(Counters jobCounters) throws IOException, InterruptedException {
+    JobStatus jobStatus = jobConf.getStatus();
```
abaranau: does it make sense to also fetch jobStatus once?

abaranau:
one Q, otherwise LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceMetricsWriter.java
anwar6953: Yes, that's what I also did in this PR.
I feel that each time we call `reportStats`, it should just report the stats as if it were a snapshot (as opposed to calling `job.getStatus` multiple times throughout the reporting iteration.

abaranau:
merge

PR #1815, opened 2015-03-25
update 2.6.0 documentation for new repo structure that affects 2.6.2

duddun:


cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -238,33 +238,41 @@ RPM using Yum
 .............
 Download the Cask Yum repo definition file::
 
-  sudo curl -o /etc/yum.repos.d/cask.repo http://repository.cask.co/downloads/centos/6/x86_64/cask.repo
+  sudo curl -o /etc/yum.repos.d/cask.repo http://repository.cask.co/centos/6/x86_64/cdap/2.6/cask.repo
```
awholegunch: we should use the ``|short-version|`` syntax of release 2.8 (so that we don't write "2.6" but instead put in a variable), but as long as this is not merged to develop, this is ok. If you want to add it, here it is:
```
.. container:: highlight

  .. parsed-literal::
    |$| sudo curl -o /etc/yum.repos.d/cask.repo |http:|//repository.cask.co/centos/6/x86_64/cdap/|short-version|/cask.repo
```

cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -281,7 +289,6 @@ Using Yum::
 
 Using APT::
 
-  sudo apt-get update
   sudo apt-get install cdap-gateway cdap-kafka cdap-master cdap-security cdap-web-app
```
awholegunch: OK: this one: the line ``sudo apt-get update`` is still in the 2.8 installation. Should it have been/be removed?

awholegunch:
Two comments, otherwise, LGTM.

awholegunch:
After discussion, LGTM.

PR #1817, opened 2015-03-26
Implemented suspend/resume for a particular run of the Workflow.

sagarkapare:
Added the rest endpoints "/workflows/<workflow-name>/runs/<run-id>/suspend" and "/workflows/<workflow-name>/runs/<run-id>/resume", to suspend/resume particular run of the Workflow. 

Planning to add "/workflows/<workflow-name>/suspend" and "/workflows/<workflow-name>/resume" in the next PR. This will suspend/resume all the Workflow runs along with the Schedules associated with the particular Workflow.


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -884,6 +885,101 @@ public void deleteFlowQueues(HttpRequest request, HttpResponder responder,
     }
   }
 
+  @POST
+  @Path("/apps/{app-id}/workflows/{workflow-name}/{run-id}/suspend")
```
bdmogal: Since workflows have some specific APIs (suspend, resume, and in future ``ExecutionContext``) should we extract them out to a ``WorkflowHttpHandler`` that extends ``ProgramLifecycleHttpHandler``? ``ProgramLifecycleHttpHandler`` is already becoming large, and these APIs are very specific to only 1 type of program - workflow.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -356,6 +361,20 @@ protected static ProgramRecord makeProgramRecord(String appId, ProgramSpecificat
     return null;
   }
 
+  protected Map<RunId, ProgramRuntimeService.RuntimeInfo> findAllRuntimeInfos(String namespaceId, String appId,
```
bdmogal: Better for this method to take ``Id.Program``?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -784,6 +789,173 @@ private void checkCurrentRuns(int retries, String url, int currentRunningProgram
     Assert.assertTrue(output.size() == currentRunningProgramsExpected);
   }
 
+  protected int getWorkflowSuspendResume(String namespaceId, String appId, String workflowId, String runId,
+                                          String action) throws Exception {
```
bdmogal: indentation

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -784,6 +789,173 @@ private void checkCurrentRuns(int retries, String url, int currentRunningProgram
     Assert.assertTrue(output.size() == currentRunningProgramsExpected);
   }
 
+  protected int getWorkflowSuspendResume(String namespaceId, String appId, String workflowId, String runId,
```
bdmogal: should we have ``suspendWorkflow(Id.Program, runId)`` and ``resumeWorkflow(Id.Program, runId)`` instead? @chtyim recently refactored this class similarly, so we should follow that pattern.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
bdmogal: actually, should we move this method to ``ProgramRuntimeService``? Instead of passing a ``ProgramRuntimeService`` to a helper method? That way even things that do not use ``AbstractAppFabricHttpHandler``, but need this functionality, can use it on a ``ProgramRuntimeService`` object?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -884,6 +885,101 @@ public void deleteFlowQueues(HttpRequest request, HttpResponder responder,
...
+    } catch (Throwable e) {
+      LOG.error("Got exception:", e);
+      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+    }
+  }
+
+  private void sendInvalidResponse(HttpResponder responder, Id.Program id) {
+    try {
+      AppFabricServiceStatus status;
+      try {
```
bdmogal: This inner try-catch is not needed since you already have a catch block on the outer try that catches a ``Throwable`` and sends an ``INTERNAL_SERVER_ERROR``?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -884,6 +885,101 @@ public void deleteFlowQueues(HttpRequest request, HttpResponder responder,
...
+      responder.sendString(status.getCode(), status.getMessage());
+    } catch (SecurityException e) {
+      responder.sendStatus(HttpResponseStatus.UNAUTHORIZED);
+    } catch (Throwable e) {
+      LOG.error("Got exception:", e);
+      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+    }
+  }
+
+  private AppFabricServiceStatus suspendResumeProgram(ProgramController controller, String action) {
```
bdmogal: Is it better to split this into ``suspendProgram`` and ``resumeProgram``? I think it may be much cleaner, also callers will not have to pass in exact strings for the particular action. Also seems like we don't have much common logic between suspend and resume.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -81,6 +83,9 @@
   private NettyHttpService httpService;
   private volatile Thread runningThread;
   private final Map<String, WorkflowActionNode> status = new ConcurrentHashMap<String, WorkflowActionNode>();
+  private volatile boolean isSuspended;
+  private ReentrantLock lock;
```
bdmogal: why does this have to be reentrant?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -114,6 +121,40 @@ protected void startUp() throws Exception {
     runningThread = Thread.currentThread();
   }
 
+  private void checkSuspended() {
+    lock.lock();
+    try {
+      while (isSuspended) {
+        condition.await();
+      }
+    } catch (InterruptedException e) {
```
bdmogal: Should we log a message here?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -884,6 +885,101 @@ public void deleteFlowQueues(HttpRequest request, HttpResponder responder,
...
+      responder.sendString(status.getCode(), status.getMessage());
+    }  catch (SecurityException e) {
+      responder.sendStatus(HttpResponseStatus.UNAUTHORIZED);
+    } catch (Throwable e) {
+      LOG.error("Got exception:", e);
+      responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+    }
+  }
+
+  private void sendInvalidResponse(HttpResponder responder, Id.Program id) {
```
bdmogal: The current calls to this method can only send ``NOT_FOUND``, right? its only called when ``runtimeInfo`` is null. Will we use this method for something else later?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Moved to `WorkflowHttpHandler`

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
sagarkapare: Added new method in the `ProgramRuntimeService` to `list` the runtime info objects for given `Id.Program`

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
sagarkapare: Fixed.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
sagarkapare: Using `suspendWorkflow` and `resumeWorkflow` methods now.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Remove the additional methods since the code was minimal.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: We need the extended capabilities such as wait and notify on the condition for suspend and resume functionality. Since implicit lock does not provide those features, we need `ReentrantLock`

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: This is only get called when the runtime info not found. Moved the method to the `WorkflowHttpHandler`

sagarkapare:
@bdmogal Addressed comments. Please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -146,10 +140,10 @@
   /**
    * Runtime program service for running and managing programs.
    */
-  private final ProgramRuntimeService runtimeService;
+  protected final ProgramRuntimeService runtimeService;
```
bdmogal: should we group the protected and private members together? Makes it more readable.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -0,0 +1,244 @@
...
+  public WorkflowHttpHandler(Authenticator authenticator, StoreFactory storeFactory, WorkflowClient workflowClient,
+                             CConfiguration configuration, ProgramRuntimeService runtimeService,
+                             DiscoveryServiceClient discoveryServiceClient, QueueAdmin queueAdmin, Scheduler scheduler,
+                             PreferencesStore preferencesStore, NamespacedLocationFactory namespacedLocationFactory) {
+    super(authenticator, storeFactory, workflowClient, configuration, runtimeService, discoveryServiceClient,
+          queueAdmin, scheduler, preferencesStore, namespacedLocationFactory);
+  }
+
+  @POST
+  @Path("/apps/{app-id}/workflows/{workflow-name}/{run-id}/suspend")
```
bdmogal: should send an API review for these APIs. Not RESTy, because of the missing "runs" component after ``{workflow-name}``.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
bdmogal: Did you miss this?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -178,6 +219,7 @@ private void executeFork(final ApplicationSpecification appSpec, WorkflowForkNod
           public String call() throws Exception {
             Iterator<WorkflowNode> iterator = branch.iterator();
             while (!Thread.currentThread().isInterrupted() && iterator.hasNext()) {
+              checkSuspended();
```
bdmogal: should we add this check in the ``executeNode`` method, instead of calling it before every call to ``executeNode()``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -114,6 +121,40 @@ protected void startUp() throws Exception {
     runningThread = Thread.currentThread();
   }
 
+  private void checkSuspended() {
```
bdmogal: so this is really a ``waitWhileSuspended()``, right? and not just ``checkSuspended``?

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -75,6 +76,17 @@ public synchronized RuntimeInfo lookup(RunId runId) {
   }
 
   @Override
+  public synchronized Map<RunId, RuntimeInfo> list(Id.Program program) {
+    Map<RunId, RuntimeInfo> runtimeInfoMap = Maps.newHashMap();
```
chtyim: Use `Maps.filterValues` instead.

```java
return Maps.filterValues(list(program.getType()), new Predicate<RuntimeInfo>() {
      @Override
      public boolean apply(RuntimeInfo info) {
        return info.getProgramId().equals(program);
      }
});

File: cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java [hunk 15]
```diff
@@ -122,16 +116,16 @@
...
     .create();
 
   /**
    * Store manages non-runtime lifecycle.
    */
-  private final Store store;
+  protected final Store store;
 
-  private final WorkflowClient workflowClient;
+  protected final WorkflowClient workflowClient;
```
chtyim: Why protected? If this field is only use by the sub-class (WorkflowHttpHandler), then do the guice injection there.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
chtyim: `Lock` class does support `newCondition()`, isn't it?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -81,6 +83,9 @@
   private NettyHttpService httpService;
   private volatile Thread runningThread;
   private final Map<String, WorkflowActionNode> status = new ConcurrentHashMap<String, WorkflowActionNode>();
+  private volatile boolean isSuspended;
```
chtyim: By common Java convention, variable name don't have `is` (otherwise the getter method for the variable will become `isIsSuspended`. Just call it `suspended`.

chtyim: Also, since you always use a lock when accessing the `suspended` variable, it doesn't need to have volatile.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
chtyim: Or `blockIfSuspended`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -237,6 +279,7 @@ protected void run() throws Exception {
     final Iterator<WorkflowNode> iterator = workflowSpec.getNodes().iterator();
     while (iterator.hasNext() && runningThread != null) {
       try {
+        checkSuspended();
```
chtyim: Looks like code duplication to me. You better have a method `executeAll(Iterator<WorkflowNode> iterator)` to encapsulate the logic.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -114,6 +121,40 @@ protected void startUp() throws Exception {
...
+        condition.await();
+      }
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  public void suspend() throws Exception {
```
chtyim: javadoc for public method. Also, seems to me the suspend contract is to pause only after all current steps are completed, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
```diff
@@ -133,6 +134,70 @@ public void recordProgramStart(Id.Program program, String pid, long startTs) {
...
+      .add(program.getNamespaceId())
+      .add(program.getApplicationId())
+      .add(program.getType().name())
+      .add(program.getId())
+      .add(pid)
+      .build();
+    write(key, new RunRecord(started.getPid(), started.getStartTs(), null, ProgramRunStatus.SUSPENDED));
+  }
+
+  public void recordProgramResumed(Id.Program program, String pid) {
```
chtyim: This looks highly duplicated with the previous method.

cdap-app-fabric/src/test/java/co/cask/cdap/PauseResumeWorklowApp.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  static final class FirstSimpleAction extends AbstractWorkflowAction {
+    private static final Logger LOG = LoggerFactory.getLogger(FirstSimpleAction.class);
+    @Override
+    public void run() {
+      LOG.info("Running FirstSimpleAction");
+      try {
+        File file = new File(getContext().getRuntimeArguments().get("first.simple.action.file"));
```
chtyim: You can simplifies all these Action classes into one that has the argument name determined by the action name:

```java
addAction(new SimpleAction("first"));

class SimpleAction extends AbstractWorkflowAction {
  SimpleAction(String name) {
    super(name);
  }
  void run() {
    File file = new File(getContext().getRuntimeArguments().get(getContext().getSpecification().getName() + ".simple.action.file"));
    ....
  }
}
```

The `getName()` in this case returns the name you passed in through the action constructor (e.g. `first`, hence the runtime argument is "first.simple.action.file").

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -0,0 +1,755 @@
...
+ * Tests for {@link WorkflowHttpHandler}
+ */
+public class WorkflowHttpHandlerTest  extends AppFabricTestBase {
+
+  private static final Gson GSON = new GsonBuilder()
+    .registerTypeAdapter(ScheduleSpecification.class, new ScheduleSpecificationCodec())
+    .registerTypeAdapter(WorkflowActionSpecification.class, new WorkflowActionSpecificationCodec())
+    .create();
+
+  private static final String PAUSE_RESUME_WORKFLOW_APP = "PauseResumeWorkflowApp";
```
chtyim: Can you guys stop defining all these constants? You can have it as local variable in the test method. Having ALL_CAPS_EVERYWHERE_IN_THE_CODE hurts the reading.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -0,0 +1,755 @@
...
+    while (trial++ < retries) {
+      response = doGet(url);
+      if (response.getStatusLine().getStatusCode() == 200) {
+        json = EntityUtils.toString(response.getEntity());
+        output = GSON.fromJson(json, LIST_WORKFLOWACTIONNODE_TYPE);
+        if (output.size() == currentRunningProgramsExpected) {
+          break;
+        }
+      }
+      TimeUnit.SECONDS.sleep(1);
```
chtyim: Don't sleep 1 second when polling in unit-test. Sleep only 50-100ms. Also, use `Tasks.waitFor()` instead, don't do custom polling loop everywhere.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -0,0 +1,755 @@
...
+    String currentUrl = String.format("apps/%s/workflows/%s/%s/current", app, workflow, runId);
+    String versionedUrl = getVersionedAPIPath(currentUrl, Constants.Gateway.API_VERSION_3_TOKEN, namespace);
+    return doGet(versionedUrl);
+  }
+
+  private String createInput(String folderName) throws IOException {
+    File inputDir = tmpFolder.newFolder(folderName);
+
+    File inputFile = new File(inputDir.getPath() + "/words.txt");
+    inputFile.deleteOnExit();
```
chtyim: no need

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -0,0 +1,755 @@
...
+
+  private Long getNextScheduledRunTime(String namespace, String app, String workflow, String schedule)
+    throws Exception {
+    String nextRunTimeUrl = String.format("apps/%s/workflows/%s/nextruntime", app, workflow);
+    String versionedUrl = getVersionedAPIPath(nextRunTimeUrl, Constants.Gateway.API_VERSION_3_TOKEN, namespace);
+    HttpResponse response = doGet(versionedUrl);
+    JsonArray array = readResponse(response, JsonArray.class);
+    JsonObject wfObject = (JsonObject) array.get(0);
+    Assert.assertNotNull(wfObject);
+    String id = wfObject.get("id").getAsString();
```
chtyim: We don't have a class in cdap-proto for this?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -0,0 +1,755 @@
...
+    Long time = wfObject.get("time").getAsLong();
+    Assert.assertTrue(id.contains(schedule));
+    return time;
+  }
+
+  private String getStatusUrl(String namespace, String appName, String schedule) throws Exception {
```
chtyim: `URL` should be all caps (by Java naming convention, common abbreviation should be all caps when use in method name).

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -0,0 +1,755 @@
...
+    String statusUrl = String.format("apps/%s/schedules/%s/status", appName, schedule);
+    return getVersionedAPIPath(statusUrl, Constants.Gateway.API_VERSION_3_TOKEN, namespace);
+  }
+
+  @Category(XSlowTests.class)
```
chtyim: Is it XSlow?? how slow it is?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -0,0 +1,755 @@
...
+    Assert.assertEquals(200, resumeSchedule(defaultNamespace, APP_WITH_CONCURRENT_WORKFLOW,
+                                            APP_WITH_CONCURRENT_WORKFLOW_SCHEDULE_2));
+
+    Map<String, String> propMap = Maps.newHashMap();
+    propMap.put(ProgramOptionConstants.CONCURRENT_RUNS_ENABLED, "true");
+    PreferencesStore store = getInjector().getInstance(PreferencesStore.class);
+    store.setProperties(defaultNamespace, APP_WITH_CONCURRENT_WORKFLOW, ProgramType.WORKFLOW.getCategoryName(),
+                        CONCURRENT_WORKFLOW_NAME, propMap);
+
+    String runsUrl = getRunsUrl(defaultNamespace, APP_WITH_CONCURRENT_WORKFLOW, CONCURRENT_WORKFLOW_NAME, "running");
```
chtyim: Don't generate URL and call `scheduleHistoryRuns`. Have a `scheduleHistoryRun(Id.Program programId)` to do the job (URL generation should be in that method). I've been fixing that in the AppFabricTestBase (go to that class and see how it should be. E.g. Look at `startProgram` or `waitState` methods there). We should stop doing things like this in new code.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: Added `Lock lock = new ReentrantLock();`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: Added log message.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Grouped the methods together based on the access modifiers.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
sagarkapare: fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Moved `WorkflowClient` object to the `WorkflowHttpHandler`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: Moved the common logic to the `executeAll` method.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: Yes. suspend happens only after all the current steps are executed. Added javadoc.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
sagarkapare: Moved to the common method.

cdap-app-fabric/src/test/java/co/cask/cdap/PauseResumeWorklowApp.java
sagarkapare: Simplified the actions as per suggestions.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: Moved the variables to the local scope.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: Using `Tasks.waitFor()` now instead of custom loops.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: removed.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: No. Currently the handler for the `nextruntime` returns the `JsonArray` as

```java
JsonArray array = new JsonArray();
for (ScheduledRuntime runtime : runtimes) {
        JsonObject object = new JsonObject();
        object.addProperty("id", runtime.getScheduleId());
        object.addProperty("time", runtime.getTime());
        array.add(object);
      }
```

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: Fixed.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: Updated the XSlow for the test cases which run for more than 30 seconds only.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: These test cases were moved here from `ProgramLifeCycleHttpHandlerTest`. Fixed the methods to match the pattern with the `AppFabricTestBase`.

sagarkapare:

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT1262-8

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
chtyim: We should have one then.

sagarkapare: Using `ScheduledRuntime` class for the same and moved it to the cdap-proto.

chtyim:
LGTM.

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT1262-12

bdmogal:
LGTM

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT1262-13. Merging.

PR #1820, opened 2015-03-26
Feature/ui set stream properties

elfenheart:
will move create stream and set properties to the admin view under namespace on separate PR


ajainarayanan:
:+1: LGTM

PR #1821, opened 2015-03-26
Adding Table.scan(Scan)

abaranau:
This PR is for discussing mergint MetricsTable and Table classes hierarchies. At minimum, it lacks unit-tests for the added scan method.

UPDATE: was finished up

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Scan.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Scan.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.dataset.table;
+
+import co.cask.cdap.api.annotation.Beta;
+import co.cask.cdap.api.common.Bytes;
+import com.google.common.base.Objects;
+
+import javax.annotation.Nullable;
+
+/**
+ * Scan configuration for {@link Table}.
+ */
+@Beta
+public class Scan {
+  @Nullable
+  private final byte[] startRow;
+  @Nullable
+  private final byte[] stopRow;
+  @Nullable
+  private final Filter filter;
+
+  /**
+   * Creates {@link Scan} for a given start and stop row keys.
+   * @param startRow start row inclusive; {@code null} means start from first row of the table
+   * @param stopRow stop row exclusive; {@code null} means scan all rows to the end of the table
+   */
+  public Scan(@Nullable byte[] startRow, @Nullable byte[] stopRow) {
+    this(startRow, stopRow, null);
+  }
+
+  /**
+   * Creates {@link Scan} for a given start and stop row keys and filter.
+   * @param startRow start row inclusive; {@code null} means start from first row of the table
+   * @param stopRow stop row exclusive; {@code null} means scan all rows to the end of the table
+   * @param filter filter to be used on scan
+   */
+  public Scan(@Nullable byte[] startRow, @Nullable byte[] stopRow, @Nullable Filter filter) {
+    this.startRow = startRow;
+    this.stopRow = stopRow;
+    this.filter = filter;
+  }
+
+  @Nullable
+  public byte[] getStartRow() {
+    return startRow;
+  }
+
+  @Nullable
+  public byte[] getStopRow() {
+    return stopRow;
+  }
+
+  @Nullable
+  public Filter getFilter() {
+    return filter;
+  }
+
+  @Override
+  public String toString() {
+    return Objects.toStringHelper(this)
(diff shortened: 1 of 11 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Scan.java
```diff
@@ -0,0 +1,82 @@
...
+    return startRow;
+  }
+
+  @Nullable
+  public byte[] getStopRow() {
+    return stopRow;
+  }
+
+  @Nullable
+  public <T> T getFilter() {
```
chtyim: Useful to have another method in this class to get the `Class` type of the filter to avoid the need of catching `ClassCastException`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
```diff
@@ -565,13 +583,28 @@ public Scanner scan(byte[] startRow, byte[] stopRow) {
...
+
+  private NavigableMap<byte[], NavigableMap<byte[], Update>> applyFilter(
+                                                        NavigableMap<byte[], NavigableMap<byte[], Update>> bufferMap,
+                                                        @Nullable Object filter) {
+    if (filter == null) {
+      return bufferMap;
+    }
+
+    // todo: currently we support only FuzzyRowFilter as an experimental feature
+    if (filter instanceof FuzzyRowFilter) {
```
chtyim: Hum.. this is the problematic part.. this means we need to re-implement different filters to be applied at the buffer table level? Shall we consider flushing the table when scan is going to be performed with a filter so that we can rely on the server side filter?

abaranau: discussed. It is ok for now to start with.

cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Scan.java
abaranau: Per discussion with Gary, we'll add a Filter "marker" interface for now

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
nitinmotgi: @abaranau can you please add the conclusion of discussion please.

awholegunch:
Is there going to be any documentation for this, perhaps marked as Beta?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
abaranau: The summary is that it is fine to start with support of one filter in this custom/special way. We'll learn from it and it will help to figure out the API for server side filters feature.

abaranau:
@awholegunch not yet. This filter is not exposed yet.

abaranau:
For completion, this PR lacks unit-tests

abaranau:
added test. Let's review again

abaranau:
http://builds.cask.co/browse/CDAP-DUT1370-1

cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Scan.java
```diff
@@ -0,0 +1,81 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Scan configuration for {@link Table}.
+ */
+@Beta
+public class Scan {
+  @Nullable
+  private final byte[] startRow;
```
ghelmling: Why final for any of these?  Why not make them settable?

ghelmling: Nevermind, it's good that these can't be modified if we pass this along.  But if we add more options we may want to adopt a builder pattern here.

abaranau: agree. Now it is almost empty :(

cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Scan.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return Objects.toStringHelper(this)
+      .add("startRow", startRow == null ? null : Bytes.toStringBinary(startRow))
+      .add("stopRow", stopRow == null ? null : Bytes.toStringBinary(stopRow))
+      .add("filter", filter)
+      .toString();
+  }
```
anwar6953: Don't need the ternary. It's the same effect without it.

```java
 @Override
 public String toString() {
     .add("startRow", Bytes.toStringBinary(startRow))
     .add("stopRow", Bytes.toStringBinary(stopRow))
     .add("filter", filter)
     .toString();
  }
```

anwar6953:
looks good to me

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/table/TableTest.java
```diff
@@ -1101,6 +1105,77 @@ public void testScanAndDelete() throws Exception {
...
+            }
+          }
+        }
+      }
+
+      // we should have 81 (3^4) rows now
+      Assert.assertEquals(81, countRows(table));
+
+      // check that filter works against data written in same tx
+      verifyScanWithFuzzlyRowFilter(table);
```
ghelmling: "Fuzzly"?

abaranau: :) fixed

ghelmling:
lgtm.  one very minor typo.

PR #1822, opened 2015-03-26
Fix/ui non secure login screen

ajainarayanan:
- Fixes usage of $http to sockets for backend status
- Fixes the event to fireup even when node server is down.

### DIFF
--- a/cdap-ui/app/main.js
+++ b/cdap-ui/app/main.js
@@ -138,12 +138,35 @@ angular
     });
   })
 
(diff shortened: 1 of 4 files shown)


cdap-ui/app/main.js
```diff
@@ -156,6 +179,7 @@ angular
 
 
     $scope.$on('$stateChangeSuccess', function (event, state) {
+      console.log("still in body controller");
```
elfenheart: Can we remove this?

elfenheart:
:+1: LGTM

PR #1823, opened 2015-03-26
CDAP-1895 fixing a bug where previewing a query would close the

albertshau:
query if there were no results returned.

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
@@ -19,7 +19,6 @@ package co.cask.cdap.explore.service.hive;
 import co.cask.cdap.api.dataset.Dataset;
 import co.cask.cdap.api.dataset.DatasetSpecification;
 import co.cask.cdap.api.dataset.lib.Partition;
-import co.cask.cdap.api.dataset.lib.PartitionKey;
 import co.cask.cdap.api.dataset.lib.TimePartitionedFileSet;
 import co.cask.cdap.app.runtime.scheduler.SchedulerQueueResolver;
 import co.cask.cdap.app.store.Store;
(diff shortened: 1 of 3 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT1265

jguery:
Where do you prevent the query from not entering the inactive state? I don't see those changes.

albertshau:
calls fetchNextResults instead of nextResults.  nextResults is the one that calls timeoutAggressively, which makes it inactive.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -766,13 +764,8 @@ public QueryStatus getStatus(QueryHandle handle) throws ExploreException, Handle
...
-    Lock nextLock = getOperationInfo(handle).getNextLock();
-    nextLock.lock();
     try {
-      // Fetch results from Hive
-      LOG.trace("Getting results for handle {}", handle);
-      OperationHandle opHandle = getOperationHandle(handle);
-      List<QueryResult> results = fetchNextResults(opHandle, size);
+      List<QueryResult> results = fetchNextResults(handle, size);
       QueryStatus status = getStatus(handle);
       if (results.isEmpty() && status.getStatus() == QueryStatus.OpStatus.FINISHED) {
```
albertshau: this is where the query handle is moved to inactive.

albertshau:
tested on cdh4, cdh5, hdp2

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
jguery: Oh ok yes I see now. Thanks.

jguery:
LGTM

PR #1824, opened 2015-03-26
Feature/ui explorer dash

nistala:
- Store user queries in settings
- Format tables UI


cdap-ui/app/features/datasets/controllers/tabs/query-model.js
```diff
@@ -0,0 +1,79 @@
...
+          self.sync();
+        });
+        return deferred.promise;
+      };
+
+      /**
+       * Merges queries stored in user settings and queries pulled from server
+       * using query_handle.
+       */
+      QueryModel.prototype.mergeQueries = function (userQueries, apiQueries) {
```
ajainarayanan: We get the query handles from two endpoints, 

- From backend: We get all the queries made by all the users 
- From user store: We get a list of query handles for the current user .

Some of the query handles that we had stored in the backend might have expired. So at time t=0th day backend might have sent 30 queries and the current user might have made 5 queries out of that.
So now we should show only those 5 queries in the UI.

And lets say the query expiration time is 1 day and after a day (time t=1day) we query for list of queries we might not get all the 30 because from the backend since some might have expired.  
When we say the queries expired it might be queries made by the current user + queries made by other users.

So now we need to check if the list of query handles that we get from user store, against the list of query handles we get from backend, are valid and are not expired and show only those that belong to the current user. 

Hope I am making some sense here.

nistala: if they are expired, should we delete them, then it would not make sense to store them in user settings. The table also allows old queries to be clicked and previewed. I've attached a sample response from the server regarding queries, should I check is active and whether it is contained in this list?
[
{
timestamp: 1427414764599,
statement: "SELECT * FROM dataset_wordcounts LIMIT 5",
status: "FINISHED",
query_handle: "71339d45-da97-4557-a3ef-95c94a4f2519",
has_results: true,
is_active: true
}
]

ajainarayanan:
@nistala :+1: LGTM Can you merge with latest develop so that we can merge it in github..

PR #1825, opened 2015-03-26
[CDAP-1293] Fix workflow program status and history

bdmogal:
This is still in progress.

Pending thorough testing (both on cluster as well as unit testing). Opening PR to get feedback on the approach.

This also does some refactoring in ``ProgramLifecycleHttpHandlerTest`` to not use deprecated methods.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/InMemoryProgramRunnerModule.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/InMemoryProgramRunnerModule.java
@@ -18,6 +18,8 @@ package co.cask.cdap.app.guice;
 
 import co.cask.cdap.app.runtime.ProgramRunner;
 import co.cask.cdap.app.runtime.ProgramRuntimeService;
+import co.cask.cdap.app.store.Store;
+import co.cask.cdap.app.store.StoreFactory;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.discovery.ResolvingDiscoverable;
@@ -38,6 +40,7 @@ import co.cask.cdap.internal.app.runtime.webapp.JarHttpHandler;
 import co.cask.cdap.internal.app.runtime.webapp.WebappHttpHandlerFactory;
 import co.cask.cdap.internal.app.runtime.webapp.WebappProgramRunner;
 import co.cask.cdap.internal.app.runtime.workflow.WorkflowProgramRunner;
+import co.cask.cdap.internal.app.store.DefaultStore;
 import co.cask.cdap.internal.app.worker.InMemoryWorkerRunner;
 import co.cask.cdap.internal.app.worker.WorkerProgramRunner;
 import com.google.common.base.Preconditions;
@@ -111,6 +114,8 @@ final class InMemoryProgramRunnerModule extends PrivateModule {
     // Create webapp http handler factory.
     install(new FactoryModuleBuilder().implement(JarHttpHandler.class, IntactJarHttpHandler.class)
               .build(WebappHttpHandlerFactory.class));
+
+    bind(Store.class).to(DefaultStore.class).in(Scopes.SINGLETON);
   }
 
   @Singleton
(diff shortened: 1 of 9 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/InMemoryProgramRunnerModule.java
```diff
@@ -111,6 +114,8 @@ protected void configure() {
     // Create webapp http handler factory.
     install(new FactoryModuleBuilder().implement(JarHttpHandler.class, IntactJarHttpHandler.class)
               .build(WebappHttpHandlerFactory.class));
+
+    bind(Store.class).to(DefaultStore.class).in(Scopes.SINGLETON);
```
bdmogal: This needs to be fixed to bind ``StoreFactory``, but its causing problems right now, will investigate and fix.

sagarkapare:
Approach looks good to me! Please let me know when you are done with the implementation changes and testing and I can take a look at it again. One more thing to keep in mind while making these changes is how the status of the underlying program will get updated. Maybe we can tackle that as part of another PR.

cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/InMemoryProgramRunnerModule.java
chtyim: We should get rid of the `StoreFactory` class. It is absolutely useless.

chtyim: And why program runner need `Store`? The design of the whole `ProgramRunner` and related class is that they should never communicate to Store.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
```diff
@@ -107,28 +117,53 @@ public RuntimeContext call() throws Exception {
...
+  protected RuntimeContext executeProgram(final Id.Program programId, final ProgramController controller,
                                           final RuntimeContext context) throws Exception {
+    // TODO: Perhaps generate a different runId for each program contained in a workflow
+    final String runId = controller.getRunId().getId();
     // Execute the program.
     final SettableFuture<RuntimeContext> completion = SettableFuture.create();
     controller.addListener(new AbstractListener() {
       @Override
+      public void init(ProgramController.State state, @Nullable Throwable cause) {
+        store.setStart(programId, runId, TimeUnit.SECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS));
```
chtyim: Why?? The existence of Controller and listener is exactly to avoid logic like this inside ProgramRunner. The ProgramRunner should only cares about the state transition, but shouldn't responsible for changing things in Store.

chtyim:
@bdmogal Please explain what you want to fix and what's the approach is.

cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/InMemoryProgramRunnerModule.java
bdmogal: I'll create a PR to remove ``StoreFactory``

bdmogal:
@chtyim The problem is that when you start/stop a workflow, the programs that it runs in its nodes never transition states - they always remain ``STOPPED``. Also, the history of those programs never gets recorded either. Is there a better approach to do this? I guess it would have to be somewhere in the workflow classes, since that's the only place which knows when the programs start/stop. 

Here is the jira - [CDAP-1293](https://issues.cask.co/browse/CDAP-1293)

bdmogal:
Also, the ``Store`` is used in this PR in ``AbstractProgramWorkflowRunner``, which is not a ``ProgramRunner``, it implements ``ProgramWorkflowRunner``.

chtyim:
I'll look into details of CDAP-1293. BTW, since you mention the `ProgramWorkflowRunner`, that class hierarchy is not that good (from last time I looked at it). Anyway, it's part of the runtime execution system which I referred to above.

bdmogal:
This is being done in #1887

PR #1826, opened 2015-03-26
CDAP-1941 fix explore on older Hive versions.

albertshau:


albertshau:
http://builds.cask.co/browse/CDAP-DUT1268

cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
```diff
@@ -134,7 +137,21 @@ public InputSplit createSplit(Path path, Path indexPath, long startTime, long en
...
+    // whereas newer versions use Utilities.deserializeExpression(String).
+    try {
+      expr = Utilities.deserializeExpression(serializedExpr);
+    } catch (NoSuchMethodError e) {
+      try {
+        expr = (ExprNodeGenericFuncDesc) Utilities.class.getMethod(
+          "deserializeExpression", String.class, Configuration.class).invoke(null, serializedExpr, conf);
+      } catch (Exception e1) {
+        LOG.warn("Could not deserialize query predicate. A full table scan will be performed.", e1);
+        return builder.setStartTime(startTime).setEndTime(endTime);
```
chtyim: Maybe better to have the whole expression analyze logic runs inside the `try-catch` block to have one `return` at the end.

albertshau: hm that's true, never know what else may have similar issues. will do

chtyim:
LGTM.

PR #1827, opened 2015-03-27
Feature/ui styling global

astewes:


cdap-ui/app/features/overview/templates/data-apps.html
```diff
@@ -0,0 +1,95 @@
...
+      </div>
+
+      <ul class="list-group" >
+        <li class="list-group-item" ng-repeat="stream in streams track by $index">
+          <a ui-sref="streams.detail.overview({streamId: stream.name})">
+            <strong ng-bind="stream.name | caskCapitalizeFilter">  </strong>
+            <div> <p ng-bind="'Type: ' + (stream.type)"></p> </div>
+          </a>
+        </li>
+        <li class="list-group-item" ng-repeat="dataset in datasets track by $index">
```
ajainarayanan: If we can move the list-group-item class to the <a> tag then the entire list item becomes clickable, instead of just the link tag (which in this case the text)

ajainarayanan:
:+1: LGTM

PR #1828, opened 2015-03-27
CDAP Tutorial Revisions for 2.8.0

awholegunch:
This is a drop-in replacement of everything at the [Current CDAP Tutorial](http://docs.cask.co/tutorial/current/en/intro.html).

The new version is staged at [cdap-tutorial](http://stg-web101.sw.joyent.continuuity.net/cdap-tutorial/2.8.0-r2.8_tutorial/en/index.html) rather than at "tutorial". 

It includes using MD5 checksums to guard against changes of the code samples rather than copying and comparing of files. It downloads the correct versions from GitHub and uses them to construct the docs.

I have tested the steps against 2.8.0 and supplied new screen shots where appropriate.

There are more files here than really should be needed. Because the currently tutorial lives off in its own directory (rather than being a sub-directory of the CDAP docs), and because 2.8.0 was already released, I simply copied the entire cdap docs common files and used them to build this small site. I would like to see this merged into and included in the cdap-docs as part of the main build as part of 3.0. A redirect can be used to help people find it. At that time, the duplication of files can be eliminated.


cdap-docs/tutorial/source/tutorial1.rst
```diff
@@ -0,0 +1,208 @@
...
+  ``libexec\bin`` directory as ``curl.exe``.
+- Other scripts referenced below are included either in the SDK or downloaded zips as ``.bat``
+  versions for Windows. Substitute these versions as appropriate in the examples below.
+
+Build and Run
+=============
+.. highlight:: console
+
+We’ll assume that you have already downloaded, installed, and have started an instance of
+CDAP, as described in the :ref:Tutorial Introduction <tutorial-intro>`.
```
albertshau: missing a tick mark in the ref, so the link doesn't render

cdap-docs/tutorial/source/tutorial2.rst
```diff
@@ -0,0 +1,684 @@
...
+
+The ``@GET`` annotation specifies the HTTP method used to reach the endpoint. The ``@Path``
+annotation defines the URL path used to reach this endpoint. This path has a single user
+parameter: ``{ip}``. It is decoded as a ``String`` in the parameters of the ``getIPCount()`` 
+method with the help of the ``@PathParam`` annotation.
+
+The ``PageViewCountHandler`` class is registered in the ``WiseService`` class, which has the
+implementation::
+
+  class WiseService extends AbstractService { @Override protected void configure() {
```
albertshau: code formatting here is off

albertshau:
couple comments, otherwise lgtm

cdap-docs/tutorial/source/tutorial2.rst
awholegunch: Oops. That's suppose to have been removed...

cdap-docs/tutorial/source/tutorial1.rst
awholegunch: fixed.

awholegunch:
Fixed per comments (thanks!); merging.

PR #1829, opened 2015-03-27
Fixes data interpolation with 0s for operations dashboard.

ajainarayanan:
- Backend send metrics only when it is emitted.
  So even if I ask for a particular metric for the last 1 min it will give me a series with some data only if that particular metric was emitted by something in the backend. Otherwise the UI gets an empty array.

- We are interpolating the rest of the values with 0s if we don't get any.

Get ```events.in``` metric for a flow in an app for the last 1 min:
Scenario1:
  Three events happend in the last one min. So we get exactly three values with three timestamps. Even though we are requesting metrics for 60 timestamps since the metric was not emitted at other time periods we will have only 3 values for corresponding 3 time stamps.

Scenario2:
  No events are in the last one min. So we get no time series data from the backend.

In both the cases we have startTime and endTime. Even if we get sporadic data for random timestamps within that startTime and endTime we need to draw a graph which is smooth. Thats the reason we are interpolating the rest of the timestamps with 0s so that the graph moves from right to left.

dashboardlive

### DIFF
--- a/cdap-ui/app/features/dashboard/widget.js
+++ b/cdap-ui/app/features/dashboard/widget.js
@@ -28,11 +28,22 @@ angular.module(PKG.name+'.feature.dashboard')
           method: 'POST'
         },
         (function (result) {
+          var data, tempMap = {};
           if(result.series && result.series.length) {
-            var data = result.series[0].data;
-            data.splice(data.length-1, 1);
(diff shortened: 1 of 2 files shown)


nistala:
LGTM

PR #1830, opened 2015-03-27
Feature/ui streams in admin

elfenheart:
add data under namespace sidebar in admin page. 

Able to create stream and edit stream properties from admin page.

### DIFF
--- a/cdap-ui/app/features/admin/admin.less
+++ b/cdap-ui/app/features/admin/admin.less
@@ -88,4 +88,10 @@ body.theme-cdap.state-admin {
       }
     }
   }
+
+  .modal-body {
+    .panel-title {
+      color: black;
+    }
+  }
 }
--- a/cdap-ui/app/features/admin/controllers/DatasetsController.js
+++ b/cdap-ui/app/features/admin/controllers/DatasetsController.js
@@ -1,9 +1,24 @@
 angular.module(PKG.name + '.feature.admin')
-  .controller('AdminDatasetsController', function ($scope) {
-    $scope.datasets = [{
-      name: 'STU',
-      type: 'Stream',
-      description: 'little miss muffet',
-      apps: '2'
-    }];
+  .controller('AdminDatasetsController', function ($scope, MyDataSource, $stateParams) {
+
+    var dataSrc = new MyDataSource($scope);
+
+    $scope.datasets = [];
+    $scope.streams = [];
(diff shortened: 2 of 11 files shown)


cdap-ui/app/features/admin/routes.js
```diff
@@ -143,6 +143,34 @@ angular.module(PKG.name + '.feature.admin')
               templateUrl: '/assets/features/admin/templates/namespace/datasets.html',
               controller: 'AdminDatasetsController'
             })
+              .state('admin.namespace.detail.datasets.streamcreate', {
```
ajainarayanan: Shouldn't this be ```admin.namespace.detail.streams.create```?

I believe we either display both streams & datasets in admin view under 'data' or we have individual streams & datasets submenus under namespace to show them individually. Right?

cdap-ui/app/features/admin/controllers/DatasetsController.js
```diff
@@ -1,9 +1,24 @@
...
+      _cdapPath: '/namespaces/' + $stateParams.nsadmin + '/data/datasets'
+    })
+      .then(function(res) {
+        $scope.datasets = res;
+      });
+
+    dataSrc.request({
+      _cdapPath: '/namespaces/' + $stateParams.nsadmin + '/streams'
+    }, function(res) {
+      if (angular.isArray(res) && res.length) {
```
ajainarayanan: Any specific reason why we are having this check here alone and not in datasets? Do we need this check?

cdap-ui/app/features/admin/admin.less
```diff
@@ -88,4 +88,10 @@ body.theme-cdap.state-admin {
       }
     }
   }
+
+  .modal-body {
```
ajainarayanan: Any specific reason why we are customizing modal in admin view here?

elfenheart: I am targeting the panel-title for the accordion (under modal). Without it, the text color is white. I am guessing this styling will get changed when we do the real styling for modals.

cdap-ui/app/features/admin/controllers/DatasetsController.js
elfenheart: updated

ajainarayanan:
:+1: LGTM

PR #1831, opened 2015-03-27
Fix Documentation mouseovers

awholegunch:
In our current documentation stylesheet, when you mouseover a link, the text changes in weight from normal to bold, causing a text-reflow and flash. It's very annoying. This replaces that with an underline and color-change instead as they style to indicate and activate link. If the link is already colored (as text links body copy are), the color goes from orange to dark-blue.

It also makes all links in body text normal rather than some normal and some in italic.
The only issue I see is this: is the orange color enough to show links in the body text? The underline only shows on mouseover.

- [The current style](http://docs.cask.co/cdap/current/en/admin-manual/installation/security.html)
- [The proposed change](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_fix_mouseovers/en/admin-manual/installation/security.html)

Fix for https://issues.cask.co/browse/CDAP-1944


cdap-docs/_common/_themes/cdap/static/cdap.css_t
```diff
@@ -148,10 +153,11 @@ div.sphinxsidebar a tt.docutils.literal {
     font-size: 1.0em;
 }
 
+/*
```
albertshau: is there a reason we are commenting out instead of just removing?

albertshau:
small question otherwise lgtm

cdap-docs/_common/_themes/cdap/static/cdap.css_t
awholegunch: Good catch. I forget to delete that, now that we aren't using it.

PR #1832, opened 2015-03-27
Removed StoreFactory since it does not provide much utility.

bdmogal:
Build: https://builds.cask.co/browse/CDAP-DUT1272


cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/AppFabricServiceRuntimeModule.java
```diff
@@ -266,10 +265,7 @@ protected void configure() {
           .build(new TypeLiteral<ManagerFactory<DeploymentInfo, ApplicationWithPrograms>>() { })
       );
 
-      install(new FactoryModuleBuilder()
-                .implement(Store.class, DefaultStore.class)
-                .build(StoreFactory.class)
-      );
+      bind(Store.class).to(DefaultStore.class).in(Scopes.SINGLETON);
```
chtyim: This change the behavior. When factory was used, a new instance is created every time upon request. Do you know if `Store` is safe to be a singleton?

anew: Store is not thread-safe... I think Terence is right and we need to provide a new instance every time.

anew:
@bdmogal I understand what you are trying to do. I do not think this is sufficient. Creating a new instance every time it is injected (instead of using a singleton) is not the same as creating a new instance every time it is used. If you inject this into a multi-threaded class, such as an HTTP handler, each thread will still use the same instance that was injected. If you inject a provider that creates a new instance every time it is invoked, makes every thread use a different instance.

Can we just drop this and leave it the way it was? :)

chtyim:
I think for all current usages of storefactory, they only do the create call in constructor, hence making the existence of factory class unnecessary. Also, in guice, for factory.create method that doesn't take any parameter, we better use the guice Provider instead if there is a need to create new instance besides in @Inject constructor. Provider binding are available by default for all bindings

anew:
I looked at the code a little more, and I concur with Terence. Good to me if you use the provider.

chtyim:
You don't need to use provider when the injection happens in construction.

chtyim:
LGTM

bdmogal:
Build passed - http://builds.cask.co/browse/CDAP-DUT1272-7

Merging.

PR #1837, opened 2015-03-30
(CDAP-1946) Make row key distributor size configurable

chtyim:


cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
```diff
@@ -492,10 +487,13 @@ public void create() throws IOException {
         addCoprocessor(htd, coprocessor, coprocessorJar.getJarLocation(), coprocessorJar.getPriority(coprocessor));
       }
 
-      // Create queue table with splits.
-      int splits = cConf.getInt(QueueConstants.ConfigKeys.QUEUE_TABLE_PRESPLITS,
-                                QueueConstants.DEFAULT_QUEUE_TABLE_PRESPLITS);
-      byte[][] splitKeys = HBaseTableUtil.getSplitKeys(splits);
+      // Create queue table with splits. The distributor bucket size is the same as splits.
+      int splits = cConf.getInt(QueueConstants.ConfigKeys.QUEUE_TABLE_PRESPLITS);
```
anew: Should we have a default here? Or is it safe to assume that this property is always there?

anew: So this is not a per-queue or per-flow setting? But global setting that applies to all queues?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
```diff
@@ -229,6 +242,15 @@ private HTable createHTable(TableId tableId) throws IOException {
     return consumerTable;
   }
 
+  private int getDistributorBuckets(HTableDescriptor htd) {
+    String value = htd.getValue(QueueConstants.DISTRIBUTOR_BUCKETS);
+    // If the setting is not in the table meta, this is a old table, hence use the value in the cConf
+    if (value == null) {
+      return cConf.getInt(QueueConstants.ConfigKeys.QUEUE_TABLE_PRESPLITS);
```
anew: hmmm interesting. So if I have a 2.8 queue where the descriptor does not carry the DISTRIBUTOR_BUCKETS, then upgrade to 3.0, and change my cdap-site and set QUEUE_TABLE_PRESPLITS to 24, what happens? We would use 24 although the queue was created with 16 (the old default). Would it be more correct to say, if it is not in the HTable descriptor, then it must be a pre-3.0 queue and therefore it is 16?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/hbase/HBaseStreamFileConsumerFactory.java
```diff
@@ -105,4 +112,19 @@ private synchronized HBaseAdmin getAdmin() throws IOException {
     }
     return admin;
   }
+
+  /**
+   * Creates a {@link AbstractRowKeyDistributor} based on the meta data in the given {@link HTableDescriptor}.
+   */
+  private AbstractRowKeyDistributor createKeyDistributor(HTableDescriptor htd) {
+    int buckets = cConf.getInt(Constants.Stream.CONSUMER_TABLE_PRESPLITS);
```
anew: similar to my comment above, is this the right default to use?

anew:
I have only one major concern (in two places).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueAdmin.java
chtyim: Yes it's a global setting. The default is defined in `cdap-default.xml`, hence this get should always return a value.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
chtyim: That's true. Probably should just hard code it to 16.

chtyim:
Updated to use 16 buckets by default if the setting is missing from the table attribute.

anew:
LGTM

chtyim:
Squashed two commits into two to make cherry pick into 2.8 branch easier.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1278-3

PR #1839, opened 2015-03-30
Removed ApplicationBundler to use Twill's instead

jwang47:
http://builds.cask.co/browse/CDAP-DUT1281


chtyim:
Don't do this until TWILL-112 is fixed and Twill 0.6.0 is released.

PR #1840, opened 2015-03-30
Feature/etl rest api

gokulavasan:
Mock Endpoints for ETL specific REST endpoints.

Build : http://builds.cask.co/browse/CDAP-DT3

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MockETLHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MockETLHandler.java
@@ -44,7 +44,6 @@ public class MockETLHandler extends AuthenticatedHttpHandler {
     super(authenticator);
   }
 
-  @Path("/")
   @GET
   public void getTemplates(HttpRequest request, HttpResponder responder) throws Exception {
     JsonArray templates = new JsonArray();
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MockETLHandler.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Inject
+  public MockETLHandler(Authenticator authenticator) {
+    super(authenticator);
+  }
+
+  @Path("/")
+  @GET
+  public void getTemplates(HttpRequest request, HttpResponder responder) throws Exception {
+    List<Map<String, String>> templates = Lists.newArrayList();
```
jwang47: May be worth it to use objects here instead of `ImmutableMap`s.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MockETLHandler.java
```diff
@@ -0,0 +1,189 @@
...
+
+  private static final String REALTIME = "etlrealtime";
+  private static final String BATCH = "etlbatch";
+
+  @Inject
+  public MockETLHandler(Authenticator authenticator) {
+    super(authenticator);
+  }
+
+  @Path("/")
```
albertshau: i think you can just leave this out

albertshau:
minor comment, then lgtm as a mock

gokulavasan:
Build passed. Merging.

PR #1841, opened 2015-03-30
CDAP-614: Support port 0 for ExternalAuthenticationServer

jwang47:
http://builds.cask.co/browse/CDAP-DUT1282

### DIFF
--- a/cdap-security/src/main/java/co/cask/cdap/security/server/ExternalAuthenticationServer.java
+++ b/cdap-security/src/main/java/co/cask/cdap/security/server/ExternalAuthenticationServer.java
@@ -111,12 +111,12 @@ public class ExternalAuthenticationServer extends AbstractExecutionThreadService
    */
   public InetSocketAddress getSocketAddress() {
     if (!server.isRunning()) {
-      return new InetSocketAddress(address, port);
-    } else {
-      // assumes we only have one connector
-      final Connector connector = server.getConnectors()[0];
-      return new InetSocketAddress(connector.getHost(), connector.getLocalPort());
+      throw new IllegalStateException("Server not started yet");
     }
+
+    // assumes we only have one connector
+    final Connector connector = server.getConnectors()[0];
+    return new InetSocketAddress(connector.getHost(), connector.getLocalPort());
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


cdap-security/src/main/java/co/cask/cdap/security/server/ExternalAuthenticationServer.java
```diff
@@ -110,11 +110,21 @@ public ExternalAuthenticationServer(CConfiguration configuration, SConfiguration
    * @return InetSocketAddress of server.
    */
   public InetSocketAddress getSocketAddress() {
-    return new InetSocketAddress(address, port);
+    if (!server.isRunning()) {
```
albertshau: does anybody call this before starting the server?  Would it be better just to add a javadoc that this needs to be called after the server is started, otherwise you can get an invalid address?

cdap-security/src/main/java/co/cask/cdap/security/server/ExternalAuthenticationServer.java
```diff
@@ -110,11 +110,21 @@ public ExternalAuthenticationServer(CConfiguration configuration, SConfiguration
...
+      return new InetSocketAddress(connector.getHost(), connector.getLocalPort());
+    }
   }
 
   @Override
   protected void run() throws Exception {
+    server.start();
+
+    // assumes we only have one connector
+    final Connector connector = server.getConnectors()[0];
```
albertshau: are connectors guaranteed to be set at this point?

albertshau:
not familiar with how it's used/works, so a couple questions

cdap-security/src/main/java/co/cask/cdap/security/server/ExternalAuthenticationServer.java
jwang47: Yup, we set them up when creating the server: https://github.com/caskdata/cdap/pull/1841/files#diff-875ec12d39f8e6b2cf4725b9ec4b4305R193 and https://github.com/caskdata/cdap/pull/1841/files#diff-875ec12d39f8e6b2cf4725b9ec4b4305R198.

cdap-security/src/main/java/co/cask/cdap/security/server/ExternalAuthenticationServer.java
jwang47: Changed to throw an exception if it's called before server is started.

cdap-security/src/main/java/co/cask/cdap/security/server/ExternalAuthenticationServer.java
albertshau: hm but is that when the 0 gets resolved to the actual local port it bound to?  Or does server.start() guarantee that?

jwang47: server.start() guarantees the actual port to be given from that connector

albertshau:
:+1:

PR #1846, opened 2015-03-31
Clean-up explore tests

poornachandra:


poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT1283

cdap-explore/src/test/java/co/cask/cdap/explore/service/ExploreServiceTestsSuite.java
```diff
@@ -31,12 +33,17 @@
   HiveExploreServiceTestRun.class,
   WritableDatasetTestRun.class,
   HiveExploreObjectMappedTableTestRun.class,
-  HiveExploreServiceStopTestRun.class
```
albertshau: why remove?

cdap-explore/src/test/java/co/cask/cdap/explore/service/ExploreServiceTestsSuite.java
```diff
@@ -31,12 +33,17 @@
...
+  HiveExploreServiceFileSetTest.class
```
albertshau: should rename to HiveExploreServiceFileSetTestRun so that maven will not run it both in the suite and outside the suite

cdap-explore/src/test/java/co/cask/cdap/explore/service/BaseHiveExploreServiceTest.java
```diff
@@ -367,6 +373,16 @@ protected void configure() {
                     .build(StoreFactory.class)
           );
           bind(StreamHttpService.class).in(Scopes.SINGLETON);
+
+          install(Modules.override(new DataFabricModules().getInMemoryModules()).with(new AbstractModule() {
```
albertshau: why is this needed?

albertshau:
couple questions and one comment

cdap-explore/src/test/java/co/cask/cdap/explore/service/ExploreServiceTestsSuite.java
poornachandra: Since `HiveExploreServiceStopTestRun` stops the service, it is a good idea not to run it as part of the test suite.

cdap-explore/src/test/java/co/cask/cdap/explore/service/BaseHiveExploreServiceTest.java
poornachandra: I changed the way transaction was asserted in `HiveExploreServiceStopTestRun`, so that we don't have conditional logic in `BaseHiveExploreServiceTest` code. Now all services get stopped, and transaction state is asserted using transaction snapshot.

cdap-explore/src/test/java/co/cask/cdap/explore/service/ExploreServiceTestsSuite.java
poornachandra: The JUnit docs don't say anything about excluding Test files that are part of TestSuite from outside the suite. But I see that our config is setup that way. I'll fix the class names.

poornachandra:
@albertshau Addressed review comments and did some more refactoring. Please take a look.

cdap-explore/src/test/java/co/cask/cdap/explore/service/BaseHiveExploreServiceTest.java
```diff
@@ -138,24 +144,23 @@
 
   protected static Injector injector;
 
-  @ClassRule
-  public static TemporaryFolder tmpFolder = new TemporaryFolder();
-
-  protected static void initialize() throws Exception {
-    initialize(CConfiguration.create());
+  protected static void initialize(TemporaryFolder tmpFolder) throws Exception {
```
albertshau: how come it has to be passed in? isn't it get a new folder each time initialize is called?

poornachandra: I'm now passing in `TemporaryFolder` in initialize, otherwise when we call `intitalize` from `BaseHiveExploreServiceTest` the `@ClassRule` may not be initiailzed.

albertshau:
:+1:

poornachandra:
Build is green http://builds.cask.co/browse/CDAP-DUT1283.

PR #1850, opened 2015-03-31
Feature/cdap 1227 pfs output committer

anew:
- adds an interface for datasets to run code after completion of a batch job. This is like onFinish() for MapReduce without the dependency on Hadoop. This interface is called from the MapReduce runner before calling onFinish()
- partitioned file sets implement that interface to add the output partition to the meta data on success
- removed the addPartition from our existing MapReduces (adapter, examples, tests)
- changed addPartition() to succeed even if the same partition already exists - if it was added in the same transaction. This is to keep existing jobs from failing if they have the addPartition in their onFinish().
- added a unit test for backward compatibility with such existing apps.
- updated the docs to remove mentions of this limitation

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/data/batch/DatasetOutputCommitter.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/data/batch/DatasetOutputCommitter.java
@@ -20,7 +20,7 @@ import co.cask.cdap.api.dataset.DataSetException;
 
 /**
  * This interface is implemented by a dataset if at the end of a batch job (MapReduce, Spark, ...)
- * the output needs to be committed, or rolled back, depending one success of the job. This is
+ * the output needs to be committed, or rolled back, depending on success of the job. This is
  * similar to what Hadoop's OutputCommitter does, however, this is dataset-centric and does not
  * assume any dependencies on Hadoop.
  */
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
@@ -401,8 +401,8 @@ final class MapReduceRuntimeService extends AbstractExecutionThreadService {
                 ((DatasetOutputCommitter) outputDataset).onFailure();
               }
             } catch (Throwable t) {
-              LOG.error("Error from {} method of output dataset {}.",
-                        succeeded ? "onSuccess" : "onFailure", outputDataset);
+              LOG.error(String.format("Error from %s method of output dataset %s.",
+                        succeeded ? "onSuccess" : "onFailure", context.getOutputDatasetName()), t);
               success = false;
             }
           }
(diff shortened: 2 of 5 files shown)


anew:
Build running here: http://builds.cask.co/browse/CDAP-DUT1290

awholegunch:
I will build the docs from this change and check the includes, as it looks to me as if the changes you have made need to be adjusted a bit. It's hard to tell from the diffs; best to build and check that way.

anew:
thanks @awholegunch. I believe that the includes should still be correct, because I removed code at the end of the file. But I agree, better check...

awholegunch:
I've added a few corrections, and staged the results at http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-cdap-1227-pfs-output-committer/en/examples-manual/examples/index.html

awholegunch:
There is an issue that when includes are added, leading spaces at the start of the line cause the included code samples to be "shifted" to the right. (They aren't actually shifted, it's just that they look like they are.) There is a fix for this; I need to upgrade the version of Sphinx we use and update the rst code.

anew:
The staging server does not respond to me. I built the docs locally with your changes. Looks good to me. Thanks

anew:
Docs are reviewed and in good shape. Waiting for review of code changes.

cdap-api/src/main/java/co/cask/cdap/api/data/batch/DatasetOutputCommitter.java
```diff
@@ -0,0 +1,38 @@
...
+ * the License.
+ */
+
+package co.cask.cdap.api.data.batch;
+
+import co.cask.cdap.api.dataset.DataSetException;
+
+/**
+ * This interface is implemented by a dataset if at the end of a batch job (MapReduce, Spark, ...)
+ * the output needs to be committed, or rolled back, depending one success of the job. This is
```
albertshau: typo: "one" -> "on"

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -389,7 +390,23 @@ private void onFinish(final boolean succeeded) throws TransactionFailureExceptio
...
+          if (outputDataset != null && outputDataset instanceof DatasetOutputCommitter) {
+            try {
+              if (succeeded) {
+                ((DatasetOutputCommitter) outputDataset).onSuccess();
+              } else {
+                ((DatasetOutputCommitter) outputDataset).onFailure();
+              }
+            } catch (Throwable t) {
+              LOG.error("Error from {} method of output dataset {}.",
+                        succeeded ? "onSuccess" : "onFailure", outputDataset);
```
albertshau: should we log the error as well?

albertshau:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
anew: oh, yes. I wanted to.

cdap-api/src/main/java/co/cask/cdap/api/data/batch/DatasetOutputCommitter.java
anew: sure

PR #1851, opened 2015-03-31
Update installation instructions

awholegunch:
Ports a change in the installation instructions from 2.6.2 to 2.8.0. and fixes command examples (makes all command start with a "$") and cleans up a description and minor formatting issues.

Staged at http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_update_installation/en/admin-manual/installation/installation.html#preparing-package-managers

### DIFF
--- a/cdap-docs/admin-manual/source/installation/installation.rst
+++ b/cdap-docs/admin-manual/source/installation/installation.rst
@@ -115,7 +115,7 @@ for Linux and Solaris must be installed in your environment; we recommend the Or
 
 To check the Java version installed, run the command::
 
-  java -version
+  $ java -version
   
 CDAP is tested with the Oracle JDKs; it may work with other JDKs such as 
 `Open JDK <http://openjdk.java.net>`__, but it has not been tested with them.
@@ -204,7 +204,7 @@ Preparing the Cluster
 To prepare your cluster so that CDAP can write to its default namespace,
 create a top-level ``/cdap`` directory in HDFS, owned by an HDFS user ``yarn``::
 
-  sudo -u hdfs hadoop fs -mkdir /cdap && hadoop fs -chown yarn /cdap
+  $ sudo -u hdfs hadoop fs -mkdir /cdap && hadoop fs -chown yarn /cdap
 
 In the CDAP packages, the default HDFS namespace is ``/cdap`` and the default HDFS user is
 ``yarn``. If you set up your cluster as above, no further changes are required.
@@ -267,7 +267,7 @@ Add the Cask Public GPG Key to your repository:
 
 Update your Yum cache::
 
-  sudo yum makecache
+  $ sudo yum makecache
 
 Debian using APT
 ................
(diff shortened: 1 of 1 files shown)


wolf31o2:
:+1: LGTM

cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -595,23 +602,23 @@ you have a use case for it, please reach out to us at `cdap-user@googlegroups.co
...
 
 #. Update the CDAP packages by running either of these methods:
 
    - Using Yum (on one line)::
 
-       sudo yum install cdap cdap-gateway
-                             cdap-hbase-compat-0.94 cdap-hbase-compat-0.96
-                             cdap-hbase-compat-0.98 cdap-kafka cdap-master
-                             cdap-security cdap-web-app
+       $ sudo yum install cdap cdap-gateway
```
duddun: We need to avoid displaying commands wrapped onto multiple lines, unless they have '\' at the end of each line that wraps to another. The user should be able to copy (from after command prompt, if applicable) and paste the whole thing into a terminal window and have it run as expected.  This may look 'prettier', but it will not work as it is written here.

cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -595,23 +602,23 @@ you have a use case for it, please reach out to us at `cdap-user@googlegroups.co
...
+                               cdap-hbase-compat-0.98 cdap-kafka cdap-master
+                               cdap-security cdap-web-app
 
    - Using APT (on one line)::
 
-       sudo apt-get install cdap cdap-gateway
-                            cdap-hbase-compat-0.94 cdap-hbase-compat-0.96
-                            cdap-hbase-compat-0.98 cdap-kafka cdap-master
-                            cdap-security cdap-web-app
+       $ sudo apt-get install cdap cdap-gateway
```
duddun: Same as wrapping comment above.
It should like: 
$ sudo apt-get install cdap cdap-gateway \
                                  cdap-hbase-compat-0.94 cdap-hbase-compat-0.96 \
                                  cdap-hbase-compat-0.98 cdap-kafka cdap-master \
                                  cdap-security cdap-web-app

duddun:
An additional comment (personal preference).
Command prompts very rarely look like '$'.  Yes, it is a convention in some (documentation) places, but if we use something that is never going to exactly mimic what people will see on their command lines, we could consider just not using anything at all, (with the assumption these commands are made obvious by their formatting or how they are introduced).  This will allow the reader/user to more easily copy paste commands from the documentation to their terminal.

awholegunch:
We have settled on a convention of indicating commands with a "$"; it means that lines that don't start with a "$" are something else, usually a code sample or a file sample. As this is a convention used in most commercial documentation manuals, I would be loathe to deviate from it.

cdap-docs/admin-manual/source/installation/installation.rst
awholegunch: Added backslashes and reformat the lines.

cdap-docs/admin-manual/source/installation/installation.rst
awholegunch: ibid.

PR #1852, opened 2015-03-31
CDAP-1298: Fix a/an in CLI command descriptions

jwang47:
https://issues.cask.co/browse/CDAP-1298
http://builds.cask.co/browse/CDAP-DUT1294


awholegunch:
Looking over the code, it strikes me that there is a lot of duplication that could be factored out. For instance, "type.getNamePlural().toString()"; could not this be replaced by a "type.getStringNamePlural()"?

And, 
"Fragment.of(Article.A, ElementType.ADAPTER.getTitleName())": could this not be simplified?

could ElementType return the correct version of itself?:

"ElementType.ADAPTER.getATitleName()"

awholegunch:
In "cdap-docs/reference-manual/source/cli-api.rst", has this been generated, or were the lines manually edited? It should be the result of running the builder command, so that we know that it works...

jwang47:
Regarding the duplication: I preferred `type.getNamePlural().toString()` over `type.getStringNamePlural()` so that there's only one way to get the plural name in `String` form.

The reason for `Fragment.of()` is that we could also use it in the future for things like `Fragment.of(Article.number(5), ElementType.ADAPTER.getTitleName())`, but it's not necessary right yet.

`cli-api.rst` has been generated.

cdap-docs/reference-manual/source/cli-api.rst
```diff
@@ -122,7 +122,7 @@ These are the available commands:
    :header: Command,Description
    :widths: 50, 50
 
-   **General**
+     **General**
```
awholegunch: These two spaces will break the table format.

cdap-docs/reference-manual/source/cli-api.rst
```diff
@@ -225,30 +225,30 @@ These are the available commands:
...
+   ``load preferences app <local-file-path> <content-type> [<app-id>]``,"Set Preferences of an Application from a local Config File (supported formats = JSON)."
+   ``load preferences flow <local-file-path> <content-type> [<app-id.flow-id>]``,"Set Preferences of a Flow from a local Config File (supported formats = JSON)."
+   ``load preferences instance <local-file-path> <content-type> [<instance-id>]``,"Set Preferences of an Instance from a local Config File (supported formats = JSON)."
+   ``load preferences mapreduce <local-file-path> <content-type> [<app-id.mapreduce-id>]``,"Set Preferences of a MapReduce Program from a local Config File (supported formats = JSON)."
+   ``load preferences namespace <local-file-path> <content-type> [<namespace-name>]``,"Set Preferences of a Namespace from a local Config File (supported formats = JSON)."
+   ``load preferences procedure <local-file-path> <content-type> [<app-id.procedure-id>]``,"Set Preferences of a Procedure from a local Config File (supported formats = JSON)."
```
awholegunch: Are Procedures going away in 3.0?

cdap-docs/reference-manual/source/cli-api.rst
jwang47: Oops, fixed.

awholegunch:
LGTM. If procedures go away in 3.0, we can assume this page will be re-built.

jwang47:
I think it will be better to use some linguistics library instead of having my own crude and incomplete implementation..

jwang47:
Actually, it's not really needed right now. Will merge as is.

PR #1853, opened 2015-03-31
CDAP-1897: Show config entry source in JSON for ConfigHandler

jwang47:
https://issues.cask.co/browse/CDAP-1897
http://builds.cask.co/browse/CDAP-DUT1295

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/conf/Configuration.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/conf/Configuration.java
@@ -16,6 +16,7 @@
 
 package co.cask.cdap.common.conf;
 
+import co.cask.cdap.api.annotation.Beta;
 import co.cask.cdap.common.utils.DirUtils;
 import com.google.common.base.Preconditions;
 import com.google.gson.stream.JsonWriter;
@@ -995,6 +996,41 @@ public class Configuration implements Iterable<Map.Entry<String, String>> {
     } else {
       set(name, pattern.pattern());
     }
+  }
+
+
+  /**
+   * Gets information about why a property was set.  Typically this is the
+   * path to the resource objects (file, URL, etc.) the property came from, but
+   * it can also indicate that it was set programatically, or because of the
+   * command line.
+   *
+   * @param name - The property name to get the source of.
+   * @return null - If the property or its source wasn't found. Otherwise,
+   * returns a list of the sources of the resource.  The older sources are
+   * the first ones in the list.  So for example if a configuration is set from
+   * the command line, and then written out to a file that is read back in the
+   * first entry would indicate that it was set from the command line, while
+   * the second one would indicate the file that the new configuration was read
+   * in from.
+   */
+  @Beta
+  public synchronized String[] getPropertySources(String name) {
+    if (properties == null) {
+      // If properties is null, it means a resource was newly added
+      // but the props were cleared so as to load it upon future
+      // requests. So lets force a load by asking a properties list.
+      getProps();
+    }
+
+    // Return a null right away if our properties still
+    // haven't loaded or the resource mapping isn't defined
+    if (properties == null || updatingResource == null) {
+      return new String[0];
+    }
+
+    String resource = updatingResource.get(name);
+    return (resource == null) ? new String[0] : new String[] { resource };
   }
 
   /**
--- a/cdap-common/src/main/java/co/cask/cdap/gateway/handlers/ConfigHandler.java
+++ b/cdap-common/src/main/java/co/cask/cdap/gateway/handlers/ConfigHandler.java
@@ -21,9 +21,7 @@ import co.cask.cdap.common.conf.Constants;
 import co.cask.http.AbstractHttpHandler;
 import co.cask.http.HttpResponder;
 import com.google.common.base.Charsets;
-import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
-import org.apache.hadoop.conf.Configuration;
 import org.jboss.netty.buffer.ChannelBuffer;
 import org.jboss.netty.buffer.ChannelBuffers;
 import org.jboss.netty.handler.codec.http.HttpRequest;
@@ -32,9 +30,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.io.StringWriter;
 import java.nio.ByteOrder;
-import java.util.Map;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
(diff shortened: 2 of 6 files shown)


sreevatsanraman:
There are no tests. Can you add a test?

nitinmotgi:
@jwang47 can you address this immediately.

nitinmotgi:
:-1: There is a major flaw with this approach. Router might be outside of firewall and might not have the configurations available for it. So, these endpoints should be dispatched to master. Please address this issue immediately.

cdap-common/src/main/java/co/cask/cdap/common/conf/Configuration.java
```diff
@@ -19,6 +19,7 @@
 import co.cask.cdap.common.utils.DirUtils;
 import com.google.common.base.Preconditions;
 import com.google.gson.stream.JsonWriter;
+import org.apache.hadoop.classification.InterfaceStability;
```
chtyim: Why use hadoop annotation in common package?

chtyim: This is an internal class, ok not to have annotation like this. If needed, use `@Beta`.

cdap-common/src/main/java/co/cask/cdap/common/conf/Configuration.java
```diff
@@ -997,6 +998,44 @@ public void setPattern(String name, Pattern pattern) {
...
+      // If properties is null, it means a resource was newly added
+      // but the props were cleared so as to load it upon future
+      // requests. So lets force a load by asking a properties list.
+      getProps();
+    }
+    // Return a null right away if our properties still
+    // haven't loaded or the resource mapping isn't defined
+    if (properties == null || updatingResource == null) {
+      return null;
+    } else {
```
chtyim: No need to have else after return statement in the if block.

cdap-common/src/main/java/co/cask/cdap/common/conf/Configuration.java
```diff
@@ -997,6 +998,44 @@ public void setPattern(String name, Pattern pattern) {
...
+      String resource = updatingResource.get(name);
+      if (resource == null) {
```
chtyim: Use `return (resources == null) ? null : new String[] { resource };`

cdap-common/src/main/java/co/cask/cdap/gateway/handlers/ConfigHandler.java
```diff
@@ -97,15 +98,60 @@ public void configHBase(@SuppressWarnings("UnusedParameters") HttpRequest reques
...
-    ImmutableMap.Builder<String, String> result = ImmutableMap.builder();
+  private String getFirstElement(String[] array) {
+    if (array != null && array.length >= 1) {
+      return array[0];
+    } else {
+      return null;
+    }
+  }
+
+  private List<ConfigEntry> toObject(Configuration configuration) {
```
chtyim: Rather than super generic name `toObject`, better call it `toConfigEntries` or `toConfigEntryList`.

cdap-common/src/main/java/co/cask/cdap/common/conf/Configuration.java
jwang47: This method was copied from Hadoop's Configuration class. Should I just add a note without refactoring it?

cdap-common/src/main/java/co/cask/cdap/common/conf/Configuration.java
```diff
@@ -997,6 +998,43 @@ public void setPattern(String name, Pattern pattern) {
     }
   }
 
+
+  /**
+   * From org.apache.hadoop.conf.Configuration, with minor modifications.
```
chtyim: No need to mention this.

jwang47:
Updated PR to run `ConfigHandler` from master.

cdap-common/src/main/java/co/cask/cdap/common/conf/CConfiguration.java
```diff
@@ -34,7 +34,7 @@
   private static final Logger LOG =
     LoggerFactory.getLogger(CConfiguration.class);
 
-  private CConfiguration() {
+  public CConfiguration() {
```
chtyim: why the change? And it's different than what the comment inside this constructor.

cdap-common/src/test/java/co/cask/cdap/gateway/handlers/ConfigServiceTest.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ *
+ */
+public class ConfigServiceTest {
+
+  @Test
+  public void testConfig() {
+
+    // cConf
+    CConfiguration cConf = new CConfiguration();
```
chtyim: Shouldn't do this. You can do a `cConf.clear()`.

cdap-common/src/main/java/co/cask/cdap/gateway/handlers/ConfigHandler.java
```diff
@@ -62,12 +55,11 @@ public ConfigHandler(CConfiguration cConf, Configuration hConf) {
   public void configCDAP(@SuppressWarnings("UnusedParameters") HttpRequest request, HttpResponder responder,
                         @DefaultValue("json") @QueryParam("format") String format) {
```
chtyim: mis-align.

cdap-common/src/main/java/co/cask/cdap/common/conf/Configuration.java
```diff
@@ -997,6 +998,41 @@ public void setPattern(String name, Pattern pattern) {
...
+   * @return null - If the property or its source wasn't found. Otherwise,
+   * returns a list of the sources of the resource.  The older sources are
+   * the first ones in the list.  So for example if a configuration is set from
+   * the command line, and then written out to a file that is read back in the
+   * first entry would indicate that it was set from the command line, while
+   * the second one would indicate the file that the new configuration was read
+   * in from.
+   */
+  @Beta
+  public synchronized String[] getPropertySources(String name) {
```
chtyim: If it will return null, add `@Nullable`. BTW, why not return an empty array if property not found?

cdap-gateway/src/main/java/co/cask/cdap/gateway/router/RouterPathLookup.java
```diff
@@ -186,6 +186,11 @@ private String getV3RoutingService(String [] uriParts, AllowedMethod requestMeth
       // /v3/namespaces/{namespace-id}/data/datasets/{name}/properties
       // /v3/namespaces/{namespace-id}/data/datasets/{name}/admin/{method}
       return Constants.Service.DATASET_MANAGER;
+    } else if (uriParts.length >= 2 && uriParts[1].equals("config")) {
+      // TODO: config endpoint only hits stream service
```
chtyim: What does this comment mean? config endpoint hits stream service??

jwang47: Router will route calls to /v3/config/* to stream service. Wasn't sure where would be a better place to route /v3/config/* requests and run `ConfigHandler`, so right now it's hitting stream service.

chtyim: You mean the stream handler? So the stream handler is handling /v3/config calls to give the cConf?

jwang47: It's the `StreamHttpService` that runs the `ConfigHandler` -- see https://github.com/caskdata/cdap/blob/develop/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamServiceModule.java#L39: `CommonHandlers.add(handlerBinder);` adds the `ConfigHandler`.

chtyim: So we are having the config handler in every http service? That seems very wrong. Why not route to master (app-fabric)?

jwang47: True, need to route to app-fabric.

sreevatsanraman:
LGTM :+1:

PR #1854, opened 2015-03-31
interpolator support and auto resolutions fix in metrics v3 api

gsps1:
https://issues.cask.co/browse/CDAP-1910
https://issues.cask.co/browse/CDAP-1922

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/metrics/MetricsConstants.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/metrics/MetricsConstants.java
@@ -48,8 +48,8 @@ public final class MetricsConstants {
   public static final int DEFAULT_TIME_SERIES_TABLE_ROLL_TIME = 3600;
   public static final long DEFAULT_RETENTION_HOURS = 2;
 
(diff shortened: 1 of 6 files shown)


cdap-watchdog/src/main/java/co/cask/cdap/metrics/MetricsConstants.java
```diff
@@ -48,8 +48,8 @@
   public static final int DEFAULT_TIME_SERIES_TABLE_ROLL_TIME = 3600;
   public static final long DEFAULT_RETENTION_HOURS = 2;
 
-  public static final long METRICS_HOUR_RESOLUTION_CUTOFF = 3610;
-  public static final long METRICS_MINUTE_RESOLUTION_CUTOFF = 610;
+  public static final long METRICS_HOUR_RESOLUTION_CUTOFF = 36000;
```
abaranau: no need to have METRICS prefix. Also might wanna call it MAX_HOUR_RESOLUTION_QUERY_INTERVAL or smth

abaranau:
very minor - fix and merge

gsps1:
done. merging.

PR #1858, opened 2015-03-31
Use provided namespace when namespace isn't provided in IntegrationTestBase.deployApplication

jwang47:
http://builds.cask.co/browse/CDAP-DUT1302


bdmogal:
:+1:

PR #1862, opened 2015-04-01
Fix datasets collection used during the creation of MapReduce Context

gokulavasan:
Bamboo : http://builds.cask.co/browse/CDAP-DUT1305

This fix includes input/output dataset names in the dataset collection used in the creation of the MapReduceContext. For example, this will enable the use case illustrated in the test added in this PR - accessing datasets without adding the names during compile time.

### DIFF
--- a/cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetWithMRApp.java
+++ b/cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetWithMRApp.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


gokulavasan:
@albertshau Please review! Thanks.

cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetWithMRApp.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public void beforeSubmit(MapReduceContext context) {
+      context.setInput(context.getRuntimeArguments().get(INPUT_KEY));
+      context.setOutput(context.getRuntimeArguments().get(OUTPUT_KEY));
+      Job hadoopJob = context.getHadoopJob();
+      hadoopJob.setMapperClass(IdentityMapper.class);
+      hadoopJob.setNumReduceTasks(0);
+    }
+
+    public static class IdentityMapper extends Mapper<byte[], byte[], byte[], byte[]>
```
albertshau: think you can remove implements ProgramLifecycle

albertshau:
minor comment, otherwise lgtm assuming build passes

gokulavasan:
Build passed. Merging.

PR #1863, opened 2015-04-01
Creates a "Wise" tutorial in the examples-manual.

awholegunch:
Modifies the build.sh to include downloading of Java and image files used by the rst source.
Creates a new tutorials directory that adds the Wise tutorial, and no longer links to the docs.cask.co/tutorials.
Modifies the conf.py to include wise_version, and re-organizes the rst_epilog to handle missing definitions.
It looks like a big PR (54 files!) but most of those are deleting (49 deletions) files that are no longer required.


sreevatsanraman:
LGTM :+1:

PR #1864, opened 2015-04-01
Fix/ui flow nested tabs

ajainarayanan:
- [x] Fixes routes issues in Flows Feature
- [x] Fixes back button in the Flows feature for straight forward navigation among views
- [x] Improves performance by caching data (make one polling between shared controllers instead of N) (https://github.com/caskdata/cdap/pull/1865)
- [ ] Fix updating the status of each runs in the accordion title.
- [x] Fix controller name usage consistent (that could be adopted in other features) (https://github.com/caskdata/cdap/pull/1865)
- [ ] Fix stream data injection through flow diagram (right now injecting data into a stream in a flow is disabled)

runsview


ajainarayanan:
Split into another PR https://github.com/caskdata/cdap/pull/1865 (for performance + controller naming convention + data caching)

elfenheart:
:+1: LGTM

PR #1865, opened 2015-04-01
Fix/ui nested tabs perf matters

ajainarayanan:
- Fixes performance of run accordions in flows feature
- Fixes Controller naming convention (sanity in insane list of controllers)
- Minor improvements on data caching in service instead of making ginormous REQUESTs to backend.

### DIFF
--- a/cdap-ui/app/features/flows/controllers/tabs/runs/tabs/status-ctrl.js
+++ b/cdap-ui/app/features/flows/controllers/tabs/runs/tabs/status-ctrl.js
@@ -27,9 +27,6 @@ angular.module(PKG.name + '.feature.flows')
       })
 
     // This controller is NOT shared between the accordions.
-    // We need to share it and cache it in a service so that
-    // all the controllers don't make the poll but one guy does.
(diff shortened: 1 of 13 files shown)


cdap-ui/app/features/flows/controllers/tabs/runs/tabs/status-ctrl.js
```diff
@@ -21,16 +21,16 @@ angular.module(PKG.name + '.feature.flows')
...
 
-    FlowDiagramData.fetchData()
+    FlowDiagramData.fetchData($state.params.appId, $state.params.programId)
       .then(function(data) {
         $scope.data = data;
       })
 
-    // This controller is shared between the accordions.
+    // This controller is NOT shared between the accordions.
     // We need to share it and cache it in a service so that
```
anwar6953: line 30 needs to be removed, it seems

ajainarayanan: :) good catch! Fixed

elfenheart:
:+1: LGTM

PR #1866, opened 2015-04-01
support limiting result metrics size

gsps1:
- user supplied limit applied on result time series
- hard limit on records to scan during a query


cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
```diff
@@ -370,6 +372,38 @@ public void testSearchMetrics() throws Exception {
...
+    // 30 second
+    metricStore.add(new MetricValue(sliceBy, "reads", start + 30, 1, MetricType.COUNTER));
+    // 1 minute
+    metricStore.add(new MetricValue(sliceBy, "reads", start + 60, 1, MetricType.COUNTER));
+    // 10 minutes
+    metricStore.add(new MetricValue(sliceBy, "reads", start + 600, 1, MetricType.COUNTER));
+    // 1 hour
+    metricStore.add(new MetricValue(sliceBy, "reads", start + 3600, 1, MetricType.COUNTER));
+
+    // just one record
```
abaranau: put in comment that you test:
* limit is 1
* limit is less than data points in time range
* limit is greater than data points in time range

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricQueryParser.java
```diff
@@ -46,6 +46,7 @@
   private static final String COUNT = "count";
   private static final String START_TIME = "start";
   private static final String RESOLUTION = "resolution";
+  private static final String LIMIT = "limit";
```
abaranau: isn't count same as limit? Do we need both params?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/cube/DefaultCube.java
```diff
@@ -52,7 +52,8 @@
   private static final Logger LOG = LoggerFactory.getLogger(DefaultCube.class);
 
   private static final TagValueComparator TAG_VALUE_COMPARATOR = new TagValueComparator();
-
+  // hard-limit on max records to scan
+  private static final int MAX_RECORDS_TO_SCAN = 10 * 1000;
```
abaranau: 100 * 1000

abaranau:
couple minor comments - one bigger about duplicating param

abaranau:
LGTM

PR #1870, opened 2015-04-01
Allow quotes to be used for keys and values in CLI map arguments

jwang47:
http://builds.cask.co/browse/CDAP-DUT1313


cdap-cli/src/main/java/co/cask/cdap/cli/util/ArgumentParser.java
```diff
@@ -47,7 +46,29 @@
     if (mapString == null || mapString.isEmpty()) {
       return ImmutableMap.of();
     }
-    return Splitter.on(" ").omitEmptyStrings().trimResults().withKeyValueSeparator("=").split(mapString);
+
+
```
gokulavasan: Remove extra line

gokulavasan:
minor nit. LGTM otherwise.

PR #1871, opened 2015-04-01
Fix/ui set stream properties

elfenheart:
streamproperties


bdmogal:
The stream format should probably be a drop down/auto complete. May not be a good idea to allow free text there?

ajainarayanan:
:+1: LGTM

PR #1872, opened 2015-04-01
CDAP-1768 Updating the adapter lifecycle apis. Almost everything

albertshau:
is moved out of the manifest, as what was present there is
expected to be done programmatically. Source and Sink were removed,
as they will be moved to an ETL specific template. Also making
ApplicationSpecification a parameterized type so that templates
will be able to define their own type of config.
Also removing commands related to
creating adapters since there is too much to express is a single
line and writing json on the command line isn't so nice. Once
everything is nailed down, will add back a command that reads the
adapter spec from a file to create it.


albertshau:
http://builds.cask.co/browse/CDAP-DUT1315

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -492,45 +417,22 @@ void registerAdapters() {
...
-
-      if (adapterType != null && sourceType != null && sinkType != null && adapterProgramType != null) {
-        return new AdapterTypeInfo(file, adapterType, Source.Type.valueOf(sourceType.toUpperCase()),
-                                   Sink.Type.valueOf(sinkType.toUpperCase()),
-                                   propertiesFromString(defaultSourceProperties),
-                                   propertiesFromString(defaultSinkProperties),
-                                   propertiesFromString(defaultAdapterProperties),
-                                   ProgramType.valueOf(adapterProgramType.toUpperCase()));
-      }
+      String name = mainAttributes.getValue(ApplicationTemplateManifestAttributes.NAME);
```
gokulavasan: Do we still need Manifest file with these info? If we have the restriction of only template-supported program per template?

gokulavasan: Oh ok nvm. I see the TODO on the top.

gokulavasan: So do we want the configure method in the ApplicationTemplate provide the name and programType?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -240,7 +237,7 @@ public void deleteApp(HttpRequest request, HttpResponder responder,
       Id.Application id = Id.Application.from(namespaceId, appId);
 
       // Deletion of a particular application is not allowed if that application is used by an adapter
-      if (adapterService.getAdapterTypeInfo(appId) != null) {
+      if (adapterService.getApplicationTemplateInfo(appId) != null) {
```
gokulavasan: Overall should we move Adapter Lifecycle endpoints to a separate Http Handler?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
albertshau: it is probably best to restrict the ApplicationTemplate api.  But my starting plan was to configure the app and then examine the ApplicationSpecification to see what programs there are and derive it from that.

gokulavasan: Oh ok. Sounds good.

gokulavasan:
Comment about moving endpoints. Rest LGTM :+1:

PR #1873, opened 2015-04-01
Process quoted input passed as command line arguments to CLIMain

jwang47:
**Note:** This also needs to go into 2.8.1, because otherwise the quickstart guide for 2.8 will not work.

http://builds.cask.co/browse/CDAP-DUT1317

The command below wasn't working because Apache commons-cli was parsing the `-0400` as a command line option. `CLIMain` needs to parse the command line arguments first, and then pass the processed arguments to Apache commons-cli.

```
cdap-cli.sh send stream logEventStream '255.255.255.185 - - [23/Sep/2014:11:45:38 -0400] "GET /cdap.html HTTP/1.0" 401 2969 " " "Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)"'
```

Fix for https://issues.cask.co/browse/CDAP-1975.

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
@@ -32,6 +32,7 @@ import co.cask.common.cli.Command;
 import co.cask.common.cli.CommandSet;
 import co.cask.common.cli.exception.CLIExceptionHandler;
 import co.cask.common.cli.exception.InvalidCommandException;
+import co.cask.common.cli.util.Parser;
 import com.google.common.base.Joiner;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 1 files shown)


awholegunch:
I removed the 2.8 label because it was confusing as this change is just going into develop; another PR is need for the other change.

awholegunch:
I tried porting this to 2.8. When I run it, this command works fine in the cli shell:
```
cdap (http://JJs-MacBook-Pro.local:10000/default)> get stream logEventStream -25m +3m 5
...
Fetched 5 events from stream logEventStream
```
but the exact same command from outside the shell fails:
```
[john@JJs-MacBook-Pro cdap-sdk-2.8.0]$ ./bin/cdap-cli.sh get stream logEventStream -25m +3m 5
Unrecognized option: -25m
```
This suggests this isn't working yet.

jwang47:
Right, this doesn't fix disambiguating `-25m` from `-d` (debug flag). Need to think of a solution.

jwang47:
This is probably why `python` and `bash` do something like

```
python -c 'print "hello"'
```

PR #1874, opened 2015-04-01
Feature/ui preferences

elfenheart:
setting preferences globally and within namespace


ajainarayanan:
@elfenheart The controllers functionality are exactly the same, except for the urls. Can we extract out the common part as a base controller and use the same for both? 

We could have the same controller being used in two views but we resolve the url from the ui-router for each specific state.  For instance,


Routes.js

```
.state('namespacespecific.property.state', {
  url: '..',
  ...
  resolve: {
    urlPath: function() {
      return '/namespaces/specific/url';
    }
  },
  controller: 'GenericController'
}
.state('cdapspecific.property.state', {
  url: '..',
  ...
  resolve: {
    urlPath: function() {
      return '/cdap/specific/url';
    }
  },
  controller: 'GenericController'
}
```

Controller.js

```
angular.module(PKG.name + '.feature.admin.something')
  .controller('GenericController', function(urlPath) {
     //Common functionality.
   });
```

Hope that helps.

ajainarayanan:
:+1: LGTM

PR #1875, opened 2015-04-01
CDAP-1724: Meta metrics for AggregatedMetricsCollectionService

jwang47:
http://builds.cask.co/browse/CDAP-DUT1322

Now we have

* total number of metric values processed
* processing delay time

```
$ curl -X POST http://localhost:10000/v3/metrics/query?context=component.metrics~processor&metric=system.processed.count | python -mjson.tool
{
    "endTime": 0,
    "series": [
        {
            "data": [
                {
                    "time": 0,
                    "value": 271
                }
            ],
            "grouping": {},
            "metricName": "system.metrics.global.processed.count"
        }
    ],
    "startTime": 0
}

$ curl -X POST http://localhost:10000/v3/metrics/query?context=component.metrics~processor&metric=system.processed.delay.avg | python -mjson.tool
{
    "endTime": 0,
    "series": [
        {
            "data": [
                {
                    "time": 0,
                    "value": 966
                }
            ],
            "grouping": {},
            "metricName": "system.metrics.global.processed.delay.ms"
        }
    ],
    "startTime": 0
}
```


cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/KafkaMetricsCollectionService.java
```diff
@@ -87,15 +88,25 @@ protected void publish(Iterator<MetricValue> metrics) throws Exception {
     while (metrics.hasNext()) {
       // Encode each MetricRecord into bytes and make it an individual kafka message in a message set.
       MetricValue value = metrics.next();
-      recordWriter.encode(value, encoder);
-      // partitioning by the context
-      preparer.add(ByteBuffer.wrap(encoderOutputStream.toByteArray()), getPartitionKey(value));
-      encoderOutputStream.reset();
+      metaMetricsComputer.visitMetric(value);
```
gsps1: so you are emitting latency metric for every record ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -86,7 +91,33 @@ protected final void runOneIteration() throws Exception {
...
+          numMetrics++;
+
+          long timestampMs = TimeUnit.MILLISECONDS.convert(metricValue.getTimestamp(), TimeUnit.SECONDS);
+          long nowMs = System.currentTimeMillis();
+          long delayMs = nowMs - timestampMs;
+
+          MetricValue processDelayMetric = new MetricValue(
+            ImmutableMap.of(Constants.Metrics.Tag.NAMESPACE, "system"),
+            "metrics.global.processed.delay.ms", timestamp, delayMs, MetricType.GAUGE);
+          metaMetrics.add(processDelayMetric);
```
gsps1: so if you call visitMetric on every record received at KafkaMetricsCollectionService, and there will be multiple records for the same timestamp, we will be calculating the latency for them and using Gauge will just override them. which seems incorrect. we should either sample or use counter. correct me if i am missing anything.

gsps1: also this line can potentially cause overflow.

cdap-watchdog/src/test/java/co/cask/cdap/metrics/collect/KafkaMetricsCollectionServiceTest.java
```diff
@@ -194,9 +194,8 @@ public void finished() {
       }
     });
 
-    Assert.assertTrue(semaphore.tryAcquire(expected.size(), 5, TimeUnit.SECONDS));
+    Assert.assertTrue(semaphore.tryAcquire(expected.size(), 15, TimeUnit.SECONDS));
```
gsps1: why 10 more seconds ? please add some comment

cdap-watchdog/src/test/java/co/cask/cdap/metrics/collect/KafkaMetricsCollectionServiceTest.java
```diff
@@ -194,9 +194,8 @@ public void finished() {
...
 
-    Assert.assertEquals(expected.size(), metrics.size());
```
gsps1: why ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/MapReduceCounterCollectionService.java
```diff
@@ -46,26 +46,35 @@ protected void startUp() throws Exception {
...
+    while (metaMetricsIterator.hasNext()) {
+      publishMetric(metaMetricsIterator.next());
+    }
+  }
 
-      // The format of the counters:
-      // * counter group name: "cdap.<tag_name>.<tag_value>[.<tag_name>.<tag_value>[...]]
-      // * counter name: metric name
+  private void publishMetric(MetricValue record) {
+    // The format of the counters:
```
gsps1: these are just formatting changes right ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
jwang47: Yeah, we should sample. Why would this cause overflow though?

jwang47: Do you mean we may be storing too many `metaMetrics` in memory? Maybe we can have meta metrics per metric and per metric batch then?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -86,7 +91,33 @@ protected final void runOneIteration() throws Exception {
...
-      publish(metricsItor);
+      publish(metricsItor, new MetaMetricsComputer() {
+        private List<MetricValue> metaMetrics = Lists.newArrayList();
+        private int numMetrics = 0;
+
+        @Override
+        public void visitMetric(MetricValue metricValue) {
+          numMetrics++;
+
+          long timestampMs = TimeUnit.MILLISECONDS.convert(metricValue.getTimestamp(), TimeUnit.SECONDS);
```
chtyim: I think using `TimeUnit.SECONDS.toMillis(metricValue.getTimestamp());` is simpler

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -86,7 +91,33 @@ protected final void runOneIteration() throws Exception {
...
+          long nowMs = System.currentTimeMillis();
```
chtyim: Doing this for every record is not free. You only need to do this if the metric value timestamp changed (in 1 second, there can be a lot of metric values).

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -145,6 +176,8 @@ protected MetricValue computeNext() {
         return endOfData();
       }
     };
+
```
chtyim: Why extra empty lines?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -75,9 +78,11 @@ public AggregatedMetricsEmitter load(EmitterKey key) throws Exception {
    * This method is guaranteed not to get concurrent calls.
    *
    * @param metrics collection of {@link co.cask.cdap.api.metrics.MetricValue} to publish.
+   * @param metaMetricsComputer computes metrics about metrics
    * @throws Exception if there is error raised during publish.
    */
-  protected abstract void publish(Iterator<MetricValue> metrics) throws Exception;
+  protected abstract void publish(Iterator<MetricValue> metrics,
+                                  MetaMetricsComputer metaMetricsComputer) throws Exception;
```
chtyim: Why need to change this? Isn't it better to have the Iterator do it internally while iterating, rather than sub-class need to calls it explicitly?

jwang47: True, I'll have some Iterator that does this.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/MapReduceCounterCollectionService.java
jwang47: yeah

cdap-watchdog/src/test/java/co/cask/cdap/metrics/collect/KafkaMetricsCollectionServiceTest.java
jwang47: Still figuring out this test.. not able to run due to missing some native libs on my machine.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -46,13 +49,17 @@
...
 
   private final LoadingCache<Map<String, String>, MetricsCollector> collectors;
   private final LoadingCache<EmitterKey, AggregatedMetricsEmitter> emitters;
+  private final boolean recordMetaMetrics;
 
   private ScheduledExecutorService executorService;
+  private long timestampLastSampled;
 
-  public AggregatedMetricsCollectionService() {
+  public AggregatedMetricsCollectionService(boolean recordMetaMetrics) {
```
gsps1: may be its better to have  ``recordMetaMetrics``` to be true by default and have a disableMetaMetrics method, so the subclasses can call them if needed. so you don't have to change all the subclass constructor, what do you think ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/KafkaMetricsCollectionService.java
```diff
@@ -87,15 +90,19 @@ protected void publish(Iterator<MetricValue> metrics) throws Exception {
     while (metrics.hasNext()) {
       // Encode each MetricRecord into bytes and make it an individual kafka message in a message set.
       MetricValue value = metrics.next();
-      recordWriter.encode(value, encoder);
-      // partitioning by the context
-      preparer.add(ByteBuffer.wrap(encoderOutputStream.toByteArray()), getPartitionKey(value));
-      encoderOutputStream.reset();
+      publishMetric(preparer, value);
```
gsps1: undo changes?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -83,7 +90,66 @@ public AggregatedMetricsEmitter load(EmitterKey key) throws Exception {
...
     LOG.trace("Start log collection for timestamp {}", timestamp);
-    Iterator<MetricValue> metricsItor = getMetrics(timestamp);
+
+    final Iterator<MetricValue> rawMetricsItor = getMetrics(timestamp);
+    Iterator<MetricValue> metricsItor;
+    if (!recordMetaMetrics) {
+      metricsItor = rawMetricsItor;
+    } else {
+      // wrap the raw metrics iterator with an iterator that will publish meta metrics
+      metricsItor = new AbstractIterator<MetricValue>() {
```
chtyim: Move this into a class/method to make the logic here simple and straightforward.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -83,7 +90,66 @@ public AggregatedMetricsEmitter load(EmitterKey key) throws Exception {
...
+                pendingMetrics.add(processDelayMetric);
+              }
+
+              return raw;
+            } else {
+              // publish pending metric
+              return pendingMetrics.poll();
+            }
+          } else {
+            if (!queuedAggregateMetrics) {
```
gsps1: can we rename the variable to be more meaningful, ```flushTotalProcessedMetrics``` ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
jwang47: Sounds good. I may also split this up into multiple iterator classes.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
jwang47: Sounds good.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
jwang47: I think `queuedAggregateMetrics` is meaningful: If true, that means we queued up the metrics representing the aggregate of the current metrics batch, and thus it's time to send out the queued up aggregate metrics.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionService.java
```diff
@@ -79,11 +84,79 @@ public AggregatedMetricsEmitter load(EmitterKey key) throws Exception {
...
+        protected MetricValue computeNext() {
+          if (rawMetricsItor.hasNext()) {
+
+            if (pendingMetrics.isEmpty()) {
+              // publish raw metric
+              MetricValue raw = rawMetricsItor.next();
+              numMetrics++;
+
+              // sample processing delay - assumes we mostly receive metrics in time order
+              if (raw.getTimestamp() - timestampLastSampled >= SAMPLING_DELAY_SEC) {
```
gsps1: should we also have another option to execute this (every 'n' seconds ?) , let's say we have ```SAMPLING_DELAY_SEC= 10s``` , and we emit 10k records at ```timestamp -1428023632``` and let's say it took 20seconds (hypothetically) to process, assuming we are continuing process at the same rate, by the time we reach ```timestamp -1428023732``` and execute the next iteration, the latency would be roughly close to 3 minutes, we might want to get this latency information more frequently right ?

jwang47:
Discussed this in person with @gsps1 and @chtyim. We decided instead of emitting a metric for processing delay in the iterator, we'd compute running statistics about the processing delay and emit min, max, avg gauges for the processing delays for that batch.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/KafkaMetricsCollectionService.java
```diff
@@ -75,7 +76,7 @@ protected void startUp() throws Exception {
   }
 
   @Override
-  protected void publish(Iterator<MetricValue> metrics) throws Exception {
+  protected void publish(MetricsIterator metrics) throws Exception {
```
chtyim: Why the change in the collector end?

jwang47: Right, this isn't needed. Will change.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/iterator/MetricsIterator.java
```diff
@@ -0,0 +1,76 @@
...
+      return endOfData();
+    }
+  }
+
+  public Iterator<MetricValue> getMetaMetrics(long timestampMs) {
+    long timestamp = TimeUnit.MILLISECONDS.toSeconds(timestampMs);
+
+    MetricValue delayAvg = new MetricValue(
+      ImmutableMap.of(Constants.Metrics.Tag.NAMESPACE, "system"),
+      "metrics.global.processed.delay.avg", timestamp, (long) stats.getAverage(), MetricType.GAUGE);
```
gsps1: you need to calculate the difference right ?
stats.getAverage() - timestamp ? as we are looking for latency ? -- same for other metrics below

cdap-watchdog/src/main/java/co/cask/cdap/metrics/iterator/MetricsIterator.java
```diff
@@ -0,0 +1,76 @@
...
+    if (rawMetricsItor.hasNext()) {
+      MetricValue metric = rawMetricsItor.next();
+      stats.gauge(metric.getValue());
+      return metric;
+    } else {
+      return endOfData();
+    }
+  }
+
+  public Iterator<MetricValue> getMetaMetrics(long timestampMs) {
```
gsps1: nit: don't pass timestamp , as its just current time, you can get it inside.

cdap-watchdog/src/test/java/co/cask/cdap/metrics/collect/AggregatedMetricsCollectionServiceTest.java
```diff
@@ -53,7 +53,7 @@ public void testPublish() throws InterruptedException {
 
     AggregatedMetricsCollectionService service = new AggregatedMetricsCollectionService() {
       @Override
-      protected void publish(Iterator<MetricValue> metrics) {
+      protected void publish(Iterator<MetricValue> metrics) throws Exception {
```
gsps1: unwanted change ? remove?

jwang47:
Tested and working in distributed cluster.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/iterator/MetricsCollectorIterator.java
```diff
@@ -0,0 +1,84 @@
...
+      return Iterators.emptyIterator();
+    }
+
+    long currentTimeMs = System.currentTimeMillis();
+    long currentTimeSec = TimeUnit.MILLISECONDS.toSeconds(currentTimeMs);
+
+    MetricValue delayAvg = new MetricValue(
+      ImmutableMap.of(Constants.Metrics.Tag.NAMESPACE, "system"),
+      "metrics.collector.processed.delay.avg", currentTimeSec,
+      currentTimeMs - TimeUnit.SECONDS.toMillis(processDelayStats.getAverage()), MetricType.GAUGE);
```
gsps1: hey, @abaranau mentioned this has to be counter during our meeting. if we receive multiple iterators within a second , we would override.

gsps1:
just one comment @jwang47 LGTM otherwise.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/iterator/MetricsCollectorIterator.java
jwang47: Did you mean histogram? I thought we were going to implement histogram later?

jwang47:
@gsps1 can you take another look?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/iterator/MetricsCollectorIterator.java
gsps1: no i don't mean histogram, if we process multiple batches within a second, then we will only have latency metrics for the last batch as we override the other batch data. we should use counter right?

sreevatsanraman:
Made the change @gsps1 suggested. Running a build here: 
http://builds.cask.co/browse/CDAP-DUT1322-18

sreevatsanraman:
One more time. http://builds.cask.co/browse/CDAP-DUT1322-19

cdap-watchdog/src/main/java/co/cask/cdap/metrics/iterator/MetricsCollectorIterator.java
```diff
@@ -0,0 +1,85 @@
...
+
+    long currentTimeMs = System.currentTimeMillis();
+    long currentTimeSec = TimeUnit.MILLISECONDS.toSeconds(currentTimeMs);
+
+    Map<String, String> tags = ImmutableMap.of(Constants.Metrics.Tag.NAMESPACE, "system",
+                                               Constants.Metrics.Tag.COMPONENT, "metrics.processor");
+
+    MetricValue delaySum = new MetricValue(
+      tags, "processed.delay.sum", currentTimeSec,
+      currentTimeMs - TimeUnit.SECONDS.toMillis(processDelayStats.getSum()), MetricType.COUNTER);
```
gsps1: when we get the sum, its the sum of timestamp, to calculate the sum of latency, we need to do
``(currentTimems * processDelayStats.getCount) - TimeUnit.SECONDS.toMillis(processDelayStats.getSum())``

or pass the currentTimems to the getSum and get the sum of latency from there.

sreevatsanraman: Fixed.

gsps1:
LGTM :+1:

sreevatsanraman:
will merge once the build passes: http://builds.cask.co/browse/CDAP-DUT1322-21

PR #1879, opened 2015-04-02
Fix/ui explorer view

elfenheart:
- Fix default SQL query to get the current streamId or datasetId
- Change styling on view Queries to give more room to results
- Change body on Stream explore to limit to 80 characters
- Cursor change on View Queries table list
- Add selected queries styling to indicate which query is currently selected

explore


cdap-ui/app/features/datasets/datasets.less
```diff
@@ -13,6 +13,20 @@ body.state-datasets-detail {
...
 
     .panel-body {
+
+      .queries {
+        font-size: 14px;
+
+        tr {
+          cursor: pointer;
+
+          &.current {
```
ajainarayanan: Do we already have a style for hover?

cdap-ui/app/features/datasets/templates/tabs/explore.html
```diff
@@ -95,18 +95,12 @@ <h4 class="panel-title" bs-collapse-toggle>
         </div>
 
         <!-- Right Panel -->
-        <div class="col-xs-6 well">
-          <div ng-show="responses.results.length">
-            <p><strong>{{ responses.request.status }}</strong></p>
+        <div class="col-xs-8 well">
+          <div ng-show="responses.request.status === 'FINISHED'">
```
ajainarayanan: Shouldn't the ```well``` css class be in this div?

cdap-ui/app/features/datasets/templates/tabs/explore.html
```diff
@@ -74,7 +74,7 @@ <h4 class="panel-title" bs-collapse-toggle>
       <div class="panel-body">
 
         <!-- Left Panel -->
-        <div class="col-xs-6">
+        <div class="col-xs-4 queries">
           <table class="table table-hover">
```
ajainarayanan: Same here. Shouldn't the ```queries``` class be in the table?

cdap-ui/app/features/streams/streams.less
```diff
@@ -49,6 +62,7 @@ body.state-streams-detail {
       }
 
       .results {
+        margin-top: 25px;
         div {
```
ajainarayanan: I feel a css selector for all the divs under results css class is too generic. We should probably revisit this soon. This is the same for both datasets and streams. We should also be re-using them. Not sure why we aren't.

ajainarayanan:
We are ideally having the same template and less file for both the datasets and streams explorer view. @elfenheart Any specific reason why we are replicating the code in multiple places? Was the view supposed to be different in datasets and streams?

elfenheart:
@ajainarayanan yes, in that case a directive will make more sense. I will make the modification.

ajainarayanan:
:+1: LGTM

PR #1880, opened 2015-04-02
Feature/r2.8 fixes to RESTful API "Metrics" and Developers' Manual "Quick Start"

awholegunch:
Both of these had errors in them.

Staged at: 

- [Quick Start](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_fix_quick_start/en/developers-manual/getting-started/quick-start.html)
- [Metrics](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_fix_quick_start/en/reference-manual/http-restful-api/http-restful-api-v3/metrics.html)

A change has been made to develop that fixes this issue in develop without changing the CLI API. This PR then fixes the documentation (to remove its errors.) The CLI fix is needed in 2.8 so that the Quick Start and CLI will work there.

### DIFF
--- a/cdap-docs/_common/common_conf.py
+++ b/cdap-docs/_common/common_conf.py
@@ -26,6 +26,8 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+wise_version = "0.3.0"
+
 import sys
 import os
 import os.path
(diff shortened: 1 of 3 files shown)


sreevatsanraman:
lgtm

PR #1881, opened 2015-04-02
Using time based UUIDs for run ids

poornachandra:


cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
```diff
@@ -0,0 +1,146 @@
...
+  // Multiplier to convert millisecond into 100ns
+  private static final long HUNDRED_NANO_MULTIPLIER = 10000;
+
+  private static final AtomicInteger COUNTER = new AtomicInteger();
+
+  /**
+   * @return UUID based on current time. If called repeatedly within the same millisecond, this is
+   * guaranteed to generate at least 10000 unique UUIDs for the millisecond.
+   */
+  public static RunId generate() {
```
chtyim: You might want the timestamp being passed in so that you can sync it with the one recorded in MDS.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
```diff
@@ -0,0 +1,146 @@
...
+      }
+
+      TimeBasedRunId that = (TimeBasedRunId) o;
+
+      return Objects.equal(this.id, that.id);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(id);
```
chtyim: Just returns the id.hashCode().

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
```diff
@@ -0,0 +1,146 @@
...
+ */
+public final class RunIds {
+  private static final Random RANDOM = new Random();
+
+  // Number of 100ns since 15 October 1582 00:00:000000000
+  private static final long NUM_100NS_INTERVALS_SINCE_UUID_EPOCH = 0x01b21dd213814000L;
+  // Multiplier to convert millisecond into 100ns
+  private static final long HUNDRED_NANO_MULTIPLIER = 10000;
+
+  private static final AtomicInteger COUNTER = new AtomicInteger();
```
chtyim: Better use `AtomicLong` to avoid potential overflow.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Generates an unique ID for a running program using type 1 and variant 2 time-based {@link UUID}.
+ */
+public final class RunIds {
+  private static final Random RANDOM = new Random();
+
+  // Number of 100ns since 15 October 1582 00:00:000000000
```
anwar6953: This comment seems to be inaccurate.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
```diff
@@ -0,0 +1,146 @@
...
+    return new TimeBasedRunId(UUID.fromString(id));
+  }
+
+  /**
+   * @return time in milliseconds from the UUID if it is a time-based UUID, -1 otherwise.
+   */
+  public static long getTimeMillis(RunId runId) {
+    UUID uuid = UUID.fromString(runId.getId());
+    if (uuid.version() == 1 && uuid.variant() == 2) {
+      return (uuid.timestamp() - NUM_100NS_INTERVALS_SINCE_UUID_EPOCH) / HUNDRED_NANO_MULTIPLIER;
```
anwar6953: It doesn't seem like a multiplier if you're dividing by it?

cdap-app-fabric/src/test/java/co/cask/cdap/app/runtime/RunIdsTest.java
```diff
@@ -0,0 +1,48 @@
...
+    Assert.assertNotEquals(uuid1.toString(), uuid2.toString());
+
+    // Generate UUID for a different time
+    long newTime = time + 1;
+    UUID uuid3 = RunIds.generateUUIDForTime(newTime);
+    Assert.assertEquals(newTime, RunIds.getTimeMillis(RunIds.fromString(uuid3.toString())));
+    Assert.assertNotEquals(uuid1.toString(), uuid3.toString());
+    Assert.assertNotEquals(uuid2.toString(), uuid3.toString());
+
+    // Time from a random UUID should be -1
```
anwar6953: Is this guaranteed?
(Whats the likelihood)?

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
poornachandra: Or should I get the time from UUID and store it as the start time in MDS? This will eliminate the need to pass the start time into `ProgramRuntimeService`. What do you say?

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
poornachandra: Good point.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
poornachandra: Agreed.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
poornachandra: Yes, good point. Will fix it.

cdap-app-fabric/src/test/java/co/cask/cdap/app/runtime/RunIdsTest.java
poornachandra: It is guaranteed 100% since we check to make sure it is a time-based UUID before returning the timestamp.

anwar6953: I mean since its 'random', there's no chance it can conflict with and be interpreted as a time-based UUI?

poornachandra: Every UUID has type information embedded in it, hence not all bits of the UUID are random in this case. More information http://docs.oracle.com/javase/7/docs/api/java/util/UUID.html

anwar6953: ok, sounds good.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
chtyim: Seems like we only update the MDS after the `ProgramController` is return. What you suggested would work :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
```diff
@@ -30,7 +31,6 @@
 import com.google.inject.name.Named;
 import org.apache.twill.api.RunId;
 import org.apache.twill.api.ServiceAnnouncer;
-import org.apache.twill.internal.RunIds;
```
sreevatsanraman: Should this timebased run-id eventually be in twill?

sreevatsanraman:
one question so that we open corresponding jira in twill. LGTM :+1:

poornachandra:
@sreevatsanraman I don't think Twill has a use case for it. @chtyim what do you say?

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
poornachandra: I will make this change in the next PR where we'll need to extract time from runId to fetch run record.

poornachandra:
@anwar6953 @chtyim Addressed review comments. Please take a look.

chtyim:
Twill doesn't have the requirement of generating time-based RunId. In fact, the concept of RunId in Twill is internal, it's not part of public API.

chtyim:
LGTM

poornachandra:
Build http://builds.cask.co/browse/CDAP-DUT1327-5

PR #1882, opened 2015-04-02
Feature/etl batch driver

gokulavasan:
* ETL Batch Template + ETL MapReduce
* Simple test case with KVTableSource/Sink and IdentityTransform

Bamboo : http://builds.cask.co/browse/CDAP-DT4-4

JIRA : https://issues.cask.co/browse/CDAP-1769


gokulavasan:
@albertshau @chtyim Addressed your comments. Please review. Had to close the old PR #1848 since it was opened against etl-template-api branch.

cdap-app-templates/cdap-etl/cdap-etl-core/pom.xml
```diff
@@ -30,12 +30,78 @@
   <name>CDAP ETL App Template Core</name>
   <packaging>jar</packaging>
 
+  <properties>
+    <app.main.class>co.cask.cdap.templates.etl.batch.ETLBatchTemplate</app.main.class>
```
albertshau: doesn't this go in the parent pom?

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+    initTable(Lists.<Class>newArrayList(KVTableSource.class, KVTableSink.class, IdentityTransform.class));
+  }
+
+  private void initTable(List<Class> classList) throws Exception {
+    for (Class klass : classList) {
+      DefaultStageConfigurer configurer = new DefaultStageConfigurer(klass);
+      if (BatchSource.class.isAssignableFrom(klass)) {
+        BatchSource source = (BatchSource) klass.newInstance();
+        source.configure(configurer);
+        nameToClass.put("source", configurer.createSpecification().getName(),
```
albertshau: these strings should be an enum

albertshau: actually, it's probably simpler just to have 3 Maps

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+    //plugins management is completed.
+    initTable(Lists.<Class>newArrayList(KVTableSource.class, KVTableSink.class, IdentityTransform.class));
+  }
+
+  private void initTable(List<Class> classList) throws Exception {
+    for (Class klass : classList) {
+      DefaultStageConfigurer configurer = new DefaultStageConfigurer(klass);
+      if (BatchSource.class.isAssignableFrom(klass)) {
+        BatchSource source = (BatchSource) klass.newInstance();
+        source.configure(configurer);
```
albertshau: shouldn't configure here, as it's called from the constructor.  For example, suppose the source creates a stream in it's configure.

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+  public ETLBatchTemplate() throws Exception {
+    nameToClass = HashBasedTable.create();
+    //TODO: Add classes from Lib here to be available for use in the ETL Adapter. Remove this when
+    //plugins management is completed.
+    initTable(Lists.<Class>newArrayList(KVTableSource.class, KVTableSink.class, IdentityTransform.class));
+  }
+
+  private void initTable(List<Class> classList) throws Exception {
+    for (Class klass : classList) {
+      DefaultStageConfigurer configurer = new DefaultStageConfigurer(klass);
```
albertshau: shouldn't have to create a Configurer anywhere in the application code.  One will be passed in to configureAdapter by the CDAP platform

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+    configurer.addRuntimeArgument(Constants.ADAPTER_NAME, adapterName);
+    configurer.setSchedule(new TimeSchedule(String.format("etl.batch.adapter.%s.schedule", adapterName),
+                                            String.format("Schedule for %s Adapter", adapterName),
+                                            cronEntry));
+  }
+
+  private void configureSource(ETLStage source, AdapterConfigurer configurer) throws Exception {
+    String sourceName = source.getName();
+    String className = nameToClass.get("source", sourceName);
+    BatchSource batchSource = (BatchSource) Class.forName(className).newInstance();
```
albertshau: this is where you should be configuring the source

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+    DefaultStageConfigurer stageConfigurer = new DefaultStageConfigurer(batchSource.getClass());
+    StageSpecification specification = stageConfigurer.createSpecification();
```
albertshau: not sure I understand what is going on here.  Isn't this where we want to be calling configure on the source?  And shouldn't it be passing the AdapterConfigurer to them?

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+  private void configureSink(ETLStage sink, AdapterConfigurer configurer) throws Exception {
+    String sinkName = sink.getName();
+    String className = nameToClass.get("sink", sinkName);
+    BatchSink batchSink = (BatchSink) Class.forName(className).newInstance();
+    DefaultStageConfigurer stageConfigurer = new DefaultStageConfigurer(batchSink.getClass());
+    StageSpecification specification = stageConfigurer.createSpecification();
+    configurer.addRuntimeArgument(Constants.Sink.SPECIFICATION, GSON.toJson(specification));
+  }
+
+  private void configureTransform(List<ETLStage> transforms, AdapterConfigurer configurer) throws Exception {
```
albertshau: nit: should be named configureTransforms

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
```diff
@@ -0,0 +1,201 @@
...
+  public void configure() {
+    setName("ETLMapReduce");
+    setDescription("MapReduce driver for Batch ETL Adapters");
+  }
+
+  @Override
+  public void beforeSubmit(MapReduceContext context) throws Exception {
+    Job job = context.getHadoopJob();
+
+    Preconditions.checkArgument(context.getRuntimeArguments().containsKey(Constants.ADAPTER_NAME));
```
albertshau: nit: ```context.getRuntimeArguments()``` is called a bunch of times in this method. Would be less verbose to get it once and store it in a variable.

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
```diff
@@ -0,0 +1,201 @@
...
+    LOG.info("Sink Class : {}", sink.getClass().getName());
+    LOG.info("Specifications of Sink : {}", sinkSpec);
+    sink.prepareJob(sinkContext);
+  }
+
+  private void prepareTransforms(MapReduceContext context, List<ETLStage> transformStages) {
+    Job job = context.getHadoopJob();
+
+    //Set list of Transform specifications to be used in the Mapper
+    job.getConfiguration().set(Constants.Transform.SPECIFICATION,
```
albertshau: what is the difference between specification and configuration?  Is configuration the thing that gets passed to the adapter? How come we need both?

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/TransformExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+public final class TransformExecutor {
+  private final List<Transform> transformList;
+  private DefaultEmitter previousEmitter;
+  private DefaultEmitter currentEmitter;
+
+  public TransformExecutor(List<Transform> transforms) {
+    this.transformList = transforms;
+    this.previousEmitter = new DefaultEmitter();
+    this.currentEmitter = new DefaultEmitter();
+
```
albertshau: extra newline

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/TransformExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  public void runOneIteration(DefaultEmitter data) throws Exception {
+    //Copy Data into previousEmitter and reset 'data'.
```
albertshau: I think we can simplify things if the method takes in a key and value and returns an ```Iterable<Map.Entry>```.  Then you can just return previousEmitter at the end, and you can reset before each call to runOneIteration.  That way, you also don't have to pass in an emitter and avoid the copies that happen at the start and end.

cdap-app-templates/cdap-etl/cdap-etl-core/src/test/java/co/cask/cdap/templates/etl/batch/ETLMapReduceTest.java
```diff
@@ -0,0 +1,86 @@
...
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.dataset.lib.KeyValueTable;
+import co.cask.cdap.api.templates.ApplicationTemplate;
+import co.cask.cdap.templates.etl.batch.config.ETLBatchConfig;
+import co.cask.cdap.templates.etl.common.config.ETLStage;
+import co.cask.cdap.test.ApplicationManager;
+import co.cask.cdap.test.DataSetManager;
+import co.cask.cdap.test.MapReduceManager;
+import co.cask.cdap.test.TestBase;
+import com.clearspring.analytics.util.Lists;
```
albertshau: use guava lists

cdap-app-templates/cdap-etl/cdap-etl-core/src/test/java/co/cask/cdap/templates/etl/batch/DefaultAdapterConfigurer.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+
+import java.util.Map;
+
+/**
+ * Default implementation of {@link AdapterConfigurer} for ETL Batch Tests.
+ */
+//TODO: Remove/Move this class else where until we figure out how to write tests without AdapterConfigurer dependency
+public class DefaultAdapterConfigurer implements AdapterConfigurer {
```
albertshau: if this is just for tests, it's better to call it MockAdapterConfigurer or something like that.

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
albertshau: ok spec is generated by the stage and config comes from the outside. Not clear if we really need spec, but keeping it for now and will remove at the end if it is really not used anywhere

gokulavasan:
@albertshau Addressed all your comments. Please take a look again. Thanks!

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/common/TransformExecutor.java
```diff
@@ -0,0 +1,53 @@
...
+  private DefaultEmitter currentEmitter;
+
+  public TransformExecutor(List<Transform> transforms) {
+    this.transformList = transforms;
+    this.previousEmitter = new DefaultEmitter();
+    this.currentEmitter = new DefaultEmitter();
+  }
+
+  public Iterable<Map.Entry> runOneIteration(Object key, Object value) throws Exception {
+    //Copy Data into previousEmitter and reset 'data'.
```
albertshau: the comment no longer applies

cdap-app-templates/cdap-etl/pom.xml
```diff
@@ -48,6 +49,117 @@
...
+            <groupId>org.mortbay.jetty</groupId>
+            <artifactId>jetty</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>org.mortbay.jetty</groupId>
+            <artifactId>jetty-util</artifactId>
+          </exclusion>
+          <exclusion>
+            <artifactId>jasper-compiler</artifactId>
+            <groupId>tomcat</groupId>
```
albertshau: what is using tomcat?

albertshau: o ignore my comment, this is an exclusion

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/common/Constants.java
```diff
@@ -0,0 +1,48 @@
...
+   */
+  public static final class Sink {
+    public static final String SPECIFICATION = "templates.etl.adapter.sink.specification";
+  }
+
+  /**
+   * Constants related to Transform.
+   */
+  public static final class Transform {
+    public static final String TRANSFORM_CONFIGS = "templates.etl.adapter.transform.configs";
```
albertshau: nit: let's rename to CONFIGS since its already in Transform

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/common/Constants.java
```diff
@@ -0,0 +1,48 @@
...
+    public static final String SPECIFICATION = "templates.etl.adapter.transform.specification";
```
albertshau: nite: lets rename to SPECIFICATIONS so its clear that its a list of specs

albertshau:

gokulavasan:
Thanks @albertshau. Addressed your comments. Will merge once build passes.

PR #1884, opened 2015-04-02
Use read/write lock instead of exclusive lock in DS framework

chtyim:


anew:
LGTM

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1329-1

PR #1885, opened 2015-04-02
Treat options and command as two separate sections of args from main()

jwang47:
http://builds.cask.co/browse/CDAP-DUT1330

The command below wasn't working because Apache commons-cli was parsing the `-0400` as a command line option. `CLIMain` needs to use the convention of having command line options first in the main() args, then the command to execute after.

```
cdap-cli.sh send stream logEventStream '255.255.255.185 - - [23/Sep/2014:11:45:38 -0400] "GET /cdap.html HTTP/1.0" 401 2969 " " "Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1)"'
```


awholegunch:
I have built and test this in develop, and so far so good. It's fixed the problem with the Quick Start not working. LGTM.

PR #1887, opened 2015-04-02
Implementation to update the history of MapReduce program running under Workflow

sagarkapare:
Moved the logic to record states for MapReduce programs from ProgramLifeCycleHttpHandler to the MapReduceRunTimeService.

Things to do yet:
1. Unit test cases
2. Some more testing on the distributed

Opening PR to get early feedback.


sagarkapare:
Added unit test cases. @chtyim @bdmogal Can you please take a look? Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1171,7 +1171,13 @@ protected ProgramStatus getProgramStatus(Id.Program id, ProgramType type) {
             // program doesn't exist
             return new ProgramStatus(id.getApplicationId(), id.getId(), HttpResponseStatus.NOT_FOUND.toString());
           } else {
```
chtyim: No need to have else after return inside `if`.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1171,7 +1171,13 @@ protected ProgramStatus getProgramStatus(Id.Program id, ProgramType type) {
...
-            // program exists and not running. so return stopped.
+            // program exists
+            if (type == ProgramType.MAPREDUCE) {
+              // check if MapReduce program is running as a part of the Workflow
+              if (store.getRuns(id, ProgramRunStatus.RUNNING, 0, Long.MAX_VALUE, 1).size() > 0) {
```
chtyim: Combine these two if into one using `&&`.

chtyim: Also, for checking collection size > 0, always use `!collection.isEmpty()`

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1340,53 +1346,59 @@ private AppFabricServiceStatus start(final Id.Program id, ProgramType type,
...
 
-      controller.addListener(new AbstractListener() {
-
-        @Override
-        public void init(ProgramController.State state, @Nullable Throwable cause) {
-          store.setStart(id, runId, TimeUnit.SECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS));
-          if (state == ProgramController.State.COMPLETED) {
-            completed();
+      if (type == ProgramType.MAPREDUCE) {
+        // For MapReduce programs recording of the states in the MDS is done by MapReduceRuntimeService
```
chtyim: Can you add a TODO and open a JIRA for making all ProgramRunners do the same?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1340,53 +1346,59 @@ private AppFabricServiceStatus start(final Id.Program id, ProgramType type,
...
+        // We only need listener here to record explicit stopping of the MapReduce program
```
chtyim: Why? If the program is stopped by user explicitly, there will be a stop signal send to the program runner so that it can interrupt/terminate whatever is running. The means the program runner does know that user request a stop, hence can record the state in MDS.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -120,12 +122,13 @@
...
   private Runnable cleanupTask;
   private volatile boolean stopRequested;
+  private final Store store;
 
   MapReduceRuntimeService(CConfiguration cConf, Configuration hConf,
                           MapReduce mapReduce, MapReduceSpecification specification,
                           DynamicMapReduceContext context,
                           Location programJarLocation, LocationFactory locationFactory,
-                          StreamAdmin streamAdmin, TransactionSystemClient txClient) {
+                          StreamAdmin streamAdmin, TransactionSystemClient txClient, Store store) {
```
chtyim: You shouldn't need to pass the store to this class. The `MapReduceRuntimeService` is a guava Service, hence you can attach a service listener in the program runner, before starting this service and do the writing to store in the listener.

chtyim:
Done review. Please address comments.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Merged into single `if`

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Added TODO

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
sagarkapare: Moved the logic to the `MapReduceProgramRunner`

sagarkapare:

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1170,10 +1170,13 @@ protected ProgramStatus getProgramStatus(Id.Program id, ProgramType type) {
...
-            // program exists and not running. so return stopped.
-            return new ProgramStatus(id.getApplicationId(), id.getId(), "STOPPED");
           }
+          if (type == ProgramType.MAPREDUCE && !store.getRuns(id, ProgramRunStatus.RUNNING, 0,
+                                                              Long.MAX_VALUE, 1).isEmpty()) {
+            // MapReduce program exists and running as a part of Workflow
+            return new ProgramStatus(id.getApplicationId(), id.getId(), "RUNNING");
+          }
+          return new ProgramStatus(id.getApplicationId(), id.getId(), "STOPPED");
         } else {
```
chtyim: You also don't need the else here since the if always return.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/inmemory/InMemoryMapReduceContextBuilder.java
```diff
@@ -86,7 +88,13 @@ private Injector createInMemoryModules() {
       new LoggingModules().getInMemoryModules(),
       new StreamAdminModules().getInMemoryModules(),
       new ExploreClientModule(),
-      new NotificationFeedServiceRuntimeModule().getInMemoryModules()
+      new NotificationFeedServiceRuntimeModule().getInMemoryModules(),
+      new AbstractModule() {
+        @Override
+        protected void configure() {
+          bind(Store.class).to(DefaultStore.class);
```
chtyim: Do we need it here? I thought only the program runner uses Store and program runner is not created by this injector, but rather by the one in AppFabric for in-memory (which already has Store binded) and in `AbstractProgramTwillRunnable` for distributed mode.

chtyim:
Mostly look two, just two comments left.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/inmemory/InMemoryMapReduceContextBuilder.java
sagarkapare: We need it here as well, because ProgramRunner is also created by this injector on  line 84 above. right?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Fixed in the last commit.

sagarkapare:
Thank you @chtyim for the review! Updated the code.
Build is running here - http://builds.cask.co/browse/CDAP-DUT1325-9

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/inmemory/InMemoryMapReduceContextBuilder.java
chtyim: I remember we never create program runner with this injector. We shouldn't have that module used. Can you try to remove that module and see what bindings are missing if there is any?

chtyim: I tried commenting out the ProgramRunner module above and it works in unit-test. I believe it will work in Standalone as well.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/inmemory/InMemoryMapReduceContextBuilder.java
```diff
@@ -107,7 +115,13 @@ private Injector createPersistentModules() {
       new LoggingModules().getStandaloneModules(),
       new ExploreClientModule(),
       new StreamAdminModules().getStandaloneModules(),
-      new NotificationFeedServiceRuntimeModule().getStandaloneModules()
+      new NotificationFeedServiceRuntimeModule().getStandaloneModules(),
+      new AbstractModule() {
+        @Override
+        protected void configure() {
+          bind(Store.class).to(DefaultStore.class);
```
chtyim: As mentioned above, we shouldn't have this.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/inmemory/InMemoryMapReduceContextBuilder.java
sagarkapare: I verified by removing it and standalone is also seems to be working fine. So you are right we do not need to use ProgramRunner module in this injector. Removed from `InMemoryMapReduceContextBuilder` and `InMemorySparkContextBuilder`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/inmemory/InMemoryMapReduceContextBuilder.java
sagarkapare: Removed.

sagarkapare:
Triggered new build is running here - http://builds.cask.co/browse/CDAP-DUT1325-10

chtyim:
LGTM

sagarkapare:
Build passed. Merging!

PR #1896, opened 2015-04-03
Moving Cube APIs into cdap-api

abaranau:
Pre-requisite for Cube dataset (https://issues.cask.co/browse/CDAP-1520). This is in separate PR from it for easier review. This PR only moves classes and adds couple of javadocs. No code/logic changes.


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/cube/CubeDeleteQuery.java
```diff
@@ -14,27 +14,37 @@
...
+  /**
+   * Creates instance of {@link CubeDeleteQuery} that defines selection of data to delete from {@link Cube}.
+   * @param startTs start time of the data selection, in seconds since epoch
+   * @param endTs end time of the data selection, in seconds since epoch
+   * @param resolution resolution of the aggregations to delete from
+   * @param measureName name of the measure to delete, {@code null} means delete all
+   * @param sliceByTagValues tag name, tag value pairs that define the data selection
+   */
+  public CubeDeleteQuery(long startTs, long endTs, int resolution,
+                         @Nullable String measureName,
```
chtyim: Better to have `@Nullable` argument as the last one. Also, it's handy to have an overloaded constructor to represent delete all so that user doesn't need to pass in `null`.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/cube/CubeDeleteQuery.java
```diff
@@ -68,8 +78,9 @@ public String toString() {
     return Objects.toStringHelper(this)
       .add("startTs", startTs)
       .add("endTs", endTs)
+      .add("resolution", resolution)
       .add("measureName", measureName)
-      .add("sliceByTags", Joiner.on(",").withKeyValueSeparator(":").useForNull("null").join(sliceByTagValues))
+      .add("sliceByTags", Joiner.on(",").withKeyValueSeparator(":").join(sliceByTagValues))
```
chtyim: No guava class in cdap-api module (we are trying to get rid of the dependency). Simple add the map and relies on the Map.toString.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/cube/CubeDeleteQuery.java
```diff
@@ -14,27 +14,37 @@
  * the License.
  */
 
-package co.cask.cdap.metrics.store.cube;
+package co.cask.cdap.api.dataset.lib.cube;
 
+import co.cask.cdap.api.annotation.Beta;
 import com.google.common.base.Joiner;
```
chtyim: No guava dependency for cdap-api module.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/cube/CubeExploreQuery.java
```diff
@@ -34,6 +33,13 @@
...
   private final List<TagValue> tagValues;
 
+  /**
+   * Creates instance of {@link CubeExploreQuery} that defines selection of data of {@link Cube} to explore in.
+   * @param startTs start time of the data selection, in seconds since epoch
+   * @param endTs end time of the data selection, in seconds since epoch
+   * @param resolution resolution of the aggregations explore
+   * @param tagValues tag name, tag value pairs that define the data selection
+   */
   public CubeExploreQuery(long startTs, long endTs, int resolution, int limit, List<TagValue> tagValues) {
```
chtyim: Is ordering important for `tagValues`? If not, use more generic form like `Collection` or `Iterable`.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/cube/CubeExploreQuery.java
```diff
@@ -34,6 +33,13 @@
   private final int limit;
   private final List<TagValue> tagValues;
 
+  /**
+   * Creates instance of {@link CubeExploreQuery} that defines selection of data of {@link Cube} to explore in.
+   * @param startTs start time of the data selection, in seconds since epoch
+   * @param endTs end time of the data selection, in seconds since epoch
```
chtyim: Should state clearly whether it's inclusive or exclusive.

chtyim:
Just couple comments, mainly about guava usage.

awholegunch:
Will there be documentation about this? I notice this is being marked as "Beta"...

abaranau:
@awholegunch when we finish up CubeDataset...

abaranau:
one more (there were green already before merge with dev) http://builds.cask.co/browse/CDAP-DUT1336-7

PR #1898, opened 2015-04-04
Treat options and command as two separate sections of args from main()

jwang47:
http://builds.cask.co/browse/CDAP-RBT214

From https://github.com/caskdata/cdap/pull/1885


albertshau:
:+1:

awholegunch:
LGTM.

PR #1900, opened 2015-04-04
Fix/ui dashboard live file move

ajainarayanan:
- Fixes dashboard tabs. Now we can,
   - View Standard & User dashboards together
   - Has minor fixes (adds new dashboard to the end + onhover tooltip) to the dashboard tabs
- Modifies the tabs styles universally.


cdap-ui/app/features/dashboard/routes.js
```diff
@@ -1,38 +1,77 @@
 angular.module(PKG.name+'.feature.dashboard')
   .config(function ($stateProvider, $urlRouterProvider, MYAUTH_ROLE) {
 
+    var path = '/assets/features/operation28/';
```
elfenheart: Are we still using operation28 now that you merge it?

ajainarayanan: Yes. The the controllers and templates are still under operations28 directory. We are just re-using stuff from that directory instead of moving them to dashboard.

astewes:
LGTM

elfenheart:
:+1: LGTM

PR #1902, opened 2015-04-04
CDAP-1768 adding separate pipelines for app template deployment

albertshau:
and adapter deployment since they are different than application
deployment.

Also removing overly complex generic for AdapterSpecification

### DIFF
--- a/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/AdapterArgs.java
+++ b/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/AdapterArgs.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.conversion.app;
+
+import co.cask.cdap.api.Resources;
+import co.cask.cdap.api.data.format.FormatSpecification;
+import co.cask.cdap.api.data.schema.Schema;
+import com.google.gson.JsonObject;
+
+import java.io.IOException;
+import java.util.Collections;
+
+/**
+ */
+public class AdapterArgs {
+  private final String format;
+  private final JsonObject schema;
+  private final String headers;
+  private final String source;
+  private final String sink;
+  private final String frequency;
+
+  public AdapterArgs(String format, JsonObject schema, String headers, String source, String sink, String frequency) {
+    this.format = format;
+    this.schema = schema;
+    this.headers = headers;
+    this.source = source;
+    this.sink = sink;
+    this.frequency = frequency;
+  }
+
+  public ConversionConfig getConfig() throws IOException {
+    return new ConversionConfig(
+      new Resources(),
+      new FormatSpecification(format, Schema.parseJson(schema.toString()), Collections.<String, String>emptyMap()),
+      source, sink, headers, frequency);
+  }
+}
--- a/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionAdapter.java
+++ b/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionAdapter.java
@@ -29,6 +29,7 @@ import org.apache.avro.mapreduce.AvroKeyOutputFormat;
  * Application that converts a stream into a partitioned file set.
  */
 public class StreamConversionAdapter extends ApplicationTemplate<AdapterArgs> {
+  static final String CONFIG_KEY = "adapter.args";
 
   @Override
   public void configure() {
@@ -59,7 +60,7 @@ public class StreamConversionAdapter extends ApplicationTemplate<AdapterArgs> {
   @Override
   public void configureAdapter(String adapterName, AdapterArgs args,
                                AdapterConfigurer configurer) throws Exception {
-    configurer.addRuntimeArgument("adapter.args", new Gson().toJson(args));
+    configurer.addRuntimeArgument(CONFIG_KEY, new Gson().toJson(args));
     configurer.setSchedule(Schedules.createTimeSchedule("test", "adapter schedule", "* * * * *"));
   }
 
--- a/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionAdapter.java
+++ b/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionAdapter.java
@@ -16,17 +16,51 @@
 
 package co.cask.cdap.conversion.app;
 
-import co.cask.cdap.api.app.AbstractApplication;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.dataset.lib.FileSetProperties;
+import co.cask.cdap.api.schedule.Schedules;
+import co.cask.cdap.api.templates.AdapterConfigurer;
+import co.cask.cdap.api.templates.ApplicationTemplate;
+import com.google.gson.Gson;
+import org.apache.avro.mapreduce.AvroKeyInputFormat;
+import org.apache.avro.mapreduce.AvroKeyOutputFormat;
 
 /**
  * Application that converts a stream into a partitioned file set.
  */
-public class StreamConversionAdapter extends AbstractApplication {
+public class StreamConversionAdapter extends ApplicationTemplate<AdapterArgs> {
 
   @Override
   public void configure() {
     setDescription("Periodically reads stream events and writes them to a time partitioned fileset");
     addMapReduce(new StreamConversionMapReduce());
     addWorkflow(new StreamConversionWorkflow());
+
+    Schema eventSchema = Schema.recordOf(
+      "streamEvent",
+      Schema.Field.of("ts", Schema.of(Schema.Type.LONG)),
+      Schema.Field.of("header1", Schema.unionOf(Schema.of(Schema.Type.NULL), Schema.of(Schema.Type.STRING))),
+      Schema.Field.of("header2", Schema.unionOf(Schema.of(Schema.Type.NULL), Schema.of(Schema.Type.STRING))),
+      Schema.Field.of("ticker", Schema.of(Schema.Type.STRING)),
+      Schema.Field.of("num", Schema.of(Schema.Type.INT)),
+      Schema.Field.of("price", Schema.of(Schema.Type.DOUBLE)));
+    createDataset("converted", "timePartitionedFileSet", FileSetProperties.builder()
+      .setBasePath("converted")
+      .setInputFormat(AvroKeyInputFormat.class)
+      .setOutputFormat(AvroKeyOutputFormat.class)
+      .setEnableExploreOnCreate(true)
+      .setSerDe("org.apache.hadoop.hive.serde2.avro.AvroSerDe")
+      .setExploreInputFormat("org.apache.hadoop.hive.ql.io.avro.AvroContainerInputFormat")
+      .setExploreOutputFormat("org.apache.hadoop.hive.ql.io.avro.AvroContainerOutputFormat")
+      .setTableProperty("avro.schema.literal", eventSchema.toString())
+      .build());
   }
+
+  @Override
+  public void configureAdapter(String adapterName, AdapterArgs args,
+                               AdapterConfigurer configurer) throws Exception {
+    configurer.addRuntimeArgument("adapter.args", new Gson().toJson(args));
+    configurer.setSchedule(Schedules.createTimeSchedule("test", "adapter schedule", "* * * * *"));
+  }
+
 }
--- a/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionMapReduce.java
+++ b/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionMapReduce.java
@@ -28,6 +28,7 @@ import co.cask.cdap.conversion.avro.Converter;
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+import com.google.gson.Gson;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.mapred.AvroKey;
@@ -48,7 +49,6 @@ import java.util.Map;
  */
 public class StreamConversionMapReduce extends AbstractMapReduce {
   private static final Logger LOG = LoggerFactory.getLogger(StreamConversionMapReduce.class);
-  public static final String ADAPTER_PROPERTIES = "adapter.properties";
   public static final String SCHEMA_KEY = "cdap.stream.conversion.output.schema";
   public static final String HEADERS_KEY = "cdap.stream.conversion.headers";
   private String sinkName;
@@ -69,16 +69,18 @@ public class StreamConversionMapReduce extends AbstractMapReduce {
     job.setMapOutputKeyClass(AvroKey.class);
     job.setMapOutputValueClass(NullWritable.class);
 
-    AdapterArguments adapterArguments = new AdapterArguments(context.getRuntimeArguments());
-    sinkName = adapterArguments.getSinkName();
+    AdapterArgs args = new Gson().fromJson(
+      context.getRuntimeArguments().get("adapter.args"), AdapterArgs.class);
+    ConversionConfig config = args.getConfig();
+    sinkName = config.getSinkName();
     partitionTime = context.getLogicalStartTime();
 
     // setup input for the job
     long endTime = context.getLogicalStartTime();
-    long startTime = endTime - adapterArguments.getFrequency();
-    StreamBatchReadable.useStreamInput(context, adapterArguments.getSourceName(), startTime, endTime,
-                                       adapterArguments.getSourceFormatSpec());
-    context.setMapperResources(adapterArguments.getMapperResources());
+    long startTime = endTime - config.getFrequency();
+    StreamBatchReadable.useStreamInput(context, config.getSourceName(), startTime, endTime,
+                                       config.getSourceFormatSpec());
+    context.setMapperResources(config.getMapperResources());
 
     // setup output for the job
     Map<String, String> sinkArgs = Maps.newHashMap();
@@ -86,14 +88,14 @@ public class StreamConversionMapReduce extends AbstractMapReduce {
     TimePartitionedFileSet sink = context.getDataset(sinkName, sinkArgs);
     outputPath = FileSetArguments.getOutputPath(sink.getEmbeddedFileSet().getRuntimeArguments());
     context.setOutput(sinkName, sink);
-    AvroJob.setOutputKeySchema(job, adapterArguments.getSinkSchema());
+    AvroJob.setOutputKeySchema(job, config.getSinkSchema());
 
     // set configuration the mappers will need
-    job.getConfiguration().set(SCHEMA_KEY, adapterArguments.getSinkSchema().toString());
-    if (adapterArguments.getHeadersStr() != null) {
-      job.getConfiguration().set(HEADERS_KEY, adapterArguments.getHeadersStr());
+    job.getConfiguration().set(SCHEMA_KEY, config.getSinkSchema().toString());
+    if (config.getHeaders() != null && !config.getHeaders().isEmpty()) {
+      job.getConfiguration().set(HEADERS_KEY, config.getHeaders());
     }
-    job.setJobName("adapter.stream-conversion." + adapterArguments.getSourceName()
+    job.setJobName("adapter.stream-conversion." + config.getSourceName()
                      + ".to." + sinkName + "." + partitionTime);
   }
 
--- a/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionMapReduce.java
+++ b/cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionMapReduce.java
@@ -70,7 +70,7 @@ public class StreamConversionMapReduce extends AbstractMapReduce {
     job.setMapOutputValueClass(NullWritable.class);
 
     AdapterArgs args = new Gson().fromJson(
-      context.getRuntimeArguments().get("adapter.args"), AdapterArgs.class);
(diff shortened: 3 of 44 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT1342

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionAdapter.java
```diff
@@ -16,17 +16,51 @@
...
+      .setExploreOutputFormat("org.apache.hadoop.hive.ql.io.avro.AvroContainerOutputFormat")
+      .setTableProperty("avro.schema.literal", eventSchema.toString())
+      .build());
   }
+
+  @Override
+  public void configureAdapter(String adapterName, AdapterArgs args,
+                               AdapterConfigurer configurer) throws Exception {
+    configurer.addRuntimeArgument("adapter.args", new Gson().toJson(args));
+    configurer.setSchedule(Schedules.createTimeSchedule("test", "adapter schedule", "* * * * *"));
```
sreevatsanraman: We had discussed that the schedule will be added on start and not while configuring adapters. this shouldn't be there right?

albertshau: this is the template configuring what the schedule for an adapter should be. The platform will generate a schedule spec from this and create it when an adapter is started.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryAdapterConfigurator.java
```diff
@@ -0,0 +1,122 @@
...
+import java.lang.reflect.Type;
+import java.util.jar.Manifest;
+
+/**
+ * In Memory Configurator doesn't spawn a external process, but
+ * does this in memory.
+ *
+ * @see SandboxConfigurator
+ */
+public final class InMemoryAdapterConfigurator {
```
sreevatsanraman: Any reason not to implement Configurator?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryAdapterConfigurator.java
```diff
@@ -0,0 +1,122 @@
...
+    this.adapterName = adapterName;
+  }
+
+  /**
+   * Executes the <code>Application.configure</code> within the same JVM.
+   * <p>
+   * This method could be dangerous and should be used only in standalone mode.
+   * </p>
+   *
+   * @return A instance of {@link ListenableFuture}.
```
sreevatsanraman: Doesn't return a ListenableFuture. Related to earlier comment. if we are not expecting to implement configurator fix docs.

sreevatsanraman: Can you document when an Exception will be thrown?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/AdapterVerificationStage.java
```diff
@@ -0,0 +1,58 @@
...
+   * and verifies both.
+   *
+   * @param input An instance of {@link AdapterSpecification}
+   */
+  @Override
+  public void process(AdapterSpecification input) throws Exception {
+    Preconditions.checkNotNull(input);
+
+    // if this adapter uses a workflow, a schedule should be set
+    if (templateSpec.getWorkflows().size() == 1) {
```
sreevatsanraman: shouldn't this be >=1. I know we won't have more than one workflow per adapter now. but it might lead to a bug later on.

sreevatsanraman: We should also check if there is one workflow or one worker. It cannot have both.

cdap-app-fabric/src/main/java/co/cask/cdap/templates/DefaultAdapterConfigurer.java
```diff
@@ -0,0 +1,180 @@
...
+    // defaults
+    this.resources = new Resources();
+    this.instances = 1;
+  }
+
+  public void addStream(Stream stream) {
+    Preconditions.checkArgument(stream != null, "Stream cannot be null.");
+    StreamSpecification spec = stream.configure();
+    streams.put(spec.getName(), spec);
+    instances = 1;
```
sreevatsanraman: why do we need to set instances to 1 here?

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionAdapter.java
sreevatsanraman: got it. SGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryAdapterConfigurator.java
albertshau: no reason will change.

cdap-app-fabric/src/main/java/co/cask/cdap/templates/DefaultAdapterConfigurer.java
albertshau: oops should not be there, good catch.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterLifecycleHttpHandler.java
```diff
@@ -157,12 +155,10 @@ public void deleteAdapter(HttpRequest request, HttpResponder responder,
     try {
       adapterService.removeAdapter(Id.Namespace.from(namespaceId), adapterName);
       responder.sendStatus(HttpResponseStatus.OK);
+    } catch (InvalidAdapterOperationException e) {
```
bdmogal: when is this thrown? DELETE responding with CONFLICT seems strange?

bdmogal: Just checked. This should probably throw a FORBIDDEN instead.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -230,44 +227,63 @@ public AdapterStatus setAdapterStatus(Id.Namespace namespace, String adapterName
...
-    AdapterSpecification adapterSpec = getAdapter(namespace, adapterName, Object.class);
-    ApplicationSpecification appSpec =
-      store.getApplication(Id.Application.from(namespace, adapterSpec.getTemplate()));
-    unschedule(namespace, appSpec, appTemplateInfos.get(adapterSpec.getTemplate()), adapterSpec);
+  public void removeAdapter(Id.Namespace namespace, String adapterName)
+    throws NotFoundException, InvalidAdapterOperationException {
+
+    AdapterStatus adapterStatus = getAdapterStatus(namespace, adapterName);
+    if (adapterStatus != AdapterStatus.STOPPED) {
+      throw new InvalidAdapterOperationException("Cannot delete an adapter unless it is stopped.");
```
bdmogal: Should throw ``CannotBeDeletedException``? Or a subclass if necessary.

albertshau: ah ok, wil do

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterLifecycleHttpHandler.java
albertshau: will do

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/AdapterVerificationStage.java
albertshau: the number of workflows or workers is checked in the template verification stage, which is why im not checking it here.  This is for adapter related checks, assuming a valid template.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/ConfigureAdapterStage.java
```diff
@@ -0,0 +1,93 @@
...
+                                            tmpLocation.toURI(), outputLocation.toURI()));
+      }
+    } catch (IOException e) {
+      // In case copy to temporary file failed, or rename failed
+      tmpLocation.delete();
+      throw e;
+    }
+
+    InMemoryAdapterConfigurator inMemoryAdapterConfigurator =
+      new InMemoryAdapterConfigurator(namespace, new LocalLocationFactory().create(input.toURI()), adapterName,
```
bdmogal: ``LocationFactory`` is not injected here because we strictly need it to be ``LocalLocationFactory``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/AdapterVerificationStage.java
albertshau: there are assumptions elsewhere that there is just one workflow so I would rather leave it in.  Maybe I can add a unit test that checks that you can't create a template with 2 workflows so that we can't allow more programs in the future without breaking tests.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/ConfigureAdapterStage.java
albertshau: yup template jar will always be a local jar

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/ConfigureAdapterStage.java
```diff
@@ -0,0 +1,93 @@
...
+   * the process of generation of {@link ApplicationSpecification}
+   *
+   * @param deploymentInfo Location of the input and output location
+   */
+  @Override
+  public void process(AdapterDeploymentInfo deploymentInfo) throws Exception {
+
+    Location outputLocation = deploymentInfo.getTempJarLoc();
+    Location parent = Locations.getParent(outputLocation);
+    Locations.mkdirsIfNotExists(parent);
```
bdmogal: Is this location namespaced? Would this ever be called when the namespace directory does not exist? If it creates the namespace directory when the namespace does not exist in the MDS, it may be a problem.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -215,9 +215,8 @@ public void deleteApp(HttpRequest request, HttpResponder responder,
 
       // Deletion of a particular application is not allowed if that application is used by an adapter
       if (adapterService.getApplicationTemplateInfo(appId) != null) {
-        responder.sendString(HttpResponseStatus.BAD_REQUEST,
-                             String.format("Cannot delete Application %s." +
-                                             " An AdapterType exists with a conflicting name.", appId));
+        responder.sendString(HttpResponseStatus.BAD_REQUEST, String.format(
+          "Cannot delete Application %s. An ApplicationTemplate exists with a conflicting name.", appId));
```
bdmogal: Is this message correct? (Although not introduced by your PR). Perhaps 'An ApplicationTemplate is using the specified application'?

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/AdapterArgs.java
```diff
@@ -0,0 +1,52 @@
...
+import co.cask.cdap.api.Resources;
+import co.cask.cdap.api.data.format.FormatSpecification;
+import co.cask.cdap.api.data.schema.Schema;
+import com.google.gson.JsonObject;
+
+import java.io.IOException;
+import java.util.Collections;
+
+/**
+ */
```
chtyim: some javadoc is missing? What kind of argument is this class representing? Is it comes from runtime arguments?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
gokulavasan: Currently the name of the Application Template (which extends AbstractApplication) is same as the name of the application that is deployed. So I think the error msg is right.

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/AdapterArgs.java
gokulavasan: Looks like this is specific to Stream Conversion Adapter. And it seems like the arguments are coming from the configuration of the StreamConversion Adapter.

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/AdapterArgs.java
```diff
@@ -0,0 +1,52 @@
...
+  private final JsonObject schema;
+  private final String headers;
+  private final String source;
+  private final String sink;
+  private final String frequency;
+
+  public AdapterArgs(String format, JsonObject schema, String headers, String source, String sink, String frequency) {
+    this.format = format;
+    this.schema = schema;
+    this.headers = headers;
```
gokulavasan: headers is a string and not a map?

gokulavasan: Oh I see you are creating a map from the string in ConversionConfig.

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionAdapter.java
```diff
@@ -16,17 +16,51 @@
...
+import co.cask.cdap.api.templates.ApplicationTemplate;
+import com.google.gson.Gson;
+import org.apache.avro.mapreduce.AvroKeyInputFormat;
+import org.apache.avro.mapreduce.AvroKeyOutputFormat;
 
 /**
  * Application that converts a stream into a partitioned file set.
  */
-public class StreamConversionAdapter extends AbstractApplication {
+public class StreamConversionAdapter extends ApplicationTemplate<AdapterArgs> {
```
gokulavasan: So StreamConversionAdapter is a different AppTemplate? Isn't it just adapter created from ETLBatch template?

gokulavasan: Or is it to test the app template deployment logic. If that is the case, then ok.

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionMapReduce.java
```diff
@@ -69,16 +69,18 @@ public void beforeSubmit(MapReduceContext context) throws Exception {
     job.setMapOutputKeyClass(AvroKey.class);
     job.setMapOutputValueClass(NullWritable.class);
 
-    AdapterArguments adapterArguments = new AdapterArguments(context.getRuntimeArguments());
-    sinkName = adapterArguments.getSinkName();
+    AdapterArgs args = new Gson().fromJson(
+      context.getRuntimeArguments().get("adapter.args"), AdapterArgs.class);
```
gokulavasan: Move "adapter.args" to some Constants class?

cdap-adapters/src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -1,14 +1,5 @@
 CDAP-Adapter-Type: stream-conversion
```
gokulavasan: Do we still need these properties in the Manifest? Will they eventually be removed?

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/AdapterArgs.java
gokulavasan: Looks like this is the class analogous to the ETLBatchConfig. POJO to represent the configuration of a given app-template (in this case StreamConversion AppTemplate).

cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterConfigurer.java
```diff
@@ -43,6 +44,13 @@
   public void setInstances(int instances);
 
   /**
+   * Set the resources the program should use.
+   *
+   * @param resources
+   */
+  public void setResources(Resources resources);
```
gokulavasan: In the case of ETL Adapters, will this be derived from the source, sink specification? For example, in Batch ETL Adapter, the source and sink set the resources required for Mapper through their context (there is a bug which needs to be fixed - to figure out if resources are already set and use the max). For realtime case, the source, sink specification will return resources which could then be used to implement the AdapterConfigurer#setResources method for ETL Realtime template. And so for Batch case we can return some constant?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalAdapterManager.java
```diff
@@ -0,0 +1,58 @@
...
+    this.store = store;
+  }
+
+  @Override
+  public ListenableFuture<AdapterSpecification> deploy(Id.Namespace namespace, String id,
+                                                       AdapterDeploymentInfo input) throws Exception {
+    Pipeline<AdapterSpecification> pipeline = pipelineFactory.getPipeline();
+    pipeline.addLast(new ConfigureAdapterStage(namespace, id));
+    pipeline.addLast(new AdapterVerificationStage(input.getTemplateSpec()));
+    // TODO: stages for creating dataset instances and streams
```
gokulavasan: Need to figure out if this will be required if we support creating dataset instances / streams during runtime through Source/Sink context.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationTemplateManager.java
```diff
@@ -0,0 +1,119 @@
...
+    this.preferencesStore = preferencesStore;
+  }
+
+  @Override
+  public ListenableFuture<ApplicationWithPrograms> deploy(Id.Namespace id, @Nullable String templateId,
+                                                          DeploymentInfo input) throws Exception {
+    Pipeline<ApplicationWithPrograms> pipeline = pipelineFactory.getPipeline();
+    pipeline.addLast(new LocalArchiveLoaderStage(store, configuration, id, templateId));
+    pipeline.addLast(new ApplicationTemplateVerificationStage(store, datasetFramework, adapterService));
+    pipeline.addLast(new DeployDatasetModulesStage(configuration, datasetFramework, inMemoryDatasetFramework));
```
gokulavasan: Why do we have DeployDatasetModule, createDatasetInstance, createStreamStage etc in ApplicationTemplateManager? App Template should not be able to do any of these, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/ConfigureAdapterStage.java
gokulavasan: In the create adapter endpoint, if we check for the existence of the namespace and return error if it doesn't exist, would that be sufficient?

cdap-app-fabric/src/main/java/co/cask/cdap/templates/AdapterSpecification.java
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Specification of an adapter.
+ */
+public final class AdapterSpecification {
+  private final String name;
+  private final String description;
+  private final String template;
+  private final ScheduleSpecification scheduleSpec;
+  private final WorkerSpecification workerSpec;
+  private final Map<String, StreamSpecification> streams;
```
gokulavasan: Same question as the before. Do we need to this info so that platform will create the streams, datasets? Or will the source/sink be able to do it themselves?

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/AdapterArgs.java
albertshau: yeah. All this is also going away, so didn't spend much time on it.

cdap-adapters/src/main/java/co/cask/cdap/conversion/app/StreamConversionAdapter.java
albertshau: yeah just temporary for testing

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalAdapterManager.java
albertshau: we should talk more about it, but i think this is the right approach.  Makes sure users are not creating tons of datasets and streams without thinking about it, and is more consistent with our other apis

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationTemplateManager.java
albertshau: Was thinking there could be a case for having a template wide dataset or something like that. Kind of like a static variable in a class.  But maybe it would be best to restrict it unless we have a hard use case, so i will take it out.

cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterConfigurer.java
albertshau: it is up to the template to call this how they want

cdap-adapters/src/main/resources/META-INF/MANIFEST.MF
albertshau: these are removed in https://github.com/caskdata/cdap/pull/1907

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/ConfigureAdapterStage.java
albertshau: the location is namespaced and the rest api checks that the namespace exists before deploying

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/AdapterVerificationStage.java
sreevatsanraman: ok

sreevatsanraman:
LGTM :+1:

PR #1903, opened 2015-04-04
fixes modal backdrops and some general padding issues

astewes:


elfenheart:
@astewes still problem with Modal at Edit Stream Properties:

screen shot 2015-04-04 at 8 56 31 pm

elfenheart:
- Label is not aligned with text input
- Metric picker should be input group

PR #1904, opened 2015-04-04
Fix get/set worker instances in ProgramClient for non-default namespace

jwang47:


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -74,6 +74,7 @@
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableList;
```
anwar6953: necessary import?

cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
```diff
@@ -323,7 +323,8 @@ public void setFlowletInstances(String appId, String flowId, String flowletId, i
    */
   public int getWorkerInstances(String appId, String workerId) throws IOException, NotFoundException,
     UnauthorizedException {
-    URL url = config.resolveURL(String.format("apps/%s/workers/%s/instances", appId, workerId));
+    URL url = VersionMigrationUtils.resolveURL(config, ProgramType.WORKER,
```
anwar6953: We can simply use `config.resolveNamespacedURLV3(path)`, since this endpoint definitely available in v3 APIs.

nitinmotgi:
@jwang47 can you address this immediately.

anwar6953:
LGTM

PR #1905, opened 2015-04-04
Feature/adapter logs

bdmogal:
Not ready for merging yet, this is just an approach of how logging for adapters could potentially work. 

- [x] Creates a ``BatchAdapterLoggingContext`` that is used if a mapreduce program was run via a schedule that started through an adapter.
- [x] Adds the ``/namespaces/[namespace-id]/adapters/[adapter-id]/logs`` API
- [x] End-to-end testing - Deployed an adapter, used the above API to get logs for the adapter.

Some stuff that is not nice in this PR, still exploring ways to fix, but some of this may require larger refactoring
- [ ] Uses ``adapterName`` encoded in ``scheduleName``
- [ ] Separation of concerns - ``ApplicationLoggingContext`` accepts an ``adapterId``, ``BasicMapreduceContext`` knows about ``adapterId`` as well.

Things that we need in future, but are not part of this PR:
* Currently gets ``templateId``, ``programType`` and ``programId`` as parameters in the ``/namespaces/[namespace-id]/adapters/[adapter-id]/logs`` API. Have to live with this until we have an ``AdapterClient`` that can be used from the log service to get this info by making a remote call?

Build: https://builds.cask.co/browse/CDAP-DUT1346-1
Issue: [CDAP-1786](https://issues.cask.co/browse/CDAP-1786)

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
@@ -68,7 +68,8 @@ public abstract class AbstractMapReduceContextBuilder {
                                      Program mrProgram,
                                      @Nullable String inputDataSetName,
                                      @Nullable List<Split> inputSplits,
-                                     @Nullable String outputDataSetName) {
+                                     @Nullable String outputDataSetName,
+                                     @Nullable String adapterName) {
     Injector injector = prepare();
 
     // Initializing Program
@@ -104,10 +105,9 @@ public abstract class AbstractMapReduceContextBuilder {
     // Creating mapreduce job context
     MapReduceSpecification spec = program.getApplicationSpecification().getMapReduce().get(program.getName());
     BasicMapReduceContext context =
-      new BasicMapReduceContext(program, type, RunIds.fromString(runId), taskId,
-                                runtimeArguments, datasets, spec, logicalStartTime,
-                                workflowBatch, discoveryServiceClient, metricsCollectionService,
-                                datasetFramework);
+      new BasicMapReduceContext(program, type, RunIds.fromString(runId), taskId, runtimeArguments, datasets, spec,
+                                logicalStartTime, workflowBatch, discoveryServiceClient, metricsCollectionService,
+                                datasetFramework, adapterName);
 
     // propagating tx to all txAware guys
     // NOTE: tx will be committed by client code
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
@@ -53,15 +53,15 @@ import javax.annotation.Nullable;
 public class BasicMapReduceContext extends AbstractContext implements MapReduceContext {
 
   private final MapReduceSpecification spec;
-  private final MapReduceLoggingContext loggingContext;
+  private final LoggingContext loggingContext;
   private final long logicalStartTime;
   private final String workflowBatch;
   private final Metrics userMetrics;
   private final MetricsCollectionService metricsCollectionService;
+  private final String adapterName;
 
   private String inputDatasetName;
   private List<Split> inputDataSelection;
-
   private String outputDatasetName;
   private Job job;
   private Dataset outputDataset;
@@ -79,12 +79,14 @@ public class BasicMapReduceContext extends AbstractContext implements MapReduceC
                                String workflowBatch,
                                DiscoveryServiceClient discoveryServiceClient,
                                MetricsCollectionService metricsCollectionService,
-                               DatasetFramework dsFramework) {
+                               DatasetFramework dsFramework,
+                               @Nullable String adapterName) {
     super(program, runId, runtimeArguments, datasets,
           getMetricCollector(metricsCollectionService, program, type, runId.getId(), taskId),
           dsFramework, discoveryServiceClient);
     this.logicalStartTime = logicalStartTime;
     this.workflowBatch = workflowBatch;
+    this.adapterName = adapterName;
     this.metricsCollectionService = metricsCollectionService;
 
     if (metricsCollectionService != null) {
@@ -92,7 +94,8 @@ public class BasicMapReduceContext extends AbstractContext implements MapReduceC
     } else {
       this.userMetrics = null;
     }
-    this.loggingContext = new MapReduceLoggingContext(getNamespaceId(), getApplicationId(), getProgramName());
+    this.loggingContext = new MapReduceLoggingContext(getNamespaceId(), getApplicationId(), getProgramName(),
+                                                      getAdapterName());
     this.spec = spec;
     this.mapperResources = spec.getMapperResources();
     this.reducerResources = spec.getReducerResources();
@@ -122,6 +125,14 @@ public class BasicMapReduceContext extends AbstractContext implements MapReduceC
    */
   public String getWorkflowBatch() {
     return workflowBatch;
+  }
+
+  /**
+   * @return the adapter name if run from within adapter, otherwise null
+   */
+  @Nullable
+  public String getAdapterName() {
+    return adapterName;
   }
 
   public void setJob(Job job) {
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
@@ -53,6 +53,7 @@ public final class MapReduceContextConfig {
   private static final String HCONF_ATTR_RUN_ID = "hconf.program.run.id";
   private static final String HCONF_ATTR_LOGICAL_START_TIME = "hconf.program.logical.start.time";
   private static final String HCONF_ATTR_WORKFLOW_BATCH = "hconf.program.workflow.batch";
+  private static final String HCONF_ATTR_ADAPTER_NAME = "hconf.program.adapter.name";
   private static final String HCONF_ATTR_ARGS = "hconf.program.args";
   private static final String HCONF_ATTR_PROGRAM_JAR_NAME = "hconf.program.jar.name";
   private static final String HCONF_ATTR_CCONF = "hconf.cconf";
@@ -70,7 +71,12 @@ public final class MapReduceContextConfig {
                   Transaction tx, String programJarName) {
     setRunId(context.getRunId().getId());
     setLogicalStartTime(context.getLogicalStartTime());
-    setWorkflowBatch(context.getWorkflowBatch());
+    if (context.getWorkflowBatch() != null) {
+      setWorkflowBatch(context.getWorkflowBatch());
+    }
+    if (context.getAdapterName() != null) {
+      setAdapterName(context.getAdapterName());
+    }
     setArguments(context.getRuntimeArguments());
     setProgramJarName(programJarName);
     setConf(conf);
@@ -117,15 +123,20 @@ public final class MapReduceContextConfig {
   }
 
   private void setWorkflowBatch(String workflowBatch) {
-    if (workflowBatch != null) {
-      jobContext.getConfiguration().set(HCONF_ATTR_WORKFLOW_BATCH, workflowBatch);
-    }
+    jobContext.getConfiguration().set(HCONF_ATTR_WORKFLOW_BATCH, workflowBatch);
   }
 
   public String getWorkflowBatch() {
     return jobContext.getConfiguration().get(HCONF_ATTR_WORKFLOW_BATCH);
   }
 
+  private void setAdapterName(String adapterName) {
+    jobContext.getConfiguration().set(HCONF_ATTR_ADAPTER_NAME, adapterName);
+  }
+
+  public String getAdapterName() {
+    return jobContext.getConfiguration().get(HCONF_ATTR_ADAPTER_NAME);
+  }
 
   private void setProgramJarName(String programJarName) {
     jobContext.getConfiguration().set(HCONF_ATTR_PROGRAM_JAR_NAME, programJarName);
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
@@ -44,6 +44,7 @@ import co.cask.cdap.internal.lang.Reflections;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.tephra.TransactionSystemClient;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
 import com.google.common.base.Throwables;
 import com.google.common.reflect.TypeToken;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -129,6 +130,7 @@ public class MapReduceProgramRunner implements ProgramRunner {
                                 : System.currentTimeMillis();
 
     String workflowBatch = arguments.getOption(ProgramOptionConstants.WORKFLOW_BATCH);
+    String adapterName = getAdapterName(arguments);
     MapReduce mapReduce;
     try {
       mapReduce = new InstantiatorFactory(false).get(TypeToken.of(program.<MapReduce>getMainClass())).create();
@@ -140,7 +142,7 @@ public class MapReduceProgramRunner implements ProgramRunner {
     final DynamicMapReduceContext context =
       new DynamicMapReduceContext(program, null, runId, null,
                                   options.getUserArguments(), spec,
-                                  logicalStartTime, workflowBatch,
+                                  logicalStartTime, workflowBatch, adapterName,
                                   discoveryServiceClient, metricsCollectionService,
                                   txSystemClient, datasetFramework);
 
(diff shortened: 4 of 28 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/ProgramOptionConstants.java
```diff
@@ -48,4 +48,6 @@
   public static final String RUN_BASE_COUNT_SIZE = "runBaseCountSize";
 
   public static final String RUN_BASE_COUNT_TIME = "runBaseCountTime";
+
+  public static final String ADAPTER_SPECIFICATION = "adapterSpeciciation";
```
bdmogal: not needed. should change to ``ADAPTER_NAME``, after moving adapter name out of schedule name

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -92,7 +97,13 @@ public BasicMapReduceContext(Program program,
     } else {
       this.userMetrics = null;
     }
-    this.loggingContext = new MapReduceLoggingContext(getNamespaceId(), getApplicationId(), getProgramName());
+    String adapterName = getAdapterName(runtimeArguments);
```
chtyim: Is the `runtimeArguments` the user runtime arguments? I think it should come from `ProgramOptions.getArguments`, but not `ProgramOptions.getUserArguments`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -92,7 +97,13 @@ public BasicMapReduceContext(Program program,
...
+    if (adapterName == null) {
+      this.loggingContext = new MapReduceLoggingContext(getNamespaceId(), getApplicationId(), getProgramName());
+    } else {
+      this.loggingContext = new BatchAdapterLoggingContext(getNamespaceId(), getApplicationId(), getProgramName(),
```
chtyim: Why need to have a separate class just for the sake of setting the system tag? Can we just have the `AbstractLoggingContext` to have a constructor that takes `Arguments` and extract adapter name from there? Then we don't need to keep modifying code when allowing more program types to be adapter program.

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/BatchAdapterLoggingContext.java
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.logging.context;
+
+/**
+ * Logging context for instances of ETL templates
```
gokulavasan: Is this specifically for ETL Batch Template? Or for both ETL batch and ETL realtime templates?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
bdmogal: You're right, I misread this. Have fixed it to follow a similar path as ``workflowBatch``, but still need to test it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
bdmogal: I was initially under the impression that adapter logs needed to be in their own partition/own directory etc. However, doesn't look like that is needed. Removed ``BatchAdapterLoggingContext``. Added an optional ``.adapterId`` tag in ``ApplicationLoggingContext``, that various programs can set if required.

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/BatchAdapterLoggingContext.java
bdmogal: The initial idea I had was to have two different logging contexts. However, after exploring a bit more, removed the necessity of a new logging context. We just have an extra, optional tag in ``ApplicationLoggingContext`` now.

bdmogal:
I have made some more progress, however, this still needs a lot of testing end-to-end. Going to do that between today and tomorrow.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
chtyim: I think it shouldn't even be set. The logging context can simply determine it from the program option. By doing so, we also enforced one way to pass in the adapter is

chtyim: Alternatively, do the extraction at the AbstractContext level so that it can set for both metrics and logging

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -124,6 +129,14 @@ public String getWorkflowBatch() {
     return workflowBatch;
   }
 
+  /**
+   * @return the adapter name if run from within adapter, otherwise null
+   */
+  @Nullable
+  public String getAdapterName() {
```
nitinmotgi: What would this be when it's not part of adapter ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
```diff
@@ -106,7 +107,7 @@ public BasicMapReduceContext build(MapReduceMetrics.TaskType type,
     BasicMapReduceContext context =
       new BasicMapReduceContext(program, type, RunIds.fromString(runId), taskId,
                                 runtimeArguments, datasets, spec, logicalStartTime,
-                                workflowBatch, discoveryServiceClient, metricsCollectionService,
+                                workflowBatch, adapterName, discoveryServiceClient, metricsCollectionService,
```
nitinmotgi: Adapter is not concept known to MapReduce right ? If we make it that, what happens when it's not associated with Adapter, but used as is in more general terms.

nitinmotgi:
Adapter is now being made as a concept similar to Application. This should not be the case.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
bdmogal: ``null``, in the current design. Also, won't be added to the ``LoggingContext``, and in future, to the ``MetricsContext``, either

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
bdmogal: when its used outside of adapters, the ``adapterName`` would be null in ``BasicMapreduceContext``. Similar pattern as ``workflowBatch``, but perhaps we need some refactoring here for ``BasicMapreduceContext`` to only receive ``Arguments`` and do the extraction that it needs within.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
```diff
@@ -63,6 +63,7 @@ public BasicMapReduceContext build(MapReduceMetrics.TaskType type,
                                      String taskId,
                                      long logicalStartTime,
                                      String workflowBatch,
+                                     String adapterName,
```
henry-cask: It is probably better to add new parameter in the constructor at the end rather than stick it in the middle. Make it easier to avoid merge conflict and update code.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -77,6 +79,7 @@ public BasicMapReduceContext(Program program,
                                MapReduceSpecification spec,
                                long logicalStartTime,
                                String workflowBatch,
+                               @Nullable String adapterName,
```
henry-cask: Probably better to just add new parameter to the end rather than sneak it in the middle to avoid merge conflict and update to existing code.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
```diff
@@ -126,6 +130,15 @@ public String getWorkflowBatch() {
     return jobContext.getConfiguration().get(HCONF_ATTR_WORKFLOW_BATCH);
   }
 
+  private void setAdapterName(@Nullable String adapterName) {
```
henry-cask: probably not a good idea to let setter to pass null. Caller to make sure not to pass null when trying to set adapter name to context config.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -178,4 +181,22 @@ public ProgramController run(Program program, ProgramOptions options) {
     }
     return controller;
   }
+
+  @Nullable
```
henry-cask: Not sure what use of annotation Nullable here? It is used primarily to indicate if we could pass null as value for input parameters.
Maybe using JavaDoc explaining it could return null instead.

henry-cask: Ah looks like it could be use by IntelliJ Idea to check if method could return null.

bdmogal: Yup, we use it pretty extensively to indicate null return values.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
bdmogal: I could go either way on this. This same class uses two different styles of setting nulls. ``setWorkflowBatch`` allows callers to pass null and handles it internally, while ``setInputSelection`` expects callers to handle nulls before calling.

Following your suggestion, but applying the same to ``setWorkflowBatch`` as well.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
bdmogal: Fixed. Wanted to keep it next to ``workflowBatch``, but don't feel too strongly about it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
bdmogal: Fixed. Again, Wanted to keep it next to ``workflowBatch``, but don't feel too strongly.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -228,4 +230,16 @@ public void error(Throwable cause) {
...
+  private String getAdapterName(Arguments arguments) {
+    // TODO: Currently this logic is super ugly, should try and pass adapter name as its own separate arg
+    if (arguments.hasOption(ProgramOptionConstants.SCHEDULE_NAME)) {
+      String scheduleName = arguments.getOption(ProgramOptionConstants.SCHEDULE_NAME);
+      if (scheduleName.contains(".")) {
```
albertshau: this won't be correct, since it is valid for a schedule name to have a dot. I see there is a todo, is that going to be fixed in a separate PR?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
```diff
@@ -84,13 +86,15 @@ public AbstractProgramWorkflowRunner(Arguments runtimeArguments, RunId runId, Pr
    * @return a {@link Callable} of {@link RuntimeContext} for this {@link Program}
    */
   protected Callable<RuntimeContext> getRuntimeContextCallable(String name, final Program program) {
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+    for (Map.Entry<String, String> systemArgument : systemArguments) {
+      builder.put(systemArgument);
+    }
+    builder.put(ProgramOptionConstants.RUN_ID, runId.getId());
+    builder.put(ProgramOptionConstants.WORKFLOW_BATCH, name);
```
albertshau: no longer adding logical start time? is that on purpose?

albertshau: ok i see it is added earlier

bdmogal: This is the issue where we were dropping the existing ``systemArguments`` and recreating then fresh - (lines 89-93 below). Now, workflows accept ``systemArguments`` from a higher layer (example: schedule) and only append more arguments to it, as opposed to recreating the entire ``systemArguments`` afresh. ``LOGICAL_START_TIME`` is already present in the systemArguments passed to this class from a schedule.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
```diff
@@ -80,6 +87,7 @@ public void set(BasicMapReduceContext context, CConfiguration conf,
     }
   }
 
+
```
albertshau: nit: extra newline

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
bdmogal: I see, so schedules don't follow the rules for other ``Id`` classes? In that case, we should probably use a different separator for the schedule name set in ``DefaultAdapterConfigurer``. Since its not easy to get ``adapterName`` separately right now. I am unsure if that TODO will get addressed anytime in the immediate future.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
albertshau: actually, i don't see it.  where is it being set?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
albertshau: yeah actually, not sure why schedule doesn't follow the rest. seems like it should.... But whatever separator is used will be a valid character so it has the same problem.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
bdmogal: So if you look at ``WorkflowDriver``, it received ``ProgramOptions``. From there, it extracted ``logicalStartTime`` via ``programOptions.getArguments().get(...``. Then it passed ``logicalStartTime`` to this class via ``ProgramWorkflowRunnerFactory``. Instead of extracting there, and adding it back here in a new ``BasicArguments`` object, I just pass the original ``BasicArguments`` object around now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
albertshau: why is it hard to get adapter? doesn't the program runner get ProgramOptions?

bdmogal: So this actually follows the following class hierarchy:

*Adapter* -> *CDAP Schedule* -> *Quartz Schedule*

The ``ProgramOption``s are set in a class (``DefaultSchedulerService``) that only gets a Quartz class - ``JobExecutionContext``. And we do not pass any context information from *CDAP Schedule* classes to the Quartz Scheduler classes other than the trigger key which is an encoded string of the form: ``namespaceId:appId:programType:programId:scheduleName``.

albertshau:
ok :+1: . let's get this in and see how we can pass the adapter name through the schedule.

bdmogal:
Build passed. Fixed nits. Merging. Will follow up with a PR with a  different approach to passing adapter name to schedule.

PR #1906, opened 2015-04-04
Fix/ui workflow nested tabs

ajainarayanan:
Fixes workflows feature similar to the flows to fix back button issue a.k.a nested tabs issue we used to have.


elfenheart:
:+1: LGTM

PR #1907, opened 2015-04-04
CDAP-1768 add a way to update template code, by adding an endpoint

albertshau:
that will deploy a template application. Also removing need to
have template name and program type in the manifest, since those
should be coming from the configure() method.


albertshau:
http://builds.cask.co/browse/CDAP-DUT1347

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -106,6 +120,27 @@ protected void shutDown() throws Exception {
...
+   */
+  public void deployTemplate(Id.Namespace namespace, String templateName)
+    throws NotFoundException, InterruptedException, ExecutionException, TimeoutException, IOException {
+    ApplicationTemplateInfo templateInfo = appTemplateInfos.get(templateName);
+    if (templateInfo == null) {
+      throw new NotFoundException("ApplicationTemplate", templateName);
+    }
+    // make sure we're up to date on template info
+    registerTemplates();
+    deployTemplate(namespace, templateInfo);
```
gokulavasan: Will this fail if a program of the previous version of the template-app is currently running (for ex, worker)? Should the user take care of stopping all adapters of the specific template before updating the template?

albertshau: currently it will not fail, and is leaving it to the user to do the right thing depending on their template and the changes they've made.  It has the same behavior as deployment of an application. For workers, the old code will still be running until they restart the adapter. For workflow, the new code will be used at the next scheduled run. I think we eventually want the behavior of workflow to be like that of workers.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
```diff
@@ -50,24 +53,55 @@
...
+  public void deployTemplate(HttpRequest request, HttpResponder responder,
+                             @PathParam("namespace-id") String namespaceId,
+                             @PathParam("template-id") String templateId) {
+    if (!namespaceAdmin.hasNamespace(Id.Namespace.from(namespaceId))) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND,
+                           String.format("Namespace '%s' does not exist.", namespaceId));
+      return;
+    }
+    try {
+      adapterService.deployTemplate(Id.Namespace.from(namespaceId), templateId);
```
sreevatsanraman: We are not returning on success. The requests will hang.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
```diff
@@ -50,24 +53,55 @@
...
     this.namespaceAdmin = namespaceAdmin;
     this.adapterService = adapterService;
   }
 
   /**
+   * Deploy a template. Allows users to update a template, but leaves upgrade specifics to the user.
+   * For example, if a program is renamed or new arguments are required, users must handle recreation of
+   * adapters themselves.
+   */
+  @POST
```
bdmogal: Should probably be a PUT, since its a named, known resource.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -395,42 +429,56 @@ private void deleteSchedule(Id.Program programId, SchedulableProgramType program
...
+    // if the file is the same, just return
+    if (existing != null && fileHash.equals(existing.getFileHash())) {
+      return existing;
+    }
 
-      String name = mainAttributes.getValue(ApplicationTemplateManifestAttributes.NAME);
-      String programType = mainAttributes.getValue(ApplicationTemplateManifestAttributes.PROGRAM_TYPE);
-      return new ApplicationTemplateInfo(file, name, ProgramType.valueOf(programType.toUpperCase()));
+    // instantiate the template application and call configure() on it to determine it's specification
+    InMemoryConfigurator configurator = new InMemoryConfigurator(
```
sreevatsanraman: It is a bit confusing to see System namespace here. If we are deploying to usernamespace, perhaps add a comment or remove the variable in constructor.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
albertshau: oo no good. i will also add a unit test for this endpoint

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
albertshau: hm in that case i should remove 'deploy' from the path then.

sreevatsanraman:
Couple of comments. otherwise looks good.

sreevatsanraman:
LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
```diff
@@ -50,24 +53,56 @@
...
+                             @PathParam("template-id") String templateId) {
+    if (!namespaceAdmin.hasNamespace(Id.Namespace.from(namespaceId))) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND,
+                           String.format("Namespace '%s' does not exist.", namespaceId));
+      return;
+    }
+    try {
+      adapterService.deployTemplate(Id.Namespace.from(namespaceId), templateId);
+      responder.sendString(HttpResponseStatus.OK, "Deploy Complete");
+    } catch (NotFoundException e) {
```
henry-cask: We could just use the catches to set HttpResponseStatus and message and add finally clause to do the responder.sendString call to make sure method sending response properly.

PR #1911, opened 2015-04-06
Fix/ui mapreduce nested tabs

ajainarayanan:
Fixes mapreduce feature for nested tabs (similar to workflow and flows)


elfenheart:
:+1: LGTM

PR #1914, opened 2015-04-06
CDAP-2012: Add application version in the ApplicationSpec

jwang47:
http://builds.cask.co/browse/CDAP-DUT1354

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/ApplicationSpecificationCodec.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/ApplicationSpecificationCodec.java
@@ -48,7 +48,9 @@ final class ApplicationSpecificationCodec extends AbstractSpecificationCodec<App
     JsonObject jsonObj = new JsonObject();
 
     jsonObj.add("name", new JsonPrimitive(src.getName()));
-    jsonObj.add("version", new JsonPrimitive(src.getVersion()));
+    if (src.getVersion() != null) {
+      jsonObj.add("version", new JsonPrimitive(src.getVersion()));
+    }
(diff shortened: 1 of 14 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/ApplicationSpecificationCodec.java
```diff
@@ -70,6 +71,7 @@ public ApplicationSpecification deserialize(JsonElement json, Type typeOfT, Json
     JsonObject jsonObj = json.getAsJsonObject();
 
     String name = jsonObj.get("name").getAsString();
+    String version = jsonObj.get("version").getAsString();
```
chtyim: What if it is missing? E.g. existing spec in MDS?

sreevatsanraman:
Add unit test.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/DefaultApplicationSpecification.java
```diff
@@ -93,6 +95,11 @@ public String getName() {
   }
 
   @Override
+  public String getVersion() {
```
chtyim: Seems like version is `nullable`? What happen if the jar manifest doesn't have `Bundle-Version`?

awholegunch:
Does this change the examples? For instance, in the example Application of the Developer's Manual: Building Block, would we add "setVersion"?

jwang47:
No, this won't allow developers to set the version of their app. It will the pull the version from the jar that they deploy (version is set from jar manifest).

sreevatsanraman:
LGTM :+1:

PR #1915, opened 2015-04-06
Feature/ui styling nitin feedback

astewes:


cdap-ui/app/features/overview/templates/data-apps.html
```diff
@@ -21,16 +21,14 @@ <h2 class="text-uppercase"> Data </h2>
...
+          <a ng-if="data.type === 'Stream'" ui-sref="streams.detail.overview({streamId: data.name})">
+            <p ng-bind="data.name | caskCapitalizeFilter"> </p>
+            <p><img src="assets/img/stream.svg" alt="" /> Stream</p>
           </a>
-        </li>
-        <li class="list-group-item" ng-repeat="dataset in datasets track by $index">
-          <a ui-sref="datasets.detail.overview({datasetId: dataset.name})">
-            <strong ng-bind="dataset.name | caskCapitalizeFilter">  </strong>
-            <div> <p ng-bind="'Type: ' + (dataset.type)"></p> </div>
+          <a ng-if="data.type !== 'Stream'" ui-sref="datasets.detail.overview({datasetId: data.name})">
```
ajainarayanan: Doesn't need to go in this PR but I feel this should be straight forward.

data.type === 'Dataset' & data.type === 'Stream'. Later if we have a new type of data (or icons we need to show or states we need to navigate that differs based on the type) then that should be addressed specifically for the data type.

ajainarayanan:
NOTE: 
  The button line-height needs some love in the dev start view. This will be fixed in the next PR. Otherwise LGTM

PR #1917, opened 2015-04-06
Set CONFIG_KEY key/value in configure method of etlbatch apptemplate an...

gokulavasan:
...d not in test

Bamboo: http://builds.cask.co/browse/CDAP-DT5-1


albertshau:
:+1:

PR #1920, opened 2015-04-06
CDAP-1756 allow templates to add datasets and streams when

albertshau:
configuring an adapter. Some refactoring along the way to make
sure that most of the code is shared between adapters and apps.


albertshau:
http://builds.cask.co/browse/CDAP-DUT1358

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationManager.java
```diff
@@ -100,19 +99,19 @@ public LocalApplicationManager(CConfiguration configuration, PipelineFactory pip
...
+    pipeline.addLast(new DeployDatasetModulesStage(configuration, namespace, datasetFramework,
+                                                   inMemoryDatasetFramework));
+    pipeline.addLast(new CreateDatasetInstancesStage(configuration, datasetFramework, namespace));
+    pipeline.addLast(new CreateStreamsStage(namespace, streamAdmin, exploreFacade, exploreEnabled));
     pipeline.addLast(new DeletedProgramHandlerStage(store, programTerminator, streamConsumerFactory,
                                                     queueAdmin, discoveryServiceClient));
     pipeline.addLast(new ProgramGenerationStage(configuration, namespacedLocationFactory));
     pipeline.addLast(new ApplicationRegistrationStage(store));
-    pipeline.addLast(new CreateSchedulesStage(scheduler));
-    pipeline.setFinally(new DeployCleanupStage());
```
gokulavasan: Why was this removed?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterServiceTests.java
```diff
@@ -144,6 +147,18 @@ public void testRedeploy() throws Exception {
...
+    String streamName = "somestream";
+    String tableName = "sometable";
+    String adapterName = "streamAdapter";
+    DataTemplate.Config config = new DataTemplate.Config(streamName, tableName);
+    AdapterConfig adapterConfig = new AdapterConfig("description", DataTemplate.NAME, GSON.toJsonTree(config));
+    adapterService.createAdapter(NAMESPACE, adapterName, adapterConfig);
+
+    Assert.assertTrue(streamExists(Id.Stream.from(NAMESPACE, streamName)));
+    Assert.assertTrue(datasetExists(Id.DatasetInstance.from(NAMESPACE, tableName)));
+  }
```
gokulavasan: Need a line break

cdap-app-fabric/src/test/java/co/cask/cdap/DataTemplate.java
```diff
@@ -0,0 +1,142 @@
...
+  @Override
+  public void configure() {
+    setName(NAME);
+    addWorker(new DWorker());
+  }
+
+  @Override
+  public void configureAdapter(String adapterName, Config config, AdapterConfigurer configurer) throws Exception {
+    configurer.addDatasetModule("customDataset", CustomDatasetModule.class);
+    if (config.streamName != null && !config.streamName.isEmpty()) {
```
gokulavasan: So the configureAdapter method will decide what streams, datasets to be created from the Adapter configuration? How will configureAdapter know what streams, datasets to create? For example, a specific Source or Sink will know what stream, dataset it needs right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationManager.java
albertshau: this was performing cleanup at the end. I changed it so the stage performs clean up on itself so that it is simpler and less error prone. For example, before, deploy dataset modules could unpack some jars, then you could have have an error in a later stage and those jars might not get cleaned up.

cdap-app-fabric/src/test/java/co/cask/cdap/DataTemplate.java
albertshau: yes, templates will have to declare them during configureAdapter.  For the etl case, this means the driver needs to allow the sources and sinks to add streams and datasets, which seems correct to me. You know everything that is needed to create required datasets at adapter creation time, and you can make sure it happens before any processing begins.

gokulavasan: Ok. Sounds good. It would be good to see how the source, sink APIs need to change to support this.

albertshau: yeah that is still up in the air. One way would be to add a configureAdapter() method to them that configureAdapter then calls.  Another way would be to pass in a config to the source/sink configure method, and add add stream/dataset methods to the StageConfigurer, which the etl driver uses to delegate to the AdapterConfigurer's methods.

albertshau: i would probably add a method like:

```configurePipeline(ETLStage config, PipelineConfigurer pipelineConfigurer)```

gokulavasan: The current configure method in Source/Sink is primarily used to convey the information about the properties that specific source/sink is expecting the user to provide. So it could/should be invoked before we have a adapter configuration.

gokulavasan:
Comment about Source/Sink API change. We can figure that out in a separate PR. Rest LGTM :+1:

PR #1922, opened 2015-04-07
Fixes dashboard remove and adding in sync with backend and user settings

ajainarayanan:
Fixes dashboard creation & deletion in sync between backend and user settings store.


elfenheart:
:+1: LGTM

PR #1923, opened 2015-04-07
adds newest dropdown styles to global nav

astewes:


cdap-ui/app/directives/navbar/navbar.html
```diff
@@ -25,8 +25,8 @@
       <li ng-class="{active: highlightTab == 'development'}">
         <a ui-sref="overview({namespace: $stateParams.namespace})"> Development </a>
       </li>
-      <li ng-class="{active: highlightTab == 'dashboard'}">
-        <a ui-sref="dashboard.standard.cdap({namespace: $stateParams.namespace})"> Operations </a>
+      <li ng-class="{active: highlightTab == 'operations'}">
+        <a ui-sref="operations({namespace: $stateParams.namespace})"> Operations </a>
```
elfenheart: This link is broken

cdap-ui/app/styles/themes/cdap/header.less
```diff
@@ -1,145 +1,226 @@
...
-    .navbar-brand {
-      position: relative;
-      top: -5px;
-
-      > strong {
-        display: inline-block;
-        visibility: hidden;
-        width: 84px;
+body {
+  header [my-navbar] {
```
ajainarayanan: So now if I change the theme would I still be getting navbar just the same? We have header.less inside /styles/themes/cdap/header.less but we are not namespace'ing it with the theme.

ajainarayanan: If this are the set of styles that will be common across themes then probably we should move this to common.less?

astewes:
Dropdown styles are separated into 2 groups - 1) theme-agnostic styling: body header [my-navbar] and 2) theme-specific styling: body.theme-cdap header [my-navbar]. Since the styles are all specific to the my-navbar directive, I think we should keep them out of common.less, which should be reserved for more global stylings, like default margins on a paragraph.

elfenheart:
:+1: LGTM

PR #1924, opened 2015-04-07
Performance improvements for dataset load

poornachandra:
@chtyim @abaranau:
I'm going to merge this PR into branch `hotfix/temp-fix-flow-change-instances` to start performance testing. Please still review the changes.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetServiceModules.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetServiceModules.java
@@ -17,8 +17,11 @@
 package co.cask.cdap.data.runtime;
 
 import co.cask.cdap.api.dataset.module.DatasetDefinitionRegistry;
+import co.cask.cdap.api.dataset.module.DatasetModule;
+import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.runtime.RuntimeModule;
+import co.cask.cdap.data2.datafabric.dataset.DatasetMetaTableUtil;
 import co.cask.cdap.data2.datafabric.dataset.service.DatasetService;
 import co.cask.cdap.data2.datafabric.dataset.service.DistributedUnderlyingSystemNamespaceAdmin;
 import co.cask.cdap.data2.datafabric.dataset.service.LocalUnderlyingSystemNamespaceAdmin;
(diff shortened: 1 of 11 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryFixedModulesDatasetFramework.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  @Override
+  public void deleteAllInstances(Id.Namespace namespaceId) throws DatasetManagementException, IOException {
+    super.deleteAllInstances(namespaceId);
+  }
+
+  @Override
+  public void addModule(Id.DatasetModule moduleId, DatasetModule module) {
+    throw new UnsupportedOperationException("");
```
abaranau: c'mon put some text in it :)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryFixedModulesDatasetFramework.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import javax.annotation.Nullable;
+
+/**
+ * In memory {@link DatasetFramework} implementation that has fixed set of {@link DatasetModule}s.
+ * The modules cannot be changed after the creation of the DatasetFramework.
+ * Since the modules are fixed, this implementation has better performance than {@link InMemoryDatasetFramework}.
+ */
+public class InMemoryFixedModulesDatasetFramework extends InMemoryDatasetFramework implements DatasetFramework {
```
abaranau: hate the name.. Hard to figure out good one..

StaticDatasetFramework?

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetServiceModules.java
```diff
@@ -171,4 +173,30 @@ protected void configure() {
       }
     };
   }
+
+  private static final class DatasetMdsProvider implements Provider<DatasetFramework> {
```
abaranau: can we NOT use provider? Seems like we can do bindings without it, no? Making guice module complex doesn't do good

abaranau:
couple minor comments - otherwise looks fine. Hard to follow the whole DatasetFramework hiearchy. We may want to look at refactoring it... Maybe later with work on dataset framework improvements

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetServiceModules.java
poornachandra: As discussed offline, we need a provider to combine default modules and dataset meta table util values.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/InMemoryFixedModulesDatasetFramework.java
poornachandra: Done

PR #1926, opened 2015-04-07
CDAP-2032: Ensure stream exists according to streamAdmin before processing its config in LocalStreamService

jwang47:
https://issues.cask.co/browse/CDAP-2032
http://builds.cask.co/browse/CDAP-DUT1363


chtyim:
LGTM

PR #1928, opened 2015-04-07
CDAP-2031: Retry for StreamConversionTest

jwang47:
https://issues.cask.co/browse/CDAP-2031
http://builds.cask.co/browse/CDAP-DUT1364

### DIFF
--- a/cdap-examples/StreamConversion/src/test/java/co/cask/cdap/examples/streamconversion/StreamConversionTest.java
+++ b/cdap-examples/StreamConversion/src/test/java/co/cask/cdap/examples/streamconversion/StreamConversionTest.java
@@ -19,6 +19,7 @@ package co.cask.cdap.examples.streamconversion;
 import co.cask.cdap.api.common.RuntimeArguments;
 import co.cask.cdap.api.dataset.lib.TimePartition;
 import co.cask.cdap.api.dataset.lib.TimePartitionedFileSet;
+import co.cask.cdap.common.utils.Tasks;
 import co.cask.cdap.test.ApplicationManager;
 import co.cask.cdap.test.DataSetManager;
 import co.cask.cdap.test.MapReduceManager;
@@ -32,7 +33,11 @@ import java.sql.ResultSet;
 import java.util.Calendar;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicMarkableReference;
 
 /**
  * Test the stream conversion example app.
@@ -60,7 +65,7 @@ public class StreamConversionTest extends TestBase {
     mapReduceManager.waitForFinish(5, TimeUnit.MINUTES);
 
     // verify the single partition in the file set
-    long partitionTime = testWithRetry(new Callable<Long>() {
+    long partitionTime = assertWithRetry(new Callable<Long>() {
       @Override
       public Long call() throws Exception {
         DataSetManager<TimePartitionedFileSet> fileSetManager = getDataset("converted");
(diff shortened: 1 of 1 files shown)


sreevatsanraman:
LGTM :+1:  Merge once the build passes.

cdap-examples/StreamConversion/src/test/java/co/cask/cdap/examples/streamconversion/StreamConversionTest.java
```diff
@@ -99,4 +105,20 @@ public void testStreamConversion() throws Exception {
     Assert.assertEquals(minute, results.getInt(5));
     Assert.assertFalse(results.next());
   }
+
+  private <T> T testWithRetry(Callable<T> callable, int timesToRetry,
```
bdmogal: Can we re-use ``Tasks.waitFor``?

anwar6953: Tasks.waitFor doesn't retry upon exception, I think. Maybe it should be changed to do so.

jwang47: Also `testWithRetry` is meant to allow multiple assertions inside of the `callable`, so it doesn't really fit the use case as it's not really waiting for a single value.

Perhaps we could add a way to specify exception types to ignore to `Tasks.waitFor`, but I don't think that's necessary for this PR.

bdmogal: I see. its a pattern that we see often, so would be good to extract it somewhere. agree that it may be beyond this PR's scope.

chtyim: Why not use the `Tasks.waitFor`?

chtyim: You can have something like this:

```java
<T> T assertWithRetry(Callable<T> callable, long timeout, TimeUnit timeoutUnit,
                                 long sleepDelay, TimeUnit sleepDelayUnit) {
  AtomicMarkableReference<T> result = new AtomicMarkableReference<T>(null, false);
  Tasks.waitFor(true, new Callable<Boolean>() {
    T call() throws Exception {
      result.set(callable.call(), true);
      return true;
    }
  }, timeout, timeoutUnit, sleepDelay, sleepDelayUnit);
  Assert.assertTrue(result.isMarked());
  return result.get();
}
```

anwar6953: I think that won't work with current implementation of Tasks.waitFor, because Tasks.waitFor doesn't do retries upon exceptions.

chtyim: Then, catch the exception in the wrapping callable, if throwing exception is the way to signal retry.

cdap-examples/StreamConversion/src/test/java/co/cask/cdap/examples/streamconversion/StreamConversionTest.java
```diff
@@ -59,11 +65,16 @@ public void testStreamConversion() throws Exception {
...
+    long partitionTime = assertWithRetry(new Callable<Long>() {
+      @Override
+      public Long call() throws Exception {
+        DataSetManager<TimePartitionedFileSet> fileSetManager = getDataset("converted");
+        TimePartitionedFileSet converted = fileSetManager.get();
+        Set<TimePartition> partitions = converted.getPartitionsByTime(startTime, System.currentTimeMillis());
+        Assert.assertEquals(1, partitions.size());
+        return partitions.iterator().next().getTime();
+      }
+    }, 15L, TimeUnit.SECONDS, 3L, TimeUnit.SECONDS);
```
chtyim: Don't sleep for 3 seconds. Sleep shorter in unit-test, e.g. 100ms.

chtyim:
Just one comment, otherwise LGTM.

PR #1929, opened 2015-04-07
CDAP-2034: Removed unnecessary actions in AdapterClientTest, as StandaloneTestBase takes care of cleanup already

jwang47:
https://issues.cask.co/browse/CDAP-2034
http://builds.cask.co/browse/CDAP-DUT1365


sreevatsanraman:
LGTM :+1:

gokulavasan:
How does this resolve the problem?

jwang47:
AdapterClientTest doesn't stop any programs, so if the adapter workflow starts before deleteAll is called, then we get the exception "program still running". I removed that part of the test, since StandaloneTestBase does stop all programs and unrecoverable reset in @AfterClass.

gokulavasan:
Ok so even if MR job is running, the cleanup in StandaloneTestBase do the required cleanup. Thanks Alvin!

PR #1932, opened 2015-04-07
Feature/3.0 ui doc string changes

awholegunch:
- Changes the label for Namespace drop-down menu from "No more Namespaces" to "No other namespaces"
- Changes the footer copyright string to include a © symbol and a self-updating date string (2014-current year)


cdap-ui/app/index.html
```diff
@@ -34,7 +34,8 @@
           <div class="col-xs-12 col-md-6 text-uppercase">
             <p>
               <img src="/assets/img/cask.png" alt="" />
-              <span>&copy; 2015 Cask Data, Inc. All Rights Reserved.</span>
+              <span>Copyright &copy; 2014-<script type="text/javascript">today = new Date();document.write(today.getFullYear());</script>
```
ajainarayanan: In angular world this should be done in a controller. 

The way we do it is, in main.js we have a ```BodyCtrl``` which will load the body element (or the home page). Right now it modifies the css class of the body based on the UI state. 

The ideal way to do this is,

Create a variable in the ```BodyCtrl``` and add it to the scope and use that variable here and use ng-bind-once which will bind that variable to this html element. That way we do javascript in js (controller) and bind the value to the DOM that gets updated only once when the app loads.

cdap-ui/app/directives/navbar/namespace.html
```diff
@@ -7,7 +7,7 @@
       </a>
     </li>
     <li role="menuitem" ng-show="namespaces.length === 1" class="disabled">
-      <a href=""> <strong> No more Namespaces </strong> </a>
+      <a href="">No other namespaces</a>
```
bdmogal: Hmm, @ajainarayanan, has the design for this been finalized? Would it be better instead that we disable the drop down when there are no namespaces?

awholegunch:
The squashing of the commits failed, as they had already been pushed, as the docs had warned about. Closing and re-creating the PR.

PR #1933, opened 2015-04-07
surface worker metrics

gsps1:
worker metrics didn't have aggregation group defined, so we weren't collecting metrics emitted from worker. this is the fix.

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
@@ -130,6 +130,14 @@ public class DefaultMetricStore implements MetricStore {
       // i.e. for service only
       ImmutableList.of(Constants.Metrics.Tag.NAMESPACE, Constants.Metrics.Tag.APP,
                        Constants.Metrics.Tag.SERVICE)));
+    // worker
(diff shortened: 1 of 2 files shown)


gokulavasan:
@gsps1 Can you add some tests?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
```diff
@@ -130,6 +130,14 @@ public Cube get() {
       // i.e. for service only
       ImmutableList.of(Constants.Metrics.Tag.NAMESPACE, Constants.Metrics.Tag.APP,
                        Constants.Metrics.Tag.SERVICE)));
+    // worker
+    aggs.add(new DefaultAggregation(
```
anwar6953: indentation seems off

anwar6953: remainder lgtm

gsps1: indentation is fine, the able comment `` for service only`` is for a parameter, that's why it seems off , but isn't.

gsps1:
http://builds.cask.co/browse/CDAP-DUT1367 - Green - Merging

PR #1934, opened 2015-04-07
change styling of system preferences

elfenheart:
screen shot 2015-04-07 at 2 28 44 pm

### DIFF
--- a/cdap-ui/app/features/admin/admin.less
+++ b/cdap-ui/app/features/admin/admin.less
@@ -2,9 +2,6 @@
 
 body.theme-cdap.state-admin {
   main.container {
-    .system-preferences {
-      margin-top: 25px;
-    }
     .clickable {
       cursor: pointer;
     }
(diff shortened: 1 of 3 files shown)


cdap-ui/app/features/admin/admin.less
```diff
@@ -2,6 +2,9 @@
 
 body.theme-cdap.state-admin {
   main.container {
+    .system-preferences {
+      margin-top: 25px;
```
ajainarayanan: Can we use ```<br/>``` tag? When we want to style the padding later it will become a custom fix just for this. I believe we have been using br tag elsewhere when we need a new line (separation of elements vertically)

cdap-ui/app/features/admin/controllers/PreferencesController.js
```diff
@@ -18,6 +18,27 @@ angular.module(PKG.name + '.feature.admin')
...
 
     $scope.loadProperties = function () {
+      if (source !== 'SYSTEM') {
+        dataSrc
+          .request({
+            _cdapPath: '/preferences'
+          })
+          .then(function (res) {
+
+            console.log('test', res);
```
ajainarayanan: Do we need this if something fails?

ajainarayanan:
:+1: LGTM

PR #1935, opened 2015-04-07
Added listener to the MapReduceRuntimeService to record the state change...

sagarkapare:
...s.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
@@ -26,6 +26,7 @@ import java.net.SocketException;
 import java.util.Enumeration;
 import java.util.Random;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
 /**
@@ -77,12 +78,13 @@ public final class RunIds {
   }
 
   /**
-   * @return time in milliseconds from the UUID if it is a time-based UUID, -1 otherwise.
+   * @return time from the UUID if it is a time-based UUID, -1 otherwise.
    */
-  public static long getTimeMillis(RunId runId) {
+  public static long getTime(RunId runId, TimeUnit timeUnit) {
     UUID uuid = UUID.fromString(runId.getId());
     if (uuid.version() == 1 && uuid.variant() == 2) {
-      return (uuid.timestamp() - NUM_100NS_INTERVALS_SINCE_UUID_EPOCH) / HUNDRED_NANO_MULTIPLIER;
+      long timeInMilliseconds = (uuid.timestamp() - NUM_100NS_INTERVALS_SINCE_UUID_EPOCH) / HUNDRED_NANO_MULTIPLIER;
+      return timeUnit.convert(timeInMilliseconds, TimeUnit.MILLISECONDS);
     }
     return -1;
   }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
@@ -156,52 +156,47 @@ public class MapReduceProgramRunner implements ProgramRunner {
     final Service mapReduceRuntimeService = new MapReduceRuntimeService(cConf, hConf, mapReduce, spec, context,
                                                                         program.getJarLocation(), locationFactory,
                                                                         streamAdmin, txSystemClient);
-    final ProgramController controller = new MapReduceProgramController(mapReduceRuntimeService, context);
-    controller.addListener(new AbstractListener() {
+    mapReduceRuntimeService.addListener(new Service.Listener() {
       @Override
-      public void init(ProgramController.State state, @Nullable Throwable cause) {
+      public void starting() {
         store.setStart(program.getId(), runId.getId(), TimeUnit.SECONDS.convert(System.currentTimeMillis(),
                                                                                 TimeUnit.MILLISECONDS));
-        if (state == ProgramController.State.COMPLETED) {
-          completed();
+      }
+
+      @Override
+      public void running() {
+        //no-op
+      }
+
+      @Override
+      public void stopping(Service.State from) {
+        //no-op
+      }
+
+      @Override
+      public void terminated(Service.State from) {
+        if (from != Service.State.STOPPING) {
+          // Service completed by itself.
+          store.setStop(program.getId(), runId.getId(),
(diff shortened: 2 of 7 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -156,53 +156,48 @@ public ProgramController run(final Program program, ProgramOptions options) {
     final Service mapReduceRuntimeService = new MapReduceRuntimeService(cConf, hConf, mapReduce, spec, context,
                                                                         program.getJarLocation(), locationFactory,
                                                                         streamAdmin, txSystemClient);
-    final ProgramController controller = new MapReduceProgramController(mapReduceRuntimeService, context);
-    controller.addListener(new AbstractListener() {
+    mapReduceRuntimeService.addListener(new Service.Listener() {
       @Override
-      public void init(ProgramController.State state, @Nullable Throwable cause) {
+      public void starting() {
         store.setStart(program.getId(), runId.getId(), TimeUnit.SECONDS.convert(System.currentTimeMillis(),
```
chtyim: Extract the timestamp from the RunId. `RunIds.getTimeMillis(runId)`.

chtyim: Also, you can add method to `RunIds` to simplify the conversion. E.g. `RunIds.getTime(RunId runId, TimeUnit unit)`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -156,53 +156,48 @@ public ProgramController run(final Program program, ProgramOptions options) {
...
-          error(controller.getFailureCause());
-        }
       }
 
       @Override
-      public void completed() {
-        store.setStop(program.getId(), runId.getId(),
-                      TimeUnit.SECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS),
-                      ProgramController.State.COMPLETED.getRunStatus());
+      public void running() {
```
chtyim: You can extend from `ServiceListenerAdapter` to simplify implementation for no-op method.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -156,53 +156,48 @@ public ProgramController run(final Program program, ProgramOptions options) {
...
       @Override
-      public void suspended() {
-        store.setSuspend(program.getId(), runId.getId());
-      }
-
-      @Override
-      public void resuming() {
-        store.setResume(program.getId(), runId.getId());
+      public void terminated(Service.State from) {
+        if (from != Service.State.STOPPING) {
```
chtyim: Usually it's easier to understand if 'positive' check is used.

```java
if (from == STOPPING) {
} else {
}
```

chtyim:
Just couple comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -156,53 +156,48 @@ public ProgramController run(final Program program, ProgramOptions options) {
...
+                        ProgramController.State.KILLED.getRunStatus());
+        }
       }
 
       @Override
-      public void error(Throwable cause) {
-        LOG.info("MapReduce Program stopped with error {}, {}", program.getId(), runId, cause);
+      public void failed(Service.State from, Throwable failure) {
         store.setStop(program.getId(), runId.getId(),
                       TimeUnit.SECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS),
```
chtyim: I find it simpler and easier to read if you use `TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis())`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
sagarkapare: Added `RunIds.getTime(RunId runId, TimeUnit unit)` mthod to the RunIds class.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
sagarkapare: Using `ServiceListenerAdapter`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
sagarkapare: Changed to used `TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis())`

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/RunIds.java
```diff
@@ -77,12 +78,13 @@ public static RunId generate(long timeMillis) {
   }
 
   /**
-   * @return time in milliseconds from the UUID if it is a time-based UUID, -1 otherwise.
+   * @return time from the UUID if it is a time-based UUID, -1 otherwise.
    */
-  public static long getTimeMillis(RunId runId) {
```
chtyim: Why remove this method? you shouldn't. I was suggesting to add the `getTime`, not to replace this one.

sagarkapare: ohh ok. Will add it. But not sure who will call `getTimeMillis`, as all calls to this method are for setting the store and I thought they will use the `getTime`.   
`AppMetadataStore` was also calling `getTimeMillis` but it was converting the returned value to the seconds.

sagarkapare:
Build passed here: http://builds.cask.co/browse/CDAP-DUT1366-2

chtyim:
LGTM

PR #1938, opened 2015-04-08
move cube and fact table implementation into data fabric

abaranau:
We'll use that in CubeDataset and in MetricStore, so proper place is in data-fabric. Only classes move, nothing is changed


abaranau:
http://builds.cask.co/browse/CDAP-DUT1360 GREEN

gokulavasan:
LGTM :+1:

PR #1940, opened 2015-04-08
Fix/ui home page

ajainarayanan:
Fixes Home view. 
- Removes fonts
- Removes font size, color, weight on headings and other text (paragraph). Uses styles that comes with bootstrap
- Removes most references of line-height. 
- Fixes border-radius to 4px almost everywhere (global styles)
- Removes style changes for h1-h6 headings.


elfenheart:
Seems like when it is empty, the header size is different for Apps vs Data.

screen shot 2015-04-08 at 1 22 33 pm

elfenheart:
:+1: LGTM

elfenheart:
:+1: LGTM

nistala:
LGTM after dry run with modals

PR #1941, opened 2015-04-08
Proper bundle jar building for unit-test

chtyim:
- Added util class AppJarHelper to create bundle jar using ApplicationBundler
- Remove JarFinder, which only able to build uber, not even honoring dependencies

This PR mainly involves moving the createDeploymentJar method from AppFabricClient to AppJarHelper and making all usages of JarFinder to use AppJarHelper instead.


cdap-common/src/main/java/co/cask/cdap/internal/test/AppJarHelper.java
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Helper class for building application jar.
+ */
+public final class AppJarHelper {
+
+  private AppJarHelper() {
+    // No-op
+  }
+
+  public static Location createDeploymentJar(LocationFactory locationFactory, Class<?> clz, Manifest manifest,
```
albertshau: may be better to take location factory in constructor and not have this as a helper class. Don't have a strong opinion about this though.

albertshau:
lgtm

cdap-common/src/main/java/co/cask/cdap/internal/test/AppJarHelper.java
chtyim: That method was moved from `AppFabricClient` to common module so that data-fabric can use it. I didn't intended to change the signature in this PR.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1378-3

PR #1943, opened 2015-04-08
fix flaky metrics query test

gsps1:
this test was failing on one of the builds , reported by @rohitsinha54  - this is a fix. 

we had gauge and increments in the same timestamp and their ordering affects the result, now we gauge after the increments separated by a timestamp and also fixed the assertion.

build running : http://builds.cask.co/browse/CDAP-DUT1380-1

### DIFF
--- a/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsQueryTestRun.java
+++ b/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsQueryTestRun.java
@@ -191,7 +191,8 @@ public class MetricsQueryTestRun extends MetricsSuiteTestBase {
                                                        "counter"));
     flowCollector.increment("reads", 2);
 
-
+    // should wait for increment to happen before gauge, so we keep ordering for test
+    TimeUnit.SECONDS.sleep(1);
     MetricsCollector mrCollector =
       collectionService.getCollector(getMapReduceTaskContext(Constants.DEFAULT_NAMESPACE, "WordCount", "CounterMapRed",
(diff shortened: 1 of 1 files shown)


abaranau:
Hate these sleeps... Created: https://issues.cask.co/browse/CDAP-2042

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsQueryTestRun.java
```diff
@@ -211,7 +212,7 @@ public void testFlowAndAppLevelMetricsAggregation() throws Exception {
     testSingleMetric(flowletMetric2, 2);
     testSingleMetric(flowMetric, 4);
     testSingleMetric(mrMetric, 10);
-    testSingleMetric(appMetric, 14);
+    testSingleMetric(appMetric, 10);
```
abaranau: so it doesn't add them up now?

gsps1:
we will be getting rid of these v2 metrics tests when we remove the v2 endpoints, so not critical in fixing v2 tests, and for v3 tests, we can just use metricstore to add entries at timestamp and avoid sleeps, but agree, yes we need to improve v3 test cases without using sleep.

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsQueryTestRun.java
gsps1: we do a gauge at mr level in the end , so that should be the latest value when we query for reads at app level.

abaranau:
LGTM

gsps1:
Green. Merging.

PR #1945, opened 2015-04-08
Updated code to use HashMap instead of ImmutableMap.Builder

sagarkapare:
In distributed mode the RunId is passed to the programs from AbstractProgramTwillRunnable. So the RunId key is always present in the options. If we use ImmutableMap.Builder to recreate the options in the ProgramRunner, in distributed mode it fails with "duplicate key" exception. There is a bigger refactoring needs to be done around how RunId should be propagated as a part of jira https://issues.cask.co/browse/CDAP-2039.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
@@ -32,10 +32,14 @@ import co.cask.cdap.internal.app.runtime.ProgramRunnerFactory;
 import co.cask.cdap.internal.app.runtime.SimpleProgramOptions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.SettableFuture;
+import jline.internal.Log;
 import org.apache.twill.api.RunId;
 import org.apache.twill.common.Threads;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
```diff
@@ -32,10 +32,14 @@
 import co.cask.cdap.internal.app.runtime.SimpleProgramOptions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.SettableFuture;
+import jline.internal.Log;
```
bdmogal: Remove all log related imports.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
```diff
@@ -54,6 +58,7 @@
  * The {@link RuntimeContext} is blocked until completion of the associated program.
  */
 public abstract class AbstractProgramWorkflowRunner implements ProgramWorkflowRunner {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractProgramWorkflowRunner.class);
```
bdmogal: unused?

bdmogal:
Minor comments, then LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
sagarkapare: My bad. Removed.

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT1381-2

sagarkapare:
Build passed. Merging.

PR #1946, opened 2015-04-08
Feature/cdap 951 explorable table

albertshau:
Adds schema as a Table property. If schema is set on a Table, calls to enable exploration on the Table will create a corresponding Hive table with the given schema.  Schema is applied only at read time with this change, and is not enforced at write time yet.  

As part of the change, refactoring some of the logic around translating a Row into a StructureRecord so that the row key can be given as an argument, and one of the schema fields can come from the row key.


albertshau:
tested on standalone and a hdp2 cluster. In the process of testing on cdh4 and cdh5

albertshau:
http://builds.cask.co/browse/CDAP-DUT1377

abaranau:
please make mergeable and add jira number

albertshau:
tested on cdh4.3 and cdh5.3 as well

cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
```diff
@@ -565,6 +565,16 @@ public boolean isNullableSimple() {
   }
 
   /**
+   * Check if this is a simple type or a nullable simple type, which is a union of a null and one other non-null
+   * simple type, where a simple type is a boolean, int, long, float, double, bytes, or string type.
+   *
+   * @return whether or not this is a nullable simple type.
+   */
+  public boolean isSimpleOrNullableSimple() {
+    return type.isSimpleType() || (type == Type.UNION && getNonNullable().getType().isSimpleType());
```
poornachandra: What if there are more than 2 types in a union?

cdap-common/src/main/java/co/cask/cdap/internal/io/ReflectionRowRecordReader.java
```diff
@@ -16,44 +16,123 @@
...
-      return builder;
+      return builder.build();
     } catch (Exception e) {
       throw propagate(e);
     }
   }
+
+  // translates a row key into some other object.
+  private abstract static class RowKeyFunction {
+    abstract Object get(byte[] rowKey);
```
poornachandra: Would it be a good idea to parametrize this? 

Also, does it need to be abstract class? Can it be an interface?

cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
albertshau: good catch, there is a isNullableSimple() method that it should just call.

cdap-common/src/main/java/co/cask/cdap/internal/io/ReflectionRowRecordReader.java
albertshau: can be an interface, will change it.  Could also be parameterized.

albertshau:
@poornachandra addressed comments, does it look good to merge?

cdap-common/src/main/java/co/cask/cdap/internal/io/ReflectionRowRecordReader.java
```diff
@@ -16,44 +16,123 @@
...
-      return builder;
+      return builder.build();
     } catch (Exception e) {
       throw propagate(e);
     }
   }
+
+  // translates a row key into some other object.
+  private interface RowKeyFunction<T> {
+    T get(byte[] rowKey);
```
poornachandra: `get` is a bit confusing, can we call it `convert`, `translate`, or something similar?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
```diff
@@ -109,14 +110,16 @@ public BufferingTable(String name) {
...
+    this(name, level, false, null, null);
   }
 
   /**
    * Creates an instance of {@link BufferingTable}.
    * @param name table name
    */
-  public BufferingTable(String name, ConflictDetection level, boolean enableReadlessIncrements) {
+  public BufferingTable(String name, ConflictDetection level, boolean enableReadlessIncrements,
+                        Schema schema, String schemaRowField) {
```
poornachandra: It would be a good idea to add `@Nullable` annotation on `schema` and `schemaRowField`

poornachandra: `schemaRowField` should be renamed as `rowFieldName`, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/ObjectMappedTableDefinition.java
```diff
@@ -90,16 +94,28 @@ public DatasetAdmin getAdmin(DatasetContext datasetContext, DatasetSpecification
...
                                                 ClassLoader classLoader) throws IOException {
+    String keyName = ObjectMappedTableProperties.getRowKeyExploreName(spec.getProperties());
+
     DatasetSpecification tableSpec = spec.getSpecification(TABLE_NAME);
+    // if the table spec did not have schema, this is an ObjectMappedTable from CDAP 2.8.
+    // add the schema and row key as arguments so that explore will work
+    if (!tableSpec.getProperties().containsKey(Table.PROPERTY_SCHEMA)) {
+      tableSpec = DatasetSpecification.builder(tableSpec.getName(), tableSpec.getType())
+        .properties(tableSpec.getProperties())
+        .property(Table.PROPERTY_SCHEMA, spec.getProperty(Table.PROPERTY_SCHEMA))
```
poornachandra: Will schema be added to old `ObjectMappedTable` during upgrade?

albertshau: created CDAP-2122, will do it separately

albertshau:
@poornachandra  addressed comments, please take another look

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
```diff
@@ -156,12 +157,11 @@ public QueryHandle enableDataset(Id.DatasetInstance datasetID, DatasetSpecificat
     String datasetType = spec.getType();
     // special casing here... but we really should clean this up
     // there are two ways to refer to each dataset type...
```
poornachandra: This comment should probably move with the check.

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
```diff
@@ -204,6 +204,13 @@ public QueryHandle enableDataset(Id.DatasetInstance datasetID, DatasetSpecificat
     }
   }
 
+  private boolean isExplorableTable(String datasetType) {
+    return ObjectMappedTableModule.FULL_NAME.equals(datasetType) ||
+      ObjectMappedTableModule.SHORT_NAME.equals(datasetType) ||
+      "table".equals(datasetType) ||
+      Table.class.getName().equals(datasetType);
```
poornachandra: This PR enables more tables like `IndexedTable`, etc. to have schema so that they can be explored, right? Should we say any dataset that implements Table and has a schema is explorable?

poornachandra:
@albertshau Final couple of comments, please address and merge :+1:

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
albertshau: good point, will change

PR #1947, opened 2015-04-08
(CDAP-2043) Make AvroRecordFormat decodes Avro binary data into StructuredRecord

chtyim:
We never expose `GenericRecord` class through Program ClassLoader, hence having the record format emit `GenericRecord` would make it unusable from the user program (there will be class cast exception if avro jar is bundled with user program jar).

### DIFF
--- a/cdap-data-fabric/src/test/java/co/cask/cdap/data/format/AvroRecordFormatTest.java
+++ b/cdap-data-fabric/src/test/java/co/cask/cdap/data/format/AvroRecordFormatTest.java
@@ -20,10 +20,12 @@ import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.data.format.FormatSpecification;
 import co.cask.cdap.api.data.format.Formats;
 import co.cask.cdap.api.data.format.RecordFormat;
+import co.cask.cdap.api.data.format.StructuredRecord;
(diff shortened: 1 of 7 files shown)


cdap-data-fabric/src/test/java/co/cask/cdap/data/format/AvroRecordFormatTest.java
```diff
@@ -210,4 +231,17 @@ private StreamEvent toStreamEvent(GenericRecord record, boolean writeSchema) thr
 
     return new StreamEvent(headers, ByteBuffer.wrap(serializedRecord));
   }
+//
```
henry-cask: Any reason why we keeping the comment out code?

albertshau:
lgtm after removing commented code that henry pointed out

cdap-data-fabric/src/test/java/co/cask/cdap/data/format/AvroRecordFormatTest.java
chtyim: Oh, that was a mistake. Good catch.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1384-1

Squashed commits into one.

PR #1949, opened 2015-04-08
Ditches angular bootstrap accordion and uses a panel group in simpler way

ajainarayanan:
Removes the use of angular bootstrap accordion and uses a simple panel-group to achieve the same thing.

TODO: 
 - Use the same in other program types too.
 - Enable injecting through streams in flows.


elfenheart:
:+1: LGTM

PR #1950, opened 2015-04-08
Fix/ui flows live metric

elfenheart:
flowlet diagram now have live metrics. Status page on flows also have live metrics.


ajainarayanan:
Except for console.log statements it LGTM :+1:

PR #1951, opened 2015-04-08
CDAP-2047: First step: Cleanup MetricsTable interface of all unused methods

abaranau:
Cleaning up MetricsTable interface makes it easier to implement (CDAP-2047) and later integrate into Table (CDAP-1193)


abaranau:
http://builds.cask.co/browse/CDAP-DUT1387-1

abaranau:
GREEN

anwar6953:
whats the replacement for batch delete?

abaranau:
it is not used now: see todo "+        // todo: do deletes efficiently, in batches, not one-by-one".

It will not be used as it was - via scan + filter, but rather it will be multi-delete

abaranau:
so old goes away anyways, as unused

anwar6953:
lgtm

PR #1878, opened 2015-04-02
Feature/mr task info endpoint

anwar6953:
JIRA: https://issues.cask.co/browse/CDAP-1958
BUILD: http://builds.cask.co/browse/CDAP-DUT1324-9

Opening PR to get feedback.

This endpoint returns job-level and task-level information for a particular MapReduce runId.
It does this by relaying information from Hadoop Job History Server, to avoid duplicating this logic in CDAP.
If JHS is not available or if there is failure to communicate with it, the endpoint returns 503 (Service Not Available). **This is up for discussion**.

Because there is no Task-level granularity in Local MR Runner, and because there is not a JHS available in Standalone, this endpoint will return Service Not Available (503) in Standalone.

The endpoint does not currently return metrics related to the job (from MetricsStore) if there is failure to communicate with the Job History Server.

Further improvements can be made:
- Audi Tests (because this functionality does not work in Standalone/InMemory) [Completed]
- Persist a (runId --> jobId) mapping to use for lookups.
- Better way of detecting that JHS is not available than catching IOE.

Example Output (for a completed MR Job with 8 mapper tasks and 3 reducer tasks).
Note that it is only returning counters from the TaskCounter group (we can probably simplify/filter the returned counters further).
```
{  
   "state":"SUCCEEDED",
   "startTime":1427852454941,
   "finishTime":1427852549292,
   "mapProgress":0.0,
   "reduceProgress":0.0,
   "counters":{  
      "MAP_OUTPUT_MATERIALIZED_BYTES":19344,
      "MAP_INPUT_RECORDS":20,
      "MERGED_MAP_OUTPUTS":24,
      "REDUCE_SHUFFLE_BYTES":19344,
      "SPILLED_RECORDS":400,
      "MAP_OUTPUT_BYTES":18800,
      "COMMITTED_HEAP_BYTES":5490343936,
      "CPU_MILLISECONDS":122570,
      "FAILED_SHUFFLE":0,
      "SPLIT_RAW_BYTES":1077,
      "COMBINE_INPUT_RECORDS":0,
      "REDUCE_INPUT_RECORDS":200,
      "REDUCE_INPUT_GROUPS":1,
      "COMBINE_OUTPUT_RECORDS":0,
      "PHYSICAL_MEMORY_BYTES":5830778880,
      "REDUCE_OUTPUT_RECORDS":1,
      "VIRTUAL_MEMORY_BYTES":15434371072,
      "MAP_OUTPUT_RECORDS":200,
[...]

PR #1955, opened 2015-04-08
CDAP-1958 Job-level and Task-level information (historical and current) ...

anwar6953:
...via Metrics system (in case retrieving from jobHistoryServer fails).

The implementation in this PR uses JobHistoryServer: https://github.com/caskdata/cdap/pull/1878

This PR is the fallback (via Metrics system) if the JHS is not available. It doesn't return all the same data (it is missing fields such as status/startime/endtime as well as majority of the counters).

https://issues.cask.co/browse/CDAP-1958
http://builds.cask.co/browse/CDAP-DUT1324-16

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
@@ -14,7 +14,7 @@
  * the License.
  */
 
-package co.cask.cdap.gateway.handlers;
+package co.cask.cdap.app.mapreduce;
 
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
@@ -22,6 +22,8 @@ import co.cask.cdap.common.exception.NotFoundException;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.MRJobInfo;
 import co.cask.cdap.proto.MRTaskInfo;
+import co.cask.cdap.proto.ProgramType;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.inject.Inject;
@@ -53,7 +55,7 @@ public class MRJobClient {
     int numRetries = cConf.getInt(Constants.AppFabric.MAPREDUCE_JOB_CLIENT_CONNECT_MAX_RETRIES);
     this.hConf = new Configuration(hConf);
     // Override a cloned hConf's configuration of IPC Client max retries based upon value in CConf to avoid longer
-    // amounts of retrying (especially when the Job History Server is not installed)
+    // amounts of retrying (this is helpful especially when the Job History Server is not installed)
     this.hConf.setInt(CommonConfigurationKeysPublic.IPC_CLIENT_CONNECT_MAX_RETRIES_KEY, numRetries);
   }
 
@@ -64,6 +66,8 @@ public class MRJobClient {
    * @throws NotFoundException if a Job with the given runId is not found.
    */
   public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundException {
+    Preconditions.checkArgument(ProgramType.MAPREDUCE.equals(runId.getProgram().getType()));
+
     JobClient jobClient;
     JobStatus[] jobs;
     try {
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
@@ -1,0 +1,191 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.app.mapreduce;
+
+import co.cask.cdap.api.dataset.lib.cube.TimeValue;
+import co.cask.cdap.api.metrics.MetricDataQuery;
+import co.cask.cdap.api.metrics.MetricStore;
+import co.cask.cdap.api.metrics.MetricTimeSeries;
+import co.cask.cdap.api.metrics.MetricType;
+import co.cask.cdap.app.metrics.MapReduceMetrics;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.MRJobInfo;
+import co.cask.cdap.proto.MRTaskInfo;
+import co.cask.cdap.proto.ProgramType;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.HashBasedTable;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Table;
+import com.google.inject.Inject;
+import org.apache.hadoop.mapreduce.TaskCounter;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Retrieves information/reports for a MapReduce run via the Metrics system.
+ */
+public class MapReduceMetricsInfo {
+
+  private final MetricStore metricStore;
+
+  @Inject
+  public MapReduceMetricsInfo(MetricStore metricStore) {
+    this.metricStore = metricStore;
+  }
+
+  /**
+   * @param runId for which information will be returned.
+   * @return a {@link MRJobInfo} containing information about a particular MapReduce program run.
+   */
+  public MRJobInfo getMRJobInfo(Id.Run runId) throws Exception {
+    Preconditions.checkArgument(ProgramType.MAPREDUCE.equals(runId.getProgram().getType()));
+
+    // baseTags has tag keys: ns.app.mr.runid
+    Map<String, String> baseTags = Maps.newHashMap();
+    baseTags.put(Constants.Metrics.Tag.NAMESPACE, runId.getNamespace().getId());
+    baseTags.put(Constants.Metrics.Tag.APP, runId.getProgram().getApplicationId());
+    baseTags.put(Constants.Metrics.Tag.MAPREDUCE, runId.getProgram().getId());
+    baseTags.put(Constants.Metrics.Tag.RUN_ID, runId.getId());
+
+    Map<String, String> mapTags = Maps.newHashMap(baseTags);
+    mapTags.put(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Mapper.getId());
+
+    Map<String, String> reduceTags = Maps.newHashMap(baseTags);
+    reduceTags.put(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Reducer.getId());
+
+    // map from RunId -> (CounterName -> CounterValue)
+    Table<String, String, Long> mapTaskMetrics = HashBasedTable.create();
+    Table<String, String, Long> reduceTaskMetrics = HashBasedTable.create();
+
+
+    // Populate mapTaskMetrics and reduce Task Metrics via MetricStore. Used to construct MRTaskInfo below.
+    // Use batch-querying when it is available on the MetricStore. https://issues.cask.co/browse/CDAP-2045
+    putMetrics(queryGroupedAggregates(mapTags, MapReduceMetrics.METRIC_INPUT_RECORDS),
+               mapTaskMetrics, TaskCounter.MAP_INPUT_RECORDS.name());
+    putMetrics(queryGroupedAggregates(mapTags, MapReduceMetrics.METRIC_OUTPUT_RECORDS),
+               mapTaskMetrics, TaskCounter.MAP_OUTPUT_RECORDS.name());
+    putMetrics(queryGroupedAggregates(mapTags, MapReduceMetrics.METRIC_BYTES),
+               mapTaskMetrics, TaskCounter.MAP_OUTPUT_BYTES.name());
+
+    Map<String, Long> mapProgress = queryGroupedAggregates(mapTags, MapReduceMetrics.METRIC_TASK_COMPLETION);
+
+    putMetrics(queryGroupedAggregates(reduceTags, MapReduceMetrics.METRIC_INPUT_RECORDS),
+               reduceTaskMetrics, TaskCounter.REDUCE_INPUT_RECORDS.name());
+    putMetrics(queryGroupedAggregates(reduceTags, MapReduceMetrics.METRIC_OUTPUT_RECORDS),
+               reduceTaskMetrics, TaskCounter.REDUCE_OUTPUT_RECORDS.name());
+
+    Map<String, Long> reduceProgress = queryGroupedAggregates(reduceTags, MapReduceMetrics.METRIC_TASK_COMPLETION);
+
+    // Construct MRTaskInfos from the information we can get from Metric system.
+    List<MRTaskInfo> mapTaskInfos = Lists.newArrayList();
+    for (Map.Entry<String, Map<String, Long>> taskEntry : mapTaskMetrics.rowMap().entrySet()) {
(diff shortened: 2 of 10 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobMetricsReporter.java
```diff
@@ -0,0 +1,235 @@
...
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Retrieves information/reports for a MapReduce run via the Metrics system.
+ */
+public class MRJobMetricsReporter {
```
abaranau: rename please. It does not report metrics, it queries metrics for MR Job info

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobMetricsReporter.java
```diff
@@ -0,0 +1,235 @@
...
+    baseTagValues.add(new TagValue(Constants.Metrics.Tag.MAPREDUCE, runId.getProgram().getId()));
+    baseTagValues.add(new TagValue(Constants.Metrics.Tag.RUN_ID, runId.getId()));
+
+    List<TagValue> mapTagValues = Lists.newArrayList(baseTagValues);
+    mapTagValues.add(new TagValue(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Mapper.getId()));
+
+    List<TagValue> reduceTagValues = Lists.newArrayList(baseTagValues);
+    reduceTagValues.add(new TagValue(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Reducer.getId()));
+
+    List<String> mapTaskIds = getTaskIds(mapTagValues);
```
abaranau: these two are redundant...

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -0,0 +1,217 @@
...
+    putMetrics(queryGroupedAggregates(mapperContext, MapReduceMetrics.METRIC_INPUT_RECORDS),
+               mapTaskMetrics, TaskCounter.MAP_INPUT_RECORDS.name());
+    putMetrics(queryGroupedAggregates(mapperContext, MapReduceMetrics.METRIC_OUTPUT_RECORDS),
+               mapTaskMetrics, TaskCounter.MAP_OUTPUT_RECORDS.name());
+    putMetrics(queryGroupedAggregates(mapperContext, MapReduceMetrics.METRIC_BYTES),
+               mapTaskMetrics, TaskCounter.MAP_OUTPUT_BYTES.name());
+
+    Map<String, Long> mapProgress = queryGroupedAggregates(mapperContext, MapReduceMetrics.METRIC_TASK_COMPLETION);
+
+
```
abaranau: 2 empty lines?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -0,0 +1,217 @@
...
+    Map<String, String> reducerContext = tagValuesToMap(reduceTagValues);
+    putMetrics(queryGroupedAggregates(reducerContext, MapReduceMetrics.METRIC_INPUT_RECORDS),
+               reduceTaskMetrics, TaskCounter.REDUCE_INPUT_RECORDS.name());
+    putMetrics(queryGroupedAggregates(reducerContext, MapReduceMetrics.METRIC_OUTPUT_RECORDS),
+               reduceTaskMetrics, TaskCounter.REDUCE_OUTPUT_RECORDS.name());
+
+    Map<String, Long> reduceProgress = queryGroupedAggregates(reducerContext, MapReduceMetrics.METRIC_TASK_COMPLETION);
+
```
abaranau: 2 empty lines?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -0,0 +1,217 @@
...
+    baseTagValues.add(new TagValue(Constants.Metrics.Tag.RUN_ID, runId.getId()));
+
+    List<TagValue> mapTagValues = Lists.newArrayList(baseTagValues);
+    mapTagValues.add(new TagValue(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Mapper.getId()));
+
+    List<TagValue> reduceTagValues = Lists.newArrayList(baseTagValues);
+    reduceTagValues.add(new TagValue(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Reducer.getId()));
+
+    // map from RunId -> (CounterName -> CounterValue)
+    Map<String, Map<String, Long>> mapTaskMetrics = Maps.newHashMap();
```
abaranau: Will it help to use Table?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -0,0 +1,217 @@
...
+    Map<String, Map<String, Long>> reduceTaskMetrics = Maps.newHashMap();
+
+    // Populate mapTaskMetrics and reduce Task Metrics via MetricStore. Used to construct MRTaskInfo below.
+    Map<String, String> mapperContext = tagValuesToMap(mapTagValues);
```
abaranau: add a todo to do in one query with link to CDAP-2045, so that we don't forget

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -0,0 +1,217 @@
...
+  private MRJobInfo getJobCounters(Map<String, String> jobContext, List<MRTaskInfo> mapTaskInfos,
+                                   List<MRTaskInfo> reduceTaskInfos) throws Exception {
+    Map<String, String> mapContext = Maps.newHashMap(jobContext);
+    mapContext.put(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Mapper.getId());
+    Map<String, String> reduceContext = Maps.newHashMap(jobContext);
+    reduceContext.put(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Reducer.getId());
+
+    HashMap<String, Long> metrics = Maps.newHashMap();
+
+    metrics.put(TaskCounter.MAP_INPUT_RECORDS.name(),
```
abaranau: add a todo to do in one query with link to CDAP-2045, so that we don't forget

abaranau:
Just couple minor comments

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
```diff
@@ -53,7 +55,7 @@ public MRJobClient(CConfiguration cConf, Configuration hConf) {
     int numRetries = cConf.getInt(Constants.AppFabric.MAPREDUCE_JOB_CLIENT_CONNECT_MAX_RETRIES);
```
gsps1: do you need a default here , if that parameter is not set ?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -0,0 +1,217 @@
...
+    MetricDataQuery metricDataQuery =
+      new MetricDataQuery(0, Integer.MAX_VALUE, Integer.MAX_VALUE, prependSystem(metric), MetricType.GAUGE, context,
+                          ImmutableList.of(Constants.Metrics.Tag.INSTANCE_ID));
+    Collection<MetricTimeSeries> query = metricStore.query(metricDataQuery);
+
+    // runId -> metricValue
+    Map<String, Long> taskMetrics = Maps.newHashMap();
+    for (MetricTimeSeries metricTimeSeries : query) {
+      List<TimeValue> timeValues = metricTimeSeries.getTimeValues();
+      TimeValue timeValue = Iterables.getOnlyElement(timeValues);
```
gsps1: do you want to throw the exception if no element is present or (log and skip) ?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+   * @param runId for which information will be returned.
+   * @return a {@link MRJobInfo} containing information about a particular MapReduce program run.
+   */
+  public MRJobInfo getMRJobInfo(Id.Run runId) throws Exception {
+    Preconditions.checkArgument(ProgramType.MAPREDUCE.equals(runId.getProgram().getType()));
+
+    // baseTagValues has tag keys: ns.app.mr.runid
+    List<TagValue> baseTagValues = Lists.newArrayList();
```
gsps1: this can be a Map right? i think you were searching before and that's why you needed list then.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -223,23 +228,29 @@ public void mapReduceInfo(HttpRequest request, HttpResponder responder,
...
       }
       if (!appSpec.getMapReduce().containsKey(mapreduceId)) {
-        responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Program not found: %s", programId));
-        return;
+        throw new NotFoundException(programId);
       }
+//      if (store.getRun(programId, runId) == null) {
+//        throw new NotFoundException(run);
+//      }
+
```
gsps1: remove?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -223,23 +228,29 @@ public void mapReduceInfo(HttpRequest request, HttpResponder responder,
...
+
+      MRJobInfo mrJobInfo;
+      try {
+        mrJobInfo = mrJobClient.getMRJobInfo(run);
+      } catch (IOException ioe) {
+        LOG.warn("Failed to get run history for runId from JobClient: {}. Falling back to Metrics system.");
+        mrJobInfo = mapReduceMetricsInfo.getMRJobInfo(run);
```
gsps1: this can throw an exception too, catch block for this ?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
anwar6953: the default is in cdap-default

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
```diff
@@ -64,6 +66,8 @@ public MRJobClient(CConfiguration cConf, Configuration hConf) {
    * @throws NotFoundException if a Job with the given runId is not found.
    */
   public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundException {
+    Preconditions.checkArgument(ProgramType.MAPREDUCE.equals(runId.getProgram().getType()));
```
gsps1: can workflow call this method?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anwar6953: Nah, I just need to uncomment this.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anwar6953: The outer catch block will get this. And log and respond 500.
The inner catch block is just so that we can fall back to mapReduceMetricsInfo.
Does that make sense?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
anwar6953: Yea. I removed the search, but didn't see the need to change this.
Will change to map now.

anwar6953: This will also allow me to remove `tagValuesToMap` method

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
anwar6953: This should always be true right? (unless there's a problem with the MetricStore, because I'm asking for aggregate).
So probably throw an exception if it isn't true.
What do you think?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
anwar6953: The runId of the workflow wouldn't be found from the JobClient.
Would have to use the MR's runId (even if its in a workflow).

cdap-app-fabric/src/test/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfoTest.java
```diff
@@ -0,0 +1,166 @@
...
+    injector = startMetricsService(conf);
+    metricStore = injector.getInstance(MetricStore.class);
+  }
+
+  public static Injector startMetricsService(CConfiguration conf) {
+    Injector injector = Guice.createInjector(new AppFabricTestModule(conf));
+    injector.getInstance(TransactionManager.class).startAndWait();
+    injector.getInstance(DatasetOpExecutor.class).startAndWait();
+    injector.getInstance(DatasetService.class).startAndWait();
+    injector.getInstance(MetricsQueryService.class).startAndWait();
```
gsps1: you dont need this service right?

anwar6953: right. removed.

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
abaranau: so if no metric, we throw exception? not good

abaranau:
one more comment - shankar's I think is pretty important

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
anwar6953: if no metric, there will be no item in the collection.

abaranau: oh, ok

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
anwar6953: I don't think Table is needed. This isn't too complex, (just a map from taskId to its metrics).
And Table might bring more complexity.

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -0,0 +1,190 @@
...
+    baseTags.put(Constants.Metrics.Tag.RUN_ID, runId.getId());
+
+    Map<String, String> mapTags = Maps.newHashMap(baseTags);
+    mapTags.put(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Mapper.getId());
+
+    Map<String, String> reduceTags = Maps.newHashMap(baseTags);
+    reduceTags.put(Constants.Metrics.Tag.MR_TASK_TYPE, MapReduceMetrics.TaskType.Reducer.getId());
+
+    // map from RunId -> (CounterName -> CounterValue)
+    HashBasedTable<String, String, Long> mapTaskMetrics = HashBasedTable.create();
```
abaranau: Use Table interface as var type

abaranau:
tiny comment and feel free to merge

PR #1956, opened 2015-04-09
Fix/ui apps overview columns switch

ajainarayanan:


elfenheart:
:+1: LGTM

cdap-ui/app/features/admin/templates/namespace/app-metadata.html
```diff
@@ -15,7 +15,7 @@ <h3 ng-bind="apps.name | caskCapitalizeFilter"></h3>
     </div>
 </div>
 
-<table class="table table-responsive" cask-sortable>
+<table class="table table-responsive">
```
elfenheart: we don't want cask-sortable here?

elfenheart:
:+1: LGTM

PR #1957, opened 2015-04-09
Fix/ui dashboard without default widget

elfenheart:
no Welcome default widget when creating new Dashboard.

Navigate to the newly created Dashboard.


ajainarayanan:
:+1: LGTM

PR #1958, opened 2015-04-09
[CDAP-1787] Added 'adapter' tag to mapreduce metrics when run through adapters

bdmogal:
When a mapreduce is run through an adapter, you should be able to query its metrics using tags: *namespace* and *adapter* as well as *namespace*, *app*, *mapreduce*

Jira: https://issues.cask.co/browse/CDAP-1787
Build: https://builds.cask.co/browse/CDAP-DUT1389

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
@@ -82,7 +82,7 @@ public class BasicMapReduceContext extends AbstractContext implements MapReduceC
                                DatasetFramework dsFramework,
                                @Nullable String adapterName) {
     super(program, runId, runtimeArguments, datasets,
-          getMetricCollector(metricsCollectionService, program, type, runId.getId(), taskId),
+          getMetricCollector(program, runId.getId(), taskId, metricsCollectionService, type, adapterName),
           dsFramework, discoveryServiceClient);
     this.logicalStartTime = logicalStartTime;
     this.workflowBatch = workflowBatch;
@@ -241,8 +241,11 @@ public class BasicMapReduceContext extends AbstractContext implements MapReduceC
     this.reducerResources = resources;
   }
 
-  private static MetricsCollector getMetricCollector(MetricsCollectionService service, Program program,
-                                                     MapReduceMetrics.TaskType type, String runId, String taskId) {
+  @Nullable
+  private static MetricsCollector getMetricCollector(Program program, String runId, String taskId,
+                                                     @Nullable MetricsCollectionService service,
+                                                     @Nullable MapReduceMetrics.TaskType type,
+                                                     @Nullable String adapterName) {
     if (service == null) {
       return null;
     }
@@ -257,6 +260,10 @@ public class BasicMapReduceContext extends AbstractContext implements MapReduceC
     } else {
       // in a runner (container that submits the job): put program info
       tags.putAll(getMetricsContext(program, runId));
+    }
+
+    if (adapterName != null) {
+      tags.put(Constants.Metrics.Tag.ADAPTER, adapterName);
     }
 
     return service.getCollector(tags);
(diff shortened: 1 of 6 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -242,7 +242,8 @@ public void setReducerResources(Resources resources) {
   }
 
   private static MetricsCollector getMetricCollector(MetricsCollectionService service, Program program,
-                                                     MapReduceMetrics.TaskType type, String runId, String taskId) {
+                                                     MapReduceMetrics.TaskType type, String runId, String taskId,
```
albertshau: not introduced by you, but some of these other arguments are nullable so we should be consistent and annotate them.

albertshau:
minor comment, otherwise lgtm.  Is there another jira open for workers and adapters?

bdmogal:
No jira yet, but I was thinking I'll keep the same jira open until I add metrics for workers. I wasn't sure yet that we have all the support needed for testing adapter metrics in workers

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
bdmogal: I'll annotate them.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
```diff
@@ -158,6 +158,12 @@ public Cube get() {
       ImmutableList.of(Constants.Metrics.Tag.NAMESPACE, Constants.Metrics.Tag.APP,
                        Constants.Metrics.Tag.SPARK)));
 
+    // batch adapters - can only contain mapreduce for now
+    aggs.add(new DefaultAggregation(
+      ImmutableList.of(Constants.Metrics.Tag.NAMESPACE, Constants.Metrics.Tag.ADAPTER),
```
anwar6953: Shouldn't we aggregate by RunId (like the other aggregations)? Or is there no notion of a runId for adapters?
(A particular instance of an adapter type).

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
```diff
@@ -168,11 +174,15 @@ public void testSearchWithTags() throws Exception {
...
                        getSearchResultExpected("flow", "WCounter",
-                                                "flow", "WordCounter",
-                                                "mapreduce", "ClassicWordCount",
-                                                "procedure", "RCounts"
+                                               "flow", "WordCounter",
+                                               "mapreduce", "ClassicWordCount",
+                                               "procedure", "RCounts"
                        ));
 
+    // Child context for adapters should be empty
```
gsps1: we don't use child context terminology anymore, so just say, no tags under adapter

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
gsps1: @anwar6953  yeah, not even program-type, its aggregated and queried only at adapter level. @bdmogal can give more details.

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
```diff
@@ -259,6 +274,10 @@ public void testSearchContext() throws Exception {
     verifySearchResult("/v3/metrics/search?target=childContext&context=namespace.myspace.app.WCount1",
                        ImmutableList.<String>of());
 
+    // no child context for adapters
```
gsps1: same

gsps1:
just minor nits, LGTM otherwise. :+1:

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
bdmogal: this one is ok, since this tests for the old API.

bdmogal:
Fixed nits. Will merge after a clean run. I had to fix ``MetricsHandlerTestRun``, so @gsps1 if you can take a look at the last commit again, that would be great.

bdmogal:
Build passed, merging.

PR #1959, opened 2015-04-09
CDAP-1756 extend the ETL template api to add a configurePipeline

albertshau:
method to sources and sinks, which will let them add streams
and datasets that should be created when the pipeline is created.


gokulavasan:
LGTM :+1:

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/config/ETLConfig.java
```diff
@@ -14,23 +14,22 @@
...
 import com.google.common.base.Objects;
 
 import java.util.List;
 
 /**
- * ETL Batch Adapter Configuration.
+ * ETL Adapter Configuration.
  */
-public class ETLBatchConfig {
+public final class ETLConfig {
```
gokulavasan: Why the class name change to ETLConfig? Since we have schedule as a field, wouldn't this be specific to a Batch adapter?

gokulavasan:
Minor comment about class name change. Rest LGTM :+1:

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/config/ETLConfig.java
albertshau: oo ok thats why it was in batch. I will move it back

gokulavasan:
@albertshau I think the PR deletes ETLBatchConfig.java. Is that right? Can you also include a link to the build?

albertshau:
ah sorry didn't add it back. Added it, build running at http://builds.cask.co/browse/CDAP-DT8

gokulavasan:
LGTM :+1:

PR #1961, opened 2015-04-09
(CDAP-1887) Fix MR ClassLoader

chtyim:
- Manipulate the MR job container classpath so that the CDAP classes get higher precedence
- Use ApplicationClassLoader, which is of same class name as the one used by MR framework to trick MR framework to use our implementation
- ApplicationClassLoader do classloading in the order of [ProgramClassLoader, CDAP system ClassLoader (which includes Hadoop classes there)]
- For local mode, simply set the ApplicationClassLoader to the configuration and use it in all MR threads
- Get rid of the unnecessary Input/OutputFormatWrapper classes that existed for manipulating ClassLoader before


chtyim:
Tested in both Standalone and on cluster.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextProvider.java
```diff
@@ -98,46 +99,43 @@ private synchronized AbstractMapReduceContextBuilder getBuilder(CConfiguration c
...
-    // It is needed so that in distributed mode, there is only one ProgramClassLoader created, even if
-    // there are multiple instances of this class being created (one from creating RecordReader/Writer, one from
-    // MapperWrapper/ReducerWrapper).
-
-    // For local mode, it doesn't matter than much, but no harm in returning the same program.
-    Program program = PROGRAM_THREAD_LOCAL.get();
-    if (program != null) {
-      return program;
+  private Program createProgram(MapReduceContextConfig contextConfig) {
+    LocationFactory locationFactory = new LocalLocationFactory();
```
henry-cask: Could we just hang this final member variable? I assume the value will be the same and no state would be change during the lifetime of the context provider?

chtyim: We could, although this method only get called once or twice per container.

cdap-app-fabric/src/main/java/org/apache/hadoop/yarn/util/ApplicationClassLoader.java
```diff
@@ -0,0 +1,168 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
```
chtyim: I am not sure if we need to keep this license header since we changed the class entirely. @henry-cask thoughts?

henry-cask: No we do not have to keep the duplicate license header because our "derivative" work is under the same ASF 2.0 license.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextProvider.java
henry-cask: Should be fine then

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1374-9

Comments addressed.

henry-cask:
LGTM +1

PR #1962, opened 2015-04-09
Starting Worker for Realtime Adapters

gokulavasan:
Implementing Worker Adapters. 

Starting an Adapter which corresponds to a Worker ProgramType will start a run of that specific Worker program with the runtime arguments from the AdapterSpecification. It will also persist the information of adapter name -> RunId map in AdapterStore. This is required in order to stop the adapter which can be achieved only by knowing the RunId.

Also, a preliminary work on ProgramLifecycleService is part of this PR. This will abstract away the logic associated with starting a program and thus can be used in different places where Programs are started in CDAP - ProgramLifecycleHttpHandler, ScheduleTaskRunner, AdapterService.

Bamboo : http://builds.cask.co/browse/CDAP-DUT1375-9
JIRA : https://issues.cask.co/browse/CDAP-2050


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterServiceTests.java
```diff
@@ -74,18 +75,49 @@ public void testInvalidAdapter() throws Exception {
...
+    // Stop Adapter
+    adapterService.stopAdapter(NAMESPACE, adapter1);
+    adapterService.stopAdapter(NAMESPACE, adapter2);
+
+    Assert.assertEquals(AdapterStatus.STOPPED, adapterService.getAdapterStatus(NAMESPACE, adapter1));
+    Assert.assertEquals(AdapterStatus.STOPPED, adapterService.getAdapterStatus(NAMESPACE, adapter2));
+
+    // Delete Adapter
+    adapterService.removeAdapter(NAMESPACE, adapter1);
+    adapterService.removeAdapter(NAMESPACE, adapter2);
```
albertshau: verify they are deleted after removing

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -0,0 +1,187 @@
...
+
+  @Override
+  protected void shutDown() throws Exception {
+    LOG.info("Shutting down ProgramLifecycleService");
+  }
+
+  private Program getProgram(Id.Program id, ProgramType programType) throws IOException {
+    Program program = store.loadProgram(id, programType);
+    if (program == null) {
+      throw new FileNotFoundException(String.format("Program not found: Id = %s; Type = %s", id, programType));
```
albertshau: is this supposed to be a FileNotFound?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -373,12 +388,75 @@ private void stopWorkflowAdapter(Id.Namespace namespace, AdapterSpecification ad
     store.deleteSchedule(workflowId, SchedulableProgramType.WORKFLOW, scheduleName);
   }
 
-  private void startWorkerAdapter(Id.Namespace namespace, AdapterSpecification adapterSpec) {
-    // TODO: implement
+  private Id.Program getWorkerId(Id.Namespace namespace, AdapterSpecification adapterSpec) throws NotFoundException {
+    Id.Application appId = Id.Application.from(namespace, adapterSpec.getTemplate());
+    ApplicationSpecification appSpec = store.getApplication(appId);
+    if (appSpec == null || appSpec.getWorkers().size() != 1) {
```
albertshau: these 2 errors are different and should be handled differently i think.  though how can the second case happen?  It seems like if it happens it's a bug in our code?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterStore.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.inject.Inject;
+import org.apache.twill.api.RunId;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Store for Adapter related information.
+ */
+//TODO: Combine simple ConfigStore abstractions together.
+public class AdapterStore {
```
albertshau: dont really like that state is split up into multiple stores.  It makes it harder to make sure things are consistent and correct.  I think it would be better to store this information where status is stored.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -373,12 +388,75 @@ private void stopWorkflowAdapter(Id.Namespace namespace, AdapterSpecification ad
...
+    String workerName = workerId.getId();
+    try {
+      Map<String, String> sysArgs = resolver.getSystemProperties(workerId, ProgramType.WORKER);
+      Map<String, String> userArgs = resolver.getUserProperties(workerId, ProgramType.WORKER);
+      // Override resolved preferences with adapter worker spec properties.
+      userArgs.putAll(adapterSpec.getRuntimeArgs());
+      store.setWorkerInstances(workerId, adapterSpec.getInstances());
+      ProgramRuntimeService.RuntimeInfo runtimeInfo = lifecycleService.start(workerId, ProgramType.WORKER,
+                                                                             sysArgs, userArgs, false);
+      adapterStore.setRunId(adapterId, runtimeInfo.getController().getRunId());
```
albertshau: not introduced by you, but we can have races where requests to start the same program are made concurrently and then things break.  should probably have some sort of check and set logic on program status before we do anything.  Can you open a jira and add a todo somewhere?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -373,12 +388,75 @@ private void stopWorkflowAdapter(Id.Namespace namespace, AdapterSpecification ad
...
+      final ProgramController controller = runtimeInfo.getController();
+      controller.addListener(new AbstractListener() {
+        @Override
```
albertshau: i think we'll also want to override init and set the adapter status there instead of afterwards to reduce the chance of a race.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterStore.java
albertshau: actually, how is current run information stored?  ideally we would be able to get run history for an adapter as well instead of just storing the current runid only for workers.

gokulavasan:
@albertshau Updated the PR. Addressed your comments. Please review. 
I have filed https://issues.cask.co/browse/CDAP-2112. Once that is resolved, I will add a test for RealtimeAdapterLifecycle to AdapterLifecycleTests.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1416,88 +1414,21 @@ private AppFabricServiceStatus start(final Id.Program id, ProgramType type,
...
-      }
-      return AppFabricServiceStatus.OK;
-    } catch (DatasetInstantiationException e) {
-      return new AppFabricServiceStatus(HttpResponseStatus.UNPROCESSABLE_ENTITY, e.getMessage());
+      ProgramRuntimeService.RuntimeInfo runtimeInfo = lifecycleService.start(id, type, sysArgs, userArgs, debug);
+      return (runtimeInfo != null) ? AppFabricServiceStatus.OK : AppFabricServiceStatus.INTERNAL_ERROR;
     } catch (Throwable throwable) {
       LOG.error(throwable.getMessage(), throwable);
-      if (throwable instanceof FileNotFoundException) {
+      if (throwable instanceof ProgramNotFoundException) {
```
albertshau: should just have a separate catch for this exception instead of checking the type here. Also be sure not to log an error for this case.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -379,12 +395,88 @@ private void stopWorkflowAdapter(Id.Namespace namespace, AdapterSpecification ad
...
+
+        @Override
+        public void killed() {
+          super.killed();
+          LOG.debug("Adapter {} killed", adapterId);
+          store.setAdapterStatus(adapterId.getNamespace(), adapterId.getId(), AdapterStatus.STOPPED);
+        }
+      }, Threads.SAME_THREAD_EXECUTOR);
+    } catch (Throwable t) {
+      if (t instanceof FileNotFoundException) {
```
albertshau: catch this separately. Shouldn't be FileNotFound right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -379,12 +395,88 @@ private void stopWorkflowAdapter(Id.Namespace namespace, AdapterSpecification ad
...
+        throw new NotFoundException(workerId);
+      } else {
+        Throwables.propagate(t);
```
albertshau: should try to avoid doing this whenever possible.  It changes checked exceptions to runtime exceptions means you lose the ability to do proper error handling.  why not include them as exceptions the method can throw?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -379,12 +395,88 @@ private void stopWorkflowAdapter(Id.Namespace namespace, AdapterSpecification ad
...
+    RunRecord adapterRun = Iterables.getFirst(runRecords, null);
+    if (adapterRun == null) {
+      throw new NotFoundException(adapterId);
+    }
+
+    RunId runId = RunIds.fromString(adapterRun.getPid());
+    try {
+      lifecycleService.stopProgram(runId);
+    } catch (Throwable t) {
+      LOG.error("Error while trying to stop Adapter {} which has RunId of {} : ", adapterId, runId, t);
```
albertshau: similar comment as above. Catching just to log an error and then propagate means callers can't handle errors differently.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/ScheduleTaskRunner.java
```diff
@@ -136,26 +113,21 @@ public ScheduleTaskRunner(Store store, ProgramRuntimeService runtimeService, Pre
...
-  private ListenableFuture<?> execute(final Program program, ProgramOptions options)
-    throws TaskExecutionException {
-    ProgramRuntimeService.RuntimeInfo runtimeInfo = runtimeService.run(program, options);
+  private ListenableFuture<?> execute(final Id.Program id, final ProgramType type, Map<String, String> sysArgs,
+                                      Map<String, String> userArgs) throws IOException, TaskExecutionException {
+    ProgramRuntimeService.RuntimeInfo runtimeInfo;
+    try {
+      runtimeInfo = lifecycleService.start(id, type, sysArgs, userArgs, false);
+    } catch (ProgramNotFoundException e) {
+      throw new TaskExecutionException(UserMessages.getMessage(UserErrors.PROGRAM_NOT_FOUND), e, false);
```
albertshau: think you should just have the method throw ProgramNotFoundException instead of converting it to TaskExecutionException.  Seems like they are two very different types of errors.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -379,12 +395,88 @@ private void stopWorkflowAdapter(Id.Namespace namespace, AdapterSpecification ad
...
-  private void stopWorkerAdapter(Id.Namespace namespace, AdapterSpecification adapterSpec) {
-    // TODO: implement
+  private void stopWorkerAdapter(Id.Namespace namespace, AdapterSpecification adapterSpec) throws NotFoundException {
+    final Id.Adapter adapterId = Id.Adapter.from(namespace, adapterSpec.getName());
+    final Id.Program workerId = getWorkerId(namespace, adapterSpec);
+    List<RunRecord> runRecords = store.getRuns(workerId, ProgramRunStatus.RUNNING, 0, Long.MAX_VALUE, Integer.MAX_VALUE,
+                                               adapterSpec.getName());
+    RunRecord adapterRun = Iterables.getFirst(runRecords, null);
+    if (adapterRun == null) {
+      throw new NotFoundException(adapterId);
```
albertshau: throwing this is misleading, stopAdapter documents that NotFound is thrown when the adapter could not be found, which means this will result in a 404 in the handler.  Should probably change the current NotFoundExceptions to AdapterNotFoundExceptions.

albertshau: also, how can this happen?  If the run is not there should this just be a no-op instead of an error?

albertshau:
happy path logic looks good, most comments are around error scenarios

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/ScheduleTaskRunner.java
gokulavasan: The Javadoc mentions that TaskExecutionException is thrown if program is already running or program is not found.

gokulavasan:

albertshau:
:+1:

PR #1963, opened 2015-04-09
Fix/ui flows streams injection

ajainarayanan:
Service to ingest data into a stream in a flow (typically should be able to ingest data anywhere given a streamid)


elfenheart:
:+1: LGTM, but we still have to talk to the back end about the quote marks added to the event

ajainarayanan:
@elfenheart True. Lets talk to @albertshau When he is back.

PR #1965, opened 2015-04-09
Implements SparkListener using Scala

chtyim:
After upgraded to latest IDE, it keeps showing error for the class `SparkProgramListener`, saying that static method on interface is only supported in Java8. It happens on the line `JobSucceeded.canEqual(jobEnd.jobResult())`.

Since the `SparkListener` is a scala trait, I think it's best to implement it with a scala class.


chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1391-2

abaranau:
LGTM

PR #1966, opened 2015-04-09
Refactor AbstractDatasetTest into DatasetFrameworkTestUtil that extends ...

abaranau:
...ExternalResource.

Only refactor, no logic changes..


abaranau:
http://builds.cask.co/browse/CDAP-DUT1393-1

cdap-data-fabric/src/test/java/co/cask/cdap/api/dataset/lib/CounterTimeseriesTableTest.java
```diff
@@ -34,24 +35,28 @@
 /**
  * Tests covering the {@link co.cask.cdap.api.dataset.lib.CounterTimeseriesTable} class.
  */
-public class CounterTimeseriesTableTest extends AbstractDatasetTest {
+public class CounterTimeseriesTableTest {
+  @ClassRule
```
chtyim: :+1:

chtyim:
LGTM

PR #1967, opened 2015-04-09
Fix/ui flowlet detail stream

elfenheart:
Flowlet detail page now accounting for STREAM input.

Change default graph to keep moving based on time when there is no data.


ajainarayanan:
:+1: LGTM

PR #1968, opened 2015-04-09
Fix/ui preferences styling

elfenheart:
Fix styling of Preferences page:
- Hiding hr tag when there is no system preferences
- Aligning buttons

screen shot 2015-04-09 at 1 29 17 pm


ajainarayanan:
:+1: LGTM

PR #1971, opened 2015-04-09
Fix build on bamboo.

awholegunch:
Removes an extra space character that is causing issues on CentOS and building the documentation.


gokulavasan:
LGTM :+1:

awholegunch:
Thanks!

PR #1972, opened 2015-04-09
Fix doc build on bamboo by removing an extra space character.

awholegunch:
Apparently, under CentOS, having even a single extra space at the end of a python file causes grief.
This change deletes that last line with that single character in it.


gokulavasan:
LGTM :+1:

PR #1974, opened 2015-04-09
Ignore zero values only for COUNTER Metric Type.

anwar6953:
Ignore zero values only for COUNTER Metric Type. Gauge of a zero value should NOT be ignored.

https://issues.cask.co/browse/CDAP-2055
http://builds.cask.co/browse/CDAP-DUT1399-1


abaranau:
Where's JIRA for this?

abaranau:
LGTM

anwar6953:
added: https://issues.cask.co/browse/CDAP-2055

abaranau:
add jira # in commit message and merge

PR #1975, opened 2015-04-09
reload current state when starting and stopping run

elfenheart:
fix to reload current state when starting a new run.

ToDo:
- research ways to achieve this without wrapping $state.go in a $timeout


ajainarayanan:
:+1: LGTM

PR #1976, opened 2015-04-09
Validation of ETLBatch Adapter - hookup of source/transform/sink stages

gokulavasan:
Validate the source, sink, transform stages during the configuration step. So that the creation of the adapter fails before the program is started.

JIRA : https://issues.cask.co/browse/CDAP-1772
Bamboo : http://builds.cask.co/browse/CDAP-DT9-2


cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -91,7 +93,8 @@ public void configureAdapter(String adapterName, ETLBatchConfig etlConfig, Adapt
     configureSink(sink, configurer);
     configureTransforms(transform, configurer);
 
-    //TODO: Validate if source, transforms, sink can be tied together
+    // Validate Adapter by making source the key-value types match.
```
albertshau: nit: typo, "making sure"

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -100,6 +103,49 @@ public void configureAdapter(String adapterName, ETLBatchConfig etlConfig, Adapt
                                             cronEntry));
   }
 
+  private void validateAdapter(ETLStage source, ETLStage sink, List<ETLStage> transform) throws Exception {
+    String sourceName = source.getName();
+    String sinkName = sink.getName();
+    String sourceClassName = sourceClassMap.get(sourceName);
+    String sinkClassName = sinkClassMap.get(sinkName);
+
+    BatchSource batchSource = (BatchSource) Class.forName(sourceClassName).newInstance();
```
albertshau: these are instantiated in configureSource, configureSink, etc. It would be better to move things around so that they're only created once.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -100,6 +103,49 @@ public void configureAdapter(String adapterName, ETLBatchConfig etlConfig, Adapt
...
+    BatchSink batchSink = (BatchSink) Class.forName(sinkClassName).newInstance();
+
+    if (transform.size() == 0) {
+      // No transforms. Check only source and sink.
+      Preconditions.checkArgument(batchSink.getKeyType().equals(batchSource.getKeyType()));
```
albertshau: should give a meaningful error message, otherwise people will just be confused by the exception.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
albertshau: same with the transforms

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLMapReduceTest.java
```diff
@@ -70,12 +70,26 @@ public void testConfig() throws Exception {
     }
   }
 
+  @Test(expected = IllegalArgumentException.class)
+  public void testInvalidConfig() throws Exception {
```
albertshau: nit: there can be many types of invalid configs (for exaple, source name is missing from properties), better to be more specific

albertshau:
small comments, otherwise lgtm

gokulavasan:
@albertshau Addressed comments. Please take a look. Thanks!

albertshau:
couple more comments

gokulavasan:
@albertshau Can you take a look again Albert? Especially exception handling during instantiation of stages part.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
```diff
@@ -17,7 +17,9 @@
 package co.cask.cdap.templates.etl.api;
 
 import co.cask.cdap.api.ProgramLifecycle;
+import com.google.common.reflect.TypeToken;
```
albertshau: i guess guava is already a dependency?  Would be good to avoid putting it in api if possible but I guess we can revisit.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
```diff
@@ -30,6 +32,47 @@
  */
 public abstract class Transform<KEY_IN, VALUE_IN, KEY_OUT, VALUE_OUT> implements ProgramLifecycle<TransformContext> {
 
+  private final TypeToken<KEY_IN> keyInTypeToken = new TypeToken<KEY_IN>(getClass()) { };
```
albertshau: a little better if we store the Type instead of TypeToken

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -106,11 +117,78 @@ public void configureAdapter(String adapterName, ETLBatchConfig etlBatchConfig,
...
+      String sinkClassName = sinkClassMap.get(sink.getName());
+      batchSource = (BatchSource) Class.forName(sourceClassName).newInstance();
+      batchSink = (BatchSink) Class.forName(sinkClassName).newInstance();
+
+      for (ETLStage etlStage : transformList) {
+        String transformName = transformClassMap.get(etlStage.getName());
+        Transform transformObj = (Transform) Class.forName(transformName).newInstance();
+        transforms.add(transformObj);
+      }
+    } catch (ClassNotFoundException e) {
```
albertshau: I think we can just catch any Exception here to catch everything. For example, an InstantiationException can happen if they give an interface, or give a class without nullary constructor, both of which would be invalid input.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -106,11 +117,78 @@ public void configureAdapter(String adapterName, ETLBatchConfig etlBatchConfig,
...
+      throw new IllegalArgumentException(String.format("Unable to load class. Check stage names. %s", e));
```
albertshau: Don't need to %s and string format. Can just have the string message as first arg and the exception as the second.

albertshau:
minor comments, then lgtm

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Transform.java
gokulavasan: Ya, since etl-api depends on cdap-api, that already brings in guava. But we need to not depend on it. Not sure if there is an easy way out. I will create a JIRA for this.

gokulavasan:

albertshau:
lgtm

PR #1977, opened 2015-04-09
reload namespace list after creating namespace

elfenheart:
force namespace list reload after creating a new namespace


ajainarayanan:
:+1: LGTM

PR #1978, opened 2015-04-10
CDAP-1928 Interface for the Condition node in the Workflow.

sagarkapare:
Jira: https://issues.cask.co/browse/CDAP-1928

This PR contains the implementation of the interface for Condition node in the Workflow. 
Following example illustrates the API for the condition:
```java
@Override
public void configure() {
   setName("PurchaseHistoryWorkflow");
   setDescription("PurchaseHistoryWorkflow description");
   addMapReduce("PurchaseHistoryBuilder");
   condition(new MyPredicate())
      .addMapReduce("UserProfileBuilder")
   .otherwise()
      .addMapReduce("CustomBuilder")
   .end();
}
```

where `MyPredicate` is the public class which implements the `Predicate` interface as
```java
public static class MyPredicate implements Predicate<Map<String, String>> {
    @Override
    public boolean apply(@Nullable Map<String, String> input) {
      if (input != null && input.containsKey("someKey")) {
        return true;
      }
      return false;
    }
}
```

If the predicate evaluates to true MapReduce program `UserProfileBuilder` will be executed, otherwise MapReduce program `CustomBuilder` will be executed.

Build is running here: http://builds.cask.co/browse/CDAP-DUT1402-1

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
@@ -30,11 +30,4 @@ public interface Predicate<T> {
    * @return the result of the evaluation of the predicate
    */
   boolean apply(@Nullable T input);
-
-  /**
-   * Indicates whether another object is equal to this predicate.
-   * @param object the object to be compared with this predicate
-   * @return the result of the comparison
-   */
-  boolean equals(@Nullable Object object);
 }
--- a/cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api;
+
+import javax.annotation.Nullable;
+
+/**
+ * Determines a true or false value for a given input.
+ *
+ * @param <T> the type of the input based on which the predicate is to be evaluated
+ */
+public interface Predicate<T> {
+  /**
+   * Returns the result of applying this predicate to {@code input}.
+   * @param input the input based on which the predicate to be evaluated
+   * @return the result of the evaluation of the predicate
+   */
+  boolean apply(@Nullable T input);
+
+  /**
+   * Indicates whether another object is equal to this predicate.
+   * @param object the object to be compared with this predicate
+   * @return the result of the comparison
+   */
+  boolean equals(@Nullable Object object);
+}
--- a/cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowConditionNode.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowConditionNode.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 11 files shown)


cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @param <T> the type of the input based on which the predicate is to be evaluated
+ */
+public interface Predicate<T> {
+  /**
+   * Returns the result of applying this predicate to {@code input}.
+   * @param input the input based on which the predicate to be evaluated
+   * @return the result of the evaluation of the predicate
+   */
+  boolean apply(@Nullable T input);
```
chtyim: Can it be null?

cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Indicates whether another object is equal to this predicate.
+   * @param object the object to be compared with this predicate
+   * @return the result of the comparison
+   */
+  boolean equals(@Nullable Object object);
```
chtyim: I would say we don't need this for now.

bdmogal:
Why don't we call these methods:
```
if(new MyPredicate())
  .addMapReduce(...)
else()
  .addSpark(...)
end()
```
?

cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
sagarkapare: Ok. will remove it.

cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
sagarkapare: I made the interface similar to the guava's `Predicate` interface. can remove it.

sagarkapare:
`if`, `else` are keywords in java, we can't use them for the function names.

bdmogal:
oops. stupid question then. ignore :)

cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
chtyim: hum.. maybe just keep it for now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConditionConfigurer.java
```diff
@@ -0,0 +1,99 @@
...
+  implements WorkflowConditionConfigurer<T>, WorkflowConditionAdder {
+
+  private final T parentConfigurer;
+  private final List<WorkflowNode> ifBranch = Lists.newArrayList();
+  private final List<WorkflowNode> elseBranch = Lists.newArrayList();
+  private List<WorkflowNode> currentBranch;
+  private boolean addingToIfBranch = true;
+  private final Predicate<Map<String, String>> predicate;
+
+  public DefaultWorkflowConditionConfigurer(T parentConfigurer, Predicate<Map<String, String>> predicate) {
```
chtyim: You only need the predicate class name in the configurer, no need to carry the instance around.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowConditionNode.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Represents the CONDITION node in the {@link Workflow}.
+ */
+public class WorkflowConditionNode extends WorkflowNode {
+  private final List<WorkflowNode> ifBranch;
+  private final List<WorkflowNode> elseBranch;
+  private final String predicateClassName;
+
+  public WorkflowConditionNode(String nodeId, String predicateClassName, List<WorkflowNode> ifBranch,
```
chtyim: Should it be public or just package visible?

chtyim:
Just some minor comments. Overall looks ok to me.

cdap-api/src/main/java/co/cask/cdap/api/Predicate.java
sagarkapare: Removed this method from the interface.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConditionConfigurer.java
sagarkapare: Fixed. Using predicate class name now.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowConditionNode.java
sagarkapare: It is used by `DefaultWorkflowConditionConfigurer` which is in package `co.cask.cdap.internal.app.workflow`. Hence it should be public.

sagarkapare:

chtyim:
LGTM

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT1402-3

sagarkapare:
Build passed. Merging.

PR #1980, opened 2015-04-10
Remove workerSpec from AdapterSpec

gokulavasan:
Removing WorkerSpec from AdapterSpec since it can be retrieved from the ApplicationSpec from Store. The only information that is required to be passed on from the Template's configureAdpater method is instances and runtime args (in case of Realtime Adapter) and ScheduleSpecification (in case of Batch Adapter) since we don't allow schedules to be part of Application Template.

More changes need to be done to AdapterSpec but they will be addressed in subsequent PRs so that PR #1962 can make use of these changes.

Bamboo : http://builds.cask.co/browse/CDAP-DUT1405-3

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
@@ -348,10 +348,7 @@ public class AdapterService extends AbstractIdleService {
   private Id.Program getWorkflowId(Id.Namespace namespace, AdapterSpecification adapterSpec) throws NotFoundException {
     Id.Application appId = Id.Application.from(namespace, adapterSpec.getTemplate());
     ApplicationSpecification appSpec = store.getApplication(appId);
-    try {
-      Preconditions.checkArgument(appSpec != null);
-      Preconditions.checkArgument(appSpec.getWorkflows().size() == 1);
-    } catch (IllegalArgumentException e) {
+    if (appSpec == null || appSpec.getWorkflows().size() != 1) {
       throw new NotFoundException(appId);
     }
     String workflowName = Iterables.getFirst(appSpec.getWorkflows().keySet(), null);
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
@@ -44,6 +44,7 @@ import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.templates.AdapterSpecification;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.hash.HashCode;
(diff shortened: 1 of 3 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -344,14 +344,18 @@ public void startAdapter(Id.Namespace namespace, String adapterName)
     setAdapterStatus(namespace, adapterName, AdapterStatus.STARTED);
   }
 
+  private Id.Program getWorkflowId(Id.Namespace namespace, AdapterSpecification adapterSpec) {
+    ApplicationSpecification appSpec = store.getApplication(Id.Application.from(namespace, adapterSpec.getTemplate()));
+    Preconditions.checkNotNull(appSpec);
```
chtyim: Do you want to throw `NullPointerException` or something more meaningful like `NotFoundException`?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -344,14 +344,18 @@ public void startAdapter(Id.Namespace namespace, String adapterName)
...
+    Preconditions.checkArgument(appSpec.getWorkflows().size() == 1);
```
chtyim: Same comment as above.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -344,14 +344,18 @@ public void startAdapter(Id.Namespace namespace, String adapterName)
...
+    String workflowName = Lists.newArrayList(appSpec.getWorkflows().keySet()).get(0);
```
chtyim: Why need to new an array list? Use `Iterables.getFirst(appSpec.getWorkflows().keySet())`.

cdap-app-fabric/src/main/java/co/cask/cdap/templates/DefaultAdapterConfigurer.java
```diff
@@ -165,33 +163,23 @@ public void addRuntimeArgument(String key, String value) {
   }
 
   public AdapterSpecification createSpecification() {
+    ScheduleSpecification scheduleSpec = null;
+    if (programType == ProgramType.WORKFLOW) {
+      String workflowName = templateSpec.getWorkflows().values().iterator().next().getName();
```
chtyim: You can use `Iterables.getFirst`.

chtyim:
Just couple minor comments. The rest LGTM.

gokulavasan:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -344,14 +345,23 @@ public void startAdapter(Id.Namespace namespace, String adapterName)
     setAdapterStatus(namespace, adapterName, AdapterStatus.STARTED);
   }
 
+  private Id.Program getWorkflowId(Id.Namespace namespace, AdapterSpecification adapterSpec) throws NotFoundException {
+    Id.Application appId = Id.Application.from(namespace, adapterSpec.getTemplate());
+    ApplicationSpecification appSpec = store.getApplication(appId);
+    try {
+      Preconditions.checkArgument(appSpec != null);
```
chtyim: maybe just a simple `if:

```java
if (appSpec == null || appSpec.getWorkflows().size() != 1) {
  throw new NotFoundException(appId);
}
```

gokulavasan:
@chtyim Addressed comments. Please take a look. Thanks!

chtyim:
LGTM

PR #1981, opened 2015-04-10
(CDAP-2036) Fix the LocalStreamService failure

chtyim:
- An AbstractScheduledService.runOneIteration shouldn’t throw exception, otherwise it will kill the service.


sreevatsanraman:
LGTM :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1406

PR #1983, opened 2015-04-10
Convert documentation build to run under python 2.6

awholegunch:
These changes modify the build scripts so that they will successfully run under Python 2.6 on CentOS 6.

The script runs successfully; the documentation built is [staged](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_convert_build_to_python_2.6/en/index.html).

These changes would need to be ported to 3.0 in order for it to build there.

### DIFF
--- a/cdap-docs/_common/common_conf.py
+++ b/cdap-docs/_common/common_conf.py
@@ -49,7 +49,6 @@ def get_sdk_version():
         p2 = subprocess.Popen(["awk", "NR==1;START{print $1}"], stdin=p1.stdout, stdout=subprocess.PIPE)
         full_version_temp = p2.communicate()[0]
 # 
-        print "full_version_temp: %s" % full_version_temp
         full_version = full_version_temp.strip().replace("<version>", "").replace("</version>", "")
         version = full_version.replace("-SNAPSHOT", "")
         short_version = '%s.%s' % tuple(version.split('.')[0:2])
--- a/cdap-docs/_common/common_conf.py
+++ b/cdap-docs/_common/common_conf.py
@@ -36,15 +36,25 @@ from datetime import datetime
 
 def get_sdk_version():
     # Sets the Build Version
-    grep_version_cmd = "grep '<version>' ../../../pom.xml | awk 'NR==1;START{print $1}'"
     version = None
     short_version = None
     full_version = None
     try:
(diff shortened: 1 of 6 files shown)


wolf31o2:
:+1: LGTM

awholegunch:
I'm doing a bit more testing before I merge: I just remembered that I had not tested building the Javadocs, nor did I test it completely under Py2.7 on Mac OS X...testing Javadocs now.
It has passed the test under Py2.7 on Mac OS X.

cdap-docs/_common/common_conf.py
```diff
@@ -36,15 +36,24 @@
...
     short_version = None
     full_version = None
     try:
-        full_version = subprocess.check_output(grep_version_cmd, shell=True).strip().replace("<version>", "").replace("</version>", "")
+# Python 2.7 commands
+#         grep_version_cmd = "grep '<version>' ../../../pom.xml | awk 'NR==1;START{print $1}'"
+#         print "grep_version_cmd: %s" % grep_version_cmd
+#         full_version_temp = subprocess.check_output(grep_version_cmd, shell=True)
+# Python 2.6 commands
+        p1 = subprocess.Popen(["grep" , "<version>", "../../../pom.xml" ], stdout=subprocess.PIPE)
```
duddun: While it is not critical, I am thinking it would be better to set a variable with the location of the pom.xml (which in this case is the one in the cdap repo directory, it looks like) than pointing to a relative directory 3 levels of up from where the script thinks it is.

duddun:
LGTM

cdap-docs/_common/common_conf.py
awholegunch: When would you see that variable being set, and by which process? It's probably a good idea to change it, as this is fragile, as you've spotted. The script is a common script that gets copied and included; as I write this, I'm thinking that the including script (which is the conf.py file in each of the manuals) would then set the variable. In practise, I would set the variable by default to this, and a conf.py could over-ride it if necessary. As they are all at the same level right now, none would need to, but I could see a circumstance where it might be required.

awholegunch:
OK, not quite right. The Javadocs built under CentOS, but the formatting is screwed up, which means the CSS file didn't get copied correctly. Need to work on that next.

awholegunch:
The Javadocs will take an hour to fix, so I will merge what's done, and then open a new branch with the Javadoc fixes.

PR #1985, opened 2015-04-10
Refactoring to allow passing both Kafka offset and log event time as LogOffset

poornachandra:
This will allow us to move between Kafka source and HDFS source while querying for logs.

JIRA https://issues.cask.co/browse/CDAP-2053

### DIFF
--- a/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/log/MockLogReader.java
+++ b/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/log/MockLogReader.java
@@ -25,6 +25,7 @@ import co.cask.cdap.logging.LoggingConfiguration;
 import co.cask.cdap.logging.filter.Filter;
 import co.cask.cdap.logging.read.Callback;
 import co.cask.cdap.logging.read.LogEvent;
+import co.cask.cdap.logging.read.LogOffset;
 import co.cask.cdap.logging.read.LogReader;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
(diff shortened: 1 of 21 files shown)


cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/log/MockLogReader.java
```diff
@@ -52,49 +53,51 @@
...
     for (int i = 0; i < MAX; ++i) {
       logMap.put(defaultNamespacedLogBaseDir + "/testApp4/userservice-testService1",
-                 new LogLine(i, "testService1<img>-" + i));
+                 new LogLine(new LogOffset(i, i), "testService1<img>-" + i));
     }
 
     // Add log lines for app testApp2, procedure testProcedure1
     for (int i = 0; i < MAX; ++i) {
       logMap.put(defaultNamespacedLogBaseDir + "/testApp2/procedure-testProcedure1",
-                 new LogLine(i, "testProcedure1<img>-" + i));
```
sreevatsanraman: nit: procedure is deprecated. you might want to call it service.

sreevatsanraman:
LGTM other than the nit. please fix and merge.

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/log/MockLogReader.java
poornachandra: We already have a test for service. If procedure is removed I can remove this test, else we can remove it when procedure gets removed.

poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT1410-3

PR #1986, opened 2015-04-10
Feature/ui workflow schedule

elfenheart:
ToDo:
- Poll for Last Run when API is ready. 

screen shot 2015-04-12 at 1 57 03 pm

### DIFF
--- a/cdap-ui/app/features/workflows/controllers/tabs/schedule-ctrl.js
+++ b/cdap-ui/app/features/workflows/controllers/tabs/schedule-ctrl.js
@@ -22,44 +22,35 @@ angular.module(PKG.name + '.feature.workflows')
           v.time.week = parse[4];
         }
         v.isOpen = false;
+
+        dataSrc.poll({
+          _cdapNsPath: '/apps/' + $state.params.appId +
+                        '/schedules/' + v.schedule.name + '/status'
+        }, function (response) {
+          v.status = response.status;
+        });
       });
 
-      $scope.toggleLine = function (schedule) {
-        schedule.isOpen = !schedule.isOpen;
-      };
     });
-    // // TODO: change to real schedule
-    // $scope.schedules = [
-    //   {
-    //     'program': {
-    //       'programName': 'PurchaseHistoryWorkflow',
-    //       'programType': 'WORKFLOW'
-    //     },
-    //     'properties': {},
-    //     'schedule': {
-    //       'cronExpression': '0 4 * * *',
-    //       'description': 'Schedule execution every day',
-    //       'name': 'DailySchedule'
-    //     },
-    //     'scheduleType': 'TIME'
-    //   },
-    //   {
-    //     'program': {
-    //       'programName': 'PurchaseHistoryWorkflow',
-    //       'programType': 'WORKFLOW'
-    //     },
-    //     'properties': {},
-    //     'schedule': {
-    //       'dataTriggerMB': 1,
-    //       'description': 'Schedule execution when 1 MB or more of data is ingested in the   purchaseStream',
-    //       'name': 'DataSchedule',
-    //       'streamName': 'purchaseStream'
-    //     },
-    //     'scheduleType': 'STREAM'
-    //   }
-    // ];
 
+    $scope.toggleLine = function (schedule) {
+      schedule.isOpen = !schedule.isOpen;
+    };
 
+    $scope.suspendSchedule = function (obj) {
+      dataSrc.request({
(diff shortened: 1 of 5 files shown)


cdap-ui/app/features/workflows/controllers/tabs/schedule-ctrl.js
```diff
@@ -0,0 +1,56 @@
...
+      angular.forEach($scope.schedules, function(v) {
+        if (v.scheduleType === 'TIME') {
+          console.log('test');
+          var parse = v.schedule.cronExpression.split(' ');
+          v.time = {};
+          v.time.min = parse[0];
+          v.time.hour = parse[1];
+          v.time.day = parse[2];
+          v.time.month = parse[3];
+          v.time.week = parse[4];
```
ajainarayanan: Are we sure that we get all 5 parse cron expressions all the time? Should we check before accessing them?

cdap-ui/app/features/workflows/templates/tabs/schedules.html
```diff
@@ -0,0 +1,109 @@
...
+  </div>
+  <div class="col-xs-2">
+    <h4>Status</h4>
+  </div>
+</div>
+
+<div class="schedule-content">
+  <div class="schedule-rows" ng-repeat="schedule in schedules">
+    <div class="row schedule-row"
+      ng-click="toggleLine(schedule)"
```
ajainarayanan: This can be inlined 

```
<div class="..."
        ng-click="schedule.isOpen = !schedule.isOpen"
> ... </div>
```

Feel a function call here is an over kill.

cdap-ui/app/features/workflows/workflow.less
```diff
@@ -12,4 +13,58 @@ body.state-workflows{
...
+        line-height: 30px;
+
+        p {
+          vertical-align: middle;
+          margin-bottom: 0;
+        }
+      }
+
+      table {
+        width: inherit;
```
ajainarayanan: Is this necessary? We don't seem to specify width anywhere? Isn't this by default happening?

cdap-ui/app/features/workflows/workflow.less
```diff
@@ -12,4 +13,58 @@ body.state-workflows{
       }
     }
   }
+
+
+  .schedule-heading {
+    background-color: white;
+    border: 1px solid @border-color;
+    border-top-left-radius: 5px;
```
ajainarayanan: Can we use ```@border-radius-base``` here? Think we use the one from bootstrap (if not we need to change that) We should probably have border-radius unified across everywhere (except top nav)

cdap-ui/app/features/workflows/templates/tabs/schedules.html
```diff
@@ -0,0 +1,109 @@
...
+        <div ng-if="schedule.scheduleType === 'TIME'">
+          <div class="col-xs-offset-2 col-xs-2">
+            <p>Crontab</p>
+          </div>
+
+          <div class="col-xs-8">
+
+            <table class="table table-responsive" cask-sortable>
+              <thead>
+                <th>min</th>
```
ajainarayanan: We use cask-sortable but have no data-predicate in the headers. Is it possible to update the header with data-predicate attribute

cdap-ui/app/features/workflows/workflow.less
elfenheart: No, the table class by default is width 100%

cdap-ui/app/features/workflows/templates/tabs/schedules.html
elfenheart: i'm mainly using cask sortable to get the styling of the table. There won't be more than 1 row of data here, so having a data-predicate is unnecessary

cdap-ui/app/features/workflows/controllers/tabs/schedule-ctrl.js
elfenheart: Confirmed with backend there are always exactly 5

cdap-ui/app/features/workflows/templates/tabs/schedules.html
ajainarayanan: Hhmm.. When we add cask-sortable it becomes a directive so angular has to now compile/link it to the directive. If we need styling then we should probably use the css class right? Just wondering if we really need to use a directive when we want only the css class .

ajainarayanan:
:+1: LGTM

PR #1987, opened 2015-04-11
use current time as endTime for aggregate queries and 

gsps1:
using empty list/map when groupBy, tags or time-range is not provided

Jira link : https://issues.cask.co/browse/CDAP-1955


anwar6953:
In comment, you mean 'is not provided', right?

gsps1:
yeah. fixed it now.

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
```diff
@@ -387,7 +387,7 @@ public void testAggregateQueryBatch() throws Exception {
     QueryRequestFormat query1 = new QueryRequestFormat(getTagsMap("namespace", "yourspace", "app", "WCount1",
                                                                   "flow", "WCounter", "flowlet", "splitter"),
                                            ImmutableList.of("system.reads"), ImmutableList.<String>of(),
-                                           ImmutableMap.of("aggregate", "true"));
+                                           ImmutableMap.<String, String>of());
```
abaranau: why? also: do we still provide timeRange as map?

gsps1: just to test it defaults to aggregate=true, when nothing is provided

abaranau:
LGTM

PR #1988, opened 2015-04-11
CDAP-2052 Restructuring adapter id in logging context

poornachandra:
Using setter for adapter id eliminates the need to have Nullable constructor parameters.

JIRA - https://issues.cask.co/browse/CDAP-2052

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/logging/context/MapReduceLoggingContext.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/logging/context/MapReduceLoggingContext.java
@@ -32,11 +32,15 @@ public class MapReduceLoggingContext extends ApplicationLoggingContext {
    * @param namespaceId namespace id
    * @param applicationId application id
    * @param mapReduceId mapreduce job id
+   * @param adapterId adapter id
    */
   public MapReduceLoggingContext(final String namespaceId, final String applicationId, final String mapReduceId,
                                  @Nullable String adapterId) {
-    super(namespaceId, applicationId, adapterId);
+    super(namespaceId, applicationId);
     setSystemTag(TAG_MAP_REDUCE_JOB_ID, mapReduceId);
+    if (adapterId != null) {
+      setAdapterId(adapterId);
+    }
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT1414-1

bdmogal:
LGTM. Reported a follow up task for me - https://issues.cask.co/browse/CDAP-2074

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/MapReduceLoggingContext.java
```diff
@@ -32,11 +30,13 @@
    * @param namespaceId namespace id
    * @param applicationId application id
    * @param mapReduceId mapreduce job id
+   * @param adapterId adapter id
    */
   public MapReduceLoggingContext(final String namespaceId, final String applicationId, final String mapReduceId,
-                                 @Nullable String adapterId) {
-    super(namespaceId, applicationId, adapterId);
+                                 String adapterId) {
+    super(namespaceId, applicationId);
```
bdmogal: Actually, ``adapterId`` should be nullable here and set only if non-null? That would be when a mapreduce is run outside of an adapter.

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/WorkerLoggingContext.java
```diff
@@ -29,9 +27,10 @@
   public static final String TAG_WORKER_ID = ".workerid";
 
   public WorkerLoggingContext(final String namespaceId, final String appId, final String workerId,
-                              @Nullable String adapterId) {
-    super(namespaceId, appId, adapterId);
+                              String adapterId) {
```
bdmogal: Same comment as above

bdmogal:
@poornachandra a couple of comments for when adapterId is null.

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/MapReduceLoggingContext.java
poornachandra: Good point.

poornachandra:
@bdmogal Addressed comments, please take a look.

bdmogal:
LGTM :+1:

poornachandra:
http://builds.cask.co/browse/CDAP-DUT1414-2

PR #1990, opened 2015-04-11
Feature/ui etl apps 2

ajainarayanan:
Phase -2 of etls apps creation
  - Fixes drag and drop functionality 
    - Adds visual ways of identifying if a draggable can be dropped in a specific zone.
  - Cleans up etl create template to work correct with properties as map.
  - Cleans up controller to a tad smaller than usual.
  - Adds an Api factory for fetch 'data' from backend (pseudo-service layer)


elfenheart:
:+1: LGTM

PR #1993, opened 2015-04-11
mapreduce metrics context fix, 

gsps1:
escape dots with tilde and replace it with dots before adding to metric system

Jira : https://issues.cask.co/browse/CDAP-2075


jwang47:
LGTM

gsps1:
http://builds.cask.co/browse/CDAP-DUT1418 - Green - merging

PR #1995, opened 2015-04-11
Stylesheet for CDAP Javadocs under Java 1.7

awholegunch:
Creates a new style sheet for CDAP Javadocs that is a mimic of the previous Java 1.6 version, but that is intended to be run under Java 1.7.

Staged at: [Revised Javadocs](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0-r2.8_fix_javadocs_under_jdk1.7/en/reference-manual/javadocs/index.html)

You can compare them to [the current Javadocs](http://stg-web101.sw.joyent.continuuity.net/cdap/2.8.0/en/reference-manual/javadocs/index.html).


sreevatsanraman:
LGTM :+1:

cdap-distributions/src/javadoc/stylesheet.css
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2014 Cask Data, Inc.
+ * Copyright © 2015 Cask Data, Inc.
```
duddun: Shouldn't this be [...] © 2014-15 Cask Data [...]?
I thought the rule was single year if it is a new file or document.

PR #1996, opened 2015-04-11
Template level properties

gokulavasan:
Extending mock endpoint to send template level properties in addition to name and description.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MockETLHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MockETLHandler.java
@@ -63,8 +63,23 @@ public class MockETLHandler extends AuthenticatedHttpHandler {
   public void getTemplate(HttpRequest request, HttpResponder responder, @PathParam("template-id") String templateId)
     throws Exception {
     JsonObject template = new JsonObject();
+    JsonArray properties = new JsonArray();
     template.addProperty("name", templateId);
     template.addProperty("description", "Template used to create ETL Adapters");
+    if (templateId.equalsIgnoreCase(BATCH)) {
+      JsonObject scheduleProperty = new JsonObject();
+      scheduleProperty.addProperty("name", "schedule");
+      scheduleProperty.addProperty("description", "Cron Schedule for Batch Adapter");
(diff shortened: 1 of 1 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MockETLHandler.java
```diff
@@ -63,8 +63,23 @@ public void getTemplates(HttpRequest request, HttpResponder responder) throws Ex
   public void getTemplate(HttpRequest request, HttpResponder responder, @PathParam("template-id") String templateId)
     throws Exception {
     JsonObject template = new JsonObject();
+    JsonArray properties = new JsonArray();
     template.addProperty("name", templateId);
     template.addProperty("description", "Template used to create ETL Adapters");
+    if (templateId.equalsIgnoreCase(BATCH)) {
+      JsonObject scheduleProperty = new JsonObject();
```
sreevatsanraman: We should add a type as well? Let's look at what Coopr returns and return relevant info for UI to use.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MockETLHandler.java
```diff
@@ -63,8 +63,23 @@ public void getTemplates(HttpRequest request, HttpResponder responder) throws Ex
   public void getTemplate(HttpRequest request, HttpResponder responder, @PathParam("template-id") String templateId)
     throws Exception {
     JsonObject template = new JsonObject();
+    JsonArray properties = new JsonArray();
     template.addProperty("name", templateId);
     template.addProperty("description", "Template used to create ETL Adapters");
+    if (templateId.equalsIgnoreCase(BATCH)) {
```
henry-cask: What other possible values for templateId?

gokulavasan: There might be many including user created ones. This is just a mock implementation so that UI can test the APIs etc.

gokulavasan:
Realized this cannot be provided by /templates/{template-id} endpoint. It should be provided by the HttpServiceHandler associated with the application template. Will modify the PR.

PR #1997, opened 2015-04-11
CDAP-1928,1611 Implementation of passing mapreduce counters to the Condi...

sagarkapare:
...tion node and execution of the condition branches based on them.

Associated Jira:
1. https://issues.cask.co/browse/CDAP-1928
2. https://issues.cask.co/browse/CDAP-1611

Hadoop counters are passed as `Map<String, String>` to the `Predicate` in the Condition node. Key of the map represents the concatenation of the counter group name and counter name by `.`, while value represents the value of the counter. 

This is currently work in progress - no unit tests yet and need more testing on the distributed cluster. 

Opening PR to get the early feedback on the approach.


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -265,12 +273,39 @@ private void executeNode(final ApplicationSpecification appSpec, WorkflowNode no
...
+  @SuppressWarnings("unchecked")
+  private void executeCondition(final ApplicationSpecification appSpec, WorkflowConditionNode node,
+                                final InstantiatorFactory instantiator, final ClassLoader classLoader,
+                                String previousNodeId) throws Exception {
+    Map<String, String> token = mapReduceJobCounters.get(previousNodeId);
+    if (previousNodeId == null || token == null) {
+      throw new IllegalStateException("Condition node is not followed by the MapReduce program.");
+    }
+
+    Class<?> clz = Class.forName(node.getPredicateClassName(), true, classLoader);
```
chtyim: Seems simpler if just use `classLoader.loadClass`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -265,12 +273,39 @@ private void executeNode(final ApplicationSpecification appSpec, WorkflowNode no
...
+        executeCondition(appSpec, (WorkflowConditionNode) node, instantiator, classLoader, previousNodeId);
+        break;
       default:
         break;
     }
   }
 
-  @Override
+  @SuppressWarnings("unchecked")
+  private void executeCondition(final ApplicationSpecification appSpec, WorkflowConditionNode node,
```
chtyim: Remove final if not needed.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
```diff
@@ -32,16 +30,16 @@
...
+   * the {@link Callable#call() call} method is invoked. When the program completes,
+   * the {@link Callable} returns the {@link Map} of execution context of the program run or {@code null} if
    * no such context exists.
    * <p/>
    * <p> An Exception is thrown from the {@link Callable#call()} method if the program fails </p>
    *
    * @throws IllegalArgumentException if no program with the specified name is defined in the workflow
    */
-  Callable<RuntimeContext> getProgramRunner(String name);
+  Callable<Map<String, String>> getProgramRunner(String name);
```
chtyim: I think it's better to have an object to represent the execution context, even though inside that object is just a Map right now. It makes extension easier later.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/MapReduceProgramWorkflowRunner.java
```diff
@@ -61,20 +67,32 @@
...
                                         "MapReduceProgramController");
     }
   }
+
+  private Map<String, String> getCounterMap(MapReduceContext context) throws Exception {
+    Map<String, String> mapCounters = Maps.newHashMap();
+    Counters counters = ((Job) context.getHadoopJob()).getCounters();
+    for (CounterGroup group : counters) {
+      for (Counter counter : group) {
+        mapCounters.put(group.getName() + "." + counter.getName(), Long.toString(counter.getValue()));
```
anwar6953: This assumes that neither group nor the counter name have `.` in it, right?

sagarkapare: We are just concatenating the counter group name and counter name to represent the key in the map. We are not recreating group name and counter name from the key, so it does not matter even if the counter name has `.` in it.
e.g. assume the group name is `some.group.name` and counter name is `some.counter.name`. The key in the map will be `some.group.name.some.counter.name`.  We will document the same.

anwar6953: But if group name and counter name can have `.` in it, then there can be conflicts right?
For instance:
group name: `some.group.name` and counter name: `foo`.
group name: `some.group` and counter name: `name.foo`

sagarkapare: ahh.. makes sense. will remove the `.` while concatenating.

anwar6953: I mean - I don't think that resolves the problem.
With those two metrics, it'll still conflict.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: using `classLoader.loadClass` now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: Removed.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
sagarkapare: Using `WorkflowToken` to wrap the hadoop counters map.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/MapReduceProgramWorkflowRunner.java
sagarkapare: Updated the map definition.

sagarkapare:
@chtyim @anwar6953 Please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/SparkProgramWorkflowRunner.java
```diff
@@ -63,20 +65,21 @@
...
     if (controller instanceof SparkProgramController) {
-      final RuntimeContext context = ((SparkProgramController) controller).getContext();
-      return executeProgram(controller, context);
+      executeProgram(controller, ((SparkProgramController) controller).getContext());
     } else {
       throw new IllegalStateException("Failed to run program. The controller is not an instance of " +
                                         "SparkProgramController");
     }
+    // TODO: currently Spark program returns the null token
+    return null;
```
anew: seems that this should return an empty token? Actually, the token should be created at start of the workflow. Then it is passed to every action, and the action can modify the token. No action should create a new token, or return null.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -265,12 +273,39 @@ private void executeNode(final ApplicationSpecification appSpec, WorkflowNode no
...
   }
 
-  @Override
+  @SuppressWarnings("unchecked")
+  private void executeCondition(ApplicationSpecification appSpec, WorkflowConditionNode node,
+                                InstantiatorFactory instantiator, ClassLoader classLoader,
+                                String previousNodeId) throws Exception {
+    WorkflowToken token = mapReduceJobCounters.get(previousNodeId);
+    if (previousNodeId == null || token == null) {
+      throw new IllegalStateException("Condition node is not followed by the MapReduce program.");
```
anew: you mean "not preceded by a MapReduce"?

cdap-app-fabric/src/test/java/co/cask/cdap/ConditionalWorkflowApp.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  static final class ConditionalWorkflow extends AbstractWorkflow {
+
+    @Override
+    protected void configure() {
+      setName("ConditionalWorkflow");
+      setDescription("Workflow to test Condition nodes");
+      addMapReduce("RecordVerifier");
+      condition(new MyVerificationPredicate()).addMapReduce("ClassicWordCount").end();
```
anew: no like this syntax. seems ambiguous to me. 
a) should be addCondition(). or ifCondition(). 
b) should be followed thenDo()...elseDo()...done();
Not sure what the right syntax is, but this one seems counter-intuitive.

chtyim: I think in this example, there is no else part. The complete syntax looks something like:

```java
condition(Predicate)
  .addMapReduce()
  .addMapReduce()
.otherwise()
  .addMapReduce()
.done()
```

there is no `then` because it is unnecessary and also more align with what the `fork-join` does as well as java syntax `if () { } else { }`.

The naming of the keywords `condition`, `otherwise` and `done` can be changed to whatever is more clear.

anew: ok

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: Yeah right. Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/SparkProgramWorkflowRunner.java
sagarkapare: Returning `null` here, since currently the `WorkflowToken` interface only has single method `getMapReduceCounters` in it which is specific to the MapReduce programs.

sagarkapare:
This PR is ready for the review. Build is running here - http://builds.cask.co/browse/CDAP-DUT1423-6. @anew Please take a look. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
```diff
@@ -27,24 +27,33 @@
...
+   * the {@link Callable} returns the {@link WorkflowToken} for the program run or {@code null} if
    * no such context exists.
    * <p/>
    * <p> An Exception is thrown from the {@link Callable#call()} method if the program fails </p>
    *
    * @throws IllegalArgumentException if no program with the specified name is defined in the workflow
+   * @throws UnsupportedOperationException if it is called from {@link Predicate}
    */
-  Callable<RuntimeContext> getProgramRunner(String name);
+  Callable<WorkflowToken> getProgramRunner(String name);
```
anew: I don't understand. Before, the program runner had access to the runtime arguments. Now it does not? How will runtime arguments be passed to programs then?

anew:
@sagarkapare please implement what we discussed in person:
- the token is created at the start of the workflow
- it is passed to every action, and actions can modify it
- for now, only map/reduce can modify it, others can only read it
- in the future we can add methods to allow adding properties by any action
This means the  program runner is not a Callable but a Runnable. 

I will review this asap as soon as you are ready. Thanks

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
sagarkapare: Creating `WorkflowToken` in the `WorkflowDriver` and passing it along to the programs running under it.

sagarkapare:
@anew Addressed the review comments. Please take a look again. Thanks!

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT1423-7

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
```diff
@@ -27,24 +26,29 @@
 
   WorkflowSpecification getWorkflowSpecification();
 
+  /**
+   * @throws UnsupportedOperationException if it is called from {@link Predicate}
```
anew: why?

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
```diff
@@ -40,9 +41,10 @@
 
   private final String name;
   private String programName;
-  private Callable<RuntimeContext> programRunner;
+  private Runnable programRunner;
   private SchedulableProgramType programType;
   private WorkflowContext context;
+  private WorkflowToken token;
```
anew: why is this needed here? It also seems that you have a getter for the token, but it never gets set.

anew: Don't you get a warning in your IDE about this? Mine warns me: token is never assigned. It also warns me that the context is assigned but never used.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
sagarkapare: Since this is the part of the `WorkflowContext`, it is possible that the user might invoke this method through the predicate. Per discussion with @chtyim, throwing `UnsupportedOperationException` in that case.

anew:
Some minor comments, and I also have a more general question: 

So, the token is a class variable. In a fork, both/all branches are executed concurrently, so they all share the same token. Is that good? Or would it be better to make a copy of the token for each branch, and merge the tokens when all branches have completed. (Merge here can mean to take an arbitrary one, but when we add more capabilities to manipulate the token, it will have to become more sophisticated). 

If two branches share the same token, and both have a MR action followed by a condition, the expectation is that each condition sees the counters of the MR in its own branch. That is not guaranteed if they share a single token.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
sagarkapare: oops... sorry!! this was missed by me. We do not need `WorkflowToken` here and context too. Removed both.

sagarkapare:
Yeah. Makes sense. Making deep copy of the `WorkflowToken` and passing the copied object to the each branch of the fork. Once the all branches of the fork are executed, currently as a part of the merge process taking the MapReduce counters from the first branch and assigning it to the merged token.

sagarkapare:
@anew Please take a look again. Thanks!

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
```diff
@@ -81,14 +78,14 @@ public void initialize(WorkflowContext context) throws Exception {
...
+      programRunner.run();
 
       // TODO (terence) : Put something back to context.
 
       LOG.info("{} Program {} workflow action completed",
                programType != null ? programType.name() : null, programName);
     } catch (Exception e) {
-      LOG.info("Failed to execute {} Program {} in workflow: {}", 
+      LOG.info("Failed to execute {} Program {} in workflow: {}",
                programType != null ? programType.name() : null, programName, e);
```
chtyim: I think `LOG` supports null value, you don't need the `?:` (if you don't call `.name()`, which is the same as `enum.toString()`.

chtyim: Meaning just write it as `LOG.info("....", programType, programName, e);`

chtyim: and by the way, when logging with an exception, you should log the stacktrace as well. Meaning it should be:

`LOG.info("Failed to execute {} Program {} in workflow", programType, programName, e);    // Remove the last {}`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Implementation of the {@link WorkflowToken} interface.
+ */
+public class BasicWorkflowToken implements WorkflowToken {
+  private Map<String, Map<String, Long>> mapReduceCounters;
+
+  @Nullable
```
chtyim: Do we need to allow null? Can it be an empty map if there is no counter? Or we do need to use `null` to distinguish no counter vs empty counter?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -0,0 +1,54 @@
...
+  public void setMapReduceCounters(Map<String, Map<String, Long>> mapReduceCounters) {
+    this.mapReduceCounters = mapReduceCounters;
+  }
+
+  /**
+   * Make a deep copy of the WorkflowToken. Currently only copies the MapReduce counters.
+   * @param token WorkflowToken to be copy
+   * @return copied WorkflowToken
+   */
+  public static WorkflowToken deepCopy(WorkflowToken token) {
```
chtyim: Why a static method? Why not an instance method that creates a clone of itself?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -0,0 +1,54 @@
...
+    BasicWorkflowToken copiedToken = new BasicWorkflowToken();
+    if (token.getMapReduceCounters() != null) {
+      Map<String, Map<String, Long>> copiedMap = new HashMap<String, Map<String, Long>>(token.getMapReduceCounters());
```
chtyim: Is this a deep copy?

sagarkapare: Fixed. My bad.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Made instance method.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: We don't use `null` to distinguish. Simpler if we store empty map for no counter. Updated the code to fix it accordingly.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
sagarkapare: Fixed.

sagarkapare:
@chtyim @anew Addressed review comments. Please take a look. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
anew: I would prefer it to return null if it is not set. The reason is that in the future we want to allow setting and retrieving arbitrary properties, not just M/R counters. If a property does not exist, it will return null. It should be the same for the M/R counters (they will be serialized into a property, too).

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
anew: ok

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Interface to represent the data that is transferred from one node to the next node in the {@link Workflow}
+ */
+public interface WorkflowToken {
+  /**
+   *
+   * @return the Hadoop MapReduce counters from the previous MapReduce program in the Workflow
+   */
+  public Map<String, Map<String, Long>> getMapReduceCounters();
```
anew: I think this should be @Nullable if the counters are not set in the token. And the javadoc should mention what it returns (null or empty map, I vote for null).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.common.collect.Maps;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Implementation of the {@link WorkflowToken} interface.
+ */
+public class BasicWorkflowToken implements WorkflowToken {
+  private Map<String, Map<String, Long>> mapReduceCounters = Maps.newHashMap();
```
anew: see my other comment, it should not an empty map.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public Map<String, Map<String, Long>> getMapReduceCounters() {
+    return mapReduceCounters;
+  }
+
+  public void setMapReduceCounters(Map<String, Map<String, Long>> mapReduceCounters) {
+    this.mapReduceCounters = mapReduceCounters;
```
anew: you should make a defensive copy into an Immutable map here, because you can't know whether the caller will modify the map later-on.

anew:
I think almost everything is good now, but please see my comments on @Nullable getMapReduceCounters() and defensive copy on setMapReduceCounters().

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
chtyim: Agree. We need to document that clearly to avoid causing unintended NPE :)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * Make a deep copy of the WorkflowToken. Currently only copies the MapReduce counters.
+   * @param token WorkflowToken to be copy
+   * @return copied WorkflowToken
+   */
+  public WorkflowToken deepCopy(WorkflowToken token) {
```
chtyim: Shouldn't take any parameter. It should create a copy of "this" (itself).

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Made it @Nullable and added javadoc

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Updated the code per comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Fixed.

sagarkapare:
@anew Please take a look again. Thank you!

anew:
I have two more comments...

sagarkapare:
@anew Fixed the code per suggestions. Please take a look again. Thank you!

anew:
LGTM

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT1423-11

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT1423-JOB1-12 Merging.

PR #1998, opened 2015-04-12
Publish runId and instanceId in log messages

poornachandra:
Adding runId and instanceId to logging context when publishing logs. Also removed some unnecessary code.

JIRA - https://issues.cask.co/browse/CDAP-2052

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/BasicFlowletContext.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/BasicFlowletContext.java
@@ -22,7 +22,6 @@ import co.cask.cdap.api.metrics.Metrics;
 import co.cask.cdap.app.metrics.ProgramUserMetrics;
 import co.cask.cdap.app.program.Program;
 import co.cask.cdap.app.runtime.Arguments;
-import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.logging.LoggingContext;
 import co.cask.cdap.common.metrics.MetricsCollectionService;
@@ -125,7 +124,8 @@ final class BasicFlowletContext extends AbstractContext implements FlowletContex
   }
 
   public LoggingContext getLoggingContext() {
-    return new FlowletLoggingContext(getNamespaceId(), getApplicationId(), getFlowId(), getFlowletId());
+    return new FlowletLoggingContext(getNamespaceId(), getApplicationId(), getFlowId(), getFlowletId(),
+                                     getRunId().getId(), String.valueOf(getInstanceId()));
   }
 
   @Override
--- a/cdap-watchdog/src/main/java/co/cask/cdap/logging/context/LoggingContextHelper.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/logging/context/LoggingContextHelper.java
@@ -78,28 +78,38 @@ public final class LoggingContextHelper {
         return null;
       }
       return new FlowletLoggingContext(namespaceId, applicationId, tags.get(FlowletLoggingContext.TAG_FLOW_ID),
-                                       tags.get(FlowletLoggingContext.TAG_FLOWLET_ID));
+                                       tags.get(FlowletLoggingContext.TAG_FLOWLET_ID),
+                                       tags.get(ApplicationLoggingContext.TAG_RUNID_ID),
+                                       tags.get(ApplicationLoggingContext.TAG_INSTANCE_ID));
     } else if (tags.containsKey(MapReduceLoggingContext.TAG_MAP_REDUCE_JOB_ID)) {
       return new MapReduceLoggingContext(namespaceId, applicationId,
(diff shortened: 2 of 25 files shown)


poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT1422-4

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/LoggingContextHelper.java
```diff
@@ -199,29 +206,4 @@ public static Filter createFilter(LoggingContext loggingContext) {
 
     }
   }
-
-  private static Filter createGenericFilter(String namespaceId, String applicationId, String entityId) {
```
bdmogal: We don't need this anymore?

cdap-watchdog/src/test/java/co/cask/cdap/logging/appender/TestResilientLogging.java
```diff
@@ -91,6 +91,8 @@ public void testResilientLogging() throws Exception {
     cConf.set(Constants.Dataset.Executor.ADDRESS, "localhost");
     cConf.setInt(Constants.Dataset.Executor.PORT, Networks.getRandomPort());
 
+    cConf.set(Constants.CFG_LOCAL_DATA_DIR, tmpFolder.newFolder().getAbsolutePath());
```
bdmogal: Changes in this file are just to use a ``@ClassRule TemporaryFolder`` instead of the ``/tmp`` directory so as to not leave side-effects in ``/tmp``, right?

cdap-watchdog/src/test/java/co/cask/cdap/logging/appender/file/TestFileLogging.java
```diff
@@ -85,16 +85,48 @@ public static void setUpContext() throws Exception {
...
     new LogAppenderInitializer(appender).initialize("TestFileLogging");
 
     Logger logger = LoggerFactory.getLogger("TestFileLogging");
-    for (int i = 0; i < 60; ++i) {
-      Exception e1 = new Exception("Test Exception1");
-      Exception e2 = new Exception("Test Exception2", e1);
-      logger.warn("Test log message " + i + " {} {}", "arg1", "arg2", e2);
+    Exception e1 = new Exception("Test Exception2");
+
+    LoggingContextAccessor.setLoggingContext(new FlowletLoggingContext("TFL_ACCT_2", "APP_1", "FLOW_1", "FLOWLET_1",
```
bdmogal: nit:  Should we change these constants to use "NS" instead of "ACCT"? Only for the new ones.

bdmogal:
Just a couple of questions and a nit. Other than that LGTM :+1:

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/LoggingContextHelper.java
poornachandra: It is not needed as `GenericLoggingContext` is not used like other logging contexts.

cdap-watchdog/src/test/java/co/cask/cdap/logging/appender/TestResilientLogging.java
poornachandra: Yes

cdap-watchdog/src/test/java/co/cask/cdap/logging/appender/file/TestFileLogging.java
poornachandra: I'll leave this as-is for now since it will cause conflict in my next branch.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/BasicFlowletContext.java
```diff
@@ -125,7 +124,8 @@ public int getInstanceId() {
   }
 
   public LoggingContext getLoggingContext() {
-    return new FlowletLoggingContext(getNamespaceId(), getApplicationId(), getFlowId(), getFlowletId());
+    return new FlowletLoggingContext(getNamespaceId(), getApplicationId(), getFlowId(), getFlowletId(),
+                                     getRunId().getId(), String.valueOf(getInstanceId()));
```
gsps1: question : is it better to have ``runId`` before ``flowletId`` ?.  as all the flowlets of a flow have the same runId.

poornachandra: It is less confusing to have all program ids in one place. What do you say?

gsps1: when we scan for logs, we would typically scan for flow + run-id ( flowlets aggregated) (at least in UI) that's the case right ? , wouldn't it be more efficient to have run-id next to flow in that scenario ?

poornachandra: The key for storing logs is derived using `FlowletLoggingContext`, and is not the exact key. So we can always rearrange the fields for the key.

PR #2004, opened 2015-04-12
[CDAP-2064] Added RDBMS Batch source

bdmogal:
Added RDBMS Batch Source

Mostly done apart from the following:
1. Testing on a cluster - DONE
2. Test with a secure database (Do username, password need to be passed separately)

Want to get the review started as I work on the above.

Outside the scope of this PR, but we probably need to think of these things nonetheless:
1. Importing multiple tables
2. Packaging driver class jars

Jira: [CDAP-2064](https://issues.cask.co/browse/CDAP-2064)
Build: https://builds.cask.co/browse/CDAP-DUT1424

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
@@ -63,6 +63,11 @@
       <artifactId>cdap-unit-test</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.hsqldb</groupId>
+      <artifactId>hsqldb</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
@@ -31,12 +31,14 @@ import co.cask.cdap.templates.etl.batch.sinks.KVTableSink;
 import co.cask.cdap.templates.etl.batch.sinks.TableSink;
 import co.cask.cdap.templates.etl.batch.sinks.TimePartitionedFileSetDatasetAvroSink;
 import co.cask.cdap.templates.etl.batch.sources.BatchReadableSource;
+import co.cask.cdap.templates.etl.batch.sources.DBSource;
 import co.cask.cdap.templates.etl.batch.sources.KVTableSource;
 import co.cask.cdap.templates.etl.batch.sources.StreamBatchSource;
 import co.cask.cdap.templates.etl.batch.sources.TableSource;
 import co.cask.cdap.templates.etl.common.Constants;
 import co.cask.cdap.templates.etl.common.DefaultPipelineConfigurer;
 import co.cask.cdap.templates.etl.common.DefaultStageConfigurer;
+import co.cask.cdap.templates.etl.transforms.DBRecordToStructuredRecordTransform;
 import co.cask.cdap.templates.etl.transforms.GenericTypeToAvroKeyTransform;
 import co.cask.cdap.templates.etl.transforms.IdentityTransform;
 import co.cask.cdap.templates.etl.transforms.RowToStructuredRecordTransform;
@@ -88,7 +90,9 @@ public class ETLBatchTemplate extends ApplicationTemplate<ETLBatchConfig> {
                                         TimePartitionedFileSetDatasetAvroSink.class,
                                         StreamToStructuredRecordTransform.class,
                                         GenericTypeToAvroKeyTransform.class,
-                                        ScriptFilterTransform.class));
+                                        ScriptFilterTransform.class,
+                                        DBSource.class,
+                                        DBRecordToStructuredRecordTransform.class));
   }
 
   private void initTable(List<Class> classList) throws Exception {
@@ -156,7 +160,7 @@ public class ETLBatchTemplate extends ApplicationTemplate<ETLBatchConfig> {
         transforms.add(transformObj);
       }
     } catch (Exception e) {
-      throw new IllegalArgumentException("Unable to load class. Check stage names. %s", e);
+      throw new IllegalArgumentException("Unable to load class. Check stage names.", e);
     }
   }
 
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
@@ -117,6 +117,15 @@
       <artifactId>twitter4j-stream</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-mapreduce-client-core</artifactId>
+    </dependency>
+    <!-- TODO: Remove when plugin management is added -->
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.11</version>
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.batch.sources;
+
+import co.cask.cdap.templates.etl.api.PipelineConfigurer;
+import co.cask.cdap.templates.etl.api.Property;
+import co.cask.cdap.templates.etl.api.StageConfigurer;
+import co.cask.cdap.templates.etl.api.batch.BatchSource;
+import co.cask.cdap.templates.etl.api.batch.BatchSourceContext;
+import co.cask.cdap.templates.etl.api.config.ETLStage;
+import co.cask.cdap.templates.etl.common.DBRecord;
+import co.cask.cdap.templates.etl.common.ETLDBInputFormat;
+import co.cask.cdap.templates.etl.common.Properties;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.mysql.jdbc.Driver;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.io.LongWritable;
+import org.apache.hadoop.mapreduce.Job;
+import org.apache.hadoop.mapreduce.lib.db.DBConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+/**
+ * Batch source to read from a Database table
+ */
+public class DBSource extends BatchSource<LongWritable, DBRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(DBSource.class);
+
+  // TODO: Remove when plugin support is enabled
+  Driver driver;
+
+  @Override
+  public void configure(StageConfigurer configurer) {
+    configurer.setDescription("Batch source for database");
+    configurer.addProperty(new Property(Properties.DB.DRIVER_CLASS, "Driver class to connect to the database", true));
+    configurer.addProperty(
+      new Property(Properties.DB.CONNECTION_STRING, "JDBC connection string including database name", true));
+    configurer.addProperty(
+      new Property(Properties.DB.USER,
+                   "User to use to connect to the specified database. " +
+                     "Required for databases that require authentication. " +
+                     "Optional for databases that do not require authentication.",
+                   false));
+    configurer.addProperty(
+      new Property(Properties.DB.PASSWORD,
+                   "Password to be use to connect to the specified database. " +
+                     "Required for databases that require authentication. " +
+                     "Optional for databases that do not require authentication.",
+                   false));
+    configurer.addProperty(new Property(Properties.DB.TABLE_NAME, "Table name to import", true));
+    configurer.addProperty(
+      new Property(Properties.DB.IMPORT_QUERY,
+                   "The SELECT query to use to import data from the specified table. " +
+                     "You can specify an arbitrary number of columns to import, or import all columns using *. " +
+                     "You can also specify a number of WHERE clauses or ORDER BY clauses. " +
+                     "However, LIMIT and OFFSET clauses should not be used in this query.",
+                   true));
+    configurer.addProperty(
+      new Property(Properties.DB.COUNT_QUERY,
+                   "The SELECT query to use to get the count of records to import from the specified table. " +
+                     "Examples: SELECT COUNT(*) from <my_table> where <my_column> 1, " +
+                     "SELECT COUNT(my_column) from my_table). " +
+                     "NOTE: Please include the same WHERE clauses in this query as the ones used in the " +
+                     "import query to reflect an accurate number of records to import.",
+                   true));
+  }
+
+  @Override
+  public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineConfigurer) {
+    Map<String, String> properties = stageConfig.getProperties();
+    String dbDriverClass = properties.get(Properties.DB.DRIVER_CLASS);
+    String dbConnectionString = properties.get(Properties.DB.CONNECTION_STRING);
+    String dbUser = properties.get(Properties.DB.USER);
+    String dbPassword = properties.get(Properties.DB.PASSWORD);
+    String dbTableName = properties.get(Properties.DB.TABLE_NAME);
+    String dbImportQuery = properties.get(Properties.DB.IMPORT_QUERY);
+    String dbCountQuery = properties.get(Properties.DB.COUNT_QUERY);
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(dbDriverClass), "dbDriverClass cannot be null");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(dbConnectionString), "dbConnectionString cannot be null");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(dbTableName), "dbTableName cannot be null");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(dbImportQuery), "dbImportQuery cannot be null");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(dbCountQuery), "dbCountQuery cannot be null");
+    Preconditions.checkArgument(!(dbUser == null && dbPassword != null),
+                                "dbUser is null. Please provide both user name and password if database requires " +
+                                  "authentication. If not, please remove dbPassword and retry.");
+    Preconditions.checkArgument(!(dbUser != null && dbPassword == null),
+                                "dbPassword is null. Please provide both user name and password if database requires" +
+                                  "authentication. If not, please remove dbUser and retry.");
+  }
+
+  @Override
+  public void prepareJob(BatchSourceContext context) {
+    Map<String, String> runtimeArguments = context.getRuntimeArguments();
+    String dbDriverClass = runtimeArguments.get(Properties.DB.DRIVER_CLASS);
+    String dbConnectionString = runtimeArguments.get(Properties.DB.CONNECTION_STRING);
+    String dbUser = runtimeArguments.get(Properties.DB.USER);
+    String dbPassword = runtimeArguments.get(Properties.DB.PASSWORD);
+    String dbTableName = runtimeArguments.get(Properties.DB.TABLE_NAME);
+    String dbImportQuery = runtimeArguments.get(Properties.DB.IMPORT_QUERY);
+    String dbCountQuery = runtimeArguments.get(Properties.DB.COUNT_QUERY);
+
+    LOG.debug("dbTableName = {}; dbDriverClass = {}; dbConnectionString = {}; dbImportQuery = {}; dbCountQuery = {}",
+              dbTableName, dbDriverClass, dbConnectionString, dbImportQuery, dbCountQuery);
+
+    Job job = context.getHadoopJob();
+    Configuration conf = job.getConfiguration();
+    job.setInputFormatClass(ETLDBInputFormat.class);
+    if (dbUser == null && dbPassword == null) {
+      DBConfiguration.configureDB(conf, dbDriverClass, dbConnectionString);
+    } else {
+      DBConfiguration.configureDB(conf, dbDriverClass, dbConnectionString, dbUser, dbPassword);
+    }
+    ETLDBInputFormat.setInput(job, DBRecord.class, dbImportQuery, dbCountQuery);
+    job.setInputFormatClass(ETLDBInputFormat.class);
+  }
+}
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
@@ -1,0 +1,353 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.common;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.data.schema.UnsupportedTypeException;
+import com.google.common.collect.Lists;
+import org.apache.hadoop.io.Writable;
+import org.apache.hadoop.mapreduce.lib.db.DBWritable;
+
+import java.io.BufferedReader;
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.sql.Blob;
+import java.sql.Clob;
+import java.sql.Date;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.sql.Types;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Writable class for DB Source/Sink
+ *
+ * {@see DBWritable}, {@see DBInputFormat}, {@see DBOutputFormat}
+ */
+public class DBRecord implements Writable, DBWritable {
+
+  private StructuredRecord record;
+
+  public void readFields(DataInput in) throws IOException {
+    // no-op, since we may never need to support a scenario where you read a DBRecord from a non-RDBMS source
+  }
+
+  /**
+   * @return the {@link StructuredRecord} contained in this object
+   */
+  public StructuredRecord getRecord() {
+    return record;
+  }
+
+  /**
+   * Builds the {@link #record} using the specified {@link ResultSet}
+   *
+   * @param resultSet the {@link ResultSet} to build the {@link StructuredRecord} from
+   */
+  public void readFields(ResultSet resultSet) throws SQLException {
+    List<Schema.Field> schemaFields = Lists.newArrayList();
+    ResultSetMetaData metadata = resultSet.getMetaData();
+    // ResultSetMetadata columns are numbered starting with 1
+    for (int i = 1; i <= metadata.getColumnCount(); i++) {
+      String columnName = metadata.getColumnName(i);
+      int columnSqlType = metadata.getColumnType(i);
+      Schema columnSchema = Schema.of(getType(columnSqlType));
+      if (ResultSetMetaData.columnNullable == metadata.isNullable(i)) {
+        columnSchema = Schema.nullableOf(columnSchema);
+      }
+      Schema.Field field = Schema.Field.of(columnName, columnSchema);
+      schemaFields.add(field);
+    }
+    Schema schema = Schema.recordOf("dbRecord", schemaFields);
+    StructuredRecord.Builder recordBuilder = StructuredRecord.builder(schema);
+    for (int i = 0; i < schemaFields.size(); i++) {
+      Schema.Field field = schemaFields.get(i);
+      int sqlColumnType = metadata.getColumnType(i + 1);
+      recordBuilder.set(field.getName(), transformValue(sqlColumnType, resultSet.getObject(field.getName())));
+    }
+    record = recordBuilder.build();
+  }
+
+  public void write(DataOutput out) throws IOException {
+    Schema recordSchema = record.getSchema();
+    List<Schema.Field> schemaFields = recordSchema.getFields();
+    for (int i = 0; i < schemaFields.size(); i++) {
+      Schema.Field field = schemaFields.get(i);
+      String fieldName = field.getName();
+      Schema.Type fieldType = field.getSchema().getType();
+      Object fieldValue = record.get(fieldName);
+      // In JDBC, field indices start with 1
+      writeToDataOut(out, fieldType, fieldValue);
+    }
+  }
+
+  /**
+   * Writes the {@link #record} to the specified {@link PreparedStatement}
+   *
+   * @param stmt the {@link PreparedStatement} to write the {@link StructuredRecord} to
+   */
+  public void write(PreparedStatement stmt) throws SQLException {
+    Schema recordSchema = record.getSchema();
+    List<Schema.Field> schemaFields = recordSchema.getFields();
+    for (int i = 0; i < schemaFields.size(); i++) {
+      Schema.Field field = schemaFields.get(i);
+      String fieldName = field.getName();
+      Schema.Type fieldType = field.getSchema().getType();
+      Object fieldValue = record.get(fieldName);
+      // In JDBC, field indices start with 1
+      writeToDB(stmt, fieldType, fieldValue, i + 1);
+    }
+  }
+
+  private Schema.Type getType(int sqlType) throws SQLException {
+    // Type.STRING covers sql types - case VARCHAR,CHAR,CLOB,LONGNVARCHAR,LONGVARCHAR,NCHAR,NCLOB,NVARCHAR
+    Schema.Type type = Schema.Type.STRING;
+    switch (sqlType) {
+      case Types.NULL:
+        type = Schema.Type.NULL;
+        break;
+
+      case Types.BOOLEAN:
+      case Types.BIT:
+        type = Schema.Type.BOOLEAN;
+        break;
+
+      case Types.TINYINT:
+      case Types.SMALLINT:
+      case Types.INTEGER:
+        type = Schema.Type.INT;
+        break;
+
+      case Types.BIGINT:
+        type = Schema.Type.LONG;
+        break;
+
+      case Types.REAL:
+      case Types.FLOAT:
+        type = Schema.Type.FLOAT;
+        break;
+
+      case Types.NUMERIC:
+      case Types.DECIMAL:
+      case Types.DOUBLE:
+        type = Schema.Type.DOUBLE;
+        break;
+
+      case Types.DATE:
+      case Types.TIME:
+      case Types.TIMESTAMP:
+        type = Schema.Type.LONG;
+        break;
+
+      case Types.BINARY:
+      case Types.VARBINARY:
+      case Types.LONGVARBINARY:
+      case Types.BLOB:
+        type = Schema.Type.BYTES;
+        break;
+
+      case Types.ARRAY:
+      case Types.DATALINK:
+      case Types.DISTINCT:
+      case Types.JAVA_OBJECT:
+      case Types.OTHER:
+      case Types.REF:
+      case Types.ROWID:
+      case Types.SQLXML:
+      case Types.STRUCT:
+        throw new SQLException(new UnsupportedTypeException("Unsupported SQL Type: " + sqlType));
+    }
+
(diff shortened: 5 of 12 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/DBRecordToByteArrayTransform.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.hadoop.io.LongWritable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+
+/**
+ * Transforms a record from a database table into a byte array containing the json representation of the record
+ */
+public class DBRecordToByteArrayTransform extends Transform<LongWritable, DBRecord, byte [], byte []> {
```
gokulavasan: Would it be better to have two separate Transforms:
DBRecord to StructuredRecord
StructuredRecordToByteArray

So that we can have other transforms like FilterTransform, ProjectionTransform inserted between the above two transforms and these transforms work on StructuredRecords.

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
```diff
@@ -63,6 +63,10 @@
       <artifactId>cdap-unit-test</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.hsqldb</groupId>
```
gokulavasan: Is this only for test? So should this have test scope?

cdap-api/src/main/java/co/cask/cdap/api/data/format/StructuredRecord.java
```diff
@@ -56,6 +56,10 @@ public Schema getSchema() {
     return (T) fields.get(fieldName);
   }
 
+  public Map<String, Object> getFields() {
+    return fields;
```
chtyim: The map is not immutable. Seems dangerous to expose it. If really needed, make the map immutable in constructor.

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
```diff
@@ -63,6 +63,10 @@
...
+      <artifactId>hsqldb</artifactId>
```
chtyim: I don't see this is added in the parent pom dependency management. Do I miss something?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/DBRecordToByteArrayTransform.java
bdmogal: Sure, I can do that. Also, in a real scenario, would we even need to transform to byte arrays? Not sure, but will split this into two for now.

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
bdmogal: yes, will add

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/DBRecordToByteArrayTransform.java
gokulavasan: Ya we would need transform to bytearrays for KeyValueTableSink.

cdap-api/src/main/java/co/cask/cdap/api/data/format/StructuredRecord.java
bdmogal: Thanks for catching! I wanted to spit out the entire structured record as a json from a ``Transform`` class. I'll make it immutable.

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
```diff
@@ -117,6 +117,14 @@
       <artifactId>twitter4j-stream</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-mapreduce-client-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
```
albertshau: how come this is being added?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/Properties.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Class to define property names for sources and sinks
+ */
+public final class Properties {
+
+  /**
+   * Class to hold properties for {@link DBSource} and {@link DBSink}
+   */
+  public static class DB {
+    public static final String DB_DRIVER_CLASS = "dbDriverClass";
```
albertshau: nit: DB_ seems redundant

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
```diff
@@ -0,0 +1,81 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(DBSource.class);
+
+  @Override
+  public void configure(StageConfigurer configurer) {
+    configurer.setName("DBBatchSource");
+    configurer.setDescription("Batch source for database");
+    configurer.addProperty(new Property(Properties.DB.DB_DRIVER_CLASS, "Driver class to connect to the database", true));
+    configurer.addProperty(new Property(Properties.DB.DB_CONNECTION_STRING, "JDBC connection string including database name", true));
+    configurer.addProperty(new Property(Properties.DB.DB_TABLE_NAME, "Table name to import", true));
+    configurer.addProperty(new Property(Properties.DB.DB_COLUMNS, "Comma-separated list of columns to import", false));
```
albertshau: should describe what happens if no value is given.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  public void prepareJob(BatchSourceContext context) {
+    Map<String, String> runtimeArguments = context.getRuntimeArguments();
+    String dbDriverClass = runtimeArguments.get(Properties.DB.DB_DRIVER_CLASS);
+    String dbConnectionString = runtimeArguments.get(Properties.DB.DB_CONNECTION_STRING);
+    String dbTableName = runtimeArguments.get(Properties.DB.DB_TABLE_NAME);
+    String dbColumns = runtimeArguments.get(Properties.DB.DB_COLUMNS);
+    Preconditions.checkArgument(dbDriverClass != null, "dbDriverClass cannot be null");
```
albertshau: this should be checked in configurePipeline() so that it fails at adapter creation instead of at runtime

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
```diff
@@ -0,0 +1,81 @@
...
+      countQueryColumn = "*";
+    } else {
+      countQueryColumn = dbColumns.substring(0, dbColumns.indexOf(","));
+    }
+
+    Job job = context.getHadoopJob();
+    Configuration conf = job.getConfiguration();
+    job.setInputFormatClass(DBInputFormat.class);
+    DBConfiguration.configureDB(conf, dbDriverClass, dbConnectionString);
+    DBInputFormat.setInput(job, DBRecord.class, String.format("SELECT %s FROM %s", dbColumns, dbTableName),
```
albertshau: it is possible to do sql injection with this.  does the input format have any way to protect against it?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
```diff
@@ -0,0 +1,135 @@
...
+      return Schema.Type.BOOLEAN;
+    } else if (Integer.class.getName().equals(columnType)) {
+      return Schema.Type.INT;
+    } else if (Long.class.getName().equals(columnType)) {
+      return Schema.Type.LONG;
+    } else if (Float.class.getName().equals(columnType)) {
+      return Schema.Type.FLOAT;
+    } else if (Double.class.getName().equals(columnType)) {
+      return Schema.Type.DOUBLE;
+    } else {
```
albertshau: what about binary columns? is there a standard type for them?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
```diff
@@ -0,0 +1,135 @@
...
+      throw new IllegalArgumentException("Unsupported datatype: " + columnType);
+    }
+  }
+
+  public void write(DataOutput out) throws IOException {
```
albertshau: what does this do

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/DBRecordToByteArrayTransform.java
albertshau: not sure if we would every transform to byte arrays.  I think we should leave it out if we dont have a use case for it.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
albertshau: also, this source will always read the entire table. What if somebody wants to read just a part of the table, and for that part to be different with each run?  For example, they have a column which is an auto-increment column and want to read a different range of ids each run?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/DBRecordToByteArrayTransform.java
bdmogal: So in this PR, I need it in a test because we only had ``KeyValueTable`` sink available, which requires byte arrays. Ideally, this ``Transform`` should be in the test class, but we can't do that right now, since all the source, sink and transform classes need to be registered in ``ETLBatchTemplate``. Maybe I'll add the transform with a note to remove it when we have plugin management enabled.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
albertshau: don't have to do it in this PR, but I think that should be a use case we target.

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
bdmogal: yeah, so without this, I had no way of testing with mysql. We need to package the jar with the ETL adapter somehow.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/Properties.java
bdmogal: removed

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Batch source to read from a Database table
+ */
+public class DBSource extends BatchSource<LongWritable, DBRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(DBSource.class);
+
+  @Override
+  public void configure(StageConfigurer configurer) {
+    configurer.setName("DBBatchSource");
```
rohitsinha54: How about using the classname here and in the unit test than a string ?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
```diff
@@ -0,0 +1,81 @@
...
+    configurer.setDescription("Batch source for database");
+    configurer.addProperty(new Property(Properties.DB.DB_DRIVER_CLASS, "Driver class to connect to the database", true));
+    configurer.addProperty(new Property(Properties.DB.DB_CONNECTION_STRING, "JDBC connection string including database name", true));
```
rohitsinha54: maybe an example of connection string

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: Added to the description.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
```diff
@@ -0,0 +1,81 @@
...
+    configurer.addProperty(new Property(Properties.DB.DB_TABLE_NAME, "Table name to import", true));
+    configurer.addProperty(new Property(Properties.DB.DB_COLUMNS, "Comma-separated list of columns to import", false));
+  }
+
```
rohitsinha54: no configurepipeline class ? you should at least check user inputs there if you assume the table exists and don't want to create it.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: will do.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
```diff
@@ -0,0 +1,81 @@
...
+  @Override
+  public void prepareJob(BatchSourceContext context) {
+    Map<String, String> runtimeArguments = context.getRuntimeArguments();
+    String dbDriverClass = runtimeArguments.get(Properties.DB.DB_DRIVER_CLASS);
+    String dbConnectionString = runtimeArguments.get(Properties.DB.DB_CONNECTION_STRING);
+    String dbTableName = runtimeArguments.get(Properties.DB.DB_TABLE_NAME);
+    String dbColumns = runtimeArguments.get(Properties.DB.DB_COLUMNS);
+    Preconditions.checkArgument(dbDriverClass != null, "dbDriverClass cannot be null");
+    Preconditions.checkArgument(dbConnectionString != null, "dbConnectionString cannot be null");
```
rohitsinha54: Shouldn't you check for empty string too ?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
```diff
@@ -0,0 +1,135 @@
...
+  private StructuredRecord record;
+
+  public void readFields(DataInput in) throws IOException {
+  }
+
+  public StructuredRecord getRecord() {
+    return record;
+  }
+
+  public void readFields(ResultSet resultSet) throws SQLException {
```
rohitsinha54: public method should have a java doc

cdap-api/src/main/java/co/cask/cdap/api/data/format/StructuredRecord.java
albertshau: making it immutable also means we can't add a set method for a field, which we may want to add for transforms that work on a single field.  I think it's better just to use get() and not expose a way to get all fields.

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
albertshau: hm ok. lets put a todo to remove once the plugin stuff is done.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/DBRecordToByteArrayTransform.java
albertshau: ah i see. or once we have a more natural sink, like a Table sink

bdmogal: yup.

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
bdmogal: we'll still need to package it though right? Since per my understanding the built in sources and sinks won't accept jar files? I'll add a note, and we can fix it when we work on packaging.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: Agreed. I think there may be a way of doing it through ``DBInputFormat``, but lets do that as a separate PR. https://issues.cask.co/browse/CDAP-2089

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
albertshau: we can't package it, as the user may want to use a specific version of the mysql connector, or may not want mysql at all.  The built-in source and sink doesn't actually require this since it uses jdbc, though it requires some sort of connector jar.  This is one of the reasons the plugin work is required.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLDBMapreduceTest.java
```diff
@@ -0,0 +1,161 @@
...
+                                   ));
+    ETLStage sink = new ETLStage("KVTableSink", ImmutableMap.of("name", "table1"));
+    ETLStage transform = new ETLStage("DBRecordToByteArrayTransform", ImmutableMap.<String, String>of());
+    List<ETLStage> transformList = Lists.newArrayList();
+    transformList.add(transform);
+    ETLBatchConfig adapterConfig = new ETLBatchConfig("", source, sink, transformList);
+    MockAdapterConfigurer adapterConfigurer = new MockAdapterConfigurer();
+    appTemplate.configureAdapter("myAdapter", adapterConfig, adapterConfigurer);
+    Map<String, String> mapReduceArgs = Maps.newHashMap();
+    for (Map.Entry<String, String> entry : adapterConfigurer.getArguments().entrySet()) {
```
rohitsinha54: See albert's comment here: 
https://github.com/caskdata/cdap/pull/2008/files#r28269348

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
bdmogal: Correct, but we won't package some default jars either?

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
bdmogal: you're right. adding it. how come nothing fails though?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLDBMapreduceTest.java
bdmogal: Fixed

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
bdmogal: Added

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
bdmogal: Fixed

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
bdmogal: Added a note

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: Fixed

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: I'd prefer not to. This is not a format that CDAP introduces. Its a well known JDBC connection string format.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: What do you mean by don't want to create it? We should never create a table in the source?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: Regarding sql injection, there is another API that ``DBInputFormat`` exposes. However, the signature for that is - 

```
public static void setInput(Job job,
                            Class<? extends DBWritable> inputClass,
                            String tableName,
                            String conditions,
                            String orderBy,
                            String... fieldNames)
```
We could use that, but it would enforce the user to always specify all columns (to send to ``fieldNames``. Thoughts?

However, even this format accepts conditions that get substituted in the ``WHERE`` clause.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
bdmogal: I'm unsure of how to handle them. Seems like most RDBMS systems expose **lob** types, but I'm unsure of how we would serialize/deserialize them.

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
bdmogal: Added to ``dependencyManagement`` in parent pom.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: Added ``configurePipeline``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: Fixed

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: Added a ``configurePipeline()`` method with these checks

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
bdmogal: So I presume that since we're writing to a DB, only implementing the ``write(PreparedStatement)`` method should be good enough. Other writables that directly extend from ``Writable`` have to implement this method, to write to the ``DataOutput``. I've tested that this works in unit test and standalone, and am about to test it on a cluster.

bdmogal:
Addressed all comments other than a couple of pending questions.

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
chtyim: The test scope should be specified in the parent pom as well.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
```diff
@@ -0,0 +1,94 @@
...
+
+    LOG.debug("dbTableName = {}; dbDriverClass = {}; dbConnectionString = {}; dbColumns = {}",
+              dbTableName, dbDriverClass, dbConnectionString, dbColumns);
+
+    String countQueryColumn;
+    if (dbColumns == null) {
+      dbColumns = "*";
+      countQueryColumn = "*";
+    } else {
+      countQueryColumn = dbColumns.substring(0, dbColumns.indexOf(","));
```
chtyim: Seems like this logic should be in `configurePipeline` so that it can fail the creation if the config is not right (e.g. dbColumns does not have ",").

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
chtyim: fieldNames is a vararg. So what happen if it is an empty array? Also, if the user doesn't provide column names, then dbColumns is `*`. In that case, we can generate the statement as `select * from` as a special case.

chtyim: Also, I just took a brief look into the DBInputFormat. Seems like even using the `setInput` method you mentioned, it just constructs the SQL statement using those arguments, but not really using prepared statement.

bdmogal: Also, @nitinmotgi suggested yesterday that like sqoop, we should allow users to provide a custom sql to import, with conditions and such. I think its possible, with an additional configuration parameter. Thoughts?

The concern I have is that then users could specify a random query (say across tables?) and it could lead to unexpected behavior. 

As opposed to that, how does the API in the above comment look? It takes a single table name, some conditions, an order by clause and field names. Is there anything we need that a custom query would provide that the above API won't?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Writable class for DB Source/Sink
+ *
+ * {@see DBWritable}, {@see DBInputFormat}, {@see DBOutputFormat}
+ */
+public class DBRecord implements Writable, DBWritable {
+
+  private StructuredRecord record;
+
+  public void readFields(DataInput in) throws IOException {
```
chtyim: Why empty?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+    Schema schema = Schema.recordOf("dbRecord", fields);
+    StructuredRecord.Builder recordBuilder = StructuredRecord.builder(schema);
+    for (Schema.Field field : fields) {
+      recordBuilder.set(field.getName(), resultSet.getObject(field.getName()));
+    }
+    record = recordBuilder.build();
+  }
+
+  public void write(DataOutput out) throws IOException {
```
chtyim: Why empty?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
```diff
@@ -0,0 +1,146 @@
...
+      Schema.Type fieldType = field.getSchema().getType();
+      Object fieldValue = record.get(fieldName);
+      writeValue(stmt, fieldType, fieldValue, jdbcFieldIndex);
+    }
+  }
+
+  private Schema.Type getType(String columnType) {
+    if (columnType == null) {
+      return Schema.Type.NULL;
+    } else if (String.class.getName().equals(columnType)) {
```
chtyim: No need to have else after return. Just have it as:

```java
if (columnType == null) {
  return Schema.Type.NULL;
}
if (..) {
  return ...;
}
```

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
```diff
@@ -0,0 +1,146 @@
...
+   * @param resultSet the {@link ResultSet} to build the {@link StructuredRecord} from
+   */
+  public void readFields(ResultSet resultSet) throws SQLException {
+    List<Schema.Field> fields = Lists.newArrayList();
+    ResultSetMetaData metadata = resultSet.getMetaData();
+    // ResultSetMetadata columns are numbered starting with 1
+    for (int i = 1; i <= metadata.getColumnCount(); i++) {
+      String columnName = metadata.getColumnName(i);
+      String columnType = metadata.getColumnClassName(i);
+      Schema.Field field = Schema.Field.of(columnName, Schema.of(getType(columnType)));
```
chtyim: Isn't that you have the check if the column type is nullable?

http://docs.oracle.com/javase/6/docs/api/java/sql/ResultSetMetaData.html#field_detail

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: @chtyim , right. internally they translate into the same set of methods I think. The ``setInput`` method above is just a convenience for setting some mapreduce configuration parameters in ``DBConfiguration``. The real parameters it needs are an input query and a count query.

Like in the previous comment, my only real concern is that we allow users to specify arbitrary queries in one API. The other (one above), even though its a convenience API, I think it helps make sure that the two queries are in sync (meaning the same filter conditions are applied to both the input query and the count query).

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
chtyim: What will be the class name returned by JDBC if the SQL type is BINARY or BLOB?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
bdmogal: yup. my bad. will fix.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
chtyim: Also, how about Byte and Short? They can map to Schema type INT, isn't it?

chtyim: Just looked at the jdbc source. Seems like BINARY, BLOB, VARBINARY and LONGVARBINARY, will become `byte[]`.
CBLO becomes `char[]` (hence can map to String schema). Also, for DATE, TIME and TIMESTAMP type, they will be of `java.util.Date` type, and we can represent them as `long` timestamp?

chtyim: BTW, for date type, it'll return subclass of `java.util.Date`. So make sure check for `isAssignable`.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
```diff
@@ -0,0 +1,146 @@
...
+    } else if (Double.class.getName().equals(columnType)) {
+      return Schema.Type.DOUBLE;
+    } else {
+      throw new IllegalArgumentException("Unsupported datatype: " + columnType);
+    }
+  }
+
+  private void writeValue(PreparedStatement stmt, Schema.Type fieldType, Object fieldValue,
+                          int fieldIndex) throws SQLException {
+    switch (fieldType) {
```
chtyim: We should support more types.

chtyim: Also, why use `if` in the read case, but `switch-case` in write case? I think it's better to be consistent (which might be easier to always use `switch-case`).

chtyim: oh. Those are strings in read case. Nevermind.

bdmogal: Only reason was that we use enums here and strings in the other method. maybe I can use ``if`` in both places.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
bdmogal: Fixed. Using a switch statement on ``ResultMetadata.getColumnType()`` instead.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/DBSource.java
bdmogal: Instead of accepting columns, accepting two queries from the user, so we don't need this check anymore.

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
bdmogal: Fixed.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
bdmogal: Like we discussed, we will not support the use-case for reading a DBRecord from a non-db source to begin with.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/DBRecord.java
bdmogal: fixed

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -252,7 +256,7 @@ private void configureTransforms(AdapterConfigurer configurer) throws Exception
 
   @Override
   public void configure() {
-    setName("etlbatch");
+    setName("etlbatch2");
```
albertshau: was this unintentional?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/StructuredRecordToPutTransform.java
```diff
@@ -154,4 +159,31 @@ private Put createPut(StructuredRecord record, Schema recordSchema) {
         throw new IllegalArgumentException("Row key is of unsupported type " + keyType);
     }
   }
+
+  private byte [] getBytes(Object key) throws IOException {
```
albertshau: this... sucks. Lets be sure to bring it up on the next pass through the APIs.

albertshau:
one comment, otherwise :+1:

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/StructuredRecordToPutTransform.java
bdmogal: agreed. completely.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
bdmogal: yea reverted

bdmogal:
Tested on

bdmogal:
Tested the following pipeline on a cluster:
Mysql Table as source feeding into a Table sink using a random SQL query as import query.

Build has been passing. Merging.

PR #2006, opened 2015-04-12
Allow adding system properties to a Schedule

gokulavasan:
We need to pass Adapter Name as a system property to the program that is invoked through the scheduler.

This is required for couple of reasons:
i) To get the Adapter Name to set the tag in the Logging Context (which is currently achieved by piggybacking on schedule name)
ii) ScheduleTaskRunner needs to get the Adapter Name so that it can record the Runs under the Adapter Name (in addition to the Program name).

Bamboo : http://builds.cask.co/browse/CDAP-DUT1383


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
```diff
@@ -45,6 +46,18 @@ public void schedule(Id.Program program, SchedulableProgramType programType, Sch
    *
    * @param program Program that needs to be run.
    * @param programType type of program.
+   * @param schedule Schedule with which the program runs.
+   * @param properties properties to be passed to the schedule
```
sagarkapare: These are only system properties for the Schedule, which are different from the properties that are associated with the Schedule when its created ,right?

gokulavasan: That is correct. Will fix the javadoc.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -119,14 +121,26 @@ void stop() throws SchedulerException {
...
   @Override
-  public void schedule(Id.Program programId, SchedulableProgramType programType, Schedule schedule)
+  public void schedule(Id.Program program, SchedulableProgramType programType, Schedule schedule)
     throws SchedulerException {
-    schedule(programId, programType, ImmutableList.of(schedule));
+    schedule(program, programType, schedule, ImmutableMap.<String, String>of());
+  }
+
+  @Override
+  public void schedule(Id.Program programId, SchedulableProgramType programType, Schedule schedule,
```
sagarkapare: Maybe we can change the name of the parameter to `program` here as well?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -119,14 +121,26 @@ void stop() throws SchedulerException {
...
   }
 
   @Override
   public void schedule(Id.Program programId, SchedulableProgramType programType, Iterable<Schedule> schedules)
     throws SchedulerException {
+    schedule(programId, programType, schedules, ImmutableMap.<String, String>of());
+  }
+
+  @Override
+  public void schedule(Id.Program programId, SchedulableProgramType programType, Iterable<Schedule> schedules,
```
sagarkapare: same here

sagarkapare:
Minor comments, otherwise LGTM :+1:

gokulavasan:
@sagarkapare Addressed comments. Will merge after build passes.

PR #2008, opened 2015-04-13
PartitionedFileset Source and Sink Implementation with tests

rohitsinha54:
Implements: 
1. PartitionedFilesetSource: https://issues.cask.co/browse/CDAP-2063
2. PartitionedFilesetSink: https://issues.cask.co/browse/CDAP-2065
3. Unit test with PartitionedFilesetSource as source and PartitionedFilesetSink as a sink

This currently does not set explore enabled while creating the filesets. We need a small discussion on it and I will update this PR after the discussion.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.batch;
+
+import co.cask.cdap.api.dataset.lib.Partition;
+import co.cask.cdap.api.dataset.lib.PartitionKey;
+import co.cask.cdap.api.dataset.lib.PartitionOutput;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSetProperties;
+import co.cask.cdap.api.dataset.lib.Partitioning;
+import co.cask.cdap.api.templates.ApplicationTemplate;
+import co.cask.cdap.templates.etl.api.config.ETLStage;
+import co.cask.cdap.templates.etl.batch.config.ETLBatchConfig;
+import co.cask.cdap.templates.etl.batch.sinks.PartitionedFileSetSink;
+import co.cask.cdap.templates.etl.batch.sources.PartitionedFileSetSource;
+import co.cask.cdap.test.ApplicationManager;
+import co.cask.cdap.test.DataSetManager;
+import co.cask.cdap.test.MapReduceManager;
+import co.cask.cdap.test.TestBase;
+import com.google.common.base.Charsets;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.gson.Gson;
+import org.apache.hadoop.mapreduce.lib.input.TextInputFormat;
+import org.apache.hadoop.mapreduce.lib.output.TextOutputFormat;
+import org.apache.twill.filesystem.Location;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Test for {@link PartitionedFileSetSource} and {@link PartitionedFileSetSink}
+ */
+public class ETLPartitionedFileSetSourceSinkTest extends TestBase {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ETLPartitionedFileSetSourceSinkTest.class);
+  private static final Gson GSON = new Gson();
+  private static final String sourceFileset = "inputFileset";
+  private static final String sinkFileset = "outputFileset";
+
+  private static final String DUMMY_RECORD = "2015/3/10,Red Falcons,Blue Bonnets,28,17\n";
+
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+    addDatasetInstance("partitionedFileSet", sourceFileset, PartitionedFileSetProperties.builder()
+      .setPartitioning(Partitioning.builder().addStringField("league").addIntField("season").build())
+      .setInputFormat(TextInputFormat.class)
+      .build());
+
+    addDatasetInstance("partitionedFileSet", sinkFileset, PartitionedFileSetProperties.builder()
+      .setPartitioning(Partitioning.builder().addStringField("league").build())
+      .setOutputFormat(TextOutputFormat.class)
+      .setOutputProperty(TextOutputFormat.SEPERATOR, ",")
+      .build());
+  }
+
+  @AfterClass
+  public static void afterClass() throws Exception {
+    clear();
+  }
+
+  @Test
+  public void testConfig() throws Exception {
+
+    writeToSource();
+
+    ApplicationManager batchManager = deployApplication(ETLBatchTemplate.class);
+
+    ApplicationTemplate<ETLBatchConfig> appTemplate = new ETLBatchTemplate();
+    ETLBatchConfig adapterConfig = constructETLBatchConfig();
+    MockAdapterConfigurer adapterConfigurer = new MockAdapterConfigurer();
+    appTemplate.configureAdapter("myAdapter", adapterConfig, adapterConfigurer);
+
+    Map<String, String> mapReduceArgs = Maps.newHashMap();
+    for (Map.Entry<String, String> entry : adapterConfigurer.getArguments().entrySet()) {
+      mapReduceArgs.put(entry.getKey(), entry.getValue());
+    }
+
+    mapReduceArgs.put("config", GSON.toJson(adapterConfig));
+    MapReduceManager mrManager = batchManager.startMapReduce("ETLMapReduce", mapReduceArgs);
+    mrManager.waitForFinish(5, TimeUnit.MINUTES);
+    batchManager.stopAll();
+
+    verySinkData();
+  }
+
+  private void verySinkData() throws Exception {
+    DataSetManager<PartitionedFileSet> sink = getDataset(sinkFileset);
+    PartitionedFileSet sinkPFS = sink.get();
+    Partition sinkPartition = sinkPFS.getPartition(PartitionKey.builder()
+                                                     .addStringField("league", "nfl")
+                                                     .build());
+
+    Assert.assertNotNull("Partition not found", sinkPartition);
+    try {
+      Location location = sinkPartition.getLocation();
+      for (Location file : location.list()) {
+        if (file.getName().startsWith("part")) {
+          BufferedReader reader = new BufferedReader(new InputStreamReader(location.getInputStream()));
+          Assert.assertEquals("The data in source and sink is not same", reader.readLine(), DUMMY_RECORD);
+        }
+      }
+    } catch (IOException e) {
+      LOG.info("Unable to read path {}", sinkPartition.getRelativePath());
+    }
+  }
+
+  private void writeToSource() throws Exception {
+    DataSetManager<PartitionedFileSet> table1 = getDataset(sourceFileset);
+    PartitionedFileSet inputFileset = table1.get();
+
+    PartitionKey key = PartitionKey.builder()
+      .addStringField("league", "nfl")
+      .addIntField("season", 1985)
+      .build();
+
+    if (inputFileset.getPartition(key) != null) {
+      LOG.info("Partition already exists");
+      throw new RuntimeException("Partition already exists");
+    }
+    PartitionOutput output = inputFileset.getPartitionOutput(key);
+
+    try {
+      Location location = output.getLocation();
+      OutputStream outputStream = location.getOutputStream();
+      try {
+        outputStream.write(DUMMY_RECORD.getBytes(Charsets.UTF_8));
+      } finally {
+        outputStream.close();
+      }
+    } catch (IOException e) {
+      LOG.warn(String.format("Unable to write path '%s'", output.getRelativePath()));
+      throw e;
+    }
+    output.addPartition();
+    table1.flush();
+  }
+
+  private ETLBatchConfig constructETLBatchConfig() {
+    ETLStage source = new ETLStage(PartitionedFileSetSource.class.getSimpleName(),
+                                   ImmutableMap.of("filesetName", sourceFileset, "filter",
+                                                   "{\"league.value\": \"nfl\",\"season.lower\": " +
+                                                     "\"1980\",\"season.upper\": \"1990\"}"));
+    ETLStage sink = new ETLStage(PartitionedFileSetSink.class.getSimpleName(),
+                                 ImmutableMap.of("filesetName", sinkFileset, "outputPath", sinkFileset,
+                                                 "partitionKey", "{\"league\" : \"nfl\"}"));
+    List<ETLStage> transformList = Lists.newArrayList();
+    return new ETLBatchConfig("", source, sink, transformList);
+  }
+}
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.batch.sinks;
+
+import co.cask.cdap.api.dataset.lib.PartitionKey;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSetArguments;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSetProperties;
+import co.cask.cdap.templates.etl.api.PipelineConfigurer;
+import co.cask.cdap.templates.etl.api.Property;
+import co.cask.cdap.templates.etl.api.StageConfigurer;
+import co.cask.cdap.templates.etl.api.batch.BatchSink;
+import co.cask.cdap.templates.etl.api.batch.BatchSinkContext;
+import co.cask.cdap.templates.etl.api.config.ETLStage;
+import co.cask.cdap.templates.etl.common.ETLUtils;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.collect.Maps;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import org.apache.hadoop.io.LongWritable;
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.mapreduce.lib.input.TextInputFormat;
+import org.apache.hadoop.mapreduce.lib.output.TextOutputFormat;
+
+import java.lang.reflect.Type;
+import java.util.Map;
+
+/**
+ * CDAP {@link PartitionedFileSet} Batch Sink
+ */
+public class PartitionedFileSetSink extends BatchSink<LongWritable, Text> {
+
+  private static final String FILESET_NAME = "filesetName";
+  private static final String PARTITION_KEY = "partitionKey";
+  private static final String PARTITIONING = "partitioning";
+
+  @Override
+  public void configure(StageConfigurer configurer) {
+    configurer.setName(PartitionedFileSetSink.class.getSimpleName());
+    configurer.setDescription("CDAP Partitioned fileSet batch sink");
+    configurer.addProperty(new Property(FILESET_NAME, "Partitioned fileset name", true));
+    configurer.addProperty(new Property(PARTITION_KEY, "Partition key in JSON format to use while writing " +
+      "ex: {\"league\" : \"nfl\", \"season\" : \"1980\"}", true));
+    configurer.addProperty(new Property(PARTITIONING, "Partitioning to use for the Partitioned FileSet in JSON " +
+      "format. Should be provided if you want the pipeline to create the fileset. ex: " +
+      "{\"league\" : \"string\", \"season\" : \"int\"}", false));
+    configurer.addProperty(new Property(TextOutputFormat.SEPERATOR, "The output format separator. Must be provided " +
+      "if you want the pipeline to create the fileset", false));
+  }
+
+  @Override
+  public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineConfigurer) {
+    // if the partitioning is provided then we should try to create the fileset here
+    if (!Strings.isNullOrEmpty(stageConfig.getProperties().get(PARTITIONING))) {
+      String pfsName = stageConfig.getProperties().get(FILESET_NAME);
+      String separator = stageConfig.getProperties().get(TextOutputFormat.SEPERATOR);
+      Preconditions.checkArgument(!Strings.isNullOrEmpty(pfsName), "PartitionedFileSet name must be given.");
+      Preconditions.checkArgument(!Strings.isNullOrEmpty(pfsName), "A separator for " +
+        TextOutputFormat.SEPERATOR + "must be given");
+      pipelineConfigurer.createDataset(pfsName, PartitionedFileSet.class.getName(),
+                                       PartitionedFileSetProperties.builder()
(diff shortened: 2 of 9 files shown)


cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    mapReduceArgs.put("config", GSON.toJson(adapterConfig));
+    MapReduceManager mrManager = batchManager.startMapReduce("ETLMapReduce", mapReduceArgs);
+    mrManager.waitForFinish(5, TimeUnit.MINUTES);
+    batchManager.stopAll();
+
+    verySinkData();
+  }
+
+  private void verySinkData() throws Exception {
```
albertshau: typo: very -> verify

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
```diff
@@ -0,0 +1,176 @@
...
+    ApplicationManager batchManager = deployApplication(ETLBatchTemplate.class);
+
+    ApplicationTemplate<ETLBatchConfig> appTemplate = new ETLBatchTemplate();
+    ETLBatchConfig adapterConfig = constructETLBatchConfig();
+    MockAdapterConfigurer adapterConfigurer = new MockAdapterConfigurer();
+    appTemplate.configureAdapter("myAdapter", adapterConfig, adapterConfigurer);
+
+    Map<String, String> mapReduceArgs = Maps.newHashMap();
+    for (Map.Entry<String, String> entry : adapterConfigurer.getArguments().entrySet()) {
+      mapReduceArgs.put(entry.getKey(), entry.getValue());
```
albertshau: you can use the putAll method instead of iterating through all entries

albertshau: you can also just pass the map into Maps.newHashMap

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
```diff
@@ -0,0 +1,176 @@
...
+                                                     .addStringField("league", "nfl")
+                                                     .build());
+
+    Assert.assertNotNull("Partition not found", sinkPartition);
+    try {
+      Location location = sinkPartition.getLocation();
+      for (Location file : location.list()) {
+        if (file.getName().startsWith("part")) {
+          BufferedReader reader = new BufferedReader(new InputStreamReader(location.getInputStream()));
+          Assert.assertEquals("The data in source and sink is not same", reader.readLine(), DUMMY_RECORD);
```
albertshau: if surprised this passes.  Usually there are several empty part files along with the one with data.  Should make sure you're not reading empty data before checking this.

albertshau: should close the reader too

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
```diff
@@ -0,0 +1,110 @@
...
+    if (!Strings.isNullOrEmpty(stageConfig.getProperties().get(PARTITIONING))) {
+      String pfsName = stageConfig.getProperties().get(FILESET_NAME);
+      String separator = stageConfig.getProperties().get(TextOutputFormat.SEPERATOR);
+      Preconditions.checkArgument(!Strings.isNullOrEmpty(pfsName), "PartitionedFileSet name must be given.");
+      Preconditions.checkArgument(!Strings.isNullOrEmpty(pfsName), "A separator for " +
+        TextOutputFormat.SEPERATOR + "must be given");
+      pipelineConfigurer.createDataset(pfsName, PartitionedFileSet.class.getName(),
+                                       PartitionedFileSetProperties.builder()
+        .setPartitioning(ETLUtils.createPartitioning(stageConfig.getProperties().get(PARTITIONING)))
+        .setInputFormat(TextInputFormat.class)
```
albertshau: as discussed, these should all be properties that the user sets when creating a pipeline

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
```diff
@@ -0,0 +1,110 @@
...
+    context.setOutput(context.getRuntimeArguments().get(FILESET_NAME), input);
+  }
+
+  /**
+   * Builds a {@link PartitionKey} from user supplied JSON string.
+   *
+   * @param paritionKeyString the partition key JSON string
+   * @return {@link PartitionKey}
+   */
+  private PartitionKey buildParitionKey(String paritionKeyString) {
```
albertshau: typo: Parition -> Partition

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
```diff
@@ -0,0 +1,110 @@
...
+    PartitionKey.Builder builder = PartitionKey.builder();
+    Type stringStringMap = new TypeToken<Map<String, String>>() {
```
albertshau: this type and the Gson instance can both be private static final variables

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
```diff
@@ -0,0 +1,110 @@
...
+import java.lang.reflect.Type;
+import java.util.Map;
+
+/**
+ * CDAP {@link PartitionedFileSet} Batch Sink
+ */
+public class PartitionedFileSetSink extends BatchSink<Text, Text> {
+
+  private static final String FILESET_NAME = "filesetName";
+  private static final String PARTITION_KEY = "partitionKey";
```
albertshau: it is more accurate for this to be plural, "partitionKeys"

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
albertshau: also more accurate for this to be plural PartitionKeys

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
```diff
@@ -0,0 +1,110 @@
...
+    configurer.addProperty(new Property(PARTITIONING, "Partitioning to use for the Partitioned FileSet in JSON " +
+      "format. Should be provided if you want the pipeline to create the fileset. ex: " +
+      "{\"league\" : \"string\", \"season\" : \"int\"}", false));
+    configurer.addProperty(new Property(TextOutputFormat.SEPERATOR, "The output format separator. Must be provided " +
+      "if you want the pipeline to create the fileset", false));
+  }
+
+  @Override
+  public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineConfigurer) {
+    // if the partitioning is provided then we should try to create the fileset here
```
albertshau: should also check that required fields are not empty (like name and partition key)

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/PartitionedFileSetSource.java
```diff
@@ -0,0 +1,127 @@
...
+      "Must be provided if you want the pipeline to create the fileset. ex: " +
+      "{\"league\" : \"string\", \"season\" : \"int\"}", false));
+    configurer.addProperty(new Property(TextOutputFormat.SEPERATOR, "The output format separator must be provided " +
+      "if you want the pipeline to create the fileset", false));
+
+  }
+
+  @Override
+  public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineConfigurer) {
+    // if the partitioning is provided then we should try to create the fileset here
```
albertshau: should also check that require fields like name are not missing

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/PartitionedFileSetSource.java
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * creates a {@link PartitionFilter} from a given string in JSON format
+   *
+   * @param filterString partition filter JSON string
+   * @return {@link PartitionFilter}
+   */
+  private PartitionFilter createPartitionFilter(String filterString) {
+    PartitionFilter.Builder builder = PartitionFilter.builder();
+    Type stringStringMap = new TypeToken<Map<String, String>>() {
```
albertshau: type and gson can be private static final variables

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/PartitionedFileSetSource.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * CDAP {@link PartitionedFileSet} Batch Source
+ */
+public class PartitionedFileSetSource extends BatchSource<LongWritable, Text> {
+
+  private static final String FILESET_NAME = "filesetName";
+  private static final String FILTER = "filter";
+  private static final String PARTITIONING = "partitioning";
+  private static final String VALUE = "value";
```
albertshau: should be '.value'

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/PartitionedFileSetSource.java
```diff
@@ -0,0 +1,127 @@
...
+  private static final String LOWER = "lower";
```
albertshau: should be '.lower'

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/PartitionedFileSetSource.java
```diff
@@ -0,0 +1,127 @@
...
+  private static final String UPPER = "upper";
```
albertshau: should be '.upper'

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/PartitionedFileSetSource.java
```diff
@@ -0,0 +1,127 @@
...
+   */
+  private PartitionFilter createPartitionFilter(String filterString) {
+    PartitionFilter.Builder builder = PartitionFilter.builder();
+    Type stringStringMap = new TypeToken<Map<String, String>>() {
+    }.getType();
+    Map<String, String> filters = new Gson().fromJson(filterString, stringStringMap);
+    for (Map.Entry<String, String> entrySet : filters.entrySet()) {
+      String key = entrySet.getKey();
+      if (key.endsWith(VALUE)) {
+        builder.addValueCondition(key.split("\\.")[0], entrySet.getValue());
```
albertshau: if it is valid for a partition key to contain a '.', this will break.  It is better to find the last to take a substring from the beginning to the end - length of VALUE

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/PartitionedFileSetSource.java
```diff
@@ -0,0 +1,127 @@
...
+      } else if (key.endsWith(LOWER)) {
+        String fieldName = key.split("\\.")[0];
```
albertshau: same comment as above

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/PartitionedFileSetSource.java
```diff
@@ -0,0 +1,127 @@
...
+        // if the upper range is found for  this lowe range add it to PartitionFilter else just ignore
+        String upperRange = Joiner.on(".").join(fieldName, UPPER);
+        if (filters.containsKey(upperRange)) {
```
albertshau: seems like we should error out if there is no upper range instead of silently ignoring it

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/ETLUtils.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.common.base.Strings;
+import com.google.common.base.Throwables;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+
+import java.lang.reflect.Type;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
```
albertshau: type and gson can be private static final instead of creating each time

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
```diff
@@ -123,6 +124,12 @@ public void deployDatasetModule(Id.Namespace namespace, String moduleName,
   }
 
   @Override
+  public <T> DataSetManager<T> getDataset(Id.Namespace namespace, String datasetInstanceName,
```
albertshau: why is this needed?

albertshau: ok i see, added to superclass. ignore this comment.

albertshau:
something is still missing from these... If you use the source or sink it will read the same data each time the workflow runs.  So for the sink, it will run once and then fail every time after that. And for the source, it will read the same data every time which does not seem useful.  In order for these to be useful, partitions must change in between runs.  The most common case I can think of is that there is some time based partition. What is the plan for this?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
```diff
@@ -0,0 +1,176 @@
...
+    Partition sinkPartition = sinkPFS.getPartition(PartitionKey.builder()
+                                                     .addStringField("league", "nfl")
+                                                     .build());
+
+    Assert.assertNotNull("Partition not found", sinkPartition);
+    try {
+      Location location = sinkPartition.getLocation();
+      for (Location file : location.list()) {
+        if (file.getName().startsWith("part")) {
+          BufferedReader reader = new BufferedReader(new InputStreamReader(location.getInputStream()));
```
chtyim: This is wrong. You always need a charset when creating `Reader` from `InputStream`. Also, better use:

`CharStreams.readFirstLine(CharStreams.newReaderSupplier(Locations.newInputSupplier(location), Charsets.UTF_8))`

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    if (inputFileset.getPartition(key) != null) {
+      LOG.info("Partition already exists");
+      throw new RuntimeException("Partition already exists");
+    }
+    PartitionOutput output = inputFileset.getPartitionOutput(key);
+
+    try {
+      Location location = output.getLocation();
+      OutputStream outputStream = location.getOutputStream();
```
chtyim: The whole block can be replaced with one line:
```java
CharStreams.write(DUMMY_RECORD, CharStreams.newWriterSupplier(Locations.newOutputSupplier(location), Charsets.UTF_8));
```

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
```diff
@@ -0,0 +1,176 @@
...
+      try {
+        outputStream.write(DUMMY_RECORD.getBytes(Charsets.UTF_8));
+      } finally {
+        outputStream.close();
+      }
+    } catch (IOException e) {
+      LOG.warn(String.format("Unable to write path '%s'", output.getRelativePath()));
```
chtyim: When logging an exception, always log with the exception as the last argument, otherwise there is no stacktrace for debug. Also, seems like you don't need to log and rethrow here. The junit show already be catching and logging and fail the test.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
```diff
@@ -0,0 +1,176 @@
...
+          Assert.assertEquals("The data in source and sink is not same", reader.readLine(), DUMMY_RECORD);
+        }
+      }
+    } catch (IOException e) {
+      LOG.info("Unable to read path {}", sinkPartition.getRelativePath());
+    }
+  }
+
+  private void writeToSource() throws Exception {
+    DataSetManager<PartitionedFileSet> table1 = getDataset(sourceFileset);
```
chtyim: It is not a "table". Call the variable appropriately.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLPartitionedFileSetSourceSinkTest.java
```diff
@@ -0,0 +1,176 @@
...
+      throw e;
+    }
+    output.addPartition();
+    table1.flush();
+  }
+
+  private ETLBatchConfig constructETLBatchConfig() {
+    ETLStage source = new ETLStage(PartitionedFileSetSource.class.getSimpleName(),
+                                   ImmutableMap.of("filesetName", sourceFileset, "filter",
+                                                   "{\"league.value\": \"nfl\",\"season.lower\": " +
```
chtyim: Instead of hand writing Json, why not use Gson?

```java
GSON.toJson(ImmutableMap.of("league.value", "nfl", "season.lower", "1980", "season.upper", "1990"))
```

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLStreamConversionTest.java
```diff
@@ -0,0 +1,161 @@
...
+    transformList.add(streamToStructuredRecord);
+    transformList.add(structuredRecordToGeneric);
+    transformList.add(genericToAvro);
+    return new ETLBatchConfig("", source, sink, transformList);
+  }
+
+  private List<GenericRecord> readOutput(TimePartitionedFileSet fileSet, Schema schema) throws IOException {
+    org.apache.avro.Schema avroSchema = new org.apache.avro.Schema.Parser().parse(schema.toString());
+    DatumReader<GenericRecord> datumReader = new GenericDatumReader<GenericRecord>(avroSchema);
+    List<GenericRecord> records = com.google.common.collect.Lists.newArrayList();
```
chtyim: why full name? You have other `Lists`?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLStreamConversionTest.java
```diff
@@ -0,0 +1,161 @@
...
+      // this level should be the day (ex: 2015-01-19)
+      for (Location timeLoc : dayLoc.list()) {
+        // this level should be the time (ex: 21-23.1234567890000)
+        for (Location file : timeLoc.list()) {
+          // this level should be the actual mapred output
+          String locName = file.getName();
+
+          if (locName.endsWith(".avro")) {
+            DataFileStream<GenericRecord> fileStream =
+              new DataFileStream<GenericRecord>(file.getInputStream(), datumReader);
```
chtyim: Close stream after used in finally block.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLStreamConversionTest.java
```diff
@@ -0,0 +1,161 @@
...
+            while (fileStream.hasNext()) {
```
chtyim: Just use `Iterables.addAll(records, fileStream)`.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/PartitionedFileSetSink.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final String PARTITION_KEY = "partitionKey";
+  private static final String PARTITIONING = "partitioning";
+
+  @Override
+  public void configure(StageConfigurer configurer) {
+    configurer.setName(PartitionedFileSetSink.class.getSimpleName());
+    configurer.setDescription("CDAP Partitioned fileSet batch sink");
+    configurer.addProperty(new Property(FILESET_NAME, "Partitioned fileset name", true));
+    configurer.addProperty(new Property(PARTITION_KEY, "Partition key in JSON format to use while writing " +
+      "ex: {\"league\" : \"nfl\", \"season\" : \"1980\"}", true));
```
chtyim: Same comment as above. Use GSON to generate json string. don't hand write it.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/ETLUtils.java
```diff
@@ -0,0 +1,89 @@
...
+ * An Utility class for ETL
+ */
+public class ETLUtils {
```
chtyim: Util class must be `final` with a private constructor.

chtyim: Also, I don't like a class call `XXXUtils`, as it means you can put anything there, which is super bad (like a random collection of stuff, which is a signal for bad OO design sometime).

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/ETLUtils.java
```diff
@@ -0,0 +1,89 @@
...
+    //TODO: replace with TimeMathParser (available only internal to cdap)
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(frequencyStr));
+    frequencyStr = frequencyStr.toLowerCase();
+
+    String value = frequencyStr.substring(0, frequencyStr.length() - 1);
+    int parsedValue = 0;
+    try {
+      parsedValue = Integer.parseInt(value);
+    } catch (NumberFormatException nfe) {
+      Throwables.propagate(nfe);
```
chtyim: Better write `throw Throwables.propagate(nfe)`.

rohitsinha54:
@albertshau : For a time based partition we have the TimeParitionedFilset and the sink for the same. Is it not something which is more appropriate for it?  If we take a step back and look at it we had the same issue with FileSets as there was no way in the current ETL architecture to write to a new inputPath file in every run as the parameters are passed just once. I brought this up in the Adapter room and we made a consensus that FileSet  are very low level and will not be so much useful which made us switch to PartitionedFileSet: 

```
[Apr-9 12:42 PM] Rohit Sinha: @terence : Not we are using FileSets to make a FileSetSource.
[Apr-9 12:47 PM] Rohit Sinha: No*
[Apr-9 12:52 PM] Terence Yim: Why not use the ParititonFileSet?
[Apr-9 12:52 PM] Invisible Joltie: +!
[Apr-9 12:52 PM] Terence Yim: or that should be a different sink?
[Apr-9 12:52 PM] Terence Yim: the raw FileSet is very limited in feature.
[Apr-9 12:52 PM] Terence Yim: and PartitionedFileSet should be used in most cases.
[Apr-9 12:53 PM] Terence Yim: The FileSet is mainly for lower level usage.. e.g. you already have some existing file on HDFS and wanted to use them in CDAP... or for being used inside a PartitionedFileSet.
[Apr-9 1:31 PM] Rohit Sinha: The requirement was to create FileSetSource, FileSetSink and PartitionedFileSetSink. I am not sure as how we came up with these.
[Apr-9 1:31 PM] Rohit Sinha: @sree ^^
[Apr-9 1:31 PM] Rohit Sinha: Should we just do Partitioned then ?
[Apr-9 1:36 PM] Sreevatsan Raman: the requirement is to integrate with Datasets. If we think FileSets doesn't make sense then let's do paritioned file sets
[Apr-9 2:01 PM] Terence Yim: Remember, the goal is to create something useful...
[Apr-9 2:02 PM] Terence Yim: not just to create a collection of semi-useful/not-useful stuff for our users.
```

More or less with this too we are at same issue. The use case which I see here with PartitionFilesetSource and Sink is to have Workflow triggered by new partition being created.

rohitsinha54:
So after discussion with @sreevatsanraman and @albertshau we decided to stay put on the PFS source and sink tell we fix the underlying architecture which will enable us to write to existing partition when new data comes in through MR or if ETL can support changing arguments in consecutive runs.
We will be closing this PR for now and will come back to it when we have architecture changes

PR #2010, opened 2015-04-13
Correct the grammar of Command Line Interface messages and its test.

awholegunch:
Corrects the tense: "send" -> "sent". Adjusted in both output and in the test that depends on the output.


jwang47:
LGTM

PR #2011, opened 2015-04-13
CDAP-1961: Remove Exception from Cube and related classes APIs

abaranau:
To be consistent with Dataset APIs


abaranau:
http://builds.cask.co/browse/CDAP-DUT1412 GREEN

gsps1:
LGTM

PR #2012, opened 2015-04-13
(CDAP-1917) Added classes for representing plugin version.

chtyim:
This is the first PR for CDAP-1917 by adding a new class in API to represent version information for template plugin.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginVersion.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginVersion.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginVersion.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * Constructs an instance by parsing the given string.
+   *
+   * @param str the version stirng.
+   * @param matchSuffix if {@code true}, try to match the version pattern by the suffix of the string. Otherwise match
+   *                    the whole string.
+   */
+  public PluginVersion(String str, boolean matchSuffix) {
```
gokulavasan: Didn't find matchSuffix to be true in any of the tests. Is that intended?

chtyim: Oh, this constructor was used in other places in the plugin system, which is not included in this PR.
Sure, I'll add some test on it in this PR.

gokulavasan:
LGTM :+1:

chtyim:
Squashed commit into one to prepare for merge.

chtyim:
Since this PR is only adding a new class and a new test class which are not used anywhere in existing code, there is no way to break existing code. All tests passed in the new test class introduced. Merging this PR.

PR #2016, opened 2015-04-13
Fix a badly designed internal ConfigResponse interface

chtyim:
- The old design made leaking file descriptor easy.
- All existing usage does leak file descriptor if the underlying Reader does not come from a String


abaranau:
LGTM

chtyim:
Build: http://builds.cask.co/browse/CDAP-DUT1438-1
1 test failed but not related to this PR changes and that was fixed on develop separately.

PR #2019, opened 2015-04-14
Adapter Runs Management in MDS

gokulavasan:
Adding Adapter Name to the RunRecord. We can then query the Runs of an Adapter (but need to know the Program Id associated with that Adapter for the query). Also register the runs started through the Scheduler as Adapter runs if a non-null Adapter Name is passed in as System Property.

Bamboo : http://builds.cask.co/browse/CDAP-DT12


cdap-proto/src/main/java/co/cask/cdap/proto/RunRecord.java
```diff
@@ -35,15 +37,27 @@
   @SerializedName("status")
   private final ProgramRunStatus status;
 
-  public RunRecord(String pid, long startTs, Long stopTs, ProgramRunStatus status) {
+  @SerializedName("adapter")
+  private final String adapterName;
```
sagarkapare: May not be a good idea to include adapter in this class, since all program runs will start showing adapter in the run records whether they were started from adapter or not, right? Possible to have class `AdapterRunRecord` which would extend the `RunRecord` to include the adapter info in it?

chtyim: I think gson won't serialize null field (unless told to do so)

gokulavasan: https://sites.google.com/site/gson/gson-user-guide#TOC-Null-Object-Support

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
```diff
@@ -174,6 +176,14 @@ private void recordProgramSuspendResume(Id.Program program, String pid, String a
       throw new IllegalArgumentException(msg);
     }
 
+    // If RunRecord has a valid adapterName, then it should match the adapter name passed in.
+    if (record.getAdapterName() != null && (!record.getAdapterName().equals(adapter))) {
```
albertshau: how can this happen?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
```diff
@@ -304,46 +361,77 @@ public boolean apply(RunRecord input) {
...
+
     MDSKey start = new MDSKey.Builder(historyKey).add(getInvertedTsScanKeyPart(endTime)).build();
     MDSKey stop = new MDSKey.Builder(historyKey).add(getInvertedTsScanKeyPart(startTime)).build();
     if (status.equals(ProgramRunStatus.ALL)) {
       //return all records (successful and failed)
-      return list(start, stop, RunRecord.class, limit, Predicates.<RunRecord>alwaysTrue());
+      return list(start, stop, RunRecord.class, limit, new Predicate<RunRecord>() {
+        @Override
+        public boolean apply(@Nullable RunRecord record) {
+          boolean normalCheck = true;
```
albertshau: nit: don't need this variable, can just return true at the bottom and return the condition inside the if.

albertshau:
minor comment and a question about why we need the record adapter checks

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
gokulavasan: Since this logic is in recordProgramStop, I will use realtime adapter scenario:
Say someone starts a realtime adapter and this will start a Worker and register a RunRecord with null as adapterName (this will also change the status of the Worker driver to RUNNING on the UI aka through Program status REST APIs). If someone hits stop on the UI or issues REST call to the Worker Driver program, then this scenario might occur. Though in this case, passed in adapterName will be `null` instead of another adapter name. I can't think of a way where both adapter names are not-null. Also, I can't think of a scenario where RunRecord has a null for adapterName but passed in adapter name is not-null.

albertshau:
:+1: though lets be sure to handle what the right behavior should be if you stop a program that has multiple runs going in a future PR. Can you open a jira for that?

gokulavasan:
@albertshau Opened JIRA : https://issues.cask.co/browse/CDAP-2111. Will merge this PR once the build completes. (http://builds.cask.co/browse/CDAP-DT12-4)

PR #1921, opened 2015-04-06
StreamBatchSource and TPFSSink

rohitsinha54:
Initial implementation of StreamBatchSource and TPFSSink. 
Also added an integration test which tests the pipeline below:

StreamBatchSource -> StreamEventToStructuredRecordTransform -> StructuredRecordToAvroTransform -> TPFS

Issues: 
https://issues.cask.co/browse/CDAP-1945
https://issues.cask.co/browse/CDAP-1947

PR #2020, opened 2015-04-14
Relax ETL Batch Adapter Validation

gokulavasan:
Check for two cases - if types are Parameterized check only the rawType, if types are classes then check assignable, else pass validation.

If TypeToken's isAssignableFrom is directly used on Parametrized types, it checks for the assignability of Type Arguments as well, and this fails validation of Pipeline used in #1921.

Fix required to unblock #1921 

Bamboo : http://builds.cask.co/browse/CDAP-DT13

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
@@ -37,10 +37,8 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import com.google.common.reflect.TypeToken;
 import com.google.gson.Gson;
 
-import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


gokulavasan:
Removed the validation logic. Will fix it properly in another PR. This PR needs to be merged to unblock #1921.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -140,8 +140,8 @@ private void validateAdapter(ETLStage source, ETLStage sink, List<ETLStage> tran
     throws IllegalArgumentException {
     if (transformList.size() == 0) {
       // No transforms. Check only source and sink.
-      if (!(TypeToken.of(batchSink.getKeyType()).isAssignableFrom(batchSource.getKeyType()) &&
-        TypeToken.of(batchSink.getValueType()).isAssignableFrom(batchSource.getValueType()))) {
+      if (!(isAssignable(batchSource.getKeyType(), batchSink.getKeyType()) &&
```
rohitsinha54: Why does the IDE complains that the condition will always be false ?

gokulavasan: That is because isAssignable is currently always returning true. I have a TODO there to fix it. So basically no validation is being done currently.

rohitsinha54: Then why no  just comment out all these and bring it in when you fix it. Its confusing to have it on develop

albertshau:
:+1:

PR #2021, opened 2015-04-14
add mappers and reducers tasks list

elfenheart:
- List view of Mappers and Reducers tasks.
- Remove Status Page
- Change behavior of Runs tab to match Mock design
- Add Angular-UI-Select for Run ID selector
- Change Start a new run button color and behavior to be able to have concurrent runs

screen shot 2015-04-14 at 2 47 21 pm

screen shot 2015-04-15 at 3 42 21 pm


ajainarayanan:
:+1: LGTM

PR #2022, opened 2015-04-14
CDAP-2088 Adding a TableSource for the ETL template. Also adding

albertshau:
a BatchReadableSource that the others extend so that there is a
catch-all for advanced users.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/common/MockEmitter.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/common/MockEmitter.java
@@ -17,30 +17,44 @@
 package co.cask.cdap.templates.etl.common;
 
 import co.cask.cdap.templates.etl.api.Emitter;
-import com.google.common.collect.Maps;
+import com.google.common.collect.Lists;
 
-import java.util.Map;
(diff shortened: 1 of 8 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DT14

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/common/MockEmitter.java
```diff
@@ -0,0 +1,46 @@
...
+import co.cask.cdap.templates.etl.api.Emitter;
+import com.google.common.collect.Maps;
+
+import java.util.Map;
+
+/**
+ * Mock emitter for unit tests
+ */
+public class MockEmitter<K, V> implements Emitter<K, V> {
+  private final Map<K, V> emitted = Maps.newHashMap();
```
gokulavasan: Technically this should be List? I can emit multiple but identical <K, V> pairs. I guess it is fine for unit tests?

albertshau: good point, will change it

gokulavasan:
One comment. Rest LGTM :+1:

PR #2025, opened 2015-04-15
add clean function, move variables to file, streamline repeated statemen...

duddun:
Changes in preparation for automating docs portion of cdap code builds.


cdap-docs/_common/common-build.sh
```diff
@@ -131,8 +131,11 @@ function usage() {
 
 function clean() {
   cd $SCRIPT_PATH
-  rm -rf $SCRIPT_PATH/$BUILD
-  mkdir $SCRIPT_PATH/$BUILD
+  if [ -d "${SCRIPT_PATH}/${BUILD}" ]
```
dereklwood: This existence check is redundant with the rm ``-f`` flag.  The ``-f`` flag causes it to return 0 if it doesn't exist.

cdap-docs/_common/common-build.sh
```diff
@@ -131,8 +131,11 @@ function usage() {
...
+  then
+    rm -rf $SCRIPT_PATH/$BUILD
```
dereklwood: convention is to put braces around all variables always: ``${SCRIPT_PATH}/${BUILD}``

cdap-docs/build.sh
```diff
@@ -149,11 +153,10 @@ function build_docs_outer_level() {
   echo "========================================================"
   echo ""
 
-  copy_html admin-manual
-  copy_html developers-manual
-  copy_html integrations
-  copy_html reference-manual
-  copy_html examples-manual
+  for i in "${manuals[@]}"
```
dereklwood: ``for i in ${MANUALS}`` would work too.  No need to have converted it to an array

cdap-docs/build.sh
```diff
@@ -92,7 +89,11 @@ function run_command() {
 
 function clean() {
   cd $SCRIPT_PATH
-  rm -rf $SCRIPT_PATH/$BUILD/*
+  if [ -d  "${SCRIPT_PATH}/${BUILD}" ]
+  then
+    echo "do nothing"
```
dereklwood: remove

cdap-docs/build.sh
```diff
@@ -171,18 +174,23 @@ function build_docs_outer_level() {
...
+  echo "========================================================================="
   echo "Building Web Docs."
-  ./build.sh docs-web $ARG_2 $ARG_3
+  echo "========================================================================="
+  run_command docs-web-part $ARG_2 $ARG_3
   echo "Replacing GitHub Docs."
   mv $SCRIPT_PATH/$BUILD_TEMP/*.zip $SCRIPT_PATH/$BUILD
   rm -rf $SCRIPT_PATH/$BUILD_TEMP
   bell
+  exit 1
```
dereklwood: why?  guessing this was for debug, pls remove

dereklwood:
Some comments.  Some with multiple instances, which i didn't repeatedly comment on, so please check those.

cdap-docs/build.sh
```diff
@@ -77,9 +71,12 @@ function usage() {
...
     docs )              build_docs; exit 1;;
-    docs-github )       build_docs_github; exit 1;;
-    docs-web )          build_docs_web; exit 1;;
+    docs-github-part )  build_docs_github $ARG_2 $ARG_3;;
+    docs-github )       build_docs_github $ARG_2 $ARG_3; exit 1;;
+    docs-web-part )     build_docs_web $ARG_2 $ARG_3;;
+    docs-web )          build_docs_web $ARG_2 $ARG_3; exit 1;;
     javadocs )          build_javadocs; exit 1;;
     licenses )          build_license_depends; exit 1;;
     sdk )               build_sdk; exit 1;;
```
wolf31o2: So... exit 1 means failed build.

cdap-docs/build.sh
```diff
@@ -68,6 +60,8 @@ function usage() {
   echo "    sdk            Build SDK"
   echo "    version        Print the version information"
   echo ""
+  echo "    clean	   Clean up (previous builds)"
```
wolf31o2: Align spacing with above/below.

cdap-docs/build.sh
wolf31o2: This function exits 1...

cdap-docs/_common/common-build.sh
```diff
@@ -175,7 +181,10 @@ function build_license_pdfs() {
   version
   cd $SCRIPT_PATH
   PROJECT_VERSION_TRIMMED=${PROJECT_VERSION%%-SNAPSHOT*}
-  rm -rf $SCRIPT_PATH/$LICENSES_PDF
+  if [ -d "${SCRIPT_PATH}/${LICENSES_PDF}" ]
+  then
+    rm -rf $SCRIPT_PATH/$LICENSES_PDF
+  fi
```
awholegunch: Why replace one line with four? Especially that have two with a path that needs to be the same in two places.

cdap-docs/_common/common-build.sh
```diff
@@ -131,8 +131,11 @@ function usage() {
 
 function clean() {
   cd $SCRIPT_PATH
-  rm -rf $SCRIPT_PATH/$BUILD
-  mkdir $SCRIPT_PATH/$BUILD
+  if [ -d "${SCRIPT_PATH}/${BUILD}" ]
+  then
+    rm -rf $SCRIPT_PATH/$BUILD
+  fi
+  mkdir -p $SCRIPT_PATH/$BUILD
```
awholegunch: I could be wrong, but this is looking more verbose for not much benefit. There's a number of places where one line is replaced with more? Especially cases that then end up with a path that needs to be the same in multiple places.

cdap-docs/build.sh
```diff
@@ -26,17 +26,9 @@
 # Builds each of these individually, and then packages them into a single zip file for distribution.
 # _common directory holds common files and scripts.
 
+source _common/vars
```
awholegunch: Are these vars actually common, or are they only used by the top-level? I think we should keep in common only variable that really are common...no?

cdap-docs/_common/vars
```diff
@@ -0,0 +1,11 @@
...
+COMMON="_common"
+COMMON_HTACCESS="$COMMON/htaccess"
+COMMON_IMAGES="$COMMON/_images"
+COMMON_SOURCE="$COMMON/_source"
+COMMON_CONF_PY="$COMMON/common_conf.py"
+COMMON_HIGHLEVEL_PY="$COMMON/highlevel_conf.py"
+COMMON_PLACEHOLDER="$COMMON/_source/placeholder_index._rst"
+HTACCESS="htaccess"
+
+MANUALS="admin-manual developers-manual examples-manual integrations reference-manual"
```
awholegunch: I don't think this should be here. This is not common to the other files; only the high-level build needs to know this. If it is here, it will be hard to find, as when a new manual is added. Such as I am currently doing.

cdap-docs/build.sh
awholegunch: I think the reason why I did it this is so that you can cancel a build in progress. Because when you do a control-c, if can get grabbed by Java or maven or Sphinx (python), and IKBW, but I think it was necessary to get the bigger script to not keep going. Plus, there are invocations inside invocations: the lower-level doc builds are run from inner shell scripts.

cdap-docs/build.sh
```diff
@@ -216,11 +224,10 @@ function _build_docs() {
 }
 
 function build() {
-  build_specific_doc admin-manual $1
-  build_specific_doc developers-manual $1
-  build_specific_doc integrations $1
-  build_specific_doc reference-manual $1
-  build_specific_doc examples-manual $1
+  for i in "${manuals[@]}"
```
awholegunch: see note above about ${MANUALS}.

cdap-docs/build.sh
```diff
@@ -281,6 +288,28 @@ function test() {
...
+  echo ""
+  if [ -d "${SCRIPT_PATH}/${BUILD}" ]
+  then
+    rm -rf $SCRIPT_PATH/$BUILD/*
+  fi
+  echo "Cleaned $SCRIPT_PATH/$BUILD directory"
+
+  # clean everything in manual directories  
+  echo ""
+  for i in "${manuals[@]}"
```
awholegunch: Ibid.

cdap-docs/build.sh
```diff
@@ -281,6 +288,28 @@ function test() {
   echo "Test completed."
 }
 
+function clean_builds() {
+  # clean everything in cdap-docs dir
```
awholegunch: # Removes all upper- and lower-level build directories

cdap-docs/_common/common-build.sh
awholegunch: Because the way JJ did it was wrong. This is probably why the script tried to eat Julien's dad drive one day when he invoked it from the wrong location, as it was probably deleting merrily away without testing first...

cdap-docs/build.sh
awholegunch: At least, this is what I recall. But that's the general idea, of what it trying to accomplish.

cdap-docs/build.sh
```diff
@@ -171,18 +174,23 @@ function build_docs_outer_level() {
 ################################################## current
 
 function build_all() {
+  echo "========================================================================="
   echo "Building GitHub Docs."
-  ./build.sh docs-github $ARG_2 $ARG_3
+  echo "========================================================================="
+  run_command docs-github-part $ARG_2 $ARG_3
```
awholegunch: I'm wondering about this: when I wrote this, I thought that I should duplicate the invocation to prevent any carry-over of environment. Is the environment going to be set the same, run this way? We don't have anything testing that Github google analytics are set correctly, for instance. (That could be added as a test, btw.)

duddun: Not sure about environment carryover.
Google analytics: If we test for that, then are a ton of things we want to test and error check :) but I do not disagree with you.

awholegunch: Here's what I would see needs to be tested, either manually or automatically: 
Outer level: 
```
build.sh docs
build.sh docs-web
build.sh docs-github
build.sh docs all
```
Check that the correct versions of Google Analytics are set in the output.
Check that builds can be killed: cntrl-c.

Inner Level (in each manual or sub-directory):
```
build.sh docs
```
If all these work, we should be good.

awholegunch:
I ran a build, and everything looked fine. LGTM.

PR #2026, opened 2015-04-15
Fix documentation build by downloading images from Github

awholegunch:
and using the downloaded version rather than using a local copy.

### DIFF
--- a/cdap-docs/examples-manual/build.sh
+++ b/cdap-docs/examples-manual/build.sh
@@ -33,13 +33,26 @@ function guide_rewrite_sed() {
   local project_version=$PROJECT_SHORT_VERSION
   
   local source1="https://raw.githubusercontent.com/cdap-guides"
-  local source2="release/cdap-$project_version-compatible/README.rst"
+  local source2="release/cdap-$project_version-compatible"
+  local rst_file="README.rst"
+  local redirect="\.\./\.\./build/_includes" # Target, 2 redirects, escaped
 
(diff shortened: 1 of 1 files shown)


cdap-docs/examples-manual/build.sh
```diff
@@ -33,13 +33,26 @@ function guide_rewrite_sed() {
...
+  local rst_file="README.rst"
+  local redirect="\.\./\.\./build/_includes" # Target, 2 redirects, escaped
 
-  local redirect="\.\./\.\./\.\./\.\./\.\." # Target, 5 redirects, escaped
-  
-  mkdir $includes_dir/$guide
-  curl --silent $source1/$guide/$source2 --output $includes_dir/$guide/README_SOURCE.rst  
-  sed -e "s|image:: docs/images|image:: $redirect/$guide/docs/images|g" -e "s|.. code:: |.. code-block:: |g" $includes_dir/$guide/README_SOURCE.rst > $includes_dir/$guide/README.rst
+  mkdir -p $includes_dir/$guide
+  curl --silent $source1/$guide/$source2/$rst_file --output $includes_dir/$guide/SOURCE_$rst_file
```
duddun: Is the output file name correct?

duddun:
LGTM other than the question above

awholegunch:
Yes, it's correct. I change the sed to use a file prefix ("SOURCE") instead of a suffix inserted between the filename and the filetype delimiter.

PR #2028, opened 2015-04-15
CDAP-2087 adding a table sink for the etl template. As part of

albertshau:
this, also adding a BatchWritableSink that will work for any
CDAP BatchWritable, but which requires users to supply the dataset
type and any properties that type requires. TableSink and
KVTableSink then extend the BatchWritableSink.


albertshau:
http://builds.cask.co/browse/CDAP-DT15

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/transforms/StructuredRecordToPutTransformTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Map;
+
+/**
+ */
+@SuppressWarnings("unchecked")
+public class StructuredRecordToPutTransformTest {
+
```
gokulavasan: remove extra line

gokulavasan:
LGTM :+1:

PR #2031, opened 2015-04-15
Fix/ui programs view consistency

ajainarayanan:
#### Long Pending simple changes

- Unifies the look (layout) of a program detail view's heading & start/stop button + active runs (if applicable)
- Unifies the behavior of start & stop buttons in all program views 
- Minor style modifications to breadcrumbs

screen shot 2015-04-15 at 2 03 23 am
screen shot 2015-04-15 at 1 58 09 am
screen shot 2015-04-15 at 1 57 56 am
screen shot 2015-04-15 at 1 57 31 am

TODO:
  Make this as a directive (if it makes sense) and simplify the start/stop function handlers in the scope.
 EDIT: 
   Starting/Stopping a concurrent run of a mapreduce/workflow is not possible yet. Will be fixed within this week.

### DIFF
--- a/cdap-ui/app/features/flows/controllers/detail-ctrl.js
+++ b/cdap-ui/app/features/flows/controllers/detail-ctrl.js
@@ -1,5 +1,6 @@
 angular.module(PKG.name + '.feature.flows')
   .controller('FlowsDetailController', function($scope, MyDataSource, $state, FlowDiagramData, myHelpers, $timeout) {
+    $scope.status = null;
     var dataSrc = new MyDataSource($scope),
         basePath = '/apps/' + $state.params.appId + '/flows/' + $state.params.programId;
 
@@ -13,11 +14,13 @@ angular.module(PKG.name + '.feature.flows')
 
 
     $scope.do = function(action) {
+      $scope.status = action;
       dataSrc.request({
         _cdapNsPath: basePath + '/' + action,
         method: 'POST'
       }).then(function() {
         $timeout(function() {
+          $scope.status = null;
           $state.go($state.current, $state.params, { reload: true });
         });
       });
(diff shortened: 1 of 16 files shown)


anwar6953:
Order of tabs seems out of order.
Service has history in the middle, while other programs have it at the end (right-most).

ajainarayanan:
@anwar6953 The order of tabs (or the actual tabs) are going to change. Those will be the next set of changes

cdap-ui/app/features/flows/controllers/detail-ctrl.js
```diff
@@ -58,11 +59,13 @@ angular.module(PKG.name + '.feature.flows')
     });
 
     $scope.do = function(action) {
+      $scope.status = action;
```
anwar6953: Say I have a flow that is 'RUNNING'.
Now, If I click stop, then the status is 'STOPPED' automatically?
What happens the very next time I poll the back end for status? (it might be 'STOPPING').

cdap-ui/app/features/services/controllers/tabs/status-ctrl.js
```diff
@@ -26,27 +26,10 @@ angular.module(PKG.name + '.feature.services')
         $scope.instances = res;
       });
 
-    dataSrc.poll({
+    dataSrc.request({
```
anwar6953: Why not poll for status?

anwar6953: One example where polling is definitely necessary is MapReduce (which stops on its own, once it completes).
I think that the UI should poll for all programs though if the user is on the page to view the status (flows/services can stop on their own given a failure condition).

cdap-ui/app/features/flows/controllers/detail-ctrl.js
ajainarayanan: This is not something that we get from backend. I believe it was said that flows cannot have multiple runs. So I am listening for active runs from backend.

if active runs is 0 && status is null
  show start button
else if active runs is gt 0  && status is null
  show start button
if status is start
  show start loading 
if status is stop
  show stop loading

astewes:
LGTM

elfenheart:
:+1: LGTM besides @anwar6953 suggestion to Poll for status

PR #2034, opened 2015-04-15
CDAP-1973: Fix InstanceURIParser for URIs with trailing slash

jwang47:
https://issues.cask.co/browse/CDAP-1973
http://builds.cask.co/browse/CDAP-DUT1452


sreevatsanraman:
LGTM :+1:

PR #2035, opened 2015-04-15
Multiple metrics per chart.

anwar6953:
This lets you choose multiple metrics when creating a widget. It uses the new metrics batch endpoint to query for all the metrics of the chart in one http request.
http://g.recordit.co/E6lzHDiwK6.gif

Not implemented (yet):
There is no legend/key (identification for which color is which metric) currently.
You can not modify the list of metrics for the chart after creating it.

### DIFF
--- a/cdap-ui/app/directives/epoch/charts.js
+++ b/cdap-ui/app/directives/epoch/charts.js
@@ -72,7 +72,7 @@ ngEpoch.controller('epochController', function ($scope, $compile, caskWindowMana
         }
         else if (newVal && newVal.length) {
           for (var i = 0; i < newVal.length; i++) {
-            $scope.me.push([newVal[i]]);
+            $scope.me.push(newVal[i]);
           }
 
         }
--- a/cdap-ui/app/directives/epoch/charts.js
+++ b/cdap-ui/app/directives/epoch/charts.js
@@ -71,10 +71,7 @@ ngEpoch.controller('epochController', function ($scope, $compile, caskWindowMana
           $scope.me.update(newVal);
         }
         else if (newVal && newVal.length) {
-          for (var i = 0; i < newVal.length; i++) {
-            $scope.me.push(newVal[i]);
-          }
-
+          $scope.me.push(newVal);
         }
       });
     }
--- a/cdap-ui/app/directives/metric-picker/metric-picker.html
+++ b/cdap-ui/app/directives/metric-picker/metric-picker.html
@@ -24,6 +24,7 @@
   </div>
 </div>
 
+<div class="metrics-container">
 <div ng-repeat="name in metric.names" class="h6">
   <div class="input-group">
     <span class="input-group-addon">
@@ -39,11 +40,12 @@
       data-container="body"
       bs-typeahead />
     <div class="input-group-addon" ng-click="deleteMetric($index)">
-      <span ng-click="deleteMetric($index)">
+      <span>
         <i class="fa fa-fw fa-trash"></i>
       </span>
     </div>
   </div>
+</div>
 </div>
 
 <button class="btn btn-primary" ng-click="addMetricName()" type="button" >
--- a/cdap-ui/app/directives/metric-picker/metric-picker.html
+++ b/cdap-ui/app/directives/metric-picker/metric-picker.html
@@ -24,17 +24,28 @@
   </div>
 </div>
 
-<div class="input-group">
-  <span class="input-group-addon">
-    <i class="fa fa-line-chart"></i>
-  </span>
-  <input type="text" class="form-control"
-    placeholder="Metric Name"
-    ng-model="metric.name"
-    bs-options="m for m in available.names"
-    data-watch-options="1"
-    data-min-length="0"
-    data-limit="20"
-    data-container="body"
-    bs-typeahead />
+<div ng-repeat="name in metric.names" class="h6">
+  <div class="input-group">
+    <span class="input-group-addon">
+      <i class="fa fa-line-chart"></i>
+    </span>
+    <input type="text" class="form-control"
+      placeholder="Metric Name"
+      ng-model="name.name"
+      bs-options="m for m in available.names"
+      data-watch-options="1"
+      data-min-length="0"
+      data-limit="20"
+      data-container="body"
+      bs-typeahead />
+    <div class="input-group-addon" ng-click="deleteMetric($index)">
+      <span ng-click="deleteMetric($index)">
+        <i class="fa fa-fw fa-trash"></i>
+      </span>
+    </div>
+  </div>
 </div>
+
+<button class="btn btn-primary" ng-click="addMetricName()" type="button" >
+  <span class="fa fa-fw fa-plus"></span>
+</button>
--- a/cdap-ui/app/directives/metric-picker/metric-picker.js
+++ b/cdap-ui/app/directives/metric-picker/metric-picker.js
@@ -16,7 +16,35 @@ angular.module(PKG.name + '.commons')
       $scope.metric = {
         context: '',
         type: ns,
-        name: ''
+        names: [{name: ''}],
+        resetNames: function() {
+          this.names = [{name: ''}];
+        },
+        getNames: function() {
+          return this.names.map(function(value) {
+            return value.name;
+          });
+        },
+        getName: function() {
+          name = '';
+          this.getNames().forEach(function(el) {
+            name += el + ', ';
+          });
+          return name;
+        }
+      };
+
+      $scope.addMetricName = function() {
+        $scope.metric.names.push({name: ''});
+      };
+
+      $scope.deleteMetric = function(idx) {
+        if ($scope.metric.names.length == 1) {
+          // If its the only metric, simply clear it instead of removing it
+          $scope.metric.resetNames();
+        } else {
+          $scope.metric.names.splice(idx, 1);
+        }
       };
 
     }
@@ -38,7 +66,15 @@ angular.module(PKG.name + '.commons')
           elem.find('input').attr('required', true);
           ngModel.$validators.metricAndContext = function (m, v) {
             var t = m || v;
-            return t && t.name && t.context;
+            if (!t || !t.names || !t.names.length || !t.context) {
+              return false;
+            }
+            for (var i = 0; i < t.names.length; i++) {
+              if (!t.names[i].length) {
+                return false;
+              }
+            }
+            return true;
           };
         }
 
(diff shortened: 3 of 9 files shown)


cdap-ui/app/directives/epoch/charts.js
```diff
@@ -72,7 +72,7 @@ ngEpoch.controller('epochController', function ($scope, $compile, caskWindowMana
         }
         else if (newVal && newVal.length) {
           for (var i = 0; i < newVal.length; i++) {
-            $scope.me.push([newVal[i]]);
+            $scope.me.push(newVal[i]);
```
ajainarayanan: Any specific reason why this is changing? This will cause javascript error to be thrown right?

anwar6953: I wrap it in an array here:
https://github.com/caskdata/cdap/pull/2035/files#diff-461271d65a10d30a3b9a8f79b05e0349R157

For multiple metrics, it's no longer a single element array, but instead an array of n elements, where n is the number of metrics being polled/displayed.

cdap-ui/app/directives/metric-picker/metric-picker.html
```diff
@@ -24,17 +24,28 @@
...
+      placeholder="Metric Name"
+      ng-model="name.name"
+      bs-options="m for m in available.names"
+      data-watch-options="1"
+      data-min-length="0"
+      data-limit="20"
+      data-container="body"
+      bs-typeahead />
+    <div class="input-group-addon" ng-click="deleteMetric($index)">
+      <span ng-click="deleteMetric($index)">
```
ajainarayanan: We probably don't need this twice. The even bubbles up and this should be caught at the parent.

anwar6953: I should remove the one on line 42?

cdap-ui/app/directives/metric-picker/metric-picker.js
```diff
@@ -16,7 +16,35 @@ angular.module(PKG.name + '.commons')
...
+          this.names = [{name: ''}];
+        },
+        getNames: function() {
+          return this.names.map(function(value) {
+            return value.name;
+          });
+        },
+        getName: function() {
+          name = '';
+          this.getNames().forEach(function(el) {
```
ajainarayanan: this.getNames().join()?

cdap-ui/app/directives/metric-picker/metric-picker.js
```diff
@@ -38,7 +66,15 @@ angular.module(PKG.name + '.commons')
           elem.find('input').attr('required', true);
           ngModel.$validators.metricAndContext = function (m, v) {
             var t = m || v;
-            return t && t.name && t.context;
+            if (!t || !t.names || !t.names.length || !t.context) {
```
ajainarayanan: This can be simplified to

```
if (!myHelper.objectQuery(t, 'names', 'length') || ! t.context) {
...
}
```

cdap-ui/app/directives/metric-picker/metric-picker.js
```diff
@@ -109,45 +145,54 @@ angular.module(PKG.name + '.commons')
...
         elem.find('input').on('blur', onBlurHandler);
 
-        scope.$watchCollection('metric', function (newVal, oldVal) {
+        var metricChanged = function (newVal, oldVal) {
           ngModel.$validate();
 
           if(newVal.type !== oldVal.type) {
             scope.metric.context = '';
-            scope.metric.name = '';
+            scope.metric.resetNames();
```
ajainarayanan: What happens in this scenario,

- Choose context 1 
- Add 5 metrics
- Erase context 1 and choose context 2

Will the 5 metrics be there but their values cleared out or are we removing all the metrics except for the one (starting back from default)?

anwar6953: Right now, it'll remove the 4 rows, and clear the first.
Should we keep the metrics and not reset them?

cdap-ui/app/directives/metric-picker/metric-picker.js
```diff
@@ -109,45 +145,54 @@ angular.module(PKG.name + '.commons')
...
-          if(newVal.context && newVal.name) {
-            if (newVal.name === 'Add All') {
+          if(newVal.context && newVal.names) {
+            var isAddAll = false;
+            for (var i = 0; i < newVal.names.length; i++) {
+              if (newVal.names[i].name === 'Add All') {
+                isAddAll = true;
+              }
+            }
+            if (isAddAll) {
```
ajainarayanan: I feel when we choose 'Add All' even as one option we should probably keep only that instead of adding multiple metrics. Not sure how we should deal with 'Add All' + adding more metrics in the metrics picker works. They should be mutually exclusive right?

- We either allow users to choose multiple metrics to be shown in the same graph 
- Or we add all metrics in the current context as independent widgets.

@nitinmotgi Any thoughts?

anwar6953: Probably 'add all' should be a checkbox that removes the metric input boxes. (I agree with the exclusivity of the two).

cdap-ui/app/directives/metric-picker/metric-picker.js
```diff
@@ -109,45 +145,54 @@ angular.module(PKG.name + '.commons')
...
+          } else {
             if(ngModel.$dirty) {
               ngModel.$setViewValue(null);
             }
 
           }
 
-        });
+        }
+        scope.$watch('metric', metricChanged, true);
```
ajainarayanan: This will be expensive. Probably not in this PR but we need to figure out a way to make this a little better.

anwar6953: We need to watch each of the metric name input boxes, right?
Is there much more that this is watching?

cdap-ui/app/features/dashboard/controllers/addwdgt-ctrl.js
```diff
@@ -50,17 +50,18 @@ function ($scope, $modalInstance, caskFocusManager, Widget) {
...
         widgets.push(
           new Widget({
             type: $scope.model.type,
-            title: value,
-            color: classes[(generateHash(value) * 13) % classes.length],
+            title: $scope.model.metric.title,
+            color: classes[(generateHash($scope.model.metric.context) * 13) % classes.length],
             metric: {
               context: $scope.model.metric.context,
+              names: [value],
```
ajainarayanan: Do we need to pass both ```name``` & ```names``` here?

cdap-ui/app/directives/metric-picker/metric-picker.js
```diff
@@ -16,7 +16,35 @@ angular.module(PKG.name + '.commons')
...
+        resetNames: function() {
+          this.names = [{name: ''}];
+        },
+        getNames: function() {
+          return this.names.map(function(value) {
+            return value.name;
+          });
+        },
+        getName: function() {
+          name = '';
```
ajainarayanan: ```name``` variable not declared?

cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -14,17 +14,39 @@ angular.module(PKG.name+'.feature.dashboard')
       this.dataSrc = null;
       this.isLive = false;
     }
+
+    // 'ns.default.app.foo' -> {'ns': 'default', 'app': 'foo'}
+    function contextToTags(context) {
+      parts = context.split('.');
```
ajainarayanan: ```parts``` not declared?

cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -14,17 +14,39 @@ angular.module(PKG.name+'.feature.dashboard')
...
+      if (parts.length % 2 != 0) {
+        throw "Metrics context has uneven number of parts: " + this.metric.context
+      }
+      tags = {};
```
ajainarayanan: ```tags``` not declared?

cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -14,17 +14,39 @@ angular.module(PKG.name+'.feature.dashboard')
...
+      for (var i = 0; i < parts.length; i+=2) {
+        tags[parts[i]] = parts[i + 1]
+      }
+      return tags;
+    }
+
+    function constructQuery(queryId, tags, metrics) {
+      timeRange = {'start': 'now-60s', 'end': 'now'};
```
ajainarayanan: I believe none of the variables is declared? Am I missing something or the initialization happening somewhere else?

cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -107,26 +135,33 @@ angular.module(PKG.name+'.feature.dashboard')
...
 
-        $scope.chartHistory = [
-          {
-            label: $scope.wdgt.metric.name,
-            values: v
+        if ($scope.chartHistory) {
+          arr = [];
+          for (var i = 0; i < vs.length; i++) {
+            var el = vs[i];
+            var lastIndex = el.length-1;
```
ajainarayanan: This could have been ```arr.push(el.slice(-1))```

cdap-ui/app/features/dashboard/templates/partials/addwdgt.html
```diff
@@ -23,7 +23,7 @@ <h4 class="modal-title pull-left">Add a Widget</h4>
 
           <div class="form-group">
             <label class="col-xs-3 control-label">Metric</label>
-            <div class="col-xs-9">
+            <div class="col-xs-8 modal-body">
```
ajainarayanan: This is not 'semantically' correct. ```modal-body``` class is something given by bootstrap and we should use it to enclose everything that we put under modal body.

cdap-ui/app/directives/epoch/charts.js
ajainarayanan: What happens if I don't get want multiple metrics? In the case of cdap dashboard? This throws me an error in the console when I am in the cdap (standard) dashboard

cdap-ui/app/directives/metric-picker/metric-picker.html
ajainarayanan: Yes sir! Just remove ```ng-click``` from span

cdap-ui/app/directives/metric-picker/metric-picker.js
ajainarayanan: So when we do ```$scope.$watch( 'something', handler , true)``` then it makes a deep-copy check.

In our case we have a object like this
```
{
  context: '',
  metric: [ ... ],
  type: ''
}
```
when we do ```$watch(something, handler)``` it will trigger a change event when we modify an array item in metric (say in our case name). So this needs the third argument to be true but on a longer run we need to figure out how to avoid that. Not necessarily in this PR.

cdap-ui/app/directives/metric-picker/metric-picker.js
ajainarayanan: :+1:

cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -14,17 +14,39 @@ angular.module(PKG.name+'.feature.dashboard')
...
+      timeRange = {'start': 'now-60s', 'end': 'now'};
+      queryObj = {tags: tags, metrics: metrics, groupBy: [], timeRange: timeRange};
+      retObj = {};
+      retObj[queryId] = queryObj;
+      return retObj;
+    }
+
+    // The queryId value does not matter, as long as we are using the same value in the request
+    // as in parsing the response.
+    var queryId = "qid";
```
ajainarayanan: I am not sure I understand this usage. we want a unique id for each query right? But we are always using ```qid``` ?

cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -14,17 +14,39 @@ angular.module(PKG.name+'.feature.dashboard')
...
         this.dataSrc = new MyDataSource(scope);
       }
       this.dataSrc.request({
-        _cdapPath: '/metrics/query' +
-          '?context=' + encodeURIComponent(this.metric.context) +
-          '&metric=' + encodeURIComponent(this.metric.name) +
-          '&start=now-60s&end=now',
-
-        method: 'POST'
+        _cdapPath: '/metrics/query',
```
ajainarayanan: Not sure if its here but for metrics call return I am getting this as response ```"Missing metrics parameter in the query"```

Am I missing something? And one more thing the live feature is no longer live. Are we disabling it from UI?

cdap-ui/app/directives/epoch/charts.js
anwar6953: If you don't want multiple metrics, It'll be a 1-element array (just as it is currently on develop).

cdap-ui/app/features/dashboard/controllers/widget.js
anwar6953: So either metrics parameter has to be in query params, or send it as the body of the request. I am changing it to use the body of the request.

cdap-ui/app/features/dashboard/controllers/widget.js
anwar6953: The queryId does not have to be unique across requests.
It only has to be across queries for different contexts in the same http request.

cdap-ui/app/directives/metric-picker/metric-picker.js
anwar6953: I think for this case, we can't avoid it, because we definitely do need to watch all elements of the metric array, right?

cdap-ui/app/directives/metric-picker/metric-picker.js
anwar6953: replaced this by:
`          return this.getNames().join(', ');`

cdap-ui/app/features/dashboard/controllers/addwdgt-ctrl.js
anwar6953: `names` is the list of metric names for the chart to poll.
`name` is the name of the widget (displays at the bottom).

ajainarayanan:
:+1: LGTM

PR #2040, opened 2015-04-15
CDAP-2059: Split auth address into "announce" and "bind" addresses

jwang47:
https://issues.cask.co/browse/CDAP-2059
http://builds.cask.co/browse/CDAP-DUT1457

Now by default you get the following when you enable security and try to make an unauthorized request:

```
$ curl http://localhost:10000/v3/namespaces
{"auth_uri":["http://Alvins-MacBook-Pro-2.local:10009/token"]}
```

You can also change this by configuring the `security.auth.server.announce.address` configuration property:

```
    <property>
      <name>security.auth.server.announce.address</name>
      <value>asdf</value>
    </property>
```

```
$ curl http://localhost:10000/v3/namespaces
{"auth_uri":["http://asdf:10009/token"]}
```


cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -1077,8 +1077,13 @@
     </property>
 
     <property>
+      <name>security.auth.server.address</name>
+      <value>asdf</value>
```
sreevatsanraman: Can we make a note that this is deprecated? We should also give devops headsup on this change.

sreevatsanraman:
:+1:

PR #2043, opened 2015-04-15
CDAP-2112 Allow deleting Template cdap app when all adapters of that template type are deleted

gokulavasan:
Bamboo : http://builds.cask.co/browse/CDAP-DUT1375-10

JIRA : https://issues.cask.co/browse/CDAP-2112 [Partial solution since automatic deletion of app is not addressed by this PR]


albertshau:
:+1:

PR #2045, opened 2015-04-16
CDAP-2069 adding a filter transform that lets the user provide

albertshau:
a javascript snippet that determines whether or not a record
should be filtered or not.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/transforms/ScriptFilterTransformTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/transforms/ScriptFilterTransformTest.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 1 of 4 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DT16

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/transforms/ScriptFilterTransformTest.java
```diff
@@ -0,0 +1,121 @@
...
+import co.cask.cdap.templates.etl.common.MockEmitter;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ */
+@SuppressWarnings("unchecked")
+public class ScriptFilterTransformTest {
```
gokulavasan: May be we can add couple of tests to test the failure scenarios - invalid script and invalid filter condition?

gokulavasan:
Comment about testing error conditions. Rest LGTM :+1:

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/transforms/ScriptFilterTransformTest.java
albertshau: sure will do

PR #2046, opened 2015-04-16
Fix/ui workflow status removal

ajainarayanan:
- Re-implements the mocks for workflows view
- Makes workflows runs available even before loading details view of workflows
    - This is to make sure we don't make workflows/workflowId/runs call at multiple places.
- Uses one template to render detailed view of a run (used both in runs & runs/runid)
   In the case of runs state, it will always show the detailed view of the latest run id
   In the case of runs/runid it will show the detailed view of the respective run id


elfenheart:
:+1: LGTM

PR #2050, opened 2015-04-16
cdap-cli script file, and cmd-line multi-cmd support

dave-finnegan:
This pull request provides support for cdap-cli.sh script file handling and for multi-command support on the cdap-cli.sh command-line.

The script file can contain cdap-cli.sh commands, blank lines, or commented lines beginning with a '#' character.  Blank lines and comment lines are ignored.

The 'break' command is interpreted by the script handling code as a command to break out of the script being interpreted.  This aids in developing scripts.

The 'exit' command will exit the script and the cdap-cli.sh process, as expected.  Without the 'exit' command being placed into the script the caller is left at the cdap shell prompt in interactive mode.

In addition, the handling of command-line entered cdap-cli.sh commands was augmented to support multiple commands separated by a semi-colon (';') character.  This must be quoted or escaped on the command line.  This facilitates commands such as: 

    cdap-cli.sh "use namespace Demo; load stream logEventStream data/accesslog.txt"

The multi-command support on the cdap-cli.sh command-line is related to Cask Issue: https://issues.cask.co/browse/CDAP-2113.


nitinmotgi:
@dave-finnegan you might want to add test cases :-)

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -91,6 +94,14 @@
   public static final Option DEBUG_OPTION = new Option(
     "d", "debug", false, "Print exception stack traces.");
 
+  @VisibleForTesting
+  public static final Option SCRIPT_FILE_OPTION = new Option(
+    "f", "scriptfile", true, "Name of script file to read and execute.  " +
+    "Blank lines and comments starting with '#' are ignored.  " +
```
jwang47: Can we simply have anything after a `#` be ignored in all cases? I think it would be confusing if you can do `# sdfk` as a comment but not `list apps # lists apps`.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -91,6 +94,14 @@
   public static final Option DEBUG_OPTION = new Option(
     "d", "debug", false, "Print exception stack traces.");
 
+  @VisibleForTesting
+  public static final Option SCRIPT_FILE_OPTION = new Option(
+    "f", "scriptfile", true, "Name of script file to read and execute.  " +
+    "Blank lines and comments starting with '#' are ignored.  " + 
+    "The 'break' keyword terminates the script.  " + 
+    "The 'exit' command terminates the cli, otherwise the user is left " +
```
jwang47: Maybe have CLI terminate-at-end be the default and then if you do a `break` at the end, it'll leave the user at the cdap-cli command prompt?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -248,10 +260,54 @@ public static void main(String[] args) {
...
+
+          ConnectionConfig connectionConfig = clientConfig.getConnectionConfig();
+          URI baseURI = connectionConfig.getURI();
+          URI uri = baseURI.resolve("/" + connectionConfig.getNamespace());
+          String cliPrompt = "\ncdap (" + uri + ")> ";
+
+          BufferedReader br = null;
+          try {
+            String line;
+            br = new BufferedReader(new FileReader(launchOptions.getScriptFile()));
```
jwang47: Use `br = Files.newReader(launchOptions.getScriptFile(), Charset.UTF_8)` instead.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -248,10 +260,54 @@ public static void main(String[] args) {
...
+            while ((line = br.readLine()) != null) {
+              String ln = line.trim();
+              if ("".equals(ln) || ln.startsWith("#")) {
+                continue;
+              }
+              if (ln.startsWith("break")) {
```
jwang47: `ln.startsWith("break ")`

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -248,10 +260,54 @@ public static void main(String[] args) {
...
+                output.println(cliPrompt + ln);
+                break; // Break from script
+              }
+              output.println(cliPrompt + ln);
+              cli.execute(ln, output);
+            }
+            output.println("\n");
+          } catch (IOException e) {
+            output.println(e.getMessage());
+            usage();
```
jwang47: Why print usage on `IOException`? This is thrown when `readLine` fails right? I think we should just print an error message like "Couldn't read from script file <path-to-script-file>"

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -248,10 +260,54 @@ public static void main(String[] args) {
...
+          }
+        }
         if (commandArgs.length == 0) {
+          // No commands following switches; just start shell
           cli.startInteractiveMode(output);
         } else {
-          cli.execute(Joiner.on(" ").join(commandArgs), output);
+          // Commands following switches; process them
+          String commandLine = Joiner.on(" ").join(commandArgs);
+          String[] commandLns = commandLine.split(";");
```
jwang47: So now you can't enter in a `;` for an individual command here. Can you add some code that allows escaping?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: I couldn't get this line to compile; sorry, java noob.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: I'm not sure that I follow this.  Is a ';' character used within the cdap-cli command space?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: Done.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: Done.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: Done.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: Done.

dave-finnegan:
I'm happy to add tests, but not quite sure how to do that.  Any pointers, or examples would be appreciated.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
jwang47: You'll need to `import com.google.common.io.Files;` first.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -248,10 +265,63 @@ public static void main(String[] args) {
...
+          String cliPrompt = "\ncdap (" + uri + ")> ";
+
+          BufferedReader br = null;
+          try {
+            String line;
+            br = new BufferedReader(new FileReader(launchOptions.getScriptFile()));
+            // br = Files.newReader(launchOptions.getScriptFile(), Charset.UTF_8);
+            while ((line = br.readLine()) != null) {
+              String ln = line.trim();
+              String[] lnArr = ln.split("#");
```
jwang47: Need a way to allow user to escape `#`

jwang47: Also probably don't want to treat it as a comment if it's inside quotes. E.g. `send stream x "#helloworld"`

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -248,10 +265,63 @@ public static void main(String[] args) {
...
+              }
+            } catch (IOException e) {
+              output.println("Couldn't close script '" + 
+                             launchOptions.getScriptFile() + "'\n" + e.getMessage());
+            }
+          }
+        }
+        if (commandArgs.length > 0 && !brokeFromScript) {
+          String commandLine = Joiner.on(" ").join(commandArgs);
+          String[] commandLns = commandLine.split(";");
```
jwang47: Need to allow user to escape `;`

jwang47: Also probably don't want to treat it as a command separator if it's inside quotes. E.g. send stream x "hello;world" or send stream x 'hello;world'

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -248,10 +265,63 @@ public static void main(String[] args) {
...
+          for (int i = 0; i < commandLns.length; i++) {
+              cli.execute(commandLns[i].trim(), output);
```
jwang47: I think the commands provided inline should be executed first, though seems better to allow the user to choose. Maybe add an option later.

cdap-cli/src/main/java/co/cask/cdap/cli/util/table/AltStyleTableRenderer.java
```diff
@@ -15,9 +15,11 @@
  */
```
jwang47: Don't think this should be part of this PR. Not sure why it's here though, since these changes are already in develop.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
jwang47: This comment feature should also be supported in other cases: Entering a command in interactive mode, and entering command from command line (e.g. `cdap-cli.sh deploy app purchase.jar`)

cdap-cli/src/main/java/co/cask/cdap/cli/util/table/AltStyleTableRenderer.java
dave-finnegan: This code should not be a part of this PR; agreed.  I had pulled in that fix to my builds, but did not intend to include it in my PR.  I may need help remedying this situation as I'm somewhat new to git.

jwang47: Seems like this shouldn't even be visible in this PR since the same commit is already merged to develop: https://github.com/caskdata/cdap/commit/12975d5daf310549c507d67eeba3769b638dad2d

Maybe you need to synchronize your fork with caskdata/cdap?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: Done.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: Done.  Cmd line args now default to run first.  Added switch to allow them to run second.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: Added comment and semi-colon (multi-command) support for command line and script file command string interpretation.  Would like to add it to command prompt string interpretation, but could use some help in understanding the cli.startInteractiveMode() call and where the command prompt strings get interpreted.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
dave-finnegan: Done.

anew:
Hmmm. I am not sure I like this approach. Typically, command line interpreters do not process line by line, but statement by statement, separated by ";". I know bash goes line by line but allows continuation into the next line. But given that you want to support ";" to separate multiple commands on the command line arguments, it seems more consistent to use ";" in the file, too. 

Overall, I am not sure how brittle the parsing will become, once you deal with escaping and quoting. It seems to me that if we want to support more complex syntax, we need to port the CLI to a parser generator (or at least a lexer generator) like flex or antlr.

sreevatsanraman:
please note, let's not fork the repo for cask developers. We could create feature branches from caskdata/cdap.git

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
jwang47: `CLI.startInteractiveMode` calls `CLI.execute` https://github.com/caskdata/common/blob/develop/common-cli/src/main/java/co/cask/common/cli/CLI.java#L129. The ideal place would be to make the changes to support comments and multiple commands in one line in `CLI.execute` in common-cli, but probably there isn't enough time for that right now.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
jwang47: How does the user escape `;`?

sreevatsanraman:
We will close this PR and open one from caskdata/cdap repo with the unit tests.

PR #2051, opened 2015-04-16
Fix/ui flow status removal

ajainarayanan:
Does the exact same thing as this PR (https://github.com/caskdata/cdap/pull/2046/files#diff-30dd5b019ff05bdd57ddb1f91b7ba466) for flows


elfenheart:
:+1: LGTM

PR #2052, opened 2015-04-16
CDAP-2137 Get JMS messages from JMS Consumer from async to sync to reduce heap usage

henry-cask:
The first drop for JMS realtime source for template uses internal Queue to support asynchronous push messages from
JMS consumer so when the CDAP template transformer call the JmsSource#poll it already should have message available.

However, this could consume some extra heap allocation given that poll is the mechanism we used to get the
realtime data from CDAP source.
The Queue could get large such that we need to cap it or start dropping older messages if CDAP template transformer
does not poll fast enough.

Need to change the async mode to get messages from JMS provider to sync mode by calling JMS MessageConsumer#receive explicitly.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
@@ -24,8 +24,6 @@ import co.cask.cdap.templates.etl.realtime.jms.JmsProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.LinkedBlockingQueue;
 import javax.annotation.Nullable;
 import javax.jms.BytesMessage;
 import javax.jms.Connection;
@@ -34,7 +32,6 @@ import javax.jms.Destination;
 import javax.jms.JMSException;
 import javax.jms.Message;
 import javax.jms.MessageConsumer;
-import javax.jms.MessageListener;
 import javax.jms.Session;
 import javax.jms.TextMessage;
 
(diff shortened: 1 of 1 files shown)


henry-cask:
Build at:

http://builds.cask.co/browse/CDAP-DUT1467-1

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
```diff
@@ -44,17 +41,18 @@
  * JMS Consumer and send the message as String to the CDAP ETL Template flow via {@link ValueEmitter}
  * </p>
  */
-public class JmsSource extends RealtimeSource<String> implements MessageListener {
+public class JmsSource extends RealtimeSource<String> {
   private static final Logger LOG = LoggerFactory.getLogger(JmsSource.class);
 
-  // TODO Need option to add Max size of the internal queue
-  private final BlockingQueue<Message> messageQueue = new LinkedBlockingQueue<Message>();
+  private final static long JMS_CONSUMER_TIMEOUT_MS = 30000;
```
bdmogal: nit: by convention, we use static before final, here and next line.

bdmogal:
1 minor nit, then LGTM. Fix and merge.

Also with the understanding that we need to eventually support an ``onMessage`` kind of mechanism in future. Could you create a jira for that?

henry-cask:
@bdmogal Sounds good, will create separate JIRA to address it. Thanks!

gokulavasan:
@henry-cask The build you have pointed to is under the regular Bamboo configuration which does not run templates profile. You can create a branch manually under Templates Bamboo (http://builds.cask.co/browse/CDAP-DT) or you can create a branch with name feature-adapter/* to be enable bamboo to pick it and build automatically.

henry-cask:
Ah ok, will try it with feature-adapeter branch name this time

henry-cask:
http://builds.cask.co/browse/CDAP-DT10-7

henry-cask:
Closing this to open
https://github.com/caskdata/cdap/pull/2072

PR #2053, opened 2015-04-16
Use RunRecords for job-level state, start/end time.

anwar6953:
1) Use RunRecords for job-level state, start/end time.
2) Add a field to the return object which returns if the return object is complete or not.

Change no. 1 helps to provide that information in standalone as well as when the JobHistory server is missing.

Build: http://builds.cask.co/browse/CDAP-DUT1324-20

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
@@ -23,7 +23,6 @@ import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.MRJobInfo;
 import co.cask.cdap.proto.MRTaskInfo;
 import co.cask.cdap.proto.ProgramType;
-import co.cask.cdap.proto.RunRecord;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
@@ -62,12 +61,11 @@ public class MRJobClient {
 
   /**
    * @param runId for which information will be returned.
-   * @param runRecord for the run (we already keep track of information such as job-level state, start/end time).
    * @return a {@link MRJobInfo} containing information about a particular MapReduce program run.
    * @throws IOException if there is failure to communicate through the JobClient.
    * @throws NotFoundException if a Job with the given runId is not found.
    */
-  public MRJobInfo getMRJobInfo(Id.Run runId, RunRecord runRecord) throws IOException, NotFoundException {
+  public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundException {
     Preconditions.checkArgument(ProgramType.MAPREDUCE.equals(runId.getProgram().getType()));
 
     JobClient jobClient;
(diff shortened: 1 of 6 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
```diff
@@ -61,11 +62,12 @@ public MRJobClient(CConfiguration cConf, Configuration hConf) {
 
   /**
    * @param runId for which information will be returned.
+   * @param runRecord for the run (we already keep track of information such as job-level state, start/end time).
    * @return a {@link MRJobInfo} containing information about a particular MapReduce program run.
    * @throws IOException if there is failure to communicate through the JobClient.
    * @throws NotFoundException if a Job with the given runId is not found.
    */
-  public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundException {
+  public MRJobInfo getMRJobInfo(Id.Run runId, RunRecord runRecord) throws IOException, NotFoundException {
```
abaranau: it seems wrong/confusing to pass RunRecord here... Maybe pass constructed MRJobInfo with those fields, and set others with setters?

abaranau: passing runrecord here mixes different things, adds unnecesary and confusing dependencies

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MapReduceMetricsInfo.java
```diff
@@ -131,7 +132,7 @@ private void putMetrics(Map<String, Long> taskMetrics, Table<String, String, Lon
     }
   }
 
-  private MRJobInfo getJobCounters(Map<String, String> mapTags, Map<String, String> reduceTags,
+  private MRJobInfo getJobCounters(RunRecord runRecord, Map<String, String> mapTags, Map<String, String> reduceTags,
```
abaranau: same here

abaranau:
one comment

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
chtyim: the job client shouldn't aware of `RunRecord`. The client should be purely just for talking to mr job. Blending of RunRecord for additional / missing information should be outside of this class.

anwar6953:
Removed RunRecord information from MRJobClient and MapReduceMetricsInfo.
Should I make all the fields of MRJobInfo non-final?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -242,6 +243,20 @@ public void mapReduceInfo(HttpRequest request, HttpResponder responder,
       } catch (IOException ioe) {
         LOG.warn("Failed to get run history from JobClient for runId: {}. Falling back to Metrics system.", run, ioe);
         mrJobInfo = mapReduceMetricsInfo.getMRJobInfo(run);
+      } catch (NotFoundException nfe) {
+        // Even if we ran the MapReduce program, there is no guarantee that the JobClient will be able to find it.
+        // For example, if the MapReduce program fails before it successfully submits the job.
+        LOG.warn("Failed to find run history from JobClient for runId: {}. Falling back to Metrics system.", run, nfe);
```
abaranau: should be info. When job history is not there - it'll always spit out.. Ops will freak out for WARNS. Maaybe even debug

abaranau:
minor commment - fix and please merge

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anwar6953: Will do debug, since it can be too common for info.

PR #2054, opened 2015-04-16
Fix MetadataStoreDataset list to handle limit correctly

poornachandra:
MetadataStoreDataset's list was applying limit on records fetched rather than on records returned.


poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT1468-2

anwar6953:
LGTM

PR #2056, opened 2015-04-16
Fix Building of Javadocs under OS X

awholegunch:
Under "OS X" ("Darwin"), always build javadocs under 1.7. This will match what we are intending to do with Bamboo.

Fixes https://issues.cask.co/browse/CDAP-523


duddun:
LGTM

PR #2057, opened 2015-04-16
Feature/multiple measurements metric value

gsps1:
Change MetricValue to support having Collection<Measurement> in a context. 
This will reduce the number of MetricValue objects created and collected at kafka when there are multiple metrics emitted from the same context.
https://issues.cask.co/browse/CDAP-2125


cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricValue.java
```diff
@@ -16,47 +16,48 @@
...
+import co.cask.cdap.api.dataset.lib.cube.MeasureType;
+import co.cask.cdap.api.dataset.lib.cube.Measurement;
+import com.google.common.collect.ImmutableList;
+
+import java.util.Collection;
 import java.util.Map;
 
 /**
- * Carries the "raw" emitted metric data point: metric name, context, type, value, etc.
+ * Carries the "raw" emitted metric data points: context, timestamp, collection<Measurement>
```
abaranau: collection<Measurement> is not valid javadoc format

abaranau: hit control+J - you'll see the output:)

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricValue.java
```diff
@@ -16,47 +16,48 @@
...
-  private final MetricType type;
+
+  public MetricValue(Map<String, String> tags, long timestamp, Collection<Measurement> metrics) {
+    this.tags = tags;
+    this.timestamp = timestamp;
+    this.metrics = metrics;
+  }
 
   public MetricValue(Map<String, String> tags, String name, long timestamp, long value, MetricType type) {
+    MeasureType measureType = type == MetricType.GAUGE ? MeasureType.GAUGE : MeasureType.COUNTER;
```
abaranau: ouch. this is very error prone

abaranau: Should we just use MeasurementType everywhere?

abaranau: Or, alternatively, you'd create specific Measurement class for metrics. MetricDataPoint or smth

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricValue.java
```diff
@@ -16,47 +16,48 @@
...
+import com.google.common.collect.ImmutableList;
+
+import java.util.Collection;
 import java.util.Map;
 
 /**
- * Carries the "raw" emitted metric data point: metric name, context, type, value, etc.
+ * Carries the "raw" emitted metric data points: context, timestamp, collection<Measurement>
  */
 public class MetricValue {
```
abaranau: Should we rename it to MetricPoints? or smth

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricValue.java
abaranau: Seems like MetricValue became exactly same as CubeFact... Maybe use it instead. Let me think - let's discuss

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsEmitter.java
```diff
@@ -33,35 +39,88 @@
   private static final Logger LOG = LoggerFactory.getLogger(AggregatedMetricsEmitter.class);
```
abaranau: javadoc says "A {@link co.cask.cdap.common.metrics.MetricsCollector}" - it is not true, can you fix that?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/collect/AggregatedMetricsEmitter.java
```diff
@@ -33,35 +39,88 @@
...
 
   private final Map<String, String> tags;
```
abaranau: Some logic is wrong. You don't have to keep map of tags in here, as you already have them in a cache entry in a key.
So what you need is
``` java
Cache<Map<String, String>, Cache<String, Measure>>
```

I.e. tags->metric-name->metric

abaranau: it'll simplify stuff a lot

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricValue.java
abaranau: Since you not here, I thought thru it. Let's keep Metrics API classes separate from Cube API classes - we serialize former ones when write data to kafka and we don't want to deal with any changes in them that may be result of e.g. Cube refactor.

That said, let's have classes:
``` java
class MetricValues { // or MetricDataPoints or smth
   long timestamp;
   Map tags;
   Collection<MetricValue> values;
}

MetricValue {
  MetricType type;
  long value;
}

abaranau:
please see comments

abaranau:
walked thru last commit that addresses previous comments. Have only minor comments. Looks good otherwise.

gsps1:
http://builds.cask.co/browse/CDAP-DUT1466 - Green - merging

PR #2058, opened 2015-04-17
CDAP-2143 adding a method to MapReduceContext to set a

albertshau:
StreamBatchReadable as the input to match the static methods
available in StreamBatchReadable and to match what is in the
javadoc for the class. Also updating the ETL template to use the
new method instead of converting StreamBatchReadable to its uri.


albertshau:
http://builds.cask.co/browse/CDAP-DUT1471

chtyim:
LGTM

PR #2060, opened 2015-04-17
CDAP-2144 getStopTs returns Long instead of long.

anwar6953:
Resolves: https://issues.cask.co/browse/CDAP-2144
                RunRecord of a RUNNING program throws NPE upon getStopTime().
                This was happening because the return type of `getStopTs` is `long`, even though its possible for `stopTs` to be null.
Build: http://builds.cask.co/browse/CDAP-DUT1472-2


poornachandra:
LGTM

PR #2061, opened 2015-04-17
Application Template implements Application

gokulavasan:
CDAP-2142 : ApplicationTemplate implements Application.

Application Template can now add datasets (and streams).


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationTemplateManager.java
```diff
@@ -97,6 +98,7 @@ public LocalApplicationTemplateManager(CConfiguration configuration, PipelineFac
     pipeline.addLast(new ApplicationTemplateVerificationStage(store, datasetFramework, adapterService));
     pipeline.addLast(new DeployDatasetModulesStage(configuration, namespace,
                                                    datasetFramework, inMemoryDatasetFramework));
+    pipeline.addLast(new CreateDatasetInstancesStage(configuration, datasetFramework, namespace));
```
albertshau: ApplicationTemplateVerificationStage will have to change too.

albertshau: should we add back stream creation as well?  dont have a use case right now, but if you can create datasets it seems weird that you cant create streams

albertshau:
couple comments otherwise lgtm

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-DUT1473-1

albertshau:
:+1:

PR #2062, opened 2015-04-17
prepend relative path (dot slash) to sourced file

duddun:
The shell bamboo drops into during scripts needs a relative path for handling files in the current directory (based on our testing). It cannot find the file otherwise.

The script runs like this, essentially:
./build.sh <options>
The first line in build.sh (source vars) is failing, because it cannot find vars in the current directory (even though it is there).
This works: 
source ./vars

### DIFF
--- a/cdap-docs/build.sh
+++ b/cdap-docs/build.sh
@@ -26,7 +26,7 @@
 # Builds each of these individually, and then packages them into a single zip file for distribution.
 # _common directory holds common files and scripts.
 
-source vars
+source ./vars
 source _common/common-build.sh
 
 ARG_1="$1"


dereklwood:
:+1 lgtm

cdap-docs/build.sh
```diff
@@ -26,7 +26,7 @@
 # Builds each of these individually, and then packages them into a single zip file for distribution.
 # _common directory holds common files and scripts.
 
-source vars
+source ./vars
 source _common/common-build.sh
```
awholegunch: Should this be ```source ./_common/common-build.sh```?

duddun: "_common/" is a relative path already.
vars implies ./vars (./ is implicit). It needs to be spelled out when we shell out through a bamboo task.

PR #2063, opened 2015-04-17
Set Adpater Worker resources before starting the program

gokulavasan:
Use ProgramOptions (system arguments) to pass instances, resources. Unfortunately this has to be replicated thrice in different places. Need to fix this cleanly in the future.

Bamboo: http://builds.cask.co/browse/CDAP-DUT1476-2


albertshau:
won't different adapters clobber each other this way?

gokulavasan:
Ok will address CDAP-2146 in this PR.

gokulavasan:
@albertshau Addressed your comment. Please review.

albertshau:
:+1:

PR #2064, opened 2015-04-17
Adding Runs REST API for Adapters

gokulavasan:
Will be used by UI to display runs of Adapters. PR has a test for Worker based Adapter runs.
But I will be opening a separate PR for AdapterHttpHandlerTests since it is not present currently.

Bamboo: http://builds.cask.co/browse/CDAP-DUT1458-2


albertshau:
:+1:

PR #2065, opened 2015-04-17
CDAP-2145 changing ETL API so that transforms take one input type

albertshau:
and one output type instead of a pair in and pair out. Also
changing BatchSource to take a key value pair and output a single
type and BatchSink to take a single type and output a key value
pair.


albertshau:
http://builds.cask.co/browse/CDAP-DT18

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchSink.java
```diff
@@ -77,4 +68,23 @@ public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineC
...
    */
   public abstract void prepareJob(BatchSinkContext context);
+
+  /**
+   * Initialize the sink. This is called once each time the Hadoop Job runs, before any
+   * calls to {@link #write} are made.
+   *
+   * @param stageConfig the configuration for the stage.
+   */
+  public void initialize(ETLStage stageConfig) {
```
gokulavasan: So BatchSinkContext here would not make sense, since it exposes hadoop job conf etc.?

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/common/PipelineExecutor.java
```diff
@@ -17,36 +17,44 @@
...
+  public PipelineExecutor(BatchSource source, List<Transform> transforms, BatchSink sink) {
+    this.source = source;
     this.transformList = transforms;
+    this.sink = sink;
     this.previousEmitter = new DefaultEmitter();
     this.currentEmitter = new DefaultEmitter();
   }
 
-  public Iterable<Map.Entry> runOneIteration(Object key, Object value) throws Exception {
+  public void runOneIteration(Object key, Object val, SinkWriter sinkWriter) throws Exception {
```
gokulavasan: I guess we want this to be reused in Worker based driver as well. We can do that in a Worker Driver PR.

gokulavasan:
Couple of observations. No action required. LGTM :100:

PR #2068, opened 2015-04-17
CDAP-2133: Upgrade tool deletes stream conversion adapters

rohitsinha54:
Does the necessary work needed to upgrade Adapters from 2.8 and 3.0 by removing stream-conversion adapter.
There are two places from where the entries need to be removed. 

1. DatasetBasedTimeScheduleStore: Stores all the schedules (triggers) and jobs. 
2. Store (App MetaData Store): Stores the meta data of the adapter.

So, we delete these entries for the stream-conversion adapter across all namespaces.

Build: http://builds.cask.co/browse/CDAP-DT20-1

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
@@ -20,12 +20,15 @@ import co.cask.cdap.api.dataset.module.DatasetModule;
 import co.cask.cdap.api.metrics.MetricStore;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
 import co.cask.cdap.api.schedule.Schedule;
+import co.cask.cdap.app.guice.AppFabricServiceRuntimeModule;
 import co.cask.cdap.app.guice.ProgramRunnerRuntimeModule;
+import co.cask.cdap.app.store.ServiceStore;
 import co.cask.cdap.app.store.Store;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.guice.ConfigModule;
 import co.cask.cdap.common.guice.DiscoveryRuntimeModule;
+import co.cask.cdap.common.guice.KafkaClientModule;
 import co.cask.cdap.common.guice.LocationRuntimeModule;
 import co.cask.cdap.common.guice.TwillModule;
 import co.cask.cdap.common.guice.ZKClientModule;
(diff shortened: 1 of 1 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -505,4 +594,52 @@ private FileMetaDataManager getFileMetaDataManager() {
...
+   *
+   * @return {@link DatasetBasedTimeScheduleStore}
+   * @throws SchedulerException
+   */
+  private DatasetBasedTimeScheduleStore getDatasetBasedTimeScheduleStore() throws SchedulerException {
+    if (datasetBasedTimeScheduleStore == null) {
+      datasetBasedTimeScheduleStore = injector.getInstance(Key.get(DatasetBasedTimeScheduleStore.class,
+                                                                   Names.named("datasetBasedTimeScheduleStore")));
+      // need to call initialize on datasetBasedTimeScheduleStore
+      ExecutorThreadPool threadPool = new ExecutorThreadPool(1);
```
rohitsinha54: This is to create and initialize Quartz scheduler which is needed to initialize the DatasetBasedTimeScheduleStore. 
This is present AppFabricServiceRuntimeModule#getScheduler which we use in injection but is not exposed and is used to get the Supplier. I talked to @albertshau and we were not able to find out a way to get/reuse that one.
@chtyim : Any ideas ? Thanks.

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -423,6 +481,37 @@ private void performUpgrade() throws Exception {
...
+   *
+   * @throws SchedulerException
+   */
+  private void upgradeAdapters() throws SchedulerException {
+    DatasetBasedTimeScheduleStore datasetBasedTimeScheduleStore = getDatasetBasedTimeScheduleStore();
+    NamespaceAdmin namespaceAdmin = injector.getInstance(NamespaceAdmin.class);
+    List<NamespaceMeta> namespaceMetas = namespaceAdmin.listNamespaces();
+    for (NamespaceMeta namespaceMeta : namespaceMetas) {
+      Collection<AdapterSpecification> adapters = getAdapterService().getAdapters(Id.Namespace
+                                                                                    .from(namespaceMeta.getName()));
```
albertshau: nit: namespaceMeta.getName() is used several places, would be less verbose to store it in a variable.

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -505,4 +594,52 @@ private FileMetaDataManager getFileMetaDataManager() {
     }
     return fileMetaDataManager;
   }
+
+  /**
+   * gets the {@link AdapterService}
+   *
+   * @return {@link AdapterService}
+   */
+  private AdapterService getAdapterService() {
```
albertshau: dont need this method, just set it after the injector is setup.

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -216,6 +247,32 @@ public LogSaverTableUtil getLogSaverTableUtil(DatasetFramework dsFramework,
 
         @Provides
         @Singleton
+        @Named("scheduleStoreTableUtil")
```
albertshau: how come there are all these provider methods? Shouldn't things be bound already in the modules already in use?

albertshau:
lgtm, though lets open a jira to clean up the guice setup in this tool...

rohitsinha54:
Build passed here: http://builds.cask.co/browse/CDAP-DT20-5

PR #2070, opened 2015-04-17
(CDAP-2148) Make adapter service thread-safe

chtyim:
- Also fixed a bug in the template upgrade logic
  - It uses file name when update the internal file->template info map, but uses full file path when looking up
- Fix a unit test that relies on the bug mentioned above.


chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1481-1

albertshau:
lgtm

PR #2071, opened 2015-04-17
Fix/ui skip login if security disabled

ajainarayanan:
- Moves login skip part if security is disabled to home feature
- Hides home view when we have security disabled (just skip to dev overview and don't show the home)
- Fixes font-icons


elfenheart:
:+1: LGTM

PR #2072, opened 2015-04-17
CDAP-2137 Get JMS messages from JMS Consumer from async to sync to reduce heap usage

henry-cask:
The first drop for JMS realtime source for template uses internal Queue to support asynchronous push messages from
JMS consumer so when the CDAP template transformer call the JmsSource#poll it already should have message available.

However, this could consume some extra heap allocation given that poll is the mechanism we used to get the
realtime data from CDAP source.
The Queue could get large such that we need to cap it or start dropping older messages if CDAP template transformer
does not poll fast enough.

Need to change the async mode to get messages from JMS provider to sync mode by calling JMS MessageConsumer#receive explicitly.


henry-cask:
got +1 from https://github.com/caskdata/cdap/pull/2052

Just create feature-adapter branch so could be picked up by right CI

henry-cask:
http://builds.cask.co/browse/CDAP-DT21-1

PR #2074, opened 2015-04-17
ETL Worker Adapter Driver

gokulavasan:
* Worker Driver for Realtime Adapters
* Simple test (will be replaced with more meaningful tests when we have more sources/sinks)

Need to fix in future PRs:
* SourceState management. Currently when adapter is deleted we don't have a way to delete the entry in the soureState table corresponding to that adapter.
* DatasetContext access to RealtimeSink through SinkContext is quite hacky. Need to figure out a better way.

Bamboo : http://builds.cask.co/browse/CDAP-DT7


cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/batch/BatchSinkWriter.java
```diff
@@ -22,7 +22,7 @@
  * @param <KEY> the type of key to write
  * @param <VAL> the type of value to write
  */
-public interface SinkWriter<KEY, VAL> {
+public interface BatchSinkWriter<KEY, VAL> {
```
albertshau: why the rename?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLStreamConversionTest.java
```diff
@@ -25,6 +25,7 @@
 import co.cask.cdap.templates.etl.batch.config.ETLBatchConfig;
 import co.cask.cdap.templates.etl.batch.sinks.TimePartitionedFileSetDatasetAvroSink;
 import co.cask.cdap.templates.etl.batch.sources.StreamBatchSource;
+import co.cask.cdap.templates.etl.common.MockAdapterConfigurer;
```
albertshau: don't need this

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
```diff
@@ -0,0 +1,166 @@
...
+    configureTransforms(configurer);
+
+    configurer.addRuntimeArgument(Constants.ADAPTER_NAME, adapterName);
+    configurer.addRuntimeArgument(Constants.CONFIG_KEY, GSON.toJson(etlConfig));
+    configurer.setInstances(instances);
+  }
+
+  private void configureSource(ETLStage sourceConfig, AdapterConfigurer configurer,
+                               PipelineConfigurer pipelineConfigurer) throws Exception {
+    source.configurePipeline(sourceConfig, pipelineConfigurer);
```
albertshau: this stuff is very similar to batch. What if we have a common class 'Stage' or something like that, which has the configure() method that is shared by all sources, sinks, and transforms?  then you could re-use a lot of this code.

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
```diff
@@ -0,0 +1,207 @@
...
+      @Override
+      public void run(DatasetContext context) throws Exception {
+        datasetContext = context;
+      }
+    });
+
+    initializeSource(context, config.getSource());
+    initializeTransforms(context, config.getTransforms());
+
+    // TODO: Execute within a transaction? Dataset operations could be performed in the Sink?
```
albertshau: yup anyplace that can do dataset operations needs to be in a transaction

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
```diff
@@ -0,0 +1,207 @@
...
+
+      final SourceState nextState = source.poll(defaultEmitter, sourceState);
+      for (Object sourceData : defaultEmitter) {
+        try {
+          final Iterable<Object> dataToSink = transformExecutor.runOneIteration(sourceData);
+          getContext().execute(new TxRunnable() {
+            @Override
+            public void run(DatasetContext context) throws Exception {
+              for (Object object : dataToSink) {
+                sink.write(object);
```
albertshau: hm do you think it makes sense to pass the iterable in to the sink? Some sinks may be able to write more efficiently that way

albertshau:
couple comments

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/templates/etl/batch/ETLStreamConversionTest.java
gokulavasan: This is required since I moved this class to etl-common since it is needed in both etl-realtime and etl-batch.

gokulavasan:
@albertshau Addressed your comments. I will add the state management in another PR. Please take a look! Thanks.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/EndPointStage.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package co.cask.cdap.templates.etl.api;
+
+import co.cask.cdap.templates.etl.api.config.ETLStage;
+
+/**
+ * Base class for ETL Source, Sink Stages.
+ */
+public class EndPointStage {
```
albertshau: I think Stage or ETLStage is more descriptive.  Perhaps we should rename the current ETLStage class to StageConfig or ETLStageConfig to avoid confusion?

albertshau: better if this is an interface. Then you can implement this plus something else

albertshau: o i see it's endpoint because transforms don't use it

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
```diff
@@ -0,0 +1,212 @@
...
+      @Override
+      public void run(DatasetContext dsContext) throws Exception {
+        datasetContext = dsContext;
+      }
+    });
+
+    initializeSource(context, config.getSource());
+    initializeTransforms(context, config.getTransforms());
+
+    // Execute within a transaction since Dataset operations could be performed in the Sink.
```
albertshau: do we need to do the same for the source?

gokulavasan: Source does not have access to DatasetContext.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/EndPointStage.java
gokulavasan: will change it to be an interface.

albertshau: since configure is common for all stages, I think it makes more sense for this to be an abstract class called Stage and have source, sink, transforms extend it.  We can move configurePipeline to source and sink.  Then the code to configure sources, sinks, and transforms is all the same rather than one way for sources and sinks, and another for transforms.

gokulavasan: I thought the configure method will be removed once plugin management is in. And so we may not need configure method. And since configurePipeline is specifically for dataset, stream creation it will be applicable only for source and sink.

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/common/StageConfigurator.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.List;
+
+/**
+ * Configurator utility class that has helper methods to configure Source/Sink and Transform stages.
+ */
+public class StageConfigurator {
+  private static final Gson GSON = new Gson();
+
+  public static void configure(EndPointStage stage, ETLStage stageConfig, AdapterConfigurer configurer,
```
albertshau: what do you think of moving this type of logic into a ```configurePipeline(AdapterConfigurer configurer)``` method on a base ETLConfig class that both ETLBatchConfig and ETLRealtimeConfig extend? That class could have source, transforms, and sink configs since they are common for both batch and realtime.  Then the configuration and instantiation logic is all in one place instead of spread out across this utility class and the drivers.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/EndPointStage.java
albertshau: hm ok, ignore the thing about interface then

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/common/StageConfigurator.java
gokulavasan: Will add a base ETLTemplate class which will have common logic that is shared by both ETLRealtimeTemplate and ETLBatchTemplate.

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * ETL Realtime Template.
+ */
+public class ETLRealtimeTemplate extends ETLTemplate<ETLRealtimeConfig> {
+  public static final String STATE_TABLE = "etlrealtimesourcestate";
+  private static final Gson GSON = new Gson();
+  private final Map<String, String> sourceClassMap;
+  private final Map<String, String> sinkClassMap;
+  private final Map<String, String> transformClassMap;
+  private RealtimeSource source;
```
albertshau: this could be a EndPointStage, then instantiation logic can be moved to parent

albertshau: same with sink. could move transforms to parent as well

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
```diff
@@ -0,0 +1,142 @@
...
+        transformClassMap.put(configurer.createSpecification().getName(),
+                              configurer.createSpecification().getClassName());
+      }
+    }
+  }
+
+  @Override
+  public void configureAdapter(String adapterName, ETLRealtimeConfig etlConfig, AdapterConfigurer configurer)
+    throws Exception {
+    ETLStage sourceConfig = etlConfig.getSource();
```
albertshau: most of this is still common between realtime and batch. can have a configureAdapter in parent, then children call super() and only do their specific parts, setting instances or setting a schedule.

albertshau:
couple more things could be consolidated, then lgtm

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
gokulavasan: Good point. Moving.

gokulavasan:
@albertshau Addressed comments. Please take a look again. Build passed : http://builds.cask.co/browse/CDAP-DT7-26

albertshau:
lgtm

PR #2077, opened 2015-04-17
Fix/ui tables styling

astewes:


cdap-ui/app/features/streams/streams.less
```diff
@@ -40,5 +40,25 @@ body.state-streams-detail {
...
+        background-color: @cdap-header;
+        th { color: white; }
+      }
+      tbody {
+        background-color: #DCDCE4;
+        color: @cdap-header;
+      }
+      td { border-top: 1px solid fade(@cdap-bluegray, 40%); }
+    }
+  }
```
elfenheart: I think this should go to theme.less?

astewes: Was initially in theme.less, but Ajai and I moved this style block to both datasets.less and streams.less since everything contained applies specifically to the accordion table styles under the Exlplore tab.

ajainarayanan:
:+1: LGTM

@astewes Going forward if we could break PR like core styles first and then each individual features as individual PRs then it will help people review and would unblock anybody else depending on the core styles.

PR #2078, opened 2015-04-18

jwang47:
https://issues.cask.co/browse/CDAP-2105
http://builds.cask.co/browse/CDAP-DUT1461

After beaming CDAP to a cluster, you can run the tool using the following java command from `/opt/cdap/master`:

```
java -cp `hbase classpath`:"lib/*:../hbase-compat-0.98/lib/*":/etc/cdap/conf/* co.cask.cdap.data2.transaction.queue.hbase.HBaseQueueDebugger queue:///default/PurchaseHistory/PurchaseFlow/reader/queue
```

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/collect/AllPairCollector.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/collect/AllPairCollector.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.common.collect;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+
+import java.util.Map;
+
+/**
+ * This collector will collect until it runs out of memory, it
+ * never returns false.
+ *
+ * @param <KEY> Type of key
+ * @param <VALUE> Type of value
+ */
+public class AllPairCollector<KEY, VALUE> implements PairCollector<KEY, VALUE> {
+
+  private final Multimap<KEY, VALUE> elements = HashMultimap.create();
+
+
+  @Override
+  public boolean addEntries(Iterable<Map.Entry<KEY, VALUE>> entries) {
+    for (Map.Entry<KEY, VALUE> entry : entries) {
+      elements.put(entry.getKey(), entry.getValue());
+    }
+    return true;
+  }
+
+  @Override
+  public <T extends Multimap<? super KEY, ? super VALUE>> T finish(T map) {
+    map.putAll(elements);
+    elements.clear();
+    return map;
+  }
+}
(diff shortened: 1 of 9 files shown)


sreevatsanraman:
Can you add an option to run this from CLI? We should hide this from auto complete so that people don't run this by mistake.

jwang47:
I don't think we should do that, since then cdap-cli would depend on cdap-data-fabric.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,290 @@
...
+
+    byte[] stateColumnName = Bytes.add(QueueEntryRow.STATE_COLUMN_PREFIX,
+                                       Bytes.toBytes(consumerConfig.getGroupId()));
+
+
+    int distributorBuckets = queueClientFactory.getDistributorBuckets(hTable.getTableDescriptor());
+    ShardedHBaseQueueStrategy queueStrategy = new ShardedHBaseQueueStrategy(distributorBuckets);
+
+    Scan scan = new Scan();
+    scan.setStartRow(start.getStartRow());
```
jwang47: Working on fixing this part.

jwang47:
Tested and working in distributed cluster. Ran from `/opt/cdap/master` and used the the following command to run the tool:

```
java -cp `hbase classpath`:"lib/*:../hbase-compat-0.98/lib/*":/etc/cdap/conf/* co.cask.cdap.data2.transaction.queue.hbase.HBaseQueueDebugger queue:///default/PurchaseHistory/PurchaseFlow/reader/queue
```

cdap-common/src/main/java/co/cask/cdap/common/collect/AllPairCollector.java
```diff
@@ -0,0 +1,50 @@
...
+
+package co.cask.cdap.common.collect;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+
+import java.util.Map;
+
+/**
+ * This collector will collect until it runs out of memory, it
```
chtyim: Runs out of memory? Why?

cdap-common/src/main/java/co/cask/cdap/common/collect/PairCollector.java
```diff
@@ -0,0 +1,50 @@
...
+ * This can be used to collect with different strategies while iterating
+ * over a stream of elements. For every element in the stream, add the
+ * element to the collector. The collector then indicates whether more
+ * elements are needed (for instance, to collect the first N elements only,
+ * use a collector that returns false after the Nth element has been added.
+ *
+ * @param <KEY> Type of key.
+ * @param <VALUE> Type of value.
+ */
+public interface PairCollector<KEY, VALUE> {
```
chtyim: Why need this? Can't use the `Collector<Map.Entry<Key, Value>>`?

chtyim: or at least extend from that one and provide an addition method for putting entries into map.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataFabricDistributedModule.java
```diff
@@ -63,44 +50,4 @@ public void configure() {
     install(new TransactionModules().getDistributedModules());
 
   }
-
-  /**
-   * Provides implementation of {@link ThriftClientProvider} based on configuration.
-   */
-  @Singleton
-  private static final class ThriftClientProviderSupplier implements Provider<ThriftClientProvider> {
```
chtyim: Can you use `TransactionClientModule` from Tephra instead? We don't need to change this module now. For the tool purpose, doesn't need to choose client based on cConf.

chtyim:
Need time to review it in details. Probably do it on Monday.

sreevatsanraman:
We will close this and open against 3.0. Since this is a tool and not a part of code. It can come in after code freeze.

PR #2081, opened 2015-04-18
Stop Workflow programs in addition to deleting schedule when stopping a Workflow based Adapter

gokulavasan:
Currently when we stop a Adapter which is based on a Workflow, we delete the schedule. But that does not stop the executing program. So we query the RunRecords and stop the ones that are running.


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterServiceTests.java
```diff
@@ -157,7 +172,56 @@ public void testWorkerAdapter() throws Exception {
   }
 
   @Test
-  public void testAdapters() throws Exception {
+  public void testBatchRunRecords() throws Exception {
```
albertshau: this test runs for a long time and is very reliant on timing, which makes it a very fragile test.  I think it's fine, we don't need to test this and can rely on existing run records and lifecycle tests.

albertshau:
one comment, then lgtm

gokulavasan:
@albertshau Removed test. Please review.

cdap-app-fabric/src/test/java/co/cask/cdap/ActionBatchTemplate.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.common.base.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * App Template to test batch adapter lifecycle.
+ */
+public class ActionBatchTemplate extends ApplicationTemplate<ActionBatchTemplate.Config> {
```
albertshau: dont need this anymore right?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterServiceTests.java
```diff
@@ -66,10 +70,19 @@ public static void setup() throws Exception {
     locationFactory = getInjector().getInstance(LocationFactory.class);
     adapterDir = new File(conf.get(Constants.AppFabric.APP_TEMPLATE_DIR));
     setupAdapters();
+    schedulerService = getInjector().getInstance(SchedulerService.class);
+    schedulerService.startAndWait();
```
albertshau: changes here can be removed.

albertshau:
more things can be removed

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterServiceTests.java
gokulavasan: Hmm true. Apologies.

gokulavasan:
@albertshau Removed all the unnecessary stuff. Build : http://builds.cask.co/browse/CDAP-DUT1486-4

albertshau:
lgtm

PR #2082, opened 2015-04-18
add new docs archive deploy script

duddun:
Preliminary script for deploying docs zip archives.
The intent is to use this as part of the bamboo docs build automation


cdap-docs/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+################################################################################
+### Deploy script for docs
+# Deploys zip files created by build scripts
+# First parameter is cdap (default)
```
dereklwood: First parameter is project (default: cdap)

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+# Second parameter is version
+# third parameter is the remote directory to be created on the web/doc server
+################################################################################
+DEBUG=${DEBUG:-no}
+PROJECT=${1:-cdap}
+PROJECT_DOCS=${PROJECT}-docs
+#PROJECT_PATH=
+VERSION=${2:-2.8.1}
```
dereklwood: does it make sense to have defaults?  Will the default ever be useful?  If I understand correctly, the build should always provide an argument, and so this script should probably just validate and fail if no args given.

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+### Deploy script for docs
+# Deploys zip files created by build scripts
+# First parameter is cdap (default)
+# Second parameter is version
+# third parameter is the remote directory to be created on the web/doc server
+################################################################################
+DEBUG=${DEBUG:-no}
+PROJECT=${1:-cdap}
+PROJECT_DOCS=${PROJECT}-docs
+#PROJECT_PATH=
```
dereklwood: remove comment

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+RSYNC_OPTS='-a --human-readable --progress --stats --rsync-path="sudo rsync"'
+WEB_FILE=${PROJECT}-docs-${VERSION}-web.zip
+GITHUB_FILE=${PROJECT}-docs-${VERSION}-github.zip
+JOB_DIR=/var/bamboo/xml-data/build-dir/CDAP-DRBD-JOB1
+FILE_PATH=${JOB_DIR}/${PROJECT}/${PROJECT_DOCS}/build
+USER=bamboo
+DOCS_SERVER1=docs1.cask.co
+DOCS_SERVER2=docs2.cask.co
+DOCS_SERVERS="${DOCS_SERVER1} ${DOCS_SERVER2}"
+STG_SERVER=docs-staging.cask.co
```
dereklwood: not used?

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+PROJECT_VERSION=${3:-2.8.1}
+RSYNC_OPTS='-a --human-readable --progress --stats --rsync-path="sudo rsync"'
+WEB_FILE=${PROJECT}-docs-${VERSION}-web.zip
+GITHUB_FILE=${PROJECT}-docs-${VERSION}-github.zip
+JOB_DIR=/var/bamboo/xml-data/build-dir/CDAP-DRBD-JOB1
+FILE_PATH=${JOB_DIR}/${PROJECT}/${PROJECT_DOCS}/build
+USER=bamboo
+DOCS_SERVER1=docs1.cask.co
+DOCS_SERVER2=docs2.cask.co
+DOCS_SERVERS="${DOCS_SERVER1} ${DOCS_SERVER2}"
```
dereklwood: These should be bamboo variables

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+
+function rsync_zip_file () {
+  decho "rsync archive"
+  decho "rsync -a -e 'ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null' --human-readable --progress --rsync-path=\"sudo rsync\" ${5}/${4} \"${1}@${2}:${3}/.\""
+  rsync -a -e 'ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null' --human-readable --progress --rsync-path="sudo rsync" ${5}/${4} "${1}@${2}:${3}/." || die "could not rsync ${4} to ${2}"
+  decho ""
+}
+
+function unzip_archive () {
+  decho "going to new staging server and unzipping the file"
```
dereklwood: "unzipping ${4} on remote server ${2}" (or similar) would be better

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+################################################################################
+decho "DEPLOYING"
+
+### DEVELOP => Staging
+deploy ${USER} ${STG_SERVER} ${REMOTE_STG_DIR} ${WEB_FILE} ${FILE_PATH}
+
+### RELEASE => Docs Servers
+for i in ${DOCS_SERVERS}; do
+  decho "not deploying to $i"
+#  deploy ${USER} ${i} ${REMOTE_STG_DIR} ${WEB_FILE} ${FILE_PATH}
```
dereklwood: i'm guessing you need to uncomment, and change the 'not deploying' line?

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+DEBUG=${DEBUG:-no}
+PROJECT=${1:-cdap}
+PROJECT_DOCS=${PROJECT}-docs
+#PROJECT_PATH=
+VERSION=${2:-2.8.1}
+PROJECT_VERSION=${3:-2.8.1}
+RSYNC_OPTS='-a --human-readable --progress --stats --rsync-path="sudo rsync"'
+WEB_FILE=${PROJECT}-docs-${VERSION}-web.zip
+GITHUB_FILE=${PROJECT}-docs-${VERSION}-github.zip
+JOB_DIR=/var/bamboo/xml-data/build-dir/CDAP-DRBD-JOB1
```
dereklwood: this can be passed in using built-in bamboo variables ( ``bamboo.buildResultKey`` I believe).  Useful in case the build name changes

awholegunch:
The existing staging script (stage.sh) lives in the documentation repo, a private repo, because it is for staging any and all docs...should this be there, or should this be here in the branch? Just wondering...it's not exactly CDAP related, in that it takes the output of the build set and does something with it. Wouldn't you want the same script for all builds? That's why I put stage.sh in the documentation repo.

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,165 @@
...
+# Deploys zip files created by build scripts
+#
+# OPT_DIR is an optional directory to use instead of derived branch name. E.g.:
+#   if branch name=feature/docs-build-testing, and we want to use testdir
+#   we would set OPT_DIR=testdir and add it to the bamboo tasks' variables
+#   and the remote directory would be 2.8.1/testdir (testdir treated as branch)
+# If OPT_DIR is not set, bamboo.planRepository.<position>.branch will be used,
+#   e.g. /var/www/html/cdap/2.8.1/feature-myfix, unless the branch name
+#   is release/* or develop/* in which case, it will just be put into
+#   the version directory, e.g. /var/www/html/cdap/2.8.1
```
awholegunch: Will this handle the difference between SNAPSHOT and releases correctly? Just wondering...

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,165 @@
...
+# parameters that can be passed to this script (as environment variables)
+DEBUG=${DEBUG:-no}
+DEPLOY_TO_STG=${DEPLOY_TO_STG:-no}
+DEPLOY_TO_DOCS=${DEPLOY_TO_DOCS:-no}
+BUILD_WORKING_DIR=${BUILD_WORKING_DIR:-/var/bamboo/xml-data/build-dir/CDAP-DRBD-JOB1}
+DOCS_SERVER1=${DOCS_SERVER1:-docs1.cask.co}
+DOCS_SERVER2=${DOCS_SERVER2:-docs2.cask.co}
+STG_SERVER=${STG_SERVER:-docs-staging.cask.co}
+REMOTE_STG_BASE=${REMOTE_STG_BASE:-/var/www/html/cdap}
+REMOTE_DOCS_BASE=${REMOTE_DOCS_BASE:-/var/www/docs/cdap}
```
awholegunch: What about having "cdap" be a variable, so that this can be used for other projects? "coopr", "tigon", and maybe someday "tephra"...

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,165 @@
...
+
+get_version
+get_project
+set_remote_dir
+
+#
+USER=bamboo
+PROJECT_DOCS=${PROJECT}-docs
+WEB_FILE=${PROJECT}-docs-${VERSION}-web.zip
+GITHUB_FILE=${PROJECT}-docs-${VERSION}-github.zip
```
awholegunch: The GitHub pages are never staged. They can be removed from this.

cdap-docs/deploy.sh
awholegunch: Should cdap be ${PROJECT}?

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,165 @@
...
+decho "BRANCH_NAME=${BRANCH_NAME}"
+
+get_version
+get_project
+set_remote_dir
+
+#
+USER=bamboo
+PROJECT_DOCS=${PROJECT}-docs
+WEB_FILE=${PROJECT}-docs-${VERSION}-web.zip
```
awholegunch: Perhaps instead of WEB_FILE, call this ZIP_FILE, as that's what it is.

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,165 @@
...
+  DOC_DIR=`echo $DOC_DIR | tr '/' '-'`
+}
+
+set_remote_dir () {
+  DOC_DIR=${OPT_DIR:-${BRANCH_NAME}}
+  convert_branch_name
+  if [[ "${DOC_DIR}" == release* || "${DOC_DIR}" == develop* ]]; then
+    DIR=''
+  else
+    DIR=branches/${DOC_DIR}
```
awholegunch: Perhaps call this REMOTE_DIR instead of just DIR. It's hard to follow in the script, because it's a portion of so many other commands.

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,165 @@
...
+  decho "ssh ${1}@${2} \"sudo unzip -o ${3}/${4} -d ${3}\""
+  ssh ${1}@${2} "sudo unzip -o ${3}/${4} -d ${3}" || die "unable to unzip ${4} in ${3} on ${2}, as ${1}"
+  decho ""
+}
+
+function deploy () {
+  decho "deploying to ${2}"
+  make_remote_dir ${1} ${2} ${3}
+  rsync_zip_file ${1} ${2} ${3} ${4} ${5}
+  unzip_archive ${1} ${2} ${3} ${4}
```
awholegunch: With numbers instead names, it's hard for me to tell if this is working or going to work. But my first take is that it's not going to.

The existing staging script rsync the file, unzips into a temp directory, copies the zip into the unzipped archive -- which will be named VERSION-[SNAPSHOT], and then moves and renames that version to the version with the branch name. It doesn't look like you are doing this. 

It looks like the branches are becoming a sub-directory, and the version goes under that? Because currently all versions are in the same directory, and named by their branch. See http://stg-web101.sw.joyent.continuuity.net/cdap/
I would prefer to keep the current structure, as it's easy to find things.

Even with that, this looks like things might end up in the wrong location. Hard to tell without running it.

I don't see the step to copy the zip file inside the new directory, so it can be downloaded.

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,165 @@
...
+function convert_branch_name () {
+  decho "converting '/' to '-' in branch name"
+  DOC_DIR=`echo $DOC_DIR | tr '/' '-'`
+}
+
+set_remote_dir () {
+  DOC_DIR=${OPT_DIR:-${BRANCH_NAME}}
+  convert_branch_name
+  if [[ "${DOC_DIR}" == release* || "${DOC_DIR}" == develop* ]]; then
+    DIR=''
```
awholegunch: Should this be:
```
DIR= ${DOC_DIR}
```

cdap-docs/deploy.sh
awholegunch: I don't think this is what we want. Won't you end up with this as the structure?
- 2.8.0
- 3.0.0
- branches
  - 2.8.0-r2.8_fix_javadocs_under_jdk1.7/
  - 3.0.0-SNAPSHOT-3.0_1827_Dataset_Building_Block/
  - 3.0.0-SNAPSHOT-3.0_2070_Metrics_changes/
  - 3.0.0-SNAPSHOT-3.0_edit_dev_man_ingesting/
  - 3.0.0-SNAPSHOT-3.0_Single_Page_RESTful_API/

cdap-docs/deploy.sh
duddun: I grab "cdap" from the project's pom.xml (artifactId).  If we were ever to change CDAP's name to something else, this script would not have to be changed.

cdap-docs/deploy.sh
duddun: this is not a staging script, it is a deployment script

cdap-docs/deploy.sh
duddun: I believe it will, because VERSION would equal 3.0.0-SNAPSHOT, etc..

cdap-docs/deploy.sh
duddun: There is a commit below (641f1a9) that moves the zip file.

cdap-docs/deploy.sh
duddun: I just tested it.  If the pom.xml has the correct version, then it works.

sreevatsanraman:
Why is this not merged? @duddun  What needs to happen next?

duddun:
I am waiting for approval

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,201 @@
...
+# output trimmer
+decho () {
+  if [[ "${DEBUG}" == 'yes' ]]; then
+    echo ${*}
+  else
+    RSYNC_QUIET='--quiet'
+  fi
+}
+
+# set the mandatory variables first (from info in project's main pom.xml)
```
dereklwood: comment out of place

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,201 @@
...
+#     OPT_DIR is not set AND branch=anything else (e.g. feature/*, hotfix/*, etc..)
+#         => remote directory=/var/www/html/cdap/${VERSION}-${BRANCH_NAME}
+set_remote_dir () {
+  DOC_DIR=${OPT_DIR:-${BRANCH_NAME}}
+  convert_branch_name
+  BRANCH=''
+  if [[ "${DOC_DIR}" == release* || "${DOC_DIR}" == develop* ]]; then
+    REMOTE_DIR=''
+  else
+    #REMOTE_DIR=branches/${DOC_DIR}
```
dereklwood: remove commented code

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,201 @@
...
+    BRANCH=yes
+  fi
+  decho "SUBDIR=${REMOTE_DIR}"
+}
+
+# parameters that can be passed to this script (as environment variables)
+DEBUG=${DEBUG:-no}
+DEPLOY_TO_STG=${DEPLOY_TO_STG:-no}
+DEPLOY_TO_DOCS=${DEPLOY_TO_DOCS:-no}
+DOCS_SERVER1=${DOCS_SERVER1:-docs1.cask.co}
```
dereklwood: no defaults for destination hosts... these should need to be explicitly set (from the build), otherwise fail.

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,201 @@
...
+# rsync local dir to remote dir
+function sync_local_dir_to_remote_dir () {
+  local _user=${1}
+  local _host=${2}
+  local _remote_dir=${3}
+  local _zip_file=${4}
+  local _local_dir=${5}
+  local _version=${6}
+  if [ "${BRANCH}" == '' ]; then
+    _remote_dir=${_remote_dir}/${_version}
```
dereklwood: seems like there is a bug here.  this function and make_remote_dir() above are called with the same _remote_dir, but it is modified here and not in make_remote_dir().  are we syncing to a directory not created yet?  also, isn't the remote_dir already set by set_remote_dir()? why is it modified again?

cdap-docs/deploy.sh
```diff
@@ -0,0 +1,201 @@
...
+get_version
+get_project
+set_remote_dir
+
+#
+USER=bamboo
+PROJECT_DOCS=${PROJECT}-docs
+ZIP_FILE=${PROJECT}-docs-${VERSION}-web.zip
+FILE_PATH=${BUILD_WORKING_DIR}/${PROJECT}/${PROJECT_DOCS}/build
+DOCS_SERVERS="${DOCS_SERVER1} ${DOCS_SERVER2}"
```
dereklwood: suggestion (not a blocker): read DOCS_SERVERS as a comma-separated env var.  that way you are not limited to 2.

dereklwood:
couple comments to address, otherwise :+1:

PR #2085, opened 2015-04-18
Feature/ui admin page changes

elfenheart:
- Adding 'source' columns for Configuration
- Fixing Preference bug (cannot delete last preference)

This PR should be checked after #2077 is merged.


ajainarayanan:
:+1: LGTM. Need to rebased or merged with latest develop.

PR #2086, opened 2015-04-18
Feature/ui explore screen change

elfenheart:
Changes to Explore screen:
- Change Dataset explore
- Add Global Explore
- Change View Queries directive

This PR should be checked after #2077 is merged.


cdap-ui/app/directives/sql-query/sql-query.js
```diff
@@ -4,20 +4,18 @@ angular.module(PKG.name + '.commons')
     return {
       restrict: 'E',
       scope: {
-        panel: '='
+        panel: '=',
+        type: '=',
+        name: '='
```
anwar6953: sorry, what does '=' mean as a name or type?

ajainarayanan: @anwar6953 http://stackoverflow.com/a/14910708/661768

cdap-ui/app/directives/view-queries/view-queries.js
```diff
@@ -28,28 +33,42 @@ angular.module(PKG.name + '.commons')
...
+              });
+
+            // request preview
+            dataSrc
+              .request({
+                _cdapPath: '/data/explore/queries/' +
+                              query.query_handle + '/preview',
+                method: 'POST'
+              })
+              .then(function (result) {
```
ajainarayanan: Not necessarily a bad thing but we already have 'result' being fetched (/schema). This is a preview of the results. Thought it would be better if this is 'results'. Feel too many results are there in this controller :smile:

ajainarayanan:
:+1: LGTM

PR #2087, opened 2015-04-18
CDAP-2067 projection transform for the etl template, which allows

albertshau:
dropping fields, renaming fields, and casting types.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/KeyValueListParser.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/KeyValueListParser.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.common;
+
+import co.cask.cdap.api.dataset.lib.KeyValue;
+import com.google.common.base.Splitter;
+
+import java.util.Iterator;
+import java.util.regex.Pattern;
+
+/**
+ * Parses a list of key value pairs.
+ */
+public class KeyValueListParser {
+  private final Pattern pairDelimiter;
+  private final Pattern keyValDelimiter;
+
+  /**
+   * Create a parser that uses the given regexes to parse a list of key value pairs.
+   *
+   * @param pairDelimiterRegex the delimiter between key value pairs.
+   * @param keyValDelimiterRegex the delimiter between a key and value.
+   */
+  public KeyValueListParser(String pairDelimiterRegex, String keyValDelimiterRegex) {
+    pairDelimiter = Pattern.compile(pairDelimiterRegex);
+    keyValDelimiter = Pattern.compile(keyValDelimiterRegex);
+  }
+
+  /**
+   * Parses the given list of key value pairs.
+   *
+   * @param kvList the string to parse
+   * @return an iterable of key values
+   */
+  public Iterable<KeyValue<String, String>> parse(String kvList) {
+    return new KeyValueIterable(kvList);
+  }
+
+  private class KeyValueIterable implements Iterable<KeyValue<String, String>> {
+    private final String kvList;
+
+    private KeyValueIterable(String kvList) {
+      this.kvList = kvList;
+    }
+
+    @Override
+    public Iterator<KeyValue<String, String>> iterator() {
+      return new KeyValueIterator(kvList);
+    }
+  }
+
+  private class KeyValueIterator implements Iterator<KeyValue<String, String>> {
+    private final Iterator<String> pairIter;
+
+    private KeyValueIterator(String kvList) {
+      pairIter = Splitter.on(pairDelimiter).split(kvList).iterator();
+    }
+
+    @Override
+    public boolean hasNext() {
+      return pairIter.hasNext();
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/KeyValueListParser.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    public boolean hasNext() {
+      return pairIter.hasNext();
+    }
+
+    @Override
+    public KeyValue<String, String> next() {
+      String pair = pairIter.next();
```
sreevatsanraman: do a `next()` only if `hasNext()` succeeds otherwise handle `pair == null`

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/ProjectionTransform.java
```diff
@@ -0,0 +1,301 @@
...
+/**
+ * Projection transform that allows dropping, renaming, and converting field types.
+ */
+public class ProjectionTransform extends Transform<StructuredRecord, StructuredRecord> {
+  private static final String DROP = "drop";
+  private static final String RENAME = "rename";
+  private static final String CONVERT = "convert";
+  private static final Pattern fieldDelimiter = Pattern.compile("\\s*,\\s*");
+  private Set<String> fieldsToDrop = Sets.newHashSet();
+  private Map<String, String> fieldsToRename = Maps.newHashMap();
```
sreevatsanraman: We should use a `BiMap` to catch renaming two source fields to one target fields which could be an error. Example X -> Y, Z -> Y

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/ProjectionTransform.java
```diff
@@ -0,0 +1,301 @@
...
+  private Map<String, Schema.Type> fieldsToConvert = Maps.newHashMap();
```
sreevatsanraman: Same thing here.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/ProjectionTransform.java
```diff
@@ -0,0 +1,301 @@
...
+      for (String dropField : Splitter.on(fieldDelimiter).split(dropStr)) {
+        fieldsToDrop.add(dropField);
+      }
+    }
+
+    KeyValueListParser kvParser = new KeyValueListParser("\\s*,\\s*", ":");
+    String renameStr = properties.get(RENAME);
+    if (renameStr != null) {
+      for (KeyValue<String, String> keyVal : kvParser.parse(renameStr)) {
+        fieldsToRename.put(keyVal.getKey(), keyVal.getValue());
```
sreevatsanraman: Check if the value already exists before insert. If so that will be an initialization error. Renaming two source fields to a single target

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/ProjectionTransform.java
```diff
@@ -0,0 +1,301 @@
...
+    String convertStr = properties.get(CONVERT);
+    if (convertStr != null) {
+      for (KeyValue<String, String> keyVal : kvParser.parse(convertStr)) {
+        String name = keyVal.getKey();
+        String typeStr = keyVal.getValue();
+        Schema.Type type = Schema.Type.valueOf(typeStr.toUpperCase());
+        if (!type.isSimpleType() || type == Schema.Type.NULL) {
+          throw new IllegalArgumentException("Only non-null simple types are supported.");
+        }
+        fieldsToConvert.put(name, type);
```
sreevatsanraman: Same check here. Check if value exists.

sreevatsanraman:
Overall very clean code! A few things needs to be fixed.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/KeyValueListParser.java
albertshau: keeping as is because the iterator contract is to throw an exception if you call next() when there isn't anything next, as nulls are valid elements to return.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/ProjectionTransform.java
albertshau: good point, I will check when populating the map so that we can give a more descriptive error message

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/ProjectionTransform.java
albertshau: this one doesn't need to be a bimap since multiple fields can be converted to a string for example. but will check that each key only shows up once when initializing

albertshau:
@sreevatsanraman addressed comments, could you take another look?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/KeyValueListParser.java
sreevatsanraman: ok.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/ProjectionTransform.java
sreevatsanraman: yeah, i meant the checks. good.

sreevatsanraman:
LGTM :+1:

PR #2088, opened 2015-04-18
Fix/ui run level logs

ajainarayanan:
Adds run level logs tab to workflows + flows + overview (no apps & no data) view

### DIFF
--- a/cdap-ui/app/features/mapreduce/mapreduce.less
+++ b/cdap-ui/app/features/mapreduce/mapreduce.less
@@ -27,7 +27,12 @@ body.state-mapreduce{
       border-right: 1px solid black;
     }
   }
+
   &.theme-cdap {
     .table.table-status { margin-top: 0; }
-  }
+    .btn.btn-default.no-border{
+      &:focus{
+        outline: none;
+      }
+    }
 }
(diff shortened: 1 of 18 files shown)


cdap-ui/app/features/mapreduce/mapreduce.less
```diff
@@ -27,4 +27,9 @@ body.state-mapreduce{
       border-right: 1px solid black;
     }
   }
+  .btn.btn-default.no-border{
+    &:focus{
+      outline: none;
+    }
+  }
```
elfenheart: Since this is repeated several times, should we extract this to either theme.less or common.less?

ajainarayanan: Thought this is only specific for program types these should be in respective programs. Was not too sure how to create a program level styling.

cdap-ui/app/features/overview/templates/data-apps.html
```diff
@@ -119,7 +122,10 @@ <h3 class="text-uppercase"> Apps </h3>
             data-button-icon="fa-plus"
             on-file-select="onFileSelected($files)">
         </my-file-select>
-        <a ui-sref="etlapps.create" class="btn btn-default">Add ETL App</a>
+        <a ui-sref="etlapps.create" class="btn btn-default">
+          <span class="fa fa-plus"></span>
+          <span>Add ETL App</span>
```
elfenheart: Change the wording to Add ETL Template?

elfenheart:
Other than the comment, :+1: LGTM

PR #2092, opened 2015-04-18
Not allow operations on an template application outside of adapter operations

gokulavasan:
* In the delete application handler method, check if the application is deployed and if it is then check if its an application template, then return 403. Unit Tests to verify the same behavior.

Bamboo : http://builds.cask.co/browse/CDAP-DUT1499


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -327,6 +330,15 @@ public void performAction(HttpRequest request, HttpResponder responder,
                             @PathParam("type") String type,
                             @PathParam("id") String id,
                             @PathParam("action") String action) {
+    // If the app is deployed and it is an Application Template, then don't allow any action.
+    // Operations are only allowed through Adapter Lifecycle management.
+    if (store.getApplication(Id.Application.from(namespaceId, appId)) != null &&
```
albertshau: don't think you need to get it from the store, just check adapter service

gokulavasan: Hmm for example, say the user wants to deploy an app which matches with a apptemplate name. Should we not allow that even though no adapter was created? Though the user won't be able to create an adapter of that type after that (until he deletes his app). Thoughts?

albertshau: yes we should not allow that.

albertshau: looks like we check on delete, but not on create. should check on create as well.

gokulavasan:

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterLifecycleTests.java
```diff
@@ -217,28 +217,6 @@ public void testInvalidConfigReturns400() throws Exception {
     // TODO: implement once adapter creation calls configureTemplate()
   }
 
-  @Test
```
gokulavasan: This test is being removed since we don't deploy the template again if the JAR contents have not changed and thus this test is not correct.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -423,6 +414,11 @@ public boolean apply(Id.Program programId) {
   }
 
   private AppFabricServiceStatus removeApplication(final Id.Application appId) throws Exception {
+    // Deletion of a particular application is not allowed if that application is used by an adapter
```
gokulavasan: This logic is moved from here so that it gets executed for both the handlers of /apps/app-id DELETE and /apps DELETE

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterLifecycleTests.java
```diff
@@ -136,7 +136,7 @@ private void testAdapterLifeCycle(String namespaceId, String templateId, String
     Assert.assertEquals("STARTED", status);
 
     // Deleting App should fail
-    deleteApplication(1, deleteURL, 400);
+    deleteApplication(1, deleteURL, 403);
```
gokulavasan: The check should be for forbidden. not bad request.

bdmogal:
So we do not restrict get operations on adapter workflows through ProgramLifecycleHttpHandler, right? If that's by design, LGTM

gokulavasan:
Ya, as long as it does not involve starting, stopping, suspending etc, it should be fine.

PR #2093, opened 2015-04-18
Making resources (mapper resources) part of the ETLBatchConfig

gokulavasan:
Instead of having it as part of the BatchContext.

Bamboo : http://builds.cask.co/browse/CDAP-DT25


cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
```diff
@@ -77,6 +74,7 @@ public void beforeSubmit(MapReduceContext context) throws Exception {
     prepareSource(context, config.getSource());
     prepareSink(context, config.getSink());
 
+    context.setMapperResources(config.getResources());
```
albertshau: resources can be null right? if so should check before setting.

albertshau:
one comment, otherwise lgtm

gokulavasan:

gokulavasan:
Build passed : http://builds.cask.co/browse/CDAP-DT25-5

albertshau:
lgtm

PR #2094, opened 2015-04-18
Remove old UI and make new UI the default UI

nitinmotgi:
NOTE: Even though there are many files in this PR, most of it because of the removal of cdap-web-app (All those node_modules)

Following are the only real changes that have been done in this PR 

- [x] cdap.sh and cdap.bat modified to use and load the new UI in standalone
- [x] pom.xml of standalone modified to not build old cdap-web-app 
- [x] remove module cdap-web-app from main project. 
- [x] Moved ConfigurationJsonTool from common to UI & removed from common.
- [x] Renamed everything from web-app to UI (Java, Bash, and BAT)

*This PR has been tested in the following environments. *

- [x] Standalone 
- [x] Distributed


awholegunch:
Need to update the usage statement for starting the cdap.sh and remove
```
      echo "--enable-alpha-ui Start new ui"
```
Probably also true for the cdap.bat.

awholegunch:
OK, cdap.bat looked OK. Copyrights should be 2014-2015 (yeah, right. 5,501 files...)

awholegunch:
LGTM

anew:
I cannot review the entire PR because it is too large to show in a browser:

``Sorry, we could not display the entire diff because too many files (5,501) changed.``

I will review what I can.

cdap-standalone/bin/cdap.bat
```diff
@@ -24,7 +24,7 @@ SET CDAP_HOME=%CDAP_HOME:~0,-5%
 SET JAVACMD=%JAVA_HOME%\bin\java.exe
 
 REM Specifies Web App Path
-SET WEB_APP_PATH=%CDAP_HOME%\web-app\local\server\main.js
+SET UI_PATH=%CDAP_HOME%\ui\alpha\server.js
```
anew: why is the path still alpha? If the old UI is gone, then this should be just /ui/

cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
```diff
@@ -114,7 +113,7 @@ private StandaloneMain(List<Module> modules, CConfiguration configuration, Strin
     serviceStore = injector.getInstance(ServiceStore.class);
     streamService = injector.getInstance(StreamService.class);
 
-    webCloudAppService = (webAppPath == null) ? null : injector.getInstance(WebCloudAppService.class);
+    userInterfaceService = (uiPath == null) ? null : injector.getInstance(UserInterfaceService.class);
```
anew: how can it ever be null?

cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
```diff
@@ -249,12 +248,12 @@ static void usage(boolean error) {
...
       out.println("  cdap.bat [options]");
     } else {
-      out.println("  ./cdap.sh [options]");
+      out.println("  cdap.sh [options]");
     }
     out.println("");
     out.println("Additional options:");
-    out.println("  --web-app-path  Path to Webapp");
-    out.println("  --help          To print this message");
+    out.println("  --ui-path  Path to UI");
```
anew: is that still needed, now that we have only one UI?

cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
```diff
@@ -249,12 +248,12 @@ static void usage(boolean error) {
     if (OSDetector.isWindows()) {
       out.println("  cdap.bat [options]");
     } else {
-      out.println("  ./cdap.sh [options]");
+      out.println("  cdap.sh [options]");
```
anew: it is kind of weird that the Java code prints the options for the shell/bat script. But since it does, should it mention --enable-debug?

anew:
This is great. Please see just a few comments from me.

cdap-standalone/bin/cdap.bat
nitinmotgi: We will changing that from alpha to non-alpha after all UI PR's are merged.

cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
nitinmotgi: You were right! cleaned up! now

cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
nitinmotgi: I thought it would be useful for later, but now that I think again, it's better to remove it now.

anew:
Merge it!

PR #2095, opened 2015-04-18
(CDAP-2160) Fix the stream coordination when deleting stream

chtyim:
- Instead of incrementing the generation, a stream deleted event should be sent.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/AbstractStreamCoordinatorClient.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/AbstractStreamCoordinatorClient.java
@@ -83,11 +83,7 @@ public abstract class AbstractStreamCoordinatorClient extends AbstractIdleServic
   protected abstract void streamCreated(Id.Stream streamId);
 
   /**
-<<<<<<< HEAD
-   * GEts invoked when a stream is deleted.
-=======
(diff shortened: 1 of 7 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/AbstractStreamCoordinatorClient.java
```diff
@@ -118,15 +114,15 @@ public void updateProperties(Id.Stream streamId, Callable<CoordinatorStreamPrope
...
     Lock lock = getLock(streamId);
     lock.lock();
     try {
-      CoordinatorStreamProperties properties = action.call();
-      if (properties != null) {
-        updateProperties(streamId, properties).get();
-        streamDeleted(streamId);
-      }
+      action.run();
+      // Ideally would be deleting the property. However it is not supported by PropertyStore right now.
```
anew: make this a TODO?

chtyim: Sure. There is a JIRA created already (see next line, CDAP-2161).

anew:
LGTM

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1501-2

PR #2096, opened 2015-04-18
Introduce non-realtime charts

anwar6953:
Rename realtime charts in order to introduce non realtime - ['basic' epoch](http://fastly.github.io/epoch/basic/) - charts.


anwar6953:
Closing.
Will reopen at a later point in time, once the capability to poll at variable frequencies is implemented (so that that can be used here).

PR #2099, opened 2015-04-18
[CDAP-2066] Added DB Batch Sink

bdmogal:
Adds a ``DBSink`` that:
1. Reads a ``StructuredRecord``
2. Makes a JDBC call to the destination table to figure out the schema of the destination table as ``ResultSetMetadata``. This is needed because Database Drivers do not reliably return the schema from a ``PreparedStatement``.
3. Constructs a ``DBRecord`` using the ``StructuredRecord`` and ``ResultSetMetadata`` from 1 and 2 above.
4. Sends the ``DBRecord`` a custom ``ETLDBOutputFormat``, which is needed to load the driver class properly, just like we do in ``ETLDBInputFormat``.

- [x] Tested most SQL column types.
- [x] Tested on a local MySQL.
- [x] Tested a CDAP Table to MySQL table ETL pipeline on a cluster.
- [ ] The unit test for this fails, because when ``DBOutputFormat`` calls ``PreparedStatement.executeBatch()``, HSQLDB throws an exception saying *Statement is not in batch mode.* I'm trying to debug this a little more, but if it does not work, we may have to rely instead on an audi test that sets up a local mysql for this.

Jira: [CDAP-2066](https://issues.cask.co/browse/CDAP-2066)
Build: https://builds.cask.co/browse/CDAP-DT26

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/DBSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/DBSink.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.batch.sinks;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.templates.etl.api.Emitter;
+import co.cask.cdap.templates.etl.api.PipelineConfigurer;
+import co.cask.cdap.templates.etl.api.Property;
+import co.cask.cdap.templates.etl.api.StageConfigurer;
+import co.cask.cdap.templates.etl.api.batch.BatchSink;
+import co.cask.cdap.templates.etl.api.batch.BatchSinkContext;
+import co.cask.cdap.templates.etl.api.config.ETLStage;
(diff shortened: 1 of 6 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/DBSink.java
```diff
@@ -0,0 +1,171 @@
...
+      new Property(Properties.DB.CONNECTION_STRING, "JDBC connection string including database name", true));
+    configurer.addProperty(
+      new Property(Properties.DB.USER,
+                   "User to use to connect to the specified database. " +
+                     "Required for databases that require authentication. " +
+                     "Optional for databases that do not require authentication.",
+                   false));
+    configurer.addProperty(
+      new Property(Properties.DB.PASSWORD,
+                   "Password to be use to connect to the specified database. " +
```
albertshau: typo: Password to use

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/ETLDBOutputFormat.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Class that extends {@link DBOutputFormat} to load the database driver class correctly.
+ *
+ * @param <K> - Key passed to this class to be written
+ * @param <V> - Value passed to this class to be written. The value is ignored.
+ *
+ * {@inheritDoc}
+ */
+public class ETLDBOutputFormat<K  extends DBWritable, V>  extends DBOutputFormat<K, V> {
```
albertshau: remove extra space between K and extends

albertshau:

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/DBSink.java
bdmogal: Fixed

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/common/ETLDBOutputFormat.java
bdmogal: Fixed

bdmogal:
Fixed nits. Rebased with develop. Tested a complete local run (unit tests with local mysql). 

Latest Build Passed - http://builds.cask.co/browse/CDAP-DT26-5. Merging.

PR #2100, opened 2015-04-18
Errors Chart on overview page

anwar6953:
System and app errors on overview page, using c3.

Perhaps a single number should also be visible (either aggregated for past 5 minutes or past hour).

Note: Because it is based off of c3 charts, it depends on: https://github.com/caskdata/cdap/pull/2184

### DIFF
--- a/cdap-ui/app/features/dashboard/controllers/widget.js
+++ b/cdap-ui/app/features/dashboard/controllers/widget.js
@@ -3,16 +3,20 @@
  */
 
 angular.module(PKG.name+'.feature.dashboard')
-  .factory('Widget', function (MyDataSource) {
+  .factory('Widget', function (MyDataSource, myHelpers) {
 
     function Widget (opts) {
       opts = opts || {};
       this.title = opts.title || 'Widget';
       this.type = opts.type;
       this.metric = opts.metric || false;
+      // TODO: reconsider this field once Epoch is removed.
       this.color = opts.color;
       this.dataSrc = null;
-      this.isLive = false;
+      this.isLive = opts.isLive || false;
+      this.interval = opts.interval;
+      this.aggregate = opts.aggregate;
+      this.metricAlias =  opts.metricAlias || {};
     }
 
     // 'ns.default.app.foo' -> {'ns': 'default', 'app': 'foo'}
@@ -29,11 +33,15 @@ angular.module(PKG.name+'.feature.dashboard')
       return tags;
     }
 
-    function constructQuery(queryId, tags, metrics) {
+    function constructQuery(queryId, tags, metric) {
       var timeRange, retObj;
-      timeRange = {'start': 'now-60s', 'end': 'now'};
+      timeRange = {'start': metric.startTime || 'now-60s',
+                  'end': metric.endTime || 'now'};
+      if (metric.resolution) {
+        timeRange['resolution'] = metric.resolution;
+      }
       retObj = {};
-      retObj[queryId] = {tags: tags, metrics: metrics, groupBy: [], timeRange: timeRange};
+      retObj[queryId] = {tags: tags, metrics: metric.names, groupBy: [], timeRange: timeRange};
       return retObj;
     }
 
@@ -47,7 +55,7 @@ angular.module(PKG.name+'.feature.dashboard')
       this.dataSrc.request({
         _cdapPath: '/metrics/query',
         method: 'POST',
-        body: constructQuery(queryId, contextToTags(this.metric.context), this.metric.names)
+        body: constructQuery(queryId, contextToTags(this.metric.context), this.metric)
       })
         .then(this.processData.bind(this))
     }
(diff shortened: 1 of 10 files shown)


cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -73,6 +77,33 @@ angular.module(PKG.name+'.feature.dashboard')
...
+        // instead of for the whole range. We have to interpolate the rest with 0s to draw the graph.
+        if (resolution == '1h') {
+          skipAmt = 60 * 60;
+        } else if (resolution == '1m') {
+          skipAmt = 60;
+        } else {
+          skipAmt = 1;
+        }
+
+        var startTime = roundUpToNearest(result.startTime, skipAmt);
```
ajainarayanan: Isn't startTime in epoch time and skipAmt plain number?

cdap-ui/app/features/overview/overview.html
```diff
@@ -99,16 +99,28 @@ <h3 class="text-uppercase"> Accelerate... </h3>
...
+        <div ng-repeat="wdgt in wdgts" class="col-xs-5">
+          <p class="text-left"> {{wdgt.title}} </p>
+
+          <div ng-controller="WidgetTimeseriesCtrl" class="{{wdgt.color}}">
+            <!--{{chartHistory[0].values.slice(-2)[0].y}}--> <!-- This would be the value for the last complete minute-->
+            <div ng-show="stream.length">
+              {{stream[0].y}}
+            </div>
+            <div ng-hide="stream.length"> N/A </div>
+            <!-- Stream not passed to Epoch because we do not currently support variable-frequency polling -->
```
ajainarayanan: We have this now right?

cdap-ui/app/features/overview/overview.html
```diff
@@ -99,16 +99,28 @@ <h3 class="text-uppercase"> Accelerate... </h3>
...
-        <p class="text-center"> App </p>
-      </div>
+
+        <div ng-repeat="wdgt in wdgts" class="col-xs-5">
+          <p class="text-left"> {{wdgt.title}} </p>
+
+          <div ng-controller="WidgetTimeseriesCtrl" class="{{wdgt.color}}">
+            <!--{{chartHistory[0].values.slice(-2)[0].y}}--> <!-- This would be the value for the last complete minute-->
+            <div ng-show="stream.length">
+              {{stream[0].y}}
```
ajainarayanan: Isn't this now removed from ```TimeSeriesCtrl```?

cdap-ui/app/features/overview/controllers/overview-ctrl.js
```diff
@@ -73,4 +73,27 @@ function ($scope, MyDataSource, $state, myLocalStorage, MY_CONFIG) {
     }
   });
 
+  $scope.wdgts = [];
+  var errLogMetric = 'system.log.error';
+  // type field is overridden by what is rendered in view because we do not use widget.getPartial()
+  $scope.wdgts.push(new Widget({title: 'Errors last minute - System', type: 'line', isLive: true,
```
ajainarayanan: Do we need ```isLive``` any more? We are no longer having constant polling right? The resolution is anyway 1m

anwar6953: Right, I wanted it to be live, with a frequency of 30s or 1m. (still would be live though, just low frequency)

cdap-ui/app/features/overview/overview.html
anwar6953: ah yeah, we didn't have it at the time.

cdap-ui/app/features/dashboard/controllers/widget.js
anwar6953: right. you can still round an epoch time to the nearest 60 or the nearest 3600 though.

cdap-ui/app/features/overview/controllers/overview-ctrl.js
ajainarayanan: So now making a chart live becomes a property of widget instead of the chart? If I use epoch-line-chart directive somewhere I still need to use widget to make it a live one?

cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -136,7 +162,6 @@ angular.module(PKG.name+'.feature.dashboard')
     });
     $scope.wdgt.fetchData($scope);
     $scope.chartHistory = null;
-    $scope.stream = null;
```
ajainarayanan: Any specific reason why we are removing this? Its generally a better practice to declare the variable inside a scope before assigning with values from backend/services

anwar6953: Ah ok. I'll undo this.
I thought by default its 'undefined', didn't see the point of setting it to null.

anwar6953:
Discussed with Ajai offline. Looks good to merge.

PR #2108, opened 2015-04-19
Make chart color a hash of metric name. Fix a bug which occurred if metr...

anwar6953:
Make chart color a hash of metric name.
Fix a bug which occurred if metric name had '>', '.\d', or other illegal jquery selector characters.


ajainarayanan:
:+1: LGTM

PR #2110, opened 2015-04-19
CDAP-2169: Delete the app if the last adapter got deleted

rohitsinha54:
Build: http://builds.cask.co/browse/CDAP-DT30-1

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
@@ -290,7 +290,8 @@ public class AdapterService extends AbstractIdleService {
   }
 
   /**
-   * Remove adapter identified by the namespace and name.
+   * Remove adapter identified by the namespace and name and also deletes the template if this was last adapter
+   * of that type
    *
    * @param namespace namespace id
    * @param adapterName adapter name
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -301,12 +302,24 @@ public void removeAdapter(Id.Namespace namespace, String adapterName)
...
-    // TODO: Delete the application if this is the last adapter
+    // delete the template if this was the last adapter of that type
+    deleteApp(Id.Application.from(namespace, adapterSpec.getTemplate()));
+  }
+
+  /**
+   * deletes the app if it has no associated adapters
+   * @param templateId the {@link Id.Application} of the template
+   */
+  private void deleteApp(Id.Application templateId) {
```
gokulavasan: Not sure if this is enough. Take a look at the AppLifecycleHttpHandler and see what all we when we delete an app. Shouldn't we cover all that?

rohitsinha54:
Develop build : http://builds.cask.co/browse/CDAP-DUT1513-1

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
rohitsinha54: Yes. Make sense.  Will do that.

rohitsinha54:
New PR: https://github.com/caskdata/cdap/pull/2204

PR #2112, opened 2015-04-19
ETL Realtime Adapter SourceState management

gokulavasan:
* Realtime Adapters state is stored in a KV Table. We assign a unique id when the adapter is created and then this is used to identify re-runs of that specific adapter and thus the old state is reused. When the user deletes the adapter and creates another one with the same name, a different id is generated in configureAdapter and when the worker is started, all the other ids for that adapter name is deleted and thus old state for the same adapter name is removed.

Open Issues:
i) Since there are no simple realtime sources, sinks yet, it is hard to test this now. Will add a test in a later PR.
ii) State of the adapter is left in the table and cleaned up only when a new etl-realtime adapter of the same name is created. Will open a JIRA for this [https://issues.cask.co/browse/CDAP-2175]

Bamboo : http://builds.cask.co/browse/CDAP-DT31


cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/templates/etl/common/Constants.java
```diff
@@ -44,4 +44,11 @@
   public static final class Transform {
     public static final String SPECIFICATIONS = "templates.etl.adapter.transform.specifications";
   }
+
+  /**
+   * Constants related to Realtime Adapter.
+   */
+  public static final class Realtime {
```
henry-cask: Just notice this, so we are creating classes for each constants for just grouping? Could we just make it like flat Strings inside the Constants class but prefix it for each module?

chtyim: It's easier to read as long as we don't create a lot of inner class for that. E.g.

```java
Constants.REAL_TIME_UNIQUE_ID
Constants.REAL_TIME_TIMEOUT
Constants.REAL_TIME_DELAY
```

vs

```java
Constants.RealTime.UNIQUE_ID
Constants.RealTime.TIMEOUT
Constants.RealTime.DELAY
```

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
```diff
@@ -52,6 +54,8 @@ public void configureAdapter(String adapterName, ETLRealtimeConfig etlConfig, Ad
     super.configureAdapter(adapterName, etlConfig, configurer);
     configurer.addRuntimeArgument(Constants.CONFIG_KEY, GSON.toJson(etlConfig));
     configurer.setInstances(etlConfig.getInstances());
+    // Generate unique id for this adapter creation.
+    configurer.addRuntimeArgument(Constants.Realtime.UNIQUE_ID, UUID.randomUUID().toString());
```
chtyim: You can just use timestamp, assuming you don't create two adapter instances in the same millisecond.

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
```diff
@@ -53,14 +54,15 @@
...
+  private static final String SEPARATOR = ":";
 
   private String adapterName;
   private RealtimeSource source;
   private RealtimeSink sink;
   private List<TransformStage> transforms;
-  private DatasetContext datasetContext;
   private TransformExecutor transformExecutor;
   private DefaultEmitter defaultEmitter;
+  private String matchKey;
```
chtyim: That's not a very good name. Maybe something like `stateStoreKey`?

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/test/java/co/cask/cdap/templates/etl/realtime/ETLWorkerTest.java
```diff
@@ -64,7 +64,7 @@ public void testSimpleConfig() throws Exception {
     }
 
     WorkerManager workerManager = templateManager.startWorker(ETLWorker.class.getSimpleName(), workerArgs);
-    TimeUnit.SECONDS.sleep(5);
+    TimeUnit.SECONDS.sleep(2);
```
chtyim: Why sleep?

gokulavasan:
@chtyim Addressed PR comments. Please take a look again.

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/test/java/co/cask/cdap/templates/etl/realtime/ETLWorkerTest.java
gokulavasan: This was added since we wanted a very preliminary check on the adapter execution logic in the worker driver. Once we have proper sinks, we will remove this test and have proper tests.

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
```diff
@@ -52,6 +52,8 @@ public void configureAdapter(String adapterName, ETLRealtimeConfig etlConfig, Ad
     super.configureAdapter(adapterName, etlConfig, configurer);
     configurer.addRuntimeArgument(Constants.CONFIG_KEY, GSON.toJson(etlConfig));
     configurer.setInstances(etlConfig.getInstances());
+    // Generate unique id for this adapter creation.
+    configurer.addRuntimeArgument(Constants.Realtime.UNIQUE_ID, String.valueOf(System.currentTimeMillis()));
```
chtyim: Maybe a good idea to have the configurer.addRuntimeArgument to support different primitive types as well. No need to be this PR, but good to note.

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
```diff
@@ -80,16 +82,32 @@ public void initialize(final WorkerContext context) throws Exception {
     Preconditions.checkArgument(runtimeArgs.containsKey(Constants.Source.SPECIFICATION));
     Preconditions.checkArgument(runtimeArgs.containsKey(Constants.Sink.SPECIFICATION));
     Preconditions.checkArgument(runtimeArgs.containsKey(Constants.Transform.SPECIFICATIONS));
+    Preconditions.checkArgument(runtimeArgs.containsKey(Constants.Realtime.UNIQUE_ID));
 
     adapterName = runtimeArgs.get(Constants.ADAPTER_NAME);
```
chtyim: What happen if it is missing?

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
```diff
@@ -80,16 +82,32 @@ public void initialize(final WorkerContext context) throws Exception {
...
-        datasetContext = dsContext;
+        KeyValueTable stateTable = dsContext.getDataset(ETLRealtimeTemplate.STATE_TABLE);
+        byte[] startKey = Bytes.toBytes(String.format("%s%s", adapterName, SEPARATOR));
+        // Scan the table for adaptername: prefixes and remove rows which doesn't match the unique id of this adapter.
+        CloseableIterator<KeyValue<byte[], byte[]>> rows = stateTable.scan(startKey, Bytes.stopKeyForPrefix(startKey));
+        try {
+          while (rows.hasNext()) {
+            KeyValue<byte[], byte[]> row = rows.next();
+            String rowKey = Bytes.toString(row.getKey());
+            if (!rowKey.equals(stateStoreKey)) {
```
chtyim: Better has a `byte[]` form of the `stateStoreKey` once and use `Bytes.compare`.

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
gokulavasan: Then we will not know what records to clean. Should we handle that case (ie, delete the key corresponding to just "adapterName") or should we just let Preconditions throw an exception?

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
gokulavasan: Created JIRA : https://issues.cask.co/browse/CDAP-2178

gokulavasan:
@chtyim Addressed comments. Please review again. Thanks!

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
chtyim: I think just use a Precondtions to check is ok.

cdap-app-templates/suppressions.xml
```diff
@@ -0,0 +1,42 @@
...
+  the License.
+  -->
+
+<!DOCTYPE suppressions PUBLIC
+        "-//Puppy Crawl//DTD Suppressions 1.1//EN"
+        "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+
+  <suppress checks="Javadoc.*" files=".*/src/test/java/.*" />
```
chtyim: Why need a separate checkstyle suppression?

gokulavasan: Oh didn't realize this got included. I will remove it. My bad.

chtyim:
LGTM

PR #2115, opened 2015-04-19
CDAP-2164 adding metrics to etl template. The template will

albertshau:
emit records.in and records.out for each stage of the pipeline.
In addition, each context has a way to get Metrics for that stage,
which can be used by plugins to emit their own metrics.


cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
```diff
@@ -98,7 +100,7 @@ private void prepareSink(MapReduceContext context, ETLStage sinkStage) throws Ex
     StageSpecification sinkSpec = GSON.fromJson(
       context.getRuntimeArguments().get(Constants.Sink.SPECIFICATION), StageSpecification.class);
     BatchSink sink = (BatchSink) Class.forName(sinkSpec.getClassName()).newInstance();
-    BatchSinkContext sinkContext = new MapReduceSinkContext(context, sinkStage, sinkSpec);
+    BatchSinkContext sinkContext = new MapReduceSinkContext(context, sinkStage, sinkSpec, null);
```
gokulavasan: Why do we pass in null for metrics for Source and Sink?

albertshau: mistake, will pass in a real one

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/MapReduceBatchContext.java
```diff
@@ -19,6 +19,7 @@
 import co.cask.cdap.api.data.DatasetInstantiationException;
 import co.cask.cdap.api.dataset.Dataset;
 import co.cask.cdap.api.mapreduce.MapReduceContext;
+import co.cask.cdap.api.metrics.Metrics;
```
gokulavasan: required?

albertshau: nope, will remove

gokulavasan:
Changes look good. But my concern is wrt to the expectations of Metrics behavior from users pt of view wrt to the behavior seen in CDAP Apps - where they can query metrics that are available at app-level, program-level etc and it will just return the metric names emitted as provided by the user. And then the user can query at any level they want and metrics with same name from different programs are aggregated. This PR is similar to what we had in the old metrics system?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/MapReduceBatchContext.java
albertshau: and now its required

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
```diff
@@ -87,7 +90,8 @@ private void prepareSource(MapReduceContext context, ETLStage sourceStage) throw
     StageSpecification sourceSpec = GSON.fromJson(
       context.getRuntimeArguments().get(Constants.Source.SPECIFICATION), StageSpecification.class);
     BatchSource source = (BatchSource) Class.forName(sourceSpec.getClassName()).newInstance();
-    BatchSourceContext sourceContext = new MapReduceSourceContext(context, sourceStage, sourceSpec);
+    BatchSourceContext sourceContext = new MapReduceSourceContext(context, sourceStage, sourceSpec,
+      new StageMetrics(mrMetrics, StageMetrics.Type.SOURCE, sourceSpec.getName()));
```
gokulavasan: We should pass in mrMetrics instead of StageMetrics?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
```diff
@@ -98,7 +102,8 @@ private void prepareSink(MapReduceContext context, ETLStage sinkStage) throws Ex
     StageSpecification sinkSpec = GSON.fromJson(
       context.getRuntimeArguments().get(Constants.Sink.SPECIFICATION), StageSpecification.class);
     BatchSink sink = (BatchSink) Class.forName(sinkSpec.getClassName()).newInstance();
-    BatchSinkContext sinkContext = new MapReduceSinkContext(context, sinkStage, sinkSpec);
+    BatchSinkContext sinkContext = new MapReduceSinkContext(context, sinkStage, sinkSpec,
+      new StageMetrics(mrMetrics, StageMetrics.Type.SINK, sinkSpec.getName()));
```
gokulavasan: Same here?

gokulavasan:

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
albertshau: yep good catch

PR #2116, opened 2015-04-19
CDAP-2176 Fix logic to get the Config Type from App Template class

gokulavasan:
Existing logic fails if we have a class' whose super class is not ApplicationTemplate. Using TypeToken's resolveType to fix it. Added a template class which satisfies that criteria to make sure we are able to create an adapter from that template.

Bamboo : http://builds.cask.co/browse/CDAP-DUT1519
JIRA : https://issues.cask.co/browse/CDAP-2176


chtyim:
LGTM.

PR #2117, opened 2015-04-19
Variable Frequency Polling from FE on WebSocket. 

nitinmotgi:
Creating PR for @anwar6953 

One can now specify an interval from FE to the backend NodeJS server the frequency with which the nodejs server can poll. 

Added a new node-uuid to generate a unique request/response ID for the FE. Added the node-uuid to bower.json & Gulpfile. Also, included the same in package.json. 

*  [TESTED] Chrome
*  [TESTED] Safari
*  [TESTED] FireFox

### DIFF
--- a/cdap-ui/app/services/data/datasource.js
+++ b/cdap-ui/app/services/data/datasource.js
@@ -1,5 +1,8 @@
 angular.module(PKG.name+'.services')
-  /*
+
+  /**
+    Example Usage:
+
     MyDataSource // usage in a controler:
 
     var dataSrc = new MyDataSource($scope);
@@ -7,7 +10,8 @@ angular.module(PKG.name+'.services')
     // polling a namespaced resource example:
     dataSrc.poll({
         method: 'GET',
-        _cdapNsPath: '/foo/bar'
+        _cdapNsPath: '/foo/bar',
+        interval: 5000 // in milliseconds.
       },
       function(result) {
         $scope.foo = result;
@@ -26,20 +30,26 @@ angular.module(PKG.name+'.services')
         $scope.foo = result;
       }
     ); // will post to <host>:<port>/v3/system/config
+
    */
+
   .factory('MyDataSource', function ($state, $log, $rootScope, caskWindowManager, mySocket,
     MYSOCKET_EVENT, $q, $filter) {
 
     var instances = {}; // keyed by scopeid
-    var counter = 0;
-    var generateId = function() {
-    //TODO: revisit
-      // Very fundamental (naive) unique id. Wouldn't
-      // work on a cluster environment if two browser sessions
-      // access the same metric from different computers
-      return Date.now() + '-' + counter++;
+
+    /**
+     * Generates unique id's for each request that is being sent on
+     * the websocket connection. 
+     */
+    var generateUUID = function() {
+      return uuid.v4();
     }
 
+    /**
+     * Start polling of the resource - sends the action 'poll-start' to 
+     * the node backend. 
+     */
     function _pollStart (resource) {
       mySocket.send({
         action: 'poll-start',
@@ -47,6 +57,10 @@ angular.module(PKG.name+'.services')
       });
     }
 
+    /**
+     * Stops polling of the resource - sends the actions 'poll-stop' to
+     * the node backend. 
+     */
     function _pollStop (resource) {
       mySocket.send({
         action: 'poll-stop',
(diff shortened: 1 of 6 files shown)


cdap-ui/app/services/data/datasource.js
```diff
@@ -80,7 +88,7 @@ angular.module(PKG.name+'.services')
           return; // errors are handled at $rootScope level
         }
         angular.forEach(self.bindings, function (b) {
-          if(angular.equals(b.resource, data.resource)) {
+          if(b.resource.id == data.resource.id) {
```
elfenheart: use '===' instead of just '=='

nitinmotgi: done!

cdap-ui/app/services/data/datasource.js
```diff
@@ -148,27 +167,24 @@ angular.module(PKG.name+'.services')
...
+     * Fetch a resource on-demand. Send the action 'request' to 
+     * the node backend.
      */
     DataSource.prototype.request = function (resource, cb) {
-      var once = false,
-          deferred = $q.defer();
+      var deferred = $q.defer();
 
+      var id = generateUUID();
+      resource.id = id;
```
ajainarayanan: We don't need the id here right?

cdap-ui/server/aggregator.js
```diff
@@ -43,59 +49,104 @@ Aggregator.prototype.log = function () {
...
+Aggregator.prototype.startPolling = function (resource) {
+  // WARN: This assumes that the browser side ids are unique for a websocket session.
+  // This check is needed for Safari.
+  if(this.polledResources[resource.id]) {
+    // console.log("Resource id " + resource.id + " already registered.");
+    return;
+  }
+  resource.interval = resource.interval || POLL_INTERVAL;
+  this.polledResources[resource.id] = resource;
+  _.bind(doPoll, this, resource)();
```
ajainarayanan: All usages of ```_``` (underscore) library usage could replaced by native javascript usage 

```
doPoll.bind(this, resource)();
```

Javascript has ```.bind``` syntax that calls a function with a specific context and with specific set of arguments.

cdap-ui/server/aggregator.js
```diff
@@ -43,59 +49,104 @@ Aggregator.prototype.log = function () {
...
-Aggregator.prototype.planPolling = function () {
-  this.timeout = setTimeout(_.bind(doPoll, this), POLL_INTERVAL);
+Aggregator.prototype.scheduleAnotherIteration = function (resource) {
+  if (resource.stop) {
+    // Don't reschedule another iteration if the resource has been stopped
+    return;
+  }
+  // console.log(Object.keys(this.polledResources).length + ':' + Math.floor(Date.now()/1000) +
+  //    ': scheduling for: ' + resource.id + ', interval: ' + resource.interval + ' - ' + resource.url);
+  resource.timerId = setTimeout(_.bind(doPoll, this, resource), resource.interval);
```
ajainarayanan: Not sure why we need this timer and need to clear out this timer.

We are not using this to check for any condition and clearing out this time, say before receiving the response, doesn't change anything for us. Do we really need to do this?

cdap-ui/app/services/data/datasource.js
anwar6953: We use it to determine which callback to execute when the response comes from the node server (line 93)

cdap-ui/server/aggregator.js
anwar6953: It's not to avoid sending a response for a resource which has stopped polling.
Rather, its to avoid scheduling another request.

cdap-ui/server/aggregator.js
```diff
@@ -2,11 +2,16 @@
 
 var _ = require('lodash'),
     request = require('request'),
-    colors = require('colors/safe'),
-    HashTable = require('object-hash').HashTable;
```
anwar6953: remove line 56 of package.json also
`    "object-hash": "^0.5.0",`

nitinmotgi: Done!

cdap-ui/server/aggregator.js
nitinmotgi: Done!

cdap-ui/server/aggregator.js
nitinmotgi: +1

cdap-ui/server/aggregator.js
anwar6953: @ajainarayanan Do you know why _.bind and other _. functions were being used, when native versions exist in javascript?

anwar6953: Might be [performance](https://jsperf.com/bind-vs-jquery-proxy/68)
2x speed difference. Not a big deal, due to the rate at which we call them.

ajainarayanan:
:+1: LGTM Awesome!

PR #2119, opened 2015-04-19
Add a single-page version of HTTP RESTful API

awholegunch:
These changes modify the documentation for the RESTful API such that a single-page version with all the APIs on one page can be created. There have been a number of requests for a page like this. It is generated by "including" together all the other pages, and so it is automatically updating with any changes in the other files.

The advantage with it is that you can search on this single page across all APIs.

It is staged as [Single Page](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_Single_Page_RESTful_API/en/reference-manual/http-restful-api/single-page.html).

Please take a look and comment on if you think this is useful addition—or not.

Caveats:
- The metrics section of the single page is not formatted correctly, because the underlying document has a different structure than the other RESTful APIs. As that file is currently being worked on in another PR, I will fix it there so that problem will disappear.
- Numerous warnings are created in the build because duplicate references are created. However, the links all seem to work fine, so we have to just put up with these warnings, unless I figure out how to get around them.

### DIFF
--- a/cdap-docs/reference-manual/source/http-restful-api/dataset.rst
+++ b/cdap-docs/reference-manual/source/http-restful-api/dataset.rst
@@ -3,13 +3,14 @@
     :description: HTTP RESTful Interface to the Cask Data Application Platform
     :copyright: Copyright © 2014 Cask Data, Inc.
 
+
+.. highlight:: console
+
 .. _http-restful-api-dataset:
 
 ===========================================================
 Dataset HTTP RESTful API
 ===========================================================
-
-.. highlight:: console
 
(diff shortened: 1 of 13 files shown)


sreevatsanraman:
I don't think it is useful. Having a multi-page and a single page is very confusing.

sreevatsanraman:
Let's close this PR. Singlepage rest docs are not useful. Useful material here would be a cheat sheet with all the endpoints.

awholegunch:
Closed per request. Moving to internal use only.

PR #2121, opened 2015-04-19
(CDAP-2177) Fix standalone log appender

chtyim:
It needs to override the lifecycle methods and the append method to correctly delegate calls from logback into underlying log appender.


sreevatsanraman:
LGTM :+1: 

Will merge once the build completes. http://builds.cask.co/browse/CDAP-DUT1523-1

chtyim:
Build passed. Merging.

PR #2122, opened 2015-04-19
Feature/ui truncate v2 api

elfenheart:
Change V2 APIs to V3


ajainarayanan:
:+1: LGTM

anwar6953:
This seems very similar to what we have in widget.js (the time series controller).
Should this be reusing the same code there?

Either way, this doesn't seem like it would work, because it is not interpolating the metric values with zeros. (I'm not sure what the chart would look like).

elfenheart:
Changing operation28 to dashboard in separate PR

anwar6953:
Ok, using widgets (the whole operation28 doesn't need to be a dashboard though) would help centralize the logic.

PR #2123, opened 2015-04-19
[CDAP-2182] fix routes for system service "streams"

anew:
The stream handler rule (which only checks whether the 4th component of the URL path is "streams") was firing on these URLs (/v3/system/services/streams/[logs|live-info|...]. Adding a test for 2nd component to be "namespaces" fixes that.


bdmogal:
:+1:

PR #2124, opened 2015-04-19
Replace MetaClient.unrecoverableReset() with NamespaceClient.deleteAll()

anew:
This is used in test cases to reset CDAP after a test. The old method only reset the default namespace. But because tests can use multiple namespaces, we should delete all namespaces. This belongs into NamespaceClient, not MetaClient. Also adds a unit test and removes the old method.


sreevatsanraman:
LGTM :+1:

PR #2125, opened 2015-04-19
Removing the configurations that are no more used.

nitinmotgi:
Removed the remnant configuration from accounts when we had sandbox.


sreevatsanraman:
LGTM :+1:

PR #2126, opened 2015-04-19
(CDAP-1917) Added plugin repository and plugin instantiate support.

chtyim:
It is the ground work for the whole plugin system for application template.

This PR includes works on the following pieces:
- New annotations defined for annotating plugin class and plugin config properties
- New data structure classes for recording plugin information and plugin class information.
- A `PluginClassLoader` for defining the class loading behavior of plugin classes
- A `PluginRepository` for inspecting and caching plugin information
- A `PluginInstantiator` for creating new instances of plugin classes

Template plugins are deployed as JAR files under the template plugin directory (i.e. under "data/plugins/[template_name]").

There are two ways of defining plugin classes in a plugin jar.

  1. Annotate class with the `@Plugin` annotation, and make the package name of the class available under the "Export-Package" jar manifest field.
    - The syntax for the "Export-Package" is the same as defined in OSGI, but only the package name part is being used. All other attributes are currently ignored
    - Plugin class can have a field with type extended from `PluginConfig` class. If such field exists, it will be inspected and information will be extracted to form the set of plugin properties being supported by that plugin class
    - At runtime, the config field of the plugin class will be injected with an instance with fields populated with plugin configurations provided by the user.
  2. Use an external json file to declare what classes are available as plugin class. In this mode, no class inspecting and config field field will happen.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginClass.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginClass.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.templates.plugins;
+
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Contains information about a plugin class.
+ */
+public class PluginClass {
+
+  private final String type;
+  private final String name;
+  private final String description;
+  private final String className;
+  private final String configFieldName;
+  private final Map<String, PluginPropertyField> properties;
+
+  public PluginClass(String type, String name, String description, String className,
+                     @Nullable String configfieldName, Map<String, PluginPropertyField> properties) {
+    if (type == null) {
+      throw new IllegalArgumentException("Plugin class type cannot be null");
+    }
+    if (name == null) {
+      throw new IllegalArgumentException("Plugin class name cannot be null");
+    }
+    if (description == null) {
+      throw new IllegalArgumentException("Plugin class description cannot be null");
+    }
+    if (className == null) {
+      throw new IllegalArgumentException("Plugin class className cannot be null");
+    }
+    if (properties == null) {
+      throw new IllegalArgumentException("Plugin class properties cannot be null");
+    }
+
+    this.type = type;
+    this.name = name;
+    this.description = description;
+    this.className = className;
+    this.configFieldName = configfieldName;
+    this.properties = properties;
+  }
+
+  /**
+   * Returns the type name of the plugin.
+   */
+  public String getType() {
+    return type;
+  }
+
+  /**
+   * Returns name of the plugin.
+   */
+  public String getName() {
+    return name;
+  }
+
+  /**
+   * Returns description of the plugin.
+   */
+  public String getDescription() {
+    return description;
+  }
+
+  /**
+   * Returns the fully qualified class name of the plugin.
+   */
+  public String getClassName() {
+    return className;
+  }
+
+  /**
+   * Returns the name of the field that extends from {@link PluginConfig} in the plugin class.
+   * If no such field, {@code null} will be returned.
+   */
+  @Nullable
+  public String getConfigFieldName() {
+    return configFieldName;
+  }
+
+  /**
+   * Returns a map from config property name to {@link PluginPropertyField} that are supported by the plugin class.
+   */
+  public Map<String, PluginPropertyField> getProperties() {
+    return properties;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    PluginClass that = (PluginClass) o;
+
+    return type.equals(that.type)
+      && name.equals(that.name)
+      && description.equals(that.description)
+      && className.equals(that.className)
+      && !(configFieldName != null ? !configFieldName.equals(that.configFieldName) : that.configFieldName != null)
+      && properties.equals(that.properties);
+  }
+
+  @Override
+  public int hashCode() {
+    int result = type.hashCode();
+    result = 31 * result + name.hashCode();
+    result = 31 * result + description.hashCode();
+    result = 31 * result + className.hashCode();
+    result = 31 * result + (configFieldName != null ? configFieldName.hashCode() : 0);
+    result = 31 * result + properties.hashCode();
+    return result;
+  }
+}
--- a/cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginInfo.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginInfo.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.templates.plugins;
+
+/**
+ * Contains plugin information.
+ */
+public final class PluginInfo implements Comparable<PluginInfo> {
+
+  private final String fileName;
+  private final String name;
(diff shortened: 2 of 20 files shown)


anew:
Nice description there - does it make sense to add that to the Jira, too? Also should it be documented somewhere in the docs?

I will start reading the code now.

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginClass.java
```diff
@@ -0,0 +1,117 @@
...
+      && className.equals(that.className)
+      && !(configFieldName != null ? !configFieldName.equals(that.configFieldName) : that.configFieldName != null)
+      && properties.equals(that.properties);
+  }
+
+  @Override
+  public int hashCode() {
+    int result = type.hashCode();
+    result = 31 * result + name.hashCode();
+    result = 31 * result + description.hashCode();
```
anew: none of the fields can ever be null? Description, for example? If they are not allowed to be null, should you check for that in the constructor?

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginInfo.java
```diff
@@ -0,0 +1,91 @@
...
+public final class PluginInfo implements Comparable<PluginInfo> {
+
+  private final String fileName;
+  private final String name;
+  private final PluginVersion version;
+
+  public PluginInfo(String fileName, String name, PluginVersion version) {
+    this.fileName = fileName;
+    this.name = name;
+    this.version = version;
```
anew: Similar here, does it make sense to protect against null values?

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginProperties.java
```diff
@@ -0,0 +1,83 @@
...
+package co.cask.cdap.api.templates.plugins;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Plugin instance properties.
+ */
+public class PluginProperties {
```
anew: naming of this class is a little misleading. It suggests that it is a collection or map of PluginProperty. But that it is only strings, whereas PluginProperty is a different thing. It is also not clear right away, what purpose this serves.

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginProperty.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+public class PluginProperty {
+
+  private final String name;
+  private final String description;
+  private final String type;
+  private final boolean required;
+
+  public PluginProperty(String name, String description, String type, boolean required) {
+    this.name = name;
```
anew: check against null values?

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginClass.java
chtyim: It is not supposed to be null. In fact the constructor is not supposed to be called by user, but by the plugin system only (which will make sure all values are correct). However, since the class needs to be in api, there is no good way to protect the constructor. I can add null check in the constructor.

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginProperties.java
chtyim: I name this class based on `DatasetProperties`. Probably better to rename the `PluginProperty` class to something else. Any suggestion?

cdap-app-fabric/src/main/java/co/cask/cdap/app/program/ManifestFields.java
```diff
@@ -33,6 +40,44 @@
...
+  public static final Attributes.Name EXPORT_PACKAGE = new Attributes.Name("Export-Package");
+
   public static final String VERSION = "1.0";   // Defines manifest version value.
   public static final String MANIFEST_SPEC_FILE = "META-INF/specification/application.json";
+
+  /**
+   * Regex for extracting one package name from the Export-Package manifest entry.
+   * Refer to OSGI spec r4v43, section 3.6.5 for details.
+   */
+  private static final Pattern EXPORT_PACKAGE_PATTERN =
```
anew: maybe add one or tow examples of valid inputs to illustrate this?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginFile.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package co.cask.cdap.internal.app.runtime.adapter;
+
+import co.cask.cdap.api.templates.plugins.PluginInfo;
+
+import java.io.File;
+
+/**
+ * A internal container for holding plugin {@link File} and the {@link PluginInfo} together.
```
anew: An internal

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginRepository.java
```diff
@@ -0,0 +1,479 @@
...
+
+      String pluginName = rawVersion == null ? plugin : plugin.substring(0, plugin.length() - rawVersion.length() - 1);
+      return new PluginFile(file, new PluginInfo(file.getName(), pluginName, version));
+    }
+  };
+
+  private final CConfiguration cConf;
+  private final File templateDir;
+  private final File tmpDir;
+  private final AtomicReference<Map<String, Multimap<PluginInfo, PluginClass>>> plugins;
```
anew: A map from each adapter name to its plugins? Why a multi-map?

chtyim: Map from template name to all plugins available. The `PlugInfo` is basically information about the plugin file (e.g. transform-1.0.jar). Inside a jar, there can be multiple plugin classes (e.g. Identity, Projection, etc), hence a multimap.

anew: Oh, one plugin can have multiple plugin classes...

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginRepository.java
```diff
@@ -0,0 +1,479 @@
...
+      return templatePlugins;
+    }
+
+    Iterable<PluginFile> pluginFiles = Iterables.transform(pluginJars, FILE_TO_PLUGIN_FILE);
+    CloseableClassLoader templateClassLoader = createTemplateClassLoader(templateJar);
+    try {
+      PluginInstantiator pluginInstantiator = new PluginInstantiator(cConf, template, templateClassLoader);
+      try {
+        for (PluginFile pluginFile : pluginFiles) {
+          if (!configureByFile(pluginFile, templatePlugins)) {
```
anew: So a plugin configured either by file or by inspection, right? It is not possible to add a file to add additional properties for plugin that is already configured by inspection. @nitinmotgi please confirm that you agree with that

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginClassLoader.java
```diff
@@ -0,0 +1,97 @@
...
+ *                  |
+ *           Plugin Lib CL (Common library for plugin)
+ * }</pre>
+ *
+ * <p/>
+ * Then, the parent of the PluginClassLoader is formed by a {@link CombineClassLoader} of
+ * {@code (Program Filter CL, Plugin Lib ClassLoader) }. It is a combine class loader because we
+ * want the cdap-api classes not affected by the "Template Filter ClassLoader" (which usually would have cdap-api
+ * classes filtered out).
+ *
```
anew: Nice documentation, this makes it really clear. Are there any test cases for this class loader?

anew:
Overall very clean code, nice job. I have only few comments. Major comment is that there are no tests for the plugin class loader.

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginClass.java
chtyim: check added

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginInfo.java
chtyim: check added.

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginProperty.java
chtyim: check added.

cdap-app-fabric/src/main/java/co/cask/cdap/app/program/ManifestFields.java
chtyim: Added more javadoc for that.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginClassLoader.java
chtyim: Added https://issues.cask.co/browse/CDAP-2190 for the task.

chtyim:
The description is added to the corresponding JIRA as well.

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginProperties.java
chtyim: Renamed `PluginProperty` to `PluginPropertyField`.

anew:
LGTM

chtyim:
Squashed commits to prepare for merge.

chtyim:
build passed: http://builds.cask.co/browse/CDAP-DUT1527-6

PR #2128, opened 2015-04-19
Fix/ui etlapps create makeup

ajainarayanan:
- Redesign part of the etl apps create view 
- Adds default place holders for source, transforms & sinks.

TODO:
Code cleanup
UI Cleanup
Sortable

### DIFF
--- a/cdap-ui/app/features/etlapps/controllers/create-ctrl.js
+++ b/cdap-ui/app/features/etlapps/controllers/create-ctrl.js
@@ -40,34 +40,29 @@ angular.module(PKG.name + '.feature.etlapps')
 
     // Source, Sink and Transform Models
     $scope.source = {
-      name: 'Drag & Drop A Source',
+      name: 'Add a Source',
       properties: {},
       placeHolderSource: true
     };
 
     $scope.sink = {
-      name: 'Drag & Drop A Transforms',
-      placeHolderTransform: true,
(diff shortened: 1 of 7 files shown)


cdap-ui/app/features/etlapps/controllers/create-ctrl.js
```diff
@@ -1,8 +1,11 @@
 angular.module(PKG.name + '.feature.etlapps')
-  .controller('ETLAppsCreateController', function($scope, MyDataSource, $alert, $bootstrapModal, $state, ETLAppsApiFactory) {
+  .controller('ETLAppsCreateController', function($scope, MyDataSource, $alert, $bootstrapModal, $state, ETLAppsApiFactory, $filter) {
     var apiFactory = new ETLAppsApiFactory($scope);
 
     $scope.ETLMetadataTabOpen = true;
+    $scope.ETLSourcesTabOpen = true;
+    $scope.ETLTransformsTabOpen = true;
+    $scope.ETLSinksTabOpen = true;
```
elfenheart: If it is always open, do we actually need to track it?

ajainarayanan: We need to be able to close them. Its just the first time we need to keep it open.

elfenheart:
:+1: LGTM!

PR #2129, opened 2015-04-19
Making sure that we don't propogate connection errors into alerts

nitinmotgi:
Currently, issues related to NodeJS not being able to communicate with CDAP are being propagated to user. This information is useless as if the NodeJS is not able to connect to CDAP, then we block the screen and show a pop-up. But, adding alerts for every time we try connecting with CDAP makes no sense. This PR includes the fix for the same.

It makes sure, that it will not push the response only if it's not a connect call failure.


ajainarayanan:
:+1: LGTM!

PR #2130, opened 2015-04-19
name tests correctly so they actually get run by maven,

albertshau:
and make sure api version and namespace are always specified in
AppFabricTestBase when deploying an app.


chtyim:
LGTM

albertshau:
http://builds.cask.co/browse/CDAP-DUT1530

PR #2131, opened 2015-04-19
Modified the error messages to reflect the correct state of the backend ...

nitinmotgi:
...to the user and propose right actions


elfenheart:
:+1: LGTM

PR #2133, opened 2015-04-19
CDAP-2184 document that throwing an IllegalArgumentException

albertshau:
when configuring an adapter will tell the framework that the
config was invalid. This will make the rest api return a 400
instead of a 500. Also making sure the exception is not lost
wrapped in layers of ExecutionExceptions.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationTemplateManager.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationTemplateManager.java
@@ -112,7 +112,7 @@ public class LocalApplicationTemplateManager implements Manager<DeploymentInfo, 
                                                     queueAdmin, discoveryServiceClient));
     pipeline.addLast(new ProgramGenerationStage(configuration, namespacedLocationFactory));
     pipeline.addLast(new ApplicationRegistrationStage(store));
-    pipeline.setFinally(new EnableConcurrentRunsStage(preferencesStore));
+    pipeline.addLast(new EnableConcurrentRunsStage(preferencesStore));
     return pipeline.execute(input);
   }
 }
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterLifecycleTests.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterLifecycleTests.java
@@ -181,7 +181,8 @@ public class AdapterLifecycleTests extends AppFabricTestBase {
   @Test
   public void testRestrictUserApps() throws Exception {
     // Testing that users can not deploy an application
-    HttpResponse response = deploy(AppWithServices.class, DummyBatchTemplate.NAME);
+    HttpResponse response = deploy(AppWithServices.class, Constants.Gateway.API_VERSION_3_TOKEN,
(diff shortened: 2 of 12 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT1534

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterLifecycleTests.java
```diff
@@ -214,7 +215,17 @@ public void testNoTemplateFieldReturns400() throws Exception {
...
+    DummyBatchTemplate.Config config = new DummyBatchTemplate.Config("somesource", null);
+    AdapterConfig adapterConfig = new AdapterConfig("description", DummyBatchTemplate.NAME, GSON.toJsonTree(config));
+    HttpResponse response = doPut(
+      String.format("%s/namespaces/%s/adapters/%s",
+        Constants.Gateway.API_VERSION_3, Constants.DEFAULT_NAMESPACE, adapterName), GSON.toJson(adapterConfig));
+    Assert.assertEquals(400, response.getStatusLine().getStatusCode());
+
+    String deleteURL = getVersionedAPIPath("apps/" + DummyBatchTemplate.NAME,
+      Constants.Gateway.API_VERSION_3_TOKEN, Constants.DEFAULT_NAMESPACE);
+    deleteApplication(60, deleteURL, 200);
```
gokulavasan: So the adapter did not get created. But the app got deployed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationTemplateManager.java
```diff
@@ -112,7 +112,7 @@ public LocalApplicationTemplateManager(CConfiguration configuration, PipelineFac
                                                     queueAdmin, discoveryServiceClient));
     pipeline.addLast(new ProgramGenerationStage(configuration, namespacedLocationFactory));
     pipeline.addLast(new ApplicationRegistrationStage(store));
-    pipeline.setFinally(new EnableConcurrentRunsStage(preferencesStore));
+    pipeline.addLast(new EnableConcurrentRunsStage(preferencesStore));
```
gokulavasan: So we don't want the last stage to be executed if there is an error thrown during the any of the preceding stages. Is that the reasoning behind this change?

gokulavasan:
One comment about adapter creation/app deployment. Rest LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/LocalApplicationTemplateManager.java
albertshau: sound out setFinally is like the finally block in a try catch statement, which was used to do pipeline cleanup. But some time ago I moved cleanup into the stage that causes the mess, but didn't change the setFinally

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/AdapterLifecycleTests.java
albertshau: yup. template gets deployed before the adapter. ideally the entire thing would be transactional, but that is not the case

PR #2134, opened 2015-04-19
Feature/ui move flowlet tab

elfenheart:
Moving Flowlet detail page to be under Runs


cdap-ui/app/features/flows/controllers/tabs/runs/runs-detail-ctrl.js
```diff
@@ -16,7 +20,11 @@ angular.module(PKG.name + '.feature.flows')
       event.targetScope.runs.selected = null;
     });
 
-    $scope.selectTab = function(tab) {
+    $scope.selectTab = function(tab, node) {
+      if (tab.title === 'Flowlets') {
```
ajainarayanan: This function should probably be ```goToFlowlet(flowletId)``` right? What we are actually doing is going to the flowlet detailed view but the function name suggests we are just switching the tab. The name kind of mis-leads here.. Thoughts?

ajainarayanan:
:+1: LGTM
Talked to @elfenheart We should re-visit the programs code and make the structure consistent so that its not too confusing for new developers + we need to make it better to interact between a parent & a child controller by introducing services/factories that can be shared.

cdap-ui/app/features/flows/controllers/tabs/runs/runs-detail-ctrl.js
elfenheart: This is switching tabs. If we are switching to Flowlets, then we add the flowlet node name. Otherwise it can go to Logs and Status tab as well

PR #2135, opened 2015-04-20
Aggregate metrics (pie chart)

anwar6953:
Pie chart (refreshes and redraws over time).
The slice values are not currently displayed on the charts.
We also need better labelling capabilities (tooltip the metric name upon hover?)

Metrics that have values (all metrics are zero are not render-able in a pie chart).

Otherwise, it looks like this:

Table and donut:


cdap-ui/app/directives/epoch/charts.js
```diff
@@ -94,7 +101,7 @@ ngEpoch.controller('epochController', function ($scope, $compile, caskWindowMana
 ngEpoch.directive('epochPie', function () {
   return angular.extend({
     link: function (scope, elem, attr) {
-      scope.initEpoch(elem, 'pie', attr);
+      scope.initEpoch(elem, 'pie', attr, {updateFull: true});
```
anwar6953: For pie chart, each time we get new data, we update the entire chart (no notion of sending in a stream).

nitinmotgi: +1

cdap-ui/app/directives/epoch/charts.js
```diff
@@ -52,6 +56,9 @@ ngEpoch.controller('epochController', function ($scope, $compile, caskWindowMana
       $scope.$watch('history', function (newVal) {
         if(newVal) {
           $scope.options.data = newVal;
+          if ($scope.options.updateFull) {
```
ajainarayanan: We could probably render the entire history again right? Why update an entire history and introduce a special flag for it?

anwar6953: For realtime-line, realtime-area and realtime-bar, we don't render the entire history each poll iteration. We just push into the stream.
Only if updateFull do we render the entire chart each poll (since nothing else makes sense for pie chart).

We can also set updateFull = true for other types of charts if we want though.

cdap-ui/app/directives/epoch/charts.js
ajainarayanan: Have added a comment above.

anwar6953:
Closed PR because this functionality has been implemented in other PR (via c3): https://github.com/caskdata/cdap/pull/2184

PR #2136, opened 2015-04-20
Adding log4js logger to be used instead of console.log

nitinmotgi:
*   Added log4js as dependency and removed colors and morgan
*   Changed all the backend logs to use log4js
*   Added a default configuration file conf/log4js.json that logs to console. 
*   Tested and made sure, the change in conf/log4js.json takes effect in 300 seconds. 
*   Added backend response times to aggregator to measure backend response times.


cdap-ui/conf/log4js.json
```diff
@@ -0,0 +1,11 @@
+{
+  "appenders": [
+    { 
+      "category" : "default",
+      "type" : "console"
+    }
+  ],
+  "levels": {
+     "default" : "TRACE"
```
elfenheart: do we want the default to be TRACE or INFO?

elfenheart:
:+1: LGTM besides the default to TRACE

PR #2080, opened 2015-04-18
Feature/query multiple measures fact table

gsps1:
FactTable to support query multiple metrics with a single scan. This will allow cube and metriStore API's to query multiple measures under a single context with a single API call. 

JIRA : https://issues.cask.co/browse/CDAP-2124

PR #2138, opened 2015-04-20
[CDAP-2163] delete metrics directly through MetricStore instead of making a HTTP call to DeleteMetricsHandler

anew:
- injects MetricStore in places where we delete Metrics 
- deletes via MetricStore instead of making an HTTP call to MetricService
- removes DeleteMetricHandler


anew:
Build is running here: http://builds.cask.co/browse/CDAP-DUT1532-2

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1739,37 +1707,17 @@ private boolean canHaveInstances(ProgramType programType) {
...
+                                           String application,
+                                           String flow) throws Exception {
+
+    long endTs = System.currentTimeMillis() / 1000;
+    String metricNamePrefix = "process";
+    Map<String, String> tags = Maps.newHashMap();
+    tags.put(Constants.Metrics.Tag.NAMESPACE, namespaceId);
+    tags.put(Constants.Metrics.Tag.APP, application);
+    tags.put(Constants.Metrics.Tag.FLOW, flow);
+    MetricDeleteQuery deleteQuery = new MetricDeleteQuery(0, endTs, metricNamePrefix, tags);
```
gsps1: our metric system does not support query/delete by metrics prefix anymore, as the entire metrics string is encoded as one value and not split by dots and encoded separately as before. so this would not work. 

I have a PR https://github.com/caskdata/cdap/pull/2080 to query/delete multiple metrics in a context, but for now you can add a todo and do delete for each individual metrics separately.

anew: That means the old code was also broken? When we were calling the HTTP endpoint in the DeleteMetricsHandler? So how would we implement this then?

anew: So this is a questionable API anyway. It is to clear the queues for a flow. I guess it was introduced to allow people to clear the queues if they had bad code (or due to an update of code) so that the queues had events that cannot be processed.

It is not clear to me why we have to delete queue metrics in this case. The metrics express how many events have been processed, and that does not change by emptying the queues. Of course, if we take the difference of enqueued minus processed to determine the size of the queue... that makes sense. 

So if we delete the metrics, do we have to delete every individual metric? Or can we just delete all metrics for the flow?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1014,21 +1008,9 @@ void getServiceInstances(HttpResponder responder,
...
-          responder.sendStatus(HttpResponseStatus.NOT_FOUND);
-          return;
-        }
-        instances = workerSpec.getInstances();
-      }
-
-      responder.sendJson(HttpResponseStatus.OK,
-                         new ServiceInstances(instances, getInstanceCount(namespaceId, appId, ProgramType.SERVICE,
-                                                                          serviceId, runnableName)));
+      int instances = specification.getInstances();
```
abaranau: so no worker anymore? or what was the deleted code about?

do you still need comment?

anew: there is no service worker support in v3 APIs.

anew:
@abaranau What I wanted your opinion on is the metrics change. You suggested this change. But according to @gsps1 the functionality has been dropped from the metrics service. So what do you suggest we do about this? How can we delete all metrics for a flow that start with "process"?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
abaranau: Current API supports deleting all metrics or specified metric (or specified multiple metrics - after PR #2080 is merged). So, yes, no delete by prefix name.

Not sure if we need to delete queue metrics in this case either. Maybe it is not needed - hence, since it was not doing anything before your change (as delete by prefix is incorrect) - was not spotted as issue..

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/DeletedProgramHandlerStage.java
```diff
@@ -125,39 +112,16 @@ public void process(ApplicationDeployable appSpec) throws Exception {
...
-        Throwables.propagate(e);
-      } finally {
-        client.close();
-      }
+      long endTs = System.currentTimeMillis() / 1000;
+      Map<String, String> tags = Maps.newHashMap();
+      tags.put(Constants.Metrics.Tag.NAMESPACE, namespace);
+      tags.put(Constants.Metrics.Tag.APP, application);
+      tags.put(Constants.Metrics.Tag.FLOW, flow);
+      MetricDeleteQuery deleteQuery = new MetricDeleteQuery(0, endTs, null, tags);
```
abaranau: seems like we should have ctor without metric name param, if null is allowed - would make this code pretty. Nothing to do in this PR, just noting - for future

abaranau:
> How can we delete all metrics for a flow that start with "process"?

see my reply. Currently, you have to list metrics you want to delete. If you need that weird delete - not sure about the logic there.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anew: OK, I will test this tomorrow and figure out what the effect is. 

By the way, when did this feature get removed from metrics? It is surprising that this code (which was only ported to v3 recently) is still using a feature that was removed...

abaranau:
other than that one piece in question, other stuff LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
awholegunch: Does this affect documentation? Do we have an issue tracking this: both the deletion and the "there is no service worker support in v3 APIs."?

anew:
I removed the dysfunctional code in question, added a TODO and jIra CDAP-2191. Will merge now since everything else was LGTM.

PR #2139, opened 2015-04-20
CDAP-2145 moving stream transform into the stream source.

albertshau:
While at it, doing some cleanup. The source now takes an offset
as well as a duration, and it goes through CDAP for stream formats
which means we dont need the copied formats anymore.


gokulavasan:
LGTM :+1:

gokulavasan:
LGTM :+1:

PR #2140, opened 2015-04-20
Removed the widget footer

nitinmotgi:
Widget footer that shows the metrics in graph is removed, but will be added in the edit window.


elfenheart:
:+1: LGTM

PR #2141, opened 2015-04-20
CDAP-2185 extend the get all rest api for adapters to support an

albertshau:
optional 'template' query param that will filter the adapters
returned by those that match the template given.


gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-DUT1538

PR #2142, opened 2015-04-20
Make sure that the ETLRealtimeTemplate gets a non-zero value for instances

gokulavasan:
Bamboo : http://builds.cask.co/browse/CDAP-DT35

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
@@ -27,6 +27,7 @@ import co.cask.cdap.templates.etl.realtime.config.ETLRealtimeConfig;
 import co.cask.cdap.templates.etl.realtime.sinks.NoOpSink;
 import co.cask.cdap.templates.etl.realtime.sources.TestSource;
 import co.cask.cdap.templates.etl.transforms.IdentityTransform;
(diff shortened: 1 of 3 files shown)


cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
```diff
@@ -50,6 +51,7 @@ public ETLRealtimeTemplate() throws Exception {
   public void configureAdapter(String adapterName, ETLRealtimeConfig etlConfig, AdapterConfigurer configurer)
     throws Exception {
     super.configureAdapter(adapterName, etlConfig, configurer);
+    Preconditions.checkArgument(etlConfig.getInstances() != 0, "Instances cannot be 0");
```
bdmogal: By this time, has default 0 has been applied right (scenario - its unspecified in configuration). Otherwise you may have to check for null?

bdmogal:
1 question, otherwise LGTM

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLRealtimeTemplate.java
gokulavasan: etlConfig.getInstances returns a int. So when it is not specified in the configuration, I think it will be initialized with 0.

henry-cask:
+1

LGTM

PR #2145, opened 2015-04-20
[CDAP-2062] Added real-time Stream Sink

bdmogal:
* Adds a real-time Stream sink
* PR looks large because of the two following changes in the CDAP platform:
1. ``BasicWorkerContext`` used to ``new DefaultStreamWriter()`` earlier. Which meant that it always depended on the Stream Service to be up and running, even in tests. In ``TestBase``, however, we do not start Stream Service, so all calls to write to streams failed because Stream Service could not be discovered. Fixed this by adding a Guice Factory for ``StreamWriter``, that accepts an ``@Assisted String namespaceId``. ``StreamWriterFactory`` is now injected into ``WorkerProgramRunner`` and made available to ``BasicWorkerContext`` to create a ``DefaultStreamWriter`` for the required namespace.
2. In ``TestBase``, there was no way to read from a stream. Introduced a ``StreamManager`` and a Guice factory ``StreamManagerFactory`` to be able to (for now) create a stream and read events from it using the ``TestBase`` classes. Eventually, as part of [CDAP-2171](https://issues.cask.co/browse/CDAP-2171), we should deprecate the existing ``StreamWriter`` and move all its functionality to ``StreamManager``, as also move streams outside of ``ApplicationManager``. However, this will come in a later PR.
* Another related change, made ``MockResponder`` update its contents on ``sendChunk`` calls.

- [x] End-to-end unit test from ``TestSource`` to ``StreamSink`` works
- [x] Tested the above scenario on Standalone as well
- [x] Cluster testing 

CDAP Build: http://builds.cask.co/browse/CDAP-DUT1516
Templates Build: http://builds.cask.co/browse/CDAP-DT32 (Compilation failure here because it needs a class in CDAP framework that was added by this PR. Investigating how to fix the build)
Jira: [CDAP-2062](https://issues.cask.co/browse/CDAP-2062)

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.templates.etl.realtime.sinks;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.data.stream.Stream;
+import co.cask.cdap.api.stream.StreamEventData;
+import co.cask.cdap.templates.etl.api.PipelineConfigurer;
+import co.cask.cdap.templates.etl.api.Property;
+import co.cask.cdap.templates.etl.api.StageConfigurer;
+import co.cask.cdap.templates.etl.api.config.ETLStage;
+import co.cask.cdap.templates.etl.api.realtime.DataWriter;
+import co.cask.cdap.templates.etl.api.realtime.RealtimeContext;
+import co.cask.cdap.templates.etl.api.realtime.RealtimeSink;
+import co.cask.cdap.templates.etl.common.Properties;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Map;
+
+/**
+ * Real-time sink for Streams
+ */
+public class StreamSink extends RealtimeSink<StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(StreamSink.class);
+  private String streamName;
+  private String headersField;
+  private String dataField;
+
+  @Override
+  public void configure(StageConfigurer configurer) {
+    configurer.setDescription("Real-time sink that outputs to the specified stream.");
+    configurer.addProperty(new Property(Properties.Stream.NAME,
+                                        "The name of the stream to output to. Must be a valid stream name. " +
+                                          "The stream will be created if it does not exist.",
+                                        true));
+    configurer.addProperty(new Property(Properties.Stream.DATA_FIELD,
+                                        "Name of the field in the record that contains the data to be written to " +
+                                          "the specified stream. The data could be in binary format as " +
+                                          "a byte array or a ByteBuffer. It can also be a String." +
+                                          "If unspecified, the 'data' key is used.",
+                                        false));
+    configurer.addProperty(new Property(Properties.Stream.HEADERS_FIELD,
+                                        "Name of the field in the record that contains headers. Headers are presumed" +
+                                          " to be a map of string to string.",
+                                        false));
+  }
+
+  @Override
+  public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineConfigurer) {
+    Map<String, String> properties = stageConfig.getProperties();
+    String streamName = properties.get(Properties.Stream.NAME);
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(streamName), "Stream name should be non-null, non-empty.");
+    pipelineConfigurer.addStream(new Stream(streamName));
+  }
+
+  @Override
+  public void initialize(RealtimeContext context) throws Exception {
+    super.initialize(context);
+    Map<String, String> runtimeArguments = context.getRuntimeArguments();
+    streamName = runtimeArguments.get(Properties.Stream.NAME);
+    headersField = runtimeArguments.containsKey(Properties.Stream.HEADERS_FIELD) ?
+      runtimeArguments.get(Properties.Stream.HEADERS_FIELD) : Properties.Stream.DEFAULT_HEADERS_FIELD;
+    dataField = runtimeArguments.containsKey(Properties.Stream.DATA_FIELD) ?
+      runtimeArguments.get(Properties.Stream.DATA_FIELD) : Properties.Stream.DEFAULT_DATA_FIELD;
+  }
+
+  @Override
+  public int write(Iterable<StructuredRecord> structuredRecords, DataWriter dataWriter) throws Exception {
+    int numRecordsWritten = 0;
+    for (StructuredRecord structuredRecord : structuredRecords) {
+      Schema schema = structuredRecord.getSchema();
+      Object data = structuredRecord.get(dataField);
+      Object headers = structuredRecord.get(headersField);
+      if (data == null) {
+        LOG.debug("Found null data. Skipping record.");
+        continue;
+      }
+
+      if (headers != null && !isHeadersSchemaPresentAndSupported(schema)) {
+        LOG.debug("Headers found in input, however either the headers schema is not provided or the provided " +
+                    "schema is not supported. Only a map of string keys and string values is supported. " +
+                    "Skipping record.");
+        continue;
+      }
+
+      Schema.Field dataSchemaField = schema.getField(dataField);
+      switch (dataSchemaField.getSchema().getType()) {
+        case BYTES:
+          numRecordsWritten += writeBytes(dataWriter, data, headers);
+          break;
+        case STRING:
+          numRecordsWritten += writeString(dataWriter, data, headers);
+          break;
+        default:
+          LOG.debug("Type {} is not supported for writing to stream", data.getClass().getName());
+          break;
+      }
+    }
+    return numRecordsWritten;
+  }
+
+  private boolean isHeadersSchemaPresentAndSupported(Schema recordSchema) {
+    Schema.Field headersSchemaField = recordSchema.getField(headersField);
+    if (headersSchemaField != null) {
+      Map.Entry<Schema, Schema> mapSchema = headersSchemaField.getSchema().getMapSchema();
+      return mapSchema.getKey().getType().equals(Schema.Type.STRING) &&
+        mapSchema.getValue().getType().equals(Schema.Type.STRING);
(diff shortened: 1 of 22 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,93 @@
...
+ * Real-time sink for Streams
+ */
+public class StreamSink extends RealtimeSink<StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(StreamSink.class);
+
+  @Override
+  public void configure(StageConfigurer configurer) {
+    configurer.setDescription("Real-time sink that outputs to the specified stream.");
+    configurer.addProperty(new Property(Properties.Stream.NAME,
+                                        "The name of the stream to output to . Must be a valid stream name. " +
```
albertshau: remove space before period 'to output to .'

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,93 @@
...
+                                          "The stream will be created if it does not exist",
```
albertshau: for consistency, add a period at the end of the sentence.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,93 @@
...
+                                        true));
+    configurer.addProperty(new Property(Properties.Stream.DATA_FIELD,
+                                        "Field name in the generic record that contains the data to be written to " +
```
albertshau: since we probably wont be able to switch to generic record in the near future, just say 'record' instead of 'generic record'

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,93 @@
...
+    String dataField = runtimeArguments.containsKey(Properties.Stream.DATA_FIELD) ?
+      runtimeArguments.get(Properties.Stream.DATA_FIELD) : Properties.Stream.DEFAULT_DATA_FIELD;
+
+    for (StructuredRecord structuredRecord : structuredRecords) {
+      Object data = structuredRecord.get(dataField);
+      if (data == null) {
+        LOG.debug("Found null data. Skipping record.");
+        return;
+      }
+      if (data.getClass().isAssignableFrom(StreamEvent.class)) {
```
albertshau: should be using the type from the schema to determine the object.  But a record should never have StreamEvent as the object unless something before this is doing something bad.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,93 @@
...
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(streamName), "Stream name should be non-null, non-empty.");
+    pipelineConfigurer.addStream(new Stream(streamName));
+  }
+
+  @Override
+  public void write(Iterable<StructuredRecord> structuredRecords) throws Exception {
+    SinkContext context = getContext();
+    Map<String, String> runtimeArguments = context.getRuntimeArguments();
+    String streamName = runtimeArguments.get(Properties.Stream.NAME);
+    String dataField = runtimeArguments.containsKey(Properties.Stream.DATA_FIELD) ?
```
albertshau: should do this in initialize rather than do it for each call to write

albertshau: hm i guess being able to get the context enables and encourages behavior like this. i wonder if we should remove that.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,93 @@
...
+  public void configure(StageConfigurer configurer) {
+    configurer.setDescription("Real-time sink that outputs to the specified stream.");
+    configurer.addProperty(new Property(Properties.Stream.NAME,
+                                        "The name of the stream to output to . Must be a valid stream name. " +
+                                          "The stream will be created if it does not exist",
+                                        true));
+    configurer.addProperty(new Property(Properties.Stream.DATA_FIELD,
+                                        "Field name in the generic record that contains the data to be written to " +
+                                          "the specified stream. The data could be either in avro/binary format or " +
+                                          "stream event byte buffer format. If unspecified, 'data' key is used and " +
```
albertshau: not sure what this means.  I think we can enforce that the field is of type bytes or string since that's what's available in the StreamWriter api.  If they need to convert to bytes they can always use the projection transform

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,93 @@
...
+
+    for (StructuredRecord structuredRecord : structuredRecords) {
+      Object data = structuredRecord.get(dataField);
+      if (data == null) {
+        LOG.debug("Found null data. Skipping record.");
+        return;
+      }
+      if (data.getClass().isAssignableFrom(StreamEvent.class)) {
+        context.write(streamName, (StreamEvent) data);
+      } else if (data instanceof ByteBuffer) {
```
albertshau: unfortunately, when the field type is bytes, it's possible the value is a byte[] or a ByteBuffer. You'll have to check which one and behave accordingly.

albertshau:
i dont really understand whats going on with the guice stuff so it would be good to get another opinion on it. otherwise, have a few comments on the actual sink.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
bdmogal: I added ``StreamEvent`` here, because that's the only way users can specify headers. Shouldn't we allow that?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
bdmogal: My bad. Some of this is remnants from a an initial thought-process from a couple of days ago. Will clean it up.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
bdmogal: Also regarding using ``Type`` from the ``StructuredRecord``, do you mean we should sink all fields from the ``StructuredRecord`` based on their type? I thought each ``StructuredRecord`` would result in one write, but I could be mistaken?

albertshau: i would assume one record translates into one stream event.  The event body would come from one field in the record.  Headers could come from another field, as it's possible to have a map<string,string> field in a record.

albertshau: maybe there is some confusion about the schema class and what it can represent?  There shouldn't be a way to have StreamEvent as a field.  It is possible to have a record that represents a StreamEvent, but it would be 3 fields, one for timestamp, one for body, and one for headers, similar to what StreamBatchSource emits.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,154 @@
...
+                                        "Name of the field in the record that contains the data to be written to " +
+                                          "the specified stream. The data could be in binary format as " +
+                                          "a byte array or a ByteBuffer. It can also be a String." +
+                                          "If unspecified, the 'data' key is used.",
+                                        false));
+    configurer.addProperty(new Property(Properties.Stream.HEADERS_FIELD,
+                                        "Name of the field in the record that contains headers. Headers are presumed" +
+                                          " to be a map of string to string.",
+                                        false));
+    configurer.addProperty(new Property(Properties.Stream.TIMESTAMP_FIELD,
```
albertshau: as discussed, timestamp is generated by cdap so we don't need this

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -0,0 +1,154 @@
...
+    ByteBuffer buffer;
+    if (data instanceof ByteBuffer) {
+      buffer = (ByteBuffer) data;
+    } else if (data instanceof byte []) {
+      buffer = ByteBuffer.wrap((byte []) data);
+    } else {
+      throw new UnsupportedTypeException(String.format("Type %s is not supported for writing to stream",
+                                                       data.getClass().getName()));
+    }
+    if (headers != null && headers instanceof Map) {
```
albertshau: we can also check the field to map sure its a map<string, string>, though it may not be worth it since it's an error either way

cdap-unit-test/src/main/java/co/cask/cdap/test/ConfigurableTestBase.java
```diff
@@ -219,7 +222,14 @@ protected static void initTestBase(Map<String, String> additionalConfiguration)
       new DiscoveryRuntimeModule().getInMemoryModules(),
       new AppFabricServiceRuntimeModule().getInMemoryModules(),
       new ServiceStoreModules().getInMemoryModule(),
-      new ProgramRunnerRuntimeModule().getInMemoryModules(),
+      Modules.override(new ProgramRunnerRuntimeModule().getInMemoryModules()).with(new AbstractModule() {
+        @Override
+        protected void configure() {
+          install(new FactoryModuleBuilder().implement(co.cask.cdap.api.data.stream.StreamWriter.class,
```
albertshau: how come it needs the full class names?

albertshau:
couple minor comments, otherwise lgtm

cdap-unit-test/src/main/java/co/cask/cdap/test/ConfigurableTestBase.java
bdmogal: Because we have a conflicting ``StreamWriter`` in ``TestBase`` as well.

bdmogal:
Tested that this works on a cluster.

Waiting for a new build - http://builds.cask.co/browse/CDAP-DUT1516-7.

@albertshau can you take one last look at the last two commits?

albertshau:
one minor comment, then lgtm

bdmogal:
Build passed - http://builds.cask.co/browse/CDAP-DUT1516-7. Merging.

PR #2146, opened 2015-04-20
Feature/3.0 2070 metrics changes

awholegunch:
Updates the HTTP Metrics RESTful API for changes due to 3.0.

The examples were reviewed for their use of queries, and modifications were made to reflect these and any other changes (table names, errors in commands, etc.) as seen fit.

This is staged, with pages of note:
- [Metrics RESTful API](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_2070_Metrics_changes/en/reference-manual/http-restful-api/metrics.html)
- [Queries RESTful API](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_2070_Metrics_changes/en/reference-manual/http-restful-api/query.html)
- [Namespace RESTful API](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_2070_Metrics_changes/en/reference-manual/http-restful-api/namespace.html)
- [Examples](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_2070_Metrics_changes/en/examples-manual/index.html)
- [Operations: Scaling Instances](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_2070_Metrics_changes/en/admin-manual/operations/scaling-instances.html)

Addresses https://issues.cask.co/browse/CDAP-2070 and https://issues.cask.co/browse/CDAP-1596


abaranau:
> A single Dataset across all Applications

"A single Dataset"

> All Datasets across all Applications

"All Datasets"

no need for "across all Applications" and may be not correct as Apps are not the only ones who can potentially use dataset

awholegunch:
Fixed.

cdap-docs/admin-manual/source/operations/scaling-instances.rst
```diff
@@ -58,37 +69,43 @@ Scaling Services
...
 
 with the arguments as a JSON string in the body::
 
   { "instances" : 2 }
   
-Example using the :ref:`CDAP Standalone SDK<standalone-index>` and ``curl`` (reformatted to fit)::
+Example using the :ref:`CDAP Standalone SDK <standalone-index>` and ``curl`` (reformatted to fit)::
 
-  curl -vw'\n' -XPUT 'http://localhost:10000/v3/namespaces/default/apps/WordCount/services/RetrieveCounts/runnables/RetrieveCounts/instances' \
+  curl -w'\n' -X PUT 'http://localhost:10000/v3/namespaces/default/apps/WordCount/services/RetrieveCounts/instances' \
```
anew: what is the \w'\n' for? Is that needed? It seems to be only distracting.

anew:
LGTM for the examples part.

cdap-docs/admin-manual/source/operations/scaling-instances.rst
awholegunch: If you copy and paste the curl call, it adds a trailing return, and so that instead of the results returned being mangled by your command line, it gives a nice clean result. It makes your next command line prompt appear on the next line, and makes the results match the output given in the example exactly. Yes, we could drop them, but I though that having curl calls that can be cleanly cut, pasted, and get clear results was worth the extra bit of code.

abaranau:
* Querying for Multiple Metrics
** Multiple Metrics with the Same Context
** Multiple Metrics with Different Contexts

this structure is confusing. There are two separate things:
* Specify multiple metrics in Query
* Execute multiple Queries

abaranau:
> For example, to retrieve multiple metrics using a curl call (results reformatted to fit):

this example shows neither multiple metrics nor multiple queries

abaranau:
Also, most of the info on querying metrics is in "Query Tips" - is that intentional? I'd expect to have at least these points, when you talk about querying metrics (after you talked about contexts and searching):
* metrics query can done by doing POST at <base-url>/metrics/query
* metrics query parameters can be provided in url or in a post body. In latter case, one can submit multiple queries with single request
* metrics query allows the following parameters: time range, metric names, tag values to filter-by, tag values to group by
* metrics query result (for each query) has the following format: {start,end, series[]}.
  * Query result can have multiple series, each identified by tag values specified in group by and a metric name; hence, there can be one, multiple time series in result, or none.

Then go to url-param examples, query-in-body examples. And then to tips.

Does it make sense?

awholegunch:
Makes sense. Let me revise...

awholegunch:
I've revised this quite a bit. Take a look at [Metrics](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_2070_Metrics_changes/en/reference-manual/http-restful-api/metrics.html).

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -68,33 +74,33 @@ enclosing context. These are the available Application contexts of CDAP:
    * - System Metric
      - Context
    * - One Flowlet of a Flow
-     - ``namespace.<namespace>.app.<app-id>.flow.<flow-id>.flowlet.<flowlet-id>``
+     - ``namespace:<namespace> app:<app-id> flow:<flow-id> flowlet:<flowlet-id>``
```
gsps1: should include run-id here, otherwise we aggregate flowlet of a flow between runs, same applies for examples below that uses program types.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -68,33 +74,33 @@ enclosing context. These are the available Application contexts of CDAP:
...
    * - All Flowlets of a Flow
-     - ``namespace.<namespace>.app.<app-id>.flow.<flow-id>``
+     - ``namespace:<namespace> app:<app-id> flow:<flow-id>``
    * - All Flowlets of all app of an Application
```
gsps1: all flowlets of all flows of an application

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -371,161 +410,193 @@ To query a metric within a given context, perform an HTTP POST request::
...
+  ``...metric=user.names.bytes?aggregate=true``
+  
+instead of the correct
 
-    POST '<base-url>/metrics/query?context=namespace.default.app.CountRandom.flow.*
-      &metric=system.process.events.processed&start=now-5s&count=5'
+  ``...metric=user.names.bytes&aggregate=true``
+  
+will return the empty result, as the metric name will be interpreted as
+``"user.names.bytes?aggregate=true"`` instead of ``"user.names.bytes"``.
```
gsps1: i think this detailed example is not needed, as its not very useful.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -371,161 +410,193 @@ To query a metric within a given context, perform an HTTP POST request::
...
+        }
+  }
 
-  POST http://<host>:<port>/v2/metrics
+For example, to retrieve multiple metrics using a ``curl`` call (results reformatted to fit)::
 
-with the arguments as a JSON string in the body. For example, to retrieve multiple metrics
-using a ``curl`` call::
+  $ curl -w'\n' -X POST 'http://localhost:10000/v3/metrics/query' -H 'Content-Type: application/json' \
+   -d '{"query1":{"tags": {"flow":"CountRandom"}, "metrics": ["system.process.events.processed"], "timeRange": {"start":"now-5s", "count":"5"}}}'
```
gsps1: since this is an example to demonstrate querying multiple metrics, i would give an example where i can query at least 2 metrics.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: Aside from MapReduce, which program types would that be? (I can't go by the UI to tell me...)

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: added.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: Modified, though it sure would have saved me sometime on Sunday...

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: I'm not sure I have these all correct though. Please review.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -371,161 +416,201 @@ To query a metric within a given context, perform an HTTP POST request::
...
+        timeRange: {“aggregate”: “true”}
+        },
+    “query2”: {
+        tags: {“namespace”: “default”},
+        metrics: [“metric1”, “metric2”],
+        groupBy: [“app”, “dataset”],
+        timeRange: {“start”: “now­2s”, “end”: “now”}
+        }
+  }
+
```
gsps1: i think its important to specify that these queries are identified by the queryId and the queryId is used in the returned result to identify the series. 

Also, i would be consistent with the example, you can have 2 metrics in the first queryId and groupBy in the second queryId and use that for the returned result as well. as that will give a good idea how the time series will look.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -612,7 +696,6 @@ difference calculated between the start and end times:
 - ``(endTime - startTime) >= 610 seconds``, resolution will be 1 minute; 
 - otherwise, resolution will be 1 second.
```
gsps1: this interval has been updated, now if difference > 36000 the resolution is in hours and > 600 the resolution will be in minutes.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -638,7 +721,7 @@ difference calculated between the start and end times:
 
 Example::
 
-  POST '<base-url>/metrics/query?context=namespace.default.app.CountRandom&
+  POST '<base-url>/metrics/query?tag=namespace:default&tag=app:CountRandom&
```
gsps1: L#728 the description is incorrect, you are saying the resolution is in seconds, while it should be in minutes

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -648,43 +731,139 @@ For aggregates, you cannot specify a time range. As an example, to return the to
...
+``PurchaseHistoryBuilder`` (reformatted to fit)::
 
-  POST '<base-url>/metrics/query?context=namespace.default.app.PurchaseHistory.mapreduce.
-      PurchaseHistoryWorkflow_PurchaseHistoryBuilder&metric=system.process.completion&aggregate=true'
-  
+  POST '<base-url>/metrics/query?tag=namespace:default&tag=app:PurchaseHistory
+    &tag=mapreduce:PurchaseHistoryBuilder&metric=system.process.completion&aggregate=true'
+    
+  {"startTime":0,"endTime":1429497700,"series":[{"metricName":"system.process.completion",
+   "grouping":{},"data":[{"time":0,"value":200}]}]}
```
gsps1: we are talking about completion percentage and the value is 200 ? that's incorrect.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -648,43 +731,139 @@ For aggregates, you cannot specify a time range. As an example, to return the to
...
+  {"startTime":0,"endTime":1429498425,"series":[{"metricName":"system.process.completion",
+   "grouping":{},"data":[{"time":0,"value":200}]}]}
+   
+  
+  POST '<base-url>/metrics/query?tag=namespace:default&tag=app:CountRandom&tag=flow:CountRandom&tag=run:
+    bca50436-9650-448e-9ab1-f1d186eb2285&tag=flowlet:splitter&metric=system.process.events.processed&aggregate=true'
 
-The last example will return something similar to (where ``"time"=0`` means aggregated total number)::
+The last example will return (where ``"time"=0`` means aggregated total number, and ``endTime`` is
+the query time |---| current time |---| or *now* at the time of the query) something similar to::
```
gsps1: just say endTime is queryTime

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -648,43 +731,139 @@ For aggregates, you cannot specify a time range. As an example, to return the to
...
+      &metric=system.process.events.processed&start=now-5s&count=5'
+
+    POST '<base-url>/metrics/query?tag=namespace:default&tag=app:CountRandom
+      &metric=system.process.events.processed&start=now-5s&count=5'
+
+    POST '<base-url>/metrics/query?tag=namespace:default
+      &metric=system.process.events.processed&start=now-5s&count=5'
+
+- To request user-defined metrics instead of system metrics, specify ``user`` instead of 
+  ``cdap`` in the URL and specify the user-defined metric at the end of the request.
```
gsps1: system, also this description can be made better, say that user metrics are prefixed with string ``user`` and need to be queried by adding user prefix to the metricName.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -648,43 +731,139 @@ For aggregates, you cannot specify a time range. As an example, to return the to
...
+
+  For example, to request a user-defined metric for the *HelloWorld* Application's *WhoFlow* Flow::
```
gsps1: to query the user-defined metric ``names.bytes`` in the *HelloWorld* Application

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: That's what I got. I was wondering about that. Bug?

gsps1:
most of the query tips are very basic , i will add this one, 

if you want to view the task level information of mappers of a mapreduce job, instead of querying for each individual tasks of a mapreduce job, you could make a single query 
```
metrics/query?tag=namespace:default&... &tag=mapreduce:PurchaseHistoryBuilder&tag=tasktype:m&groupBy=instance&metric=system.process.completion 
```
which will return multiple series each with grouping:instanceId and value : completion status. 

this demonstrates how to use groupBy metrics and can be a useful tip.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: Edited.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: Fixed.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: Fixed.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: Fixed.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: fixed.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: I tried that with my example, and got no results. "There's no data aggregated for specified tags to satisfy the query"...do you have an example that works?

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
gsps1: i guess since the tasktype is not specified, we add the process completion for map and reduce task type, making is 200

awholegunch: Fixed. Added changes to make it clear that the Mappers and Reducers context is to be used with those metrics.

awholegunch: Fixed.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -343,26 +358,57 @@ To search for the available metrics within a given context, perform an HTTP POST
...
+     - 
+   * - HTTP Method
+     - ``POST '<base-url>/metrics/search?target=metric&tag=namespace:default&tag=app:HelloWorld&tag=flow:WhoFlow&tag=flowlet:saver'``
+   * - Returns
+     - | ``["system.dataset.store.bytes","system.dataset.store.ops","system.dataset.store.writes",``
+       | `` "system.process.events.in","system.process.events.processed","system.process.instance",``
+       | `` "system.process.tuples.attempt.read","system.process.tuples.read","system.store.bytes",``
+       | `` "system.store.ops","system.store.writes","user.names.bytes"]``
+   * - Description
+     - Returns all metrics in the context of the flowlet *saver* of the application *PurchaseHistory* of the
```
gsps1: app name is HelloWorld , not PurchaseHistory

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -371,161 +417,205 @@ To query a metric within a given context, perform an HTTP POST request::
...
      - 
    * - HTTP Method
-     - ``POST '<base-url>/metrics/query?context=namespace.default.app.HelloWorld.flow.``
-       ``WhoFlow.run.13ac3a50-a435-49c8-a752-83b3c1e1b9a8.flowlet.saver&metric=user.names.bytes&aggregate=true'``
+     - ``POST '<base-url>/metrics/query?tag=namespace:default&tag=app:HelloWorld&tag=flow.WhoFlow``
+       ``&tag=run:13ac3a50-a435-49c8-a752-83b3c1e1b9a8&tag=flowlet:saver&metric=user.names.bytes&aggregate=true'``
+   * - Returns
+     - ``{"startTime":0,"endTime":1429328212,"series":[{"metricName":"user.names.bytes","grouping":{},"data":[{"time":0,"value":8}]}]}``
    * - Description
      - Querying the *User-defined* metric *names.bytes*, of the Flow *saver*, by its run-ID
```
gsps1: Flow *WhoFlow*

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -371,161 +417,205 @@ To query a metric within a given context, perform an HTTP POST request::
...
+.............................
+
+Retrieving multiple metrics at once can be accomplished by issuing an HTTP POST request
+with a JSON list as the request body that enumerates the name and attributes for each
+metric. The format of the request and the JSON body depends on whether the metrics share
+the same context or are being called for different contexts. 
+
+In both cases, queries are identified by a ``<query-id>`` (in the examples below,
+*query1*, *query2*, *eventsIn*, *eventsOut*); the ``<query-id>`` is then used in the
+returned result to identify the series.
```
gsps1: description is incorrect, we don't have query-id's when when we query multiple metrics in the same context using the url and not the body. only when the queries are batched and specified in the body of request, we use query-ids.

gsps1: you might want to move this to ``Multiple Metrics with Different Contexts``

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -608,11 +697,10 @@ points for a metric. By default, 1 second resolution is used. Acceptable values
 above. If ``resolution=auto``, the resolution will be determined based on a time
 difference calculated between the start and end times:
 
-- ``(endTime - startTime) >= 3610 seconds``, resolution will be 1 hour; 
-- ``(endTime - startTime) >= 610 seconds``, resolution will be 1 minute; 
+- ``(endTime - startTime) > 3600 seconds``, resolution will be 1 hour;
```
gsps1: i mentioned 36000 (10 hours), not 3600 (1hr)

gsps1:
@awholegunch why did you replace 2.8 API, example, target=childContext for search and context=<context-name> for query , they have to be deprecated before being removed.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: fixed.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: Fixed, and added "ten hours" and "ten minutes" so the reason for the particular limits is clearer.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: Fixed and re-written for the new context.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: fixed.

awholegunch:
Restored old endpoint, deprecated it, added note to release notes and linked back to replacement endpoint. Restaged docs. Please review again!

gsps1:
hey @awholegunch , i see you have only brought back the target=childContext endpoint, while the ``metrics/search?target=metric&context=<context> ``and ``metrics/query?context=<context>&metric=<metric-name>`` are still not been restored.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
gsps1: I still see 200, not fixed ?

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -638,53 +732,158 @@ difference calculated between the start and end times:
...
+  Flowlet level, the resulting values retrieved will be a sum across all Flowlets of the Flow.
+
+- If you want the number of input objects processed across all Flowlets of a Flow, you 
+  address the Metrics API at the Flow context::
+
+    POST '<base-url>/metrics/query?tag=namespace:default.app:CountRandom?tag=flow:CountRandom?tag=flowlet:*
+      &metric=system.process.events.processed&start=now-5s&count=5'
+
+  Similarly, you can address the context of all Flows of an Application, an entire Application, 
+  or the entire  namespace of a CDAP instance::
```
gsps1: this description is very confusing,  can you make it clearer ?

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -638,53 +732,158 @@ difference calculated between the start and end times:
...
+
+    POST '<base-url>/metrics/query?tag=namespace:default&tag=app:CountRandom&tag=flow:*
+      &metric=system.process.events.processed&start=now-5s&count=5'
```
gsps1: the below query is same as the this, as only flows emit the metric ``system.process.events.processed`` , may be you can remove this.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
awholegunch: There are four descriptions here: which one are you thinking of? The first or last?

awholegunch:
This has been updated and restaged. Please take another look.

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
gsps1: this one : "Similarly, you can address the context of all Flows of an Application, an entire Application, or the entire  namespace of a CDAP instance::"

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -638,53 +746,155 @@ difference calculated between the start and end times:
...
+  |---| and thus across all Flowlets |---| since this metric was actually emitted at the
+  Flowlet level, the resulting values retrieved will be a sum across all Flowlets of the Flow.
+
+- To see events processed by all Flowlets of a Flow in an Application, instead of querying
+  for each individual Flowlet of the Flow, you can perform a single query, using 
+  ``groupBy=flowlet``.
+
+  For example, to request the information for each of the Flowlets of the
+  *PurchaseHistory* Application, this query will return a multiple series, each grouped by
+  the instance (in this case a Flowlet) and with the returned value being the number of
```
gsps1: just say grouped by flowlet

awholegunch: Fixed.

gsps1:
please fix this - https://github.com/caskdata/cdap/pull/2146#discussion_r28929782 
and merge!

awholegunch:
Whew!

PR #2147, opened 2015-04-20
(CDAP-1917) Refactor program TwillApplication and distributed ProgramRunner

chtyim:
- Removed bunch of duplicated code by introducing an AbstractProgramTwillApplication
- Generalize file localize rather than a hard coded three files (program JAR, hConf and cConf)
  - Prepare the path for adding plugin files localization.


chtyim:
Tested on cluster.

albertshau:
lgtm

PR #2151, opened 2015-04-20
remove info logging in pingHandler 

gsps1:
ping endpoint is called frequently and was causing log.saver overflow. so removing the log message.

### DIFF
--- a/cdap-gateway/src/test/resources/logback-test.xml
+++ b/cdap-gateway/src/test/resources/logback-test.xml
@@ -24,13 +24,13 @@
 <!--                                                                         -->
 
 <configuration>
+    <logger name="co.cask.cdap.gateway.handlers.PingHandler" level="TRACE"/>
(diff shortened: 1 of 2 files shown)


chtyim:
You can still log it as trace if you want.

cdap-gateway/src/test/resources/logback-test.xml
```diff
@@ -24,6 +24,7 @@
 <!--                                                                         -->
 
 <configuration>
+    <logger name="co.cask.cdap.gateway.handlers.PingHandler" level="TRACE"/>
```
poornachandra: Can you put all over-rides in the same place. Helps readability.

poornachandra:
Just one minor comment, please fix and merge.

gsps1:
@poornachandra i missed that, have updated it now.

PR #2153, opened 2015-04-20
Dir to place template jars is plugins

gokulavasan:
- [x] Tested InMemory
- [x] Tested SDK


cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -590,12 +590,6 @@
     </property>
 
     <property>
-        <name>app.template.dir</name>
```
albertshau: should not remove from default

sreevatsanraman:
LGTM :+1:

PR #2154, opened 2015-04-20
CDAP-1963 & CDAP-1964: Cube aggregation to have name & query in specific aggregation

abaranau:


cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/cube/DefaultCube.java
```diff
@@ -140,19 +140,30 @@ SELECT count('read.ops')                                     << measure name and
...
-    Aggregation agg = findAggregation(query);
-    if (agg == null) {
-      throw new IllegalArgumentException("There's no data aggregated for specified tags to satisfy the query: " +
-                                           query.toString());
+    // 1) find aggregation to query
+    Aggregation agg;
+    if (query.getAggregation() != null) {
+      agg = aggregations.get(query.getAggregation());
+      if (agg == null) {
+        throw new IllegalArgumentException("Specified aggregation does not exist in cube: " + query.getAggregation());
```
gsps1: should we list the available aggregation names here ? , so its useful for the user if there is a typo.

gsps1:
not needed right now, but should we allow query by aggregation in MetricsHandler ?

abaranau:
Probably not in metrics handler. But MetricStore should use it I think... We should eventually remove auto-choosing of aggregation

gsps1:
just one minor comment, LGTM :+1:

abaranau:
waiting for build: http://builds.cask.co/browse/CDAP-DUT1494-3

PR #2155, opened 2015-04-20
Feature/admin refined

astewes:


cdap-ui/app/features/admin/templates/preferences.html
```diff
@@ -1,8 +1,10 @@
+<h2>System Preferences</h2>
+
 <div class="panel panel-default">
-  <div class="panel-heading">
+<!--   <div class="panel-heading">
     <h3 class="panel-title">{{heading}}</h3>
   </div>
-
+ -->
```
elfenheart: The reason this is using {{ heading }} is because we want to separate between system preferences and namespace preferences. Do we not want to distinguish that anymore?

elfenheart:
:+1: LGTM

PR #2157, opened 2015-04-20
CDAP-2196 Add bounded message buffer to JMS poll to emit some messages instead of just one per poll from adapter

henry-cask:
The PR contains some changes to help buffer some bulk messages from JMS sync:
-) Fix the consumer timeout from 30s to 2s
-) Add small buffer for 50 messages or if no more messages within the receive call.
   The default message could be set by "jms.messages.receive" runtime arg.
   This will allow more buffering to emitted messages when adapter poll to the JMS source.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
@@ -142,13 +142,7 @@ public class JmsSource extends RealtimeSource<String> {
             LOG.trace("Process JMS TextMessage : ", text);
           } else if (message instanceof BytesMessage) {
             BytesMessage bytesMessage = (BytesMessage) message;
-            int bodyLength = (int) bytesMessage.getBodyLength();
-            byte[] data = new byte[bodyLength];
-            int bytesRead = bytesMessage.readBytes(data);
-            if (bytesRead != bodyLength) {
-              LOG.warn("Number of bytes read {} not same as expected {}", bytesRead, bodyLength);
-            }
-            text = new String(data).intern();
+            text = bytesMessage.readUTF();
             LOG.trace("Processing JMS ByteMessage : {}", text);
           } else {
             // Different kind of messages, just get String for now
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
@@ -44,15 +44,18 @@ import javax.jms.TextMessage;
 public class JmsSource extends RealtimeSource<String> {
   private static final Logger LOG = LoggerFactory.getLogger(JmsSource.class);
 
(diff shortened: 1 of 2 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
```diff
@@ -44,15 +44,18 @@
 public class JmsSource extends RealtimeSource<String> {
   private static final Logger LOG = LoggerFactory.getLogger(JmsSource.class);
 
-  private static final long JMS_CONSUMER_TIMEOUT_MS = 30000;
+  private static final long JMS_CONSUMER_TIMEOUT_MS = 2000;
```
chtyim: Can the timeout be configurable?

henry-cask:
http://builds.cask.co/browse/CDAP-DT39

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
```diff
@@ -118,43 +125,46 @@ private void initializeJMSConnection() {
...
+            LOG.trace("Process JMS TextMessage : ", text);
+          } else if (message instanceof BytesMessage) {
+            BytesMessage bytesMessage = (BytesMessage) message;
+            int bodyLength = (int) bytesMessage.getBodyLength();
+            byte[] data = new byte[bodyLength];
+            int bytesRead = bytesMessage.readBytes(data);
+            if (bytesRead != bodyLength) {
+              LOG.warn("Number of bytes read {} not same as expected {}", bytesRead, bodyLength);
+            }
+            text = new String(data).intern();
```
chtyim: Why call intern? This will add the string to the intern String pool of the JVM. Is that what we wanted?

chtyim: Also, when newly a String from `byte[]`, a `Charset` should always be provided.

henry-cask: Yeah, originally I wanted to add the new String into intern buffer pool. 
And as for the byte[] to String convert, I will just use ByteMessage.readUTF to make sure get UTF-8 String.
So will modify the code so no worries about both =)

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
henry-cask: Not right now. The idea since we are polling the data anyway, we just control the timeout from JMS server to let the caller poll again if no messages or timeout.

henry-cask:
Addressed the issues raised

henry-cask:
Build: http://builds.cask.co/browse/CDAP-DT39-2

henry-cask:
Will merge this EOD if no more comments or concerns.

chtyim:
LGTM

henry-cask:
I owe you beer @chtyim  =)

PR #2159, opened 2015-04-20
Added the missing `toLocation()` method that broken develop branch

chtyim:


chtyim:
The change was ported from the https://github.com/caskdata/cdap/pull/2126, which get reviewed. Tested with "mvn test-compile" locally.

PR #2161, opened 2015-04-20
Use DataWriter and pass it to the write method of RealtimeSink

gokulavasan:
Use DataWriter to pass in the correct DatasetContext for every transaction wrapper for Realtime Sink.

Bamboo : http://builds.cask.co/browse/CDAP-DT41


gokulavasan:
Build passed : http://builds.cask.co/browse/CDAP-DT41-4

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/realtime/RealtimeContext.java
```diff
@@ -20,12 +20,12 @@
 import co.cask.cdap.templates.etl.api.StageSpecification;
 
 /**
- * Context passed to the Source stages.
+ * Context passed to the Source and Sink stages of Realtime Adapter.
  */
-public interface SourceContext extends RuntimeContext {
+public interface RealtimeContext extends RuntimeContext {
```
henry-cask: Looks like there is a side effect commits to change from SourceContext to RealtimeContext as part of this PR?

gokulavasan: Since we removed StreamWriter, DatasetContext interfaces from SinkContext, both SourceContext and SinkContext became identical. So I replaced them with RealtimeContext.

henry-cask: Would batch has the something called BatchContext then?

gokulavasan: Batch has BatchSourceContext and BatchSinkContext. Because in batch sources we can setInputs and in batch sinks we can setOutputs.

henry-cask: Ah never mind, just notice that SourceContext is under realtime package.

henry-cask: Should the new context be called RealtimeSourceContext to be consistent with the batch one?

gokulavasan: RealtimeContext is used in both RealtimeSource and RealtimeSink. Hence the name RealtimeContext

albertshau:
:+1:

gokulavasan:
Build passed. Merging : http://builds.cask.co/browse/CDAP-DT41-5

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/realtime/RealtimeContext.java
henry-cask: Ah thanks!

PR #2163, opened 2015-04-20
Feature/flow refined

astewes:


cdap-ui/app/directives/log-viewer/log-viewer.less
```diff
@@ -1,21 +1,23 @@
...
+    border-radius: 4px;
     min-height: 240px;
     max-height: 600px;
 
-    li {
-      margin-top: 10px;
-    }
+    // li {
+    //   margin-top: 10px;
+    // }
```
elfenheart: Let's just remove this instead of commenting it out

ajainarayanan:
:+1: LGTM

PR #2166, opened 2015-04-20
UsageHandler: HTTP handler for UsageDataset

jwang47:
https://issues.cask.co/browse/CDAP-2095
http://builds.cask.co/browse/CDAP-DUT1552


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/UsageHandler.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+@Path(Constants.Gateway.API_VERSION_3)
+@Singleton
+public class UsageHandler extends AbstractHttpHandler {
+
+  private final Supplier<UsageDataset> usageDataset;
+
+  @Inject
+  public UsageHandler(final DatasetFramework datasetFramework) {
+    this.usageDataset = Suppliers.memoize(new Supplier<UsageDataset>() {
```
poornachandra: Datasets are not thread-safe. One way to fix this is to make supplier thread-local.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/UsageHandler.java
```diff
@@ -0,0 +1,174 @@
...
+import javax.ws.rs.PathParam;
+
+/**
+ * The {@link co.cask.http.HttpHandler} for handling REST calls to the usage registry.
+ */
+@Path(Constants.Gateway.API_VERSION_3)
+@Singleton
+public class UsageHandler extends AbstractHttpHandler {
+
+  private final Supplier<UsageDataset> usageDataset;
```
poornachandra: Calling it `usageDatasetSupplier` will be clearer.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/UsageHandler.java
```diff
@@ -0,0 +1,174 @@
...
+    });
+  }
+
+  @GET
+  @Path("/namespaces/{namespace-id}/apps/{app-id}/datasets")
+  public void getAppDatasetUsage(HttpRequest request, HttpResponder responder,
+                                 @PathParam("namespace-id") String namespaceId,
+                                 @PathParam("app-id") String appId) {
+    Id.Application id = Id.Application.from(namespaceId, appId);
+    Set<? extends Id> ids = usageDataset.get().getDatasets(id);
```
poornachandra: Does having `Set<Id.DatasetInstance>` cause errors?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/UsageHandlerTest.java
```diff
@@ -0,0 +1,209 @@
...
+  public void testProgramStreamUsage() throws Exception {
+    Id.Application app = Id.Application.from("somespace", "noapp");
+    Id.Program program = Id.Program.from(app, ProgramType.FLOW, "noprogram");
+    Id.Stream stream = Id.Stream.from("somespace", "nostream");
+
+    Assert.assertEquals(0, getAppStreamUsage(app).size());
+    Assert.assertEquals(0, getProgramStreamUsage(program).size());
+    Assert.assertEquals(0, getStreamProgramUsage(stream).size());
+
+    getUsageDataset().register(program, stream);
```
poornachandra: Will you be deploying an app to register usage instead of directly calling `register` later?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/UsageHandler.java
jwang47: No, this was just to make it easier to copy paste, since these handler methods are the same except for the `UsageDataset` method it calls.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/UsageHandlerTest.java
jwang47: Yes, this test is only for the `UsageHandler`, not testing usage registry integration with apps/streams/etc.

poornachandra:
@jwang47 Done with review.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/UsageHandlerTest.java
poornachandra: You'll re-write the same test to use deploy, right?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/UsageHandler.java
```diff
@@ -0,0 +1,260 @@
...
+    responder.sendJson(HttpResponseStatus.OK, ids);
+  }
+
+  @VisibleForTesting
+  public <R> R executeUsageDatasetOp(TransactionExecutor.Function<UsageDatasetIterable, R> func) {
+    return txnl.executeUnchecked(func);
+  }
+
+  /**
+   *
```
chtyim: shouldn't have empty javadoc.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/UsageHandler.java
```diff
@@ -0,0 +1,260 @@
...
+        @Override
+        public Set<? extends Id> apply(UsageDatasetIterable input) throws Exception {
+          return input.getUsageDataset().getAdapters(id);
+        }
+      });
+    responder.sendJson(HttpResponseStatus.OK, ids);
+  }
+
+  @VisibleForTesting
+  public <R> R executeUsageDatasetOp(TransactionExecutor.Function<UsageDatasetIterable, R> func) {
```
chtyim: Should be package visible only.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/AppFabricServer.java
```diff
@@ -61,6 +61,7 @@
  */
 public final class AppFabricServer extends AbstractIdleService {
 
+  public static final String HANDLERS_BINDING = "appfabric.http.handler";
```
chtyim: Better not to define it in here. Define it in `Constants.AppFabric`.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/AppFabricTestBase.java
```diff
@@ -193,6 +197,10 @@ public static void afterClass() throws Exception {
     txManager.stopAndWait();
   }
 
+  protected UsageHandler getUsageHandler() throws IOException, DatasetManagementException {
```
chtyim: Why? Seems hacky.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/UsageHandlerTest.java
```diff
@@ -0,0 +1,267 @@
...
+      });
+  }
+
+  // app/program/adapter -> dataset/stream
+
+  private Set<Id.DatasetInstance> getAppDatasetUsage(Id.Application app) throws Exception {
+    HttpResponse response = doGet(
+      String.format("/v3/namespaces/%s/apps/%s/datasets", app.getNamespaceId(), app.getId()));
+    Assert.assertEquals(200, response.getStatusLine().getStatusCode());
+    return GSON.fromJson(new InputStreamReader(response.getEntity().getContent()),
```
chtyim: You don't need to close the reader? Also, when creating Reader from InputStream, always need to provide charset.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/AppFabricTestBase.java
poornachandra: You can  use injector to fetch usageHandler in UsageHandlerTest.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/UsageHandlerTest.java
poornachandra: Either close the reader, or use `EntityUtils.toString`.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/AppFabricTestBase.java
jwang47: This is b/c getting the same `UsageDataset` from `UsageHandler` and from `UsageHandlerTest` results in two different instances of the `UsageDataset`, so they aren't really interacting with the same dataset. Should probably file a bug for this.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/UsageHandler.java
jwang47: Needs to be public right now b/c for some reason all the handler tests are in a different package..

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/AppFabricTestBase.java
poornachandra: If table name is the same - you should get the same table. Can you try again?

jwang47: Actually, the issue was that some of the operations on `UsageDataset` wasn't in a transaction, so the operations weren't getting flushed.

poornachandra:
LGTM, please merge after build passes.

PR #2167, opened 2015-04-20
Stream usage registry integration

jwang47:
Depends on https://github.com/caskdata/cdap/pull/2166


cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/UsageDatasetUtil.java
```diff
@@ -0,0 +1,54 @@
...
+import co.cask.cdap.data2.dataset2.DatasetFramework;
+import co.cask.cdap.data2.dataset2.DatasetManagementException;
+import co.cask.cdap.proto.Id;
+
+import java.io.IOException;
+
+/**
+ * Utility to create {@link UsageDataset}.
+ */
+public class UsageDatasetUtil {
```
chtyim: Util class shouldn't be "newable" and should be final.

jwang47:
Working on writing tests for this.

anew:
At a quick glance, this seems a really large PR for as little as it is doing. Can you add a quick summary of what changes were made to help understand it?

jwang47:
Most of it is from https://github.com/caskdata/cdap/pull/2166 (+541 −327), but I'll add a quick summary, yes.

anew:
Ok, I just don't want to double-review 2166. I'll wait for 2166 and then review this.

jwang47:
Reopend PR here https://github.com/caskdata/cdap/pull/2188

PR #2168, opened 2015-04-21
CDAP-1966: dding cube sink and transform to ETLBatchTemplate

abaranau:
Missed a piece: adding sink and transform to ETLBatchTemplate registry.

Also: fixed bug with passing ms instead of sec


cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/templates/etl/transforms/StructuredRecordToCubeFactTransformTest.java
```diff
@@ -199,8 +199,8 @@ public void testTransform() throws Exception {
     transformed = emitter.getEmitted().get(0);
     // note: our date format doesn't include millis
```
gsps1: i don't understand this note, is this relevant btw ?

abaranau: we can remove now. Thogh still relevant - the date format used in test doesn't have millis

albertshau:
:+1:

PR #2169, opened 2015-04-21
Add private constructors to classes in the etl Constants class to enforce noninstantiability

henry-cask:
To enforce noninstantiability add private constuctors to Contants and its nested classes.

Throw AssertionError as reccomended in the Effective Java by Joshua Bloch.
This means that I am asserting that this constructor will never be called so if somehow it is called
then it violates the assumption.


henry-cask:
Build:

http://builds.cask.co/browse/CDAP-DT43-1

albertshau:
:+1:

henry-cask:
Thx! Merged to develop

PR #2172, opened 2015-04-21
changing instance controller for services to be the same as flowlets

elfenheart:
Change instance controller styling to be the same as flowlets

screen shot 2015-04-20 at 6 19 59 pm


nitinmotgi:
LGTM :+1:

PR #2173, opened 2015-04-21
Update to Workflow and Schedules documentation

awholegunch:
Address https://issues.cask.co/browse/CDAP-1128

Pages of interest:

- [Developers' Manual > Building Block > Workflows](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_1128_Workflow_docs/en/developers-manual/building-blocks/workflows.html)
- [Lifecycle RESTful API > Run Records and Schedules](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT-3.0_1128_Workflow_docs/en/reference-manual/http-restful-api/lifecycle.html#run-records-and-schedules)

In the second page, the "Schedules and Workflows: Suspend and Resume" rubric has been added, and all examples in that section have been modified to use a different table structure in attempt to improve the readability. The sections has been reorganized in attempt to show the information pattern more clearly.


cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,89 @@ The diagram for this code would be:
    :width: 8in
    :align: center
 
+Conditional Node
```
bdmogal: Unrelated to this PR, but on the very first line of this page, s/thousand/thousands/

bdmogal: very first paragraph I mean.

bdmogal: Also, in the conditional nodes section, we should perhaps upper case D, A, and G and add the acronym (DAG), since its very well-known

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,89 @@ The diagram for this code would be:
...
+----------------
+
+You can provide a *conditional* node in your structure that allows for branching based on 
+a true-false predicate.
```
bdmogal: a *boolean* predicate is better?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,89 @@ The diagram for this code would be:
...
+      String outputPath = context.getRuntimeArguments().get("outputPath");
+      FileInputFormat.addInputPath(job, new Path(inputPath));
+      FileOutputFormat.setOutputPath(job, new Path(outputPath));
+    }
+  }
+
+  public static class MyVerifier extends Mapper<LongWritable, Text, Text, NullWritable> {
+    public void map(LongWritable key, Text value, Context context)
+      throws IOException, InterruptedException {
+      if (value.toString() and value.toString().equals("BuildProductProfile")) {
```
bdmogal: This won't compile in Java. ``value.toString() != null``, perhaps?

bdmogal:
@awholegunch reviewed the two sections and they mostly look good, apart from some minor comments.

One major comment about workflows:
I think the example makes it clear that you can use map reduce counters in ``Predicate`` classes. However, I think we should call this out more explicitly in documentation as well. 
Something on the lines of - 
In ``Predicates``, for the 3.0 release, CDAP only supports conditions based on counters from your MapReduce jobs. This also implies that condition nodes can *strictly only follow MapReduce nodes*. In a later version, we will expose a lot more kinds of conditions.

Perhaps, we should add a **Predicates** section in the docs as well.

Also, I do not know if this may be too much information for users, but this is implemented via a new concept introduced in CDAP in 3.0, called ``Workflow Tokens``. As I said earlier, the only thing available in ``Workflow Tokens`` in the 3.0 release is MapReduce counters, and therefore you can only have condition nodes after MapReduce programs, but we will add a lot more data in ``Workflow Tokens`` in a later release.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Used.

awholegunch:
ok...when you say "Predicates section", are you thinking of an additional Building Blocks section? Or were you just thinking another part of this page? It seems to me we have enough for the latter at present; the former might come with the next release, but I can see it would be good to set the context, so that people can start to see (and request) possibilities...

But this background is very good, and I will add this in to the page.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: How about 
```
if (value and value.toString().equals("BuildProductProfile"))
```

awholegunch:
Updated and revised. Please take another look, @sagarkapare @bdmogal

bdmogal:
@awholegunch section in this document is just fine for now.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,95 @@ The diagram for this code would be:
...
+is greater than zero, the logic will follow the path of *BuildProductProfileMR*;
+otherwise, the other path will be taken. The diagram for this code would be:
+
+.. image:: /_images/conditional-workflow.png
+   :width: 8in
+   :align: center
+
+For this 3.0 release, CDAP only supports predicate conditions based on counters from your
+MapReduce programs. Because of this, condition nodes can currently only follow MapReduce
+nodes that make use of one or more counters. In a later version, we will expose many more
```
bdmogal: or not, right? I mean a condition node can follow any mapreduce node, irrespective of whether it emits counters. If it doesn't emit counters, it probably just won't be very useful.

awholegunch: Maybe add something to indicate that they will only work if they follow a MapReduce that emits a counter...

bdmogal: I mean condition nodes as a CDAP feature will still work right (in the sense that the condition will be executed, will most likely always make it to one branch of the condition node), it probably may not make sense for applications to use them?

bdmogal: I think its ok if we do not mention it at all. Just mentioning that they must follow MapReduce nodes should be enough. Thoughts?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,95 @@ The diagram for this code would be:
...
+    public boolean apply(@Nullable WorkflowContext input) {
+      if (input != null && input.containsKey("BuildProductProfile")) {
+        Map<String, Long> customCounters = input.getToken().getMapReduceCounters().get("MyCustomCounters");
+        if (customCounters.get("BuildProductProfile") > 0) {
+          return true;
+        }
+      }
+      return false;
+    }
+  }
```
sagarkapare:
```java
public static class MyPredicate implements Predicate<WorkflowContext> {

   @Override
    public boolean apply(@Nullable WorkflowContext input) {
       if (input == null) {
          return false;
       }
       Map<String, Map<String, Long>> mapReduceCounters = input.getToken().getMapReduceCounters();
       if (mapReduceCounters == null) {
          return false;
       }
       Map<String, Long> customCounters = mapReduceCounters.get("MyCustomCounters");
       if (customCounters.get("BuildProductProfile") > 0) {
         return true;
       }
       return false;
   }
}
```

sagarkapare: restructured some code.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,95 @@ The diagram for this code would be:
...
+  
+In the ``JoinWithCatalogMR`` MapReduce, it could have in its Mapper class code that
```
sagarkapare: In the ``JoinWithCatalogMR`` MapReduce, custom Hadoop counters can be emitted that

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Updated.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,100 @@ The diagram for this code would be:
...
+
+In this case, if the predicate finds that the a ``MapReduceCounter`` *BuildProductProfile*
+is greater than zero, the logic will follow the path of *BuildProductProfileMR*;
+otherwise, the other path will be taken. The diagram for this code would be:
+
+.. image:: /_images/conditional-workflow.png
+   :width: 8in
+   :align: center
+
+For this 3.0 release, CDAP only supports predicate conditions based on counters from your
```
sagarkapare: For this 3.0 release, CDAP only supports predicate conditions based on counters from previous

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,100 @@ The diagram for this code would be:
...
+MapReduce programs. Because of this, condition nodes are currently only useful when they
+follow MapReduce nodes that make use of one or more counters. In a later version, we will
```
sagarkapare: follow MapReduce nodes, so that they can make use of counters emitted by the MapReduce program to take the decision.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -185,6 +185,100 @@ The diagram for this code would be:
...
+expose many more kinds of conditions, for other components, to allow other possible
```
sagarkapare: In a later version, we will add more information to the WorkflowContext which will allow to make more complex decisions.

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -604,42 +631,146 @@ For Workflows, you can also retrieve:
...
+       - Retrieves the next runtime of the Workflow *PurchaseHistoryWorkflow* of the Application *PurchaseHistory*
+      
+     * - HTTP Method
+       - ``GET <base-url>/namespaces/default/apps/PurchaseHistory/workflows/PurchaseHistoryWorkflow/nextruntime``
+         
+     * - Returns
+       - | ``[{"id":"DEFAULT.WORKFLOW:developer:PurchaseHistory:PurchaseHistoryWorkflow:0:DailySchedule","time":1415102400000}]``
+       
+
+Schedules and Workflows: Suspend and Resume
```
sagarkapare: Schedules and Workflows are different and so there suspend/resume operations. Its better to put the suspend/resume for the schedule in the building block page for `Schedules`?

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -604,42 +631,146 @@ For Workflows, you can also retrieve:
...
+
+To *resume* means that activity will start up where it was left off, beginning with the start
+of the next program in the sequence.
+
+In the case of the Workflow mentioned above, resuming it after suspension would start up with the
+second of the three MapReduce programs, which is where it would have left off when it was suspended.
+
+With Workflows, *suspend* and *resume* require a *run-id* as the action takes place on
+either a currently running or suspended Workflow.
+
```
sagarkapare: The above description from `To *suspend* means that the current activity ...` is only applicable to the Workflows. For Schedules, suspending it would mean the program associated with that schedule will not trigger again until the schedule is resumed.

awholegunch: Maybe I need to describe them separately?

sagarkapare: yeah it will work

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
awholegunch: Yeah, I'm starting to lean that way. May need it in both locations.

sagarkapare:
LGTM :+1:

PR #2174, opened 2015-04-21
Add cdap-api and cdap-unit-test modules to templates mvn profile.

henry-cask:
Just in case we make changes to the class in those module dependencies we do not need to
rely on snapshots build to happen to make the CI build for feature-adapter branches to pass.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1456,6 +1456,8 @@
     <profile>
       <id>templates</id>
       <modules>
+        <module>cdap-api</module>
+        <module>cdap-unit-test</module>
         <module>cdap-app-templates</module>
       </modules>
     </profile>


henry-cask:
CC @gokulavasan

pom.xml
```diff
@@ -1456,6 +1456,8 @@
     <profile>
       <id>templates</id>
       <modules>
+        <module>cdap-api</module>
```
chtyim: I maybe a bit out of context. But why needed?

chtyim: the cdap-api and cdap-unit-test are in the top level module section. They are always included. For not using snapshot version from `mvn install`, just make sure the version we set for the dependency is `${project.version}`.

henry-cask:
Per discussion with @chtyim with the mvn command we have in the CI we may just need to add version number to unit test dependency.

PR #2175, opened 2015-04-21
Feature/ui buttons tables accordions fixes

elfenheart:
1. Change button on All Applications page to say + Add App
2. Add Caret to Flowlet accordions
3. Fix No Result showing when there is result in View Queries

screen shot 2015-04-21 at 10 53 56 am
screen shot 2015-04-21 at 10 54 08 am
screen shot 2015-04-21 at 10 54 20 am


ajainarayanan:
:+1: LGTM

PR #2177, opened 2015-04-21
create namespace before initializing DSFramework

rohitsinha54:
The system namespace must be created before we setup DSFramework


bdmogal:
:+1:

However, I would recommend to create a bug, and add it to the commit message before merging.

PR #2178, opened 2015-04-21
[CDAP-2171] Deprecated StreamWriter in favor of StreamManager. This will...

bdmogal:
... also allow users to create/manage streams outside apps in TestBase

Issues this PR addresses:
1. In ``TestBase``, you could only interact with streams using the ``StreamWriter`` interface. Also, it wasn't strictly a writer, since it also had a ``create()`` method. However, it was missing a method to read events. #2145 already added a ``StreamManager`` that had ``create`` and ``read`` methods. This PR ports all the write functionality from ``StreamWriter`` into ``StreamManager``.
2. Currently, the only way to get access to the ``StreamWriter`` is through an ``ApplicationManager``. With this PR, you can directly access ``StreamManager`` via a method in ``TestBase``.

Also ported all usages of ``StreamWriter`` in the current tests to use ``StreamManager`` instead. We can remove ``StreamWriter`` in a later release.

Jira: [CDAP-2171](https://issues.cask.co/browse/CDAP-2171)
Build: http://builds.cask.co/browse/CDAP-DUT1566

### DIFF
--- a/cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteStreamManager.java
+++ b/cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteStreamManager.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *  
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *  
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.test.remote;
+
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.flow.flowlet.StreamEvent;
+import co.cask.cdap.client.StreamClient;
+import co.cask.cdap.client.config.ClientConfig;
+import co.cask.cdap.common.exception.StreamNotFoundException;
+import co.cask.cdap.test.StreamManager;
+import com.google.common.base.Charsets;
+import com.google.common.base.Throwables;
+import com.google.common.collect.Lists;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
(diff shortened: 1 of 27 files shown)


jwang47:
This will conflict with https://github.com/caskdata/cdap/pull/2167

bdmogal:
It shouldn't, right? This is the ``StreamWriter`` in ``TestBase``. Not the one in app-fabric.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteStreamManager.java
```diff
@@ -0,0 +1,133 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Implementation of {@link StreamManager} that interacts with stream service via HTTP calls
+ */
+public class RemoteStreamManager implements StreamManager {
```
bdmogal: This entire class is just a copy of the existing ``RemoteStreamWriter``, with the addition of the ``getEvents`` method. The ``RemoteStreamWriter`` now delegates to this class

cdap-test/src/main/java/co/cask/cdap/test/StreamManager.java
```diff
@@ -38,6 +38,74 @@
   void createStream() throws IOException;
 
   /**
+   * Sends a UTF-8 encoded string to the stream.
```
bdmogal: These methods again from from ``StreamWriter`` as is.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteStreamManager.java
```diff
@@ -0,0 +1,133 @@
...
+    } catch (IOException e) {
+      throw e;
+    } catch (Exception e) {
+      throw Throwables.propagate(e);
+    }
+  }
+
+  @Override
+  public void send(Map<String, String> headers, String content) throws IOException {
+    throw new UnsupportedOperationException("TODO");
```
anwar6953: Is there a JIRA for these?
Is it not supported by StreamClient?

bdmogal: Not currently, doesn't look like ``StreamClient`` allows you to pass in headers. Created [CDAP-2211](https://issues.cask.co/browse/CDAP-2211)

anwar6953:
LGTM

bdmogal:
Build Passed. Merging.

PR #2179, opened 2015-04-21
Fix/etl apps save as draft

ajainarayanan:
- Adds ability to save etl adapters as draft
- Syncs (deletes) adapters from user store once saved.
- Changes etlapps routes to adapters route
- Adds a non-functional sort for adapters list (in list view)

### DIFF
--- a/cdap-ui/app/features/etlapps/controllers/create-ctrl.js
+++ b/cdap-ui/app/features/etlapps/controllers/create-ctrl.js
@@ -1,5 +1,5 @@
 angular.module(PKG.name + '.feature.etlapps')
-  .controller('ETLAppsCreateController', function($scope, MyDataSource, $alert, $bootstrapModal, $state, ETLAppsApiFactory, $filter) {
+  .controller('ETLAppsCreateController', function($scope, MyDataSource, $alert, $bootstrapModal, $state, ETLAppsApiFactory, mySettings, $filter) {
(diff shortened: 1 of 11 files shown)


cdap-ui/app/features/etlapps/controllers/create-ctrl.js
```diff
@@ -18,6 +17,29 @@ angular.module(PKG.name + '.feature.etlapps')
...
+        $scope.metadata = $scope.etlDrafts[item].config.metadata;
+        $scope.source = $scope.etlDrafts[item].config.source;
+        $scope.sink = $scope.etlDrafts[item].config.sink;
+        $scope.transforms = $scope.etlDrafts[item].config.transforms;
+      } else {
+        $scope.metadata.name = item;
+        $scope.metadata.type = $scope.metadata.type;
+        $scope.transforms = defaultTransforms;
+        $scope.source = defaultSource;
+        $scope.sink = defaultSink;
```
elfenheart: We are starting to have a lot of $scope variables. Can we group them together and turn them into objects?

ajainarayanan: :+1: We need to break the controller into smaller chunks. Will do it in the next PR

elfenheart:
:+1: LGTM

PR #2180, opened 2015-04-21
Bug fix: Ingesting a string into a stream via the UI would cause the str...

anwar6953:
Bug fix: Ingesting a string into a stream via the UI would cause the string to be JSONified before posting to CDAP.
This resolves that issue by opting for the payload not to be type JSON.


ajainarayanan:
:+1: LGTM

PR #2181, opened 2015-04-21
Make Flow start/stop button dependent on .../flows/<flow-id>/status endp...

anwar6953:
Make Flow start/stop button dependent on `.../flows/<flow-id>/status` endpoint, instead of count of RunRecords that are `RUNNING`.

Note: It is now very similar to the corresponding Services details view/controller.

This is needed because RunRecords can be inaccurate.
For instance, if CDAP master is stopped before the program stops, in which case the stop Run Record is not accurate. The RunRecords would give the impression that the Flow is running and only a STOP button would appear.

[JIRA issue CDAP-2202 created.](https://issues.cask.co/browse/CDAP-2202)


ajainarayanan:
:+1: LGTM

PR #2182, opened 2015-04-21
add Application level preferences

elfenheart:
Application Level Preferences

screen shot 2015-04-21 at 12 50 41 pm
^ Link to the admin page preferences

screen shot 2015-04-21 at 12 50 30 pm

### DIFF
--- a/cdap-ui/app/features/admin/routes.js
+++ b/cdap-ui/app/features/admin/routes.js
@@ -70,7 +70,7 @@ angular.module(PKG.name + '.feature.admin')
             templateUrl: '/assets/features/admin/templates/preferences.html',
             controller: 'PreferencesController',
             resolve: {
-              source: function() {
+              rSource: function() {
                 return 'SYSTEM';
               }
             }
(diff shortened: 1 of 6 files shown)


cdap-ui/app/features/admin/routes.js
```diff
@@ -227,8 +227,23 @@ angular.module(PKG.name + '.feature.admin')
                 parent: 'admin.namespace.detail',
                 url: '/:appId',
                 templateUrl: '/assets/features/admin/templates/namespace/app-metadata.html',
-                controller: 'AdminNamespaceAppMetadataController'
-              });
+                controller: 'AdminNamespaceAppMetadataController',
+                resolve: {
+                    source: function () {
```
ajainarayanan: If we add a parameter to resolve object in a route we usually add a 'r' as a prefix (like rDashboardModel or rNsList) may be we should follow that convention? Just noticing it. We need to probably change in adapters too.

ajainarayanan:
:+1: LGTM

PR #2183, opened 2015-04-21
Fix/ui view queries bug

elfenheart:
- Fix bug when the actual result is 0.
- Move Download All button to be inline with Results

screen shot 2015-04-21 at 1 51 50 pm


ajainarayanan:
:+1: LGTM

PR #2185, opened 2015-04-21
adding handler name to user service metrics ccontext tags

gsps1:
we have metrics of requests to /response from user services, 
this PR is to add the user service handler detail to the metrics context. 
https://issues.cask.co/browse/CDAP-2181

NOTE: we don't add method information to the context right now, but all user services metrics currently are emitted at handler level.


cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
```diff
@@ -432,6 +432,14 @@ public void testAppWithServices() throws Exception {
     Assert.assertEquals(2, serviceMetrics.getProcessed());
     Assert.assertEquals(1, serviceMetrics.getException());
 
+    RuntimeMetrics handlerMetrics = RuntimeStats.getServiceHandlerMetrics(AppWithServices.APP_NAME,
+                                                                          AppWithServices.SERVICE_NAME,
+                                                                          AppWithServices.SERVICE_NAME);
```
abaranau: AppWithServices.SERVICE_NAME twice? typo? how does test work then? :)

abaranau:
just one Q, after fix - LGTM

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
gsps1: in the test app, they use the same name for service and handler, that's why.

gsps1:
http://builds.cask.co/browse/CDAP-DUT1571-4/test - GREEN

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
abaranau: add a comment then :) Since someone asked - it is confusing :)

abaranau:
please add comment and merge!

PR #2187, opened 2015-04-21
Adds the ability to delete a transform from etl adapter create view

ajainarayanan:


cdap-ui/app/features/etlapps/controllers/create-ctrl.js
```diff
@@ -196,6 +196,19 @@ angular.module(PKG.name + '.feature.etlapps')
       });
     };
 
+    $scope.deleteTrasformProperty = function(transform) {
```
elfenheart: mistype on transform

cdap-ui/app/features/etlapps/controllers/create-ctrl.js
```diff
@@ -196,6 +196,19 @@ angular.module(PKG.name + '.feature.etlapps')
...
+      var filterFilter = $filter('filter');
```
elfenheart: we don't need this filter anymore right?

cdap-ui/app/features/etlapps/etlapps.less
```diff
@@ -27,13 +27,16 @@ body.theme-cdap.state-adapters-create {
...
     }
-    .fa-pencil {
+    .fa-pencil,.fa-times {
       font-size: 12px;
       vertical-align: middle;
       &:hover {
         cursor: pointer;
       }
     }
+    .fa-times {
```
elfenheart: Can we move this into the .template.transforms.fa-times? (line 73-76)

cdap-ui/app/features/etlapps/controllers/create-ctrl.js
```diff
@@ -196,6 +196,19 @@ angular.module(PKG.name + '.feature.etlapps')
...
+      var index = $scope.transforms.indexOf(transform);
+      $scope.transforms.splice(index, 1);
```
elfenheart: We are assuming that we can always find the element. Is there a case where we might not find a match when clicking delete (ie. syncing problem)?

ajainarayanan: No. We show the list and we delete it. So if the user clicked on the item to be deleted from an array then we are sure that its there.

cdap-ui/app/features/etlapps/controllers/create-ctrl.js
ajainarayanan: Fixed.

cdap-ui/app/features/etlapps/controllers/create-ctrl.js
ajainarayanan: Fixed.

cdap-ui/app/features/etlapps/etlapps.less
ajainarayanan: :+1:

elfenheart:
:+1: LGTM!

PR #2188, opened 2015-04-21
CDAP-2216: Program/App-Stream usage registry integration

jwang47:
https://issues.cask.co/browse/CDAP-2216
http://builds.cask.co/browse/CDAP-DUT1550-8

TODO:
- [ ] Test in distributed cluster. Some guice bindings may be missing.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
@@ -22,6 +22,7 @@ import co.cask.cdap.api.stream.StreamEventData;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.discovery.EndpointStrategy;
 import co.cask.cdap.common.discovery.RandomEndpointStrategy;
+import co.cask.cdap.data2.registry.UsageRegistry;
 import co.cask.cdap.proto.Id;
 import co.cask.common.http.HttpMethod;
 import co.cask.common.http.HttpRequest;
@@ -29,12 +30,15 @@ import co.cask.common.http.HttpRequests;
 import co.cask.common.http.HttpResponse;
 import com.google.common.base.Charsets;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
 import com.google.common.net.HttpHeaders;
 import com.google.inject.Inject;
 import com.google.inject.assistedinject.Assisted;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import org.apache.twill.discovery.Discoverable;
 import org.apache.twill.discovery.DiscoveryServiceClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
@@ -50,14 +54,27 @@ import java.util.concurrent.TimeUnit;
  */
 public class DefaultStreamWriter implements StreamWriter {
 
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultStreamWriter.class);
+
   private final String namespaceId;
   private final EndpointStrategy endpointStrategy;
+  /**
+   * The program that is using this {@link StreamWriter}.
+   */
+  private final Id.Program program;
+  private final Map<Id.Stream, Boolean> isStreamRegistered;
+  private final UsageRegistry usageRegistry;
 
   @Inject
   public DefaultStreamWriter(@Assisted("namespaceId") String namespaceId,
+                             @Assisted("programId") Id.Program program,
+                             UsageRegistry usageRegistry,
                              DiscoveryServiceClient discoveryServiceClient) {
     this.namespaceId = namespaceId;
+    this.program = program;
     this.endpointStrategy = new RandomEndpointStrategy(discoveryServiceClient.discover(Constants.Service.STREAMS));
+    this.isStreamRegistered = Maps.newConcurrentMap();
+    this.usageRegistry = usageRegistry;
   }
 
   private URL getStreamURL(String stream) throws IOException {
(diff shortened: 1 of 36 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -557,6 +562,12 @@ private void configureStreamInput(Job job, StreamBatchReadable stream) throws IO
 
     job.setInputFormatClass(StreamInputFormat.class);
 
+    try {
```
anew: Is this the only way that a MR can read a stream? Isn't there a way to specify it via ``setInputDataset("stream:<name>"``?

anew:
Overall looks good. I am just not sure whether this captures all possible ways of consuming a stream. See my comment.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
poornachandra: In that case too `configureStreamInput` is called - https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java#L456

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
```diff
@@ -50,14 +54,27 @@
...
 
   @Inject
   public DefaultStreamWriter(@Assisted("namespaceId") String namespaceId,
+                             @Assisted("programId") Id.Program program,
+                             UsageRegistry usageRegistry,
                              DiscoveryServiceClient discoveryServiceClient) {
     this.namespaceId = namespaceId;
+    this.program = program;
     this.endpointStrategy = new RandomEndpointStrategy(discoveryServiceClient.discover(Constants.Service.STREAMS));
+    this.isStreamRegistered = Maps.newHashMap();
```
poornachandra: `write` can be called concurrently, right? If so we should make the map `ConcurrentMap`.

jwang47: I was assuming
1. `HashMap` won't fail if `write` is called concurrently
2. `ConcurrentMap` would incur more overhead, which would become significant

poornachandra: Using data structures not designed for concurrency can lead to undefined behavior. I don't think using `ConcurrentMap` will add any significant overhead to stream operations.

jwang47: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
```diff
@@ -50,14 +54,27 @@
...
 
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultStreamWriter.class);
+
   private final String namespaceId;
   private final EndpointStrategy endpointStrategy;
+  /**
+   * The program that is using this {@link StreamWriter}.
+   */
+  private final Id.Program program;
+  private final Map<Id.Stream, Boolean> isStreamRegistered;
```
chtyim: Why a map to boolean value? Why not use a `Set`?

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
```diff
@@ -80,6 +97,15 @@ private URL getStreamURL(String stream, boolean batch) throws IOException {
   }
 
   private void writeToStream(Id.Stream stream, HttpRequest request) throws IOException {
+    if (!isStreamRegistered.containsKey(stream)) {
```
chtyim: So there is race here if there are concurrent calls to `writeToStream` (both threads `containsKey` return false, they'll both register and both update the map). I guess it's ok to make more than one register call, but strictly speaking it's promoting a wrong assumption here. If you want to make it atomic, you should use `putIfAbsent`

chtyim: Strictly speaking, you'll need a check->sync->check->update logic.

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/StreamWriterFactory.java
```diff
@@ -25,8 +26,9 @@
 public interface StreamWriterFactory {
   /**
    * @param namespaceId the namespaceId for which to return a {@link StreamWriter}
+   * @param programId the programId that will be using the {@link StreamWriter}
    * @return a {@link StreamWriter} for the specified namespaceId
    */
-  StreamWriter create(@Assisted("namespaceId") String namespaceId);
+  StreamWriter create(@Assisted("namespaceId") String namespaceId, @Assisted("programId") Id.Program programId);
```
chtyim: Isn't Id.Program contains namespaceId? Why pass it twice?

anew:
@jwang47 please do not squash all your commits into a single commit until the review is done. Here you only needed a few lines of change to address comments, but reviewers now have to read 500 lines of changes again. It is much easier to do an incremental review if the comments are addressed in a separate commit.

jwang47:
Opened new PR for unaddressed comments: https://github.com/caskdata/cdap/pull/2205

PR #2191, opened 2015-04-21
(CDAP-1917) Integrate plugin support to adapter

chtyim:
Public API changes:

- Modifed AdapterConfigurer to have methods for declaring plugins usage.
- Introduce a PluginSelector class for selecting plugin at configure adapter time
- Added AdapterContext to expose access to plugin classes at runtime.
- Have MapReduceContext and WorkerContext to implements AdapterContext for exposing plugin access

Implementation details:

- Have AbstractContext implements AdapterContext.
  - Children classes can provide Adapter spec and PluginInstantiator to enable plugin support
- Modify AbstractDistributedProgramRunner to localize plugin jars to twill container
- Modify AbstractProgramTwillRunnable to setup plugin path in cConf
- Update WorkerProgramRunner to create WorkerContext with plugin support
- Update MapReduceProgramRunner to enable plugin support
  - Involves updating multiple MapReduce context related classes
  - Have the plugin jars send to MR DistributedCache so that all task containers have access to plugins

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterConfigurer.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterConfigurer.java
@@ -31,35 +31,35 @@ import java.util.Map;
  * Currently, only Worker or Workflow can be used in the execution.
  */
 @Beta
-public interface AdapterConfigurer {
+public interface AdapterConfigurer extends AdapterPluginRegistry {
 
   /**
    * Set the schedule for the program. Must be set for Workflows and is not valid for other program types.
    *
    * @param schedule {@link Schedule}
    */
-  public void setSchedule(Schedule schedule);
+  void setSchedule(Schedule schedule);
 
   /**
    * Set the number of instances of the program. Valid only for Workers and defaults to 1.
    *
    * @param instances number of instances
    */
-  public void setInstances(int instances);
+  void setInstances(int instances);
 
   /**
    * Set the resources the program should use.
    *
    * @param resources the resources the program should use
    */
-  public void setResources(Resources resources);
+  void setResources(Resources resources);
 
   /**
    * Add arguments to be passed to the program as runtime arguments.
    *
    * @param arguments the runtime arguments to add to the program
    */
-  public void addRuntimeArguments(Map<String, String> arguments);
+  void addRuntimeArguments(Map<String, String> arguments);
 
   /**
    * Add argument to be passed to the program as runtime arguments.
@@ -67,7 +67,7 @@ public interface AdapterConfigurer {
    * @param key the runtime argument key
    * @param value the runtime argument value
    */
-  public void addRuntimeArgument(String key, String value);
+  void addRuntimeArgument(String key, String value);
 
   /**
    * Adds a {@link Stream} to the Adapter. The stream will be created during adapter creation if it does not
@@ -112,7 +112,5 @@ public interface AdapterConfigurer {
    * @param datasetClass dataset class to create the Dataset type from
    * @param props dataset instance properties
    */
-  void createDataset(String datasetName,
-                     Class<? extends Dataset> datasetClass,
-                     DatasetProperties props);
+  void createDataset(String datasetName, Class<? extends Dataset> datasetClass, DatasetProperties props);
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractContext.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractContext.java
@@ -23,6 +23,8 @@ import co.cask.cdap.api.data.DatasetInstantiationException;
 import co.cask.cdap.api.dataset.Dataset;
 import co.cask.cdap.api.metrics.Metrics;
 import co.cask.cdap.api.metrics.MetricsCollector;
+import co.cask.cdap.api.templates.AdapterContext;
+import co.cask.cdap.api.templates.plugins.PluginProperties;
 import co.cask.cdap.app.program.Program;
 import co.cask.cdap.app.runtime.Arguments;
 import co.cask.cdap.app.services.AbstractServiceDiscoverer;
@@ -30,21 +32,30 @@ import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.data.dataset.DatasetInstantiator;
 import co.cask.cdap.data2.dataset2.DatasetFramework;
 import co.cask.cdap.internal.app.program.ProgramTypeMetricTag;
+import co.cask.cdap.internal.app.runtime.adapter.PluginInstantiator;
+import co.cask.cdap.templates.AdapterPlugin;
+import co.cask.cdap.templates.AdapterSpecification;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+import com.google.common.io.Closeables;
 import org.apache.twill.api.RunId;
 import org.apache.twill.discovery.DiscoveryServiceClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.Closeable;
+import java.io.IOException;
 import java.util.Map;
 import java.util.Set;
+import javax.annotation.Nullable;
 
 /**
  * Base class for program runtime context
  */
-public abstract class AbstractContext extends AbstractServiceDiscoverer implements DatasetContext, RuntimeContext {
+public abstract class AbstractContext extends AbstractServiceDiscoverer
+                                      implements DatasetContext, RuntimeContext, AdapterContext {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractContext.class);
 
   private final Program program;
@@ -57,12 +68,27 @@ public abstract class AbstractContext extends AbstractServiceDiscoverer implemen
   private final DatasetInstantiator dsInstantiator;
   private final DiscoveryServiceClient discoveryServiceClient;
 
-  public AbstractContext(Program program, RunId runId,
-                         Arguments arguments,
-                         Set<String> datasets,
-                         final MetricsCollector metricsCollector,
-                         DatasetFramework dsFramework,
-                         DiscoveryServiceClient discoveryServiceClient) {
+  private final AdapterSpecification adapterSpec;
+  private final PluginInstantiator pluginInstantiator;
+
+  /**
+   * Constructs a context without application template adapter support.
+   */
+  protected AbstractContext(Program program, RunId runId, Arguments arguments,
+                            Set<String> datasets, MetricsCollector metricsCollector,
+                            DatasetFramework dsFramework, DiscoveryServiceClient discoveryServiceClient) {
+    this(program, runId, arguments, datasets, metricsCollector, dsFramework, discoveryServiceClient, null, null);
+  }
+
+  /**
+   * Constructs a context. To have application template adapter support,
+   * both the {@code adapterSpec} and {@code pluginInstantiator} must not be null.
+   */
+  protected AbstractContext(Program program, RunId runId, Arguments arguments,
+                            Set<String> datasets, MetricsCollector metricsCollector,
+                            DatasetFramework dsFramework, DiscoveryServiceClient discoveryServiceClient,
+                            @Nullable AdapterSpecification adapterSpec,
+                            @Nullable PluginInstantiator pluginInstantiator) {
     super(program);
     this.program = program;
     this.runId = runId;
@@ -76,9 +102,16 @@ public abstract class AbstractContext extends AbstractServiceDiscoverer implemen
     // todo: this should be instantiated on demand, at run-time dynamically. Esp. bad to do that in ctor...
     // todo: initialized datasets should be managed by DatasetContext (ie. DatasetInstantiator): refactor further
     this.datasets = Datasets.createDatasets(dsInstantiator, datasets, runtimeArguments);
+    this.adapterSpec = adapterSpec;
+    this.pluginInstantiator = pluginInstantiator;
   }
 
   public abstract Metrics getMetrics();
+
+  @Nullable
+  public AdapterSpecification getAdapterSpec() {
+    return adapterSpec;
+  }
 
   @Override
   public String toString() {
@@ -147,8 +180,14 @@ public abstract class AbstractContext extends AbstractServiceDiscoverer implemen
    * method to release additional resources.
    */
   public void close() {
-    for (Closeable ds : datasets.values()) {
-      closeDataSet(ds);
+    try {
+      for (Closeable ds : datasets.values()) {
+        closeDataSet(ds);
+      }
+    } finally {
+      if (pluginInstantiator != null) {
+        Closeables.closeQuietly(pluginInstantiator);
+      }
     }
   }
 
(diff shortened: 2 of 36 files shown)


chtyim:
unit-test are yet to be added. Will come in next PR.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/ProgramOptionConstants.java
```diff
@@ -51,5 +51,7 @@
 
   public static final String ADAPTER_NAME = "adapterName";
```
bdmogal: do we need both ``ADAPTER_NAME`` and ``ADAPTER_SPEC``?

chtyim: I keep it for now because the `ProgramLifecycleService.start` method is using it for updating `RunRecord` in store. Don't want to introduce too big of a refactoring around the lifecycle service in this PR.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/DynamicMapReduceContext.java
```diff
@@ -72,13 +74,16 @@ public DynamicMapReduceContext(Program program,
...
                                  Arguments runtimeArguments,
                                  MapReduceSpecification spec,
-                                 long logicalStartTime, String workflowBatch, String adapterName,
+                                 long logicalStartTime, String workflowBatch,
                                  DiscoveryServiceClient discoveryServiceClient,
                                  MetricsCollectionService metricsCollectionService,
                                  TransactionSystemClient txClient,
-                                 DatasetFramework dsFramework) {
+                                 DatasetFramework dsFramework,
+                                 @Nullable AdapterSpecification adapterSpec,
```
bdmogal: nit: Looks like ``type`` and ``taskId`` are also ``@Nullable``. Unrelated, but if it doesn't increase the size of this PR, maybe we can mark them so as well?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -818,13 +847,80 @@ private Location copyProgramJar() throws IOException {
...
+   */
+  @Nullable
+  private Location createPluginArchive(@Nullable AdapterSpecification adapterSpec,
+                                       Id.Program programId) throws IOException {
+    if (adapterSpec == null) {
+      return null;
+    }
+
+    Set<PluginInfo> pluginInfos = adapterSpec.getPluginInfos();
+    if (pluginInfos.isEmpty()) {
```
bdmogal: When does this happen - I mean that an adapter does not contain any ``pluginInfos``?

chtyim: It's not required for adapter to have plugin. Someone can write an app template that doesn't use plugin at all.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -818,13 +847,80 @@ private Location copyProgramJar() throws IOException {
...
+    }
+
+    File templateDir = new File(cConf.get(Constants.AppFabric.APP_TEMPLATE_DIR));
+    File pluginDir = new File(templateDir, adapterSpec.getTemplate());
+    File jarFile = File.createTempFile("plugin", ".jar");
+    try {
+      JarOutputStream output = new JarOutputStream(new FileOutputStream(jarFile));
+      try {
+        // Create the directory entries
+        output.putNextEntry(new JarEntry(adapterSpec.getTemplate() + "/"));
```
bdmogal: Just checked that ``JarEntry`` doesn't take a ``File`` or something like that as input :(

Will this break on windows?

chtyim: Jar Entry is always "/" separated. It's abstract path.

chtyim: See http://docs.oracle.com/javase/6/docs/api/java/util/zip/ZipEntry.html#isDirectory()

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -164,6 +169,50 @@ public TwillController launch(TwillApplication twillApplication) {
   }
 
   /**
+   * Gets plugin files that needs to be localized for the adapter. If the given run is not an adapter, no
+   * modification will be done to the map.
+   */
+  private Map<String, File> addAdapterPluginFiles(ProgramOptions options, Map<String, File> localizeFiles) {
```
bdmogal: Perhaps call this ``filesToLocalize`` to make it a noun?

chtyim: Why needs to be a noun? This method is to add files that needs to be localized to the given map. Shouldn't it call "addXXX" ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/worker/WorkerProgramRunner.java
```diff
@@ -110,6 +120,23 @@ public ProgramController run(Program program, ProgramOptions options) {
     return controller;
   }
 
+  @Nullable
+  private AdapterSpecification getAdapterSpecification(Arguments arguments) {
```
bdmogal: Thinking out loud - These methods would potentially be required in every program that is supported through app templates, right? with very similar logic? Is there a way to abstract them?

Perhaps static methods ``AdapterSpecification.from(Arguments)`` (or better ``AdapterSpecification.from(@Nullable String json)``, ``PluginInstantiator.from(@Nullable AdapterSpecification, ClassLoader)``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
bdmogal: oh... I meant the map, not the method

cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterConfigurer.java
```diff
@@ -112,7 +115,64 @@
...
+   *
+   * @param pluginType plugin type name
+   * @param pluginName plugin name
+   * @param properties properties for the plugin. The same set of properties will be used to instantiate the plugin
+   *                   instance at execution time
+   * @param <T> type of the plugin class
+   * @return A new instance of the plugin class or {@code null} if no plugin was found
+   */
+  @Nullable
+  <T> T usePlugin(String pluginType, String pluginName, PluginProperties properties);
```
gokulavasan: Will this be exposed to plugins through PipelineConfigurer for them to depend on other plugins? For example, DBSource plugin might want to add the driver class that it depends on during the configurePipeline stage?

cdap-app-fabric/src/main/java/co/cask/cdap/templates/AdapterPlugin.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Returns the set of properties available for the plugin when the adapter was created.
+   */
+  public PluginProperties getProperties() {
+    return properties;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
bdmogal: can we use Guice here since this is not in API?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/worker/WorkerProgramRunner.java
chtyim: 1. No static method (mainly is not good, and partly because the AdapterSpecification is going to move into cdap-api).
2. The current ProgramRunner hierarachy doesn't have an abstract base class. When we need it in many program runner, we can abstract the program runner logic into some abstract base. I can put a TODO there (just don't wanna do it in the PR and don't have time to create a separate PR to refactor it properly).

cdap-app-fabric/src/main/java/co/cask/cdap/templates/DefaultAdapterConfigurer.java
```diff
@@ -122,6 +147,101 @@ public void createDataset(String datasetInstanceName,
...
+  public <T> Class<T> usePluginClass(String pluginType, String pluginName, PluginProperties properties) {
+    return usePluginClass(pluginType, pluginName, properties, new PluginSelector());
+  }
+
+  @Override
+  public <T> Class<T> usePluginClass(String pluginType, String pluginName,
+                                     PluginProperties properties, PluginSelector selector) {
+
+    Preconditions.checkArgument(hasPlugin(pluginType, pluginName),
+                                "Plugin of type %s, name %s was already added.", pluginType, pluginName);
```
bdmogal: This message is misleading? Should it be "...should have been already added" or "...does not exist"?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/worker/WorkerProgramRunner.java
bdmogal: :+1:

cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterConfigurer.java
chtyim: PipelineConfigurer is internal and specific for ETL app. The CDAP system doesn't know about it. It's the job of the ETL program to expose it through the `PipelineConfigurer` if needed.

cdap-app-fabric/src/main/java/co/cask/cdap/templates/AdapterPlugin.java
chtyim: This class is used inside `AdapterSpecification`, which we are planning to move it to api so that app can access it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractContext.java
```diff
@@ -77,10 +103,17 @@ public AbstractContext(Program program, RunId runId,
...
     // todo: initialized datasets should be managed by DatasetContext (ie. DatasetInstantiator): refactor further
     this.datasets = Datasets.createDatasets(dsInstantiator, datasets, runtimeArguments);
+    this.adapterSpec = adapterSpec;
+    this.pluginInstantiator = pluginInstantiator;
   }
 
   public abstract Metrics getMetrics();
 
+  @Nullable
+  public AdapterSpecification getAdapterSpec() {
```
gokulavasan: Should AdapterContext have this method?

cdap-app-fabric/src/main/java/co/cask/cdap/templates/DefaultAdapterConfigurer.java
chtyim: If `hasPlugin` return true, meaning the plugin is already added in this configurer from some prior call to one of those `usePlugin*` method. This is to avoid adding the same plugin with different properties (and potentially have the selector to select a different plugin).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractContext.java
chtyim: AdapterSpecification is not in cdap-api yet. Hence no for now. We can move it there when we move AdapterSpecification.

cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterConfigurer.java
gokulavasan: Oh got it. ok.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
bdmogal: sounds good!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
bdmogal: oh ok.

cdap-app-fabric/src/main/java/co/cask/cdap/templates/AdapterPlugin.java
bdmogal: ok

cdap-app-fabric/src/main/java/co/cask/cdap/templates/DefaultAdapterConfigurer.java
gokulavasan: So we can't we have two Transform plugins one after another? One say does filtering and another does say renaming?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -164,6 +169,50 @@ public TwillController launch(TwillApplication twillApplication) {
...
+
+    // Get all unique PluginInfo from the adapter spec
+    Set<PluginInfo> plugins = adapterSpec.getPluginInfos();
+
+    // If there is no plugin used by the adapter, nothing need to be localized
+    if (plugins.isEmpty()) {
+      return localizeFiles;
+    }
+
+    File templateDir = new File(cConf.get(Constants.AppFabric.APP_TEMPLATE_DIR));
```
gokulavasan: so if original app.template.dir was set to /opt/cdap/master/plugins, templateDir will be plugins and pluginDir will be plugins/etlbatch ?

gokulavasan:
Couple of questions. Primarily - https://github.com/caskdata/cdap/pull/2191#discussion_r28835717
Rest LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/templates/DefaultAdapterConfigurer.java
chtyim: hum.. so they are the same plugin but configure differently? hum.. probably need that.. Meaning when calling this method, you need to provide a unique name so that it can be used to associate with the property. Let me think about it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
chtyim: The `PluginInstantiator` works with this dir structure, hence maintaining the dir structure make the PluginInstantiator much simpler to be used in different mode.

gokulavasan:
LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/DynamicMapReduceContext.java
chtyim: Do it in separate PR. In our current code base, usage of Nullable is not uniform.

chtyim:
Addressed comments, fixed unit-test and rebased on develop.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1576-3

chtyim:
Import statement conflict resolved against develop. Squashed commits into one.

PR #2192, opened 2015-04-21
Adding support for bcookie - Browser cookie

nitinmotgi:
This cookie is needed even in non secured or logged in mode to avoid to storing preferences of the user efficiently.

### DIFF
--- a/cdap-ui/server/express.js
+++ b/cdap-ui/server/express.js
@@ -12,10 +12,12 @@ module.exports = {
 
 var pkg = require('../package.json'),
     express = require('express'),
+    cookieParser = require('cookie-parser'),
     compression = require('compression'),
     finalhandler = require('finalhandler'),
     serveFavicon = require('serve-favicon'),
     request = require('request'),
+    uuid = require('node-uuid'),
     log4js = require('log4js'),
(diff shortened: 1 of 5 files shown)


cdap-ui/server/express.js
```diff
@@ -161,7 +164,17 @@ function makeApp (authAddress, cdapConfig) {
...
-      res.sendFile(DIST_PATH + '/index.html');
+      // BCookie is the browser cookie, that is generated and will live for a year.
+      // This cookie is always generated to provide unique id for the browser that 
+      // is being used to interact with the CDAP backend.
+      if(! req.cookies.bcookie) {
+        var date = new Date();
+        date.setDate(date.getDate() + 365); // Expires after a year.
+        res.cookie('bcookie', uuid.v4(), { expires: date});
+      } else {
+        res.cookie('bcookie', req.cookies.bcookie, { expires: date});
```
bdmogal: ``date`` seems undefined here, no?

bcookie intrigued me, so checked this PR :)

nitinmotgi: Fixed!

elfenheart:
:+1: LGTM

PR #2194, opened 2015-04-22
CDAP-2198 fix a bug in stop adapter where the deletion of the

albertshau:
schedule could succeed but the stopping of active runs could fail.
In that scenario, the stop fails and every subsequent stop will
fail because the schedule isn't found when we try to delete it.

Also fixing a bug in DistributedProgramRuntimeService where it
could not look up runtime info of programs that were not already
in memory.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
@@ -452,6 +452,8 @@ public class AdapterService extends AbstractIdleService {
       // was some failure stopping the active run.  In that case, the next time stop is called
       // the schedule will not be present. We don't want to fail in that scenario, so its ok if the
       // schedule was not found.
+      LOG.trace("Could not delete adapter workflow schedule {} because it does not exist. Ignoring and moving on.",
+                workflowId, e);
     }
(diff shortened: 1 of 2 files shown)


gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-DUT1580

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -443,9 +443,16 @@ private void stopWorkflowAdapter(Id.Namespace namespace, AdapterSpecification ad
...
     String scheduleName = adapterSpec.getScheduleSpec().getSchedule().getName();
-    scheduler.deleteSchedule(workflowId, SchedulableProgramType.WORKFLOW, scheduleName);
-    //TODO: Scheduler API should also manage the MDS.
-    store.deleteSchedule(workflowId, SchedulableProgramType.WORKFLOW, scheduleName);
+    try {
+      scheduler.deleteSchedule(workflowId, SchedulableProgramType.WORKFLOW, scheduleName);
+      //TODO: Scheduler API should also manage the MDS.
+      store.deleteSchedule(workflowId, SchedulableProgramType.WORKFLOW, scheduleName);
+    } catch (NotFoundException e) {
+      // its possible a stop was already called and the schedule was deleted, but then there
```
henry-cask: Could we add trace log just in case we need debuggibility for this case?

albertshau: sure will do

PR #2196, opened 2015-04-22
CDAP-2220: Merge StructuredRecordToCubeFactTransform and CubeSink

abaranau:


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/StructuredRecordToCubeFact.java
```diff
@@ -14,7 +14,7 @@
  * the License.
  */
 
-package co.cask.cdap.templates.etl.transforms;
+package co.cask.cdap.templates.etl.batch.sinks;
```
albertshau: this probably makes more sense to be in common, that way the real time sink can use it too

albertshau:
1 comment, otherwise lgtm

abaranau:
http://builds.cask.co/browse/CDAP-DUT1581-1

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/StructuredRecordToCubeFact.java
abaranau: true!

abaranau: can I move it along with https://issues.cask.co/browse/CDAP-2221 (Implement realtime CubeSink)?

gokulavasan:
@abaranau Start a build under http://builds.cask.co/browse/CDAP-DT since DUT does not include templates profile.

abaranau:
@gokulavasan Thanks! Started! http://builds.cask.co/browse/CDAP-DT50-1

PR #2197, opened 2015-04-22
CDAP-2219: Program/App-Dataset usage registry integration

jwang47:
https://issues.cask.co/browse/CDAP-2219
http://builds.cask.co/browse/CDAP-DUT1583

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
@@ -26,6 +26,7 @@ import co.cask.cdap.data2.datafabric.dataset.instance.DatasetInstanceManager;
 import co.cask.cdap.data2.datafabric.dataset.service.executor.DatasetAdminOpResponse;
 import co.cask.cdap.data2.datafabric.dataset.service.executor.DatasetOpExecutor;
 import co.cask.cdap.data2.datafabric.dataset.type.DatasetTypeManager;
+import co.cask.cdap.data2.registry.UsageRegistry;
 import co.cask.cdap.explore.client.ExploreFacade;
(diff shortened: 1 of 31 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
```diff
@@ -119,6 +128,17 @@ public void getInfo(HttpRequest request, HttpResponder responder, @PathParam("na
       }
       // typeMeta is guaranteed to be non-null now.
       DatasetMeta info = new DatasetMeta(spec, typeMeta, null);
+      if (ownerType != null && ownerId != null) {
+        try {
+          if (ownerType.equals(Id.getType(Id.Program.class))) {
+            usageRegistry.register(Id.Program.fromStrings(ownerId.split("/")), datasetId);
```
anew: Is usageRegistry thread safe?

jwang47: Should be, as it simply executes dataset ops within a transaction like so:

```
  public void register(final Id.Program programId, final Id.DatasetInstance datasetInstanceId) {
    txnl.executeUnchecked(new TransactionExecutor.Function<UsageDatasetIterable, Void>() {
      @Override
      public Void apply(UsageDatasetIterable input) throws Exception {
        input.getUsageDataset().register(programId, datasetInstanceId);
        return null;
      }
    });
  }
```

anew: Transactions don't make it thread-safe. But since you get a new instance of the dataset for each transactions, it should be fine.

anew:
One comment is that there are datasets that are declared to be used statically in the program specs. These should be registered when the app is deployed. Per our discussion, please implement that in a new PR.

LGTM.

PR #2201, opened 2015-04-22
[CDAP-2225] make /v3/namespaces/{namespace-id}/streams return content type application/json

anew:
- refactor list .../streams endpoint, also removes some old unneeded code that was left over from v2
- the old code did not use sendJson() but constructed a Json String and used sendBytes(), which made the content type binary/octet-stream


sreevatsanraman:
LGTM :+1:

nitinmotgi:
LGTM :+1:

PR #2202, opened 2015-04-22
Twitter Source emits StructuredRecord

gokulavasan:
Removed custom Tweet class and emit all the data as a StructuredRecord. 

One potential issue is the memory consumed by the InMemory queue can grow unbounded if the rate of arrival of tweets is lot more than the rate at which we poll. Any suggestions on if we should address that and how would be useful. One thought was to take in an optional field for size of the queue and check the size and not enqueue if the size is greater but the size call on a concurrentLinkedQ is not a constant.

Build: http://builds.cask.co/browse/CDAP-DT48


sreevatsanraman:
LGTM :+1:

sreevatsanraman:
Build passes merging

PR #2203, opened 2015-04-22
Add default JmsProvider implementation that tries to get ConnectionFactory and Destination via JNDI

henry-cask:
This PR add default JmsProvider that tries to get ConnectionFactory and Destination from JNDI lookup.

The JniBasedJmsProvider will create ConnectionFactory and Destination from JNDI context.
The JmsSource need to be passed several runtime properties for the JNDI JmsProvider:
-) java.naming.factory.initial
-) java.naming.provider.url
-) jms.destination.name

Also update the JmsSource#configure to add more information about config that could be
passed to the JmsSource#initialize.


henry-cask:
Build success:
http://builds.cask.co/browse/CDAP-DT49

PR #2204, opened 2015-04-22
CDAP-2169 delete app through AppLifeCycleHttpHandler

rohitsinha54:
Delete the template (app) when all the associated adapters are deleted.
I am doing it through the AppLifecycleHttpHandler and want to get feedback if its a clean way to do it.
Testing on cluster pending.


chtyim:
Shouldn't do it through handler. Should be done through the AdapterService?

rohitsinha54:
Yes, I thought about doing that by injecting AppLifecycleHttpHandler to AdapterService but then AdapterService does not have HTTPRequest and Responder which is needed to call deleteApp in AppLifecycleHttpHandler. So, should we make the AdapterHTTPHandler pass that too AdapterService too or some other way ?

rohitsinha54:
New PR open according to the suggestions given by @chtyim here: https://github.com/caskdata/cdap/pull/2221

PR #2206, opened 2015-04-22
Use new names for different adapter tests

sreevatsanraman:


bdmogal:
Both tests seem to pass for me locally. It seems like on Bamboo the first test failed because of some edge case/timing issue. As a side effect, the second test failed as well. That should not have happened.

We should get this fix in, so the second failing test does not depend on the outcome of the first.

@sreevatsanraman , Let's create a jira for the first failure and address it if it happens again?

:+1:

sreevatsanraman:
Opened a JIRA: https://issues.cask.co/browse/CDAP-2232

PR #2207, opened 2015-04-22
Packaging adapter artifacts in standalone and master

sreevatsanraman:
Directory structure for packaging

<pre>
plugins/
            < etl-batch-jar >
            < etl-realtime-jar >
            etlBatch/
                          < etl-lib-jar >
            etlRealtime/
                          < etl-lib-jar >
</pre>

- [x] Packaged part of cdap-standlone
- [x] Packaged part of cdap-master rpm and debian
- [x] Pending testing


cdap-master/pom.xml
```diff
@@ -157,24 +152,69 @@
...
+                <id>copy-batch-plugin</id>
+                <phase>prepare-package</phase>
+                <goals>
+                  <goal>copy-resources</goal>
+                </goals>
+                <configuration combine.self="override">
+                  <outputDirectory>${stage.opt.dir}/plugins/${etl.batch.plugin.name}</outputDirectory>
+                  <resources>
+                    <resource>
+                      <directory>${project.parent.basedir}/cdap-app-templates/cdap-etl/cdap-etl-lib/target</directory>
```
gokulavasan: So etl lib JAR is copied to both realtime and batch? ok for now, do we need to split etl-lib into batch-lib, realtime-lib, transforms-lib?

gokulavasan: Or in fact they could all be separate. RealtimeSource.jar, RealtimeSinks.jar, Transforms.jar etc.

gokulavasan: Also how will this work inmemory? If I build app-templates, will I get template and plugin jars separately? That I can manually copy into plugins directory ?

cdap-master/pom.xml
```diff
@@ -138,16 +138,11 @@
     <profile>
       <id>dist</id>
       <properties>
+        <stage.plugins.dir>${project.build.directory}/plugins</stage.plugins.dir>
```
dereklwood: looks like this isn't used anywhere?

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
```diff
@@ -102,6 +105,7 @@
               <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
               <Embed-Transitive>true</Embed-Transitive>
               <Embed-Directory>lib</Embed-Directory>
+              <Export-Package>co.cask.cdap.etl.api.*</Export-Package>
```
gokulavasan: co.cask.cdap.etl.templates.api

cdap-app-templates/cdap-etl/cdap-etl-realtime/pom.xml
```diff
@@ -89,6 +90,7 @@
               <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
               <Embed-Transitive>true</Embed-Transitive>
               <Embed-Directory>lib</Embed-Directory>
+              <Export-Package>co.cask.cdap.etl.api.*</Export-Package>
```
gokulavasan: co.cask.cdap.etl.templates.api

cdap-master/pom.xml
sreevatsanraman: This will have to be done in code.

cdap-app-templates/cdap-etl/cdap-etl-realtime/pom.xml
```diff
@@ -59,6 +59,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <scope>test</scope>
```
chtyim: This should be set in dependencyManagement in the parent, especially as for junit, which is almost always never used other than in test scope.

cdap-master/pom.xml
```diff
@@ -138,16 +138,10 @@
     <profile>
       <id>dist</id>
       <properties>
+        <etl.batch.plugin.name>etlBatch</etl.batch.plugin.name>
+        <etl.realtime.plugin.name>etlRealtime</etl.realtime.plugin.name>
```
chtyim: You can define a properties file in the src/main/resources/etl.properties have content as:

```
etl.batch.plugin.name=${etl.batch.plugin.name}
etl.realtime.plugin.name=${etl.realtime.plugin.name}
```

then in the Java profile (e.g. ETLReadltimeTemplate.java), you can get it by:

```java
Properties prop = new Properties();
InputStream input = getClass().getResourceAsStream("/etl.properties");
prop.load(input);
...
configurer.setName(prop.get("etl.realtime.plugin.name"));
...
```

This ensure the name is defined in once place.

chtyim: This is the standard way of having maven to do properties substitution. See `ProjectInfo` class in cdap-common on the usage.

cdap-app-templates/cdap-etl/cdap-etl-realtime/pom.xml
sreevatsanraman: You are right. Removed.

cdap-master/pom.xml
sreevatsanraman: Great, learnt a new feature :-). Fixed.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLBatchTemplate.java
```diff
@@ -86,9 +95,23 @@ public void configureAdapter(String adapterName, ETLBatchConfig etlBatchConfig,
 
   @Override
   public void configure(ApplicationConfigurer configurer, ApplicationContext context) {
-    configurer.setName("etlbatch");
-    configurer.setDescription("Batch Extract-Transform-Load (ETL) Adapter");
-    configurer.addMapReduce(new ETLMapReduce());
-    configurer.addWorkflow(new ETLWorkflow());
+    Properties prop = new Properties();
```
chtyim: I think you can move this logic into `ETLTemplate`. Something like `getAppName`.

sreevatsanraman: good point. fixe.d

chtyim:
LGTM. Merge if tested.

PR #2209, opened 2015-04-22
CDAP-2227 fixing bug where streams were not getting created at

albertshau:
adapter creation time. Turns out its because the stream files get
created, but the stream metadata is not created.

To fix, refactored so that StreamAdmin implementations write the
metadata as well as create the filesystem directories so that
you can't call create stream on the admin without metadata getting
written. Also removing unused leveldb and hbase stream admins.


albertshau:
http://builds.cask.co/browse/CDAP-DUT1585

henry-cask:
Would feature branch build app template? I thought we need to use feature-adapter prefix?

albertshau:
this is almost all platform code, so wanted it to run the full set of unit tests

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -393,7 +398,8 @@ public void run() {
           }
           alterExploreStream(StreamUtils.getStreamIdFromLocation(streamLocation), false);
           configLocation.delete();
-        } catch (IOException e) {
+          streamMetaStore.removeStream(streamId);
+        } catch (Exception e) {
```
bdmogal: Does this change behavior? Catching ``Exception`` instead of ``IOException``?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/inmemory/InMemoryStreamAdmin.java
```diff
@@ -35,18 +37,24 @@
...
+                             StreamMetaStore streamMetaStore) {
     super(queueService);
     this.usageRegistry = usageRegistry;
+    this.streamMetaStore = streamMetaStore;
   }
 
   @Override
   public void dropAllInNamespace(Id.Namespace namespace) throws Exception {
     queueService.resetStreamsWithPrefix(QueueName.prefixForNamedspacedStream(namespace.getId()));
+    for (StreamSpecification spec : streamMetaStore.listStreams(namespace)) {
```
bdmogal: Should we instead add a ``removeAll(Id.Namespace)`` method in ``StreamMetastore``, so that it can be made transactional?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
albertshau: it does a little bit in that any run time exception that would have happened before will now get wrapped again because of Throwables.propagate

bdmogal:
@albertshau just 1 major comment about adding ``StreamMetastore.removeAll``, and 1 minor question. 

Also, I think I can already see it, but can you confirm with @chtyim that the Queue-based Stream implementation is not needed anymore? Are there any other tests that we could remove since these classes are gone now?

Other than that, the rest LGTM :+1:

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/inmemory/InMemoryStreamAdmin.java
albertshau: the real implementation isn't transactional though, it goes one by one and deletes them.

bdmogal: you mean in the ``FileStreamAdmin``? Yeah, saw that... your call :)

albertshau: think will leave as is until there is a hard requirement.

PR #2210, opened 2015-04-22
[CDAP-2238] Add StreamWriterFactory binding to AbstractTwillRunnable

bdmogal:
``StreamWriterFactory`` binding was missing from ``AbstractTwillRunnable``

Jira: [CDAP-2238](https://issues.cask.co/browse/CDAP-2238)
Build: https://builds.cask.co/browse/CDAP-DUT1593

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/DistributedProgramRunnerModule.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/DistributedProgramRunnerModule.java
@@ -15,11 +15,8 @@
  */
 package co.cask.cdap.app.guice;
 
-import co.cask.cdap.api.data.stream.StreamWriter;
 import co.cask.cdap.app.runtime.ProgramRunner;
 import co.cask.cdap.app.runtime.ProgramRuntimeService;
-import co.cask.cdap.app.stream.DefaultStreamWriter;
-import co.cask.cdap.app.stream.StreamWriterFactory;
 import co.cask.cdap.internal.app.runtime.ProgramRunnerFactory;
 import co.cask.cdap.internal.app.runtime.distributed.DistributedFlowProgramRunner;
 import co.cask.cdap.internal.app.runtime.distributed.DistributedMapReduceProgramRunner;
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/DistributedProgramRunnerModule.java
```diff
@@ -63,6 +63,7 @@ protected void configure() {
     // Create StreamWriter factory.
     install(new FactoryModuleBuilder().implement(StreamWriter.class, DefaultStreamWriter.class)
               .build(StreamWriterFactory.class));
+    expose(StreamWriterFactory.class);
```
chtyim: Why? Who uses it?

bdmogal: Not needed. Removed the binding.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
```diff
@@ -380,11 +384,25 @@ public Cancellable announce(String serviceName, int port) {
...
     );
   }
 
+  private Module createStreamFactoryModule() {
+    return new PrivateModule() {
+      @Override
+      protected void configure() {
+        install(new FactoryModuleBuilder().implement(StreamWriter.class, DefaultStreamWriter.class)
+                  .build(StreamWriterFactory.class));
+        expose(StreamWriterFactory.class);
```
chtyim: If you only have one binding and is always exposed, you don't need to use PrivateModule. Just using AbstractModule is fine.

chtyim: But I guess is ok to keep it like this. No need to change.

nitinmotgi:
LGTM :+1:

bdmogal:
Unrelated build failures because of unstable develop. I'm going to merge this, to avoid hogging build resources for this change, since it does not impact any build/unit test scenarios. It only impacts distributed mode, which I have already tested this change on.

PR #2041, opened 2015-04-15
CDAP-2039 Generating RunId in the AbstractProgramRuntimeService and pass...

sagarkapare:
...ing it as ProgramOptions to the programs.

Jira: https://issues.cask.co/browse/CDAP-2039

Move the RunId generation logic to the AbstractProgramRuntimeService. All programs that are started explicitly will now receive the RunId as a part of ProgramOptions. 
When the program is started as a part of the Workflow, workflow will create the new RunId and then pass it as a part of ProgramOptions to the underlying program.

PR #2212, opened 2015-04-22
Generating RunId in the AbstractProgramRuntimeService and passing it as ProgramOptions to the programs.

sagarkapare:
Jira: https://issues.cask.co/browse/CDAP-2039
This is continuation of the PR #2041. I closed that PR since I was no longer able to update it. Copying the description here -
The PR contains following changes: 
1. Move the RunId generation logic to the AbstractProgramRuntimeService. All programs that are started explicitly will now receive the RunId as a part of ProgramOptions. 
2. When the program is started as a part of the Workflow, workflow will create the new RunId and then pass it as a part of ProgramOptions to the underlying program.
3. RunRecords of the program now contains two id's - one is generated by the CDAP and another is generated by Twill. In standalone the twill run id will be null.
4. Updated the RuntimeInfo object to add the twill RunId. This is required to do mapping between CDAP RunId and TwillRunId.


sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT1592-1

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -42,7 +46,7 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(AbstractProgramRuntimeService.class);
 
-  private final Table<ProgramType, RunId, RuntimeInfo> runtimeInfos;
+  protected final Table<ProgramType, RunId, RuntimeInfo> runtimeInfos;
```
chtyim: This is dangerous. I don't like exposing raw data structure directly to children classes.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -54,12 +58,28 @@ protected AbstractProgramRuntimeService(ProgramRunnerFactory programRunnerFactor
...
+    programStarted(runtimeInfo);
     runtimeInfos.put(runtimeInfo.getType(), runtimeInfo.getController().getRunId(), runtimeInfo);
     return runtimeInfo;
   }
 
+  private ProgramOptions createProgramOptionsWithRunId(ProgramOptions options) {
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+    builder.putAll(options.getArguments().asMap());
+    // Generate the new RunId and add it to the system arguments
+    RunId runId = RunIds.generate();
```
chtyim: You are overloading this method to do two things (generation of RunId and copying/update the ProgramOptions). It's better to have this method as:

`addRunId(ProgramOptions options, RunId runId);`

and call it like:

`ProgramOptions opt = addRunId(options, RunId.generate());`

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -58,12 +58,13 @@
...
-   * @param startTime start timestamp in seconds; if run id is time-based pass the time from the run id
-   * @param adapter   name of the adapter associated with the run
+   * @param id         Info about program
+   * @param pid        run id
+   * @param startTime  start timestamp in seconds; if run id is time-based pass the time from the run id
+   * @param adapter    name of the adapter associated with the run
+   * @param twillRunId twill run id
    */
-  void setStart(Id.Program id, String pid, long startTime, String adapter);
+  void setStart(Id.Program id, String pid, long startTime, String adapter, @Nullable String twillRunId);
```
chtyim: Shouldn't just update the method (as that would introduce unnecessary changes to pass in `null`). Should just add an overloaded method and have the existing one modified to call this one with `null`.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -127,6 +128,12 @@
   List<RunRecord> getRuns(Id.Program id, ProgramRunStatus status, long startTime, long endTime, int limit);
 
   /**
+   * Fetches all active run records.
```
chtyim: What do mean by active? I guess you mean `ProgramRunStatus.RUNNING`, right? better be consistent with naming. Also better have this method more generic as:

`List<RunRecord> getRuns(ProgramRunStatus status)`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -133,7 +131,9 @@ public ProgramController run(final Program program, ProgramOptions options) {
                                 : System.currentTimeMillis();
 
     String workflowBatch = arguments.getOption(ProgramOptionConstants.WORKFLOW_BATCH);
-    String adapterName = arguments.getOption(ProgramOptionConstants.ADAPTER_NAME);
+    final String adapterName = arguments.getOption(ProgramOptionConstants.ADAPTER_NAME);
+    final String twillRunId = arguments.getOption(ProgramOptionConstants.TWILL_RUN_ID);
```
chtyim: Why a `String`? Twill RunId is also of `RunId` type.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
```diff
@@ -128,18 +132,45 @@
   }
 
   @Override
+  protected RuntimeInfo createRuntimeInfo(ProgramController controller, Program program) {
+    RunId twillRunId = ((AbstractTwillProgramController) controller).getTwillRunId();
```
chtyim: Better check `instanceOf` because casting. You should see a yellow warning in the IntelliJ for this statement. We should go for all green, instead of full of yellow warning.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
```diff
@@ -128,18 +132,45 @@
...
+  @Nullable
+  private synchronized RunId getTwillRunId(RunId runId) {
+    RuntimeInfo runtimeInfo = super.lookup(runId);
+    if (runtimeInfo != null) {
+      return runtimeInfo.getTwillRunId();
+    }
+    return null;
+  }
+
+  private synchronized boolean twillRunIdCached(RunId twillRunId) {
```
chtyim: Name it as `isTwillRunIdCached`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
```diff
@@ -128,18 +132,45 @@
...
+    for (RuntimeInfo runtimeInfo : runtimeInfos.values()) {
+      if (runtimeInfo.getTwillRunId() != null && runtimeInfo.getTwillRunId().equals(twillRunId)) {
```
chtyim: Just do `if (twillRunId.equals(runtimeInfo.getTwillRunId()) { ... }`. Java `equals()` works with null value.

chtyim: Assuming the `twillRunId` in the parameter never null, which is ok in this case since this is a private method.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
sagarkapare: The `twillRunId` is only used for passing it to store, which accepts in the string format, hence using `String` here.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
sagarkapare: We always should pass the twillrunid to the `setStart` method right? In standalone it would be null, but in distributed it will have some value.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
sagarkapare: Renamed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
sagarkapare: Updated the code.

sagarkapare:
@chtyim Can you please take a look again? Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -54,20 +60,33 @@ protected AbstractProgramRuntimeService(ProgramRunnerFactory programRunnerFactor
...
-    final SimpleRuntimeInfo runtimeInfo = new SimpleRuntimeInfo(runner.run(program, options), program);
-    addRemover(runtimeInfo);
+    ProgramOptions optionsWithRunId = addRunId(options, RunIds.generate());
+    final RuntimeInfo runtimeInfo = createRuntimeInfo(runner.run(program, optionsWithRunId), program);
+    programStarted(runtimeInfo);
     runtimeInfos.put(runtimeInfo.getType(), runtimeInfo.getController().getRunId(), runtimeInfo);
     return runtimeInfo;
   }
 
+  private ProgramOptions addRunId(ProgramOptions options, RunId runId) {
```
chtyim: It'll be good write a line of comment saying that the returned object is a copy of the given one, with runId added.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -54,20 +60,33 @@ protected AbstractProgramRuntimeService(ProgramRunnerFactory programRunnerFactor
...
+
+    return new SimpleProgramOptions(options.getName(), new BasicArguments(builder.build()), options.getUserArguments(),
+                                    options.isDebug());
+  }
+
+  protected RuntimeInfo createRuntimeInfo(ProgramController controller, Program program) {
+    return new SimpleRuntimeInfo(controller, program);
+  }
+
+  protected List<RuntimeInfo> getRuntimeInfos() {
```
chtyim: I believe this needs to be synchronized since it access the `runtimeInfos` table.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -127,6 +128,14 @@
   List<RunRecord> getRuns(Id.Program id, ProgramRunStatus status, long startTime, long endTime, int limit);
 
   /**
+   * Fetches the run records for the particular status.
+   * @param status  status of the program running/completed/failed or all
+   * @param limit   max number of entries to fetch for this history call
+   * @return        lis of logged runs
+   */
+  List<RunRecord> getRuns(ProgramRunStatus status, int limit);
```
chtyim: Not very useful to have limit without a start. How is it being used?

chtyim: Shouldn't it more like it takes a Predicate to determine what to include?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
```diff
@@ -128,55 +132,70 @@
...
+  public synchronized RuntimeInfo lookup(Id.Program programId, final RunId runId) {
+    RuntimeInfo runtimeInfo = super.lookup(programId, runId);
+    if (runtimeInfo != null) {
+      return runtimeInfo;
     }
 
-    Matcher matcher = APP_NAME_PATTERN.matcher(appName);
-    if (!matcher.matches()) {
-      LOG.warn("Unrecognized application name pattern {}", appName);
+    RunRecord record = store.getRun(programId, runId.getId());
```
chtyim: This logic should be done inside the for loop down below and it only need to consult Store if a Twill program that matches the Id.Program is running (i.e. from the twill live list).

chtyim: Remember that hitting Store has much higher cost that looping the twill live list. Hitting Store is transactionally and involves quite a number of RPC round trip and scan on HBase. Looping twill live list is all happening in-memory.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
sagarkapare: Added comment.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
sagarkapare: Yeah. synchronized the method.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
sagarkapare: Moved logic to access the store inside the loop.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
```diff
@@ -128,55 +132,70 @@
...
-      for (TwillController c : liveInfo.getControllers()) {
-        if (c.getRunId().equals(runId)) {
-          appName = liveInfo.getApplicationName();
-          controller = c;
-          break;
-        }
-      }
-      if (controller != null) {
-        break;
+  private synchronized boolean isTwillRunIdCached(RunId twillRunId) {
```
chtyim: Add a TODO to have a better data structure to support this more efficiently. It can be as simple as a object that contains a Table and a Set.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
```diff
@@ -196,38 +218,71 @@ public synchronized RuntimeInfo lookup(final RunId runId) {
...
-        RuntimeInfo runtimeInfo = createRuntimeInfo(type, programId, controller);
+        twillProgramInfo.put(programId, twillRunId, controller);
+      }
+    }
+
+    if (twillProgramInfo.isEmpty()) {
+      return ImmutableMap.copyOf(result);
+    }
+
+    List<RunRecord> activeRunRecords = store.getRuns(ProgramRunStatus.RUNNING, 100);
```
chtyim: Why 100? As I suggested above, should use a Predicate and use the Predicate while scanning (meaning don't get a full list and filter).

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
sagarkapare: Currently it is only called from the `DistributedProgramRuntimeService` and limit is set to 100. We can add predicate as a parameter, which would filter based on the list of twillRunId. Will add TODO for this and do cleanup in another PR, if thats ok.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
sagarkapare: Yes. Will add a predicate. It will be simpler for me to do in another PR.

chtyim: The Predicate should look something like this:

```java
Set<RunId> twillRunIds = twillProgramInfos.columnKeySet();
store.getRuns(RUNNING, new Predicate<RunRecord>() {
  boolean accept(RunRecord record) {
    return twillRunIds.contains(record.getTwillRunId());
  }
});
```

chtyim: With that list of `RunRecord`, you can just loop over the run record list (which shouldn't be too big, and should be of the same size as the `twillProgramInfos` table) and constructs RuntimeInfo by performing lookup into the `twillProgramInfos` table.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
chtyim: See my comment below. Using Predicate would simplify the code.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
sagarkapare: Added note.

sagarkapare:
@chtyim Please take a look again. Thanks!

chtyim:
LGTM. We may need to do more testing of it in distributed mode.

sagarkapare:
Thanks a lot @chtyim for the review and suggestions. Yes will do more testing on distributed.

PR #2213, opened 2015-04-22
Fix StreamClient to adapt to /streams change

jwang47:
http://builds.cask.co/browse/CDAP-DUT1594

### DIFF
--- a/cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
+++ b/cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
@@ -26,6 +26,7 @@ import co.cask.cdap.common.exception.UnauthorizedException;
 import co.cask.cdap.common.stream.StreamEventTypeAdapter;
 import co.cask.cdap.internal.io.SchemaTypeAdapter;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.StreamDetail;
 import co.cask.cdap.proto.StreamProperties;
 import co.cask.cdap.proto.StreamRecord;
 import co.cask.cdap.security.authentication.client.AccessToken;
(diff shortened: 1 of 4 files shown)


cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -253,10 +254,10 @@ public void setTTL(String streamId, long ttlInSeconds)
    * @return list of {@link StreamRecord}s
    * @throws IOException if a network error occurred
    */
-  public List<StreamRecord> list() throws IOException, UnauthorizedException {
+  public List<StreamSpecification> list() throws IOException, UnauthorizedException {
     URL url = config.resolveNamespacedURLV3("streams");
     HttpResponse response = restClient.execute(HttpMethod.GET, url, config.getAccessToken());
-    return ObjectResponse.fromJsonBody(response, new TypeToken<List<StreamRecord>>() { }).getResponseObject();
+    return ObjectResponse.fromJsonBody(response, new TypeToken<List<StreamSpecification>>() { }).getResponseObject();
```
anew: actually it is a StreamDetail... seems we have some duplication in our classes. See AppFabricDataHttpHandler

jwang47: Oh, looks like there's a typo in the handler code then, b/c I see it's return `specs` rather than `result` which actually isn't returned. https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppFabricDataHttpHandler.java#L76

anew: Oh right. Can you fix that along with this PR, please?

jwang47: Opened new PR here: https://github.com/caskdata/cdap/pull/2223 since no change is needed if return type is `StreamDetail`

anew: But shouldn't we use StreamDetail here?

sreevatsanraman:
LGTM :+1:

PR #2216, opened 2015-04-22
Dataset Status View, fix Streams Type bug, Data Metadata tab

elfenheart:
- Fix bug where type stream is no longer returned by the backend
- Add Metadata page on Data detail
- Add Status page on Dataset

### DIFF
--- a/cdap-ui/app/features/datasets/controllers/tabs/status-ctrl.js
+++ b/cdap-ui/app/features/datasets/controllers/tabs/status-ctrl.js
@@ -3,16 +3,13 @@ angular.module(PKG.name + '.feature.datasets')
     function($scope, MyDataSource, $state, myHelpers) {
       $scope.writes = 0;
       $scope.reads = 0;
+      $scope.storage = 0;
       $scope.transactions = 0;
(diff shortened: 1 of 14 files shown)


cdap-ui/app/features/datasets/controllers/tabs/status-ctrl.js
```diff
@@ -30,59 +28,35 @@ angular.module(PKG.name + '.feature.datasets')
...
+          var data = query(metricData, 'series', 0, 'data', 0, 'value');
           $scope[metric.scopeProperty] = data;
         });
       }
 
-      var query = myHelpers.objectQuery;
-      // Temporary API until we get a status API for each dataset.
+      dataSrc.poll({
+        _cdapPath : '/metrics/query?metric=system.dataset.store.bytes&context=namespace.' +
+                    $state.params.namespace + '.dataset.' + currentDataset + '&aggregate=true',
```
ajainarayanan: Is this the new API(modified metrics API with tags as query params)?

ajainarayanan:
Apart from the metrics path LGTM.

PR #2217, opened 2015-04-22
Minor fixes to UI

anwar6953:
1) amDateFormat expected millisecond time values.
2) When using metrics endpoints that use tags, '~' does not get replaced with '.' in the backend. Only when using context (2.8 APIs).
3) Unneeded file.
4) Malformed query url and improper metric name.


elfenheart:
:+1: LGTM

cdap-ui/app/features/dashboard/controllers/widget.js
```diff
@@ -24,7 +24,9 @@ angular.module(PKG.name+'.feature.dashboard')
       }
       tags = {};
       for (i = 0; i < parts.length; i+=2) {
-        tags[parts[i]] = parts[i + 1]
+        // In context, '~' is used to represent '.'
+        var tagValue = parts[i + 1].replace(/~/g, '.');
```
gsps1: would be great if we use the new search API, but okay for now.  please add a todo

anwar6953: Yea, I created this for now:
https://issues.cask.co/browse/CDAP-2236

PR #2220, opened 2015-04-22
remove duplicate aggregation, we already have handler + method aggregation

gsps1:
https://github.com/caskdata/cdap/pull/1775/ added handler + method aggregation in DefaultMetricStore and my recent merge https://github.com/caskdata/cdap/pull/2185 added a duplicate for that. so removing the duplicate aggregation rule.


abaranau:
LGTM. Tests should have cought it,, hm

gsps1:
how ? we have aggregation for {namespace} and {namespace, app} , this aggregation is similar, they are not exact duplicate. one aggregation had {... service, handler, method} while other had {..service, handler}, also if it was exact duplicate the second one would have replaced the first as we add it to a map. am i missing something ?

PR #2221, opened 2015-04-22
CDAP-2169: Delete the app if the last adapter got deleted

rohitsinha54:
- Introduces ApplicationLifecycleService which is responsible for managing ApplicationLifecycle. Currently, I have refactored out the code of removal of application from AppLifecycleHttpHandler to this class and AppLifecycleHttpHandler delegates to ApplicationLifecycleService for removing applications.

- This also allow AdapterService to delete the application using ApplicationLifecycleService if the deleted adapter was the last associated adapter of the application (template)

Build: http://builds.cask.co/browse/CDAP-DUT1598-1

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
@@ -18,16 +18,10 @@ package co.cask.cdap.gateway.handlers;
 
 import co.cask.cdap.api.ProgramSpecification;
 import co.cask.cdap.api.data.stream.StreamSpecification;
-import co.cask.cdap.api.flow.FlowSpecification;
-import co.cask.cdap.api.flow.FlowletConnection;
-import co.cask.cdap.api.metrics.MetricDeleteQuery;
-import co.cask.cdap.api.metrics.MetricStore;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
-import co.cask.cdap.api.workflow.WorkflowSpecification;
 import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.deploy.Manager;
 import co.cask.cdap.app.deploy.ManagerFactory;
-import co.cask.cdap.app.program.Programs;
 import co.cask.cdap.app.runtime.ProgramController;
 import co.cask.cdap.app.runtime.ProgramRuntimeService;
 import co.cask.cdap.app.store.Store;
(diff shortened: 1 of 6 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -216,7 +186,13 @@ public void deleteApp(HttpRequest request, HttpResponder responder,
                         @PathParam("app-id") final String appId) throws Exception {
 
     Id.Application id = Id.Application.from(namespaceId, appId);
-    AppFabricServiceStatus appStatus = removeApplication(id);
+    // Deletion of a particular application is not allowed if that application is used by an adapter
+    AppFabricServiceStatus appStatus;
+    if (!adapterService.canDeleteApp(id)) {
+      appStatus = AbstractAppFabricHttpHandler.AppFabricServiceStatus.ADAPTER_CONFLICT;
+    } else {
+      appStatus= applicationLifecycleService.removeApplication(id);
```
chtyim: space before '='. Does it pass checkstyle?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -93,7 +93,7 @@
   /**
    * Class to represent status of programs.
    */
-  protected static final class AppFabricServiceStatus {
+  public static final class AppFabricServiceStatus {
```
chtyim: Why? If it is needed to be used across multiple classes, turn it into a top level class (which seems unlikely, since this class uses HttpResponseStatus inside, which is very http handler specific).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -307,12 +310,29 @@ public void removeAdapter(Id.Namespace namespace, String adapterName)
     throws AdapterNotFoundException, CannotBeDeletedException {
 
     AdapterStatus adapterStatus = getAdapterStatus(namespace, adapterName);
+    AdapterSpecification adapterSpec = getAdapter(namespace, adapterName);
```
chtyim: Can you add a TODO. The logic is not transactional and there can be race that one thread is creating and the other thread is deleting.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
rohitsinha54: But now the ApplicationLifecycleService will return these codes to Handler that is why I made it public.

chtyim:
Done review.

rohitsinha54:
Working on fixing the return codes properly.

rohitsinha54:
@chtyim : Fixed comments. Please have a look

chtyim:
LGTM

sreevatsanraman:
Test failures. This PR needs more work.

rohitsinha54:
Against 3.0 : https://github.com/caskdata/cdap/pull/2256

PR #2222, opened 2015-04-22
Set the templates dir path for unit tests

gokulavasan:


gokulavasan:
Will use TEMP_DIR instead of fixed "plugins"

cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
```diff
@@ -97,6 +97,11 @@
     public static final String MAPREDUCE_JOB_CLIENT_CONNECT_MAX_RETRIES = "mapreduce.jobclient.connect.max.retries";
 
     /**
+     * Default Template Dir.
+     */
+    public static final String DEFAULT_TEMPLATE_DIR = "plugins";
```
albertshau: this isn't the default, the default is in cdap-default.xml right?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/adapter/PluginTest.java
```diff
@@ -91,6 +91,7 @@ public static void setup() throws IOException, ClassNotFoundException {
 
     cConf = CConfiguration.create();
     cConf.set(Constants.CFG_LOCAL_DATA_DIR, TMP_FOLDER.newFolder().getAbsolutePath());
+    cConf.set(Constants.AppFabric.APP_TEMPLATE_DIR, TMP_FOLDER.newFolder().getAbsolutePath());
```
albertshau: this should have "plugins" too right

albertshau: o nevermind i see ignore

albertshau:
lgtm

PR #2225, opened 2015-04-22
Fixed issue with running from IDE

nitinmotgi:
There was issue with running standalone from IDE due to the path to UI. So, in this PR this issue for resolved. 

Test
- [x] Standalone
- [x] From IDE.

### DIFF
--- a/cdap-standalone/src/main/java/co/cask/cdap/UserInterfaceService.java
+++ b/cdap-standalone/src/main/java/co/cask/cdap/UserInterfaceService.java
@@ -56,10 +56,13 @@ final class UserInterfaceService extends AbstractExecutionThreadService {
     // When run from IDE, the base is "ui".
     File base = new File("ui");
     if (!base.isDirectory()) {
+      base = new File("cdap-ui");
       // It's ok as the path might get pass from StandaloneMain
       LOG.warn("Unable to determine UI directory");
+    } else {
+      base = new File(base, UI_VERSION);
     }
-    UI = new File(new File(base, UI_VERSION), "server.js").getAbsolutePath();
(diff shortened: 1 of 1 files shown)


abaranau:
checkboxes. sexy

abaranau:
it's a guice error - any way to test injection at least

cdap-standalone/src/main/java/co/cask/cdap/UserInterfaceService.java
```diff
@@ -53,13 +53,14 @@
   static final String UI;
   static {
     // Determine what's the path to the server.js, based on what's on the directory
-    // When run from IDE, the base is "ui".
+    // When run from IDE, the base is "cdap-ui" and from standalone it's ui.
     File base = new File("ui");
     if (!base.isDirectory()) {
```
abaranau: should you check if it exist?

sreevatsanraman:
LGTM :+1:

cdap-standalone/src/main/java/co/cask/cdap/UserInterfaceService.java
abaranau: Also, since you overwrite base variable anyways, should you just test
``` java
if (new File("ui").isDirectory)
```

abaranau: ignore last one

nitinmotgi: Not always - In else I use the base
```
base = new File(base, UI_VERSION);
```

PR #2227, opened 2015-04-22
Copy Guava pkg private classes for Type validation

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-DT51


henry-cask:
I would recommend adding huge TODO to make sure we remember to remove this when we move to Guava 15 since we move so fast this could be easily be forgotten later.

gokulavasan:
@henry-cask Created JIRA : https://issues.cask.co/browse/CDAP-2243

sreevatsanraman:
LGTM :+1:

PR #2230, opened 2015-04-22
Feature/skip processing records meta table unavailable

gsps1:
when the metaTable is unavailable we still go ahead and try to consume metrics records and we will eventually fail when we try to write the consumer offset to the kafka metaTable. 

Instead we should return if we cannot get the metaTable and not continue consuming the metrics records as this will cause those records to be processed again when the metaTable becomes available.

https://issues.cask.co/browse/CDAP-1710


cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/KafkaMetricsProcessorService.java
```diff
@@ -133,6 +133,12 @@ private void subscribe() {
     KafkaConsumer.Preparer preparer = kafkaClient.getConsumer().prepare();
 
     String topic = topicPrefix;
+
+    if (getMetaTable() == null) {
```
abaranau: there's a possible race - hbase can become unaccessible after that..

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/KafkaMetricsProcessorService.java
```diff
@@ -149,11 +155,6 @@ private void subscribe() {
   private long getOffset(String topic, int partition) {
     LOG.info("Retrieve offset for topic: {}, partition: {}", topic, partition);
     try {
```
abaranau: move this try-catch up - catch error in caller and return before subscribing

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/KafkaMetricsProcessorService.java
```diff
@@ -133,33 +133,30 @@ private void subscribe() {
...
-      }
-      long offset = metaTable.get(new TopicPartition(topic, partition));
-      LOG.info("Offset for topic: {}, partition: {} is {}", topic, partition, offset);
-      return offset;
-    } catch (Exception e) {
-      LOG.error("Failed to get offset from meta table. Defaulting to beginning. {}", e.getMessage(), e);
-    }
-    return -1L;
+    KafkaConsumerMetaTable metaTable = getMetaTable();
+    long offset = metaTable.get(new TopicPartition(topic, partition));
```
abaranau: can be nullpointer on metaTable?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/KafkaMetricsProcessorService.java
```diff
@@ -133,33 +133,30 @@ private void subscribe() {
...
+    try {
+      for (int i : partitions) {
+        long offset = getOffset(topic, i);
+        if (offset >= 0) {
+          preparer.add(topic, i, offset);
+        } else {
+          preparer.addFromBeginning(topic, i);
+        }
       }
+    } catch (Exception e) {
```
abaranau: move to getOffset

abaranau: disccussing....

abaranau:
LGTM

PR #2232, opened 2015-04-22
Send Events to Stream without Flows

elfenheart:
screen shot 2015-04-22 at 4 45 20 pm

### DIFF
--- a/cdap-ui/app/features/streams/templates/detail.html
+++ b/cdap-ui/app/features/streams/templates/detail.html
@@ -16,6 +16,11 @@
       </button>
       <ul class="dropdown-menu" role="menu">
         <li>
+          <a href="" ng-click="sendEvents()">
+            Send Events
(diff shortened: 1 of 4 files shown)


cdap-ui/app/features/streams/templates/detail.html
```diff
@@ -16,6 +16,11 @@
       </button>
       <ul class="dropdown-menu" role="menu">
         <li>
+          <a href="" ng-click="sendEvents()">
```
anwar6953: Does it make sense to have this button on the Status page? Like you can just click on a stream on the flow page to ingest?

elfenheart: That needs to be discussed, but can be easily added.

nitinmotgi:
LGTM :+1: @ajainarayanan

PR #2233, opened 2015-04-22
(CDAP-1917) Adding template REST endpoints

chtyim:
- Supports listing of templates
- Supports getting of a template
- Supports listing of all plugin types
- Supports listing of all plugins of a particular type
- Supports listing details of a plugin


chtyim:
Tested in standalone, unit test yet to come.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ApplicationTemplateHandler.java
```diff
@@ -0,0 +1,231 @@
...
+   *
+   * @param templateId the template id to query for
+   */
+  @Path("/{template-id}")
+  @GET
+  public void getTemplate(HttpRequest request, HttpResponder responder,
+                          @PathParam("template-id") String templateId) throws NotFoundException {
+    ApplicationTemplateInfo templateInfo = adapterService.getApplicationTemplateInfo(templateId);
+    if (templateInfo == null) {
+      throw new NotFoundException(Id.ApplicationTemplate.from(templateId));
```
sreevatsanraman: should return a 404 and return.

sreevatsanraman:
LGTM :+1:

PR #2237, opened 2015-04-23
Rename AdapterSpecification to AdapterDefinition

chtyim:
- This is a step to have AdapterSpecification to be available in cdap-api
- The rename is needed so that the AdapterDefinition can be kept as an internal class
  - It has things like Id.Program, DatasetCreationSpec, etc that are internal
- A new AdapterSpecification class will be introduced in cdap-api, which the AdapterDefinition will be extended from.


chtyim:
This PR has no logic changes, just rename and reorder of imports.
- Ran checkstyle locally and all pass.

sreevatsanraman:
LGTM :+1:

PR #2238, opened 2015-04-23
Expose AdapterSpecification through AdapterContext.

chtyim:
- Define the AdapterSpecification interface in cdap-api
- Make AdapterDefinition to implement AdapterSpecification
- Also piggy back to fix a small bug in DefaultAdapterConfigurer for wrong check on plugin usage.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterSpecification.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterSpecification.java
@@ -37,22 +37,27 @@ public interface AdapterSpecification {
   String getName();
 
   /**
-   * Returns the description of the adapter.
+   * Returns the description of the adapter or {@code null} if there is no description.
    */
+  @Nullable
   String getDescription();
 
   /**
-   * Returns the string representation of the configuration used to create the adapter.
+   * Returns the string representation of the configuration used to create the adapter
+   * or {@code null} if there is no configuration.
    */
+  @Nullable
   String getConfigString();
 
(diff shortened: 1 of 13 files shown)


cdap-api/src/main/java/co/cask/cdap/api/templates/AdapterSpecification.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Returns the name of the adapter.
+   */
+  String getName();
+
+  /**
+   * Returns the description of the adapter.
+   */
+  String getDescription();
```
bdmogal: The description can never be null, right?

cdap-app-fabric/src/main/java/co/cask/cdap/templates/AdapterDefinition.java
```diff
@@ -145,6 +152,16 @@ public JsonElement getConfig() {
...
   }
 
+  @Override
+  public <T> T getConfig(Type configType) {
+    return GSON.fromJson(config, configType);
+  }
+
+  @Override
+  public String getConfigString() {
+    return config.toString();
```
bdmogal: This is the one that extends ``PluginConfig``, right? Its ``toString()`` returns a JSON?

chtyim: config is of type `JsonElement`, not the user defined class.

bdmogal:
:+1:

PR #2239, opened 2015-04-23
Fix program archive creation

chtyim:
- Copy all existing manifest from the user JAR into program jar
  - It is needed for adapter to run correctly since it relies on the Export-Package attribute


gokulavasan:
LGTM :+1:

PR #2248, opened 2015-04-23
Fix READMEs

awholegunch:
This fixes minor errors in the READMEs, both in the source, in the Standalone, and on the web at GitHub.


sreevatsanraman:
LGTM :+1:

PR #2249, opened 2015-04-23
Pass ProgramOptions to Worker/Service in inmemory mode

gokulavasan:
We were passing only userArguments that were passed in through ProgramRunner and were skipped systemArguments. Need to pass in the sysArguments (figured this out while passing Adapter Name/Adapter Spec).

Builds : http://builds.cask.co/browse/CDAP-DUT1628
JIRA : https://issues.cask.co/browse/CDAP-2250

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
@@ -18,7 +18,6 @@ package co.cask.cdap.internal.app;
 
 import co.cask.cdap.api.ProgramSpecification;
 import co.cask.cdap.app.program.Program;
-import co.cask.cdap.app.runtime.Arguments;
 import co.cask.cdap.app.runtime.ProgramController;
 import co.cask.cdap.app.runtime.ProgramOptions;
 import co.cask.cdap.app.runtime.ProgramRunner;
@@ -62,21 +61,21 @@ public abstract class AbstractInMemoryProgramRunner implements ProgramRunner {
    * @param name Name of the component
    * @param instances Number of instances to start
    * @param runId The runId
-   * @param userArguments User runtime arguments
+   * @param options System and User runtime arguments
    * @param components A Table for storing the resulting ProgramControllers
    * @param type Type of ProgramRunnerFactory
    */
-  protected void startComponent(Program program, String name, int instances, RunId runId, Arguments userArguments,
+  protected void startComponent(Program program, String name, int instances, RunId runId, ProgramOptions options,
                               Table<String, Integer, ProgramController> components, ProgramRunnerFactory.Type type) {
     for (int instanceId = 0; instanceId < instances; instanceId++) {
-      ProgramOptions componentOptions = createComponentOptions(name, instanceId, instances, runId, userArguments);
+      ProgramOptions componentOptions = createComponentOptions(name, instanceId, instances, runId, options);
       ProgramController controller = programRunnerFactory.create(type).run(program, componentOptions);
       components.put(name, instanceId, controller);
     }
   }
 
   protected abstract ProgramOptions createComponentOptions(String name, int instanceId, int instances, RunId runId,
-                                                           Arguments userArguments);
+                                                           ProgramOptions options);
 
   /**
    * ProgramController to manage multiple in-memory instances of a Program.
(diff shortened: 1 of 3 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
```diff
@@ -62,21 +61,21 @@ public AbstractInMemoryProgramRunner(ProgramRunnerFactory programRunnerFactory)
...
    * @param instances Number of instances to start
    * @param runId The runId
-   * @param userArguments User runtime arguments
+   * @param options System and User runtime arguments
    * @param components A Table for storing the resulting ProgramControllers
    * @param type Type of ProgramRunnerFactory
    */
-  protected void startComponent(Program program, String name, int instances, RunId runId, Arguments userArguments,
+  protected void startComponent(Program program, String name, int instances, RunId runId, ProgramOptions options,
                               Table<String, Integer, ProgramController> components, ProgramRunnerFactory.Type type) {
```
bdmogal: Indentation seems off

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/InMemoryServiceProgramRunner.java
```diff
@@ -123,11 +122,13 @@ public ProgramController run(Program program, ProgramOptions options) {
...
   @Override
   public ProgramOptions createComponentOptions(String name, int instanceId, int instances, RunId runId,
-                                               Arguments userArguments) {
-    Map<String, String> options = ImmutableMap.of(ProgramOptionConstants.INSTANCE_ID, Integer.toString(instanceId),
-                                                  ProgramOptionConstants.INSTANCES, Integer.toString(instances),
-                                                  ProgramOptionConstants.RUN_ID, runId.getId(),
-                                                  ProgramOptionConstants.HOST, host);
-    return new SimpleProgramOptions(name, new BasicArguments(options), userArguments);
+                                               ProgramOptions options) {
+    Map<String, String> systemOptions = Maps.newHashMap();
```
bdmogal: I don't think this should change anything, but to keep the behavior the same as before, can we use ``ImmutableMap.Builder`` here instead of a ``newHashmap()``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/worker/InMemoryWorkerRunner.java
```diff
@@ -62,11 +61,13 @@
 
   @Override
   protected ProgramOptions createComponentOptions(String name, int instanceId, int instances, RunId runId,
-                                                  Arguments userArguments) {
-    Map<String, String> options = ImmutableMap.of(ProgramOptionConstants.INSTANCE_ID, Integer.toString(instanceId),
-                                                  ProgramOptionConstants.INSTANCES, Integer.toString(instances),
-                                                  ProgramOptionConstants.RUN_ID, runId.getId());
-    return new SimpleProgramOptions(name, new BasicArguments(options), userArguments);
+                                                  ProgramOptions options) {
+    Map<String, String> systemOptions = Maps.newHashMap();
```
bdmogal: same comment as above

bdmogal:
Minor comments, LGTM otherwise

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
sreevatsanraman: Fixe.d

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/InMemoryServiceProgramRunner.java
sreevatsanraman: Not fixing it. Since we cannot override instances otherwise.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/worker/InMemoryWorkerRunner.java
sreevatsanraman: Not fixing it. Since we cannot override instances otherwise.

sreevatsanraman:
Build passed. Merging.

PR #2254, opened 2015-04-23
Enhancement to allow external plugin to omit the properties field

chtyim:
- Help configuring external plugin easier.


sreevatsanraman:
LGTM :+1:

PR #2255, opened 2015-04-23
Sample Plugin Source and Sink using Plugin support

gokulavasan:
Changes to pom files and modified Twitter Source, Stream Sink to use plugins framework. Tested manually.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
@@ -44,6 +44,7 @@
       <groupId>co.cask.cdap</groupId>
       <artifactId>cdap-etl-lib</artifactId>
       <version>${project.version}</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>co.cask.cdap</groupId>
@@ -104,7 +105,7 @@
               <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
               <Embed-Transitive>true</Embed-Transitive>
               <Embed-Directory>lib</Embed-Directory>
-              <Export-Package>co.cask.cdap.etl.templates.api.*</Export-Package>
+              <Export-Package>co.cask.cdap.templates.etl.api.*</Export-Package>
             </instructions>
           </configuration>
           <executions>
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
@@ -27,7 +27,6 @@ import co.cask.cdap.api.templates.plugins.PluginConfig;
 import co.cask.cdap.templates.etl.api.PipelineConfigurer;
 import co.cask.cdap.templates.etl.api.config.ETLStage;
 import co.cask.cdap.templates.etl.api.realtime.DataWriter;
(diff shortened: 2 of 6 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -76,12 +87,6 @@ public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineC
   @Override
   public void initialize(RealtimeContext context) throws Exception {
```
bdmogal: no need to override now, right?

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
```diff
@@ -44,6 +44,7 @@
       <groupId>co.cask.cdap</groupId>
       <artifactId>cdap-etl-lib</artifactId>
       <version>${project.version}</version>
+      <scope>test</scope>
```
chtyim: This is temporary, right? When we have the test framework supports adapter and plugin then we can remove this dependency entirely, right?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sinks/StreamSink.java
```diff
@@ -36,33 +38,42 @@
...
-                                        false));
+  private static final String NAME_DESC = "The name of the stream to output to. Must be a valid stream name. " +
+    "The stream will be created if it does not exist.";
+  private static final String DATA_FIELD_DESC = "Name of the field in the record that contains the data to be " +
+    "written to the specified stream. The data could be in binary format as a byte array or a ByteBuffer. " +
+    "It can also be a String. If unspecified, the 'data' key is used.";
+  private static final String HEADERS_FIELD_DESC = "Name of the field in the record that contains headers. " +
+    "Headers are presumed to be a map of string to string.";
+
+  private StreamConfig streamConfig;
```
chtyim: You can have this field `final` and have a constructor `StreamSink(StreamConfig config)` (you might need to turn `StreamConfig` class public for that.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/TwitterSource.java
```diff
@@ -68,19 +72,26 @@
...
-  @Override
-  public void configure(StageConfigurer configurer) {
-    configurer.setName(TwitterStreamSource.class.getSimpleName());
-    configurer.setDescription("Twitter Realtime Source");
-    configurer.addProperty(new Property(CONSUMER_KEY, "Consumer Key", true));
-    configurer.addProperty(new Property(CONSUMER_SECRET, "Consumer Secret", true));
-    configurer.addProperty(new Property(ACCESS_TOKEN, "Access Token", true));
-    configurer.addProperty(new Property(ACCESS_SECRET, "Access Token Secret", true));
+
+  private TwitterConfig twitterConfig;
```
chtyim: Same applies here

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
gokulavasan: Yes that is right.

bdmogal:
:+1:

PR #2264, opened 2015-04-23
(CDAP-332) Added a REST endpoint for deleting a stream

chtyim:
- Added the DELETE /streams/{stream-id} endpoint
- Includes some code cleanup to unify StreamMetaStore usage
- For stream not found, throw the NotFoundException to simplify code.
- Implements delete by moving the stream directory into a special .deleted directory
  - Stream file janitor is responsible to cleanup that .delete directory
  - Adjusted code that do raw list of stream directories to one filterd out special directory


chtyim:
Open against release for now. If we don't want this to go to 3.0, then I can reopen against develop.

chtyim:
Reopen against develop

PR #2265, opened 2015-04-23
if no metrics are provided in metrics query, send bad request

gsps1:
bugfix: https://issues.cask.co/browse/CDAP-2157

we send badRequest when there are no metric query param for the /query endpoint, but missed that for the /query endpoint with context query param (deprecated). this solves that issue. 

build running : http://builds.cask.co/browse/CDAP-RBT218-1

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
@@ -138,7 +138,7 @@ public class MetricsHandler extends AuthenticatedHttpHandler {
     this.metricStore = metricStore;
   }
 
-  // Deprecate search with context param
+  // todo supporting 2.8 format - context param should be removed after deprecation (CDAP-1998)
   @POST
   @Path("/search")
   public void search(HttpRequest request, HttpResponder responder,
@@ -212,7 +212,7 @@ public class MetricsHandler extends AuthenticatedHttpHandler {
       tagsQuerying(request, responder, tags, metrics, groupBy);
     } else {
       // context querying support for 2.8 compatibility.
-      contextQuerying(request, responder, context, metrics.get(0), groupBy.size() > 0 ? groupBy.get(0) : null);
+      contextQuerying(request, responder, context, metrics, groupBy.size() > 0 ? groupBy.get(0) : null);
     }
   }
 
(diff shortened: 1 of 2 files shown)


cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -270,11 +270,11 @@ private void tagsQuerying(HttpRequest request, HttpResponder responder, List<Str
   }
 
   private void contextQuerying(HttpRequest request, HttpResponder responder,
-                               String context, String metric, String groupBy) {
+                               String context, List<String> metrics, String groupBy) {
     try {
       List<String> groupByTags = parseGroupBy(groupBy);
       MetricQueryResult queryResult = executeQuery(request, parseTagValuesAsMap(context),
-                                                   groupByTags, ImmutableList.of(metric));
+                                                   groupByTags, metrics);
```
anew: what does executeQuery() do if metrics is null? It is noyt annotated as @Nullable, and neither are the corresponding parameters of methods called from there. Are you sure it is ok to send null?

anew: Or is it an empty list? Still would like to know how that makes it a 400

gsps1: its an empty list and in executeQuery() we check if the list is empty for metrics and throw IllegalArgumentException there which gets caught and thrown as BadRequest

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -374,8 +374,7 @@ private Integer getResolution(String resolution, Long start, Long end) {
 
   private MetricQueryResult executeQuery(QueryRequest queryRequest) {
     try {
-
-      if (queryRequest.getMetrics() == null || queryRequest.getMetrics().size() == 0) {
+      if (queryRequest.getMetrics().size() == 0) {
```
gsps1: @anew here.

anew: OK. IllegalArgumentException is not really a good indicator for a bad request, because many internal errors are also IllegalArgumentException (whenever Preconditions.checkArgument throws). But for now I will say it's ok.

anew:
LGTM

PR #2266, opened 2015-04-23
(CDAP-332) Added a REST endpoint for deleting a stream

chtyim:
- Added the DELETE /streams/{stream-id} endpoint
- Includes some code cleanup to unify StreamMetaStore usage
- For stream not found, throw the NotFoundException to simplify code.
- Implements delete by moving the stream directory into a special .deleted directory
  - Stream file janitor is responsible to cleanup that .delete directory
  - Adjusted code that do raw list of stream directories to one filterd out special directory

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
@@ -28,8 +28,10 @@ import co.cask.cdap.data2.util.TableId;
 import co.cask.cdap.proto.Id;
 import com.google.common.base.CharMatcher;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
 import org.apache.twill.filesystem.Location;
 
 import java.io.DataInput;
@@ -46,6 +48,10 @@ import java.util.concurrent.TimeUnit;
  * TODO: Usage of this class needs to be refactor, as some methods are temporary (e.g. encodeMap/decodeMap).
  */
 public final class StreamUtils {
+
+  // The directory name for storing stream files that are pending for deletion
+  // StreamId cannot have "." there, so it's safe that it won't clash with any stream name
+  private static final String DELETED = ".deleted";
 
   /**
    * Decode a map.
@@ -454,6 +460,30 @@ public final class StreamUtils {
     String namespace = namespaceDir.getName();
     String streamName = streamBaseLocation.getName();
     return Id.Stream.from(namespace, streamName);
+  }
+
+  /**
+   * Returns the location of the stream deleted location.
+   *
+   * @param streamRootLocation the root location that all streams go under
+   */
+  public static Location getDeletedLocation(Location streamRootLocation) throws IOException {
+    return streamRootLocation.append(DELETED);
+  }
+
+  /**
+   * Lists all stream locations under the given root.
+   *
+   * @param streamRootLocation the root location that all streams go under
+   */
+  public static Iterable<Location> listAllStreams(Location streamRootLocation) throws IOException {
+    return Iterables.filter(streamRootLocation.list(), new Predicate<Location>() {
+      @Override
+      public boolean apply(Location location) {
+        // Any directories started with "." is special system directory, which is not regular stream directory
+        return !location.getName().startsWith(".");
+      }
+    });
   }
 
   private StreamUtils() {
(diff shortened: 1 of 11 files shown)


anwar6953:
Relevant: https://issues.cask.co/browse/CDAP-2316
It's also an issue with truncate, though.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
```diff
@@ -47,6 +49,10 @@
  */
 public final class StreamUtils {
 
+  // The directory name for storing stream files that are pending for deletion
+  // StreamId cannot have "." there, so it's safe that it won't crash with any stream name
```
jwang47: Why would it crash with "." in the stream name?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamHandler.java
```diff
@@ -178,15 +173,12 @@ public void getInfo(HttpRequest request, HttpResponder responder,
...
 
-    if (streamMetaStore.streamExists(streamId)) {
-      StreamConfig streamConfig = streamAdmin.getConfig(streamId);
-      StreamProperties streamProperties = new StreamProperties(streamConfig.getTTL(), streamConfig.getFormat(),
-                                                               streamConfig.getNotificationThresholdMB());
-      responder.sendJson(HttpResponseStatus.OK, streamProperties, StreamProperties.class, GSON);
-    } else {
-      responder.sendStatus(HttpResponseStatus.NOT_FOUND);
-    }
+    StreamConfig streamConfig = streamAdmin.getConfig(streamId);
```
jwang47: Maybe `streamAdmin.getConfg()` be throwing `NotFoundException` instead of having to remember to do `checkStreamExists` before every call.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
anwar6953: I think he means 'clash'.

jwang47: That makes sense.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamHandler.java
chtyim: The `getConfig` call simply read the config.json file. It doesn't hit MDS to check for stream existence. It does, however, throw `IOException` if failed to read the file (including file not found, or other form of IOException).

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
chtyim: fixed the typo.

awholegunch:
Can we add this to /cdap/cdap-docs/reference-manual/source/http-restful-api/stream.rst ?

chtyim:
@awholegunch Doc will be in separate PR.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
```diff
@@ -47,6 +49,10 @@
  */
 public final class StreamUtils {
 
+  // The directory name for storing stream files that are pending for deletion
+  // StreamId cannot have "." there, so it's safe that it won't clash with any stream name
+  private static final String DELETED = ".deleted";
```
bdmogal: Should we call this ``.Trash``?

bdmogal:
@chtyim just one minor question. Otherwise, LGTM

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamUtils.java
chtyim: Does it matter?

chtyim: The call is to delete stream, hence I call it deleted

bdmogal: Doesn't really matter, other than people may be used to ``.Trash`` in hadoop.

chtyim: CDAP is also for people who are not familiar with Hadoop or don't want/need to deal with Hadoop :)

bdmogal: Fair point... :) :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1780-4

PR #2268, opened 2015-04-23

jwang47:
Same as https://github.com/caskdata/cdap/pull/2290 (merged).

Continued from https://github.com/caskdata/cdap/pull/2078

https://issues.cask.co/browse/CDAP-2105
http://builds.cask.co/browse/CDAP-RBT219

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueConsumer.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueConsumer.java
@@ -331,7 +331,7 @@ public abstract class AbstractQueueConsumer implements QueueConsumer, Transactio
         }
 
         // Row key is queue_name + writePointer + counter
-        long writePointer = Bytes.toLong(rowKey, queueRowPrefix.length, Longs.BYTES);
+        long writePointer = QueueEntryRow.getWritePointer(rowKey, queueRowPrefix.length);
 
         // If it is first row returned by the scanner and was written before the earliest in progress,
         // it's safe to advance scanStartRow to current row because nothing can be written before this row.
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
@@ -22,7 +22,9 @@ import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Scanner;
 import co.cask.cdap.api.dataset.table.Table;
 import co.cask.cdap.common.collect.AllCollector;
+import co.cask.cdap.common.collect.AllPairCollector;
 import co.cask.cdap.common.collect.Collector;
+import co.cask.cdap.common.collect.PairCollector;
 import co.cask.cdap.common.queue.QueueName;
 import co.cask.cdap.data2.queue.ConsumerConfig;
 import co.cask.cdap.data2.queue.ConsumerGroupConfig;
@@ -33,7 +35,10 @@ import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import com.google.common.primitives.Ints;
 import com.google.common.primitives.Longs;
@@ -53,7 +58,7 @@ import javax.annotation.Nullable;
 /**
  * Class for persisting consumer state per queue consumer.
  */
-final class HBaseConsumerStateStore extends AbstractDataset implements QueueConfigurer {
+public final class HBaseConsumerStateStore extends AbstractDataset implements QueueConfigurer {
 
   private static final Gson GSON = new Gson();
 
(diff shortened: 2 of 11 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/AbstractQueueConsumer.java
```diff
@@ -306,6 +306,11 @@ private boolean fetchFromCache(SortedMap<byte[], SimpleQueueEntry> entries, int
     return true;
   }
 
+  public static long getWritePointer(byte[] rowKey, int queueRowPrefixLenght) {
```
chtyim: Don't add random static method like this to non-util type class. Move this to `QueueEntryRow` as

`getEnqueueWritePointer(byte[] rowKey, byte[] queueRowPrefix)`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -316,6 +328,15 @@ public void configureGroups(Iterable<? extends ConsumerGroupConfig> groupConfigs
     return scanBarriers(groupId, new AllCollector<QueueBarrier>()).finish(new ArrayList<QueueBarrier>());
   }
 
+  /**
+   * Gets all barrier information for all groups. The information are sorted in the order of
+   * the barrier changes.
+   */
+  Multimap<Long, QueueBarrier> getAllBarriers() {
+    return scanBarriers(new AllPairCollector<Long, QueueBarrier>()).finishMultimap(
+      HashMultimap.<Long, QueueBarrier>create());
```
chtyim: Is `HashMutlimap` guarantee sorting order? Shouldn't you use `LinkedHashMultimap`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -447,6 +487,20 @@ private int getInstanceIdFromColumn(byte[] column) {
   }
 
   @Nullable
+  private Map<Long, QueueBarrier> decodeBarrierInfo(Row row) {
+    Map<Long, QueueBarrier> barrierInfo = Maps.newHashMap();
+    Map<byte[], byte[]> columns = row.getColumns();
+    for (byte[] columnKey : columns.keySet()) {
+      long groupId = Bytes.toLong(columnKey);
+      QueueBarrier barrier = decodeBarrierInfo(row, groupId);
```
chtyim: That's a bit wasteful to encode groupId again and use it to get from row again. You can add a new method:

```java
QueueBarrier decodeBarrierInfo(byte[] rowKey, byte[] columnValue);
```

and have both existing `decodeBarrierInfo` call that.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,309 @@
...
+
+  @Inject
+  public HBaseQueueDebugger(HBaseQueueAdmin queueAdmin,
+                            HBaseQueueClientFactory queueClientFactory,
+                            TransactionServiceClient txClient,
+                            ZKClientService zkClientService) {
+    this.queueAdmin = queueAdmin;
+    this.queueClientFactory = queueClientFactory;
+    this.txClient = txClient;
+    this.zkClientService = zkClientService;
```
chtyim: I don't see the zkClient is used anywhere. What is it for?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,309 @@
...
+    stateStore.startTx(tx);
+
+    Multimap<Long, QueueBarrier> barriers = stateStore.getAllBarriers();
+    LOG.info("Got {} barriers", barriers.size());
+
+    QueueStatistics stats = new QueueStatistics();
+    int currentSection = 1;
+    for (Long groupId : barriers.keySet()) {
+      LOG.info("Scanning barriers for group {}", groupId);
+      Collection<QueueBarrier> groupBarriers = barriers.get(groupId);
```
chtyim: You can loop over `barries.entries()`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,309 @@
...
+      Iterator<QueueBarrier> barrierIterator = groupBarriers.iterator();
+
+      QueueBarrier previous;
+      QueueBarrier current = null;
+      while (barrierIterator.hasNext()) {
```
chtyim: You should be able to use a for-each loop instead of the iterator.

jwang47:
Going to address comments then reopen against hotfix/temp-fix-flow-change-instances branch.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: I remember something needed zkClientService to be started.. I can check what it is.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: Not sure how I do that and still maintain previous and current `QueueBarrier` cleanly. Seems I'd have to check every time if the `groupId` changed and do some logic to reset previous and current `QueueBarrier`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: For-each loop is possible, but I didn't like it b/c I'm handling the last barrier outside the for-loop, like so:

```

      LOG.info("Scanning section {}/{}...", currentSection, barriers.size());
      stats = QueueStatistics.of(stats, scanQueue(tx, queueName, lastBarrier, null));
      LOG.info("Scanning complete", currentSection, barriers.size());
```

jwang47: Though maybe `Multimap`/`LinkedMultimap` is the wrong class to use here, since I need the entries to be ordered per key. Maybe `Map<Long, List>` would be better?

chtyim: You don't need order by key (group ID), you just need to maintain the ordering of the barriers with the same order of insertion (scanning from HBase gives you from earliest to latest barrier).

jwang47:
Reopened against hotfix/temp-fix-flow-change-instances branch, squashed commits: https://github.com/caskdata/cdap/pull/2290 (you can still review how I addressed comments in this PR)

jwang47:
Merging after build passes.

PR #2269, opened 2015-04-23
add programs list on app metadata page

elfenheart:
Add programs list under Apps Metadata page

screen shot 2015-04-23 at 2 04 12 pm


nitinmotgi:
LGTM :+1:

PR #2270, opened 2015-04-23
Fix/ui styling system health default color

astewes:


anwar6953:
seems you're branched off develop
"Bump version to 3.1.0-SNAPSHOT on develop"

PR #2272, opened 2015-04-23
Add routing rule for dataset usage

jwang47:
http://builds.cask.co/browse/CDAP-RBT221


cdap-gateway/src/main/java/co/cask/cdap/gateway/router/RouterPathLookup.java
```diff
@@ -123,6 +123,10 @@ private String getV3RoutingService(String [] uriParts, AllowedMethod requestMeth
       // /v3/namespaces/{namespace-id}/data/datasets/{name}/workers
       // /v3/namespaces/{namespace-id}/data/datasets/{name}/mapreduce
       return Constants.Service.APP_FABRIC_HTTP;
+    } else if (
```
chtyim: The convention should be:

```java
if (matches(...)
 || matches(...)) {
}
```

sreevatsanraman:
LGTM :+1:

PR #2273, opened 2015-04-23
Fix/ui styling system health defaults

astewes:


cdap-ui/app/features/overview/overview.html
```diff
@@ -101,29 +66,30 @@ <h3 class="text-uppercase"> Accelerate... </h3>
 
     <div class="col-xs-12 col-lg-8">
```
ajainarayanan: Moving the ```text-center``` here would be better since after col-lg-8 we are directly going to col-xs-12 and the status symbol appears centered but the charts seem left aligned. If we have text-center here then it will make everything center aligned which would be good for all screen sizes.

ajainarayanan:
One minor comment apart from that :+1: LGTM

cdap-ui/app/features/overview/overview.html
ajainarayanan: For now this is not done. Moving this to another PR since the layout becomes un-necessarily complicated to be addressed in one PR.

PR #2274, opened 2015-04-23
Feature/ui styling dashboard widgets

astewes:


ajainarayanan:
@astewes dashboard.html has list element for user dashboards. The li element should have a class 'dropdown-toggle' that is being checked by the ```activeTabClick``` function. So if that is not there then we end up getting two different menus. This is happening because of the place where the drop down menu gets added. We have our styles namespaced to ```.cdap-subnav-end + ul.slanted-tabs``` classes but since we click on the edge of the tab header it is getting added outside of the tabs so we get the default menu.. 
dashboard

ajainarayanan:
Otherwise looks Lookxy!! Awesome!

ajainarayanan:
:+1: LGTM!

PR #2275, opened 2015-04-23
Converting Identity Transform

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-RT0-2

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
@@ -190,7 +190,7 @@
         <extensions>true</extensions>
         <configuration>
           <instructions>
-            <Export-Package>co.cask.cdap.templates.etl.realtime.*;co.cask.cdap.templates.etl.batch.*;co.cask.cdap.templates.etl.transforms</Export-Package>
+            <Export-Package>co.cask.cdap.templates.etl.realtime.*;co.cask.cdap.templates.etl.batch.*;co.cask.cdap.templates.etl.transforms.*</Export-Package>
             <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
             <Embed-Transitive>true</Embed-Transitive>
             <Embed-Directory>lib</Embed-Directory>
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
@@ -190,7 +190,7 @@
         <extensions>true</extensions>
         <configuration>
           <instructions>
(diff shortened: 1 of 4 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
```diff
@@ -190,7 +190,7 @@
         <extensions>true</extensions>
         <configuration>
           <instructions>
-            <Export-Package>co.cask.cdap.templates.etl.realtime.*;co.cask.cdap.templates.etl.batch.*</Export-Package>
+            <Export-Package>co.cask.cdap.templates.etl.realtime.*;co.cask.cdap.templates.etl.batch.*;co.cask.cdap.templates.etl.transforms</Export-Package>
```
chtyim: No need to have `.*`?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/IdentityTransform.java
```diff
@@ -16,23 +16,23 @@
...
 import co.cask.cdap.templates.etl.api.TransformStage;
 
 /**
- * Simple Identity Transform for testing.
+ * Simple Identity Transform.
  *
  * @param <T> any type
  */
+@Plugin(type = "transform")
+@Name("IdentityTransform")
```
chtyim: Actually, can we just call it `Identity`, because the plugin already tells it's "transform" ? What should be the naming convention?

chtyim:
The two questions.

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
gokulavasan: Oh yes. Will add.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/transforms/IdentityTransform.java
gokulavasan: But we have TwitterSource, etc. Hmm we have to think about naming convention.

bdmogal: Should we enforce a name? Can it default to ``Class.getSimpleName()``?

chtyim: Default is class name, not simple name.

chtyim:
LGTM

gokulavasan:
Build passed : http://builds.cask.co/browse/CDAP-RT0-4

PR #2276, opened 2015-04-23
Fix/ui styling move icons

astewes:


cdap-ui/app/features/admin/admin.less
```diff
@@ -80,12 +80,14 @@ body.theme-cdap.state-admin {
     }
     h2:first-child { margin-top: 0; }
     .panel-default { margin-top: 18px; }
-    td > span {
+    td > span[class^="status-"] {
```
ajainarayanan: Can I know where this rule is being used? I am not able to find a usage for this.

ajainarayanan:
Apart from one minor comment LGTM :+1:

PR #2279, opened 2015-04-23
Feature/docs build apply 28

duddun:


cdap-docs/_common/common-build.sh
```diff
@@ -304,6 +312,7 @@ function test_an_include_diff() {
 
 function build_standalone() {
   cd $PROJECT_PATH
+  set_mvn_environment
   MAVEN_OPTS="-Xmx512m" mvn clean package -DskipTests -P examples -pl cdap-examples -am -amd && MAVEN_OPTS="-Xmx512m" mvn package -pl cdap-standalone -am -DskipTests -P dist,release
```
awholegunch: Looks like this has changed: see BUILD.md in release/3.0 for the latest.

cdap-docs/admin-manual/source/conf.py
```diff
@@ -10,6 +10,10 @@
 sys.path.insert(0, os.path.abspath('../../_common'))
 from common_conf import * 
 
+print_sdk_version()
+
+print get_sdk_version()
+
```
awholegunch: These two were de-bug lines and can be removed.

cdap-docs/examples-manual/build.sh
```diff
@@ -62,7 +62,13 @@ function download_file() {
...
+
+  if [ "$OSTYPE" == "darwin"* ]; then
+    local new_md5_hash=`md5 -q $target`
+  else
+    local new_md5_hash=`md5sum $target | awk '{print $1}'`
+  fi
+
   if [ "x$md5_hash" != "x$new_md5_hash" ]; then
     echo -e "$WARNING MD5 Hash for $file_name has changed! Compare files and update hash!"  
     echo "Old md5_hash: $md5_hash New md5_hash: $new_md5_hash"
```
awholegunch: What's really nice is to have this line as:
```
echo -e "Old MD5 Hash: ${md5_hash} New MD5 Hash: ${RED}${BOLD}${new_md5_hash}${NC}"
```
then the new MD5 hash is in red and easy to find and copy.

cdap-docs/vars
```diff
@@ -0,0 +1,11 @@
...
+COMMON="_common"
+COMMON_HTACCESS="$COMMON/htaccess"
+COMMON_IMAGES="$COMMON/_images"
+COMMON_SOURCE="$COMMON/_source"
+COMMON_CONF_PY="$COMMON/common_conf.py"
+COMMON_HIGHLEVEL_PY="$COMMON/highlevel_conf.py"
+COMMON_PLACEHOLDER="$COMMON/_source/placeholder_index._rst"
+HTACCESS="htaccess"
+
+MANUALS="admin-manual developers-manual examples-manual integrations reference-manual"
```
awholegunch: One problem is that this line needs to be kept in sync with the _common/common_conf.py.
Should that be the source of it? Is that easy/possible?

cdap-docs/_common/common_conf.py
```diff
@@ -36,15 +36,24 @@
...
+# Python 2.7 commands
+#         grep_version_cmd = "grep '<version>' ../../../pom.xml | awk 'NR==1;START{print $1}'"
+#         print "grep_version_cmd: %s" % grep_version_cmd
+#         full_version_temp = subprocess.check_output(grep_version_cmd, shell=True)
+# Python 2.6 commands
+        p1 = subprocess.Popen(["grep" , "<version>", "../../../pom.xml" ], stdout=subprocess.PIPE)
+        p2 = subprocess.Popen(["awk", "NR==1;START{print $1}"], stdin=p1.stdout, stdout=subprocess.PIPE)
+        full_version_temp = p2.communicate()[0]
+#
+	full_version = full_version_temp.strip().replace("<version>", "").replace("</version>", "")
```
awholegunch: This line is not lined up with the line above. I think it needs a leading space.

cdap-docs/_common/common_conf.py
```diff
@@ -36,15 +36,24 @@
...
-        full_version = subprocess.check_output(grep_version_cmd, shell=True).strip().replace("<version>", "").replace("</version>", "")
+# Python 2.7 commands
+#         grep_version_cmd = "grep '<version>' ../../../pom.xml | awk 'NR==1;START{print $1}'"
+#         print "grep_version_cmd: %s" % grep_version_cmd
+#         full_version_temp = subprocess.check_output(grep_version_cmd, shell=True)
+# Python 2.6 commands
+        p1 = subprocess.Popen(["grep" , "<version>", "../../../pom.xml" ], stdout=subprocess.PIPE)
+        p2 = subprocess.Popen(["awk", "NR==1;START{print $1}"], stdin=p1.stdout, stdout=subprocess.PIPE)
+        full_version_temp = p2.communicate()[0]
+#
```
awholegunch: Maybe change to
```
# Python 2.6 commands end
```
to make it clearer?

cdap-docs/build.sh
```diff
@@ -26,35 +26,26 @@
...
-COMMON_IMAGES="$COMMON/_images"
-COMMON_SOURCE="$COMMON/_source"
-COMMON_CONF_PY="$COMMON/common_conf.py"
-COMMON_HIGHLEVEL_PY="$COMMON/highlevel_conf.py"
-COMMON_PLACEHOLDER="$COMMON/_source/placeholder_index._rst"
-HTACCESS="htaccess"
-
 ARG_1="$1"
 ARG_2="$2"
 ARG_3="$3"
```
awholegunch: Do these need to be ${1}, ${2}, ${3}, etc?

cdap-docs/build.sh
```diff
@@ -26,35 +26,26 @@
...
 
 function set_project_path() {
   if [ "x$ARG_2" == "x" ]; then
```
awholegunch:
```
"x${ARG_2}" == "x"
```

cdap-docs/build.sh
```diff
@@ -135,18 +131,18 @@ function build_docs_outer_level() {
   copy_source examples-manual     "Examples, Guides, and Tutorials"
```
awholegunch: Should this be changed to a list, similar to line 153? The titles for these manuals are define in the common conf.py.

cdap-docs/build.sh
```diff
@@ -154,41 +150,43 @@ function copy_docs_lower_level() {
...
   mv $SCRIPT_PATH/$BUILD/*.zip $SCRIPT_PATH/$BUILD_TEMP
+  echo "========================================================================="
   echo "Building Web Docs."
-  ./build.sh docs-web $ARG_2 $ARG_3
+  echo "========================================================================="
+  run_command docs-web-part $ARG_2 $ARG_3
   echo "Replacing GitHub Docs."
   mv $SCRIPT_PATH/$BUILD_TEMP/*.zip $SCRIPT_PATH/$BUILD
   rm -rf $SCRIPT_PATH/$BUILD_TEMP
   bell
```
awholegunch: You might want to make this an option. I have had complaints about this from..someone. Other people find it annoying. But it's very handy. So, put it in a function, and use an environment variable?

awholegunch:
After you've had a chance to look at the comments and fix a few things, I can test under CentOS 6, if you'd like.

cdap-docs/_common/common_conf.py
awholegunch: I would expect this to break.

cdap-docs/vars
duddun: I am thinking there should be one source from various scripts that need it.  common_conf.py is a script, so it makes less sense to me to have it there (a bash script sourcing a python script for variables seems odd to me).

cdap-docs/_common/common_conf.py
duddun: I am not sure why, but in the script itself it is lined up correctly (and the build does not break there, from what I have seen).

cdap-docs/build.sh
duddun: Let us revisit this after 3.0 is released.

awholegunch:
I have fixed the build so that it works on Mac OS X. I also added a missing manual ("introductions"). Tested under both Mac OS X and CentOS 6: passed build-docs. Now testing javadocs.

awholegunch:
The javadocs were using the wrong spreadsheet when run under JDK 1.7. Fixed. Now testing under Mac OS X.

awholegunch:
The Javadocs now build correctly under 1.7 on both Platforms. However, only on the Mac OS X is there any guards to make sure that the Javadocs are built under 1.7. If they are built on CentOS 6 under JDK 1.6, they won't be correct.

awholegunch:
It looks ok. But can we merge it after 5pm Monday, in case anything breaks?

duddun:
Per JJ -- merging now.

PR #2280, opened 2015-04-23
draw leaf shape and scale correctly

nistala:
Draw leaf shape for event counts.
screenshot from 2015-04-23 15 56 33


sreevatsanraman:
Open against release/3.0 branch

PR #2281, opened 2015-04-23
CDAP-2257 Port KVTableSink to use Plugin framework

rohitsinha54:
Manual Testing Pending.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java
@@ -16,38 +16,64 @@
 
 package co.cask.cdap.templates.etl.batch.sinks;
 
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.dataset.lib.KeyValueTable;
+import co.cask.cdap.api.templates.plugins.PluginConfig;
 import co.cask.cdap.templates.etl.api.Emitter;
-import co.cask.cdap.templates.etl.api.Property;
-import co.cask.cdap.templates.etl.api.StageConfigurer;
 import co.cask.cdap.templates.etl.api.config.ETLStage;
 import co.cask.cdap.templates.etl.common.Properties;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Strings;
 
 import java.nio.ByteBuffer;
+import javax.annotation.Nullable;
 
 /**
  * CDAP Table Dataset Batch Sink.
  */
+@Plugin(type = "sink")
+@Name("KVTableSink")
+@Description("CDAP Key Value Table Dataset Batch Sink")
 public class KVTableSink extends BatchWritableSink<StructuredRecord, byte[], byte[]> {
-  private String keyField;
-  private String valueField;
 
-  @Override
-  public void configure(StageConfigurer configurer) {
-    configurer.setName(getClass().getSimpleName());
-    configurer.setDescription("CDAP Key Value Table Dataset Batch Sink");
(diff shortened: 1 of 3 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java
```diff
@@ -16,63 +16,77 @@
...
-    configurer.setDescription("CDAP Key Value Table Dataset Batch Sink");
-    configurer.addProperty(new Property(NAME, "Dataset Name", true));
-    configurer.addProperty(new Property(
-      Properties.KeyValueTable.KEY_FIELD,
-      "The name of the field to use as the key. Its type must be bytes. Defaults to 'key'.", false));
-    configurer.addProperty(new Property(
-      Properties.KeyValueTable.VALUE_FIELD,
-      "The name of the field to use as the value. Its type must be bytes. Defaults to 'value'.", false));
+  private static final String NAME_DESC = "Dataset Name";
+  private static final String KEY_FIELD_DESC = "The name of the field to use as the key. Its type must be bytes. " +
```
chtyim: Type is `bytes`?? The type is `string`, which get converted to UTF-8 encoded bytes, right?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java
```diff
@@ -16,63 +16,77 @@
...
+    "Defaults to 'key'.";
+  private static final String VALUE_FIELD_DESC = "The name of the field to use as the value. Its type must be bytes. " +
```
chtyim: Same here.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java
```diff
@@ -16,63 +16,77 @@
...
+  /**
+   * Config class for KVTableSink
+   */
+  public static class KVTableConfig extends PluginConfig {
+    @Description(NAME_DESC)
+    private String name;
+
+    @Name(Properties.KeyValueTable.KEY_FIELD)
+    @Description(KEY_FIELD_DESC)
+    @Nullable
```
chtyim: Why optional? It seems like it is not checked in the transform method that it can be null?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java
```diff
@@ -16,63 +16,77 @@
...
-import com.google.common.base.Strings;
 
 import java.nio.ByteBuffer;
+import javax.annotation.Nullable;
 
 /**
  * CDAP Table Dataset Batch Sink.
  */
+@Plugin(type = "sink")
+@Name("KVTableSink")
```
gokulavasan: KVTableSink ==> KVTable

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java
rohitsinha54: I referred to this line: 
https://github.com/caskdata/cdap/blob/develop/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java#L60

Am I missing something here ?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java
rohitsinha54: Here it says false 
https://github.com/caskdata/cdap/blob/develop/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sinks/KVTableSink.java#L47
So I marked as that. Shouldn't it be ?

PR #2282, opened 2015-04-23
Feature/ui flow fixes

nistala:
Draw leaf shape for flows and scale correctly.
screenshot from 2015-04-23 15 56 33


elfenheart:
:+1: LGTM

PR #2283, opened 2015-04-23
CDAP-2025 Updated RunRecord of the Workflow and underlying MapReduce pro...

sagarkapare:
...grams.

Opening this PR against release branch.
Jira: https://issues.cask.co/browse/CDAP-2025

Summary of changes:
1. Passing the Workflow related information (Workflow name, Workflow RunId and node id of the program in the Workflow) to the MapReduceProgramRunner which uses this information to record and update the state for the MapReduce program and Workflow in the MDS.

2. Updated the RunRecord to include generic properties field which is of type Map. The value contained in the properties depends on the program for which the RunRecord is created.
e.g. For Workflow properties represents the map of node id to run id of the programs inside the Workflow. For MapReduce program which started as a part of the Workflow, properties currently represents map of single entry for the Workflow run id.

e.g. Assume that the `PurchaseHistoryWorkflow` has runs two MapReduce program node, each runs the same MapReduce program `PurchaseHistoryBuilder`.
Then the RunRecord for the Workflow would look -
```json
[
    {
        "end": 1429831538,
        "properties": {
            "0": "f3922dd4-ea0f-11e4-a768-c200d05e85ef",
            "1": "009b46b5-ea10-11e4-8e50-c200d05e85ef"
        },
        "runid": "f3807a93-ea0f-11e4-b6eb-c200d05e85ef",
        "start": 1429831495,
        "status": "COMPLETED"
    }
] 
```
For single run of the Workflow, there will be two RunRecords for the MapReduce program as -
```json
[
    {
        "end": 1429831538,
        "properties": {
            "workflowrunid": "f3807a93-ea0f-11e4-b6eb-c200d05e85ef"
        },
        "runid": "009b46b5-ea10-11e4-8e50-c200d05e85ef",
        "start": 1429831517,
        "status": "COMPLETED"
    },
    {
        "end": 1429831517,
        "properties": {
            "workflowrunid": "f3807a93-ea0f-11e4-b6eb-c200d05e85ef"
        },
        "runid": "f3922dd4-ea0f-11e4-a768-c200d05e85ef",
        "start": 1429831495,
        "status": "COMPLETED"
    }
] 
```


sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-RBT222-1

sagarkapare:
@anew @chtyim Please take a look. Thanks!

sagarkapare:
Build passes here - http://builds.cask.co/browse/CDAP-RBT222-2

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -175,7 +178,13 @@ public void starting() {
           startTimeInSeconds = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis());
         }
         String adapterName = adapterSpec == null ? null : adapterSpec.getName();
-        store.setStart(program.getId(), runId.getId(), startTimeInSeconds, adapterName, twillRunId);
+        if (workflowName == null) {
+          store.setStart(program.getId(), runId.getId(), startTimeInSeconds, adapterName, twillRunId);
+        } else {
+          // Program started by Workflow
+          store.setWorkflowProgramStart(program.getId(), runId.getId(), workflowName, workflowRunId, workflowNodeId,
+                                        startTimeInSeconds, twillRunId);
```
anew: adapterName is not needed here?

sagarkapare: `adapterName` is fetched from the `Workflow` run record. However, for consistency its better to pass it from here?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/ProgramWorkflowRunnerFactory.java
```diff
@@ -60,17 +60,18 @@ public ProgramWorkflowRunnerFactory(WorkflowSpecification workflowSpec, ProgramR
    * @param token      The {@link WorkflowToken}
    * @return the appropriate concrete implementation of {@link ProgramWorkflowRunner} for the program
    */
-  public ProgramWorkflowRunner getProgramWorkflowRunner(WorkflowActionSpecification actionSpec, WorkflowToken token) {
+  public ProgramWorkflowRunner getProgramWorkflowRunner(WorkflowActionSpecification actionSpec, WorkflowToken token,
```
anew: doesn't a workflow action spec contain the node id? If not, why not?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
```diff
@@ -548,4 +551,53 @@ private MDSKey getNamespaceKey(@Nullable String name) {
...
+    if (record == null) {
+      String msg = String.format("No meta found for associated Workflow %s run record %s, while recording run for the" +
+                                   " namespace %s app %s type %s program %s runid %s", workflow, workflowRunId,
+                                 program.getNamespaceId(), program.getApplicationId(), program.getType().name(),
+                                 program.getId(), programRunId);
+      LOG.error(msg);
+      throw new IllegalArgumentException(msg);
+    }
+
+    deleteAll(key);
```
anew: why delete all? please explain

cdap-proto/src/main/java/co/cask/cdap/proto/RunRecord.java
```diff
@@ -44,14 +46,25 @@
...
   private final String adapterName;
 
+  @Nullable
+  @SerializedName("properties")
+  private final Map<String, String> properties;
+
   public RunRecord(String pid, long startTs, @Nullable Long stopTs, ProgramRunStatus status,
-                   @Nullable String adapterName, @Nullable String twillRunId) {
+                   @Nullable String adapterName, @Nullable String twillRunId,
+                   @Nullable Map<String, String> properties) {
```
anew: It seems better to require that properties is not null, but rather an empty map if there is nothing there. That would be consistent with what we do for other similar parameters, such as RuntimeArguments or DatasetProperties.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
anew: Oh, ok. That make sense.

anew:
@sagarkapare the approach is good, I have only a few comments.

sagarkapare:
Thank you @anew for the review comments. Updating the PR as per suggestions.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
sagarkapare: Passing adapterName now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/ProgramWorkflowRunnerFactory.java
sagarkapare: `WorkflowActionSpecification` does not contain the node id. 

The hierarchy is, at the top level we have `WorkflowNode`, which contains `nodeId` and the `nodeType`. Action nodes are represented by the `WorkflowActionNode` class which inherits the `WorkflowNode` and include the `WorkflowActionSpecification` as well. 

Please let me know if nodeId should also be included in the action spec.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
sagarkapare: Even if the method is named as `deleteAll`, the key formed here will cause the deletion of the single `RunRecord`entry corresponding to the runId specified by the parameters.

cdap-proto/src/main/java/co/cask/cdap/proto/RunRecord.java
sagarkapare: Yeah. Makes sense. Fixed the code to set to the empty map now instead of null.

sagarkapare:
@anew Addressed comments and answered the questions. Please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
anew: But why delete it? You are gonna overwrite it in the next operation.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/ProgramWorkflowRunnerFactory.java
anew: I see. I was not aware of these details.

anew:
I still don't understand why the run record must be deleted before it can be overwritten. Other than that, LGTM.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
sagarkapare: Ohh I see what you are saying. Yes, we are not changing the key also, so we do not need `deleteAll`. Removed in the last commit.

sagarkapare:
Removed the redundant `deleteAll`. Build is running here http://builds.cask.co/browse/CDAP-RBT222-4. Will merge this once the build passes.

sagarkapare:
Un-related test case failed which is failing on release branch as well. Merging this PR.

PR #2285, opened 2015-04-24
[CDAP-2265] Convert StreamBatchSource into a plugin

bdmogal:
Jira: [CDAP-2265](https://issues.cask.co/browse/CDAP-2265)
Build: http://builds.cask.co/browse/CDAP-RT1

Tested that the plugin gets recognized. Need a batch sink to test end-to-end. Will do that after I convert DBSink or something else.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/StreamBatchSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/StreamBatchSource.java
@@ -16,6 +16,9 @@
 
 package co.cask.cdap.templates.etl.batch.sources;
 
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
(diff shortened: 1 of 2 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/StreamBatchSource.java
```diff
@@ -60,73 +68,59 @@
...
   public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineConfigurer) {
-    Config config = new Config(stageConfig.getProperties());
-    pipelineConfigurer.addStream(new Stream(config.name));
+    pipelineConfigurer.addStream(new Stream(streamBatchConfig.name));
   }
 
   @Override
   public void prepareJob(BatchSourceContext context) {
-    config = new Config(context.getPluginProperties().getProperties());
+    config = new Config(streamBatchConfig);
```
chtyim: You shouldn't need the extra `Config` object. Just do the conversion when needed since `prepareJob` only get called once per run.

chtyim:
LGTM

PR #2288, opened 2015-04-24
CDAP-1773: Added new Adapter CLI commands

jwang47:
- [ ] Test `AdapterClient.getLogs()` for template using a workflow once get workflow logs is implemented


cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
```diff
@@ -66,38 +68,26 @@ public String getCronEntry() {
   }
 
   @Override
-  public boolean equals(Object o) {
-    if (this == o) {
+  public int hashCode() {
+    return Objects.hashCode(name, description, cronEntry);
```
chtyim: Don't use guava in API. Please revert.

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
```diff
@@ -66,38 +68,26 @@ public String getCronEntry() {
...
-    result = 31 * result + cronEntry.hashCode();
-    return result;
+    final Schedule other = (Schedule) obj;
+    return Objects.equal(this.name, other.name) && Objects.equal(this.description, other.description) &&
+      Objects.equal(this.cronEntry, other.cronEntry);
   }
 
   @Override
   public String toString() {
-    StringBuilder sb = new StringBuilder("Schedule{");
```
chtyim: Revert the change.

cdap-api/src/main/java/co/cask/cdap/api/schedule/ScheduleSpecification.java
```diff
@@ -59,36 +60,26 @@ public Schedule getSchedule() {
   }
 
   @Override
-  public boolean equals(Object o) {
-    if (this == o) {
```
chtyim: Same. Revert.

cdap-proto/src/main/java/co/cask/cdap/proto/AdapterDetail.java
```diff
@@ -16,24 +16,28 @@
...
- * Adapter details that come back from get calls.
- * TODO: finalize what should be present here
+ * Represents an adapter returned for /adapters/{adapter-id}.
  */
-public final class AdapterDetail {
+public class AdapterDetail {
   private final String name;
   private final String description;
   private final String template;
+  private final Id.Program program;
```
chtyim: What does it look like for the serialized object? It seems like it'd be pretty nested and not good to be used as response of the REST endpoint. It looks something like this:

```
{
  "application": {
    "namespace": {
      "id": "ns"
    },
    "applicationId": "appId"
  },
  "type": "Worker",
  "id": "pid"
}
```

chtyim: May we should have a flat representation for program id in proto for all REST endpoints. E.g. Just

```java
class ProgramId {
  String namespace;
  String application;
  String type;
  String id;
}
```

jwang47: Yeah, that would be better.

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
```diff
@@ -67,33 +67,29 @@ public String getCronEntry() {
...
     Schedule schedule = (Schedule) o;
 
-    if (cronEntry.equals(schedule.cronEntry) && description.equals(schedule.description)
-      && name.equals(schedule.name)) {
-      return true;
-    }
-    return false;
+    if (cronEntry != null ? !cronEntry.equals(schedule.cronEntry) : schedule.cronEntry != null) return false;
+    if (description != null ? !description.equals(schedule.description) : schedule.description != null) return false;
+    if (name != null ? !name.equals(schedule.name) : schedule.name != null) return false;
```
chtyim: Isn't that would fail checkstyle? Can you just revert the change by checking out the old commit for the file?

cdap-api/src/main/java/co/cask/cdap/api/schedule/ScheduleSpecification.java
```diff
@@ -60,31 +60,29 @@ public Schedule getSchedule() {
...
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
 
     ScheduleSpecification that = (ScheduleSpecification) o;
-    if (program.equals(that.program) && properties.equals(that.properties) && schedule.equals(that.schedule)) {
-      return true;
-    }
-    return false;
+
+    if (program != null ? !program.equals(that.program) : that.program != null) return false;
```
chtyim: Same here. Checkout old commit to revert.

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
jwang47: Changed to use internal (in `cdap-api`) `Objects`. This change was needed b/c in the list adapters endpoint, some of the `Schedule` fields can be empty.

cdap-api/src/main/java/co/cask/cdap/api/schedule/ScheduleSpecification.java
jwang47: Changed to use internal (in `cdap-api`) `Objects`. This change was needed b/c in the list adapters endpoint, some of the `ScheduleSpecification` fields can be empty.

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
chtyim: IntelliJ can generate code that can deal with null. Can you not copy the `Objects` class?

cdap-api/src/main/java/co/cask/cdap/internal/Objects.java
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+package co.cask.cdap.internal;
+
+import java.util.Arrays;
+import javax.annotation.Nullable;
+
+/**
+ * Helper functions that can operate on any {@code Object}.
```
chtyim: Better not to copy from guava. This class will become not useful soon (Java 7 has java.uti.Objects).

PR #2290, opened 2015-04-24

jwang47:
Also opened here: https://github.com/caskdata/cdap/pull/2268

After installing CDAP on a cluster, you can run the tool using the following java command from `/opt/cdap/master`:

```
java -cp `hbase classpath`:"lib/*:../hbase-compat-0.98/lib/*":/etc/cdap/conf/* co.cask.cdap.data2.transaction.queue.hbase.HBaseQueueDebugger queue:///default/PurchaseHistory/PurchaseFlow/reader/queue
```

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/QueueEntryRow.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/QueueEntryRow.java
@@ -318,4 +318,12 @@ public class QueueEntryRow {
     byte state = stateBytes[Longs.BYTES + Ints.BYTES];
     return state == ConsumerEntryState.PROCESSED.getState();
   }
+
+  /**
+   * Gets the write pointer for a row.
+   */
+  public static long getWritePointer(byte[] rowKey, int queueRowPrefixLength) {
+    // Row key is queue_name + writePointer + counter
+    return Bytes.toLong(rowKey, queueRowPrefixLength, Longs.BYTES);
+  }
 }
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
@@ -22,7 +22,9 @@ import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Scanner;
 import co.cask.cdap.api.dataset.table.Table;
 import co.cask.cdap.common.collect.AllCollector;
+import co.cask.cdap.common.collect.AllPairCollector;
 import co.cask.cdap.common.collect.Collector;
+import co.cask.cdap.common.collect.PairCollector;
 import co.cask.cdap.common.queue.QueueName;
 import co.cask.cdap.data2.queue.ConsumerConfig;
 import co.cask.cdap.data2.queue.ConsumerGroupConfig;
@@ -35,6 +37,7 @@ import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import com.google.common.primitives.Ints;
@@ -55,7 +58,7 @@ import javax.annotation.Nullable;
 /**
  * Class for persisting consumer state per queue consumer.
  */
-final class HBaseConsumerStateStore extends AbstractDataset implements QueueConfigurer {
+public final class HBaseConsumerStateStore extends AbstractDataset implements QueueConfigurer {
 
   private static final Gson GSON = new Gson();
 
@@ -360,12 +363,25 @@ final class HBaseConsumerStateStore extends AbstractDataset implements QueueConf
     }
   }
 
+  public Transaction getTransaction() {
+    return transaction;
+  }
+
   /**
    * Gets all barrier information for the given group. The information are sorted in the order of
    * the barrier changes.
    */
-  List<QueueBarrier> getAllBarriers(long groupId) {
+  public List<QueueBarrier> getAllBarriers(long groupId) {
     return scanBarriers(groupId, new AllCollector<QueueBarrier>()).finish(new ArrayList<QueueBarrier>());
+  }
+
+  /**
+   * Gets all barrier information for all groups. The information are sorted in the order of
+   * the barrier changes.
+   */
+  public Multimap<Long, QueueBarrier> getAllBarriers() {
+    return scanBarriers(new AllPairCollector<Long, QueueBarrier>()).finishMultimap(
+      LinkedHashMultimap.<Long, QueueBarrier>create());
   }
 
   void getLatestConsumerGroups(Collection<? super ConsumerGroupConfig> result) {
@@ -391,6 +407,25 @@ final class HBaseConsumerStateStore extends AbstractDataset implements QueueConf
     }
   }
 
+  private PairCollector<Long, QueueBarrier> scanBarriers(PairCollector<Long, QueueBarrier> collector) {
+    Scanner scanner = table.scan(barrierScanStartRow, barrierScanEndRow);
+    try {
+      Row row;
+      while ((row = scanner.next()) != null) {
+        Map<Long, QueueBarrier> info = decodeBarrierInfo(row);
+        if (info != null) {
+          for (Map.Entry<Long, QueueBarrier> entry : info.entrySet()) {
+            if (!collector.addElement(entry)) {
+              return collector;
+            }
+          }
+        }
+      }
+    } finally {
+      scanner.close();
+    }
+    return collector;
+  }
 
   private Collector<QueueBarrier> scanBarriers(long groupId, Collector<QueueBarrier> collector) {
     Scanner scanner = table.scan(barrierScanStartRow, barrierScanEndRow);
@@ -498,16 +533,29 @@ final class HBaseConsumerStateStore extends AbstractDataset implements QueueConf
     return column;
   }
 
+  private Map<Long, QueueBarrier> decodeBarrierInfo(Row row) {
+    Map<Long, QueueBarrier> barrierInfo = Maps.newHashMap();
+    Map<byte[], byte[]> columns = row.getColumns();
+    for (byte[] columnKey : columns.keySet()) {
+      byte[] groupInfo = row.get(columnKey);
+      QueueBarrier barrier = decodeBarrierInfo(row.getRow(), groupInfo);
+      if (barrier != null) {
+        barrierInfo.put(barrier.getGroupConfig().getGroupId(), barrier);
+      }
+    }
+    return barrierInfo;
+  }
+
   @Nullable
   private QueueBarrier decodeBarrierInfo(Row row, long groupId) {
-    byte[] groupInfo = row.get(Bytes.toBytes(groupId));
+    return decodeBarrierInfo(row.getRow(), row.get(Bytes.toBytes(groupId)));
+  }
+
+  @Nullable
+  private QueueBarrier decodeBarrierInfo(byte[] rowKey, @Nullable byte[] groupInfo) {
     if (groupInfo == null) {
       return null;
     }
-    return decodeBarrierInfo(row.getRow(), groupInfo);
-  }
-
-  private QueueBarrier decodeBarrierInfo(byte[] rowKey, byte[] groupInfo) {
     ConsumerGroupConfig groupConfig = GSON.fromJson(new String(groupInfo, Charsets.UTF_8), ConsumerGroupConfig.class);
     byte[] startRow = Arrays.copyOfRange(rowKey, queueName.toBytes().length, rowKey.length);
     return new QueueBarrier(groupConfig, startRow);
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
@@ -223,7 +223,7 @@ public class HBaseQueueClientFactory implements QueueClientFactory {
    * @throws IOException
    */
   private HBaseQueueAdmin ensureTableExists(QueueName queueName) throws IOException {
-    HBaseQueueAdmin admin = queueName.isStream() ? streamAdmin : queueAdmin;
+    HBaseQueueAdmin admin = get(queueName);
     try {
       if (!admin.exists(queueName)) {
         admin.create(queueName);
@@ -234,7 +234,16 @@ public class HBaseQueueClientFactory implements QueueClientFactory {
     return admin;
   }
 
-  private HTable createHTable(TableId tableId) throws IOException {
+  /**
+   * Helper method to select the queue or stream admin.
+   * @param queueName name of the queue to be opened.
+   * @return the queue admin for that queue.
+   */
+  public HBaseQueueAdmin get(QueueName queueName) {
+    return queueName.isStream() ? streamAdmin : queueAdmin;
+  }
+
+  public HTable createHTable(TableId tableId) throws IOException {
     HTable consumerTable = hBaseTableUtil.createHTable(hConf, tableId);
     // TODO: make configurable
     consumerTable.setWriteBufferSize(DEFAULT_WRITE_BUFFER_SIZE);
@@ -242,7 +251,7 @@ public class HBaseQueueClientFactory implements QueueClientFactory {
     return consumerTable;
   }
 
-  private int getDistributorBuckets(HTableDescriptor htd) {
+  public int getDistributorBuckets(HTableDescriptor htd) {
     String value = htd.getValue(QueueConstants.DISTRIBUTOR_BUCKETS);
     // If the setting is not in the table meta, this is a old table, hence use the value in the cConf
     if (value == null) {
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
@@ -84,7 +84,7 @@ public final class ShardedHBaseQueueStrategy implements HBaseQueueStrategy, Clos
   /**
    * Constructs a new instance with the given number of buckets for distributed scan.
    */
-  ShardedHBaseQueueStrategy(int distributorBuckets) {
+  public ShardedHBaseQueueStrategy(int distributorBuckets) {
     this.distributorBuckets = distributorBuckets;
     this.rowKeyDistributor = new RowKeyDistributorByHashPrefix(
       new RowKeyDistributorByHashPrefix.OneByteSimpleHash(distributorBuckets));
@@ -104,6 +104,12 @@ public final class ShardedHBaseQueueStrategy implements HBaseQueueStrategy, Clos
   @Override
   public QueueScanner createScanner(ConsumerConfig consumerConfig,
                                     HTable hTable, Scan scan, int numRows) throws IOException {
+    ResultScanner scanner = createHBaseScanner(consumerConfig, hTable, scan, numRows);
+    return new HBaseQueueScanner(scanner, numRows, ROW_KEY_CONVERTER);
+  }
+
+  private ResultScanner createHBaseScanner(ConsumerConfig consumerConfig, HTable hTable, Scan scan,
+                                           int numRows) throws IOException {
     // Modify the scan with sharded key prefix
     Scan shardedScan = new Scan(scan);
 
@@ -111,11 +117,16 @@ public final class ShardedHBaseQueueStrategy implements HBaseQueueStrategy, Clos
     int caching = (int) (1.1 * numRows / distributorBuckets);
     shardedScan.setCaching(caching);
 
-    shardedScan.setStartRow(getShardedKey(consumerConfig, consumerConfig.getInstanceId(), scan.getStartRow()));
-    shardedScan.setStopRow(getShardedKey(consumerConfig, consumerConfig.getInstanceId(), scan.getStopRow()));
+    if (scan.getStartRow().length > 0) {
+      byte[] rowKey = getShardedKey(consumerConfig, consumerConfig.getInstanceId(), scan.getStartRow());
+      shardedScan.setStartRow(rowKey);
+    }
 
-    ResultScanner scanner = DistributedScanner.create(hTable, shardedScan, rowKeyDistributor, scansExecutor);
-    return new HBaseQueueScanner(scanner, numRows, ROW_KEY_CONVERTER);
+    if (scan.getStopRow().length > 0) {
+      byte[] rowKey = getShardedKey(consumerConfig, consumerConfig.getInstanceId(), scan.getStopRow());
+      shardedScan.setStopRow(rowKey);
+    }
+    return DistributedScanner.create(hTable, shardedScan, rowKeyDistributor, scansExecutor);
   }
 
   @Override
@@ -159,15 +170,16 @@ public final class ShardedHBaseQueueStrategy implements HBaseQueueStrategy, Clos
     scansExecutor.shutdownNow();
   }
 
-  private byte[] getShardedKey(ConsumerGroupConfig groupConfig, int instanceId, byte[] originalRowKey) {
+  private byte[] getShardedKey(ConsumerGroupConfig groupConfig, int instanceId,
+                               byte[] originalRowKey) {
     // Need to subtract the SALT_BYTES as the row key distributor will prefix the key with salted bytes
     byte[] result = new byte[PREFIX_BYTES - SaltedHBaseQueueStrategy.SALT_BYTES + originalRowKey.length];
     Bytes.putBytes(result, PREFIX_BYTES - SaltedHBaseQueueStrategy.SALT_BYTES,
                    originalRowKey, 0, originalRowKey.length);
+    Bytes.putLong(result, 0, groupConfig.getGroupId());
 
     // Default for FIFO case.
     int shardId = groupConfig.getDequeueStrategy() == DequeueStrategy.FIFO ? -1 : instanceId;
-    Bytes.putLong(result, 0, groupConfig.getGroupId());
     Bytes.putInt(result, Bytes.SIZEOF_LONG, shardId);
 
     return result;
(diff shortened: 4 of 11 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/QueueEntryRow.java
```diff
@@ -318,4 +318,12 @@ public static boolean isCommittedProcessed(byte[] stateBytes, Transaction tx) {
     byte state = stateBytes[Longs.BYTES + Ints.BYTES];
     return state == ConsumerEntryState.PROCESSED.getState();
   }
+
+  /**
+   * Gets the write pointer for a row.
+   */
+  public static long getWritePointer(byte[] rowKey, int queueRowPrefixLenght) {
```
chtyim: Lenght => Length

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -368,6 +371,15 @@ private void deleteRemovedGroups(Row row, Set<Long> existingGroups) {
     return scanBarriers(groupId, new AllCollector<QueueBarrier>()).finish(new ArrayList<QueueBarrier>());
   }
 
+  /**
+   * Gets all barrier information for all groups. The information are sorted in the order of
+   * the barrier changes.
+   */
+  LinkedHashMultimap<Long, QueueBarrier> getAllBarriers() {
```
chtyim: The return type can still be `Multimap`. Just the implementation is `LinkedHashMultimap` to guarantee the ordering (there is other implementation can do that, that's why the return type should be just Multimap).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -68,7 +69,7 @@
   // Number of bytes as the row key prefix, including salt bytes added by the row key distributor
   public static final int PREFIX_BYTES = SaltedHBaseQueueStrategy.SALT_BYTES + Bytes.SIZEOF_LONG + Bytes.SIZEOF_INT;
 
-  private static final Function<byte[], byte[]> ROW_KEY_CONVERTER = new Function<byte[], byte[]>() {
+  static final Function<byte[], byte[]> ROW_KEY_CONVERTER = new Function<byte[], byte[]>() {
```
chtyim: Can you move some of the scan logic into this class so that you don't have to turn things in this class from private to package protected? It's not a very good practice for encapsulation.

chtyim: E.g. you can have something like `createHBaseScanner` for returning the `ResultScanner` instead of `QueueScanner` and have only that method package protected, but not other field/method in this class.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+
+  /**
+   * Only works for {@link co.cask.cdap.data2.transaction.queue.hbase.ShardedHBaseQueueStrategy}.
+   */
+  private void scanQueue(URI queueNameURI) throws Exception {
+    QueueName queueName = QueueName.from(queueNameURI);
+    HBaseConsumerStateStore stateStore = queueAdmin.getConsumerStateStore(queueName);
+
+    Transaction tx = txClient.startLong();
+    stateStore.startTx(tx);
```
chtyim: Don't do manual start/commit of transaction as there are bunch of cases need to handle to not leaving an active transaction in the tx system. Use `TransactionExecutor` instead. E.g.

```java
TransactionExecutorFactory txExecutorFactory = .. // from injector
...
Transactions.createTransactionExecutor(txExecutorFactory, stateStore).execute(...);
```

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+
+    Multimap<Long, QueueBarrier> barriers = stateStore.getAllBarriers();
+    LOG.info("Got {} barriers", barriers.size());
```
chtyim: Since this runs as a tool, output should be using `System.out` / `System.err`, not logger.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+    int currentSection = 1;
+
+    for (Long groupId : barriers.keySet()) {
+      LOG.info("Scanning barriers for group {}", groupId);
+      Collection<QueueBarrier> groupBarriers = barriers.get(groupId);
+      Iterator<QueueBarrier> barrierIterator = groupBarriers.iterator();
+
+      QueueBarrier previous;
+      QueueBarrier current = null;
+      while (barrierIterator.hasNext()) {
```
chtyim: Maybe the logic can be simplified to:

```java
while (barrierIter.hasNext()) {
  QueueBarrier start = barrierIter.next();
  QueueBarrier end = barrierIter.hasNext() ? barrierIter.next() : null;
  // log start
  QueueStatistics sectionStats = scanQueue(tx, queueName, start, end);
  // log end
  currentSection++;
}
```

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+    private QueueStatistics() {
+    }
+
+    private QueueStatistics(long unprocessed, long processedAndVisible, long processedAndNotVisible) {
+      this.unprocessed = unprocessed;
+      this.processedAndVisible = processedAndVisible;
+      this.processedAndNotVisible = processedAndNotVisible;
+    }
+
+    public static QueueStatistics of(QueueStatistics...stats) {
```
chtyim: Instead of having this method, why not have the `scanQueue` method takes a `QueueStatistics` object as argument and modify it instead of returning a new one?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+
+  private QueueStatistics scanQueue(Transaction tx, QueueName queueName, QueueBarrier start,
+                                    @Nullable QueueBarrier end) throws IOException {
+
+    byte[] queueRowPrefix = QueueEntryRow.getQueueRowPrefix(queueName);
+
+    ConsumerGroupConfig consumerConfig = start.getGroupConfig();
+    LOG.info("Got consumer group config: {}", consumerConfig);
+
+    HBaseQueueAdmin admin = queueClientFactory.ensureTableExists(queueName);
```
chtyim: If the queue table doesn't exists, this tool shouldn't creating it. It should just return (as if there is no queue entry).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+    if (end != null) {
+      scan.setStopRow(end.getStartRow());
+    }
+    scan.addColumn(QueueEntryRow.COLUMN_FAMILY, QueueEntryRow.DATA_COLUMN);
+    scan.addColumn(QueueEntryRow.COLUMN_FAMILY, QueueEntryRow.META_COLUMN);
+    scan.addColumn(QueueEntryRow.COLUMN_FAMILY, stateColumnName);
+    scan.setCaching(100);
+    scan.setMaxVersions(1);
+
+    LOG.info("Scanning section with scan: {}", GSON.toJson(scan));
```
chtyim: `Scan` supports `toString()`. Better not to GSON it ourselves.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+    QueueStatistics stats = new QueueStatistics();
+
+    ConsumerConfig consConfig = new ConsumerConfig(consumerConfig, 0);
```
chtyim: So the scan would only return entries for instanceId == 0 only?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+    ResultScanner scanner = queueStrategy.createHBaseScanner(consConfig, hTable, scan, 100, true);
```
chtyim: Why only look for 100 entries?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,308 @@
...
+
+  @Override
+  protected void shutDown() throws Exception {
+    zkClientService.stopAndWait();
+  }
+
+  /**
+   * Only works for {@link co.cask.cdap.data2.transaction.queue.hbase.ShardedHBaseQueueStrategy}.
+   */
+  private void scanQueue(URI queueNameURI) throws Exception {
```
chtyim: It's better for this method to take a `QueueName` object and leave the construction outside of this method.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -159,16 +177,18 @@ public void close() throws IOException {
...
     byte[] result = new byte[PREFIX_BYTES - SaltedHBaseQueueStrategy.SALT_BYTES + originalRowKey.length];
     Bytes.putBytes(result, PREFIX_BYTES - SaltedHBaseQueueStrategy.SALT_BYTES,
                    originalRowKey, 0, originalRowKey.length);
+    Bytes.putLong(result, 0, groupConfig.getGroupId());
 
     // Default for FIFO case.
-    int shardId = groupConfig.getDequeueStrategy() == DequeueStrategy.FIFO ? -1 : instanceId;
-    Bytes.putLong(result, 0, groupConfig.getGroupId());
-    Bytes.putInt(result, Bytes.SIZEOF_LONG, shardId);
+    if (instanceId != null) {
```
chtyim: When it will be null? And what does it mean? I don't see any doc or comment explaining it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: I thought the barriers indicated where one section ends and the other starts? For example, if we have the barriers with the `startRow`s 100, 200, 300, and 400, then I thought we should scan 100-200, 200-300, 300-400, and 400-end.

Looks like your code snippet would scan 100-200, 300-400, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: Different style I guess, though current approach may not be as gc-friendly.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: Is there a way to fetch list of instanceIds? Didn't understand what instanceId would be in this case. If groupSize were 5, then would the instanceIds be 0, 1, 2, 3, 4?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: Doesn't this configure how many rows to fetch at a time? Looked at `Scan.setCaching` javadoc.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
chtyim: yes.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
chtyim: In that case, you can simplify it with the help of a PeekingIterator:

```java
PeekingIterator<QueueBarrier> barrierIter = Iterators.peekingIterator(groupBarriers.iterator());
while (barrierIter.hasNext()) {
  QueueBarrier start = barrierIter.next();
  QueueBarrier end = barrierIter.hasNext() ? barrierIter.peek() : null;

  QueueStatistics sectionStats = scanQueue(tx, queueName, start, end);
  currentSection++;
}
```

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,288 @@
...
+        int currentSection = 1;
+
+        for (Long groupId : barriers.keySet()) {
+          System.out.printf("Scanning barriers for group %s", groupId).println();
+          Collection<QueueBarrier> groupBarriers = barriers.get(groupId);
+          Iterator<QueueBarrier> barrierIterator = groupBarriers.iterator();
+
+          QueueBarrier previous;
+          QueueBarrier current = null;
+          while (barrierIterator.hasNext()) {
```
chtyim: Refer to my comment in https://github.com/caskdata/cdap/pull/2290#discussion_r29283524

chtyim:
Can you not squash the commits before it's all reviewed and going to be merged? It makes it hard to spot the changes.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: Thanks, that's what I was missing.. will fix it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseConsumerStateStore.java
```diff
@@ -499,15 +534,29 @@ private int getInstanceIdFromColumn(byte[] column) {
   }
 
   @Nullable
```
chtyim: I don't think this method every returns null.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,280 @@
...
+
+    ConsumerGroupConfig consumerConfig = start.getGroupConfig();
+    System.out.printf("Got consumer group config: %s", consumerConfig).println();
+
+    if (!queueClientFactory.tableExists(queueName)) {
+      System.out.printf("Found no flow queue table for %s", queueName).println();
+      return;
+    }
+
+    HBaseQueueAdmin admin = queueClientFactory.ensureTableExists(queueName);
```
chtyim: I think I made a comment earlier. For the debug tool, it shouldn't create the table if the table doesn't not exists.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
```diff
@@ -222,7 +222,7 @@ private HBaseQueueProducer createProducer(HTable hTable, QueueName queueName, Qu
    * @return the queue admin for that queue.
    * @throws IOException
    */
-  private HBaseQueueAdmin ensureTableExists(QueueName queueName) throws IOException {
+  HBaseQueueAdmin ensureTableExists(QueueName queueName) throws IOException {
```
chtyim: Shouldn't make this non-private.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueClientFactory.java
```diff
@@ -234,7 +234,22 @@ private HBaseQueueAdmin ensureTableExists(QueueName queueName) throws IOExceptio
...
   }
 
-  private HTable createHTable(TableId tableId) throws IOException {
+  /**
+   * Helper method to check if table exists.
+   * @param queueName name of the queue to check.
+   * @return true if the table exists.
+   * @throws IOException
+   */
+  boolean tableExists(QueueName queueName) throws IOException {
```
chtyim: Do you need this? Can you just rely on the `createHTable` to throw a `TableNotFoundException` to indicate the queue table doesn't not exists?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,280 @@
...
+   */
+  private void scanQueue(final QueueName queueName) throws Exception {
+    HBaseConsumerStateStore stateStore = queueAdmin.getConsumerStateStore(queueName);
+
+    TransactionExecutor txExecutor = Transactions.createTransactionExecutor(txExecutorFactory, stateStore);
+    txExecutor.execute(new TransactionExecutor.Procedure<HBaseConsumerStateStore>() {
+      @Override
+      public void apply(HBaseConsumerStateStore stateStore) throws Exception {
+        Multimap<Long, QueueBarrier> barriers = stateStore.getAllBarriers();
+        System.out.printf("Got %d barriers", barriers.size()).println();
```
chtyim: Just use `\n` in printf to print a newline. It will converted to a platform dependent newline.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  /**
+   * Only works for {@link co.cask.cdap.data2.transaction.queue.hbase.ShardedHBaseQueueStrategy}.
+   */
+  private void scanQueue(final QueueName queueName) throws Exception {
+    HBaseConsumerStateStore stateStore = queueAdmin.getConsumerStateStore(queueName);
+
+    TransactionExecutor txExecutor = Transactions.createTransactionExecutor(txExecutorFactory, stateStore);
+    txExecutor.execute(new TransactionExecutor.Procedure<HBaseConsumerStateStore>() {
```
chtyim: This will actually run with "short" transaction, which scanning queue can get a timeout. I notice that currently there is no way to run long running transaction with TransactionExecutor or transaction context. Can you file a JIRA for Tephra to add that support?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
chtyim: I notice you check tableExists earlier, however it's dangerous to have code like this because if someone just move the code around a bit, it will start creating new table. See my comment above to just rely on the `TableNotFoundException` from HBase to tell if the table exists or not.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,280 @@
...
+      new LocationRuntimeModule().getDistributedModules(),
+      new DataSetsModules().getDistributedModules(),
+      new TransactionModules().getDistributedModules()
+    );
+
+    HBaseQueueDebugger debugger = injector.getInstance(HBaseQueueDebugger.class);
+    debugger.startAndWait();
+
+    // e.g. "queue:///default/PurchaseHistory/PurchaseFlow/reader/queue"
+    final QueueName queueName = QueueName.from(URI.create(args[0]));
```
chtyim: It's not that friendly to require someone to pass in the queue name URI (which is something internal and not known to user. It's better having the arguments as: Namespace (optional, default to "default"), App, Flow, Flowlet and queue name (optional, default to "out"). May consider using `commons-cli` for parsing posix style arguments. I.e.

```--app AppName --flow FlowName --flowlet FlowetName```

Comment (chtyim) [inline cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java]: At least mark it as a TODO. For now, if the args.length == 0, prints a help message to tell how to construct the queue name URI.

File: cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java [hunk 22]
```diff
@@ -0,0 +1,280 @@
...
+    if (end != null) {
+      scan.setStopRow(end.getStartRow());
+    }
+    scan.addColumn(QueueEntryRow.COLUMN_FAMILY, QueueEntryRow.DATA_COLUMN);
+    scan.addColumn(QueueEntryRow.COLUMN_FAMILY, QueueEntryRow.META_COLUMN);
+    scan.addColumn(QueueEntryRow.COLUMN_FAMILY, stateColumnName);
+    scan.setMaxVersions(1);
+
+    System.out.printf("Scanning section with scan: %s", scan.toString()).println();
+    for (int instanceId = 0; instanceId < consumerConfig.getGroupSize(); instanceId++) {
```
chtyim: It is rather inefficient to do a HBase scan per instanceId (consider that there are users that can have hundreds of instances). It's better be one scan, without filtering by instanceId, and do the checking on client side (using the QueueEntryRow.canConsumer() method).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -159,15 +183,16 @@ public void close() throws IOException {
     scansExecutor.shutdownNow();
   }
 
-  private byte[] getShardedKey(ConsumerGroupConfig groupConfig, int instanceId, byte[] originalRowKey) {
+  private static byte[] getShardedKey(ConsumerGroupConfig groupConfig, int instanceId,
```
chtyim: Why changed to static?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/ShardedHBaseQueueStrategy.java
```diff
@@ -111,11 +124,22 @@ public QueueScanner createScanner(ConsumerConfig consumerConfig,
...
+      }
+      shardedScan.setStartRow(rowKey);
+    }
 
-    ResultScanner scanner = DistributedScanner.create(hTable, shardedScan, rowKeyDistributor, scansExecutor);
-    return new HBaseQueueScanner(scanner, numRows, ROW_KEY_CONVERTER);
+    if (scan.getStopRow().length > 0) {
+      byte[] rowKey = getShardedKey(consumerConfig, consumerConfig.getInstanceId(), scan.getStopRow());
+      if (useActualRowKeys) {
+        rowKey = rowKeyDistributor.getDistributedKey(rowKey);
```
chtyim: How does it work now if you set the Scan start/stop row to the actual HBase rowKey (meaning included the salted byte)? Isn't that the DistributedScanner.create() expects unsalted rowKey?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,280 @@
...
+      System.out.printf("Processing instance %d", instanceId).println();
+      ConsumerConfig consConfig = new ConsumerConfig(consumerConfig, instanceId);
+      QueueScanner scanner = queueStrategy.createScannerWithActualRowKeys(consConfig, hTable, scan, 100);
```
chtyim: So the `Scan` contains barrier start/stop row, which is never be a actual row key, but rather just the un-salted, un-partitioned rowkey (Hash + QueueName + Write pointer + counter). How does it work now?

chtyim:
Done review.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
jwang47: Created TEPHRA-95.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/HBaseQueueDebugger.java
```diff
@@ -0,0 +1,337 @@
...
+    public String getReport() {
+      return String.format("unprocessed: %d; processed and visible: %d; processed and not visible: %d; total: %d",
+                           getUnprocessed(), getProcessedAndVisible(), getProcessedAndNotVisible(), getTotal());
+    }
+  }
+
+  /**
+   * Generates a queue consumer groupId for the given flowlet in the given program id.
+   */
+  public static long generateConsumerGroupId(Id.Program program, String flowletId) {
```
chtyim: Move it to cdap-master and use FlowUtils to generate consumer group id.

chtyim:
Just one comment left. Rest looks good to me.

PR #2291, opened 2015-04-24
CDA-2269 use source:sourceId and sink:sinkId as the pluginId for source,...

gokulavasan:
... sink instead of sourceId and sinkId since sourceId and sinkId can be the same

Build : http://builds.cask.co/browse/CDAP-RT3-1
JIRA : https://issues.cask.co/browse/CDAP-2269


bdmogal:
:+1:

PR #2292, opened 2015-04-24
Add Maven dependency plugin

sreevatsanraman:
* Add maven dependency plugin to package dependencies in master package


ajainarayanan:
:+1: LGTM

PR #2293, opened 2015-04-24
Fix/ui etl view urlfix

ajainarayanan:
- Fixes api end points for etl adapter create view
- Fixes create/list views based on end point changes


sreevatsanraman:
Create against release/3.0 if you want the feature in 3.0

PR #2296, opened 2015-04-24
Removing service handler handler method

gokulavasan:
We don't need the method in Application Templates since plugins are recognized and supported by the platform.

Build: http://builds.cask.co/browse/CDAP-RBT224-1


albertshau:
:+1:

PR #2303, opened 2015-04-24
CDAP-1784 adding a way to unit test adapters. TestBase now has

albertshau:
access to methods to deploy a template, deploy template plugins,
and deploy an adapter. Deploying a template will build the
application jar, place it in the correct directory, and deploy
the template application. Deploying a template plugin will
build the plugin jar and place it in the correct diectory.
Deploying an adapter will create the adapter and return an
AdapterManager, which can be used to manage the adapter lifecycle.


albertshau:
http://builds.cask.co/browse/CDAP-RBT225

cdap-test/src/main/java/co/cask/cdap/test/TestManager.java
```diff
@@ -45,6 +50,37 @@ ApplicationManager deployApplication(Id.Namespace namespace,
                                        Class<? extends Application> applicationClz, File... bundleEmbeddedJars);
 
   /**
+   * Deploys an {@link ApplicationTemplate}. Only supported in unit tests.
```
gokulavasan: Curious about the comment - "only supported in unit tests". Where else could this could be used?

albertshau: everything here has a local implementation for unit tests and a remote implementation that just hits rest endpoints and could be used for tests against a distributed instance.

cdap-app-fabric/src/main/java/co/cask/cdap/test/internal/AppFabricClient.java
```diff
@@ -303,4 +312,54 @@ public Location deployApplication(Id.Namespace namespace, String appName,
...
+    DefaultHttpRequest request = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.PUT,
+      String.format("/v3/namespaces/%s/templates/%s", namespace.getId(), templateId.getId()));
+    MockResponder responder = new MockResponder();
+    adapterHttpHandler.deployTemplate(request, responder, namespace.getId(), templateId.getId());
+    verifyResponse(HttpResponseStatus.valueOf(200), responder.getStatus(), "Failed to deploy template.");
+  }
+
+  public void createAdapter(Id.Adapter id, AdapterConfig config) {
+    DefaultHttpRequest request = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.PUT,
+      String.format("/v3/namespaces/%s/adapters/%s", id.getNamespaceId(), id.getId()));
```
henry-cask: Would hard code with /v3/ cause any trouble with next release API version change?

albertshau: yep, i will change them to use the version constant we have, as well as the other hardcoded v3s in this class.

cdap-test/src/main/java/co/cask/cdap/test/TestManager.java
gokulavasan: Oh ok.

cdap-test/src/main/java/co/cask/cdap/test/TestManager.java
```diff
@@ -45,6 +50,37 @@ ApplicationManager deployApplication(Id.Namespace namespace,
...
+
+  /**
+   * Creates an adapter.
+   *
+   * @param adapterId The id of the adapter to create
+   * @param config The configuration for the adapter
+   * @return An {@link AdapterManager} to manage the deployed adapter.
+   * @throws Exception if there was an exception deploying the adapter.
+   */
+  AdapterManager deployAdapter(Id.Adapter adapterId, AdapterConfig config) throws Exception;
```
gokulavasan: should this be createAdapter?

cdap-unit-test/src/main/java/co/cask/cdap/test/ConfigurableTestBase.java
```diff
@@ -373,6 +377,44 @@ protected static ApplicationManager deployApplication(Class<? extends Applicatio
...
 
   /**
+   * Creates an adapter.
+   *
+   * @param adapterId The id of the adapter to create
+   * @param adapterConfig The configuration for the adapter
+   * @return An {@link AdapterManager} to manage the deployed adapter.
+   * @throws Exception if there was an exception deploying the adapter.
+   */
+  protected static AdapterManager deployAdapter(Id.Adapter adapterId, AdapterConfig adapterConfig) throws Exception {
```
gokulavasan: createAdapter instead of deployAdapter?

cdap-test/src/main/java/co/cask/cdap/test/TestManager.java
albertshau: hm yeah i think that is a better name since we're not deploying a jar or anything. will change.

cdap-unit-test/src/main/java/co/cask/cdap/test/UnitTestManager.java
```diff
@@ -110,6 +131,33 @@ public ApplicationManager deployApplication(Id.Namespace namespace,
...
+  @Override
+  public void addTemplatePlugin(Id.ApplicationTemplate templateId, Class<?> pluginClz,
+                                String jarName) throws IOException {
+    Manifest manifest = new Manifest();
+    manifest.getMainAttributes().put(ManifestFields.EXPORT_PACKAGE, pluginClz.getPackage().getName());
+    Location pluginJar = AppJarHelper.createDeploymentJar(locationFactory, pluginClz, manifest);
+    File templateDir = new File(pluginsDir, templateId.getId());
+    DirUtils.mkdirs(templateDir);
+    File destination = new File(templateDir, jarName);
+    Files.copy(Locations.newInputSupplier(pluginJar), destination);
```
gokulavasan: Do we need to call deployTemplate or something equivalent to refresh, so that the plugin jar is picked up?

gokulavasan: Or do we expect users to add all template plugins before deploying a template? If yes, then we should add it to javadoc

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
```diff
@@ -168,11 +168,18 @@ public void schedule(Id.Program programId, SchedulableProgramType programType, S
...
-      try {
-        scheduler.suspendSchedule(programId, programType, schedule.getName());
-      } catch (NotFoundException e) {
-        // Should not happen - we just created it. Could have been deleted just in between
-        LOG.info("Schedule could not be suspended - it did not exist: {}", schedule.getName());
+      // TODO: CDAP-2281 figure out a better way to handle schedules in unit tests
+      String ignoreLazy = properties.get(Constants.Scheduler.IGNORE_LAZY_START);
+      boolean scheduleIgnoresLazy = ignoreLazy != null && Boolean.valueOf(ignoreLazy);
+      if (scheduleIgnoresLazy) {
+        lazyStart(scheduler);
```
gokulavasan: where was lazyStart being called before this change? I couldn't find it in the code removed below.

nit: Also this check and method name is confusing, if "schedule ignores lazy", then "lazy start the schedule".

cdap-unit-test/src/main/java/co/cask/cdap/test/UnitTestManager.java
albertshau: that is a good point, plugins are not visible until you deploy the template. Since this is the behavior in the real system, will add to docs over refreshing in unit tests.

gokulavasan:
Few comments. Rest looks great!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
albertshau: it's normally called when you resume a schedule.  Agree about the name, I'll at least add a comment, maybe change the name.

cdap-unit-test/src/main/java/co/cask/cdap/test/UnitTestManager.java
gokulavasan: But if users add new plugins, they can refresh the plugin/template info in the system by hitting the update-template endpoint isn't it?

albertshau: yup.

PR #2304, opened 2015-04-24
[CDAP-2266] use version 1.2.0 of cdap-clients, which uses v3 APIs

anew:
title says it all. 
Build runs here: http://builds.cask.co/browse/CDAP-RBT226-1


henry-cask:
+1

PR #2305, opened 2015-04-24
Feature/ui list programs under dataset stream

elfenheart:
Programs List under Dataset and Streams

screen shot 2015-04-24 at 1 34 04 pm


ajainarayanan:
LGTM. Could just be my opinion but we already name the tab as "Programs" Do we want the first column to be programs? We have it as "Name" under data tab of a program.

PR #2306, opened 2015-04-24
Update support of CDH to be only through CDH 5.3.x.

awholegunch:


cdap-docs/reference-manual/source/release-notes.rst
```diff
@@ -64,6 +64,8 @@ API Changes
 Known Issues
 ------------
 
+- CDAP has been tested on and supports CDH 4.2.x through CDH 5.3.x. It has not been tested on later versions.
```
dereklwood: maybe it's appropriate here to mention that it hasn't been tested on HBase 1.0, which is the reason why it's limited to CDH 5.3 and below.  ("but testing is underway" or similar)

dereklwood:
just a suggestion (up to you).  otherwise, :+1: LGTM

cdap-docs/reference-manual/source/release-notes.rst
awholegunch: But maybe what's more germane is to change
```
HBase 0.94.2+, 0.96.0+, 0.98.0+
```
to something such as
```
HBase 0.94.2.x, 0.96.x, 0.98.x
```
Is that correct? Did we stop at 0.98?

awholegunch: And, while we are at, what about Hadoop? 2.7.0 is available, but what about us (for this and the following)?

"Zookeeper 3.4.3"; "3.4.6" its current "stable" release.

"Hive 12.0" -> should probably be written "Hive 0.12.0". Latest Hive is "1.1.0".

dereklwood: Sure... yeah that's correct.  Except for the lower bound, I don't think ``0.94.2.x`` is the correct nomenclature, since the versions we support would be ``0.94.2``, ``0.94.3``, and so on...  ``0.94.x where x >= 2`` is a verbose but correct description :)  

FYI ``0.94.2`` is the version of Hbase included in CDH 4.2.0, which is why it's our minimum.

dereklwood: I don't think we've tested Hadoop 2.7 or Zookeeper 3.4.6 at this point, as they are not yet in any of the major distributions.  Agreed on the Hive version suggestion.

cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -145,25 +145,33 @@ Hadoop/HBase Environment
...
++===============+===================+==================================================+
+| **HDFS**      | Apache Hadoop DFS | 2.0.2-alpha through 2.5.0                        |
++               +-------------------+--------------------------------------------------+
+|               | CDH or HDP        | (CDH) 4.2.x through 5.3.3 or (HDP) 2.0 or 2.1    |
++---------------+-------------------+--------------------------------------------------+
+| **YARN**      | Apache Hadoop DFS | 2.0.2-alpha through 2.5.0                        |
++               +-------------------+--------------------------------------------------+
+|               | CDH or HDP        | (CDH) 4.2.x through 5.3.3 or (HDP) 2.0 or 2.1    |
++---------------+-------------------+--------------------------------------------------+
+| **HBase**     | Apache Hadoop DFS | 0.94.2.x, 0.96.0.x, and 0.98.6                   |
```
dereklwood: hbase versions still not technically correct, i'd suggest: ``0.94.2+, 0.96.x, 0.98.x``

cdap-docs/reference-manual/source/faq.rst
```diff
@@ -96,7 +96,7 @@ Yes. You can install CDAP on your Hadoop cluster. See :ref:`install`.
 
 .. rubric:: What Hadoop distributions can CDAP run on?
 
-CDAP has been tested on and supports CDH 4.2.x or later, HDP 2.0 or later, and Apache Hadoop/HBase 2.0.2-0.4 and 2.1.0. 
+CDAP has been tested on and supports CDH 4.2.x through 5.3.x, HDP 2.0 through 2.1, and Apache Hadoop/HBase 2.0.2, 2.0.4, and 2.1.0.
```
dereklwood: let's add Apache Bigtop 0.8.0.  And I suggest removing Apache Hadoop/HBase as it's confusing (I am guessing this is referring to when HBase used to be bundled with Apache?)

cdap-docs/reference-manual/source/release-notes.rst
```diff
@@ -64,6 +64,9 @@ API Changes
 Known Issues
 ------------
 
+- CDAP has been tested on and supports CDH 4.2.x through CDH 5.3.x. It has not been tested on
```
dereklwood: mention HDP here too?

dereklwood:
few minor comments, otherwise :+1: LGTM

cdap-docs/reference-manual/source/faq.rst
awholegunch: I'm wondering if even that's correct: should it not be (too match the chart above):
```
 ...and Apache Hadoop 2.0.2-alpha through 2.5.0
```
otherwise, it would seem odd to me to see once sequence in one place and a different one elsewhere.

cdap-docs/reference-manual/source/faq.rst
```diff
@@ -96,7 +96,8 @@ Yes. You can install CDAP on your Hadoop cluster. See :ref:`install`.
 
 .. rubric:: What Hadoop distributions can CDAP run on?
 
-CDAP has been tested on and supports CDH 4.2.x or later, HDP 2.0 or later, and Apache Hadoop/HBase 2.0.2-0.4 and 2.1.0. 
+CDAP has been tested on and supports CDH 4.2.x through 5.3.x, HDP 2.0 through 2.1, 
+Apache Bigtop 0.8.0, and Apache Hadoop 2.0.2, 2.0.4, and 2.1.0.
```
dereklwood: I think we should leave out "Apache Hadoop 2.0.2, 2.0.4, and 2.1.0."  It's not a full distribution, ie no Hbase.

cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -145,25 +145,33 @@ Hadoop/HBase Environment
...
++---------------+-------------------+--------------------------------------------------+
+| **YARN**      | Apache Hadoop DFS | 2.0.2-alpha through 2.5.0                        |
++               +-------------------+--------------------------------------------------+
+|               | CDH or HDP        | (CDH) 4.2.x through 5.3.3 or (HDP) 2.0 or 2.1    |
++---------------+-------------------+--------------------------------------------------+
+| **HBase**     | Apache Hadoop DFS | 0.94.2+, 0.96.x, and 0.98.x                      |
++               +-------------------+--------------------------------------------------+
+|               | CDH or HDP        | (CDH) 4.2.x through 5.3.3 or (HDP) 2.0 or 2.1    |
++---------------+-------------------+--------------------------------------------------+
+| **Zookeeper** | Apache Hadoop DFS | Version 3.4.3 through 3.4.5                      |
```
dereklwood: just "Apache"

cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -145,25 +145,33 @@ Hadoop/HBase Environment
...
++===============+===================+==================================================+
+| **HDFS**      | Apache Hadoop DFS | 2.0.2-alpha through 2.5.0                        |
++               +-------------------+--------------------------------------------------+
+|               | CDH or HDP        | (CDH) 4.2.x through 5.3.3 or (HDP) 2.0 or 2.1    |
++---------------+-------------------+--------------------------------------------------+
+| **YARN**      | Apache Hadoop DFS | 2.0.2-alpha through 2.5.0                        |
++               +-------------------+--------------------------------------------------+
+|               | CDH or HDP        | (CDH) 4.2.x through 5.3.3 or (HDP) 2.0 or 2.1    |
++---------------+-------------------+--------------------------------------------------+
+| **HBase**     | Apache Hadoop DFS | 0.94.2+, 0.96.x, and 0.98.x                      |
```
dereklwood: just "Apache"

dereklwood:
:+1: LGTM

PR #2311, opened 2015-04-24
CDAP-1347: change mapreduce logs to be using runid

elfenheart:
Change logs for mapreduce to use runid


ajainarayanan:
:+1: LGTM

PR #2313, opened 2015-04-25
MR metrics, use separate names for job level and task level metrics 

gsps1:
Since Standalone doesn't have any tasks, aggregating task level metrics for a job as job level metrics wouldn't work.
so having separate metric names for job and task level metrics. 

In UI code, we were using number of tasks and the number of tasks completed to calculate the percentage completion, since the numTasks is always zero in standalone, completion percentage is also always zero.

mapProgress and reduceProgress returned by the info endpoint is correct and using that to show the progress completion instead.
https://issues.cask.co/browse/CDAP-2138


anwar6953:
Is the `getCompletedPercentage` function still in use, then?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceMetricsWriter.java
```diff
@@ -95,7 +95,15 @@ private void reportMapredStats() throws IOException, InterruptedException {
     int memoryPerMapper = jobConf.getConfiguration().getInt(Job.MAP_MEMORY_MB, Job.DEFAULT_MAP_MEMORY_MB);
     int memoryPerReducer = jobConf.getConfiguration().getInt(Job.REDUCE_MEMORY_MB, Job.DEFAULT_REDUCE_MEMORY_MB);
 
+    Counters counters = jobConf.getCounters();
```
anwar6953: If this is not local (rpc call?), then just use the same instance that's used in `reportSystemStats` method.

anwar6953:
Hmm?

anwar6953:
LGTM

gsps1:
http://builds.cask.co/browse/CDAP-RBT227 - build was green and the last build failed due to known issue in release tracked here https://issues.cask.co/browse/CDAP-2333 - so merging now.

PR #2316, opened 2015-04-25
CDAP System Dashboard.

nitinmotgi:


cdap-ui/app/features/dashboard/templates/staticdashboard.html
```diff
@@ -1,5 +1,5 @@
 <div class="row" ng-controller="WidgetColCtrl">
-  <div ng-class="col-xs-12 col-md-6"
+  <div ng-class="{'col-xs-12 col-md-6' : true}"
```
elfenheart: if this is always true, then we should change this to just class="col-xs-12 col-md-6"  instead of ng-class

cdap-ui/server/config/development/cdap.json
```diff
@@ -110,7 +110,7 @@
   "app.exec.threads": "20",
   "stream.bind.address": "127.0.0.1",
   "data.tx.snapshot.dir": "/cdap/tx.snapshot",
-  "dashboard.bind.port": "9999",
+  "dashboard.bind.port": "8080",
```
elfenheart: Are we always binding the port to 8080 instead of 9999?

anwar6953: Pretty sure he just accidentally left it in.

cdap-ui/app/features/dashboard/templates/staticdashboard.html
nitinmotgi: done!

ajainarayanan: @nitinmotgi Is this changed? if we want 3 columns in the system dashboards then ```class="col-xs-12 col-md-4"``` is what we needed.. I see ```widgetsPerRow``` in controller to be 3. Just verifying..

cdap-ui/app/features/dashboard/templates/staticdashboard.html
```diff
@@ -1,5 +1,5 @@
 <div class="row" ng-controller="WidgetColCtrl">
-  <div ng-class="col-xs-12 col-md-6"
+  <div class="col-xs-12 col-md-4"
```
nitinmotgi: now done!

ajainarayanan:
:+1: LGTM

PR #2317, opened 2015-04-25
[CDAP 2259] pending events queue metric

anew:
- Adds a new metric context: namespace.{ns}.app.{appid}.flow.{flowid}.producer.{flowlet}.queue.{name}.consumer.{flowlet}
- Adds a new metric in that context: system.queue.pending
- Flowlet runner emits an increment for that metric when it enqueues, and a decrement when it dequeues
- Adds a unit test with a specially tailored flow to validate interim queue.pending metrics

Bamboo build: http://builds.cask.co/browse/CDAP-RBT228-1


anew:
@abaranau please take a look at the change to metrics (new tags, new aggregation)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/BasicFlowletContext.java
```diff
@@ -83,6 +85,18 @@ public MetricsCollector load(String key) throws Exception {
...
 
+    this.producerMetrics = CacheBuilder.newBuilder()
+      .expireAfterAccess(1, TimeUnit.HOURS)
+      .build(new CacheLoader<ImmutablePair<String, String>, MetricsCollector>() {
+        @Override
+        public MetricsCollector load(ImmutablePair<String, String> key) throws Exception {
+          return getProgramMetrics()
+            .childCollector(Constants.Metrics.Tag.QUEUE_PRODUCER, key.getFirst())
+            .childCollector(Constants.Metrics.Tag.FLOWLET_QUEUE, key.getSecond())
+            .childCollector(Constants.Metrics.Tag.QUEUE_CONSUMER, BasicFlowletContext.this.flowletId);
```
gsps1: childCollector accepts a Map<String, String> , its better to use that than multiple calls , as each of the childCollector call adds a map to the collectorCache.

anew: oh, good to know.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -119,6 +119,9 @@
       .put(Constants.Metrics.Tag.FLOWLET, "flowlet")
       .put(Constants.Metrics.Tag.FLOWLET_QUEUE, "queue")
 
+      .put(Constants.Metrics.Tag.QUEUE_PRODUCER, "producer")
```
abaranau: Does it make sense to have generic "PRODUCER" and "CONSUMER" tag? to be reused, e.g. like HANDLER?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
```diff
@@ -113,6 +113,11 @@ public Cube get() {
       // i.e. for flows only
       ImmutableList.of(Constants.Metrics.Tag.NAMESPACE, Constants.Metrics.Tag.APP,
                        Constants.Metrics.Tag.FLOW)));
+    // queue
+    aggs.put("queue", new DefaultAggregation(
```
abaranau: you don't need instance or run id, right? just confirming..

abaranau: do you still need FLOWLET_QUEUE in "flow" aggregation?

abaranau: "queue" -> "flow.queue" ?

anew: I don't really need instance id and run id, at least not to show total pending for the queue in the flow status. But an operator may want to see these by flowlet instance, or by run id, so I thought it is useful to keep them here.

anew: FLOWLET_QUEUE in flow aggregation is still used for the system.process.events.out/in metrics. I am not sure whether it is ok to remove it.

anew: I renamed it to flow.queue. Didn't know that matters. The other aggregation names don't have dots in them... But fine.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
anew: OK, that makes sense.

anew:
@abaranau I addressed your comments

anew:
@chtyim please review this

gsps1:
@anew something is wrong, 71 files changed.

anew:
oh shoot did I pull develop? Let me fix that

anew:
allright, back to 14 changed files

anew:
wanna take another look @sreevatsanraman ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/BasicFlowletContext.java
```diff
@@ -83,6 +86,18 @@ public MetricsCollector load(String key) throws Exception {
         }
       });
 
+    this.producerMetrics = CacheBuilder.newBuilder()
```
gsps1: why do you need a cache? MetricsCollectionService has a cache for collectors, the only difference is there the expiry time is for a minute, while its an hour here. is that why you added this cache - longer time for expiry ?

anew: No, because there was an existing cache for the other flowlet metrics. I figured that I should  follow the existing design of that class. But I don't mind at all removing the cache. @chtyim do you know why there was cache in the first place?

cdap-app-fabric/src/main/java/co/cask/cdap/app/metrics/FlowMetrics.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(FlowMetrics.class);
+
+  /**
+   * Delete the "system.queue.pending" metrics for a flow or for all flows in an app or a namespace.
+   *
+   * @param namespace the namespace id; may only be null if the appId and flowId are null
+   * @param appId the application id; may only be null if the flowId is null
+   */
+  public static void deleteFlowPendingMetrics(MetricStore metricStore,
```
chtyim: Maybe move to `FlowUtils` to minimize number of util class?

cdap-app-fabric/src/main/java/co/cask/cdap/app/metrics/FlowMetrics.java
```diff
@@ -0,0 +1,70 @@
...
+    if (namespace != null) {
+      tags.put(Constants.Metrics.Tag.NAMESPACE, namespace);
+      if (appId != null) {
+        tags.put(Constants.Metrics.Tag.APP, appId);
+        if (flowId != null) {
+          tags.put(Constants.Metrics.Tag.FLOW, flowId);
+        }
+      }
+    }
+    LOG.info("Deleting 'system.queue.pending' metric for context " + tags);
```
chtyim: Use `{}` instead of `+` when logging.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/BasicFlowletContext.java
chtyim: I think the newer metrics implementation has incorporated the cache (which it wasn't before). I believe we can remove cache usage and just let the metrics api to do so.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowletProgramRunner.java
```diff
@@ -429,34 +430,57 @@ private OutputEmitterFactory outputEmitterFactory(final BasicFlowletContext flow
...
-                                                           flowletName, type));
-
+          // create a metric collector for this queue, and also one for each consumer flowlet
+          final MetricsCollector metrics = flowletContext.getProgramMetrics()
+            .childCollector(Constants.Metrics.Tag.FLOWLET_QUEUE, outputName);
+          final MetricsCollector producerMetrics = metrics.childCollector(
+            Constants.Metrics.Tag.PRODUCER, flowletContext.getFlowletId());
+          final Iterable<MetricsCollector> consumerMetrics =
+            Iterables.transform(consumerFlowlets, new Function<String, MetricsCollector>() {
+              @Nullable
```
chtyim: Can remove this.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowletProgramRunner.java
```diff
@@ -429,34 +430,57 @@ private OutputEmitterFactory outputEmitterFactory(final BasicFlowletContext flow
...
+              @Override
+              public MetricsCollector apply(@Nullable String consumer) {
```
chtyim: Can remove the `@Nullable` as I believe the list that it's transforming won't have null entry. This can eliminate unnecessary warning from IDE (if childCollector() doesn't accept null).

chtyim:
Overall LGTM. Just couple minor comments.

anew:
@chtyim did you see @gsps1 's question?

cdap-app-fabric/src/main/java/co/cask/cdap/app/metrics/FlowMetrics.java
anew: oh, I did not know about that. will do.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowletProgramRunner.java
anew: this (and the previous one) were generated by IDEA when I added the override method. Does it harm?

chtyim:
Yes. I think I replied. The caching can be removed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowletProgramRunner.java
chtyim: Should be ok. The parent class has it nullable for general case. For this implementation, however, we know for sure it cannot be null, removing should be fine.

PR #2319, opened 2015-04-25
Fix/ui etl fancy edit properties

ajainarayanan:
- Minor cleanup of ETL App Create view
- First draft of ```widget-container```. A basic widget container that given a type, property and config renders the appropriate ui element.
- Adds json & javascript editor widgets to be used in editing properties.

NOTE:
 - Right now all input type elements are bound to a model. Javascript editor is yet to be bounded to a model (ng-model). Will split it in next PR.
 - Doing it only for Source Properties. Will add Sink configuration json files to be used in editing sink properties (and eventually transforms).

### DIFF
--- a/cdap-ui/app/features/etlapps/services/etl-plugin-config-factory.js
+++ b/cdap-ui/app/features/etlapps/services/etl-plugin-config-factory.js
@@ -1,0 +1,24 @@
+angular.module(PKG.name + '.feature.etlapps')
+  .service('PluginConfigFactory', function(MyDataSource, $q) {
+    this.plugins = {};
+
+    this.fetch = function(scope, templateid, pluginid) {
+      var dataSrc = new MyDataSource(scope);
+      var defer = $q.defer();
+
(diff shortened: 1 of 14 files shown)


cdap-ui/app/features/etlapps/services/etl-plugin-config-factory.js
```diff
@@ -0,0 +1,24 @@
...
+
+      dataSrc.config({
+        templateid: templateid, //'etlRealtime',
+        pluginid: pluginid //'TwitterSource'
+      })
+        .then(function(res) {
+          this.plugins[templateid+pluginid] = res;
+          defer.resolve(this.plugins[templateid+pluginid]);
+        }.bind(this));
+      return defer.promise;
```
elfenheart: If you are already returning a promise, do you need to bind(this) as well to the 'then' function?

sreevatsanraman:
This is against develop. Please change to release/3.0

ajainarayanan:
Oh Snap! Sorry again...

PR #2320, opened 2015-04-25
(CDAP-2275) Make plugin classes exposed to MR framework.

chtyim:
- Changes in the MapReduceClassLoader to support exposing Plugin classes to the MR framework
  - It is needed when a plugin set classes needed by the framework, e.g. OutputFormat
  - Currently only classes under Export-Package are exposed to provide ClassLoading isolation


chtyim:
Closing and recreate against release branch.

PR #2322, opened 2015-04-25
(CDAP-2275) Make plugin classes exposed to MR framework.

chtyim:


chtyim:
Test passed: http://builds.cask.co/browse/CDAP-RBT229-2

anew:
I only skimmed through the code in a first pass. I need more time to review this throughly. 
One thing I can say already is that we need tests for our class loading. It is becoming quite complex, and any change in future releases can easily break it, so it would be good to get test coverage at least for the known cases.

chtyim:
Testing of the regular mapreduce class loader is already there. However the testing of the plugin support in mapreduce (which this PR is about) is pending, mainly waiting for the adapter support in unit test (because only adapter can use plugin right now). Also testing of classloading in cluster mode currently is done through integration test, not unit test

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceClassLoader.java
```diff
@@ -34,49 +48,238 @@
...
+   * Constructs a ClassLoader that load classes from the programClassLoader, then from the plugin lib ClassLoader,
+   * followed by plugin Export-Package ClassLoader and with the sytem ClassLoader last.
+   */
+  public MapReduceClassLoader(ClassLoader programClassLoader,
+                              @Nullable AdapterDefinition adapterSpec,
+                              @Nullable PluginInstantiator pluginInstantiator) {
+    this(new Parameters(programClassLoader, adapterSpec, pluginInstantiator));
+  }
+
+  private MapReduceClassLoader(Parameters paramters) {
```
anew: typo: parameters

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceClassLoader.java
```diff
@@ -34,49 +48,238 @@
...
 
   private static final Logger LOG = LoggerFactory.getLogger(MapReduceClassLoader.class);
+  private static final Function<String, String> CLASS_TO_RESOURCE_NAME = new Function<String, String>() {
+    @Override
+    public String apply(String className) {
+      return className.replace('.', '/') + ".class";
+    }
+  };
+
+  private final Parameters paramters;
```
anew: parameters

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceClassLoader.java
chtyim: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceClassLoader.java
chtyim: Fixed.

chtyim:
New build pass: http://builds.cask.co/browse/CDAP-RBT229

anew:
LGTM. Please test this on clusters with all versions of Hadoop that we support before merging. It is hard to tell just from looking at the code, whether this will work in all cases.

chtyim:
Verify on Hadoop 2.4 and 2.6. For 2.2, it's still remain as an open issue, which is unrelated to this change (see MAPREDUCE-5751)

chtyim:
Commits squashed. Merging.

PR #2326, opened 2015-04-25
Feature/move delete button

astewes:


cdap-ui/app/features/admin/templates/namespace/app-metadata.html
```diff
@@ -3,16 +3,30 @@
...
+            <li role="presentation">
+              <a role="menuitem" ui-sref="admin.namespace.detail.apps.metadata.preference">
+                <span class="fa fa-fw fa-cog pull-right"></span>
+                Preferences
+              </a>
+            </li>
+            <li role="presentation">
+              <a role="menuitem" href=""
+                ng-click="caskConfirm()"
+                cask-confirmable="deleteApp(app.id)"
```
ajainarayanan: In the context of "All Applications" app.id will be null. This is delete behavior/button should be added either to all the rows of the app or should atleast link to the admin page that allows the user to delete the app (which is the desired behavior I believe)

cdap-ui/app/features/apps/templates/list.html
```diff
@@ -4,9 +4,28 @@
...
+      <ul class="dropdown-menu dropdown-menu-right" role="menu">
+        <li role="presentation">
+          <my-file-select on-file-select="onFileSelected($files)"
+            data-button-label="Add App"
+            data-button-icon="fa-plus"
+            data-dropdown="true"> </my-file-select>
+        </li>
+        <li role="presentation">
+          <a ng-click="caskConfirm()"
+             cask-confirmable="deleteApp(app.id)"
```
ajainarayanan: In the context of "All Applications" app.id will be null. This is delete behavior/button should be added either to all the rows of the app or should atleast link to the admin page that allows the user to delete the app (which is the desired behavior I believe)

cdap-ui/app/features/admin/templates/partials/app-dd.html
```diff
@@ -1,20 +1,21 @@
-<ul class="dropdown-menu" role="menu">
+<!-- <ul class="dropdown-menu dropdown-menu-right" role="menu">
```
ajainarayanan: Just a minor comment but if we are not going to use the commented out code can we remove them? For some reason its a little distracting.

cdap-ui/app/features/apps/templates/list.html
```diff
@@ -4,9 +4,28 @@
     <h2> All Applications </h2>
   </div>
   <div class="col-xs-6 h2 text-right">
-    <my-file-select on-file-select="onFileSelected($files)"
-      data-button-label="Add App"
-      data-button-icon="fa-plus"> </my-file-select>
+    <div class="btn-group" dropdown="">
```
ajainarayanan: Do we need ```dropdown=""``` Can't we just have 

```<div class="btn-group" dropdown>
      <button type="button" dropdown-toggle>
      ...
   ...
```
We can specify them as tags since its an angular directive (I suppose we are using angular bootstrap's dropdown)

cdap-ui/app/styles/themes/cdap/mixins.less
```diff
@@ -55,3 +55,59 @@
...
+      &:last-child a {
+        .border-radius(0 0 8px 8px);
+        &:hover { .border-radius(0 0 8px 8px); }
+      }
+      &:only-child a { .border-radius(8px 0 8px 8px); }
+      > a {
+        padding: 10px;
+        cursor: pointer;
+      }
+      my-file-select {
```
ajainarayanan: Shouldn't this go into myfileselect.less?

cdap-ui/app/styles/themes/cdap/mixins.less
```diff
@@ -55,3 +55,59 @@
...
+    > li {
+      > a {
+        background-color: @cdap-header;
+        color: white;
+        &:hover, &:focus {
+          background-color: white;
+          color: @cdap-header;
+        }
+      }
+      my-file-select {
```
ajainarayanan: Same here too. The assumption is to use the same mixin everywhere for a dropdown. So this would be valid everywhere right?

cdap-ui/app/features/flows/templates/detail.html
```diff
@@ -37,16 +37,16 @@ <h2 class="programTitle">
         <span class="caret"></span>
       </button>
 
-      <ul class="dropdown-menu" role="menu" ng-if="status === 'STOPPED'">
+      <ul class="dropdown-menu dropdown-menu-right" role="menu" ng-if="status === 'STOPPED'">
         <li>
           <a href="#" ng-click="openRuntime()">
-            <span class="fa fa-wrench"></span>
+            <span class="fa fa-wrench pull-right"></span>
             Runtime Arguments
```
ajainarayanan: If we are adding the icon in span, I feel we should wrap the text in span too. (Makes it slightly easier to style in the future) Or we could use ```<i>``` tags for the icons.

cdap-ui/app/directives/fileselect/fileselect.html
```diff
@@ -1,5 +1,10 @@
 
-<button class="btn btn-default" ng-class="buttonClass" ng-disabled="buttonDisabled">
+<button class="btn btn-default" ng-class="buttonClass" ng-disabled="buttonDisabled" ng-if="!isDropdown">
   <span class="fa fa-plus"> </span>
   <span> {{buttonLabel}} </span>
 </button>
+
+<a role="menuitem" ng-if="isDropdown">
```
ajainarayanan: Same here too. wrapping ```{{buttonLabel}}``` in a span would be better. But I like this simplicity! Nice!

cdap-ui/app/styles/themes/cdap/mixins.less
astewes: I recommend keeping these mixins outside of fileselect.less. More often than not, dropdown menus contain actions beyond the fileselect directive, or don't contain a fileselect at all. One option would be to account for the filleselect style rules separately, in fileselect.less, but this seems like a fragmented approach.

ajainarayanan:
:+1: LGTM

PR #2330, opened 2015-04-26
Fix/ui etl edit properties 2

nitinmotgi:


nitinmotgi:
opened for @ajainarayanan. LGTM :+1:

PR #2333, opened 2015-04-26
Removes Save/Close buttons from Source/Sink Property views

ajainarayanan:
Removes Save/Close buttons from Source/Sink Property views


nitinmotgi:
LGTM :+1:

PR #2337, opened 2015-04-27
Disable twitter4j logs

sreevatsanraman:
http://builds.cask.co/browse/CDAP-DT60
https://issues.cask.co/browse/CDAP-2290


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/TwitterSource.java
```diff
@@ -192,6 +192,9 @@ public void onException(Exception e) {
       }
     };
 
+    // Disable chatty logging from twitter4j.
```
chtyim: It's better to do it at the first statement in this method (maybe right after `super.initialize`).

sreevatsanraman: Fixed.

gokulavasan:
LGTM :+1:

PR #2342, opened 2015-04-27
Fix/ui flowlet set instance

elfenheart:
Fix bug with flowlet set instance box:
- Typo on ng-click
- Instance number field can go to negative


ajainarayanan:
:+1: LGTM

PR #2343, opened 2015-04-27
CDAP-2291 return null if RunRecord is not present

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-2291

Bamboo : http://builds.cask.co/browse/CDAP-RBT232-1


sreevatsanraman:
LGTM :+1:

PR #2344, opened 2015-04-27
Feature/fix docs deploy ssh

duddun:
This fixes the issue with creating directories on the remote docs/staging server.


dereklwood:
:+1:

PR #2345, opened 2015-04-27
CDAP-1773: Added new Adapter CLI commands

jwang47:
http://builds.cask.co/browse/CDAP-RBT217
Continued from https://github.com/caskdata/cdap/pull/2259
* Fully tested for realtime and batch adapters


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
```diff
@@ -29,11 +29,13 @@
 import co.cask.cdap.internal.app.runtime.adapter.InvalidAdapterOperationException;
 import co.cask.cdap.internal.app.runtime.schedule.SchedulerException;
 import co.cask.cdap.proto.AdapterConfig;
+import co.cask.cdap.proto.AdapterDetail;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.ProgramRunStatus;
 import co.cask.cdap.proto.RunRecord;
 import co.cask.cdap.templates.AdapterDefinition;
 import co.cask.http.HttpResponder;
+import com.clearspring.analytics.util.Lists;
```
gokulavasan: Use Guava Lists

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/GetAdapterRunsCommand.java
```diff
@@ -0,0 +1,84 @@
...
+    long now = System.currentTimeMillis();
+    long startTs = getTimestamp(arguments.get(ArgumentName.START_TIME.toString(), "min"), now);
+    long endTs = getTimestamp(arguments.get(ArgumentName.END_TIME.toString(), "max"), now);
+    String resultLimitStr = arguments.get(ArgumentName.LIMIT.toString(), "");
+    Integer resultLimit = resultLimitStr.isEmpty() ? null : Integer.parseInt(resultLimitStr);
+
+    List<RunRecord> runs = adapterClient.getRuns(adapterName, adapterStatus, startTs, endTs, resultLimit);
+
+    Table table = Table.builder()
+      .setHeader("pid", "status", "twillRunId", "start", "stop")
```
gokulavasan: Should this be RunId instead of twillRunId?

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/DescribeAdapterCommand.java
```diff
@@ -0,0 +1,78 @@
...
+    this.adapterClient = adapterClient;
+  }
+
+  @Override
+  public void perform(Arguments arguments, PrintStream output) throws Exception {
+    String adapterName = arguments.get(ArgumentName.ADAPTER.toString());
+    AdapterDetail spec = adapterClient.get(adapterName);
+
+    Table table = Table.builder()
+      .setHeader("name", "description", "template", "config", "schedule")
```
gokulavasan: so for realtime adapters, schedule column will have null? What about "instances" which is the Realtime adapter's property similar to "schedule" for Batch adapter's property.

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/ListAdaptersCommand.java
```diff
@@ -51,12 +51,13 @@ public void perform(Arguments arguments, PrintStream output) throws Exception {
     List<AdapterDetail> list = adapterClient.list();
 
     Table table = Table.builder()
-      .setHeader("name", "description", "template", "config")
+      .setHeader("name", "description", "template", "config", "schedule")
```
gokulavasan: Schedule for Batch Adapter and instances for Realtime Adapter?

cdap-client/src/main/java/co/cask/cdap/client/AdapterClient.java
```diff
@@ -68,14 +76,14 @@ public AdapterClient(ClientConfig config) {
   /**
    * Lists all adapters.
    *
-   * @return list of {@link AdapterDetail}.
+   * @return list of {@link co.cask.cdap.api.templates.AdapterSpecification}.
```
gokulavasan: Why full class name?

cdap-proto/src/main/java/co/cask/cdap/proto/ProgramId.java
```diff
@@ -0,0 +1,103 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.proto;
+
+/**
+ * Represents an adapter returned for /adapters/{adapter-id}.
```
gokulavasan: Is the right javadoc? If yes, then ProgramId should be AdapterProgramId right?

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/GetAdapterRunsCommand.java
jwang47: `pid` is our run id, whereas `twillRunId` is from Twill. See https://github.com/caskdata/cdap/blob/develop/cdap-proto/src/main/java/co/cask/cdap/proto/RunRecord.java. Not sure how to present this info.

gokulavasan: Oh ok. Hmm, so we have two RunIds?

gokulavasan:
Few comments. Rest LGTM :+1:

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/DescribeAdapterCommand.java
jwang47: "instances" should be exposed through `AdapterDetail`. Will fix this.

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/ListAdaptersCommand.java
jwang47: How do I determine which properties should be listed from here though? How about having a `properties` column which has stuff like "schedule=.." if schedule is not null and "instances=.." if instances is not null?

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/GetAdapterRunsCommand.java
gokulavasan: @sagarkapare confirmed that we need not show the twillRunId. Since it is internal and not of any use to the user.

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/ListAdaptersCommand.java
gokulavasan: Oh may be have two columns. Null for one and has info for another?

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/ListAdaptersCommand.java
```diff
@@ -51,12 +52,20 @@ public void perform(Arguments arguments, PrintStream output) throws Exception {
     List<AdapterDetail> list = adapterClient.list();
 
     Table table = Table.builder()
-      .setHeader("name", "description", "template", "config")
+      .setHeader("name", "description", "template", "config", "properties")
```
gokulavasan: This will be RunRecord's properties?

cdap-watchdog/src/main/java/co/cask/cdap/logging/gateway/handlers/LogHandler.java
```diff
@@ -110,16 +109,24 @@ private void doGetLogs(HttpResponder responder, LoggingContext loggingContext,
...
                          @Nullable RunRecord runRecord) {
     try {
-      Filter filter = FilterParser.parse(filterStr);
-      long fromTimeMs = TimeUnit.MILLISECONDS.convert(fromTimeSecsParam, TimeUnit.SECONDS);
-      long toTimeMs = TimeUnit.MILLISECONDS.convert(toTimeSecsParam, TimeUnit.SECONDS);
 
-      if (fromTimeMs < 0 || toTimeMs < 0 || toTimeMs <= fromTimeMs) {
-        responder.sendString(HttpResponseStatus.BAD_REQUEST, "Invalid time range. 'start' and 'stop' should be " +
-          "greater than zero and 'stop' should be greater than 'start'.");
+      if (fromTimeSecsParam < -1 || toTimeSecsParam < 0) {
```
gokulavasan: So, fromTimeSecsParam = -1, toTimeSecsParam=0 is a valid combination?

gokulavasan:
One minor comment. Rest LGTM :+1:

cdap-watchdog/src/main/java/co/cask/cdap/logging/gateway/handlers/LogHandler.java
jwang47: Yes, actually default for fromTimeSecsParam is -1 if you don't provide it to the handler.

PR #2346, opened 2015-04-27
Ui/fix c3 memory leak

anwar6953:
1) Remove resize event handling from c3 charts because c3 has its own onResize event handling.
2) Destroy c3 chart when scope is destroyed (helps reduce memory leakage by a lot).


elfenheart:
:+1: LGTM

PR #2348, opened 2015-04-27
Fix/ui widget simplification

ajainarayanan:
- Simplifies widget-container to have a simpler way to add a new widget.
- Adds a WidgetFactory registry.
   - Any newly added widget can add itself to the registry and if need additional functionality can create its own controller. It should ideally be like creating a directive which accepts an ng-model and optional attributes it requires and render. Once there is a change in the model it should be serialized (say we show a date time picker, once the user choose a date/time we should update the model with the respective epoch time).
- Provides javascript-editor as a simple example for creating custom widgets.

### DIFF
--- a/cdap-ui/app/directives/widget-container/widget-factory.js
+++ b/cdap-ui/app/directives/widget-container/widget-factory.js
@@ -21,7 +21,9 @@ angular.module(PKG.name + '.commons')
       'json-editor': {
         element: '<textarea></textarea>',
         attributes: {
-          'cask-json-edit': 'model'
+          'cask-json-edit': 'model',
+          'class': 'form-control',
+          placeholder: 'myconfig.description'
         }
       },
       'javascript-editor': {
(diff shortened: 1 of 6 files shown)


cdap-ui/app/directives/widget-container/widget-factory.js
```diff
@@ -0,0 +1,35 @@
...
+          'ng-model': 'model',
+          type: 'password',
+          placeholder: 'myconfig.description'
+        }
+      },
+      'json-editor': {
+        element: '<textarea></textarea>',
+        attributes: {
+          'cask-json-edit': 'model'
+        }
```
elfenheart: Don't this need class form-control as well?

ajainarayanan: Fixed.

elfenheart:
:+1: LGTM

PR #2349, opened 2015-04-27
Fix/ui dashboard columns

elfenheart:
fix bug with changeColumn calling on different dashboard

### DIFF
--- a/cdap-ui/app/features/dashboard/controllers/dashboard-ctrl.js
+++ b/cdap-ui/app/features/dashboard/controllers/dashboard-ctrl.js
@@ -105,16 +105,20 @@ function ($scope, $state, $dropdown, rDashboardsModel, MY_CONFIG) {
   }
 
   $scope.changeColumn = function (n) {
-    $scope.currentBoard.changeColumn(n);
+    rDashboardsModel.current().changeColumn(n);
   };
 
   $scope.toggleDragDrop = function() {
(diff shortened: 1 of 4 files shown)


cdap-ui/app/features/dashboard/controllers/dashboard-ctrl.js
```diff
@@ -105,16 +105,20 @@ function ($scope, $state, $dropdown, rDashboardsModel, MY_CONFIG) {
   }
 
   $scope.changeColumn = function (n) {
-    $scope.currentBoard.changeColumn(n);
```
anwar6953: Should we remove this attribute off the scope object, if it can be outdated?
Alternative is to keep it updated?

elfenheart: No, because $scope.currentBoard is being used in different context

nitinmotgi:
Any update @elfenheart @ajainarayanan about this PR.

ajainarayanan:
:+1: LGTM. Talked to @elfenheart We can fix this later. Not a priority to stall this change. 

Original Issue:
  The ```$scope.currentBoard``` that we pass to the dropdown doesn't get updated on dashboard switch as its a javascript object (reference). So the currentBoard is being re-used even when we change the dashboard. This needs to be fixed but for now we are directly asking the model to give the current dashboard every time we click the tab header for the dropdown.

PR #2352, opened 2015-04-28
CDAP-2322: Set formatSpec in prepareJob

rohitsinha54:
Tested it with source as Stream and sink as Table and KVTable.
Build: http://builds.cask.co/browse/CDAP-DT62-1


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/StreamBatchSource.java
```diff
@@ -98,6 +98,18 @@ public StreamBatchSource(StreamBatchConfig streamBatchConfig) {
 
   @Override
   public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineConfigurer) {
+    pipelineConfigurer.addStream(new Stream(streamBatchConfig.name));
```
albertshau: still want to check the schema here. If it's an invalid schema we want to fail early at adapter creation

albertshau: similar comment about duration and delay. If we can't parse, should fail here.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/batch/sources/StreamBatchSource.java
```diff
@@ -119,17 +131,6 @@ public void configurePipeline(ETLStage stageConfig, PipelineConfigurer pipelineC
     } else {
       this.formatSpec = null;
```
albertshau: there is no need for formatSpec to be a class variable, should be local to this method.

albertshau:
mostly comments about failing early

rohitsinha54:
LGTM

PR #2354, opened 2015-04-28
CDAP-2224: measure type per each measure in cube query

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-RBT234-1

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricQueryParser.java
```diff
@@ -154,7 +153,8 @@ static MetricDeleteQuery parseDelete(URI requestURI, String metricPrefix) throws
     builder.setMetricName(metricPrefix);
 
     MetricDataQuery query = builder.build();
-    return new MetricDeleteQuery(query.getStartTs(), query.getEndTs(), query.getMetricNames(), query.getSliceByTags());
+    return new MetricDeleteQuery(query.getStartTs(), query.getEndTs(),
+                                 query.getMetrics().keySet(), query.getSliceByTags());
```
gsps1: this class should have been removed right ? as its for v2 API.

gsps1:
LGTM :+1:

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricQueryParser.java
abaranau: we do still deletion of metrics. not thru REST though

PR #2355, opened 2015-04-28
CDAP-2324: Allow specifying aggregation function per measurement in CubeQuery

abaranau:
Based on https://github.com/caskdata/cdap/pull/2354


abaranau:
http://builds.cask.co/browse/CDAP-RBT235-1

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/cube/AggregationFunction.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.dataset.lib.cube;
+
+/**
+ * Function to be used when aggregating multiple
```
gsps1: multiple measures

gsps1:
LGTM :+1:  but would be good to have some test cases for the new aggregation functions though.

abaranau:
added some bits of unit-testing. We'll need more though for Cube: https://issues.cask.co/browse/CDAP-2326

PR #2357, opened 2015-04-28
Check for null keys when pulling messages for Kafka topic.

henry-cask:
Kafka messages could have null keys, so we need to check before calling KafkaSimpleApiConsumer#decodeKey method when processing the messages.

Also removed some unused imports.


henry-cask:
Successful build: http://builds.cask.co/browse/CDAP-RT8-1

albertshau:
:+1:

henry-cask:
Thanks @albertshau =)

PR #2358, opened 2015-04-28
CDAP-2199: Builder to construct CubeQuery

abaranau:


gsps1:
I was trying it out, 
``CubeQuery.builder().select().measurement("metrics", AggregationFunction.SUM).from(null)
      .resolution(1, TimeUnit.HOURS).where().
      dimensions(ImmutableMap.<String, String>of()).
      timeRange(100, 800).groupBy().dimension("abc").limit(10).
      interpolator(new Interpolators.Step()).build();``

1) would be good to have the from() as optional , inferring that aggregation is null. 
2) measurement - accept a map of <String, AggregateFun>

gsps1:
just those two points, otherwise LGTM :+1:

PR #2361, opened 2015-04-28
Feature/ui etladapters detailedview

ajainarayanan:
- Adds detailed view of an adapter.
   - Modifies routes to add detailed view
   - Adds Runs status + Logs for each run
   - Adds History for each adapter detailed view.


elfenheart:
:+1: LGTM!

PR #2362, opened 2015-04-28
Fix/ui etlapps adapters rename

ajainarayanan:
Renames all 'etl*' stuff to 'adapters' to be consistent with the feature naming.


elfenheart:
:+1: LGTM

PR #2365, opened 2015-04-28
fix script transform compilation

albertshau:


albertshau:
http://builds.cask.co/browse/CDAP-RT9

bdmogal:
:+1:

PR #2368, opened 2015-04-28
temporarily disabling test that fails on java that bamboo uses.

albertshau:


albertshau:
http://builds.cask.co/browse/CDAP-RT11

sreevatsanraman:
LGTM :+1:

PR #2369, opened 2015-04-28
cdap-api should not be in provided scope

rohitsinha54:


rohitsinha54:
@albertshau : Have a look.

bdmogal:
:+1: since compilation fails without this, but please check with @chtyim about this.

chtyim:
It's because the multi-module dependency shouldn't be in dependency management. I'll fix the pom properly.

PR #2372, opened 2015-04-28
Fix the dependency on cdap-api for the ETL app.

chtyim:
- For modules that build bundle jar, we should explicitly declare dependency on cdap-api and have it as provided scope.
- For inter-module dependency, shouldn’t use dependency management.


henry-cask:
+1
LGTM

PR #2374, opened 2015-04-28
Fix/ui styling modal close

astewes:


cdap-ui/app/styles/themes/cdap/theme.less
```diff
@@ -364,14 +369,15 @@ body.theme-cdap {
...
-          font-size: 24px;
+          .cask-btn(@color: @cdap-header, @background: white, @border: 0, @border-radius: 16px, @padding: 5px 7px);
+          font-size: 18px;
           line-height: 1;
         }
       }
       // Small and large modal styling
       &.modal-sm, &.modal-lg {
         color: white;
+        h3, h4 { color: white; }
```
ajainarayanan: We already have this under .modal right? do we need to explicitly add it to .model-sm and .model-lg too?

ajainarayanan:
:+1: LGTM

PR #2375, opened 2015-04-28
read scp vars from environment

dereklwood:
- [x] read parcel SCP settings from environment (set by bamboo) instead of a local file (one less dependency for buildagent setup)


wolf31o2:
:+1: LGTM

PR #2376, opened 2015-04-28
The realtime module shouldn’t haven’t dependency on avro.

chtyim:
- Only plugin (etl-lib) uses avro classes.


rohitsinha54:
LGTM

PR #2377, opened 2015-04-28
Fixes the inclusion of examples and guides.

awholegunch:
Tested under OS X and CentOS 6.

- Fixed style of variable reference usage to always have variables in brace-brackets.
- Remove an unused function.
- Changed the test_an_include function to be a generic test function 
  that does not manipulate a relative path.
- Changed the version function to determine a branch type taking into account the parent branch.
- Branches types are either *develop, master, release, develop-feature, release-feature*, depending on branch and parent
- In manuals, changed the branch tests to use the new type recognition to set the includes correctly.


cdap-docs/_common/common-build.sh
```diff
@@ -244,13 +244,13 @@ function set_mvn_environment() {
 }
 
 function check_includes() {
-  if [ $CHECK_INCLUDES == $TRUE ]; then
+  if [ ${CHECK_INCLUDES} == ${TRUE} ]; then
```
duddun: Let us enclose the left hand side of the == in double quotes.

duddun:
LGTM -- Looks Ok to me. I will test it in Bamboo once it is merged

PR #2378, opened 2015-04-28
Split sections into VM-only and common... prep for Docker

wolf31o2:
This splits provisioners into smaller chunks, based on their use. Sections which are specific to the VM use case, such as installing and configuring X11, IntelliJ, and Firefox, are configured to only run for that build.


duddun:
LGTM

wolf31o2:
GRR... I made this PR against the wrong branch.

PR #2379, opened 2015-04-28
Ordering the Data and Apps on Overview page

anwar6953:
Ordering the Data and Apps on Overview page by most recently clicked first.
This is useful especially because we limit the number of elements shown.

Whenever someone clicks on an app or data element on the overview page, 'All apps' page, or 'App Datasets' page, it gets marked as most recently used. This information is then used to order the limited-length lists on the Overview page.

Click the gif for a larger view.

### DIFF
--- a/cdap-ui/app/features/apps/controllers/list-ctrl.js
+++ b/cdap-ui/app/features/apps/controllers/list-ctrl.js
@@ -1,5 +1,6 @@
 angular.module(PKG.name + '.feature.apps')
-  .controller('CdapAppListController', function CdapAppList( $timeout, $scope, MyDataSource, myAppUploader, $alert, $state) {
+  .controller('CdapAppListController', function CdapAppList( $timeout, $scope, MyDataSource, myAppUploader, $alert, $state, MyOrderings) {
+    $scope.MyOrderings = MyOrderings;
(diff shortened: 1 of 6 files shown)


cdap-ui/app/services/appsOrdering.js
```diff
@@ -0,0 +1,52 @@
+angular.module(PKG.name+'.services')
+  .service('myOrderings', function(myLocalStorage) {
+    var APP_KEY = 'appOrdering';
+    var DATA_KEY = 'dataOrdering';
+    myLocalStorage.get(APP_KEY).then(function(value) {
+      if (value === undefined) {
```
ajainarayanan: Could be better if we have ```typeof value === 'undefined'```

anwar6953: Since `value` is declared as the argument, isn't that not needed?

ajainarayanan: Just for reference http://stackoverflow.com/a/4726198/661768

anwar6953: updated & pushed.

ajainarayanan:
:+1: LGTM

PR #2381, opened 2015-04-28
Fixes editing properties of plugin with no hard dependency on configuration.

ajainarayanan:
- Enables editing properties of a plugin with no configuration served from nodejs
- Enables applying configuration to transform properties if present.
- Handles the use case of no property for a plugin.


cdap-ui/app/features/adapters/services/etl-create-factory.js
```diff
@@ -53,13 +53,12 @@ angular.module(PKG.name + '.feature.adapters')
           var s = res[0];
           this.scope.source.name = s.name;
           var obj = {};
-          angular.forEach(source.properties, function(property) {
+          angular.forEach(s.properties, function(property) {
             obj[property.name] = '';
           });
+          debugger;
```
elfenheart: remove? :smile:

elfenheart:
Minor comment, otherwise :+1: LGTM

cdap-ui/app/features/adapters/services/etl-create-factory.js
ajainarayanan: Wow. Good catch.

PR #2385, opened 2015-04-29
Fix the 'stop-polling' websocket event. The timerId of the resource was ...

anwar6953:
...being cleared (in order to avoid sending it to browser), and as a side effect, we lost a handle the timer id.

To avoid that, just avoid pushing these fields in the JSON, but leave them on the resource object.


ajainarayanan:
:+1: LGTM

PR #2388, opened 2015-04-29
Split sections into VM-only and common... prep for Docker

wolf31o2:
This is the same as #2378 but against the `release/3.0` branch.


wolf31o2:
Since this was already approved in the other PR and the changes are identical, I am merging this.

PR #2389, opened 2015-04-29
Fix CubeQuery Javadoc.

awholegunch:
Fixes broken Javadocs by surrounding code sample in \<code> and \</code> in addition to \<pre> and \</pre>.

The problem can be seen in the currently staged javadocs: [CubeQuery](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT/en/reference-manual/javadocs/co/cask/cdap/api/dataset/lib/cube/CubeQuery.html)

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/cube/CubeQuery.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/cube/CubeQuery.java
@@ -31,13 +31,15 @@ import javax.annotation.Nullable;
  * </p>
  * Another way to think about the query is to map it to the following statement::
  * <pre>
+ * <code>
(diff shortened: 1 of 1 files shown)


abaranau:
After change:

screen shot 2015-05-01 at 11 12 53 am

abaranau:
BEfore change it was good. Are you sure we need the change?

awholegunch:
It breaks the Javadoc build. Look at this link: [CubeQuery](http://stg-web101.sw.joyent.continuuity.net/cdap/3.0.0-SNAPSHOT/en/reference-manual/javadocs/co/cask/cdap/api/dataset/lib/cube/CubeQuery.html). Let me see if I can get it to work with both.

abaranau:
LGTM

PR #2390, opened 2015-04-29
CDAP-2317 Update Dockerfile

wolf31o2:
- [x] Switch to using the scripts for the VM build via Packer to increase reuse
- [x] Remove build from source for SDK, use release artifacts
- [x] Move to `cdap-distributions/src` to reduce build context


duddun:
LGTM

PR #2391, opened 2015-04-29
ETL CSV Widget

elfenheart:
Adding DSV and CSV

screen shot 2015-04-29 at 11 39 26 am

### DIFF
--- a/cdap-ui/app/directives/widget-container/widget-container.less
+++ b/cdap-ui/app/directives/widget-container/widget-container.less
@@ -17,5 +17,6 @@
 
   textarea {
     resize: vertical;
+    min-height: 100px;
   }
 }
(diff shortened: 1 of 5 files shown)


nitinmotgi:
this is not right! @ajainarayanan this is not how we had talked about right.

cdap-ui/app/directives/widget-container/widget-container.less
```diff
@@ -10,5 +10,6 @@
   }
   textarea {
     resize: vertical;
+    min-height: 100px;
```
ajainarayanan: Do we still need this?

ajainarayanan:
One minor comment. Apart from that :+1: LGTM

ajainarayanan:
Talked to @nitinmotgi the textarea has been replaced with textboxes and a delimiter for the user to configure.

PR #2395, opened 2015-04-29
CDAP-2267 set instances to 1 if null, if user provides then assert > 0

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-2267


albertshau:
:+1:

PR #2396, opened 2015-04-29
Fix creating new StructuredRecord builder per conversion.

henry-cask:
The StructuredRecord builder need to be created per instance of StructuredRecord because calling the build method does not reset the internal fields.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/KafkaSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/KafkaSource.java
@@ -58,9 +58,11 @@ public class KafkaSource extends RealtimeSource<StructuredRecord> {
   public static final String KAFKA_BROKERS = "kafka.brokers";
   public static final String KAFKA_DEFAULT_OFFSET = "kafka.default.offset";
 
-  private KafkaSimpleApiConsumer kafkaConsumer;
+  private static final Schema SCHEMA = Schema.recordOf("Kafka Message",
+                                                       Schema.Field.of(MESSAGE, Schema.of(Schema.Type.BYTES)),
+                                                       Schema.Field.of(KEY, Schema.of(Schema.Type.STRING)));
 
-  private StructuredRecord.Builder recordBuilder;
+  private KafkaSimpleApiConsumer kafkaConsumer;
 
   private KafkaPluginConfig config;
 
@@ -78,10 +80,6 @@ public class KafkaSource extends RealtimeSource<StructuredRecord> {
 
     kafkaConsumer = new Kafka08SimpleApiConsumer(this);
     kafkaConsumer.initialize(context);
-
-    Schema.Field messageField = Schema.Field.of(MESSAGE, Schema.of(Schema.Type.BYTES));
-    Schema.Field keyField  = Schema.Field.of(KEY, Schema.of(Schema.Type.STRING));
-    recordBuilder = StructuredRecord.builder(Schema.recordOf("Kafka Message", messageField, keyField));
   }
 
   @Nullable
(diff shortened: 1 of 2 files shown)


henry-cask:
http://builds.cask.co/browse/CDAP-RT16-3

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/kafka/Kafka08SimpleApiConsumer.java
```diff
@@ -238,13 +238,13 @@ public void destroy() {
   }
 
   @Override
-  protected void processMessage(String key, ByteBuffer payload, Emitter<StructuredRecord> emitter) {
+  protected void processMessage(String key, byte[] payload, Emitter<StructuredRecord> emitter) {
```
chtyim: Why change the type to `byte[]`. As far I as know, the Kafka message has payload in ByteBuffer, decoding ByteBuffer into byte[] almost always involve array creation and copying (because of the position and limit), while passing ByteBuffer around is pretty simple and the overhead is usually pretty small.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/KafkaSource.java
```diff
@@ -108,11 +105,12 @@ public SourceState poll(Emitter<StructuredRecord> writer, SourceState currentSta
    * @param payload the ByteBuffer of the Kafka message.
    * @return instance of {@link StructuredRecord} representing the message.
    */
-  public StructuredRecord byteBufferToStructuredRecord(@Nullable String key, ByteBuffer payload) {
+  public StructuredRecord byteBufferToStructuredRecord(@Nullable String key, byte[] payload) {
+    StructuredRecord.Builder recordBuilder = StructuredRecord.builder(SCHEMA);
     if (key != null) {
       recordBuilder.set(KEY, key);
     }
-    recordBuilder.set(MESSAGE, payload.array());
```
chtyim: Using `ByteBuffer.array()` is actually wrong. We should always take `arrayOffset`, `position` and `remaining` into account.

henry-cask: Yes, I should not have used ByteBuffer.array to set the message field. 
That is why I was thinking about sending the byte array instead of ByteBuffer to simplify the downstream consumption of KafkaSource output.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/kafka/Kafka08SimpleApiConsumer.java
henry-cask: I was thinking to just pass the bytes array to make sure dowan stream code to avoid consider position and remaining of the ByteBuffer instead of calling ByteBuffer.array by mistake.
On the other hand, we could just assume dowan stream sinks knew how to deal with ByteBuffer properly.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/KafkaSource.java
henry-cask: I reverted the byte array fix so lets stay with serializing the payload as ByteBuffer.

chtyim:
LGTM

henry-cask:
@chtyim Thanks!

PR #2401, opened 2015-04-29
Ui/restrict chart timerange

anwar6953:
...s.
Limit the number of widgets on a dashboard to 12.

Time Range Limit

-------
-------
Widget Count Limit


ajainarayanan:
:+1: LGTM.. One minor comment about moving WIDGET_LIMIT to the model. But apart from that it can be merged.

PR #2407, opened 2015-04-29
Remove static variable for JmsPluginConfig class in JmsSource

henry-cask:
Need to remove DEFAULT_MESSAGE_RECEIVE static variable in JmsPluginConfig since it is showing up to UI via reflection.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
@@ -279,25 +279,24 @@ public class JmsSource extends RealtimeSource<StructuredRecord> {
    * Config class for {@link JmsSource}.
    */
   public static class JmsPluginConfig extends PluginConfig {
-    private static final Integer DEFAULT_MESSAGE_RECEIVE = 50;
-
     @Name(JMS_DESTINATION_NAME)
     @Description("Name of the destination to get messages")
-    private final String destinationName;
+    private String destinationName;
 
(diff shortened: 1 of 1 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
```diff
@@ -279,25 +279,20 @@ public JmsProvider getJmsProvider() {
...
     @Name(JMS_DESTINATION_NAME)
     @Description("Name of the destination to get messages")
-    private final String destinationName;
+    private String destinationName;
 
     @Name(JMS_MESSAGES_TO_RECEIVE)
     @Description("Max number messages should be retrieved per poll. The default value is 50.")
     @Nullable
-    private final Integer messagesToReceive;
+    private Integer messagesToReceive = 50;
```
gokulavasan: I think you need to have a default constructor that assigns this value to messagesToReceive.

henry-cask:
Successful build: http://builds.cask.co/browse/CDAP-RT17-1

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
henry-cask: Ah ok, we need to document this behavior

henry-cask:
@gokulavasan Updated the patch with default constructor.

gokulavasan:
LGTM :+1:

henry-cask:
Thanks @gokulavasan, merging it now.

PR #2408, opened 2015-04-29
(CDAP-2366) call destroy() appropriately

chtyim:
- Introduced Destroyable to unify destroy call.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Destroyable.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Destroyable.java
@@ -17,31 +17,12 @@
 package co.cask.cdap.templates.etl.api;
 
 /**
- * Transform Stage.
- *
- * @param <IN> Type of input object
- * @param <OUT> Type of output object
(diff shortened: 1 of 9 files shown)


gokulavasan:
LGTM :+1:

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Destroyables.java
```diff
@@ -0,0 +1,39 @@
...
+
+package co.cask.cdap.templates.etl.api;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Helper class to deal with {@link Destroyable}.
+ */
+public final class Destroyables {
```
gokulavasan: Does this need to be in etl-api?

gokulavasan:
One comment. Rest LGTM :+1:

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/templates/etl/api/Destroyables.java
chtyim: Moved

PR #2409, opened 2015-04-29
CDAP-2038 Allow users to configure a widget's metric to have only the na...

anwar6953:
...mespace tag.

https://issues.cask.co/browse/CDAP-2038


ajainarayanan:
:+1: LGTM

PR #2410, opened 2015-04-29
Fix/ui dashboard inputs grid

astewes:


anwar6953:
Hey @astewes 
Can the dropdown be the same width as the text?
image

anwar6953:
Also, in the staticdashboard.html, there is no longer a gap between above the widgets.

image

PR #2411, opened 2015-04-30
moves tooltip icon to correct position

astewes:


nitinmotgi:
Closing PR as we have moved to use bs-popover.

PR #2412, opened 2015-04-30
(CDAP-2369) Scan all package resources to find plugin classes

chtyim:
- When the plugin jar has jar inside, it’s possible that there is more than one URL that matches with the package resource. We should search for the one that is in local directory (vs in JAR) and pickup Plugin classes form there.


bdmogal:
:+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT240-1

PR #2420, opened 2015-04-30
Fixes making backend call for logs when we have no runs in flows, workfl...

ajainarayanan:
...ows & mapreduce


elfenheart:
:+1: LGTM

ajainarayanan:
Will add the same change for adapters in the same PR

PR #2426, opened 2015-04-30
Fix log output for the auto build of docs

awholegunch:
These changes improve the output of the logs for the building of documentation.

There is now a clear start and finish header, and all headers follow the same pattern: a string of "=" followed by a string of "-". That way you can search for either pattern, and "hop" from header to header.

Additional quote marks have been added to make it clear which targets are being built.


duddun:
LGTM

PR #2431, opened 2015-04-30
Fix batch source archetype

sreevatsanraman:
* Batch source archetype uses realtime. This PR addresses that.


gokulavasan:
LGTM :+1:

PR #2432, opened 2015-04-30
[CDAP-2330] Disallow hyphens in namespace names

bdmogal:
Disallows hyphens in namespace names only for now. Retains current behavior for other ids. 
This is needed because hive complains if the database name has a hyphen in it.

Jira: [CDAP-2330](https://issues.cask.co/browse/CDAP-2330)
Build: http://builds.cask.co/browse/CDAP-RBT242

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
@@ -105,6 +105,15 @@ public class NamespaceHttpHandler extends AbstractAppFabricHttpHandler {
   @PUT
   @Path("/namespaces/{namespace-id}")
   public void create(HttpRequest request, HttpResponder responder, @PathParam("namespace-id") String namespaceId) {
+    Id.Namespace namespace;
+    try {
+      namespace = Id.Namespace.from(namespaceId);
+    } catch (IllegalArgumentException e) {
+      responder.sendString(HttpResponseStatus.BAD_REQUEST,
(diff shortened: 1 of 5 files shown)


anew:
wouldn't it be easy to replace them with _ ? Just wondering what we are doing for dataset names, because Hive probably does not allow hyphens in tables names ewther

bdmogal:
Yes, we replace '.' and '-' with '_' for Hive table names. We can do the same for hive databases too. Will update the PR.

bdmogal:
Updated the PR to only replace hyphens with underscores in the hive database name.

```
cdap (http://Bhooshans-MacBook-Pro.local:10000/default)> create namespace my-namespace
Namespace 'my-namespace' created successfully.

cdap (http://Bhooshans-MacBook-Pro.local:10000/default)> execute 'show databases'
+=======================+
| database_name: STRING |
+=======================+
| cdap_my_namespace     |
| default               |
+=======================+
Fetched 2 rows
cdap (http://Bhooshans-MacBook-Pro.local:10000/default)> use namespace my-namespace
Now using namespace 'my-namespace'

cdap (http://Bhooshans-MacBook-Pro.local:10000/my-namespace)> deploy app ../../../../cdap-examples/Purchase/target/Purchase-3.0.0-SNAPSHOT.jar
Successfully deployed application

sreevatsanraman:
Changes LGTM. Please merge once the testing is done on standalone and cluster.

bdmogal:
Hbase also does not allow hyphens in namespace names. I've reverted my last commit to only allow ``[a-zA-Z0-9_] as the character set for namespaces.

Tested on a cluster. Will wait for a successful build before I merge.

bdmogal:

PR #2436, opened 2015-04-30
CDAP-2297 make adapter operations synchronized to protect against

albertshau:
race conditions. There are better ways to do this, but doing
something simple in the short term.


albertshau:
http://builds.cask.co/browse/CDAP-RBT244

sreevatsanraman:
LGTM :+1:

PR #2440, opened 2015-04-30
Make the key to become optional for schema definition in Kafka source.

henry-cask:
Since key could be null, need to make the StructuredRecord's schema to be nullable for Kafka realtime source plugin.


gokulavasan:
LGTM :+1:

henry-cask:
Thanks @gokulavasan

PR #2442, opened 2015-04-30
Batch Source/Sink initialize method takes in Context

gokulavasan:
Batch Source and Batch Sink should have access to their Contexts in the initialize method instead of just PluginProperties. 

JIRA : https://issues.cask.co/browse/CDAP-2368

Build : http://builds.cask.co/browse/CDAP-RT18

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
@@ -124,8 +124,8 @@ public class ETLMapReduce extends AbstractMapReduce {
     public void initialize(MapReduceContext context) throws Exception {
       Map<String, String> runtimeArgs = context.getRuntimeArguments();
       ETLBatchConfig etlConfig = GSON.fromJson(runtimeArgs.get(Constants.CONFIG_KEY), ETLBatchConfig.class);
-      String sourceId = runtimeArgs.get(Constants.Source.PLUGINID);
-      String sinkId = runtimeArgs.get(Constants.Sink.PLUGINID);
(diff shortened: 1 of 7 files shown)


cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
```diff
@@ -136,15 +136,17 @@ public void initialize(MapReduceContext context) throws Exception {
       List<StageMetrics> stageMetrics = Lists.newArrayListWithCapacity(stageList.size() + 2);
       transforms = Lists.newArrayListWithCapacity(stageList.size());
 
-      BatchSource source = context.newPluginInstance(sourceId);
-      source.initialize(context.getPluginProperties(sourceId));
+      BatchSource source = context.newPluginInstance(sourcePluginId);
+      BatchSourceContext batchSourceContext = new MapReduceSourceContext(context, mapperMetrics, sourcePluginId);
```
anew: so as a side effect, the source can now also emit custom metrics?

anew:
For consistency, should Transform get a TransformContext in initialize(). To make the naming consistent?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/main/java/co/cask/cdap/templates/etl/batch/ETLMapReduce.java
gokulavasan: Yes. Earlier batch sources/sinks were only configuring hadoop job and nothing else. now, since they have initialize, transform, destroy methods, emitting custom metrics would be useful.

gokulavasan:
@anew Yes, we should rename StageContext to TransformContext. Will do that in a separate PR.

gokulavasan:
Build passed. @anew Addressed your questions/comments.

anew:
LGTM

PR #2443, opened 2015-04-30
For MRJobinfo endpoint, call Metrics System directly in Standalone/Inmem...

anwar6953:
...ory (avoid JobClient because we know it will not exist)

http://builds.cask.co/browse/CDAP-RBT245-4
http://issues.cask.co/browse/CDAP-2359

Tested on Standalone. Testing in progress on Distributed.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/AppFabricServiceRuntimeModule.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/AppFabricServiceRuntimeModule.java
@@ -17,6 +17,9 @@ package co.cask.cdap.app.guice;
 
 import co.cask.cdap.app.deploy.Manager;
 import co.cask.cdap.app.deploy.ManagerFactory;
+import co.cask.cdap.app.mapreduce.MRJobClientWithMetricsFallback;
+import co.cask.cdap.app.mapreduce.MRJobInfoFetcher;
+import co.cask.cdap.app.mapreduce.MapReduceMetricsInfo;
 import co.cask.cdap.app.store.Store;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
@@ -126,6 +129,7 @@ public final class AppFabricServiceRuntimeModule extends RuntimeModule {
                              protected void configure() {
                                bind(SchedulerService.class).to(LocalSchedulerService.class).in(Scopes.SINGLETON);
                                bind(Scheduler.class).to(SchedulerService.class);
+                               bind(MRJobInfoFetcher.class).to(MapReduceMetricsInfo.class);
 
                                MapBinder<String, MasterServiceManager> mapBinder = MapBinder.newMapBinder(
                                  binder(), String.class, MasterServiceManager.class);
@@ -172,6 +176,7 @@ public final class AppFabricServiceRuntimeModule extends RuntimeModule {
                                bind(AppFabricServer.class).to(StandaloneAppFabricServer.class).in(Scopes.SINGLETON);
                                bind(SchedulerService.class).to(LocalSchedulerService.class).in(Scopes.SINGLETON);
                                bind(Scheduler.class).to(SchedulerService.class);
+                               bind(MRJobInfoFetcher.class).to(MapReduceMetricsInfo.class);
 
                                MapBinder<String, MasterServiceManager> mapBinder = MapBinder.newMapBinder(
                                  binder(), String.class, MasterServiceManager.class);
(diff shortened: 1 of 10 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/AppFabricServiceRuntimeModule.java
```diff
@@ -217,6 +222,7 @@ public Module getDistributedModules() {
                              protected void configure() {
                                bind(SchedulerService.class).to(DistributedSchedulerService.class).in(Scopes.SINGLETON);
                                bind(Scheduler.class).to(SchedulerService.class);
+                               bind(MRJobInfoFetcher.class).to(MRJobClientWithMetricsFallback.class);
```
chtyim: Shouldn't the implementation class call `LocalMRJobInfoFetcher` and `DistributedMRJobInfoFetcher`?

anwar6953: I figured the name should be more closely related to its implementation.
So `Metrics` in local and `JobClient` in distributed.
Is that not a good idea?

chtyim: A class name should ties to what it does, not necessarily how is it implemented, hence I think `DistributedMRJobInfoFetcher` is better in this case, unless you are going to have multiple implementations of `MRJobInfoFetcher` for distributed mode.

anwar6953: Updated.

cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/AppFabricServiceRuntimeModule.java
```diff
@@ -217,6 +222,7 @@ public Module getDistributedModules() {
                              protected void configure() {
                                bind(SchedulerService.class).to(DistributedSchedulerService.class).in(Scopes.SINGLETON);
                                bind(Scheduler.class).to(SchedulerService.class);
+                               bind(MRJobInfoFetcher.class).to(DistributedMrJobInfoFetcher.class);
```
chtyim: MR, not Mr.

chtyim: Still not fixed

cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/AppFabricServiceRuntimeModule.java
```diff
@@ -217,6 +222,7 @@ public Module getDistributedModules() {
                              protected void configure() {
                                bind(SchedulerService.class).to(DistributedSchedulerService.class).in(Scopes.SINGLETON);
                                bind(Scheduler.class).to(SchedulerService.class);
+                               bind(MRJobInfoFetcher.class).to(DistributedMRJobInfoFetcher.class);
```
chtyim: Oh.. now is fine. Browser refresh issue.

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/DistributedMrJobInfoFetcher.java
```diff
@@ -0,0 +1,63 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+
+/**
+ * Retrieves information about a run of a MapReduce job, using {@link MRJobClient} and
+ * then {@link LocalMRJobInfoFetcher} if necessary.
+ */
+public class DistributedMRJobInfoFetcher implements MRJobInfoFetcher {
```
chtyim: Seems like the file name doesn't match with class name. (Mr in File name). you have to turn on the git case aware.

chtyim: this is because Mac FS is case insensitive.

anwar6953: Ok, I did this:
`git config core.ignorecase false`
Now, only one is in my local repo, but both show up in this git diff...

anwar6953: Ok, fixed by removing both files from git, and adding just the one.

chtyim:
LGTM

PR #2444, opened 2015-04-30
Transforms cannot load Plugins

gokulavasan:
Currently Transforms have access to AdapterContext but not to AdapterPluginRegistry. Removing AdapterContext access in Transforms so that Transform Plugins cannot load external Plugins in 3.0.

JIRA : https://issues.cask.co/browse/CDAP-2301

Build : http://builds.cask.co/browse/CDAP-RT19


cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/RealtimeStageContext.java
```diff
@@ -31,7 +27,8 @@
 public class RealtimeStageContext implements StageContext {
```
albertshau: how come this class is needed if there is a WorkerRealtimeContext or vice versa?

albertshau: o is this is for transforms? is the rename happening in a separate PR?

albertshau:
one question, if rename happening later, lgtm

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/RealtimeStageContext.java
gokulavasan: Yes its for Transform. Renaming of Transform related classes will happen in a separate PR.

gokulavasan:
Yes, renaming happening later. Will merge. Thanks.

PR #2446, opened 2015-04-30
CDAP-2345 mark all adapter related apis as beta and add some more

albertshau:
javadocs to the ApplicationTemplate class.


gokulavasan:
LGTM :+1:

PR #2447, opened 2015-04-30
Do not pass in queue names if they are empty

sreevatsanraman:
- Add the YARN queue only if the queue name is not empty


poornachandra:
LGTM

PR #2448, opened 2015-04-30
Reduce log level to debug in KafkaLogWriterPlugin

sreevatsanraman:


cdap-watchdog/src/main/java/co/cask/cdap/logging/save/KafkaLogWriterPlugin.java
```diff
@@ -158,7 +158,7 @@ public void init(Set<Integer> partitions) {
     countDownLatch = new CountDownLatch(1);
 
     if (partitions.contains(0)) {
-      LOG.info("Scheduling cleanup task");
+      LOG.debug("Scheduling cleanup task");
```
bdmogal: This is probably not needed. @poornachandra  confirmed that this happens once, and is good to know.

bdmogal:
LGTM other than 1 comment

sreevatsanraman:
Fixed. merging.

PR #2449, opened 2015-04-30
CDAP-2371 Added the verification that the program name should not be emp...

sagarkapare:
...ty

Jira: https://issues.cask.co/browse/CDAP-2371

If the programs are added in the application as `anonymous inner classes`, there name is always empty. Added verification to throw an exception if the name is not empty. 

This means the flow, flowlets, datasets, mapreduce, workflows should always have name otherwise app deployment would fail.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/verification/AbstractVerifier.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/verification/AbstractVerifier.java
@@ -26,7 +26,7 @@ import com.google.common.base.CharMatcher;
 public abstract class AbstractVerifier<T> implements Verifier<T> {
 
   protected boolean isId(final String name) {
-    return CharMatcher.inRange('A', 'Z')
+    return !name.isEmpty() && CharMatcher.inRange('A', 'Z')
              .or(CharMatcher.inRange('a', 'z'))
              .or(CharMatcher.is('-'))
              .or(CharMatcher.is('_'))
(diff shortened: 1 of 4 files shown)


sagarkapare:
Build running here - http://builds.cask.co/browse/CDAP-RBT246-1

cdap-app-fabric/src/main/java/co/cask/cdap/app/verification/AbstractVerifier.java
```diff
@@ -26,7 +26,7 @@
 public abstract class AbstractVerifier<T> implements Verifier<T> {
 
   protected boolean isId(final String name) {
```
bdmogal: Is there any way you can re-use methods from the Id class for this?

sagarkapare: Yeah. Removed the `isId` from here and using it from `Id` class.

sagarkapare:
@bdmogal please take a look again. Thanks!

sagarkapare:
Reverted the last commit. Build passed. @bdmogal please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/app/verification/AbstractVerifier.java
bdmogal: Ignore this for now, I think we need to refactor the ``Id`` class a little to share this ``CharMatcher``. Perhaps, create a jira for now with a TODO here.

bdmogal:
Just one comment for future. We need to somehow share the ``CharMatcher``. Infact, the validation using ``CharMatcher`` should happen only inside the ``Id`` class. Outside the Id class, we should only have to create an Id, and it should throw an exception if it is not valid. Looks like we may have to refactor ``AbstractVerifier`` and ``Id`` class a little bit to achieve that. Let's open a jira for that to fix in a future PR.

Rest LGTM

sreevatsanraman:
Build passes merging.

PR #2455, opened 2015-04-30
Ui/fix adapter status

ajainarayanan:
- Fixes status on adapters list view
- Fixes Worker name to be rendered in programs view of an adapter deployed app.


nitinmotgi:
LGTM :+1:

PR #2456, opened 2015-04-30
At least list workers (even though there's no workers page). Previously,...

anwar6953:
... workers program row is malformed (simply missing the name).


anwar6953:
Without this, there is a malformed row:
image

ajainarayanan:
:+1: LGTM

PR #2460, opened 2015-05-01
(CDP-2311) Fix MapReduce memory leak issue in local mode

chtyim:
- Introduce a WeakReferenceDelegatorClassLoader to hold a weak reference to the actual MapReduceClassLoader. The MR Framework will keep a reference to the WeakReferenceDelegatorClassLoader in a static variable, but it won’t prevent the GC of the actual MR ClassLoader
- Fix the MR ProgramRunner so that in local mode it won’t create new UGI for each run. It is needed to avoid the cache inside the Hadoop FileSystem object keep building up.
- Fix the LocalJobRunnerWithFix to not creating a new ClassLoader, but always use the one set in the Configuration object.

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/lang/WeakReferenceDelegatorClassLoader.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/lang/WeakReferenceDelegatorClassLoader.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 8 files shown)


cdap-common/src/main/java/co/cask/cdap/common/lang/WeakReferenceDelegatorClassLoader.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Override
+  public ClassLoader getDelegate() {
+    return delegate.get();
+  }
+
+  private ClassLoader ensureDelegateExists() {
+    ClassLoader classLoader = delegate.get();
+    if (classLoader == null) {
+      throw new IllegalStateException("Delegating ClassLoader is already Garbage Collected.");
```
anew: how do you prevent that the classloader is gc'ed while the MR job is still running?

chtyim: Take a look at the MapReduceRuntimeService, it holds a strong reference to the MapReduceClassLoader.

anew: oh, right. cool

anew:
LGTM

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT248-2

PR #2461, opened 2015-05-01
[CDAP-2393] emit process.events.processed tagged with the simple queue name 

anew:
... instead of the full queue name


chtyim:
LGTM

PR #2462, opened 2015-05-01
Show N/A instead of 0 or empty string, if the information from backend i...

anwar6953:
...s not available

Two main things:
1) Before the /info endpoint returns (while UI is waiting for initial response), the UI was showing 0 for the input and output records (and nothing for the other fields, because of the myNumbers filter being activated before any response even arrived.
2) Mapper and Reducer states were empty in the table in the case that they are not available. Now, it properly shows 'N/A' as [shown here](https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/i801XCBDbwIghEZ/Screen%20Shot%202015-04-30%20at%205.28.57%20PM.png).
Also, since task states are not reported when JobHistoryServer is not available (when using metrics system as fallback), now shows 'N/A' instead of 0 for the following fields:
Number of tasks that are: completed, running, pending, killed, failed
as [shown here](https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/6XW2rQ08DL8HedJ/Screen%20Shot%202015-04-30%20at%205.28.34%20PM.png).


elfenheart:
:+1: LGTM

PR #2463, opened 2015-05-01
Feature/allow main class from elsewhere

dereklwood:
Our common-env.sh was forcing ``$MAIN_CLASS`` to empty, making it impossible to wrap these scripts.  The Cloudera CSD also sources these scripts, and should be able to set ``$MAIN_CLASS`` for the upgrade tool


sreevatsanraman:
LGTM :8ball:

wolf31o2:
:+1: LGTM

PR #2467, opened 2015-05-01
Miscellanous UI changes

anwar6953:
1) debounce of 1000ms after clicking on a data/app before registering it in the ordering service.
2) hover on stream/flowlets on a flow page with out runs.
3) show stream metrics on a flow page with out runs (previously wasn't retrieving stream metrics on the flow page, if there are no runs).
4) remove BETA
5) Update Management page to look like [this](https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/RdNQt0sMSBqb6xx/Screen%20Shot%202015-04-30%20at%206.39.00%20PM.png) regardless of enterprise or standalone.
6) remove 2014 from footers


anwar6953:
Reopened against release.
https://github.com/caskdata/cdap/pull/2468

PR #2468, opened 2015-05-01
Miscellanous UI changes

anwar6953:
1) debounce of 1000ms after clicking on a data/app before registering it in the ordering service.
2) hover on stream/flowlets on a flow page with out runs.
3) show stream metrics on a flow page with out runs (previously wasn't retrieving stream metrics on the flow page, if there are no runs).
4) remove BETA
5) Update Management page to look like [this](https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/RdNQt0sMSBqb6xx/Screen%20Shot%202015-04-30%20at%206.39.00%20PM.png) regardless of enterprise or standalone.
6) remove 2014 from footers
7) Changed 'RUNNING' to 'DURATION' on flow/spark run-detail page.
8) 'controler' -> 'controller'
9) Make loading bar thicker (upon app deploy): [like so](https://s3.amazonaws.com/uploads.hipchat.com/26476/1011205/PL7ctSTCJnY6Dmv/Screen%20Shot%202015-04-30%20at%208.04.32%20PM.png)
10) System Services circle on overview page is now grey instead of red, until the request is received.

### DIFF
--- a/cdap-ui/app/features/admin/templates/overview.html
+++ b/cdap-ui/app/features/admin/templates/overview.html
@@ -5,12 +5,9 @@
         Welcome CDAP Admin!
       </h3>
     </div>
-    <div class="panel-body" ng-show="isEnterprise">
-      <p>This is the control console for CDAP (Enterprise mode). Here, you can view and manage your system configuration and define and manage namespaces, entities and access permissions to keep your team's work organized and secure.</p>
-      <p>Begin by configuring your default namespace.</p>
-    </div>
-    <div class="panel-body" ng-hide="isEnterprise">
-      <p>This is the control console for CDAP (Standalone mode).</p>
+    <div class="panel-body">
+      <p>This is the control console for CDAP. Here, you can view and manage your system configuration and define and manage namespaces and entities.
+         Begin by configuring your default namespace.</p>
     </div>
   </div>
 </div>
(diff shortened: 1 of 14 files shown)


cdap-ui/app/features/admin/templates/overview.html
```diff
@@ -5,12 +5,9 @@
...
       </h3>
     </div>
-    <div class="panel-body" ng-show="isEnterprise">
-      <p>This is the control console for CDAP (Enterprise mode). Here, you can view and manage your system configuration and define and manage namespaces, entities and access permissions to keep your team's work organized and secure.</p>
-      <p>Begin by configuring your default namespace.</p>
-    </div>
-    <div class="panel-body" ng-hide="isEnterprise">
-      <p>This is the control console for CDAP (Standalone mode).</p>
+    <div class="panel-body">
+      <p>This is the control console for CDAP. Here, you can view and manage your system configuration and define and manage namespaces and entities.
```
anwar6953: Here, I removed the part about access permissions, because that is not true. Ok with Nitin.

cdap-ui/app/features/flows/controllers/tabs/runs/tabs/status-ctrl.js
```diff
@@ -27,12 +27,12 @@ angular.module(PKG.name + '.feature.flows')
     }
 
     function pollMetrics() {
-      if (!$scope.runs.length) {
-        return;
-      }
       var nodes = $scope.data.nodes;
       // Requesting Metrics data
       angular.forEach(nodes, function (node) {
```
anwar6953: Previously, we were not hitting metrics for streams, if the flow had no runs. That was improper.

ajainarayanan:
:+1: LGTM

PR #2469, opened 2015-05-01
CDAP-2390: Use non-txnl MetricsTable for entity table in CubeDataset

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-RBT249-1

chtyim:
LGTM

abaranau:
tiny test fix: use cube of different name for different tests

abaranau:
will merge when build passes

PR #2472, opened 2015-05-01
[CDAP-2192] Aggregate adapter metrics by run id

bdmogal:
Jira: [CDAP-2192](https://issues.cask.co/browse/CDAP-2192)
Build: http://builds.cask.co/browse/CDAP-RBT250

### DIFF
--- a/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
+++ b/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
@@ -119,8 +119,20 @@ public class MetricsHandlerTestRun extends MetricsSuiteTestBase {
     collector.increment("areads", 3);
     collector.increment("awrites", 4);
 
+    collector = collectionService.getCollector(getAdapterContext("yourspace", "WCount1", "ClassicWordCount",
+                                                                 MapReduceMetrics.TaskType.Mapper,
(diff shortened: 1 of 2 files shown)


bdmogal:
Build :green_apple:

gsps1:
@bdmogal are you sure this PR solves the JIRA issue ?  when {metric-> value} is emitted in both adapter and mapreduce context,  and when you are querying without run-ids the result should be the same. as all the metrics from different run-ids would be aggregated by the adapter context as well, adding run-ids will give more resolution but i am not sure how that solves the issue?

bdmogal:
Yes, it does. Lets say we run a batch adapter 3 times (3 runs). When you query its metrics using the tags namespace, app and mapreduce, you get aggregated result across the 3 runs. However, without this fix, when you query its metrics using just the namespace and adapter tags, you get metrics from only one of the runs. 

The unit test updates should help explain.

sreevatsanraman:
LGTM :+1:

gsps1:
I was under the impression the metrics in the JIRA example was counter but the metric is a gauge and since for aggregates table, we add all the metrics to the same resolution, we keep updating the single resolution, that's the reason why we see the metrics for the latest run for aggregate queries.

for time-range queries you would get the same result for map-reduce and adapter even without run-ids, as they are gauged at unique timestamps. (verified)

we need run-ids for aggregate queries. so LGTM :+1:

bdmogal:
Thanks @gsps1. Build has been passing. Merging.

PR #2477, opened 2015-05-01
Fix JMS realtime source to pass the right properties via the PluginConfig

henry-cask:
Add more required parameters for JMS JNDI naming via the JmsPluginConfig fields rather than
pass through from PluginConfig properties to help testing.

This PR makes testing easier because the parent PluginConfig's properties is a private memeber
and the method to get the properties is final so need to push some of the java naming
properties as top level fields in JmsPluginConfig class which extends the PluginConfig

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/jms/JndiBasedJmsProvider.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/jms/JndiBasedJmsProvider.java
@@ -44,13 +44,18 @@ public class JndiBasedJmsProvider implements JmsProvider {
   private final ConnectionFactory connectionFactory;
   private final Destination destination;
 
-  public JndiBasedJmsProvider(Hashtable<String, String> initialContextEnv, String destinationName) {
+  public JndiBasedJmsProvider(Hashtable<String, String> initialContextEnv, final String destinationName,
+                              final String connectionFactoryName) {
     if (initialContextEnv == null) {
       throw new IllegalArgumentException("Cannot pass null initial context.");
     }
 
     if (destinationName == null || destinationName.trim().isEmpty()) {
       throw new IllegalArgumentException("Cannot pass null or empty destination name.");
+    }
+
+    if (connectionFactoryName == null || connectionFactoryName.trim().isEmpty()) {
+      throw new IllegalArgumentException("Cannot pass null or empty connection factory name.");
     }
 
     Context jndiContext = null;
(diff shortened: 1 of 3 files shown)


henry-cask:
Build:
http://builds.cask.co/browse/CDAP-RT21-1

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/jms/JndiBasedJmsProvider.java
```diff
@@ -44,7 +44,8 @@
   private final ConnectionFactory connectionFactory;
   private final Destination destination;
 
-  public JndiBasedJmsProvider(Hashtable<String, String> initialContextEnv, String destinationName) {
+  public JndiBasedJmsProvider(Hashtable<String, String> initialContextEnv, final String destinationName,
```
chtyim: Should use `Map` instead of `Hashtable` for type.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
```diff
@@ -95,13 +99,17 @@ public JmsSource(JmsPluginConfig config) {
...
     messagesToReceive = configMessagesToReceive.intValue();
 
-    // Try get the destination name
-    String destinationName = config.destinationName;
+    // Set initial context factory name and provider URL
+    runtimeArguments.put(Context.INITIAL_CONTEXT_FACTORY, config.initialContextFactory);
+    runtimeArguments.put(Context.PROVIDER_URL, config.providerUrl);
 
     // Get environment vars - this would be prefixed with java.naming.*
     final Hashtable<String, String> envVars = new Hashtable<String, String>();
```
chtyim: Type should be Map.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/jms/JndiBasedJmsProvider.java
henry-cask: Unfortunately the stupid javax InitialContext only takes HashTable =(

So I have to use HashTable all the way.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/templates/etl/realtime/sources/JmsSource.java
henry-cask: This will eventually be passed to JmsProvider which later be pushed to InitialContext which only take HashTable. Either do HashTable now or convert later =(

This  is what happen when we use 10 year old tech =P

chtyim: I see.

chtyim:
LGTM.

henry-cask:
Thx @chtyim, merging ...

PR #2483, opened 2015-05-01
Setting snappy dependency to 1.1.1.7 since 1.0.5 causes problems with Kafka RealtimeSource

gokulavasan:
Bamboo : http://builds.cask.co/browse/CDAP-RT23-1

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
@@ -179,6 +179,10 @@
       <groupId>org.apache.kafka</groupId>
       <artifactId>kafka_2.10</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.xerial.snappy</groupId>
+      <artifactId>snappy-java</artifactId>
+    </dependency>
     <!-- End for Kafka Source -->
   </dependencies>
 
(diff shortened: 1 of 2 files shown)


henry-cask:
+1
LGTM

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
```diff
@@ -179,6 +179,10 @@
       <groupId>org.apache.kafka</groupId>
       <artifactId>kafka_2.10</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.xerial.snappy</groupId>
+      <artifactId>snappy-java</artifactId>
```
henry-cask: I remember ordering of dependency matter in certain maven version. Probably should add this dependency before Kafka just to be safe?

chtyim: Better be exclude the transitive one from Kafka. Maven now depends on the depth of the dependency tree to determine the version

henry-cask: Good idea,

gokulavasan:
@chtyim Excluded snappy dep in etl-lib pom. Should this be done in cdap pom instead?

chtyim:
LGTM.

PR #2488, opened 2015-05-01
delete remote dir before recreating, and add protection for when there i...

duddun:
...s no branch appended


wolf31o2:
:+1: LGTM

PR #2489, opened 2015-05-01
Fix/ui adapter styling 2

ajainarayanan:
- Fixes support link
- Fixes namespace dropdown to appear in adapters view
- Fixes tooltips in property edits to be on right
- Fixes no programs view in streams & datasets to have some margin


elfenheart:
:+1: LGTM

PR #2495, opened 2015-05-01
CDAP-2202 Validate RunRecord for running program agains ProgramRuntimeService as source of truth

henry-cask:
This PR is proposed fix for CDAP-2202.

Currently the calls to check run status of a program is run against run records' store.
In case of abrupt shutdown, there is no chance to update the state of the run record of the programs.
For example if a Flow program is running in standalone then the programs dies, then the run record
for that program will indicate still running when the CDAP is restarted.

So, the call for
  ```
http://localhost:10000/v3/namespaces/default/apps/CountRandom/flows/CountRandom/runs?status=running
```
and
  ```
http://localhost:10000/v3/namespaces/default/apps/CountRandom/flows/CountRandom/status
could return different results.
````

The proposed fix is to make the Program lifecycle access to run records to check against the runtime service when querying for run records of a program.

For Adapter we just need to check for worker one because the workflow is done via scheduler so
when cdap start the workflow adapter should start the scheduled programs accordingly.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -440,7 +440,7 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     }
     long start = (startTs == null || startTs.isEmpty()) ? 0 : Long.parseLong(startTs);
     long end = (endTs == null || endTs.isEmpty()) ? Long.MAX_VALUE : Long.parseLong(endTs);
-    getRuns(responder, Id.Program.from(namespaceId, appId, type, programId), status, start, end, resultLimit, type);
+    getRuns(responder, Id.Program.from(namespaceId, appId, type, programId), status, start, end, resultLimit);
   }
 
   /**
@@ -1412,10 +1412,10 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     }
   }
 
-  private void getRuns(HttpResponder responder, Id.Program programId, String status,
-                       long start, long end, int limit, final ProgramType programType) {
+  private void getRuns(HttpResponder responder, Id.Program programId, String status, long start, long end, int limit) {
     try {
       try {
+        final ProgramType programType = programId.getType();
         final ProgramRunStatus runStatus = (status == null) ? ProgramRunStatus.ALL :
           ProgramRunStatus.valueOf(status.toUpperCase());
 
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -62,7 +62,10 @@ import co.cask.cdap.proto.codec.ScheduleSpecificationCodec;
 import co.cask.http.HttpResponder;
 import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
 import com.google.common.base.Throwables;
+import com.google.common.collect.Collections2;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -232,6 +235,9 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
       if (runRecord == null) {
         throw new NotFoundException(run);
       }
+
+      // Try to validate the run record with runtime service
+      runRecord = lifecycleService.validateRunRecord(runRecord, programId, ProgramType.MAPREDUCE);
 
       MRJobInfo mrJobInfo = mrJobInfoFetcher.getMRJobInfo(run);
 
@@ -434,7 +440,7 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     }
     long start = (startTs == null || startTs.isEmpty()) ? 0 : Long.parseLong(startTs);
     long end = (endTs == null || endTs.isEmpty()) ? Long.MAX_VALUE : Long.parseLong(endTs);
-    getRuns(responder, Id.Program.from(namespaceId, appId, type, programId), status, start, end, resultLimit);
+    getRuns(responder, Id.Program.from(namespaceId, appId, type, programId), status, start, end, resultLimit, type);
   }
 
   /**
@@ -457,6 +463,11 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     try {
       RunRecord runRecord = store.getRun(Id.Program.from(namespaceId, appId, type, programId), runid);
       if (runRecord != null) {
+
+        // Try to validate the run record with runtime service
+        Id.Program program  = Id.Program.from(namespaceId, appId, type, programId);
+        runRecord = lifecycleService.validateRunRecord(runRecord, program, type);
+
         responder.sendJson(HttpResponseStatus.OK, runRecord);
         return;
       }
@@ -1402,12 +1413,26 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
   }
 
   private void getRuns(HttpResponder responder, Id.Program programId, String status,
-                       long start, long end, int limit) {
+                       long start, long end, int limit, final ProgramType programType) {
     try {
       try {
-        ProgramRunStatus runStatus = (status == null) ? ProgramRunStatus.ALL :
+        final ProgramRunStatus runStatus = (status == null) ? ProgramRunStatus.ALL :
           ProgramRunStatus.valueOf(status.toUpperCase());
-        responder.sendJson(HttpResponseStatus.OK, store.getRuns(programId, runStatus, start, end, limit));
+
+        // Need to validate whether the run records are reflecting the actual running state of the program.
+        List<RunRecord> runRecords = lifecycleService.validateRunRecords(
+          store.getRuns(programId, runStatus, start, end, limit), programId, programType);
+
+        // Filter all the validated run records based onf run status
+        List<RunRecord> filteredRunRecords = ImmutableList.copyOf(
+          Collections2.filter(runRecords, new Predicate<RunRecord>() {
+            @Override
+            public boolean apply(RunRecord record) {
+              return (record.getStatus() == runStatus);
+            }
+          }));
+
+        responder.sendJson(HttpResponseStatus.OK, filteredRunRecords);
       } catch (IllegalArgumentException e) {
         responder.sendString(HttpResponseStatus.BAD_REQUEST,
                              "Supported options for status of runs are running/completed/failed");
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
@@ -394,15 +394,30 @@ public class AdapterService extends AbstractIdleService {
    * @throws IOException if there was some error starting worker adapter
    */
   public synchronized void startAdapter(Id.Namespace namespace, String adapterName)
-    throws NotFoundException, InvalidAdapterOperationException, SchedulerException, IOException {
+    throws NotFoundException, InvalidAdapterOperationException, SchedulerException, IOException, ExecutionException,
+    InterruptedException {
+    AdapterDefinition adapterSpec = getAdapter(namespace, adapterName);
+    ProgramType programType = adapterSpec.getProgram().getType();
+
     AdapterStatus adapterStatus = getAdapterStatus(namespace, adapterName);
     if (AdapterStatus.STARTED.equals(adapterStatus)) {
-      throw new InvalidAdapterOperationException("Adapter is already started.");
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by henry-cask** [comment]
Build:
http://builds.cask.co/browse/CDAP-RBT257-1

File: cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java [hunk 1]
```diff
@@ -1402,12 +1413,26 @@ private AppFabricServiceStatus stop(Id.Program identifier, ProgramType type) {
   }
 
   private void getRuns(HttpResponder responder, Id.Program programId, String status,
-                       long start, long end, int limit) {
+                       long start, long end, int limit, final ProgramType programType) {
```
anew: Doesn't the programId contain the type?

henry-cask: Ah you are right, did not know that. Will update to get it from the Program ID instead

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -191,4 +195,48 @@ public void stopProgram(Id.Program programId, RunId runId) throws ExecutionExcep
...
+   * @return validated {@link RunRecord} instance
+   */
+  public RunRecord validateRunRecord(final RunRecord runRecord, Id.Program programId, ProgramType programType)
+    throws ExecutionException, InterruptedException {
+    ProgramRuntimeService.RuntimeInfo runtimeInfo = findRuntimeInfo(programId, programType);
+    // check if it is not running  but the run record says it is update the run record for that program
+    ProgramRunStatus status = runRecord.getStatus();
+    if (status == ProgramRunStatus.RUNNING && runtimeInfo == null) {
+      RunId runId = RunIds.fromString(runRecord.getPid());
+      store.setStop(programId, runId.getId(), System.currentTimeMillis(),
```
anew: there is a possible race condition here. The program may have been about to stop (or already stopped, but not updated in the run records yet) when you read the run records. So its status was RUNNING. Between then and now, it has properly stopped, and its status is now STOPPED. But you will overwrite it with KILLED. That would be incorrect.

anew: Also it is a little misleading to update the run record in a method that is called validate... the method name does not suggest that this will update it.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1401,13 +1412,27 @@ private AppFabricServiceStatus stop(Id.Program identifier, ProgramType type) {
...
+        // Need to validate whether the run records are reflecting the actual running state of the program.
+        List<RunRecord> runRecords = lifecycleService.validateRunRecords(
+          store.getRuns(programId, runStatus, start, end, limit), programId, programType);
+
+        // Filter all the validated run records based onf run status
+        List<RunRecord> filteredRunRecords = ImmutableList.copyOf(
+          Collections2.filter(runRecords, new Predicate<RunRecord>() {
+            @Override
+            public boolean apply(RunRecord record) {
+              return (record.getStatus() == runStatus);
```
anew: this will always be false if status==null, that is runStatus==ALL

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1401,13 +1412,27 @@ private AppFabricServiceStatus stop(Id.Program identifier, ProgramType type) {
...
       try {
-        ProgramRunStatus runStatus = (status == null) ? ProgramRunStatus.ALL :
+        final ProgramType programType = programId.getType();
+        final ProgramRunStatus runStatus = (status == null) ? ProgramRunStatus.ALL :
           ProgramRunStatus.valueOf(status.toUpperCase());
-        responder.sendJson(HttpResponseStatus.OK, store.getRuns(programId, runStatus, start, end, limit));
+
+        // Need to validate whether the run records are reflecting the actual running state of the program.
+        List<RunRecord> runRecords = lifecycleService.validateRunRecords(
+          store.getRuns(programId, runStatus, start, end, limit), programId, programType);
```
anew: This works if the query was for RUNNING, you can list the running records and post-filter and remove the ones that are not actually running. But if the query is for KILLED, then this won't work, because getRuns will only return the runs that are already marked as KILLED; not the ones that are marked as RUNNING but have actually failed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -394,15 +394,30 @@ public synchronized void stopAdapter(Id.Namespace namespace, String adapterName)
...
-    throws NotFoundException, InvalidAdapterOperationException, SchedulerException, IOException {
+    throws NotFoundException, InvalidAdapterOperationException, SchedulerException, IOException, ExecutionException,
+    InterruptedException {
+    AdapterDefinition adapterSpec = getAdapter(namespace, adapterName);
+    ProgramType programType = adapterSpec.getProgram().getType();
+
     AdapterStatus adapterStatus = getAdapterStatus(namespace, adapterName);
     if (AdapterStatus.STARTED.equals(adapterStatus)) {
-      throw new InvalidAdapterOperationException("Adapter is already started.");
+      // For Adapter with type of Worker, need to check whether the program is running since Workflow should
```
anew: I don't understand this comment. Can we schedule workers?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
henry-cask: Yes, but we do not care for those that are is KILLED. The check is done for programs that at the execution time of this method is not actually running via call to the runtime service but the run records said it is running.

In the case of KILLED, the runtime service will tell that there is no program running for that id and hence the run records are not modified.
We do not handle the case for programs that are actually RUNNING but the run records state otherwise.

I shoule have just called the new method something like compareAndSetRunningRunRecord

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
henry-cask: Good point, will add check for ALL status

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -394,15 +394,30 @@ public synchronized void stopAdapter(Id.Namespace namespace, String adapterName)
...
+      // already added to scheduler
+      if (programType == ProgramType.WORKER) {
+        Id.Program program = getProgramId(namespace, adapterName);
+        ProgramRuntimeService.RuntimeInfo runtimeInfo = lifecycleService.findRuntimeInfo(program, programType);
+        // check if it is not running
+        if (runtimeInfo == null) {
+          // not running, lets update the RunRecords
+          stopWorkerAdapter(namespace, adapterSpec);
```
anew: this will not do anything. It will call lifecycleService.stopProgram(), which will fail to find the runtime info and then do nothing. If I understand correctly, this is the case where the worker crashed, thus its runtime info does not exist any longer, but its run record has not been updated. You need to correct its run record to FAILED, similar to what is done in the program lifecycle handler.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
henry-cask: Agree about the method name. Maybe something like compareAndSetRunningRunRecord because we are specifically want to deal with case where runtime service said the program is not currently running but RunRecord said otherwise.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -191,4 +195,48 @@ public void stopProgram(Id.Program programId, RunId runId) throws ExecutionExcep
...
+                    ProgramController.State.KILLED.getRunStatus());
```
anew: Seems that this situation represents a crash (either of standalone, or of the program controller/master). So I think the correct status should be FAILED, not KILLED.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anew: On further thinking, it should be FAILED, not KILLED.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
anew: or validateAndCorrectRunRecord()?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anew: What I meant was that this endpoint can be used to show all runs of a particular run status. It is perfectly valid to call this to list all failed runs. With your change, the runs that have failed but have not been updated as failed in the run record will not be listed. This will lead to perceived inconsistencies. Some runs will show as failed when listing ALL, but would not show up when listing FAILED. Or, if you list FAILED ones first, RUNNING ones next, you will not see some runs. But if you first list RUNNING runs, then FAILED ones, these runs will show as failed.

anew:
@henry-cask I am done with my first pass. Please see my comments

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
henry-cask: I do not think we have the race condition issue here. 
Changing a RunRecord for a particular program will only happen when the program runtime service said the program is NOT in RUNNING state AND the RunRecord said it IS in a RUNNING state. 
So, the RunRecord for that particular program is not valid anyway and need to sync with actual state (aka the runtime info) from the program runtime service. This is kind of like RAFT concensus protocol where the replica is out of sync from the leader and hence need to be updated to keep up with leader (in this case running info from runtime service).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
henry-cask: Currently we do not schedule Worker adapter

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
henry-cask: FAILED will work for me

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
henry-cask: Yeah, looks like I need to do it manually similar to program lifecyle service.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
anew: Sorry, I disagree. Consider this sequence of events:

1. You query the run record. Program is RUNNING. 
2. Program stops and updates its run record to COMPLETED (and its runtime info is now gone).
3. You query the runtime info and don't find it. You conclude that this run crashed and update it to FAILED. 

Now a successful run is marked as failed in the run history.

anew: To do this properly, you need to do a checkAndPut-like operation. Or simply, within a transaction, read the run status in the run record, and if it is still running, update it to failed; or if it is not running any more, then its status has update since you queried the first time, and the correct status is now already in the run record.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
anew: Then what does this comment say about schedule?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
henry-cask: Yes, currently we are doing it lazily which is we fix it when request for RUNNING is being sent for a particular program.

Another way to do it is to correct all run records state in the bootstrap of CDAP master. 
I am not sure where it is actually, but that would fix this consistency issues with penalty of bit longer bootstrap to scan all RunRecords records and ping runtime service for each of them.

henry-cask:
Closing this one to revise the solution

PR #2497, opened 2015-05-01
CDAP-2413 Restore config tool to determine java version

poornachandra:
Restoring logic to determine java version, etc. from https://github.com/caskdata/cdap/blob/d3e276ab1c2d45d2e4ea357ecd8cd16fb70740ce/cdap-web-app/bin/config-tool

### DIFF
--- a/cdap-ui/bin/config-tool
+++ b/cdap-ui/bin/config-tool
@@ -16,40 +16,64 @@
 # the License.
 #
 
-bin=`dirname ${BASH_SOURCE-$0}`
-bin=`cd ${bin} >/dev/null && pwd`
-lib=${bin}/../lib
-conf=${bin}/../conf
-script=`basename $0`
+# Attempt to set APP_HOME
+# Resolve links: $0 may be a link
+PRG="$0"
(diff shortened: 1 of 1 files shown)


cdap-ui/bin/config-tool
```diff
@@ -16,40 +16,64 @@
...
+    link=`expr "$ls" : '.*-> \(.*\)$'`
+    if expr "$link" : '/.*' > /dev/null; then
+        PRG="$link"
+    else
+        PRG=`dirname "$PRG"`"/$link"
+    fi
+done
+SAVED="`pwd`"
+cd "`dirname \"$PRG\"`/.."
+APP_HOME="`pwd -P`"
```
dereklwood: Looks like ``APP_HOME`` isn't used, so i think this whole block can be removed

wolf31o2:
This restored way too much. We only needed the JAVA_HOME/JAVA bits. The rest should be left alone.

poornachandra:
Closing this PR, the changes are now in https://github.com/caskdata/cdap/pull/2498

PR #2499, opened 2015-05-01
CDAP-2264: FlowQueuePendingCorrector tool for correcting the queue.pending metric

jwang47:
http://builds.cask.co/browse/CDAP-RBT264

```
/opt/cdap/master/bin/svc-master run co.cask.cdap.data.tools.FlowQueuePendingCorrector -n default -a PurchaseHistory -f PurchaseFlow
```

Help message for the new tool:

```
usage: --namespace <namespace> --app <app> --flow <flow>
                   [--producer-flowlet <flowlet> --consumer-flowlet
                   <flowlet> [--queue <queue>]]
 -a,--app <arg>                app
 -c,--consumer-flowlet <arg>   consumer flowlet (optional, leave empty to
                               correct entire flow)
 -f,--flow <arg>               flow
 -n,--namespace <arg>          namespace
 -p,--producer-flowlet <arg>   producer flowlet (optional, leave empty to
                               correct entire flow)
 -q,--queue <arg>              flowlet queue (optional, defaults to
                               "queue")
```

* Tested UpgradeTool changes in a cluster
* Tested FlowQueuePendingCorrector in a cluster

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/queue/QueueName.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/queue/QueueName.java
@@ -71,6 +71,10 @@ public final class QueueName {
     return new QueueName(URI.create(new String(bytes, Charsets.US_ASCII)));
   }
 
+  public static QueueName fromFlowlet(Id.Flow flow, String flowlet, String output) {
+    return fromFlowlet(flow.getNamespaceId(), flow.getApplicationId(), flow.getId(), flowlet, output);
+  }
+
   public static QueueName fromFlowlet(String namespace, String app, String flow, String flowlet, String output) {
     URI uri = URI.create(String.format("queue:///%s/%s/%s/%s/%s", namespace, app, flow, flowlet, output));
     return new QueueName(uri);
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
@@ -15,50 +15,45 @@
  */
 package co.cask.cdap.data.tools;
 
-import co.cask.cdap.api.dataset.module.DatasetDefinitionRegistry;
-import co.cask.cdap.api.dataset.module.DatasetModule;
-import co.cask.cdap.api.metrics.MetricStore;
-import co.cask.cdap.api.metrics.MetricsCollectionService;
+import co.cask.cdap.api.flow.FlowSpecification;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
+import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.guice.AppFabricServiceRuntimeModule;
 import co.cask.cdap.app.guice.ProgramRunnerRuntimeModule;
-import co.cask.cdap.app.store.ServiceStore;
+import co.cask.cdap.app.guice.ServiceStoreModules;
 import co.cask.cdap.app.store.Store;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.guice.ConfigModule;
 import co.cask.cdap.common.guice.DiscoveryRuntimeModule;
+import co.cask.cdap.common.guice.IOModule;
 import co.cask.cdap.common.guice.KafkaClientModule;
 import co.cask.cdap.common.guice.LocationRuntimeModule;
 import co.cask.cdap.common.guice.TwillModule;
 import co.cask.cdap.common.guice.ZKClientModule;
-import co.cask.cdap.common.metrics.NoOpMetricsCollectionService;
 import co.cask.cdap.common.namespace.NamespacedLocationFactory;
 import co.cask.cdap.common.utils.ProjectInfo;
-import co.cask.cdap.config.DefaultConfigStore;
 import co.cask.cdap.data.runtime.DataFabricDistributedModule;
+import co.cask.cdap.data.runtime.DataSetsModules;
 import co.cask.cdap.data.runtime.SystemDatasetRuntimeModule;
 import co.cask.cdap.data.stream.StreamAdminModules;
-import co.cask.cdap.data2.datafabric.dataset.DatasetMetaTableUtil;
-import co.cask.cdap.data2.dataset2.DatasetDefinitionRegistryFactory;
+import co.cask.cdap.data.tools.flow.FlowQueuePendingCorrector;
 import co.cask.cdap.data2.dataset2.DatasetFramework;
 import co.cask.cdap.data2.dataset2.DatasetManagementException;
-import co.cask.cdap.data2.dataset2.DefaultDatasetDefinitionRegistry;
-import co.cask.cdap.data2.dataset2.InMemoryDatasetFramework;
-import co.cask.cdap.data2.dataset2.lib.kv.HBaseKVTableDefinition;
+import co.cask.cdap.data2.queue.QueueClientFactory;
+import co.cask.cdap.data2.transaction.queue.QueueAdmin;
+import co.cask.cdap.data2.transaction.queue.hbase.HBaseQueueAdmin;
+import co.cask.cdap.data2.transaction.queue.hbase.HBaseQueueClientFactory;
+import co.cask.cdap.data2.util.hbase.HBaseTableUtil;
+import co.cask.cdap.data2.util.hbase.HBaseTableUtilFactory;
 import co.cask.cdap.explore.guice.ExploreClientModule;
 import co.cask.cdap.gateway.auth.AuthModule;
-import co.cask.cdap.gateway.handlers.DatasetServiceStore;
 import co.cask.cdap.internal.app.namespace.NamespaceAdmin;
 import co.cask.cdap.internal.app.runtime.adapter.AdapterService;
 import co.cask.cdap.internal.app.runtime.schedule.AbstractSchedulerService;
 import co.cask.cdap.internal.app.runtime.schedule.ExecutorThreadPool;
 import co.cask.cdap.internal.app.runtime.schedule.store.DatasetBasedTimeScheduleStore;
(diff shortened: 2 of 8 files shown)


cdap-common/src/main/java/co/cask/cdap/common/queue/QueueName.java
```diff
@@ -71,6 +71,12 @@ public static QueueName from(byte[] bytes) {
     return new QueueName(URI.create(new String(bytes, Charsets.US_ASCII)));
   }
 
+  public static QueueName fromFlowlet(Id.Flow flow, String flowlet, String output) {
+    URI uri = URI.create(String.format("queue:///%s/%s/%s/%s/%s",
+                                       flow.getNamespace(), flow.getApplicationId(), flow.getId(), flowlet, output));
+    return new QueueName(uri);
```
anew: better return fromFlowlet(flow.getNamespace(), flow.getApplicationId(), flow.getId(), flowlet, output)

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricDataQuery.java
```diff
@@ -41,8 +41,20 @@
...
+  /**
+   * Start timestamp, in seconds.
+   */
   private final long startTs;
+
+  /**
+   * End timestamp, in seconds.
+   */
   private final long endTs;
+
```
anew: Thanks for adding that. Javadocs has been missing from this class, and you probably tried to use millis first? 
While you're at it: It would make sense to add this information to the constructor, too.

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  private static CommandLine parseArgs(String[] args) {
+    Options options = new Options();
+    options.addOption(createOption("n", "namespace", true, "namespace", true));
+    options.addOption(createOption("a", "app", true, "app", true));
+    options.addOption(createOption("f", "flow", true, "flow", true));
+    options.addOption(createOption("p", "producer-flowlet", true, "producer flowlet", true));
+    options.addOption(createOption("c", "consumer-flowlet", true, "consumer flowlet", true));
+    options.addOption(createOption("q", "queue", true, "flowlet queue (defaults to \"queue\"", false));
```
anew: can we just do it for the entire flow if the details are not provided?

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
```diff
@@ -0,0 +1,229 @@
...
+    this.kafkaClientService = kafkaClientService;
+  }
+
+  public void run(Id.Flow flow, String producerFlowlet, String consumerFlowlet,
+                  String flowletQueue) throws Exception {
+
+    QueueName queueName = QueueName.fromFlowlet(flow, producerFlowlet, flowletQueue);
+    long consumerGroupId = FlowUtils.generateConsumerGroupId(flow, consumerFlowlet);
+    HBaseQueueDebugger.QueueStatistics stats = queueDebugger.scanQueue(queueName, consumerGroupId);
+    long correctQueuePendingValue = stats.getUnprocessed() + stats.getProcessedAndNotVisible();
```
anew: what is processed and not visible? Can that ever happen if the flow is stopped?

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
```diff
@@ -0,0 +1,229 @@
...
+    this.zkClientService = zkClientService;
+    this.metricsCollectionService = metricsCollectionService;
+    this.metricStore = metricStore;
+    this.kafkaClientService = kafkaClientService;
+  }
+
+  public void run(Id.Flow flow, String producerFlowlet, String consumerFlowlet,
+                  String flowletQueue) throws Exception {
+
+    QueueName queueName = QueueName.fromFlowlet(flow, producerFlowlet, flowletQueue);
```
anew: you should make sure that the flow is not running, Otherwise there will be race conditions with metrics emitted by the flow.

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
```diff
@@ -0,0 +1,229 @@
...
+      .build();
+
+    MetricDataQuery query = new MetricDataQuery(
+      0, 0, Integer.MAX_VALUE, 1, ImmutableMap.of("system.queue.pending", AggregationFunction.SUM),
+      tags, ImmutableList.<String>of(), null);
+
+    Collection<MetricTimeSeries> results = metricStore.query(query);
+    System.out.println("Got results: " + GSON.toJson(results));
+
+    if (results.isEmpty()) {
```
anew: this is a valid case. After upgrade to 3.0, existing flows will not have this metric. Then we have to emit a correction once.

anew: But you should make sure that the flow, consumer and producer flowlet and queue actually exist. Otherwise you would emit a metric for a queue that does not exist.

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
```diff
@@ -0,0 +1,229 @@
...
+ * Corrects the "queue.pending" metric emitted in {@link co.cask.cdap.internal.app.runtime.flow.FlowletProgramRunner}.
+ */
+public class FlowQueuePendingCorrector extends AbstractIdleService {
+
+  private static final Gson GSON = new Gson();
+
+  private final MetricsCollectionService metricsCollectionService;
+  private final MetricStore metricStore;
+  private final KafkaClientService kafkaClientService;
+  private final HBaseQueueDebugger queueDebugger;
```
anew: This runs only with Hbase queues? So in standalone, we won't have a way to correct it... do we have a queue introspection for standalone?

anew:
Where is the test for this tool?

anew:
Done with my first pass. Please see comments. Thanks

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
jwang47: Nope, no queue introspection for standalone right now.

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
jwang47: This means the queue entry is processed, but not visible according to the transaction it runs inside.

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
jwang47: fixed

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
jwang47: fixed

cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricDataQuery.java
```diff
@@ -51,18 +63,45 @@
...
                          Map<String, AggregationFunction> metrics,
                          Map<String, String> sliceByTagValues, List<String> groupByTags) {
     this(startTs, endTs, resolution, -1, metrics, sliceByTagValues, groupByTags, null);
   }
 
+  /**
+   * @param startTs Start timestamp, in seconds.
+   * @param endTs End timestamp, in seconds.
+   * @param resolution Resolution in seconds.
+   * @param limit
```
anew: if you don't want to document a parameter, don't add a @param tag.

cdap-master/src/main/java/co/cask/cdap/data/tools/flow/FlowQueuePendingCorrector.java
anew: Because this should only run when the flow is stopped, there should not be any invisible processed entries. Can you add a check that it is zero and print a warning if it is not?

anew:
Looks much better now, please see my new comment. And I am still not seeing a test case.

anew:
All looks good now, except that the test does not do much. Please create a more meaningful test in a new PR, and merge this one.

anew:
LGTM

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -371,36 +376,6 @@ private void upgradeAdapters() throws SchedulerException {
...
-      upgradeTool.doMain(args);
-    } catch (Throwable t) {
-      LOG.error("Failed to upgrade ...", t);
-    }
-  }
-
-  /**
-   * Sets up a {@link DatasetFramework} instance for standalone usage.  NOTE: should NOT be used by applications!!!
-   */
-  private void initializeDSFramework(CConfiguration cConf, DatasetFramework datasetFramework) throws IOException,
```
rohitsinha54: Why was this done ?

rohitsinha54:
Did the upgrade tool ran after these changes. It doesn't for me. I have opened a JIRA here: 

https://issues.cask.co/browse/CDAP-2440

PR #2502, opened 2015-05-01
CDAP-2415 improving etl worker performance 

gokulavasan:
Mainly by reducing dataset access.

i) Fetch the source state only once in the beginning.
ii) Update that state only if there is a change in state.
iii) Invoke the sink's write method only if there is data to be persisted
iv) Start a Transaction only if there is sink data to be persisted or state has to be persisted
v) Accumulate all data from TransformExecutor before invoking sink's write

Other functional changes:
i) If there is an exception thrown when polling source, then continue without processing data

JIRA: https://issues.cask.co/browse/CDAP-2415
Build : http://builds.cask.co/browse/CDAP-RT27-1

- [x] Test on Cluster

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
@@ -90,7 +90,9 @@ public class ETLWorker extends AbstractWorker {
     Preconditions.checkArgument(runtimeArgs.containsKey(Constants.Realtime.UNIQUE_ID));
 
     adapterName = runtimeArgs.get(Constants.ADAPTER_NAME);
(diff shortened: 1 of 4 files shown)


cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/templates/etl/realtime/ETLWorker.java
```diff
@@ -172,47 +174,82 @@ private void initializeSink(WorkerContext context, ETLStage stage) throws Except
...
+          }
+        } catch (Exception e) {
+          LOG.warn("Adapter {} : Exception thrown while processing data {}", adapterName, sourceData, e);
+        }
+      }
+      sourceEmitter.reset();
 
+      // Start a Transaction if there is data to persist or if the Source state has changed.
+      try {
+        if ((dataToSink.size() != 0) || (!nextState.equals(currentState))) {
```
poornachandra: `dataToSink.size()` can be replaced by `dataToSink.empty()` here and and in line 231.

poornachandra:
Minor comment, fix and merge.

gokulavasan:
Tested on cluster. Merging.

PR #2503, opened 2015-05-01
Batch Sink by default sets the input object as both key and value of the...

gokulavasan:
... key-value pair expected by the Batch Output

JIRA : https://issues.cask.co/browse/CDAP-2414

Bamboo : http://builds.cask.co/browse/CDAP-RT28-1


sreevatsanraman:
LGTM :+1:

PR #2505, opened 2015-05-01
[CDAP-2363] Provide a reason for adapter deletion failure. Also change t...

bdmogal:
...he response code to CONFLICT from FORBIDDEN

Jira: [CDAP-2363](https://issues.cask.co/browse/CDAP-2363)
Build: http://builds.cask.co/browse/CDAP-RBT259

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
@@ -276,7 +276,7 @@ public class AdapterHttpHandler extends AbstractAppFabricHttpHandler {
       adapterService.removeAdapter(Id.Namespace.from(namespaceId), adapterName);
       responder.sendStatus(HttpResponseStatus.OK);
     } catch (CannotBeDeletedException e) {
-      responder.sendString(HttpResponseStatus.FORBIDDEN, e.getMessage());
+      responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
     } catch (NotFoundException e) {
       responder.sendString(HttpResponseStatus.NOT_FOUND, e.getMessage());
     } catch (Throwable t) {
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
@@ -195,7 +195,7 @@ public class NamespaceHttpHandler extends AbstractAppFabricHttpHandler {
     } catch (NotFoundException e) {
       responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Namespace %s not found.", namespace));
     } catch (NamespaceCannotBeDeletedException e) {
-      responder.sendString(HttpResponseStatus.FORBIDDEN, e.getMessage());
+      responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
     } catch (Exception e) {
       LOG.error("Internal error while deleting namespace.", e);
       responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
@@ -325,7 +325,8 @@ public class AdapterService extends AbstractIdleService {
     AdapterSpecification adapterSpec = getAdapter(namespace, adapterName);
     Id.Application applicationId = Id.Application.from(namespace, adapterSpec.getTemplate());
     if (adapterStatus != AdapterStatus.STOPPED) {
-      throw new CannotBeDeletedException(Id.Adapter.from(namespace, adapterName));
+      throw new CannotBeDeletedException(Id.Adapter.from(namespace, adapterName), "The adapter has not been stopped." +
+        " Please stop it first.");
     }
     store.removeAdapter(namespace, adapterName);
     try {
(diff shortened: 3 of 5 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
```diff
@@ -276,7 +276,7 @@ public void deleteAdapter(HttpRequest request, HttpResponder responder,
       adapterService.removeAdapter(Id.Namespace.from(namespaceId), adapterName);
       responder.sendStatus(HttpResponseStatus.OK);
     } catch (CannotBeDeletedException e) {
-      responder.sendString(HttpResponseStatus.FORBIDDEN, e.getMessage());
+      responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
```
jwang47: Are most `CannotBeDeletedException` supposed to result in CONFLICT? If so, you can remove this catch-block and simply catch it in https://github.com/caskdata/cdap/blob/develop/cdap-common/src/main/java/co/cask/cdap/common/exception/HttpExceptionHandler.java#L35

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
```diff
@@ -276,7 +276,7 @@ public void deleteAdapter(HttpRequest request, HttpResponder responder,
...
     } catch (NotFoundException e) {
```
jwang47: Not part of your PR, but you can remove this since `HttpExceptionHandler` handles this. Possibly also the `catch (Throwable t)` block below too.

anew: We should probably do a walk through all the handlers and remove unnecessary catch clauses. I created https://issues.cask.co/browse/CDAP-2422. @jwang47 can you add some more details to that, please?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
bdmogal: This is what I see so far:

```
$ gj "catch.*\\(CannotBeDeletedException.*\)" -A2 *
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java:    } catch (CannotBeDeletedException e) {
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java-      responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java-    } catch (NotFoundException e) {
--
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java:      } catch (CannotBeDeletedException cbde) {
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java-        appStatus = AppFabricServiceStatus.PROGRAM_STILL_RUNNING;
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java-      } catch (Exception e) {
--
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java:    } catch (CannotBeDeletedException cbde) {
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java-      status = AppFabricServiceStatus.PROGRAM_STILL_RUNNING;
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java-    } catch (Exception e) {
--
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java:    } catch (NamespaceCannotBeDeletedException e) {
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java-      responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java-    } catch (Exception e) {
--
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java:    } catch (NamespaceCannotBeDeletedException e) {
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java-      responder.sendString(HttpResponseStatus.FORBIDDEN, e.getMessage());
cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java-    } catch (Exception e) {
```

So its a mix of ``FORBIDDEN`` (delete app, delete namespaces - which has one ``FORBIDDEN`` and another ``CONFLICT``) and ``CONFLICT`` currently. If we were to change everything to ``CONFLICT``, we'd have two API changes. I am fine with standardizing all of these in this PR.

Thoughts?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
bdmogal: I'm ok with changing this too - about 7 occurrences of ``NotFoundException`` in ``AdapterHttpHandler``.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
bdmogal: Ok, it is too late in the release cycle to change everything right now, I'll add some comments to @anew's jira below and we can do all of these changes together at once during the next release.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AdapterHttpHandler.java
bdmogal: @jwang47 let's do this in a separate PR. Tried it, its increasing this PR's scope, not sure I want to do that late in the cycle. Let's do this as part of a bigger cleanup.

sreevatsanraman:
LGTM :+1:  Unrelated failure in build which is already fixed in release.

PR #2506, opened 2015-05-01
CDAP-2400 adding ability to add a list of classes to include in

albertshau:
the bundled template jar to simulate the plugins lib directory.
Also adding a method to add a plugin config json file in the
test framework. Lastly, updating etl test cases that needed this
functionality in order to pass.


albertshau:
http://builds.cask.co/browse/CDAP-RBT260

PR #2507, opened 2015-05-01
CDAP-2411 tweak the filter transform so that the property must

albertshau:
supply a shouldFilter function instead of placing you inside the
shouldFilter function. This allows users to define their own
helper functions, and makes it more explicit what is going on.


albertshau:
http://builds.cask.co/browse/CDAP-RT29

PR #2510, opened 2015-05-01
Fix routing for explore service status endpoints

anwar6953:
Explore Service Status endpoint was not working properly (which I noticed while trying to run JDBC driver against CDAP). The issue was a failure in routing.

Simply pulled in the following code from 2.8, to get it working again:
https://github.com/caskdata/cdap/blob/release/2.8/cdap-gateway/src/main/java/co/cask/cdap/gateway/router/RouterPathLookup.java#L91-L92

http://builds.cask.co/browse/CDAP-RBT261-1


jwang47:
LGTM

sreevatsanraman:
Unrelated test failure. Merging.

PR #2515, opened 2015-05-02
CDAP-2416: Make SecurityRequestContext.getUserId() consistent

jwang47:
https://issues.cask.co/browse/CDAP-2416
http://builds.cask.co/browse/CDAP-RBT268


gokulavasan:
@jwang47 Can you briefly describe how this change makes the SecurityRequestContext.getUserId() consistent?

jwang47:
`AuthenticationChannelHandler` intercepts the HTTP request before it is given to the one of the handler methods to process. Before, due to where `AuthenticationChannelHandler` was being executed in the `NettyHttpService` pipeline, sometimes `AuthenticationChannelHandler` would not be able to set the `SecurityRequestContext.getUserId()` before the handler method executed.

This change in the pipeline setup makes `AuthenticationChannelHandler` guaranteed to run before any of the handler methods are executed.

sreevatsanraman:
LGTM :+1:

PR #2518, opened 2015-05-02
[CDAP-2426] Add adapter name to workflow logging context

bdmogal:
Jira: [CDAP-2426](https://issues.cask.co/browse/CDAP-2426)
Build: http://builds.cask.co/browse/CDAP-RBT269

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
@@ -43,6 +43,7 @@ import co.cask.cdap.internal.app.runtime.ProgramRunnerFactory;
 import co.cask.cdap.internal.workflow.DefaultWorkflowActionSpecification;
 import co.cask.cdap.internal.workflow.ProgramWorkflowAction;
 import co.cask.cdap.logging.context.WorkflowLoggingContext;
+import co.cask.cdap.templates.AdapterDefinition;
 import co.cask.http.NettyHttpService;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
@@ -53,6 +54,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.reflect.TypeToken;
 import com.google.common.util.concurrent.AbstractExecutionThreadService;
+import com.google.gson.Gson;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -107,10 +110,15 @@
...
-                                                options.getArguments().getOption(ProgramOptionConstants.RUN_ID));
+    this.lock = new ReentrantLock();
+    this.condition = lock.newCondition();
+    String adapterSpec = arguments.getOption(ProgramOptionConstants.ADAPTER_SPEC);
+    String adapterName = null;
+    if (adapterSpec != null) {
+      adapterName = GSON.fromJson(adapterSpec, AdapterDefinition.class).getName();
+    }
+    this.loggingContext = new WorkflowLoggingContext(program.getNamespaceId(), program.getApplicationId(), program.getName(),
+                                                arguments.getOption(ProgramOptionConstants.RUN_ID), adapterName);
```
gokulavasan: fix alignment

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/log/LogHandlerTestRun.java
```diff
@@ -207,6 +207,49 @@ public void testAdapterLogsPrev() throws Exception {
   }
 
   @Test
+  public void testWorkflowLogs() throws Exception {
+    testLogs("testTemplate1", "workflows", "testWorkflow1", MockLogReader.TEST_NAMESPACE, "testAdapter1");
+    try {
+      // in default namespace, we did not produce logs through adapters
+      testNext("testTemplate1", "workflows", "testWorkflow1", true, Constants.DEFAULT_NAMESPACE, "testAdapter1");
+      Assert.fail();
+    } catch (AssertionError e) {
```
gokulavasan: Here we call Assert.fail() to throw an AssertionError and then we catch it and do nothing. So what is the test trying to test? Same comment for this pattern used below.

gokulavasan:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
bdmogal: Fixed.

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/log/LogHandlerTestRun.java
bdmogal: Great catch! This was a pattern throughout this test. Basically, we're expecting an ``AssertionError`` on line 214 above.  But throwing an ``Assert.fail()`` in the try defeats the purpose. I've fixed it in the next commit by moving this verification to the end of each test, and calling a return from the catch clause, then adding an ``Assert.fail()`` at the end of the test... also fixed it throughout this class.

bdmogal:
@gokulavasan addressed comments in the second commit

gokulavasan:
LGTM :+1:

bdmogal:

PR #2520, opened 2015-05-02
change formatting of start time on runs

elfenheart:
screen shot 2015-05-01 at 8 17 20 pm

### DIFF
--- a/cdap-ui/app/features/mapreduce/templates/tabs/runs/run-detail.html
+++ b/cdap-ui/app/features/mapreduce/templates/tabs/runs/run-detail.html
@@ -19,7 +19,7 @@
         </thead>
         <tbody>
           <td ng-class="{'txt-failed': info.state === 'FAILED', 'txt-running': info.state === 'RUNNING'}"> {{info.state || 'Not Available'}} </td>
-          <td> {{info.startTime || 'Not Available' | date: 'medium' }} </td>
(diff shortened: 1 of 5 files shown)


anwar6953:
https://github.com/caskdata/cdap/blob/develop/cdap-ui/app/features/spark/templates/tabs/runs/run-detail.html#L25

cdap-ui/app/features/mapreduce/templates/tabs/runs/run-detail.html
```diff
@@ -19,7 +19,7 @@
         </thead>
         <tbody>
           <td ng-class="{'txt-failed': info.state === 'FAILED', 'txt-running': info.state === 'RUNNING'}"> {{info.state || 'Not Available'}} </td>
-          <td> {{info.startTime || 'Not Available' | date: 'medium' }} </td>
+          <td> {{info.startTime || 'Not Available' | amDateFormat: 'MM/DD/YY h:mm:ss a' }} </td>
```
anwar6953: Why is this one different with 'Not Available'?

Also, what happens if you push that string through amDateFormat?

anwar6953: Hmm, can you remove Not Available from line 21 and 22? It will always be available.

anwar6953:
LGTM

PR #2522, opened 2015-05-02
Admin Namespace name ellipsis

elfenheart:
screen shot 2015-05-01 at 9 01 19 pm


ajainarayanan:
@elfenheart May not be in this PR but if we can add a tooltip to whereever we add ellipsis it would be helpful to know the full name (Not that anybody is going to have a name that is going to span more than 300px width :smile: )

ajainarayanan:
:+1: LGTM

PR #2528, opened 2015-05-02
Fixes source/sink & tranforms edit properites to popover default descriptions if not provided by nodejs config

ajainarayanan:
Fixes plugin property label in adapter create view provides a valid popover with default description if the backend doesn't provide one.


nitinmotgi:
LGTM :+1:

PR #2529, opened 2015-05-02
Remove alpha from the directory structure for UI

nitinmotgi:
- [x] Changed Service file for running in distributed mode. 
- [x] Cleaned up Singlenode UserInterface source to remove any references to alpha
- [x] Changed the pom.xml in cdap-ui and cdap-standalone to not reference 'alpha'


sreevatsanraman:
LGTM

PR #2530, opened 2015-05-02
Bugfix/cdap 2202 update invalid runrecords running

henry-cask:
Proposed fix for CDAP-2202 to update invalid RunRecords with RUNNING state 

Currently the calls to check run status of a program is run against run records' store.
In case of abrupt shutdown, there is no chance to update the state of the run record of the programs.
For example if a Flow program is running in standalone then the programs dies, then the run record
for that program will indicate still running when the CDAP is restarted.

So, the call for
```
http://localhost:10000/v3/namespaces/default/apps/CountRandom/flows/CountRandom/runs?status=running
```
and
```
http://localhost:10000/v3/namespaces/default/apps/CountRandom/flows/CountRandom/status
could return different results.
```

Adding schedule thread in ProgramLifecycleService to manage the Store updating the RunRecords with invalid state.

Makes ProgramLifecycleService to be Singleton to make sure only one instance of such thread exists.
At start of the service, it will check immediately for all invalid RunRecords.

Add CAS like method in the app Store to do the compare and set operation to make the update to be
atomic.

For Adapter we also need to check runtime status for the programs bc the program may not actually running.


henry-cask:
Build at:
http://builds.cask.co/browse/CDAP-RBT271-1

sreevatsanraman:
LGTM :+1: 
Changes looks good to me. But there are no tests. Can you add that in a follow up PR?

PR #2532, opened 2015-05-02
CDAP-2400 adding method to test adapter framework to add a

albertshau:
json config file for a template plugin.
Also updating the db etl test case to make use of the ability
and re-enabling the test.


gokulavasan:
LGTM :+1:

PR #2533, opened 2015-05-02
CDAP-2253 some etl cleanup. making package names singular instead

albertshau:
of plural, annotating everything in etl-api as beta, and removing
unused modules and classes.


albertshau:
http://builds.cask.co/browse/CDAP-RT30

gokulavasan:
LGTM :+1:

PR #2534, opened 2015-05-02
CDAP-2200 use a more descriptive job name for mapreduce so that

albertshau:
its easier to distinguish between jobs when looking at the
resource manager. Also change some log messages to debug.


albertshau:
apparently you aren't allowed to set job name... since it messes up falls cdap does to the job history server. I am going to open a bug for this.

PR #2536, opened 2015-05-02
Fixes viewing access token in user profile page when security is on

ajainarayanan:
-This was missed because when security is not enabled we don't show the login view and skip directly to the overview page.
- But when it is enabled and if the current user is not authenticated we show the login form.
- This login form is being reused(only the template) in the profile page which wanted the security to be enabled (which needed to be set in the current scope)


nitinmotgi:
LGTM :+1:

PR #2540, opened 2015-05-03
CDAP-2440: Use InMemoryDatasetFramework rather than RemoteDatasetFramework in UpgradeTool

jwang47:
https://issues.cask.co/browse/CDAP-2440
http://builds.cask.co/browse/CDAP-RBT273

Recent changes to UpgradeTool for upgrading queue.pending metrics wasn't working when cdap-master wasn't being run, so I reverted those changes to UpgradeTool.

Instead, you'll have to run FlowQueuePendingCorrector when CDAP is running like so:

```
/opt/cdap/master/bin/svc-master run co.cask.cdap.data.tools.flow.FlowQueuePendingCorrector
```


sreevatsanraman:
We should not have the dependency on kafka server being up. Let's decouple the command to correct pending metrics from the upgrade.

conf/generated/cdap-security-config.json
```diff
@@ -0,0 +1 @@
+{}
```
sreevatsanraman: Remove this file. not needed.

conf/generated/cdap-config.json
```diff
@@ -0,0 +1,182 @@
...
+  "metrics.processor.num.cores": "1",
+  "router.forward.rule": "10000:gateway,20000:webapp/$HOST",
+  "hdfs.user": "yarn",
+  "security.token.digest.keylength": "128",
+  "data.tx.timeout": "30",
+  "metrics.data.table.retention.resolution.60.seconds": "2592000",
+  "zookeeper.session.timeout.millis": "40000",
+  "security.token.digest.key.expiration.ms": "3600000",
+  "app.output.dir": "/programs"
+}
```
sreevatsanraman: Remove this file. not needed.

sreevatsanraman:
LGTM :+1: Build passed merging.

PR #2544, opened 2015-05-03
Fixes node proxy config path to be relative to server (after removal of alpha folder)

ajainarayanan:


sreevatsanraman:
LGTM :+1:

PR #2548, opened 2015-05-03
Fixes tables default header color + datasets non-explorable view + add in apps list view

ajainarayanan:
- Fixes add app in apps list view.
- Fixes Dataset detailed view of explorable datasets.
- Fixes global table header view (reverses the background color with text color)


elfenheart:
:+1: LGTM

PR #2554, opened 2015-05-04
Fixes global table header font-weight + datasets list styling fixes

ajainarayanan:
- Fixes font-weight of global tables 
- Fixes datasets list view to have a margin top from the tab header
- Fixes font-weight on datasets list view in datasets tab of programs


elfenheart:
:+1: LGTM

PR #2556, opened 2015-05-04
Check against all programs and cross check it against the run record Store to get the right program id to update

henry-cask:
Previous fix on CDAP-2202 did not cover case the runtime info contain info about program based on invalid run record id.

Need to cross check it against the actual list of programs across apps and namesapces to get the right program id.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
@@ -16,6 +16,10 @@
 
 package co.cask.cdap.internal.app.services;
 
+import co.cask.cdap.api.ProgramSpecification;
+import co.cask.cdap.api.flow.FlowSpecification;
+import co.cask.cdap.api.mapreduce.MapReduceSpecification;
+import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.program.Program;
 import co.cask.cdap.app.runtime.ProgramController;
 import co.cask.cdap.app.runtime.ProgramRuntimeService;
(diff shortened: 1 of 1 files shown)


henry-cask:
CC @sreevatsanraman could you try this out?

henry-cask:
Do not merge this one first until confirmation from Sree

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -245,9 +245,22 @@ public boolean apply(@Nullable RunRecord input) {
     // Now lets correct the invalid RunRecords
     for (RunRecord rr : invalidRunRecords) {
       String runId = rr.getPid();
-      RuntimeInfo runtimeInfo = runIdToRuntimeInfo.get(RunIds.fromString(runId));
```
sreevatsanraman: Why don't we just do 
```
RuntimeInfo runtimeInfo = runIdToRuntimeInfo.get(RunIds.fromString(runId));

if (runtimeInfo != null) {

store.compareAndSetStatus(runtimeInfo.getProgramId(), runId, ProgramController.State.ALIVE.getRunStatus(),
                          ProgramController.State.ERROR.getRunStatus());
}
```

sreevatsanraman:
@henry-cask - i tried this out. It works for the case where we shutdown cdap cleanly. if the process crashes then the runhistory is not fixed. We might want to fix that in a followup PR in 3.1

henry-cask:
Close this for now

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
henry-cask: We could do that but it would just work for properly start app and its program.

henry-cask:
@sreevatsanraman could you try this one. Should cover the abrupt stop for master or any container that runs program in cluster.

henry-cask:
Build:
http://builds.cask.co/browse/CDAP-RBT277-1

sreevatsanraman:
LGTM :+1: Will merge once the build passes.

sreevatsanraman:
Build passed, merging.

PR #2566, opened 2015-05-04
Sort lists in alphabetical order, as opposed to reverse alphabetical ord...

anwar6953:
...er.

I have removed 'reverse' as being the default order, because that seems unusual.
Natural alphabetical order seems more appropriate.

It remains only in one place - `program-history.html` - because that's how it is showing most recent runs first.


anwar6953:
Reopening against release.

PR #2570, opened 2015-05-04
CDAP-2441 - Override MR job name in case user set it in beforeSubmit.

anwar6953:
Does not completely resolve CDAP-2441, so will leave the issue open.

https://issues.cask.co/browse/CDAP-2441
http://builds.cask.co/browse/CDAP-RBT281-1


chtyim:
LGTM

PR #2571, opened 2015-05-04
Give bar charts tick culling, to avoid crowded timestamps.

anwar6953:
Before
image
---------
After
image


elfenheart:
:+1: LGTM

PR #2572, opened 2015-05-04
Adapter Detail styling + small fixes

elfenheart:


cdap-ui/app/features/admin/controllers/NamespaceDatasetMetadataController.js
```diff
@@ -1,15 +1,27 @@
...
     _cdapPath: '/namespaces/' + $state.params.nsadmin
-                  + '/data/explore/tables/dataset_' + $state.params.datasetId + '/info'
-  }).then(function (res) {
-    $scope.metadata = res;
+                  + '/data/explore/tables'
+  }).then(function (tables) {
+    var match = $filter('filter')(tables, $state.params.datasetId);
+    if (match.length > 0) {
+      dataSrc.request({
+        _cdapPath: '/namespaces/' + $state.params.nsadmin
```
anwar6953: It's weird that we call this variable `nsadmin`. It's a string that is the namespace, right?

cdap-ui/app/features/admin/controllers/NamespaceDatasetMetadataController.js
```diff
@@ -1,15 +1,27 @@
 angular.module(PKG.name + '.feature.admin').controller('AdminNamespaceDatasetMetadataController',
```
anwar6953: What issue does the change in this file fix?
Also, Why couldn't we do the equivalent simply by setting `$scope.metadata = null` in the case of a 404?

cdap-ui/app/features/admin/controllers/NamespaceDatasetMetadataController.js
elfenheart: That is the the namespace ID on the admin level. We have to differentiate with just namespace on other pages.

cdap-ui/app/features/admin/controllers/NamespaceDatasetMetadataController.js
elfenheart: We can't do that, because if it ever throw a 404 error, it will show up in myAlert, since 404 error is handled globally.

anwar6953: Ok. In 3.1 maybe, we should change that, right?
When making a request, we should have some flexibility in allowing 404s, and handling them appropriately in the controller ourself.

cdap-ui/app/features/admin/controllers/NamespaceDatasetMetadataController.js
anwar6953: Ok.

cdap-ui/app/features/admin/templates/namespace/dataset-metadata.html
```diff
@@ -63,4 +63,8 @@
     <td>Is Dataset</td>
     <td>{{ metadata.from_dataset }}</td>
   </tr>
-</table>
\ No newline at end of file
+</table>
+
+<div class="well text-center" ng-if="!metadata">
+  <h4>No metadata for this dataset</h4>
```
anwar6953: If metadata is null, doesn't that mean that the dataset does not exist? (as opposed to its metadata simply missing)?

elfenheart: It means dataset is not explorable, not in Hive

cdap-ui/app/features/admin/controllers/NamespaceDatasetMetadataController.js
anwar6953: Not something for now, then.

cdap-ui/app/features/admin/templates/namespace/dataset-metadata.html
anwar6953: got it.

anwar6953:
LGTM.

PR #2580, opened 2015-05-04
fix template name

albertshau:


gokulavasan:
LGTM :+1:

gokulavasan:
http://builds.cask.co/browse/CDAP-RT31-1

PR #2581, opened 2015-05-04
CDAP-2463 Using right log offset to fetch logs from Kafka

poornachandra:
Also, closing log callback in handler instead of log readers since now each call to fetch logs can span multiple readers.


poornachandra:
Build - http://builds.cask.co/browse/CDAP-RBT282-2

sreevatsanraman:
LGTM :+1:

PR #2586, opened 2015-05-04
set namespace to default when there is no session storage

elfenheart:
initial to default namespace

### DIFF
--- a/cdap-ui/app/features/home/routes.js
+++ b/cdap-ui/app/features/home/routes.js
@@ -39,7 +39,7 @@ angular.module(PKG.name+'.feature.home')
             return myNamespace.getList();
           }
         },
-        controller: function ($state, rNsList, mySessionStorage, myLoadingService) {
+        controller: function ($state, rNsList, mySessionStorage, myLoadingService, myAlert, $filter) {
           // check that $state.params.namespace is valid
           var n = rNsList.filter(function (one) {
             return one.name === $state.params.namespace;
@@ -51,7 +51,20 @@ angular.module(PKG.name+'.feature.home')
           if(!n.length) {
             mySessionStorage.get(PREFKEY)
               .then(function (latest) {
-                var d = latest || 'default';
+
+                var def = $filter('filter')(rNsList, {name: 'default'}, true);
+
+                if (def.length < 1) {
+                  def = rNsList[0];
+                  myAlert({
+                    title: 'Cannot find default namespace',
+                    content: 'Reverting to ' + def.name + ' namespace.'
+                  });
+                } else {
+                  def = def[0];
+                }
+
+                var d = latest || def.name;
                 console.warn('invalid namespace, defaulting to', d);
                 $state.go(
                   $state.current,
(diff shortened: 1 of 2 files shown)


cdap-ui/app/features/home/routes.js
```diff
@@ -51,7 +51,7 @@ angular.module(PKG.name+'.feature.home')
           if(!n.length) {
             mySessionStorage.get(PREFKEY)
               .then(function (latest) {
-                var d = latest || rNsList[0].name;
+                var d = latest || 'default';
```
bdmogal: isn't there a variable somewhere? Also, I'm not sure about UI code, but we should make sure that the default namespace exists, if not, bail out with a dialog saying 'default namespace not found. Please try restarting CDAP.' or something to that effect.

Thoughts?

cdap-ui/app/features/home/routes.js
```diff
@@ -51,7 +51,20 @@ angular.module(PKG.name+'.feature.home')
           if(!n.length) {
             mySessionStorage.get(PREFKEY)
               .then(function (latest) {
-                var d = latest || rNsList[0].name;
+
+                var def = $filter('filter')(rNsList, {name: 'default'}, true);
```
bdmogal: nit: call it ``defaultNs``?

cdap-ui/app/features/home/routes.js
```diff
@@ -51,7 +51,20 @@ angular.module(PKG.name+'.feature.home')
...
+                  def = rNsList[0];
+                  myAlert({
+                    title: 'Cannot find default namespace',
+                    content: 'Reverting to ' + def.name + ' namespace.'
+                  });
+                } else {
+                  def = def[0];
+                }
+
+                var d = latest || def.name;
```
bdmogal: nit: use something more meaningful than 'd'?

cdap-ui/app/features/home/routes.js
anwar6953: The third parameter is only useful when the 2nd parameter is a string, right?
Otherwise, substring match doesn't make sense, right?

cdap-ui/app/features/home/routes.js
```diff
@@ -51,7 +51,20 @@ angular.module(PKG.name+'.feature.home')
           if(!n.length) {
             mySessionStorage.get(PREFKEY)
               .then(function (latest) {
-                var d = latest || rNsList[0].name;
+
+                var def = $filter('filter')(rNsList, {name: 'default'}, true);
+
+                if (def.length < 1) {
```
anwar6953: Can we just use === 0? It's not a big deal, but most other places, we do zero comparison, rather than lessThan comparison with 1.

anwar6953:
Also, this comment got hidden I guess.
https://github.com/caskdata/cdap/pull/2586#discussion_r29637101

cdap-ui/app/features/home/routes.js
elfenheart: No, it is calling angular.equals which does an object comparison.

anwar6953:
LGTM

PR #2590, opened 2015-05-05
updated 3.0 screens for CSD Wizard

dereklwood:
- [x] updated screenshots of CSD wizard to reflect 3.0 (CDAP-UI)
- [ ] does not include the final UI shot, which is cross-referenced to this page:  http://docs.cdap.io/cdap/current/en/admin-manual/operations/cdap-console.html


awholegunch:
What I notice is that in the Second image, there is a link to "CDAP Console" in both on the left side. Should that be "CDAP UI" instead?

dereklwood:
I was not aware we were changing Console to UI?  I can change it in the CSD, but I still see Console all over the docs.

awholegunch:
That's because the change is still in a PR. :)

PR #2596, opened 2015-05-05
Return back dependency on JMS for realtime lib bc it is breaking class logic

henry-cask:
Seemed like omitting JMS dependency spec caused problem with ETLAdapter tests


gokulavasan:
LGTM :+1:

gokulavasan:
Merge if build passes.

henry-cask:
http://builds.cask.co/browse/CDAP-RT33-1

PR #2600, opened 2015-05-05
modify remote incoming dir for snapshot artifacts only

duddun:
Redirect snapshot artifacts to where they should go on remote server.


wolf31o2:
:+1: LGTM

PR #2601, opened 2015-05-05
Fix/ui adapter optional property

ajainarayanan:
- Handles in UI if property of a plugin is optional and if it is null remove the property from config before saving.
- Minor fixes to json config files from node proxy.
- Minor styling fixes in adapter list view


elfenheart:
:+1: LGTM

PR #2602, opened 2015-05-05
Exclude JMS spec from ETL lib bundle to avoid duplicate dependencies.

henry-cask:
With both JSM spec jar and ActiveMQ jar in the classloader, we could get duplicate declaration of ConnectionFactory when calling:
```
connectionFactory = (ConnectionFactory) jndiContext.lookup(connectionFactoryName);
```
could generate error org.apache.activemq.ActiveMQConnectionFactory cannot be cast to javax.jms.ConnectionFactory


henry-cask:
http://builds.cask.co/browse/CDAP-RT34

sreevatsanraman:
Let's not merge this in 3.0

henry-cask:
Wont be needed if we copy the JMS jar to templates/plugins/ETLRealtime/lib.
See:
https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginClassLoader.java#L33

PR #2607, opened 2015-05-05
Remove an extra schema element in the schema sent to the backend

bdmogal:


bdmogal:
will open against release

PR #2616, opened 2015-05-05
[CDAP-2386] Checkstyle improvements

anew:
- adds a rule to checkstyle to find redundant modifiers (JLS recommends not have these):
  - public in interfaces, 
  - static for interfaces
  - abstract in interfaces
  - final in final classes
- fixes all places where that new rule triggers
- fixes a bunch of other checkstyle warnings


chtyim:
LGTM.

PR #2620, opened 2015-05-05
Remove unused variables

anwar6953:
Perfection is reached not when there’s nothing left to add, but when there’s nothing left to remove.

### DIFF
--- a/cdap-ui/app/directives/instance-control/instance-control.js
+++ b/cdap-ui/app/directives/instance-control/instance-control.js
@@ -9,7 +9,7 @@ angular.module(PKG.name + '.commons')
         basePath: '='
       },
       templateUrl: 'instance-control/instance-control.html',
-      link: function (scope, element, attrs) {
+      link: function (scope) {
 
         scope.processing = false;
 
@@ -30,9 +30,9 @@ angular.module(PKG.name + '.commons')
             method: 'PUT',
             _cdapPath: scope.basePath + '/instances',
             body: {'instances': scope.model.requested}
-          }).then(function success (response) {
+          }).then(function success () {
             scope.model.provisioned = scope.model.requested;
-          }, function error (err) {
+          }, function error () {
             //pass
           }).finally(function () {
             scope.processing = false;
--- a/cdap-ui/app/directives/instance-control/instance-control.js
+++ b/cdap-ui/app/directives/instance-control/instance-control.js
@@ -32,8 +32,6 @@ angular.module(PKG.name + '.commons')
             body: {'instances': scope.model.requested}
           }).then(function success () {
             scope.model.provisioned = scope.model.requested;
-          }, function error () {
-            //pass
           }).finally(function () {
             scope.processing = false;
           });
--- a/cdap-ui/app/features/workflows/controllers/tabs/runs/tabs/status-ctrl.js
+++ b/cdap-ui/app/features/workflows/controllers/tabs/runs/tabs/status-ctrl.js
@@ -106,6 +106,61 @@ angular.module(PKG.name + '.feature.workflows')
 
   });
 
+/**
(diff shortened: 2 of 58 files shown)


ajainarayanan:
@anwar6953 fileselect.js , status-ctrl.js in workflows can be removed. @elfenheart Can you check why we are not using runids in flowlet input-ctrl.js ?

Otherwise LGTM

elfenheart:
@anwar6953 @ajainarayanan in pending metrics, there is no concept of runid. It is always an aggregation. This is the fix that Andreas created for system.queue.pending metrics for flowlet

anwar6953:
It's strange then that we show outgoing metrics by runId, but pending metrics across runIds.
Can be quite confusing when looking at the values.

elfenheart:
I think the fix should be to show Not Available or 0 for past runs. That information is only useful for the currently running flows.

anwar6953:
Even if only showing information for the currently running flow, the pending is showing across all runs, while the outgoing is showing only for that run.

elfenheart:
how else would you do it? We have the information for outgoing for specific run, but there is no run specific pending metrics

anwar6953:
I'm not sure how I'd do it. But I definitely think it needs to be revisited. I'll remove runId from the code for now, and add it to Daily Tasks doc.

anwar6953:
Ok, removed the remaining 4.

cdap-ui/app/directives/instance-control/instance-control.js
```diff
@@ -30,9 +30,9 @@ angular.module(PKG.name + '.commons')
             method: 'PUT',
             _cdapPath: scope.basePath + '/instances',
             body: {'instances': scope.model.requested}
-          }).then(function success (response) {
+          }).then(function success () {
             scope.model.provisioned = scope.model.requested;
-          }, function error (err) {
+          }, function error () {
```
elfenheart: I think you can remove this entire error function. It's not doing anything

cdap-ui/app/features/workflows/controllers/tabs/runs/tabs/status-ctrl.js
```diff
@@ -108,93 +106,6 @@ angular.module(PKG.name + '.feature.workflows')
...
-      sourceName: nodes[nodes.length - 2].name,
-      sourceType: nodes[nodes.length - 2].nodeType,
-      targetName: nodes[nodes.length - 1].name
-    });
-
-  }
-
-}
-
-/**
```
ajainarayanan: This comment belongs to the function below it.. I don't believe we should remove this one?

cdap-ui/app/features/workflows/controllers/tabs/status-ctrl.js
```diff
@@ -52,100 +49,13 @@ angular.module(PKG.name + '.feature.workflows')
...
-        sourceName: nodes[nodes.length - 2].name,
-        sourceType: nodes[nodes.length - 2].nodeType,
-        targetName: nodes[nodes.length - 1].name
-      });
-
-    }
-
-  }
-
-  /**
```
ajainarayanan: Same here too.

cdap-ui/app/features/workflows/controllers/tabs/runs/tabs/status-ctrl.js
anwar6953: Good point. I didn't realize that. Added back.

cdap-ui/app/directives/instance-control/instance-control.js
anwar6953: removed.

ajainarayanan:
:+1: LGTM

PR #2623, opened 2015-05-06
Fixes stylings in adapter create view

ajainarayanan:
Don't know how this got removed. Anyways adding it back in develop (https://github.com/caskdata/cdap/blob/release/3.0/cdap-ui/app/features/adapters/etlapps.less#L70)

### DIFF
--- a/cdap-ui/app/features/adapters/etlapps.less
+++ b/cdap-ui/app/features/adapters/etlapps.less
@@ -51,7 +51,7 @@ body.theme-cdap.state-adapters-create {
       .default { display: none;}
       .template,.default {
         padding: 20px;
-        border: 2px dashed white;
+        border: 1px dashed white;
         color: white;
         height: 100%;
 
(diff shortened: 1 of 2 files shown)


anwar6953:
Is it possible that the change in release hadn't been merged to develop, yet? (If it went into release/3.0 branch quite late)?

anwar6953:
Nevermind, these changes aren't showing up in the following diff, so I don't think that was the case.
https://github.com/caskdata/cdap/compare/release/3.0

cdap-ui/app/features/adapters/etlapps.less
```diff
@@ -179,7 +185,14 @@ body.theme-cdap.state-adapters-create {
...
         }
-        .control-label { font-weight: 400; }
+        .row-padding {
+          padding: 0 30px;
+          .popover {
+            max-width: 550px;
+            width: 500px;
+          }
+        }
+        // .control-label { font-weight: 400; }
```
astewes: Since commented out, can it be removed?

cdap-ui/app/features/adapters/etlapps.less
```diff
@@ -67,7 +67,13 @@ body.theme-cdap.state-adapters-create {
         }
       }
       .template {
+        &.plugin-added {
+          background-color: #2c498b;
```
astewes: Nit-picky, but is either of these background colors (lines 70, 75) a variable? If not, feel free to disregard.

astewes:
A couple of very minor comments otherwise LGTM

elfenheart:
:+1: LGTM

PR #2624, opened 2015-05-06
Add downloading of any image files referenced in the docs.

awholegunch:
This changes the examples-manual build to replace any local references to image files with a reference to a downloaded image file. After an rst file is downloaded, all image references in it are extracted, downloaded and adjusted.

Tested on both Mac OS X and CentOS 6, with no other directories but cdap available.

### DIFF
--- a/cdap-docs/examples-manual/build.sh
+++ b/cdap-docs/examples-manual/build.sh
@@ -26,24 +26,38 @@ CHECK_INCLUDES=${TRUE}
 TUTORIAL_WISE="tutorial-wise"
 
 function guide_rewrite_sed() {
-  echo "Re-writing using sed ${1} ${2}"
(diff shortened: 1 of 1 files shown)


awholegunch:
The one question I have is which directories were you using in addition to cdap? It should have been all the Guides directories. If there were any others, then this code would need adjusting.

awholegunch:
The other two repos were cdap-ingest and cdap-clients; I don't think these are required anymore. I do see errors on the new staging-server builds. Those might require resolving some other way.

duddun:
The only three used in the bamboo automated doc build plans are: cdap, cdap-ingest, cdap-clients.
If we don't need cdap-ingest or cdap-clients, those portions can easily be removed from the plans.

duddun:
LGTM

PR #2625, opened 2015-05-06
Rewrite try-finally-close block using Java7 ARM construct, which is the best practice.

chtyim:
- Better error handling
  - Exception from Closeable.close() won’t shadow the actual IOException
- Not using Closelables.closeQuietly so that error on close is still
  available in the stacktrace (record as suppressed exception)

Mainly main classes are done. Most test classes are not changed

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/archive/ArchiveBundler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/archive/ArchiveBundler.java
@@ -112,48 +112,42 @@ public final class ArchiveBundler {
     Preconditions.checkNotNull(files);
 
     // Create a input stream based on the original archive file.
-    JarInputStream zin = new JarInputStream(archive.getInputStream());
-    try {
-      // Create a new output JAR file with new MANIFEST.
+    // Create a new output JAR file with new MANIFEST.
+    try (
+      JarInputStream zin = new JarInputStream(archive.getInputStream());
       JarOutputStream zout = new JarOutputStream(new BufferedOutputStream(output.getOutputStream()),
-                                                 combineManifest(zin.getManifest(), manifest));
-
-      try {
-        // Iterates through the input zip entry and make sure, the new files
-        // being added are not already present. If not, they are added to the
-        // output zip.
-        JarEntry entry = zin.getNextJarEntry();
-        while (entry != null) {
-          // Invoke the predicate to see if the entry needs to be filtered.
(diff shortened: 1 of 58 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/archive/ArchiveBundler.java
```diff
@@ -112,48 +112,42 @@ public void clone(Location output, Manifest manifest,
...
+        try {
+          ByteStreams.copy(toAdd.getValue(), zout);
+        } finally {
+          zout.closeEntry();
         }
-      } finally {
-        zout.close();
       }
-    } finally {
-      zin.close();
```
awholegunch: zin and zout don't get closed now?

anwar6953: They automatically get closed now. They just need to be declared in the `try()`

chtyim: Yes, close is no longer needed when using the new `try () { }` syntax. That's the new and better syntax supported by Java7. See http://www.oracle.com/technetwork/articles/java/trywithresources-401775.html for details

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/hbase/HBaseStreamConsumerStateStoreFactory.java
```diff
@@ -63,8 +63,6 @@ public synchronized StreamConsumerStateStore create(StreamConfig streamConfig) t
 
         tableUtil.createTableIfNotExists(admin, streamStateStoreTableId, htd, null,
                                          QueueConstants.MAX_CREATE_TABLE_WAIT, TimeUnit.MILLISECONDS);
-      } finally {
-        admin.close();
```
jwang47: We weren't closing this before all the time, huh?

jwang47:
LGTM

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/hbase/HBaseStreamConsumerStateStoreFactory.java
chtyim: Right, the old code has bug.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1805-4

PR #2627, opened 2015-05-06
add livereload for gulp

elfenheart:
Adding livereload back to gulp

### DIFF
--- a/cdap-ui/app/index.html
+++ b/cdap-ui/app/index.html
@@ -62,5 +62,12 @@
     <script type="text/javascript" src="/assets/bundle/app.js"></script>
     <script type="text/javascript" src="/assets/bundle/tpl.js"></script>
 
+    <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


anwar6953:
Why do we check if its localhost?
If gulp is running also on a different host, shouldn't it also do then?

anwar6953:
Btw whats the reason we remove it before release?

elfenheart:
improving performance. That script is a deadweight on a production environment, adding a pub/sub events that is doing nothing.

anwar6953:
Makes sense. Is there any way we can simply inject it if its built by gulp in development mode?
(That way we don't have to remove and bring it back every release/patch).

elfenheart:
I believe we can do that, but let's do that in separate PR. Have to restructure the Gulp file, and have to dig through the build script.

anwar6953:
Ok, this LGTM

PR #2628, opened 2015-05-06
Worker pages.

anwar6953:
Worker had no page in the UI. Adding the most basic functionalities to it.
Start, stop, status, runs, history, logs.

image
---------------------
Note: The code duplication is an issue across all program types. It will be resolved in a follow up PR that touches all program types.

### DIFF
--- a/cdap-ui/app/features/programs/module.js
+++ b/cdap-ui/app/features/programs/module.js
@@ -5,4 +5,5 @@ angular.module(PKG.name + '.feature.programs', [
   PKG.name + '.feature.flows',
   PKG.name + '.feature.workflows',
   PKG.name + '.feature.spark',
+  PKG.name + '.feature.worker',
 ]);
--- a/cdap-ui/app/features/programs/module.js
+++ b/cdap-ui/app/features/programs/module.js
@@ -5,5 +5,5 @@ angular.module(PKG.name + '.feature.programs', [
   PKG.name + '.feature.flows',
   PKG.name + '.feature.workflows',
   PKG.name + '.feature.spark',
-  PKG.name + '.feature.worker',
+  PKG.name + '.feature.worker'
 ]);
--- a/cdap-ui/app/features/workers/controllers/tabs/runs/runs-detail-ctrl.js
+++ b/cdap-ui/app/features/workers/controllers/tabs/runs/runs-detail-ctrl.js
@@ -13,7 +13,6 @@ angular.module(PKG.name + '.feature.worker')
     $scope.activeTab = $scope.tabs[0];
 
     $scope.$on('$destroy', function(event) {
-      debugger;
       event.targetScope.runs.selected = null;
     });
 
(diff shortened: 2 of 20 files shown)


cdap-ui/app/features/programs/module.js
```diff
@@ -5,4 +5,5 @@ angular.module(PKG.name + '.feature.programs', [
   PKG.name + '.feature.flows',
   PKG.name + '.feature.workflows',
   PKG.name + '.feature.spark',
+  PKG.name + '.feature.worker',
```
ajainarayanan: trailing comma?

cdap-ui/app/features/workers/controllers/tabs/runs/runs-detail-ctrl.js
```diff
@@ -0,0 +1,24 @@
...
+    },
+    {
+      title: 'Logs',
+      template: '/assets/features/workers/templates/tabs/runs/tabs/log.html'
+    }];
+
+    $scope.activeTab = $scope.tabs[0];
+
+    $scope.$on('$destroy', function(event) {
+      debugger;
```
ajainarayanan: remove?

cdap-ui/app/features/workers/templates/tabs/runs/tabs/status.html
```diff
@@ -0,0 +1,2 @@
+<div ng-controller="WorkersRunsDetailStatusController" class="container">
```
ajainarayanan: why container? we don't seem to be using container in other places right?

cdap-ui/app/features/workers/templates/tabs/runs.html
```diff
@@ -0,0 +1,32 @@
...
+</div>
+<br/>
+<div ng-if="runs.length">
+  <div ng-include="'/assets/features/workers/templates/tabs/runs/run-detail.html'"
+        ng-if="!$stateParams.runid">
+  </div>
+  <div ui-view  ng-if="$stateParams.runid"> </div>
+</div>
+<div ng-if="!runs.length">
+  <div class="well well-lg text-center h3">
```
ajainarayanan: Might not be in this PR but we should probably not use h3 class here. Seems to be a hacky way of adding some margin on the top of a well.. we should probably have the tabs-content to have a padding top and have all children to use that instead of each children having a margin top.

cdap-ui/app/features/workers/controllers/tabs/runs/runs-detail-ctrl.js
anwar6953: removed.

cdap-ui/app/features/programs/module.js
anwar6953: removed.

cdap-ui/app/features/workers/templates/tabs/runs/tabs/status.html
anwar6953: removed.

cdap-ui/app/features/workers/templates/tabs/runs.html
anwar6953: Removed the ng-if, to be more consistent with flows page.

ajainarayanan:
:+1: LGTM

PR #2633, opened 2015-05-06
Fix/ui json editor

ajainarayanan:
- Adds json editor to return a string instead of json
- Fixes adapter create view when template type changes (default plugins)

### DIFF
--- a/cdap-ui/app/directives/widget-container/widget-container.less
+++ b/cdap-ui/app/directives/widget-container/widget-container.less
@@ -1,6 +1,5 @@
 .my-widget-container {
   .form-control {
-    // width: 90%;
     display: inline-block;
     vertical-align: middle;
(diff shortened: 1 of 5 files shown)


cdap-ui/app/directives/widget-container/widget-container.less
```diff
@@ -1,6 +1,6 @@
 .my-widget-container {
   .form-control {
-    width: 90%;
+    // width: 90%;
```
elfenheart: Can we just remove this? With the full width, we don't need to reduce the size of the input elements right?

ajainarayanan: Fixed.

elfenheart:
:+1: LGTM

PR #2635, opened 2015-05-06
consolidate streams and datasets array

elfenheart:
Consolidate Streams and Datasets array into one dataList array so that sorting will work properly in admin page.

### DIFF
--- a/cdap-ui/app/features/admin/templates/namespace/datasets.html
+++ b/cdap-ui/app/features/admin/templates/namespace/datasets.html
@@ -3,43 +3,36 @@
     <div class="col-xs-6">
       <h2> Datasets </h2>
     </div>
-    <div class="col-xs-6" ng-if="datasets.length > 0 || streams.length > 0">
+    <div class="col-xs-6" ng-if="dataList.length > 0">
       <button class="btn btn-default pull-right" ui-sref="admin.namespace.detail.data.streamcreate({nsadmin: $stateParams.nsadmin})"><span class="fa fa-plus"></span> Add Stream</button>
     </div>
   </div>
-  <div ng-if="datasets.length > 0 || streams.length > 0" datasets-list>
+  <div ng-if="dataList.length > 0" datasets-list>
     <div class="table-responsive">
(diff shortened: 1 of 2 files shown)


cdap-ui/app/features/admin/templates/namespace/datasets.html
```diff
@@ -3,43 +3,36 @@
...
           </tr>
         </thead>
         <tbody>
-          <tr ng-repeat="dataset in datasets | orderBy:sortable.predicate:sortable.reverse">
-            <td class="col-xs-5">
-              <strong class="icon-datasets"></strong>
-              <a ui-sref="admin.namespace.detail.data.datasetmetadata({datasetId: dataset.name})">
-
-                <strong ng-bind="dataset.name"></strong>
+          <tr ng-repeat="data in dataList | orderBy:sortable.predicate:sortable.reverse">
```
ajainarayanan: We switched default sorting to be alphabetical instead of reverse right?

ajainarayanan:
One minor comment otherwise LGTM

cdap-ui/app/features/admin/templates/namespace/datasets.html
anwar6953: Yea, this doesn't define that it it is reverse by default - simply that the list is reversable.

elfenheart: This is already consistent with the other orderBy's and defaulting to alphabetical sorting. I think that sortable.reverse gets enabled when you click on the sortable dropdown.

ajainarayanan: If the list is not reversible then what are we sorting on? I am not sure if I understand why we need it in the first place.

anwar6953: If clicking the column A, then it will sort by that column. Clicking that column A again will reverse the elements, only if you have `sortable.reverse`. Without that, it won't sort in reverse order.

PR #2636, opened 2015-05-06
Feature/r3.0 fix Introduction example

awholegunch:
Staged at: [Introduction](http://docs-staging.cask.co/cdap/3.0.0-r3.0_fix_examples/en/introduction/index.html)

There are a number of errors in the Introduction. This fixes them, including providing instructions on downloading a pre-built version of the Wise App.


duddun:
LGTM

PR #2638, opened 2015-05-07
CDAP-1868: Set dataset properties support for client and CLI

jwang47:
https://issues.cask.co/browse/CDAP-1868
http://builds.cask.co/browse/CDAP-DUT1879

* Also removes requirement for passing in dataset type to PUT /data/datasets/{name}/properties endpoint (not backwards-compatible)

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/SetDatasetInstancePropertiesCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/SetDatasetInstancePropertiesCommand.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.cli.command;
+
+import co.cask.cdap.cli.ArgumentName;
+import co.cask.cdap.cli.CLIConfig;
+import co.cask.cdap.cli.ElementType;
+import co.cask.cdap.cli.english.Article;
+import co.cask.cdap.cli.english.Fragment;
+import co.cask.cdap.cli.util.AbstractCommand;
+import co.cask.cdap.cli.util.ArgumentParser;
+import co.cask.cdap.client.DatasetClient;
+import co.cask.common.cli.Arguments;
+import com.google.gson.Gson;
+import com.google.inject.Inject;
+
+import java.io.PrintStream;
+import java.util.Map;
+
+/**
+ * Sets properties for a dataset instance.
+ */
+public class SetDatasetInstancePropertiesCommand extends AbstractCommand {
+
+  private static final Gson GSON = new Gson();
+  private final DatasetClient datasetClient;
+
+  @Inject
+  public SetDatasetInstancePropertiesCommand(DatasetClient datasetClient, CLIConfig cliConfig) {
+    super(cliConfig);
+    this.datasetClient = datasetClient;
+  }
+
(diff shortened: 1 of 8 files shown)


cdap-cli/src/main/java/co/cask/cdap/cli/command/SetDatasetInstancePropertiesCommand.java
```diff
@@ -0,0 +1,69 @@
...
+    this.datasetClient = datasetClient;
+  }
+
+  @Override
+  public void perform(Arguments arguments, PrintStream output) throws Exception {
+    String name = arguments.get(ArgumentName.DATASET.toString());
+    Map<String, String> properties = ArgumentParser.parseMap(
+      arguments.get(ArgumentName.DATASET_PROPERTIES.toString()));
+
+    datasetClient.update(name, properties);
```
chtyim: Does it requires a complete set of properties? I remember the REST call did (which is not nice, as if someone wants to update, he actually needs to fetch the existing properties, modifies those he wanted and send back the whole map. It is, however, acceptable since it's a PUT call). It will especially not nice to do it from CLI though.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
```diff
@@ -100,52 +103,50 @@ public DatasetInstanceHandler(DatasetTypeManager implManager, DatasetInstanceMan
...
+
+    Id.DatasetType datasetTypeId = Id.DatasetType.from(namespaceId, spec.getType());
+    DatasetTypeMeta typeMeta = getTypeInfo(Id.Namespace.from(namespaceId), spec.getType());
+    if (typeMeta == null) {
+      // TODO: This shouldn't happen unless CDAP is in an invalid state - maybe give different error
+      throw new NotFoundException(datasetTypeId);
+    }
+
+    // register ownership
+    for (String owner : owners) {
```
chtyim: Not good to have logic in handler (I know there are existing one, but we all know we want to migrate away from the current way).

cdap-cli/src/main/java/co/cask/cdap/cli/command/SetDatasetInstancePropertiesCommand.java
jwang47: Yes, it will currently. Should I add a new endpoint to allow updating existing properties, or fetch the properties in the CLI?

chtyim: I think for now it's ok to have the CLI fetch the existing one, modify it and send it back. Also open a JIRA for evaluating the existing dataset properties APIs.

jwang47: Opened https://issues.cask.co/browse/CDAP-2553

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
jwang47: Opened https://issues.cask.co/browse/CDAP-2557. Rather not do it in this PR.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
```diff
@@ -226,15 +219,34 @@ public void update(HttpRequest request, HttpResponder responder, @PathParam("nam
...
-    if (creationProperties.getProperties().containsKey(Table.PROPERTY_TTL)) {
-      long ttl = TimeUnit.SECONDS.toMillis(Long.parseLong
-        (creationProperties.getProperties().get(Table.PROPERTY_TTL)));
-      creationProperties.getProperties().put(Table.PROPERTY_TTL, String.valueOf(ttl));
+    fixProperties(creationProperties.getProperties());
+    return creationProperties;
+  }
+
+  private Map<String, String> getProperties(HttpRequest request) {
+    Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()));
```
chtyim: When creating `Reader` from `InputStream`, always provide a `Charset`.

chtyim:
LGTM

PR #2639, opened 2015-05-07
Introduce MyCallbackHelper's doOnce and doWithDelay helper methods, t…

anwar6953:
…o avoid duplicate actions.

Currently, it is applied to the following actions.
- Create Namespace
- Create Dashboard
- Create Widget

Suggestions for other places to apply this are welcome.

### DIFF
--- a/cdap-ui/app/features/dashboard/controllers/dashboard-ctrl.js
+++ b/cdap-ui/app/features/dashboard/controllers/dashboard-ctrl.js
@@ -75,12 +75,12 @@ function ($scope, $state, $dropdown, rDashboardsModel, MY_CONFIG, $alert) {
     $state.go('dashboard.user.addwdgt', {tab: $state.params.tab}, {reload: false});
   };
 
-  $scope.addDashboard = function (title) {
+  $scope.addDashboard = _.throttle(function (title) {
     rDashboardsModel.add({title: title}).then(function() {
       var tabDest = rDashboardsModel.data.length - 1;
       $state.go('dashboard.user', {tab: tabDest}, {reload: true});
     });
-  };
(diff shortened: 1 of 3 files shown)


cdap-ui/app/services/callback-helper.js
```diff
@@ -0,0 +1,34 @@
...
+    this.doWithDelay = function(callback, delay) {
+      var disabled = false;
+      if (typeof delay === 'undefined') {
+        delay = DEFAULT_DELAY;
+      }
+      return function() {
+        if (!disabled) {
+          disabled = true;
+          $timeout(function () { disabled = false; }, delay, false);
+          callback.apply(this, arguments);
```
elfenheart: don't you want this callback inside the $timeout?

cdap-ui/app/features/dashboard/controllers/dashboard-ctrl.js
```diff
@@ -75,12 +75,12 @@ function ($scope, $state, $dropdown, rDashboardsModel, MY_CONFIG, $alert) {
     $state.go('dashboard.user.addwdgt', {tab: $state.params.tab}, {reload: false});
   };
 
-  $scope.addDashboard = function (title) {
+  $scope.addDashboard = MyCallbackHelper.doWithDelay(function (title) {
```
elfenheart: Why does adding dashboard needs a delay?

cdap-ui/app/services/callback-helper.js
anwar6953: Sorry, maybe I should rename this. It's more like 'call the callback if it has been X milliseconds f delay since the last call'.

elfenheart:
:+1:

PR #2642, opened 2015-05-07
Fix the log class name for MasterTwillApplication from MasterServiceMain to MasterTwillApplication.

henry-cask:
Looks like the log class for MasterTwillApplication has it sets to use MasterServiceMain class.


bdmogal:
:+1:

henry-cask:
Thanks @bdmogal!

Merging ...

PR #2645, opened 2015-05-07
Fixes etlapps.less with incorrect git merge. Should highlight the added transforms with different color

ajainarayanan:


elfenheart:
:+1: LGTM

PR #2647, opened 2015-05-08
CDAP-2489 Fixed the invalid RunId comparison and simplified the condi…

sagarkapare:
…tion.
Jira: https://issues.cask.co/browse/CDAP-2489

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
@@ -238,10 +238,8 @@ public final class DistributedProgramRuntimeService extends AbstractProgramRunti
     List<RunRecord> activeRunRecords = store.getRuns(ProgramRunStatus.RUNNING, new Predicate<RunRecord>() {
       @Override
       public boolean apply(RunRecord record) {
-        if (record.getTwillRunId() == null) {
-          return false;
-        }
-        return twillRunIds.contains(record.getTwillRunId());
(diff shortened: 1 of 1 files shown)


sagarkapare:
@henry-cask Please take a look. Thanks!

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT1818-2

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedProgramRuntimeService.java
```diff
@@ -238,10 +238,8 @@ public synchronized RuntimeInfo lookup(Id.Program programId, final RunId runId)
     List<RunRecord> activeRunRecords = store.getRuns(ProgramRunStatus.RUNNING, new Predicate<RunRecord>() {
       @Override
       public boolean apply(RunRecord record) {
-        if (record.getTwillRunId() == null) {
-          return false;
-        }
-        return twillRunIds.contains(record.getTwillRunId());
+        return record.getTwillRunId() != null
+          && twillRunIds.contains(org.apache.twill.internal.RunIds.fromString(record.getTwillRunId()));
```
chtyim: Why contains? I think should be startsWith

sagarkapare: `twillRunIds` is of type `Set<RunId>`. We need to check if the `twillRunId` we get from the run record exists in this set.

chtyim:
LGTM

sagarkapare:
Thank you @chtyim for the review. Build passed. Merging.

PR #2648, opened 2015-05-08
CDAP-2490 Check if invalid Run Records for Spark and MapReduce part of run from Workflow child programs

henry-cask:
In 3.0 we have a scheduler thread to check and update Run Record store for those with RUNNING state but does not have the real program running via check to RuntimeService.

However, for Run Records for Workflow child programs, the runtime service never return runtime info since those runs are run from Workflow driver. Which means that there is no Twill application for the internal child programs run records to check against.

The effect is that the UI will display children program runs for a workflow as FAILED even though it may be COMPLETED or still RUNNING.

This PR contains fix to update the RunRecord checker to account for this situation:
-) Refactor the code to reuse the logic to get the right Program id for run records with RUNNING state but not actually running.
-) Check the RunRecord properties for "workflowrunid" property which contains Workflow parent run id then check the status of that run. If it is in RUNNING status but no actual program still running then it is in invalid state to update children run records status accordingly.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
@@ -239,7 +239,7 @@ public class ProgramLifecycleService extends AbstractIdleService {
     });
 
     if (!invalidRunRecords.isEmpty()) {
-      LOG.debug("Found {} RunRecords with RUNNING status but the program not actually running for type {}",
+      LOG.warn("Found {} RunRecords with RUNNING status but the program not actually running for type {}",
                 invalidRunRecords.size(), programType.getPrettyName());
     }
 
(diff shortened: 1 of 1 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -242,107 +239,153 @@ public boolean apply(@Nullable RunRecord input) {
     });
 
     if (!invalidRunRecords.isEmpty()) {
-      LOG.debug("Found {} RunRecords with RUNNING status but the program not actually running.",
-                invalidRunRecords.size());
+      LOG.debug("Found {} RunRecords with RUNNING status but the program not actually running for type {}",
```
sagarkapare: should it be `warn` by default?

sagarkapare:
Minor comment, otherwise LGTM.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
henry-cask: Sounds good, will update to warn level.

henry-cask:
Build:
http://builds.cask.co/browse/CDAP-DUT1819-2

henry-cask:
Thanks @sagarkapare, build successful. Merging it now ...

PR #2649, opened 2015-05-08
Fixes flow diagrams (label change in workflows caused flows to break as they are using same directive)

ajainarayanan:


elfenheart:
:+1: LGTM

PR #2651, opened 2015-05-08
Fixes uppercase title in stream ingestion + fixed top offset for modal everywhere

ajainarayanan:
- Fixes modal(s) to have a fixed offset from the top of the viewport.
- Fixes uppercase stream name in stream ingestion modal.


elfenheart:
:+1: LGTM

PR #2654, opened 2015-05-12
Change name and add more vboxmanage settings

wolf31o2:
These settings improve performance and properly display a help menu in the first xterm.

PR #2655, opened 2015-05-12
Change name and add more vboxmanage settings (develop)

wolf31o2:
These settings improve performance and properly display a help menu in the first xterm.

Same as #2654 for develop branch


duddun:
LGTM

PR #2657, opened 2015-05-12
Fixes margin top for all modal in cdap

ajainarayanan:
Fixes modal's margin-top to be a constant instead of using translate css property. (Doesn't play well with both angular-strap's modal and angular-bootstrap modal)


elfenheart:
:+1: LGTM

PR #2658, opened 2015-05-12
Add the ability to forcibly kill the Standalone CDAP process.

awholegunch:
Under Windows, the running process is forcibly killed; this adds an option to force that in the equivalent shell script.

### DIFF
--- a/cdap-standalone/bin/cdap.sh
+++ b/cdap-standalone/bin/cdap.sh
@@ -253,12 +253,17 @@ start() {
 }
 
 stop() {
+    if [[ "x${1}" == "xtrue" ]]; then
+      killSignal="-9"
+    else
+      killSignal=""
+    fi
     echo -n "Stopping Standalone CDAP ..."
     if [ -f $pid ]; then
       pidToKill=`cat $pid`
       # kill -0 == see if the PID exists
       if kill -0 $pidToKill > /dev/null 2>&1; then
-        kill $pidToKill > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


wolf31o2:
Pretty sure Windows doesn't have the equivalent signal to `kill` on *NIX and must use the equivalent of `kill -9`... What was the reason for making this change?

awholegunch:
Under Windows (in cdap.bat), we force-kill the process: ``taskkill /F /PID %%i >NUL 2>&1``
(see https://technet.microsoft.com/en-us/library/bb491009.aspx; "/f   : Specifies that process(es) be forcefully terminated.")

We don't have a force under Unix...so I find I am often having to find the process ID, and then use kill -9...this means if the process won't stop, I can use cdap.sh to force kill it.

chtyim:
First, why the process cannot stop normally? We shouldn't allow that to happen.

awholegunch:
@chtyim True, the process should stop normally, but because of bugs or issues, sometimes it doesn't.

Then one is left with using kill -9. This makes it easier in those cases.

sreevatsanraman:
We should not be doing this. If we provide this option, we won't get to the underlying issue. Let's open JIRAs for when process doesn't shutdown cleanly, attach stack trace and required details than masking the problem. 

I suggest we close this PR.

awholegunch:
As there is a consensus to **not** do this, I will close the PR. Opened an issue to check if this should be done under Windows or not...(https://issues.cask.co/browse/CDAP-2555)

PR #2661, opened 2015-05-12
Show available content types for 'load stream' command

jwang47:
http://builds.cask.co/browse/CDAP-DUT1828


cdap-docs/reference-manual/source/cli-api.rst
```diff
@@ -282,7 +284,7 @@ These are the available commands:
    **Explore**
    ``execute <query> [<timeout>]``,"Executes a Query with optional <timeout> in minutes (default is no timeout)."
    **Ingest**
-   ``load stream <stream-id> <local-file-path> [<content-type>]``,"Loads a file to a Stream. The contents of the file will become multiple events in the Stream, based on the content type. If <content-type> is not provided, it will be detected by the file extension."
+   ``load stream <stream-id> <local-file-path> [<content-type>]``,"Loads a file to a Stream. The contents of the file will become multiple events in the Stream, based on the content type. If <content-type> is not provided, it will be detected by the file extension. Supported <content-type>s: avro, csv, log, tsv, txt"
```
awholegunch: I don't think this is correct:
```
$ cdap-cli.sh load stream logEventStream cdap-wise-0.4.0/resources/apache.accesslog "log"
Successfully connected CDAP instance at http://JJs-MacBook-Pro.local:10000
Error: Unsupported content type log
$ cdap-cli.sh load stream logEventStream cdap-wise-0.4.0/resources/apache.accesslog "text/plain"
Successfully connected CDAP instance at http://JJs-MacBook-Pro.local:10000
Successfully sent stream event to stream 'logEventStream'
```
Content types needs to be what the extensions are mapped to, not the extensions themselves.

awholegunch: Either that, or change the code so an extension can be passed in.

cdap-docs/reference-manual/source/cli-api.rst
```diff
@@ -282,7 +284,7 @@ These are the available commands:
    **Explore**
    ``execute <query> [<timeout>]``,"Executes a Query with optional <timeout> in minutes (default is no timeout)."
    **Ingest**
-   ``load stream <stream-id> <local-file-path> [<content-type>]``,"Loads a file to a Stream. The contents of the file will become multiple events in the Stream, based on the content type. If <content-type> is not provided, it will be detected by the file extension."
+   ``load stream <stream-id> <local-file-path> [<content-type>]``,"Loads a file to a Stream. The contents of the file will become multiple events in the Stream, based on the content type (avro/binary, text/csv, text/plain, text/tsv, text/plain). If <content-type> is not provided, it will be detected by the file extension. Supported file extensions: avro, csv, log, tsv, txt."
```
awholegunch: you have text/plain twice...so  you need the unique values of the keys in the map

awholegunch:
LGTM.

PR #2662, opened 2015-05-12
Expose RemoteStreamManager from IntegrationTestManager. It is a capab…

anwar6953:
…ility existing in RemoteApplicationManager#getStreamWriter, which is now deprecated.

http://builds.cask.co/browse/CDAP-DUT1827-3


jwang47:
LGTM

jwang47:
LGTM

PR #2663, opened 2015-05-12
Simplify the JavaDoc for ApplicationTemplate

henry-cask:
Simple PR to reduce use of word "configure" in the JavaDoc for the ApplicationTemplate class.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
@@ -33,9 +33,8 @@ import javax.annotation.Nullable;
 public abstract class ApplicationTemplate<T> implements Application {
 
   /**
-   * Configure an Adapter with the given configuration.
-   * This method is called when an adapter is created in order to define what Datasets, Streams, and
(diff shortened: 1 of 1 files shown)


albertshau:
:100:

cdap-api/src/main/java/co/cask/cdap/api/templates/ApplicationTemplate.java
```diff
@@ -33,7 +33,7 @@
 public abstract class ApplicationTemplate<T> implements Application {
 
   /**
-   * Using the given configuration, configure an Adapter with the given configuration.
+   * Configure an Adapter with the given configuration.
    * This method is called when an adapter is created in order to define what Datasets, Streams, and
    * Plugins, and runtime arguments should be available to the adapter.
    *
```
awholegunch: How about instead:
```
Called when an adapter is created in order to define what Datasets, Streams,
Plugins, and runtime arguments should be available to the adapter, as determined
by the given configuration.
```

(The duplicate "and" in the existing second sentence should be removed...if the first sentence is removed, this is the sentence that will be shown on method summary.)

henry-cask: That will work too.

PR #2664, opened 2015-05-12
Hotfix/temp fix flow change instances

chtyim:
It includes fixes for CDAP-1951 and CDAP-1996.

- CDAP-1951. It just a temp fix to make changing flowlet instances less easy to fail.
- CDAP-1996. It fixes the problem of wrongly deleting queue states when two queues share the same queue name prefix.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/FlowletTwillRunnable.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/FlowletTwillRunnable.java
@@ -16,11 +16,19 @@
 package co.cask.cdap.internal.app.runtime.distributed;
 
 import co.cask.cdap.internal.app.runtime.flow.FlowletProgramRunner;
+import com.google.common.base.Throwables;
+import org.apache.twill.api.Command;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.ExecutionException;
 
 /**
  *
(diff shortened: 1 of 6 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/FlowletTwillRunnable.java
```diff
@@ -30,4 +38,18 @@
...
     return FlowletProgramRunner.class;
   }
+
+  @Override
+  public void handleCommand(Command command) throws Exception {
+    try {
+      super.handleCommand(command);
+    } catch (ExecutionException e) {
+      Throwable rootCause = Throwables.getRootCause(e);
+      if (!"Suspension not allowed".equals(rootCause.getMessage())
```
henry-cask: Just curious why checking for specific error messages?

chtyim:
Ropen as #2666

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/FlowletTwillRunnable.java
chtyim: That's not a very great fix, but a quick fix. The IllegalStateException is thrown from the AbstractProgramController when the controller state cannot transit to suspend/resume state. Unfortunately, there are other failure that also throw IllegalStateException which we don't want to ignore. Any better suggestion?

PR #2672, opened 2015-05-13
move program status into a directive

elfenheart:


anwar6953:
Any idea why Service program type is not applied here?

elfenheart:
Service program type status container is a little different. I didn't want to change it just yet before I confirm

anwar6953:
Ok, my understanding is that it's status should be inferred in the same way as other programs (like flow).

elfenheart:
Right now services have instance container as the status. This will actually have to go through redesign first. We need to be able to increase the instance for the service as well. Will need to go to a different PR

anwar6953:
Ok, so probably - the idea should be that instance container does not contain status, right?
And that the status should be displayed through the same way as other programs.
Maybe a different PR, but that should definitely be done, if agreed upon.

elfenheart:
Yes, it should be consistent with CDAP services, flows, and services program.

cdap-ui/app/directives/program-status/program-status.js
```diff
@@ -0,0 +1,34 @@
...
+
+    return {
+      restrict: 'E',
+      scope: {
+        type: '@',
+        runid: '@'
+      },
+      templateUrl: 'program-status/program-status.html',
+      controller: function($scope, MyDataSource, $state) {
+        if ($scope.runid !== 'No Runs!') {
```
ajainarayanan: Can we use $scope.runs.length property? At some point in time if we want to change the label for No Runs! to something else then this reference needs to be changed.

ajainarayanan:
One Minor comment. Otherwise LGTM :+1:

anwar6953:
@elfenheart https://github.com/caskdata/cdap/pull/2672#issuecomment-101817200
Can you then make it a point to do that in a follow up PR, then?

elfenheart:
@anwar6953 Will be done with services instance

ajainarayanan:
:+1: LGTM

PR #2673, opened 2015-05-13
CDAP-2502 call upgrade on QueueAdmin

rohitsinha54:
Upgrade queues through admin. 

Currently testing it on a cluster with Purchase app.


chtyim:
LGTM. This also need to be back-port to release/3.0 branch as well.

PR #2675, opened 2015-05-13
CDAP-2506 SDK VM requires writable /opt/cdap

wolf31o2:
- Stop CDAP SDK
- Delete data and logs directories
- Set ownership on `/opt/cdap`

### DIFF
--- a/cdap-distributions/src/packer/scripts/sdk-cleanup.sh
+++ b/cdap-distributions/src/packer/scripts/sdk-cleanup.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+#
+# Copyright © 2015 Cask Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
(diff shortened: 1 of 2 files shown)


cdap-distributions/src/packer/scripts/sdk-cleanup.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+
+# Stop SDK
+/etc/init.d/cdap-sdk stop
+
+# Remove data and logs directories
+rm -rf /opt/cdap/sdk/data /opt/cdap/sdk/logs
+
+# Make cdap own /opt/cdap
+chown -R cdap:cdap /opt/cdap
```
bdmogal: should we also ``rm -rf /opt/cdap/conf`` if it exists?

bdmogal:
1 quick question, other than that :+1:

cdap-distributions/src/packer/scripts/sdk-cleanup.sh
wolf31o2: Yeah, let's do that. It's not 100% necessary, but let's just be sure.

PR #2678, opened 2015-05-14
Feature/ui data modeling post3.0

ajainarayanan:
Finishing up what was started here (https://github.com/caskdata/cdap/pull/1498)

The last update from me in PR#1498 (https://github.com/caskdata/cdap/pull/1498#issuecomment-82712166) is still valid except for 1st TODO.

In addition: 
- We need to modify ```BuildUrl``` function in datasource.js to parse params for metrics (tags based url.)
   The reasoning behind this is that the metrics url will have the same query string (tags) and we cannot pass a json map with same key as property multiple times.

```
  { 
    tag: 'apps:PurchaseHistory',
    tag: 'workflows:PurchaseHistoryWorkflow',
    tag: 'metric:Something'
  } 
```
the snippet is not a valid json. We need to find way to pass an array of objects to construct query string.( For further details check https://docs.angularjs.org/api/ngResource/service/$resource USAGE section)

### DIFF
--- a/cdap-ui/app/features/workflows/controllers/tabs/runs/tabs/log-ctrl.js
+++ b/cdap-ui/app/features/workflows/controllers/tabs/runs/tabs/log-ctrl.js
@@ -1,19 +1,20 @@
 angular.module(PKG.name + '.feature.workflows')
-  .controller('WorkFlowsRunDetailLogController', function($scope, MyDataSource, $state) {
-
-    var dataSrc = new MyDataSource($scope),
-        basePath = '/apps/' + $state.params.appId +
-                   '/workflows/' + $state.params.programId +
-                   '/runs/' + $scope.runs.selected.runid;
+  .controller('WorkFlowsRunDetailLogController', function($scope, myWorkFlowApi, $state) {
+    var params = {
+      appId: $state.params.appId,
+      workflowId: $state.params.programId,
+      runId: $scope.runs.selected.runid,
+      scope: $scope
+    };
 
     $scope.logs = [];
     if (!$scope.runs.length) {
       return;
     }
-    dataSrc.poll({
-      _cdapNsPath: basePath + '/logs/next?max=50'
-    }, function(res) {
-      $scope.logs = res;
-    });
-
+    params.max = 50;
+    myWorkFlowApi.logs(params)
+      .$promise
+      .then(function(res) {
+        $scope.logs = res;
+      });
 });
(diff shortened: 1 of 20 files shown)


cdap-ui/app/features/workflows/controllers/tabs/runs/tabs/log-ctrl.js
```diff
@@ -1,19 +1,20 @@
...
     if (!$scope.runs.length) {
       return;
     }
-    dataSrc.poll({
-      _cdapNsPath: basePath + '/logs/next?max=50'
-    }, function(res) {
-      $scope.logs = res;
-    });
-
+    params.max = 50;
```
elfenheart: Can we include this params.max within line 3-8?

ajainarayanan: Fixed.

anwar6953:
In response to what you said about the query tags, can we instead have a field/object that is `tags`, and it simply be a string to string mapping.
For instance:
` { 
    'apps' : 'PurchaseHistory',
    'workflows' : 'PurchaseHistoryWorkflow',
    'metric' : 'Something'
  } `
Then, building the url can take that dictionary, and for each key-value pairing, build a segment of the url like `tag=key:value`.
This avoids having the same string for multiple key-value pairings, right?

cdap-ui/app/services/data/datasource.js
```diff
@@ -199,30 +276,92 @@ angular.module(PKG.name+'.services')
...
+function buildUrl(url, params) {
+  if (!params) return url;
+  var parts = [];
+
+  function forEachSorted(obj, iterator, context) {
+    var keys = Object.keys(params).sort();
+    for (var i = 0; i < keys.length; i++) {
+      iterator.call(context, obj[keys[i]], keys[i]);
+    }
+  return keys;
```
anwar6953: indent is off

ajainarayanan:
@anwar6953 : It is something that is done only in operations. We could have generic method that accepts a js object and constructs the URL for us. Once we have the url with proper query parameters, then we can directly provide the url to ```$resource``` object which will make the request.

The contention is: is it worth abstracting out the operations part because eventually the controller is constructing the url instead of the factory or service (no re-use). Probably we end up operations using raw MyDataSource and everything else using ```$resource```

cdap-ui/app/services/data/datasource.js
ajainarayanan: Fixed.

anwar6953:
Would something like this be useful?
https://github.com/caskdata/cdap/pull/2718/files#diff-f5be262a9ebc0b228914ac38fba4796bR164

elfenheart:
:+1: LGTM

PR #2679, opened 2015-05-14
Feature/r3.0 remove console

awholegunch:
Removes remaining usage of the word "Console" (in the context of "CDAP Console"; there are other instances in different contexts that remain).

Updates the usage in the StandaloneMain to include JDK 1.7.

Fixes https://issues.cask.co/browse/CDAP-2235


bdmogal:
:+1:

awholegunch:
Need to get a build done before merging...

PR #2684, opened 2015-05-14
Fixes runtime args and preferences on program level to be persistent for the duration of the browser session

ajainarayanan:
- Provides the ability to store the preferences and run time arguments for the duration of the browser session.
- Modifies my-start-stop-button directive to use the preferences & runtime arguments services (pass in runtime args and preferences to be displayed in the modal)


elfenheart:
Talked to Ajai that Preferences persist from the backend when saved.

ajainarayanan:
@elfenheart Updated the PR. No longer saving preferences in js memory as it is persisted in backend.

elfenheart:
:+1: LGTM

PR #2687, opened 2015-05-14
CDAP-2515: MetricsClient changes for query and search by tags

jwang47:
http://builds.cask.co/browse/CDAP-DUT1841
https://issues.cask.co/browse/CDAP-2515

### DIFF
--- a/cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
+++ b/cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
@@ -20,16 +20,20 @@ import co.cask.cdap.client.app.FakeApp;
 import co.cask.cdap.client.app.FakeFlow;
 import co.cask.cdap.client.common.ClientTestBase;
 import co.cask.cdap.common.conf.Constants;
-import co.cask.cdap.common.metrics.MetricsContexts;
+import co.cask.cdap.common.metrics.MetricsTags;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.MetricQueryResult;
+import co.cask.cdap.proto.MetricTagValue;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.test.XSlowTests;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -67,11 +71,18 @@ public class MetricsClientTestRun extends ClientTestBase {
       Id.Program programId = Id.Program.from(appId, ProgramType.FLOW, FakeFlow.NAME);
       String flowlet = FakeFlow.FLOWLET_NAME;
 
-      MetricQueryResult result = metricsClient.query(Constants.Metrics.Name.Flow.FLOWLET_INPUT,
-                                                     MetricsContexts.forFlowlet(programId, flowlet));
+      MetricQueryResult result = metricsClient.query(
+        ImmutableList.of(Constants.Metrics.Name.Flow.FLOWLET_INPUT),
+        ImmutableList.<String>of(),
+        MetricsTags.flowlet(programId, flowlet));
       Assert.assertEquals(1, result.getSeries()[0].getData()[0].getValue());
 
-      // TODO: more tests
+      List<MetricTagValue> tags = metricsClient.searchTags(MetricsTags.flowlet(programId, flowlet));
+      Assert.assertEquals(1, tags.size());
+      Assert.assertEquals("run", tags.get(0).getName());
+
+      List<String> metrics = metricsClient.searchMetrics(MetricsTags.flowlet(programId, flowlet));
+      Assert.assertTrue(metrics.contains(Constants.Metrics.Name.Flow.FLOWLET_INPUT));
     } finally {
       programClient.stop(FakeApp.NAME, ProgramType.FLOW, FakeFlow.NAME);
       appClient.delete(FakeApp.NAME);
--- a/cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
+++ b/cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
@@ -21,18 +21,23 @@ import co.cask.cdap.client.config.ClientConfig;
 import co.cask.cdap.client.util.RESTClient;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.exception.UnauthorizedException;
-import co.cask.cdap.common.metrics.MetricsContexts;
+import co.cask.cdap.common.metrics.MetricsTags;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.MetricQueryResult;
+import co.cask.cdap.proto.MetricTagValue;
 import co.cask.common.http.HttpMethod;
 import co.cask.common.http.HttpResponse;
 import co.cask.common.http.ObjectResponse;
 import com.google.common.base.Joiner;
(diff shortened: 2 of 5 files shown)


cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -59,6 +63,62 @@ public MetricsClient(ClientConfig config) {
...
+      queryParts.add("tag=" + tag);
+    }
+
+    URL url = config.resolveURLV3(String.format("metrics/query?%s", Joiner.on("&").join(queryParts)));
+    HttpResponse response = restClient.execute(HttpMethod.POST, url, config.getAccessToken());
+    return ObjectResponse.fromJsonBody(response, MetricQueryResult.class).getResponseObject();
+  }
+
+  /**
+   * @deprecated As of 3.1.0, replaced by {@link #query}
```
gsps1: we will remove the context endpoints in metrics in 3.1,  as we already deprecated them in 3.0, so when we remove that, this query method would also be removed not deprecated.

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -59,6 +63,62 @@ public MetricsClient(ClientConfig config) {
...
+   * @param tags the tags to match
+   * @return the metrics matching the given tags
+   * @throws IOException if a network error occurred
+   * @throws UnauthorizedException if the request is not authorized successfully in the gateway server
+   */
+  public List<MetricTagValue> search(List<String> tags)
+    throws IOException, UnauthorizedException {
+
+    List<String> queryParts = Lists.newArrayList();
+    queryParts.add("target=tag");
```
gsps1: the javadoc says searches for metrics matching the tags, while here the target is tag, should fix the doc and also add searchMetrics method as well.

gsps1:
just couple comments

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
```diff
@@ -690,6 +690,10 @@ public String getId() {
         return id;
       }
 
+      public Flow getFlow() {
```
gsps1: do you need this change ?

cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsTags.java
```diff
@@ -0,0 +1,54 @@
...
+
+import co.cask.cdap.proto.Id;
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+/**
+ * Provides ways to obtain the context for certain metrics.
+ */
+public final class MetricsTags {
```
anwar6953: This seems very similar to MetricsContexts. What's the main difference, and ok to keep both around?

jwang47: We'll be removing MetricsContexts in 3.1 since we're also removing querying/searching metrics by context rather than tags.

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
jwang47: It's no longer being used here, but I used it earlier and don't see a reason to remove.

cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsTags.java
gsps1: maybe we can rename that to be MetricsTags now and use it?

jwang47: Can't, until we refactor RuntimeStats too, which I don't want to do in this PR.

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -59,46 +62,92 @@ public MetricsClient(ClientConfig config) {
...
+    for (String metric : metrics) {
+      queryParts.add("metric=" + metric);
+    }
+    for (String groupBy : groupBys) {
+      queryParts.add("groupBy=" + groupBy);
+    }
+    for (String tag : tags) {
+      queryParts.add("tag=" + tag);
+    }
```
gsps1: we should support time range queries too right ? if we have to add time series, have  options to add interpolate and resolution as well.

cdap-common/src/main/java/co/cask/cdap/common/metrics/MetricsTags.java
gsps1: no i meant, we don't have to remove MetricContexts, the name is misleading, we are just returning the map of tags there and not a context string, so we can just rename that to be MetricTags right? the functions are similar to MetricTags as @anwar6953  mentioned. let me know if i am missing something.

jwang47: Not possible to simply rename, since MetricsTags uses human-friendly metrics names whereas MetricsContexts uses non-human-friendly metrics names.

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -59,78 +64,159 @@ public MetricsClient(ClientConfig config) {
...
+    URL url = config.resolveURLV3(String.format("metrics/search?%s", Joiner.on("&").join(queryParts)));
+    HttpResponse response = restClient.execute(HttpMethod.POST, url, config.getAccessToken());
+    ObjectResponse<List<String>> result = ObjectResponse.fromJsonBody(
+      response, new TypeToken<List<String>>() { }.getType());
+    return result.getResponseObject();
   }
 
-  public MetricQueryResult query(String metric, Map<String, String> context)
+  /**
+   * Gets the value of the given metrics.
```
awholegunch: This and the next method have the same description; it would be good if that first sentence gave enough context to explain why there are two methods with otherwise the same description. I guess it's because you haven't detailed all the parameters in the second one. Maybe refer to the first as wrapper?

cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
```diff
@@ -67,11 +70,32 @@ public void testAll() throws Exception {
       Id.Program programId = Id.Program.from(appId, ProgramType.FLOW, FakeFlow.NAME);
       String flowlet = FakeFlow.FLOWLET_NAME;
 
-      MetricQueryResult result = metricsClient.query(Constants.Metrics.Name.Flow.FLOWLET_INPUT,
-                                                     MetricsContexts.forFlowlet(programId, flowlet));
+      MetricQueryResult result = metricsClient.query(
+        ImmutableList.of(Constants.Metrics.Name.Flow.FLOWLET_INPUT),
+        ImmutableList.<String>of(),
+        ImmutableMap.of(
+          "namespace", programId.getNamespaceId(),
```
anwar6953: Why not use `MetricTags.forFlowlet` here?

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
jwang47: I think this is fine since the only reason it's another method is that Java doesn't support optional arguments, but they're basically the same method conceptually. It's true that ideally we'd have all the parameters detailed in the javadoc, but I'll need @gsps1's help for that.. the HTTP handler code isn't very transparent about what values are valid for what params.

cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
jwang47: Will use it.

anwar6953: Remainder LGTM

anwar6953:
LGTM

PR #2695, opened 2015-05-15
remove RemoteApplicationManager#ProgramId and DefaultApplicationId#Pr…

anwar6953:
Remove RemoteApplicationManager#ProgramId and DefaultApplicationId#ProgramId.
Move common method to AbstractApplicationManager.java

http://builds.cask.co/browse/CDAP-DUT1846-2

### DIFF
--- a/cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
+++ b/cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
@@ -27,8 +27,8 @@ import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.ProgramRecord;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.RunRecord;
+import co.cask.cdap.test.AbstractApplicationManager;
 import co.cask.cdap.test.AbstractWorkerManager;
-import co.cask.cdap.test.ApplicationManager;
 import co.cask.cdap.test.DataSetManager;
 import co.cask.cdap.test.FlowManager;
 import co.cask.cdap.test.MapReduceManager;
@@ -39,7 +39,6 @@ import co.cask.cdap.test.StreamWriter;
 import co.cask.cdap.test.WorkerManager;
 import co.cask.cdap.test.WorkflowManager;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Stopwatch;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 
@@ -51,7 +50,7 @@ import java.util.concurrent.TimeoutException;
 /**
  *
  */
-public class RemoteApplicationManager implements ApplicationManager {
+public class RemoteApplicationManager extends AbstractApplicationManager {
 
   protected final Id.Application application;
 
@@ -86,7 +85,7 @@ public class RemoteApplicationManager implements ApplicationManager {
 
   @Override
   public FlowManager startFlow(final String flowName, Map<String, String> arguments) {
-    final ProgramId flowId = startProgram(flowName, arguments, ProgramType.FLOW);
+    final Id.Program flowId = startProgram(flowName, arguments, ProgramType.FLOW);
     return new FlowManager() {
       @Override
       public void setFlowletInstances(String flowletName, int instances) {
@@ -101,7 +100,7 @@ public class RemoteApplicationManager implements ApplicationManager {
       @Override
       public RuntimeMetrics getFlowletMetrics(String flowletId) {
         return metricsClient.getFlowletMetrics(
-          Id.Program.from(application, ProgramType.FLOW, flowId.getProgramId()), flowletId);
+          Id.Program.from(application, ProgramType.FLOW, flowId.getId()), flowletId);
       }
 
       @Override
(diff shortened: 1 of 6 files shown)


cdap-test/src/main/java/co/cask/cdap/test/AbstractApplicationManager.java
```diff
@@ -0,0 +1,55 @@
...
+                                      Id.Program programId) throws InterruptedException, TimeoutException {
+    programWaitForState(timeout, timeoutUnit, false, programId);
+  }
+
+  protected void programWaitForRunning(long timeout, TimeUnit timeoutUnit,
+                                       Id.Program programId) throws InterruptedException, TimeoutException {
+    programWaitForState(timeout, timeoutUnit, true, programId);
+  }
+
+  protected void programWaitForState(long timeout, TimeUnit timeoutUnit, boolean waitForRunning,
```
bdmogal: Is it better to use the ``ProgramRunState`` interface or something similar instead of a ``boolean`` here? I think it may make for a clearer interface, plus you can use it to test for more states than just running and finish.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
```diff
@@ -128,13 +127,12 @@ public MapReduceManager startMapReduce(final String jobName) {
 
   @Override
   public MapReduceManager startMapReduce(final String jobName, Map<String, String> arguments) {
-    return getMapReduceManager(jobName, arguments, ProgramType.MAPREDUCE);
+    final Id.Program jobId = startProgram(jobName, arguments, ProgramType.MAPREDUCE);
```
bdmogal: rename to programId, since there is a concerted effort to not use job anywhere in CDAP.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
```diff
@@ -128,13 +127,12 @@ public MapReduceManager startMapReduce(final String jobName) {
...
+    return getMapReduceManager(jobId);
   }
 
-  private MapReduceManager getMapReduceManager(final String jobName, Map<String, String> arguments,
-                                               final ProgramType programType) {
+  private MapReduceManager getMapReduceManager(final Id.Program jobId) {
```
bdmogal: same here, programId instead of jobId.

cdap-test/src/main/java/co/cask/cdap/test/AbstractApplicationManager.java
anwar6953: You mean `ProgramRunStatus`, right?
That seems to describe the status of a particular run, not of the program. So it doesn't have states that a program might have such as `STOPPED`

cdap-test/src/main/java/co/cask/cdap/test/AbstractApplicationManager.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Instance of this class is for managing deployed application.
+ */
+public abstract class AbstractApplicationManager implements ApplicationManager {
+
+  protected abstract boolean isRunning(Id.Program programId);
+
+  protected void programWaitForFinish(long timeout, TimeUnit timeoutUnit,
```
bdmogal: should we just call these methods waitFor* instead of programWaitFor*? 
or maybe waitForProgram*?

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultApplicationManager.java
```diff
@@ -212,8 +188,8 @@ public void waitForFinish(long timeout, TimeUnit timeoutUnit) throws TimeoutExce
     }
   }
 
-  private ProgramId startProgram(String jobName, Map<String, String> arguments, ProgramType programType) {
-    final ProgramId jobId = new ProgramId(applicationId.getId(), jobName, programType);
+  private Id.Program startProgram(String jobName, Map<String, String> arguments, ProgramType programType) {
```
bdmogal: s/jobName/programName/

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultFlowManager.java
```diff
@@ -28,20 +29,20 @@
 public class DefaultFlowManager implements FlowManager {
 
   private final AppFabricClient appFabricClient;
-  private final DefaultApplicationManager.ProgramId flowID;
+  private final Id.Program flowID;
```
bdmogal: nit: lower case D to be consistent with the constructor argument.

bdmogal:
Only minor comments/nits.

cdap-test/src/main/java/co/cask/cdap/test/AbstractApplicationManager.java
bdmogal: May be ``ProgramController.State`` then.

anwar6953: That doesn't work either, because some status returned by the REST API aren't mapped to the ProgramController.State (such as STOPPED).
I think for now, just having two possibilities (running vs not) in test case's waitFor methods is fine.

cdap-test/src/main/java/co/cask/cdap/test/AbstractApplicationManager.java
anwar6953: the program prefix is just to avoid conflicting with the method names on the program manager vs the method names on the application managers. I tried removing it, and even though the signature is different, there's conflicts.

bdmogal:
:+1:

jwang47:
LGTM

PR #2696, opened 2015-05-15
Update Release Notes for version 2.6.3.

awholegunch:
Update conf.py with latest information.

Staged at: http://docs-staging.cask.co/cdap/2.6.3-r2.6.3_release_notes/en/index.html

Page of interest: [Release Notes, 2.6.3](http://docs-staging.cask.co/cdap/2.6.3-r2.6.3_release_notes/en/reference-manual/release-notes.html#release-2-6-3)


cdap-docs/reference-manual/source/release-notes.rst
```diff
@@ -23,6 +23,66 @@ Cask Data Application Platform Release Notes
...
+--------------
+
+- Replace use of CDAP transaction snapshot codecs with the Tephra versions instead
+  (`CDAP-2496 <https://issues.cask.co/browse/CDAP-2496>`__)
+  
+  
+Upgrade Instructions
+--------------------
+
+- **Update Memory Setting for Transaction Service before Upgrading**
```
sreevatsanraman: We don't need this section.

cdap-docs/reference-manual/source/release-notes.rst
```diff
@@ -23,6 +23,66 @@ Cask Data Application Platform Release Notes
...
+  As part of the upgrade, we recommend that the memory configuration for transaction
+  service be increased from its default value of 512MB. We suggest adjusting it in
+  ``cdap-site.xml`` to 4096mb::
+
+    <property>
+    <name>data.tx.memory.mb</name>
+    <value>4096</value>
+    </property>
+
+- **Upgrading from 2.6.2 to 2.6.3**
```
sreevatsanraman: This should just be our standard upgrade instructions. No need to have special one. I suggest we remove this altogether

cdap-docs/reference-manual/source/release-notes.rst
ghelmling: This section was recommended in the case of one very large transaction snapshot file.  Not sure if this should be a general recommendation for us.  This is only necessary if the transaction snapshot that needs to be loaded is very substantial in size (100's of MB).

awholegunch:
Correct me if I am wrong: but looking over the upgrade instructions: why would they upgrade to 2.6.3? Why wouldn't they upgrade to 3.0.0? What makes them use 2.6.3?

ghelmling:
2.6.3 is a bug fix release for those who have stabilized on the 2.6.x release line.  Upgrading to 3.0 involves backward incompatible changes, major new features, no binary compatibility.  It's a whole different ball game, and involves a completely different amount of effort around updating applications, retesting and requalifying as a release.  I don't think that's reasonable to require for a simple bug fix.

awholegunch:
I apologize Gary: I wasn't clear in my question. It's not about why to or not to upgrade to 3.0 — that I understand. I don't understand our instructions: what in them does the upgrade from 2.6.2 to 2.6.3? They are no different than the instructions for 2.6.0 to 2.8.0, or 2.8.0 to 3.0.0..I was wondering if these were run, why would you not end up with 3.0.0? How do you specify the version you are upgrading to? 

This came about as I was comparing the instructions with some changes we have made in the installation instructions, where we advise people to "upgrade packages". We do so in these update instructions. But we don't in the current update instructions. This should be resolved.

awholegunch:
"they" in my original comment meant the instructions, not the user...

ghelmling:
Oh, sorry I misunderstood.  I believe the minor version release is embedded in the YUM/APT repo URL.  So, for 2.6 it contains .../2.6, for 2.8 it contains .../2.8.  So you never automatically upgrade to a new major or minor line without explicitly changing your repo settings.

cdap-docs/reference-manual/source/release-notes.rst
```diff
@@ -23,6 +23,55 @@ Cask Data Application Platform Release Notes
...
+=============================================================
+
+CDAP Bug Fixes
+--------------
+
+- Replace use of CDAP transaction snapshot codecs with the Tephra versions instead
+  (`CDAP-2496 <https://issues.cask.co/browse/CDAP-2496>`__)
+  
+  
+Upgrade Instructions
```
sreevatsanraman: We don't need to specify any additional information here. Installation and Configuration section should already cover it.

sreevatsanraman: Remove the entire Upgrade Instructions.

duddun:
That is correct (what ghelmling said).  Once the repo files are in place, unless/until they get ones for new major/minor revision, their updates will be to the same major/minor (e.g. 2.6.x for 2.6) 

ubuntu repo file (cask.list):
deb [ arch=amd64 ] http://repository.cask.co/ubuntu/precise/amd64/cdap/2.6 precise cdap

centos repo file (cask.repo):
[cask]
name=Cask Packages
baseurl=http://repository.cask.co/centos/6/x86_64/cdap/2.6
enabled=1
gpgcheck=1

awholegunch:
I've replaced the instructions with a link. Opening an issue regarding the 3.0 doc set.

sreevatsanraman:
LGTM :+1:

PR #2701, opened 2015-05-15
update angular and fix breakage in services within admin

elfenheart:
Update Angular and angular components to 1.3.15 (1.4 is not a stable release yet). There are some performance improvement especially in regards to memory leakage.

You need to remove bower_components and clean up bower cache (**bower cache clean**). Then do **bower install**


ajainarayanan:
:+1: LGTM

PR #2702, opened 2015-05-15
Releases 3.0.0 and 2.6.3.

awholegunch:
Adds 2.6.3 as a version in the list, and adds the pages to docs.cdap.io.
Updates the 3.0.0 version to the latest.


wolf31o2:
This is pretty much impossible to code review. I see you added `2.6.3` to the README.rst but everything else... in the future, may be easier to separate bulk imports of other data from actual changes being made, since bulk imports don't need review. Reluctant :+1:

awholegunch:
Well, this is typically how it happens. You bulk import the new pages, and at the same time you update the index page to provide a link to them and update the information. I don't see that having two PRs will improve that; but I can highlight the page to look at in the comments so that you know which page really needs to be reviewed. It's the first page: README.rst. I think in all cases it will be the first..

PR #2705, opened 2015-05-16
Refactor program managers.

anwar6953:
(Refactor and simplify the program manager classes, without removing any APIs.)

Mostly:
Refactor program managers to extend a new `AbstractProgramManager` for code de-duplication and simplification. This also makes it much easier to add APIs to the program manager classes.

Minor:
Extract all program manager classes out as concrete classes (as opposed to anonymous classes inside the application managers).
`FlowManager` was missing a `getFlowletInstances` method.

http://builds.cask.co/browse/CDAP-DUT1846-6


anwar6953:
The main idea is to the manager classes, so that API changes can happen more easily in the future.
I'm trying not to mix in API changes (removals) in this PR.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
```diff
@@ -127,25 +77,7 @@ public MapReduceManager startMapReduce(final String programName) {
 
   @Override
   public MapReduceManager startMapReduce(final String programName, Map<String, String> arguments) {
```
bdmogal: How come we don't do ``startProgram()`` like the other methods in this class?

bdmogal: Oh oops, just saw that the call is embedded inside the ``DefaultMapReduceManager`` constructor. Maybe pull it out for consistency?

anwar6953: We do, I guess it's the result isn't stored in a temp variable; that's the only difference.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
```diff
@@ -254,48 +116,14 @@ public ServiceManager startService(String serviceName) {
...
   public ServiceManager startService(final String serviceName, Map<String, String> arguments) {
-    startProgram(serviceName, arguments, ProgramType.SERVICE);
-    return new RemoteServiceManager(Id.Service.from(application, serviceName), clientConfig);
+    Id.Program programId = startProgram(serviceName, arguments, ProgramType.SERVICE);
+    return new RemoteServiceManager(programId, clientConfig, this);
   }
 
   @Override
   public WorkerManager startWorker(final String workerName, Map<String, String> arguments) {
     final Id.Program workerId = Id.Program.from(application, ProgramType.WORKER, workerName);
```
bdmogal: This should have a ``startProgram`` too, no?

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteWorkflowManager.java
```diff
@@ -0,0 +1,98 @@
...
+      public void resume() {
+        try {
+          scheduleClient.resume(programId.getApplicationId(), schedName);
+        } catch (Exception e) {
+          throw Throwables.propagate(e);
+        }
+      }
+
+      @Override
+      public String status(int expectedCode) {
```
bdmogal: perhaps ``getStatus``?

cdap-test/src/main/java/co/cask/cdap/test/AbstractProgramManager.java
```diff
@@ -16,12 +16,38 @@
...
  */
-public abstract class AbstractServiceManager implements ServiceManager {
+public abstract class AbstractProgramManager implements ProgramManager {
+  protected final Id.Program programId;
+  private final ApplicationManager applicationManager;
+
+  public AbstractProgramManager(Id.Program programId, ApplicationManager applicationManager) {
+    this.applicationManager = applicationManager;
+    this.programId = programId;
+  }
```
bdmogal: missing newline

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
```diff
@@ -193,58 +105,8 @@ public void waitForFinish(long timeout, TimeUnit timeoutUnit) throws TimeoutExce
...
-            } catch (Exception e) {
-              throw Throwables.propagate(e);
-            }
-          }
-        };
-      }
-
-    };
+    Id.Program workflowId = Id.Program.from(application, ProgramType.WORKFLOW, workflowName);
+    return new RemoteWorkflowManager(workflowId, clientConfig, this);
```
jwang47: startProgram here too

anwar6953: In this PR, I'm going to leave this behavior as is, because there is code that depends on it working this way.
https://github.com/caskdata/cdap/pull/2705/files#diff-0147c2712afdf71eec4b4fc988a99db7R107

I think we can remove it if we introduce a `getManager` method, without starting the program, but that will be a whole new set of APIs, and I don't want to mix that in here.

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultApplicationManager.java
```diff
@@ -211,40 +184,7 @@ public void waitForFinish(long timeout, TimeUnit timeoutUnit) throws TimeoutExce
   public WorkflowManager startWorkflow(final String workflowName, Map<String, String> arguments) {
     final Id.Program workflowId = Id.Program.from(applicationId, ProgramType.WORKFLOW, workflowName);
```
bdmogal: ``startProgram``? Or perhaps not, because workflows are started via schedules? Not completely sure.

anwar6953: https://github.com/caskdata/cdap/pull/2705#discussion_r30538711

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
bdmogal: Yes, saw that. nit: pull it out only for consistency.

bdmogal:
The changes look good. Only a few comments.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
jwang47: Sounds good.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteWorkflowManager.java
anwar6953: Hmm, this is a part of `WorkflowManager` API, and I don't think a small change like that warrants an API change right now. What do you think?

anwar6953:
Addressed comments. @bdmogal @jwang47 
Please take another look.
Thanks.

cdap-test/src/main/java/co/cask/cdap/test/AbstractProgramManager.java
anwar6953: added newline on line 35.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteApplicationManager.java
anwar6953: pulled out.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteWorkflowManager.java
bdmogal: Perhaps open a jira to change it. Good for now.

bdmogal:
:+1:

PR #2707, opened 2015-05-16
Fix/ui widget organization

ajainarayanan:
- Re-organizes code to separate out service layer and utility layer and controller
- Reduces size of Widget factory (Should have only additive things, say managing metrics, add or remove them, like a settings for the underlying chart)
- Creates a ChartsHelper factory to group all utility function required for constructing query and post-processing data received from backend.
- The aim of the above two changes is to do post-processing of data in one place (rather than at Widget Factory and C3WidgetController and in Charts directive) and try to reduce #of watchers used by dashboard.
- Temporarily Modify C3 library to avoid listening on Window.resize (which gets called like a 256 times on each resize :smile: ) There is an associated PR already open (https://github.com/masayuki0812/c3/pull/860). Once this is merged we can remove this and directly fetch from bower.


cdap-ui/app/directives/c3/charts.js
```diff
@@ -16,36 +17,40 @@ ngC3.factory('c3', function ($window) {
...
-ngC3.controller('c3Controller', function ($scope, c3, myHelpers, $filter, $timeout) {
+ngC3.controller('c3Controller', function ($scope, c3, myHelpers, $filter, $timeout, MyChartHelpers, MyDataSource) {
   // We need to bind because the init function is called directly from the directives below
   // and so the function arguments would not otherwise be available to the initC3 and render functions.
-  var c3 = c3;
-  var myHelpers = myHelpers;
-  var $filter = $filter;
+  var c3 = c3,
+      myHelpers = myHelpers,
+      $filter = $filter,
```
elfenheart: I think we can use Angular injector instead of declaring it like this?

anwar6953:
Can you open a JIRA, so we don't lose track/forget of replacing it with the bower version?

ajainarayanan:
@anwar6953 https://issues.cask.co/browse/CDAP-2544

cdap-ui/app/features/dashboard/controllers/addwdgt-ctrl.js
```diff
@@ -65,7 +65,9 @@ function ($scope, $modalInstance, caskFocusManager, Widget) {
           new Widget({
             type: $scope.model.type,
             title: $scope.model.metric.title,
-            color: classes[(generateHash(value) * 13) % classes.length],
+            settings: {
+              color: classes[(generateHash(value) * 13) % classes.length]
```
elfenheart: Are we still using colors at all?

cdap-ui/app/features/dashboard/controllers/addwdgt-ctrl.js
```diff
@@ -76,7 +78,8 @@ function ($scope, $modalInstance, caskFocusManager, Widget) {
       });
       $scope.currentBoard.addWidget(widgets);
     } else {
-      $scope.model.color = classes[(generateHash($scope.model.metric.name) * 13) % classes.length];
+      $scope.model.settings = {};
+      $scope.model.settings.color = classes[(generateHash($scope.model.metric.name) * 13) % classes.length];
```
elfenheart: Same comment as before

cdap-ui/app/features/dashboard/service/chart-helpers.js
```diff
@@ -0,0 +1,267 @@
...
+      skipAmtFromResolution: skipAmtFromResolution,
+      zeroFill: zeroFill,
+      c3ifyData: c3ifyData,
+      convertDashboardToNewWidgets: convertDashboardToNewWidgets
+    }
+
+
+  })
+
+  .factory('MyMetricsQueryHelper', function() {
```
elfenheart: Let's separate this into a different file?

cdap-ui/app/features/operations/ops-ctrl.js
```diff
@@ -169,6 +175,7 @@ angular.module(PKG.name+'.feature.dashboard')
         angular.extend(widget, panel[4]);
         widgets.push(widget);
       });
+      console.log(widgets);
```
elfenheart: remove?

cdap-ui/app/features/dashboard/controllers/addwdgt-ctrl.js
ajainarayanan: As discussed this is being used in overview-ctrl and we can have this option open if the user wants to have a custom color for the chart.

ajainarayanan:
@elfenheart Addressed other review comments. Please review when you have some time.

cdap-ui/app/features/dashboard/controllers/addwdgt-ctrl.js
anwar6953: Hmm, actually, this can probably be removed.
Currently, this is how overview page's chart's colors are being set:
https://github.com/caskdata/cdap/blob/develop/cdap-ui/app/features/overview/overview.html#L84

ajainarayanan: Done. Removed.

elfenheart:
:+1: LGTM

PR #2686, opened 2015-05-14
Check if the Plugin config params are null or empty

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-RT35-1

PR #2708, opened 2015-05-18
Check if the config param is null or empty

gokulavasan:
Same PR as https://github.com/caskdata/cdap/pull/2686 but against develop instead of release/3.0.

Build : http://builds.cask.co/browse/CDAP-DT93-1


bdmogal:
:+1:

PR #2709, opened 2015-05-18
Correctly determine programatic source for hadoop conf in hadoop 2.3 …

bdmogal:
…and before because of a typo

[Hadoop 2.3 version with typo](https://github.com/apache/hadoop/blob/branch-2.3.0/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java#L967)

[Trunk version with typo fixed](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java#L1206)

Build: http://builds.cask.co/browse/CDAP-DUT1858

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
@@ -97,6 +97,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.jar.JarEntry;
 import java.util.jar.JarOutputStream;
+import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


anwar6953:
lgtm

henry-cask:
+1

LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -121,6 +121,8 @@
 
   // Name of configuration source if it is set programmatically. This constant is not defined in Hadoop
   private static final String PROGRAMMATIC_SOURCE = "programmatically";
+  // Hadoop 2.3.0 and before has a typo, but we need to work with it as well.
+  private static final String PROGRAMATIC_SOURCE = "programatically";
```
chtyim: Maybe a regex `Pattern pattern = Pattern.compile("program{1,2}atically");` ?

bdmogal: Fixed to use above regex.

chtyim:
LGTM

bdmogal:
Build passed: http://builds.cask.co/browse/CDAP-DUT1858-2

Merging.

PR #2710, opened 2015-05-18
On application page, show RUNNING programs with green icon; SUSPENDED…

anwar6953:
… programs with yellow; all else having grey icon.

image


ajainarayanan:
:+1: LGTM

PR #2712, opened 2015-05-18
(CDAP-2495) Make master service HA

chtyim:
- Rewrite/refactor the MasterServiceMain to be more structural
  - Start/stop app-fabric and dataset sevices when leadership changed
- Introduce a wrapping Service class that can survives through startup failure by performaing retry
- Cleanup unnecessary stacktrace logs


henry-cask:
Should this go to 3.0 branch? 
This seems like new feature or enhancement that could just go to 3.1 (i.e: develop branch) release instead.

chtyim:
There is no feature but a fix that makes master works in HA correctly (CDAP-2495) and it was intended for 3.0.1

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/AppFabricServer.java
```diff
@@ -200,20 +190,19 @@ public void failed(State from, Throwable failure) {
         LOG.info("AppFabric HTTP service stopped with failure.", failure);
         for (Cancellable cancellable : cancellables) {
           if (cancellable != null) {
-            cancellable.cancel();
+              cancellable.cancel();
```
bdmogal: extra indentation?

cdap-common/src/main/java/co/cask/cdap/common/service/RetryOnStartFailureService.java
```diff
@@ -0,0 +1,119 @@
...
+import com.google.common.util.concurrent.Service;
+import com.google.common.util.concurrent.Uninterruptibles;
+import org.apache.twill.common.Threads;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ *
```
bdmogal: should we add some Javadoc?

chtyim: Oh.. I thought I added. Sure.

bdmogal:
@chtyim I see that retry ``Thread``/``Runnable``s were converted to guava services in this PR. For my understanding - why is this better?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/AppFabricServer.java
chtyim: Removed.

chtyim:
The existing retry implementing is non-stoppable. Instead of fixing the Runnable implementation, it's simpler to use the Service implementation that can handle stop correctly.

cdap-common/src/main/java/co/cask/cdap/common/service/RetryOnStartFailureService.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+public class RetryOnStartFailureService extends AbstractService {
+
+  private static final Logger LOG = LoggerFactory.getLogger(RetryOnStartFailureService.class);
+
+  private final Thread startupThread;
+  private final String delegateServiceName;
+  private volatile Service currentDelegate;
+
+  public RetryOnStartFailureService(final Supplier<Service> delegate, final RetryStrategy retryStrategy) {
```
bdmogal: Does it make sense to add a default retry strategy so the retryStrategy here becomes nullable?

chtyim: Not sure if we need that now? Also, preferably not nullable but just a overloaded constructor.

bdmogal:
Yup, agreed, we have issues like https://issues.cask.co/browse/CDAP-1933 because of the current implementation, although not sure that will be completely fixed by this.

cdap-common/src/main/java/co/cask/cdap/common/service/RetryOnStartFailureService.java
bdmogal: Sure...

cdap-common/src/main/java/co/cask/cdap/common/service/RetryStrategies.java
```diff
@@ -0,0 +1,116 @@
...
+public final class RetryStrategies {
+
+  /**
+   * @return A {@link RetryStrategy} that doesn't do any retry.
+   */
+  public static RetryStrategy noRetry() {
+    return new RetryStrategy() {
+      @Override
+      public long nextRetry(int failureCount, long startTime) {
+        return -1;
```
bdmogal: nit: ``-1L``?

cdap-common/src/main/java/co/cask/cdap/common/service/RetryStrategy.java
```diff
@@ -17,25 +17,17 @@
...
-  public void stop(final String reason);
-
-  /**
-   * Returns status about whether the thread was stopped.
-   *
-   * @return true if stopped; false otherwise.
-   */
-  public boolean isStopped();
-
+  long nextRetry(int failures, long startTime);
```
bdmogal: nit: should we call this ``nextTry``? since nextTry already implies retry?

chtyim: Retry strategy is for retry and this method only get called if there is failure (hence need to retry).

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
```diff
@@ -353,8 +248,8 @@ private InetAddress getLocalHost() {
     return configKeys;
   }
 
-  private Map<String, Integer> getSystemServiceInstances() {
-    Map<String, Integer> instanceCountMap = new HashMap<String, Integer>();
+  private Map<String, Integer> getSystemServiceInstances(CConfiguration cConf) {
+    Map<String, Integer> instanceCountMap = new HashMap<>();
     for (Map.Entry<String, Map<String, String>> entry : getConfigKeys().entrySet()) {
```
henry-cask: Should this read from cConf instead then?

chtyim: The existing logic is to read from `ServiceStore` first and uses the value in cConf as default. The logic didn't change.

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
```diff
@@ -381,121 +276,359 @@ private InetAddress getLocalHost() {
...
-        public void terminated(Service.State from) {
-          LOG.warn("{} was terminated; restarting with back-off", serviceName);
-          backOffRun();
+  /**
+   * Returns the {@link TwillController} for the current master service or {@code null} if none is running.
+   */
+  @Nullable
+  private TwillController getCurrentTwillController() {
+    int count = 100;
+    long sleepMs = LOOKUP_ATTEMPT_TIMEOUT_MS / count;
```
henry-cask: Since this value is actually static final maybe could just move to become static like LOOKUP_ATTEMPT_TIMEOUT_MS ?

henry-cask:
Did a pass and add comments and questions.

Once resolved LGTM.

Good looking patch +1

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
chtyim: you mean which value? count?

cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/ServiceStoreModules.java
```diff
@@ -19,69 +19,65 @@
...
-  public Module getDistributedModule() {
-    return new PrivateModule() {
-      @Override
-      protected void configure() {
-        install(new FactoryModuleBuilder()
-                  .implement(DatasetDefinitionRegistry.class, DefaultDatasetDefinitionRegistry.class)
-                  .build(DatasetDefinitionRegistryFactory.class));
+  private static final class ServiceStoreModule extends PrivateModule {
+
+    private final DatasetModule tableModule;
```
bdmogal: should we call this ``kvTableModule``?

chtyim: Why? It is creating bindings for the ServiceStore, not any KVTable.

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
```diff
@@ -109,204 +115,82 @@
...
-    }
-
-    // This check will throw an exception if Hive is not present or if it's distribution is unsupported
-    ExploreServiceUtils.checkHiveSupport();
+    login();
   }
 
   @Override
   public void start() {
+    // this should probably be done in NamespaceService#init()
```
bdmogal: This comment is not valid, could you remove it? :)

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
```diff
@@ -381,121 +276,359 @@ private InetAddress getLocalHost() {
...
+  /**
+   * Creates HBase namespace for the cdap system namespace.
+   */
+  private void createSystemHBaseNamespace() {
+    HBaseTableUtil tableUtil = baseInjector.getInstance(HBaseTableUtil.class);
+    try {
+      HBaseAdmin admin = new HBaseAdmin(hConf);
+      tableUtil.createNamespaceIfNotExists(admin, Constants.SYSTEM_NAMESPACE_ID);
+    } catch (IOException e) {
+      Throwables.propagate(e);
```
bdmogal: probably not your PR, but we should ``throw`` this?

cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/ServiceStoreModules.java
bdmogal: I only thought so, because we only use modules of ``KVTable`` in ``ServiceStore``, right?

bdmogal:
Only minor comments. LGTM :+1:

cdap-common/src/main/java/co/cask/cdap/common/service/RetryStrategies.java
chtyim: ok

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
chtyim: This was your comment right? Sure can remove it.

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
chtyim: Sure.

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
bdmogal: Yes :) When I had added that, there was a thought of creating a ``NamespaceService``. But I don't think we need that anymore.

chtyim:
Build passed. http://builds.cask.co/browse/CDAP-RBT292-6

bdmogal:
:+1:

chtyim:
Squashed commits and merging.

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
henry-cask: Oopps forgot to reply. I meant the sleepMs because it is always constant.

PR #2713, opened 2015-05-19
Create new StructuredRecord.Builder instance per Tweet conversions in TwitterSource

henry-cask:
Currently the TwitterSource create one instance of StructuredRecord.Builder and reuse it
to convert Tweets to StructuredRecord instances.

This will cause problems because StructuredRecord.Builder#build reuse the same fields
member variable to create new StructuredRecord.

All converted Tweets to StructuredRecords will have the same values when being written to Emitter.


henry-cask:
Cherry pick for 3.0 branch.

Please see review at PR#2711

henry-cask:
CC @gokulavasan  for FYI

PR #2714, opened 2015-05-19
CDAP-1483 QueueAdmin to use Id.Namespace instead of String

anwar6953:
https://issues.cask.co/browse/CDAP-1483
http://builds.cask.co/browse/CDAP-DUT1859-6


chtyim:
LGTM

anwar6953:
Pushed a removal of 2 lines of test code; it was failling.
We don't use QueueAdmin for LevelDB-based streams, so we don't need to be testing that.

bdmogal:
:+1:

PR #2715, opened 2015-05-19
[CDAP-2534] number of flowlet instances does not update automatically in UI

elfenheart:
Polling for flowlet instances for flow graph


ajainarayanan:
:+1: LGTM

PR #2717, opened 2015-05-19
CDAP-2173 increase metric-picker limit

anwar6953:
100 is a larger limit, while still being performant.

Helps with: https://issues.cask.co/browse/CDAP-2173


ajainarayanan:
:+1: LGTM It would be nicer if we have a better solution ( where when we don't specify a data-limit and everything loads). We need to check with angular-strap library (If possible make a PR :wink: )

anwar6953:
I think the concern with loading everything is that autocomplete with too many options might be laggy, then. I think that's why they have a low default limit.

PR #2722, opened 2015-05-19
CDAP-2547 Kakfa Source - Doesn't use the persisted offset when the Adapter is restarted.

henry-cask:
In Adapter realtime source plugin flow the SourceState is used as the source of offset for the topics and only passed when poll is called.
The initialization of Kafka Consumer infos for topics need to be suspended until first call to poll is called.

Updated the unit test to relfect this.


henry-cask:
Build:
http://builds.cask.co/browse/CDAP-DT95-1

gokulavasan:
LGTM :+1: I guess we need the same PR against release/3.0 ?

henry-cask:
Yes, will need to cherry pick to 3.0

PR #2723, opened 2015-05-19
Fixes node proxy to be startable from command line (without cConf and sConf

ajainarayanan:


elfenheart:
:+1: LGTM.

Pulled to local branch and verified

PR #2724, opened 2015-05-20
CDAP-1912 Updated examples to use Writable instead of JSON as mapper output.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-1912

Currently the `map` function in the examples converts the object to `JSON` before outputting it. Made the objects `Writable`, so that `map` can directly emit those objects.


sagarkapare:
Build is running here : http://builds.cask.co/browse/CDAP-DUT1865-1

sagarkapare:
@anew Please help review. Thanks!

cdap-examples/Purchase/src/main/java/co/cask/cdap/examples/purchase/PurchaseHistoryBuilder.java
```diff
@@ -134,8 +134,9 @@ public void reduce(Text customer, Iterable<Text> values, Context context)
 
       PurchaseHistory purchases = new PurchaseHistory(customer.toString(), userProfile);
       int numPurchases = 0;
-      for (Text val : values) {
-        purchases.add(new Gson().fromJson(val.toString(), Purchase.class));
+      for (Purchase val : values) {
+        purchases.add(new Purchase(val.getCustomer(), val.getProduct(), val.getQuantity(), val.getPrice(),
```
anew: Is this needed? Or can we just use ``val`` directly? Or write ``purchases`` directly to the context? Is it necessary to do defensive copy here? If so, can you add a constructor Purchase(Purchase other)? I think that is nicer.

anew:
Only one comment, otherwise LGTM.

cdap-examples/Purchase/src/main/java/co/cask/cdap/examples/purchase/PurchaseHistoryBuilder.java
sagarkapare: The reducer does not store all values in memory. While looping through the Iterable value list, each Object instance is re-used, so it only keeps one instance around at a given time. Because of this we need to make the copy of the object - http://stackoverflow.com/questions/11024638/hadoop-reducer-values-in-memory

sagarkapare: Updated the code to use `Purchase(Purchase other)` constructor.

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT1865-4 @anew Can you please take a look again? Thanks!

anew:
LGTM

PR #2726, opened 2015-05-20
Cleanup all tests that needs to start Standalone.

chtyim:
- Introduced StandaloneTester for starting/stopping Standalone through ExternalResource support in junit
- Introduced SingletonExternalResource for starting/stopping any ExternalResource in a singletone manner
  - Useful for writing test suites that can have individual test runnable in IDE as well
- Added a logback-test.xml in cdap-standalone test resources so that Standalone logs are output to test output, not in separate file
  - It makes debugging much easier
- Make standalone router starts on random port for tests
- Removed StandaloneContainer and StandaloneTestBase as they are highly redundant and half-baked
- Cleanup all pom.xml to make sure the right scope is used for cdap-common-unit-test module
- Removed the cdap-unit-test-standalone module


cdap-cli-tests/pom.xml
```diff
@@ -35,15 +35,28 @@
...
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-standalone</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-standalone</artifactId>
```
jwang47: Why is cdap-standalone dependency specified twice, except one has test-jar? Seems like we should only need to specify once with test-jar.

chtyim: test-jar dependency doesn't bring in regular dependency, hence need to specify both.

jwang47:
LGTM

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT1868

chtyim:
Squashed commits and merging.

PR #2727, opened 2015-05-20
[CDAP-1584] show username in cli prompt

jayfjin:
Build - http://builds.cask.co/browse/CDAP-DUT1866
JIRA - https://issues.cask.co/browse/CDAP-1584
Added username into cli prompt.

### DIFF
--- a/cdap-cli-tests/src/test/java/co/cask/cdap/cli/CLIMainTest.java
+++ b/cdap-cli-tests/src/test/java/co/cask/cdap/cli/CLIMainTest.java
@@ -30,6 +30,7 @@ import co.cask.cdap.client.app.FakeFlow;
 import co.cask.cdap.client.app.FakeSpark;
 import co.cask.cdap.client.app.FakeWorkflow;
 import co.cask.cdap.client.app.PrefixedEchoHandler;
+import co.cask.cdap.client.config.AuthenticatedConnectionConfig;
 import co.cask.cdap.client.config.ClientConfig;
 import co.cask.cdap.client.config.ConnectionConfig;
 import co.cask.cdap.common.conf.Constants;
@@ -144,6 +145,23 @@ public class CLIMainTest {
     testCommandOutputContains(cli, "list dataset instances", FakeApp.DS_NAME);
     testCommandOutputContains(cli, "list streams", FakeApp.STREAM_NAME);
     testCommandOutputContains(cli, "list flows", FakeApp.FLOWS.get(0));
+  }
+
+  @Test
+  public void testPrompt() throws Exception {
+    String prompt = cliMain.getPrompt(cliConfig.getClientConfig());
+    Assert.assertFalse(prompt.contains("@"));
+    Assert.assertTrue(prompt.contains(STANDALONE.getBaseURI().getHost()));
+    Assert.assertTrue(prompt.contains(cliConfig.getCurrentNamespace().getId()));
+
+    ConnectionConfig oldConnectionConfig = clientConfig.getConnectionConfig();
+    ConnectionConfig authConnectionConfig = new AuthenticatedConnectionConfig(oldConnectionConfig, "test-username");
+    cliConfig.setConnectionConfig(authConnectionConfig);
+    prompt = cliMain.getPrompt(cliConfig.getClientConfig());
+    Assert.assertTrue(prompt.contains("test-username@"));
+    Assert.assertTrue(prompt.contains(STANDALONE.getBaseURI().getHost()));
+    Assert.assertTrue(prompt.contains(cliConfig.getCurrentNamespace().getId()));
+    cliConfig.setConnectionConfig(oldConnectionConfig);
   }
 
   @Test
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
@@ -22,6 +22,7 @@ import co.cask.cdap.cli.util.table.AltStyleTableRenderer;
 import co.cask.cdap.cli.util.table.TableRenderer;
 import co.cask.cdap.cli.util.table.TableRendererConfig;
 import co.cask.cdap.client.MetaClient;
+import co.cask.cdap.client.config.AuthenticatedConnectionConfig;
 import co.cask.cdap.client.config.ClientConfig;
 import co.cask.cdap.client.config.ConnectionConfig;
 import co.cask.cdap.common.exception.UnauthorizedException;
@@ -31,16 +32,18 @@ import co.cask.cdap.security.authentication.client.AuthenticationClient;
 import co.cask.cdap.security.authentication.client.Credential;
 import co.cask.cdap.security.authentication.client.basic.BasicAuthenticationClient;
 import com.google.common.base.Charsets;
-import com.google.common.base.Joiner;
 import com.google.common.base.Supplier;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
 import com.google.common.io.CharStreams;
 import com.google.common.io.Files;
 import com.google.common.io.InputSupplier;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
 import jline.TerminalFactory;
 import jline.console.ConsoleReader;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -59,6 +62,7 @@ public class CLIConfig implements TableRendererConfig {
   private static final int DEFAULT_LINE_WIDTH = 80;
   private static final int MIN_LINE_WIDTH = 40;
 
+  private static final Gson GSON = new Gson();
   private final ClientConfig clientConfig;
   private final FilePathResolver resolver;
   private final String version;
@@ -76,6 +80,27 @@ public class CLIConfig implements TableRendererConfig {
       }
     }
   };
+
+  /*
+   * Wrapper class for reading/writing of username + accessToken
+   */
+  private class UserAccessToken {
+    private final AccessToken accessToken;
+    private final String username;
+
+    public UserAccessToken(AccessToken accessToken, String username) {
+      this.accessToken = accessToken;
+      this.username = username;
+    }
+
+    public AccessToken getAccessToken() {
+      return accessToken;
+    }
+
+    public String getUsername() {
+      return username;
+    }
+  }
 
   /**
    * @param clientConfig client configuration
@@ -116,7 +141,12 @@ public class CLIConfig implements TableRendererConfig {
 
   public void tryConnect(ConnectionConfig connectionConfig, PrintStream output, boolean debug) throws Exception {
     try {
-      AccessToken accessToken = acquireAccessToken(clientConfig, connectionConfig, output, debug);
+      UserAccessToken userToken = acquireAccessToken(clientConfig, connectionConfig, output, debug);
+      AccessToken accessToken = null;
+      if (userToken != null) {
+        accessToken = userToken.getAccessToken();
+        connectionConfig = new AuthenticatedConnectionConfig(connectionConfig, userToken.getUsername());
+      }
       checkConnection(clientConfig, connectionConfig, accessToken);
       setConnectionConfig(connectionConfig);
       output.printf("Successfully connected CDAP instance at %s", connectionConfig.getURI().toString());
@@ -128,8 +158,8 @@ public class CLIConfig implements TableRendererConfig {
   }
 
   public void updateAccessToken(PrintStream output) throws IOException {
-    AccessToken newAccessToken = getNewAccessToken(clientConfig.getConnectionConfig(), output, false);
-    clientConfig.setAccessToken(newAccessToken);
+    UserAccessToken newAccessToken = getNewAccessToken(clientConfig.getConnectionConfig(), output, false);
+    clientConfig.setAccessToken(newAccessToken.getAccessToken());
   }
 
   private void checkConnection(ClientConfig baseClientConfig,
@@ -147,17 +177,21 @@ public class CLIConfig implements TableRendererConfig {
     return getAuthenticationClient(connectionInfo).isAuthEnabled();
   }
 
-  private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConfig connectionInfo,
-                                         PrintStream output, boolean debug) throws IOException {
+  @Nullable
+  private UserAccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConfig connectionInfo,
+                                             PrintStream output, boolean debug) throws IOException {
 
     if (!isAuthenticationEnabled(connectionInfo)) {
       return null;
     }
 
     try {
-      AccessToken savedAccessToken = getSavedAccessToken(connectionInfo.getHostname());
-      checkConnection(clientConfig, connectionInfo, savedAccessToken);
-      return savedAccessToken;
+      UserAccessToken savedToken = getSavedAccessToken(connectionInfo.getHostname());
+      if (savedToken == null) {
+        throw new UnauthorizedException();
+      }
+      checkConnection(clientConfig, connectionInfo, savedToken.getAccessToken());
+      return savedToken;
     } catch (UnauthorizedException ignored) {
       // access token invalid - fall through to try acquiring token manually
     }
@@ -165,14 +199,15 @@ public class CLIConfig implements TableRendererConfig {
     return getNewAccessToken(connectionInfo, output, debug);
   }
 
-  private AccessToken getNewAccessToken(ConnectionConfig connectionInfo, PrintStream output,
-                                        boolean debug) throws IOException {
-
+  private UserAccessToken getNewAccessToken(ConnectionConfig connectionInfo,
+                                            PrintStream output, boolean debug) throws IOException {
     AuthenticationClient authenticationClient = getAuthenticationClient(connectionInfo);
 
     Properties properties = new Properties();
     properties.put(BasicAuthenticationClient.VERIFY_SSL_CERT_PROP_NAME,
                    String.valueOf(clientConfig.isVerifySSLCert()));
+
+    String username = "";
 
     // obtain new access token via manual user input
     output.printf("Authentication is enabled in the CDAP instance: %s.\n", connectionInfo.getHostname());
@@ -186,18 +221,21 @@ public class CLIConfig implements TableRendererConfig {
         credentialValue = reader.readLine(prompt);
       }
       properties.put(credential.getName(), credentialValue);
+      if (credential.getName().contains("username")) {
+        username = credentialValue;
+      }
     }
 
     authenticationClient.configure(properties);
     AccessToken accessToken = authenticationClient.getAccessToken();
-
+    UserAccessToken userToken = new UserAccessToken(accessToken, username);
     if (accessToken != null) {
-      if (saveAccessToken(accessToken, connectionInfo.getHostname()) && debug) {
+      if (saveAccessToken(userToken, connectionInfo.getHostname()) && debug) {
         output.printf("Saved access token to %s\n", getAccessTokenFile(connectionInfo.getHostname()).getAbsolutePath());
       }
     }
 
-    return accessToken;
+    return userToken;
   }
 
   private AuthenticationClient getAuthenticationClient(ConnectionConfig connectionInfo) {
@@ -206,26 +244,23 @@ public class CLIConfig implements TableRendererConfig {
                                            connectionInfo.isSSLEnabled());
     return authenticationClient;
   }
-
-  private AccessToken getSavedAccessToken(String hostname) {
+  @Nullable
+  private UserAccessToken getSavedAccessToken(String hostname) {
     File file = getAccessTokenFile(hostname);
-    if (file.exists() && file.canRead()) {
-      try {
-        String tokenString = Joiner.on("").join(Files.readLines(file, Charsets.UTF_8));
-        return new AccessToken(tokenString, -1L, null);
-      } catch (IOException ignored) {
-        // Fall through
-      }
+    try (BufferedReader reader = Files.newReader(file, Charsets.UTF_8)) {
+      return GSON.fromJson(reader, UserAccessToken.class);
+    } catch (IOException | JsonSyntaxException ignored) {
+      // Fall through
     }
+
     return null;
   }
 
-  private boolean saveAccessToken(AccessToken accessToken, String hostname) {
+  private boolean saveAccessToken(UserAccessToken accessToken, String hostname) {
     File accessTokenFile = getAccessTokenFile(hostname);
 
     try {
-      accessTokenFile.createNewFile();
-      Files.write(accessToken.getValue(), accessTokenFile, Charsets.UTF_8);
+      Files.write(GSON.toJson(accessToken), accessTokenFile, Charsets.UTF_8);
       return true;
     } catch (IOException ignored) {
       // NO-OP
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
@@ -56,6 +56,7 @@ import org.apache.commons.cli.ParseException;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.net.ConnectException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
@@ -143,7 +144,7 @@ public class CLIMain {
         } else if (e instanceof InvalidCommandException) {
           InvalidCommandException ex = (InvalidCommandException) e;
           output.printf("Invalid command '%s'. Enter 'help' for a list of commands\n", ex.getInput());
-        } else if (e instanceof DisconnectedException) {
+        } else if (e instanceof DisconnectedException || e instanceof ConnectException) {
           cli.getReader().setPrompt("cdap (DISCONNECTED)> ");
         } else {
           output.println("Error: " + e.getMessage());
--- a/cdap-client/src/main/java/co/cask/cdap/client/config/AuthenticatedConnectionConfig.java
+++ b/cdap-client/src/main/java/co/cask/cdap/client/config/AuthenticatedConnectionConfig.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.client.config;
+
+import java.net.URI;
+
+/**
+ * {@link ConnectionConfig} that stores a username when available. Helps keep connectionConfig thread safe.
+ */
+public class AuthenticatedConnectionConfig extends ConnectionConfig {
+  private final String username;
+
+  public AuthenticatedConnectionConfig(ConnectionConfig connectInfo, String username) {
+    super(connectInfo.getNamespace(), connectInfo.getHostname(), connectInfo.getPort(), connectInfo.isSSLEnabled());
+    this.username = username + "@";
+  }
+
+  @Override
+  public URI getURI() {
+    return URI.create(String.format("%s://%s%s:%d", super.isSSLEnabled() ? "https" : "http", username,
+      super.getHostname(), super.getPort()));
+  }
+}
+


cdap-client/src/main/java/co/cask/cdap/client/config/ConnectionConfig.java
```diff
@@ -66,10 +67,11 @@ public ConnectionConfig(Id.Namespace namespace, String hostname, int port, boole
     this.hostname = hostname;
     this.port = port;
     this.sslEnabled = sslEnabled;
+    this.username = System.getProperty("user.name");
```
jwang47: The purpose of having a username in the CLI prompt was for when interacting with a secure CDAP cluster, so that the user knows who they've been authenticated as. This will be useful when we implement per-user permissions and probably also useful now, just so the user knows who they are. Why are we showing the "user.name" property instead?

bdmogal: Ok then I perhaps misunderstood the requirement. What should be display rather then ``user.name``?

anwar6953: And in the case of insecure CDAP, don't show username at all, right?
Since the CLI user is not logged into a CDAP user.

jwang47: I was thinking whatever the user is authenticated as. For example if you login to a secure CDAP cluster as the user "sdf", you'd see "cdap(http://sdf@host:port)> " as the prompt. It may be tricky to implement since the existing CDAP authentication server doesn't really tell you who you are (and it may not make sense to have a username in some cases, e.g. when using a different authentication scheme that doesn't accept simple username/password).

jwang47: @anwar6953 yeah, no username for insecure CDAP.

jayfjin:
Updated to only show username in secure CDAP mode.

Build - http://builds.cask.co/browse/CDAP-DUT1866-5

cdap-client/src/main/java/co/cask/cdap/client/config/ConnectionConfig.java
```diff
@@ -66,10 +68,15 @@ public ConnectionConfig(Id.Namespace namespace, String hostname, int port, boole
     this.hostname = hostname;
     this.port = port;
     this.sslEnabled = sslEnabled;
+    if (SecurityRequestContext.getUserId().isPresent()) {
```
jwang47: Have you tried this out? This will never be present in the CLI, as we don't set it. It's only set in CDAP Master (server-side). To get the username, you'll need to capture it when the user provides username/password upon connecting to a secure CDAP instance (see `ConnectCommand.execute()` which calls `CLIConfig.tryConnect`).

jayfjin:
Tested and works on cdap-singlenode 

http://builds.cask.co/browse/CDAP-DUT1866-6

cdap-client/src/main/java/co/cask/cdap/client/config/ConnectionConfig.java
```diff
@@ -66,10 +68,11 @@ public ConnectionConfig(Id.Namespace namespace, String hostname, int port, boole
     this.hostname = hostname;
     this.port = port;
     this.sslEnabled = sslEnabled;
+    this.username = "";
```
bdmogal: Should we add the username field to ``hashCode`` and ``equals`` as well?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -77,6 +77,22 @@ public Integer get() {
     }
   };
 
+  private class AccessTokenWrapper {
+    private AccessToken accessToken;
```
chtyim: Use `final` fields.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -77,6 +77,22 @@ public Integer get() {
     }
   };
 
+  private class AccessTokenWrapper {
```
chtyim: In general, try not to name a class "Wrapper", as it doesn't tell what it wraps. Better call it `UserAccessToken` and have a short javadoc saying what does it contain.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -155,14 +171,19 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
     }
 
     try {
-      AccessToken savedAccessToken = getSavedAccessToken(connectionInfo.getHostname());
-      checkConnection(clientConfig, connectionInfo, savedAccessToken);
-      return savedAccessToken;
+      AccessTokenWrapper savedToken = getSavedAccessToken(connectionInfo.getHostname());
+
+      checkConnection(clientConfig, connectionInfo, savedToken.getAccessToken());
+      connectionInfo.setUsername(savedToken.getUsername());
```
chtyim: So username is not needed for checkConnection? It seems wrong to me.

jayfjin: username isn't needed for checkConnection. I didn't modify that line anyways in terms on needing a username or not.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -186,13 +209,17 @@ private AccessToken getNewAccessToken(ConnectionConfig connectionInfo, PrintStre
         credentialValue = reader.readLine(prompt);
       }
       properties.put(credential.getName(), credentialValue);
+      if (credential.getName().contains("username")) {
```
chtyim: Why use "contains"?

jayfjin: the actual value was something like cdap.security.authentication.basic.username and I wasn't sure if this would change if the authentication method changed.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -207,12 +234,15 @@ private AuthenticationClient getAuthenticationClient(ConnectionConfig connection
...
 
-  private AccessToken getSavedAccessToken(String hostname) {
+  private AccessTokenWrapper getSavedAccessToken(String hostname) {
     File file = getAccessTokenFile(hostname);
     if (file.exists() && file.canRead()) {
       try {
-        String tokenString = Joiner.on("").join(Files.readLines(file, Charsets.UTF_8));
-        return new AccessToken(tokenString, -1L, null);
+        String fileString = Joiner.on("").join(Files.readLines(file, Charsets.UTF_8));
+        int splitIndex = fileString.indexOf("username:");
```
chtyim: Looks like the existing code is bad. The logic for writing and reading of AccessToken is different (stripping newlines on read). In fact, on write we should just serialize the `AccessToken` (or `UserAccessToken` in the new code) using `Gson` and deserialize it back on read.

cdap-client/src/main/java/co/cask/cdap/client/config/ConnectionConfig.java
```diff
@@ -58,6 +59,7 @@ private static String tryResolveAddress(String addressString) {
   private final int port;
   private final boolean sslEnabled;
   private final Id.Namespace namespace;
+  private String username;
```
chtyim: Can it be `final`? Meaning not set through setter method but have to be passed in through the constructor. Adding mutation field turn this class from immutable class to mutable one, making it not thread safe.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -155,14 +171,19 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
     }
 
     try {
-      AccessToken savedAccessToken = getSavedAccessToken(connectionInfo.getHostname());
-      checkConnection(clientConfig, connectionInfo, savedAccessToken);
-      return savedAccessToken;
+      UserAccessToken savedToken = getSavedAccessToken(connectionInfo.getHostname());
+
+      checkConnection(clientConfig, connectionInfo, savedToken.getAccessToken());
+      connectionInfo.setUsername(savedToken.getUsername());
```
chtyim: I think the user name should already been set when the `ConnectionInfo` object was instantiated. See my comment below.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -77,6 +77,22 @@ public Integer get() {
     }
   };
 
+  private class UserAccessToken {
+    private final AccessToken accessToken;
+    private final String username;
+    public UserAccessToken(AccessToken accessToken, String username) {
```
chtyim: Insert an empty line above to separate fields and methods.

jayfjin:
new build - http://builds.cask.co/browse/CDAP-DUT1866-7

cdap-client/src/main/java/co/cask/cdap/client/config/ConnectionConfig.java
```diff
@@ -17,6 +17,7 @@
 
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.http.SecurityRequestContext;
```
anwar6953: remove unused import

cdap-cli-tests/src/test/java/co/cask/cdap/cli/CLIMainTest.java
```diff
@@ -176,6 +177,14 @@ public void testList() throws Exception {
   }
 
   @Test
+  public void testPrompt() throws Exception {
```
jwang47: Can you also add a test case that has username?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -77,6 +80,27 @@ public Integer get() {
     }
   };
 
+  /*
+   * Wrapper class for reading/writing of username + accessToken
+   */
+  private class UserAccessToken {
```
chtyim: After reading the code again, why not just change the "AccessToken" class to include username (maybe as a nullable final field) as well?

jayfjin: accesstoken is imported in a jar since it's source is in a different repo.

jwang47: We have control over it here: https://github.com/caskdata/cdap-clients/tree/develop/cdap-authentication-clients/java.

chtyim: Now I understand what this code does. this class is more like a Client side info (like cookies in browser) to save states across multiple runs of the CLI. It's ok to keep this class as it (probably need to rename to something more appropriate when more fields are added in future).

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -128,7 +153,7 @@ public void tryConnect(ConnectionConfig connectionConfig, PrintStream output, bo
   }
 
   public void updateAccessToken(PrintStream output) throws IOException {
-    AccessToken newAccessToken = getNewAccessToken(clientConfig.getConnectionConfig(), output, false);
+    AccessToken newAccessToken = getNewAccessToken(clientConfig, clientConfig.getConnectionConfig(), output, false);
```
chtyim: If you are passing the `clientConfig`, no need to pass `clientConfig.getConnnectionConfig()`.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -207,12 +242,13 @@ private AuthenticationClient getAuthenticationClient(ConnectionConfig connection
...
   }
 
-  private AccessToken getSavedAccessToken(String hostname) {
+  private UserAccessToken getSavedAccessToken(String hostname) {
     File file = getAccessTokenFile(hostname);
     if (file.exists() && file.canRead()) {
       try {
-        String tokenString = Joiner.on("").join(Files.readLines(file, Charsets.UTF_8));
-        return new AccessToken(tokenString, -1L, null);
+        FileReader reader = new FileReader(file);
```
chtyim: The reader is not closed after used. Also, always provides a Charset when creating `Reader`.

```java
try (Reader reader = Files.newReader(file, Charsets.UTF_8)) {
  return GSON.fromJson(reader, UserAccessToken.class);
}
```

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -207,12 +242,13 @@ private AuthenticationClient getAuthenticationClient(ConnectionConfig connection
     return authenticationClient;
   }
 
-  private AccessToken getSavedAccessToken(String hostname) {
+  private UserAccessToken getSavedAccessToken(String hostname) {
     File file = getAccessTokenFile(hostname);
     if (file.exists() && file.canRead()) {
```
chtyim: Shouldn't need this check. Simply try to read from the file and it'll throw IOException if the file is not readable.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -220,12 +256,12 @@ private AccessToken getSavedAccessToken(String hostname) {
     return null;
   }
 
-  private boolean saveAccessToken(AccessToken accessToken, String hostname) {
+  private boolean saveAccessToken(UserAccessToken accessToken, String hostname) {
     File accessTokenFile = getAccessTokenFile(hostname);
 
     try {
       accessTokenFile.createNewFile();
```
chtyim: Not necessary to create the file first.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -155,25 +180,31 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
     }
 
     try {
-      AccessToken savedAccessToken = getSavedAccessToken(connectionInfo.getHostname());
-      checkConnection(clientConfig, connectionInfo, savedAccessToken);
-      return savedAccessToken;
+      UserAccessToken savedToken = getSavedAccessToken(connectionInfo.getHostname());
+      connectionInfo = new AuthenticatedConnectionConfig(connectionInfo, savedToken.getUsername());
+      clientConfig.setConnectionConfig(connectionInfo);
+      checkConnection(clientConfig, connectionInfo, savedToken.getAccessToken());
```
chtyim: Same here. You already have ClientConfig passed in, no need to pass ConnectionInfo separately.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -155,25 +180,31 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
...
+      return savedToken.getAccessToken();
     } catch (UnauthorizedException ignored) {
       // access token invalid - fall through to try acquiring token manually
     }
 
-    return getNewAccessToken(connectionInfo, output, debug);
+    AccessToken token = getNewAccessToken(clientConfig, output, debug);
+    return token;
```
chtyim: You should see a warning in IDE here. Simply do `return getNewAccessToken(...);`

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -117,7 +141,8 @@ public void setConnectionConfig(@Nullable ConnectionConfig connectionConfig) {
   public void tryConnect(ConnectionConfig connectionConfig, PrintStream output, boolean debug) throws Exception {
     try {
       AccessToken accessToken = acquireAccessToken(clientConfig, connectionConfig, output, debug);
-      checkConnection(clientConfig, connectionConfig, accessToken);
+      connectionConfig = clientConfig.getConnectionConfig();
```
jwang47: This introduces a bug, where you can't connect to any cluster other than the default localhost one. The `connectionConfig` arg is intended to be the new connection, whereas `clientConfig.getConnectionConfig()` is the current connection. Also, try to avoid modifying method args.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -128,13 +153,13 @@ public void tryConnect(ConnectionConfig connectionConfig, PrintStream output, bo
   }
 
   public void updateAccessToken(PrintStream output) throws IOException {
-    AccessToken newAccessToken = getNewAccessToken(clientConfig.getConnectionConfig(), output, false);
+    AccessToken newAccessToken = getNewAccessToken(clientConfig, output, false);
     clientConfig.setAccessToken(newAccessToken);
   }
 
   private void checkConnection(ClientConfig baseClientConfig,
-                               ConnectionConfig connectionInfo,
```
jwang47: Revert changes to this method. `connectionInfo` is intended to override the `baseClientConfig`'s connection info, where `connectionInfo` is intended to be the new connection we want to check whereas `baseClientConfig` has timeout config + current connection.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
chtyim: I think the old "checkConnection" and ClientConfig is not good. If we use checkConnection to validate a connection before it can be set into ClientConfig, then checkConnection should take an incomplete "ClientConfig" (ideally the ClientConfig should be an immutable object). Refactoring is needed on both of them.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -155,25 +180,30 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
     }
 
     try {
-      AccessToken savedAccessToken = getSavedAccessToken(connectionInfo.getHostname());
-      checkConnection(clientConfig, connectionInfo, savedAccessToken);
-      return savedAccessToken;
+      UserAccessToken savedToken = getSavedAccessToken(connectionInfo.getHostname());
+      connectionInfo = new AuthenticatedConnectionConfig(connectionInfo, savedToken.getUsername());
+      clientConfig.setConnectionConfig(connectionInfo);
```
jwang47: Don't set connection config here. This method is only intended to acquire an access token, and can't assume that the process of using the new connection will run without failure after this method is executed.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -155,25 +180,30 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
...
+      checkConnection(clientConfig, savedToken.getAccessToken());
+      return savedToken.getAccessToken();
     } catch (UnauthorizedException ignored) {
       // access token invalid - fall through to try acquiring token manually
     }
 
-    return getNewAccessToken(connectionInfo, output, debug);
+    return getNewAccessToken(clientConfig, output, debug);
```
jwang47: Pass `connectionInfo` instead. You don't need the entire `clientConfig`.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
chtyim: Seems like this method should be more like "createClientConfig" that takes a ClientConfig.Builder and return a ClientConfig by setting the connectionConfig to the builder.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -186,13 +216,18 @@ private AccessToken getNewAccessToken(ConnectionConfig connectionInfo, PrintStre
         credentialValue = reader.readLine(prompt);
       }
       properties.put(credential.getName(), credentialValue);
+      if (credential.getName().contains("username")) {
+        username = credentialValue;
+        connectionInfo = new AuthenticatedConnectionConfig(connectionInfo, username);
+        clientConfig.setConnectionConfig(connectionInfo);
```
jwang47: Don't set connection config here.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -155,25 +180,30 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
...
       // access token invalid - fall through to try acquiring token manually
     }
 
-    return getNewAccessToken(connectionInfo, output, debug);
+    return getNewAccessToken(clientConfig, output, debug);
   }
 
-  private AccessToken getNewAccessToken(ConnectionConfig connectionInfo, PrintStream output,
-                                        boolean debug) throws IOException {
+  private AccessToken getNewAccessToken(ClientConfig clientConfig,
```
jwang47: Instead of modifying `clientConfig` for setting `UserAccessToken`, you should just return `UserAccessToken`. This method isn't intended to modify anything, just return an `AccessToken`.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -207,25 +242,22 @@ private AuthenticationClient getAuthenticationClient(ConnectionConfig connection
     return authenticationClient;
   }
 
-  private AccessToken getSavedAccessToken(String hostname) {
+  private UserAccessToken getSavedAccessToken(String hostname) {
```
jwang47: Annotate this method with `@Nullable` since it can return null.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -155,25 +180,30 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
     }
 
     try {
-      AccessToken savedAccessToken = getSavedAccessToken(connectionInfo.getHostname());
-      checkConnection(clientConfig, connectionInfo, savedAccessToken);
-      return savedAccessToken;
+      UserAccessToken savedToken = getSavedAccessToken(connectionInfo.getHostname());
+      connectionInfo = new AuthenticatedConnectionConfig(connectionInfo, savedToken.getUsername());
```
jwang47: Check for null `savedToken`.

cdap-client/src/main/java/co/cask/cdap/client/config/AuthenticatedConnectionConfig.java
```diff
@@ -0,0 +1,38 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.client.config;
+
+import java.net.URI;
+
+/**
+ * Ensures that ConnectionConfig remains thread safe
```
jwang47: What's this comment about?

jwang47:
There's a lot of connection/auth logic in `CLIConfig` that we should move into a separate class. Can you move it out?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
jayfjin: I need the connectionInfo to be modified here. Java uses pass by value so any changes I make in the function getNewAccessToken to the connectionConfig won't be saved unless I can set the connectionConfig via clientConfig. (specifically here: connectionInfo = new AuthenticatedConnectionConfig(connectionInfo, username);)

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
jayfjin: Again, connectionConfig won't be changed in the parent function unless I set it here.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
jwang47: I agree we need refactoring here, but `checkConnection` doesn't take an incomplete `ClientConfig` though, it's intended to be the current `ClientConfig` that has various timeout, etc. props that we want override with the new connection.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
jwang47: Actually, the new connection is set here: https://github.com/caskdata/cdap/blob/feature/show-username/cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java#L146. You don't need to set it again in this method. Also this method is only intended to return a new access token that is obtained by asking the user for credentials, so it doesn't need to modify anything.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
jwang47: +1

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
jwang47: Oh, so the reason for your change is to set `clientConfig`'s connection to your `AuthenticatedConnectionConfig` instance. However, it doesn't make sense to set `clientConfig`'s connection before the connection is tested and successful. Otherwise, the connection in the CLI can be set to a connection that doesn't work.

jwang47: Instead of setting `clientConfig`'s connection, you should simply return `UserAccessToken` instead of returning `AccessToken`. Then, in `tryConnect`, you'll call `checkConnection` which will create your `AuthenticatedConnectionConfig` and returns it.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
jwang47: @jayfjin Let's make `ClientConfig` immutable.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -147,7 +176,8 @@ private boolean isAuthenticationEnabled(ConnectionConfig connectionInfo) throws
     return getAuthenticationClient(connectionInfo).isAuthEnabled();
   }
 
-  private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConfig connectionInfo,
+  @Nullable
+  private UserAccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConfig connectionInfo,
                                          PrintStream output, boolean debug) throws IOException {
```
jwang47: This looks misaligned.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -165,15 +198,16 @@ private AccessToken acquireAccessToken(ClientConfig clientConfig, ConnectionConf
     return getNewAccessToken(connectionInfo, output, debug);
   }
 
-  private AccessToken getNewAccessToken(ConnectionConfig connectionInfo, PrintStream output,
-                                        boolean debug) throws IOException {
-
+  private UserAccessToken getNewAccessToken(ConnectionConfig connectionInfo,
+                                        PrintStream output, boolean debug) throws IOException {
```
jwang47: misaligned

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -207,25 +244,22 @@ private AuthenticationClient getAuthenticationClient(ConnectionConfig connection
...
-    if (file.exists() && file.canRead()) {
-      try {
-        String tokenString = Joiner.on("").join(Files.readLines(file, Charsets.UTF_8));
-        return new AccessToken(tokenString, -1L, null);
-      } catch (IOException ignored) {
-        // Fall through
-      }
+    try (BufferedReader reader = Files.newReader(file, Charsets.UTF_8)) {
+      return GSON.fromJson(reader, UserAccessToken.class);
+    } catch (IOException ignored) {
```
jwang47: Add `JsonSyntaxException` to this block. Otherwise we'll run into issues if we try to read an old version of the access token.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -143,7 +144,8 @@ public boolean handleException(PrintStream output, Exception e, int timesRetried
         } else if (e instanceof InvalidCommandException) {
           InvalidCommandException ex = (InvalidCommandException) e;
           output.printf("Invalid command '%s'. Enter 'help' for a list of commands\n", ex.getInput());
-        } else if (e instanceof DisconnectedException) {
+        } else if (e instanceof DisconnectedException || (e instanceof ConnectException
+                   && e.getMessage().equals("Connection refused"))) {
```
jwang47: I think we don't need to check the message for `ConnectException`. See the javadoc.

jwang47:
LGTM

PR #2729, opened 2015-05-20
CDAP-2520 allow zero for delay and duration

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-DT96-1


albertshau:
0 makes sense for delay, but not for duration.  If we're not checking that in the stream source, we should check it. Otherwise lgtm

gokulavasan:
@albertshau Updated PR. Build : http://builds.cask.co/browse/CDAP-DT96-2

albertshau:
lgtm

PR #2730, opened 2015-05-20
CDAP-2447 adapter config can have no transforms property

gokulavasan:
Can contain only source and sink if no transforms are required.

Build : http://builds.cask.co/browse/CDAP-DT97

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/ETLConfig.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/ETLConfig.java
@@ -17,6 +17,7 @@
 package co.cask.cdap.template.etl.common;
 
 import co.cask.cdap.api.Resources;
+import com.google.common.base.Optional;
 
 import java.util.List;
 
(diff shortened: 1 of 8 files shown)


cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/ETLConfig.java
```diff
@@ -44,8 +45,8 @@ public ETLStage getSink() {
     return sink;
   }
 
-  public List<ETLStage> getTransforms() {
-    return transforms;
+  public Optional<List<ETLStage>> getTransforms() {
+    return Optional.fromNullable(transforms);
```
chtyim: Do we need to return an Optional? Can it be just `return transforms == null ? new ArrayList<>() : transforms;`?

gokulavasan: Yes, that will work too. Will update the PR.

gokulavasan:
@chtyim Build passed. Please take a look. Thanks!

chtyim:
LGTM

PR #2731, opened 2015-05-20
Un-static IntegrationTestBase

jwang47:
http://builds.cask.co/browse/CDAP-DUT1871

Also make instanceUri and accessToken overridable by subclasses.


chtyim:
LGTM

PR #2732, opened 2015-05-20
CDAP-2549 Delete suspended workflow run record

dtzeng:
http://builds.cask.co/browse/CDAP-DUT1867-2
https://issues.cask.co/browse/CDAP-2549

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/store/DefaultStoreTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/store/DefaultStoreTest.java
@@ -126,6 +126,31 @@ public class DefaultStoreTest {
   }
 
   @Test
+  public void testDeleteSuspendedWorkflow() {
+    Id.Namespace namespaceId = new Id.Namespace("namespace1");
+
+    // Test delete application
+    Id.Application appId1 = new Id.Application(namespaceId, "app1");
+    Id.Program programId1 = Id.Program.from(appId1, ProgramType.WORKFLOW, "pgm1");
+    RunId run1 = RunIds.generate();
+    store.setStart(programId1, run1.getId(), runIdToSecs(run1));
+    store.setSuspend(programId1, run1.getId());
+    store.removeApplication(appId1);
+    Assert.assertTrue(store.getRuns(programId1, ProgramRunStatus.ALL, 0, Long.MAX_VALUE, Integer.MAX_VALUE).isEmpty());
+
+    // Test delete namespace
+    Id.Application appId2 = new Id.Application(namespaceId, "app2");
+    Id.Program programId2 = Id.Program.from(appId2, ProgramType.WORKFLOW, "pgm2");
+    RunId run2 = RunIds.generate();
+    store.setStart(programId2, run2.getId(), runIdToSecs(run2));
+    store.setSuspend(programId2, run2.getId());
+    store.removeAll(namespaceId);
+    store.deleteNamespace(namespaceId);
+    Assert.assertTrue(store.getRuns(programId2, ProgramRunStatus.ALL, 0, Long.MAX_VALUE, Integer.MAX_VALUE).isEmpty());
+  }
+
+
+  @Test
   public void testConcurrentStopStart() throws Exception {
     // Two programs that start/stop at same time
     // Should have two run history.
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/store/DefaultStoreTest.java
```diff
@@ -126,6 +126,31 @@ public void testStopBeforeStart() throws RuntimeException {
...
+    Id.Namespace namespaceId = new Id.Namespace("namespace1");
+
+    // Test delete application
+    Id.Application appId1 = new Id.Application(namespaceId, "app1");
+    Id.Program programId1 = Id.Program.from(appId1, ProgramType.WORKFLOW, "pgm1");
+    RunId run1 = RunIds.generate();
+    store.setStart(programId1, run1.getId(), runIdToSecs(run1));
+    store.setSuspend(programId1, run1.getId());
+    store.removeApplication(appId1);
+    Assert.assertEquals(store.getRuns(programId1, ProgramRunStatus.ALL, 0, Long.MAX_VALUE, Integer.MAX_VALUE).size(), 0);
```
henry-cask: The Assert.assertEquals first parameter is the expected one, so probably better to switch the input vars:
```Assert.assertEquals(0, store.getRuns(programId1, ProgramRunStatus.ALL, 0, Long.MAX_VALUE, Integer.MAX_VALUE).size());```

File: cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/store/DefaultStoreTest.java [hunk 2]
```diff
@@ -126,6 +126,31 @@ public void testStopBeforeStart() throws RuntimeException {
...
+
+    // Test delete namespace
+    Id.Application appId2 = new Id.Application(namespaceId, "app2");
+    Id.Program programId2 = Id.Program.from(appId2, ProgramType.WORKFLOW, "pgm2");
+    RunId run2 = RunIds.generate();
+    store.setStart(programId2, run2.getId(), runIdToSecs(run2));
+    store.setSuspend(programId2, run2.getId());
+    store.removeAll(namespaceId);
+    store.deleteNamespace(namespaceId);
+    Assert.assertEquals(store.getRuns(programId2, ProgramRunStatus.ALL, 0, Long.MAX_VALUE, Integer.MAX_VALUE).size(), 0);
```
henry-cask: Similar to prev comment

dtzeng:
@henry-cask fixed the assertion and checkstyle

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/store/DefaultStoreTest.java
```diff
@@ -126,6 +126,33 @@ public void testStopBeforeStart() throws RuntimeException {
...
+    Id.Namespace namespaceId = new Id.Namespace("namespace1");
+
+    // Test delete application
+    Id.Application appId1 = new Id.Application(namespaceId, "app1");
+    Id.Program programId1 = Id.Program.from(appId1, ProgramType.WORKFLOW, "pgm1");
+    RunId run1 = RunIds.generate();
+    store.setStart(programId1, run1.getId(), runIdToSecs(run1));
+    store.setSuspend(programId1, run1.getId());
+    store.removeApplication(appId1);
+    Assert.assertEquals(0, store.getRuns(programId1, ProgramRunStatus.ALL, 0,
```
chtyim: Not major, but using `isEmpty` is better than checking size is 0.

```Assert.assertTrue(store.getRuns(...).isEmpty());```

dtzeng: cleaned it up

chtyim:
LGTM

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/store/DefaultStoreTest.java
anwar6953: @chtyim One thing to note then, is that if it fails, you then have to debug further to see what the actual size was, if using `isEmpty()`.

henry-cask:
Call to isEmpty maybe better for some types of List impl but if the test failed the log report would not tell you what was the wrong value.

dtzeng:
http://builds.cask.co/browse/CDAP-DUT1867-3

Build passed, squashing and merging

PR #2733, opened 2015-05-21
CDAP-1998: Remove deprecated Context - Query param in Metrics v3 API

abaranau:


cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -273,33 +253,11 @@ private void tagsQuerying(HttpRequest request, HttpResponder responder, List<Str
...
-    }
-  }
-
   private MetricQueryResult executeQuery(HttpRequest request, Map<String, String> sliceByTags,
-                                         List<String> groupByTags, List<String> metrics) {
-    try {
-      MetricQueryRequest queryRequest = new MetricQueryRequest(sliceByTags, metrics, groupByTags);
-      setTimeRangeInQueryRequest(queryRequest, new QueryStringDecoder(request.getUri()).getParameters());
-      return executeQuery(queryRequest);
-    } catch (IllegalArgumentException e) {
```
gsps1: we don't want to differentiate between invalid request and internal_server_error ?

abaranau: it is old code. It is handled one level up now

abaranau: I.e. it *was* not differentiating, now it is

gsps1:
LGTM

abaranau:
http://builds.cask.co/browse/CDAP-DUT1946-2

abaranau:
http://builds.cask.co/browse/CDAP-DUT1946-11 GREEN

PR #2735, opened 2015-05-21
fix hoisting issue with dataset status page

elfenheart:
because of hoisting in JavaScript, datasetTags variable is undefined when pollMetric function is called.


anwar6953:
LGTM

PR #2742, opened 2015-05-21
Avoid starting standalone in IntegrationTestBase when instanceUri is provided

jwang47:
http://builds.cask.co/browse/CDAP-DUT1882

### DIFF
--- a/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
+++ b/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
@@ -66,13 +66,16 @@ import javax.annotation.Nullable;
 public class IntegrationTestBase {
 
   @ClassRule
-  public static final SingletonExternalResource STANDALONE = new SingletonExternalResource(new StandaloneTester());
-
-  @ClassRule
   public static final TemporaryFolder TEMP_FOLDER = new TemporaryFolder();
 
+  public static final SingletonExternalResource STANDALONE = new SingletonExternalResource(new StandaloneTester());
+
   @Before
-  public void setUp() throws Exception {
+  public void setUp() throws Throwable {
+    if (startStandalone()) {
+      STANDALONE.before();
+    }
+
     assertNoApps();
     assertNoUserDatasets();
     // TODO: check metrics, streams, etc.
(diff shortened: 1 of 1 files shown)


cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -108,6 +117,10 @@ protected TestManager getTestManager() {
     }
   }
 
+  protected boolean startStandalone() {
+    return System.getProperty("instanceUri", "").isEmpty();
```
anwar6953: call `getInstanceURI`

anwar6953:
LGTM

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -66,13 +66,16 @@
...
 
   @ClassRule
-  public static final SingletonExternalResource STANDALONE = new SingletonExternalResource(new StandaloneTester());
-
-  @ClassRule
   public static final TemporaryFolder TEMP_FOLDER = new TemporaryFolder();
 
+  public static final SingletonExternalResource STANDALONE = new SingletonExternalResource(new StandaloneTester());
+
   @Before
```
chtyim: This gets execute per test method. It's different that using `@ClassRule`.

chtyim: It's better implements as:

```java
@ClassRule
public static final SingletonExternalResource STANDALONE = new SingletonExternalResource(new StandaloneTester()) {
  @Override
  public void before() throws Throwable {
    if (getInstanceURI().isEmpty()) {
      super.before();
    }
  }

  @Override
  public void after() {
    if (getInstanceURI().isEmpty()) {
      super.after();
    }
  }
};

jwang47: The problem is that I don't want `getInstanceURI` to be `static` because we want our subclass to override how we obtain the instanceURI. This is so that we can read the instanceURI and accessToken from a file.

PR #2743, opened 2015-05-21
CDAP-2016: CLI support for starting programs in debug mode

jwang47:
https://issues.cask.co/browse/CDAP-2016
http://builds.cask.co/browse/CDAP-DUT1883-4

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2014 Cask Data, Inc.
+ * Copyright © 2015 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
@@ -32,16 +32,16 @@ import java.io.PrintStream;
 import java.util.Map;
 
 /**
- * Starts a program.
+ * Starts a program in debug mode.
  */
-public class StartProgramCommand extends AbstractAuthCommand {
+public class StartDebugProgramCommand extends AbstractAuthCommand {
 
   private static final Gson GSON = new Gson();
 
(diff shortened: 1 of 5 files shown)


cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright © 2014 Cask Data, Inc.
```
awholegunch: 2015!

cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
```diff
@@ -92,15 +93,22 @@ public ProgramClient(ClientConfig config) {
    * @throws ProgramNotFoundException if the program with the specified name could not be found
```
awholegunch: missing the debug parameter in the Javadocs. Goes above runtimeArgs; can be copied from the method after this one.

cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
```diff
@@ -112,6 +120,23 @@ public void start(String appId, ProgramType programType, String programName, Map
    * @param appId ID of the application that the program belongs to
```
awholegunch: Should change the method description to indicate why this method has been added. How about:
"Starts a program, giving debug mode, using the stored runtime arguments."

cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
```diff
@@ -0,0 +1,88 @@
...
+    String[] programIdParts = arguments.get(elementType.getArgumentName().toString()).split("\\.");
+    if (programIdParts.length < 2) {
+      throw new CommandInputError(this);
+    }
+
+    String appId = programIdParts[0];
+    String programId = programIdParts[1];
+
+    String runtimeArgsString = arguments.get(ArgumentName.RUNTIME_ARGS.toString(), "");
+    if (runtimeArgsString == null || runtimeArgsString.isEmpty()) {
```
sreevatsanraman: I don't think we need to fetch the run time args and pass it. The preferences API would resolve the stored runtime arguments and pass it to the program. Can you open a JIRA for fixing this in 3.1, there are other places where this needs to change as well and not only for debug,

sreevatsanraman:
Open a JIRA for the issue mentioned and merge. 

LGTM :+1:

PR #2744, opened 2015-05-21
CDAP-2529: CLI print error on failed autoconnect

jwang47:
https://issues.cask.co/browse/CDAP-2529

Now when you try to execute a command in the CLI but the provided connection cannot be used, you see the following:

```
CDAP instance at 'http://localhost:10000' could not be reached: Connection refused
```

instead of

```
```


sreevatsanraman:
LGTM :+1:

PR #2745, opened 2015-05-21
CDAP-2007: Render CLI output for CallServiceCommand in a copy-pastable way

jwang47:
https://issues.cask.co/browse/CDAP-2007
http://builds.cask.co/browse/CDAP-DUT1885

```
cdap (http://localhost:10000/default)> call service PurchaseHistory.PurchaseHistoryService GET /history/joe
< 200 OK
< Content-Length: 33
< Connection: keep-alive
< Content-Type: text/plain; charset=UTF-8
{"aslkdjfjklad":"lksjdfklsdf"}
```


sreevatsanraman:
LGTM :+1:

PR #2749, opened 2015-05-21
change service program type to be consistent with other programs

elfenheart:
Create Runs tabs on Service program type.

screen shot 2015-05-21 at 4 38 33 pm


cdap-ui/app/directives/instance-control/instance-control.js
```diff
@@ -16,22 +15,12 @@ angular.module(PKG.name + '.commons')
...
-            myAlert({
-              title: 'Invalid instances: ',
-              content: 'you must request a valid number of instances.',
-              type: 'danger'
-            });
-            scope.processing = false;
-            return;
-          }
-
           scope.myDataSrc.request({
```
ajainarayanan: If we are having a separate controller for GETing the instances, shouldn't this POST be moved to the same controller? Link functions are generally used to hook any event listeners on DOM elements (generally, could be used for multiple different things but in general used for accessing the DOM). Thoughts?

ajainarayanan:
One minor comment. Apart from that :+1: LGTM.

ajainarayanan:
:+1: LGTM

PR #2750, opened 2015-05-22
Correct Application Template documentation

awholegunch:
Issue 2509 (https://issues.cask.co/browse/CDAP-2509) described how a link for third-party plugins is bad. On examining the problem, it was concluded that the problem was more the organization of the material rather than merely a bad link. This PR addresses that.

The original material is the last part of the section titled ["Plugin Configuration for Prebuilt Jars"](http://docs.cdap.io/cdap/current/en/application-templates/etl/custom.html#plugin-configuration-for-prebuilt-jars), beginning with the text "Prebuilt JARs".  The first link leads to a place that doesn't really make much sense.

The material has been moved around, so that:
- the info on Prebuilt jars is on its [own page](http://docs-staging.cask.co/cdap/3.0.0-r3.0_fix_app_template_docs/en/application-templates/etl/third-party.html); 
- the material that was *Plugin Configuration for Prebuilt Jars* has been renamed and moved to [Sample Adaptor Configurations](http://docs-staging.cask.co/cdap/3.0.0-r3.0_fix_app_template_docs/en/application-templates/etl/creating.html#sample-adaptor-configurations);
- It's now at the end of [Configuration File Format](http://docs-staging.cask.co/cdap/3.0.0-r3.0_fix_app_template_docs/en/application-templates/etl/creating.html) of *Creating an ETL Adaptor*.

The PR also changes a couple of titles so they don't show in the table of contents.


gokulavasan:
LGTM :+1:

PR #2751, opened 2015-05-22
CDAP-2208 fixing dataset and explore classloading issues.

albertshau:
The first issue is that the FilterClassLoader does not filter
when getResource is called. When a dataset jar is built, this
can result in system jars getting leaked into the dataset jar.
The second issue is that in standalone mode, the classloader used
to instantiate DatasetType is a URLClassLoader, which means the
bundled lib jars are not visible. Finally, in explore, the
classloader a dataset is the same as the classloader for that
dataset's type, with the system classloader as a parent. This means
that system classes are leaked. Instead, the parent classloader
must be a FilterClassLoader. To support this, adding a couple
new ways to create a FilterClassLoader.

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/lang/FilterClassLoader.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/lang/FilterClassLoader.java
@@ -16,16 +16,20 @@
 
 package co.cask.cdap.common.lang;
 
+import co.cask.cdap.proto.ProgramType;
 import com.google.common.base.Predicate;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterators;
+import com.google.common.base.Predicates;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.Set;
 
 /**
  * ClassLoader that filters out certain resources.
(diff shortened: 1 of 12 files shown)


cdap-common/src/main/java/co/cask/cdap/common/lang/FilterClassLoader.java
```diff
@@ -97,7 +126,7 @@ protected Package getPackage(String name) {
 
   @Override
   public URL getResource(String name) {
-    return super.getResource(name);
+    return resourceAcceptor.apply(name) ? super.getResource(name) : null;
```
chtyim: You have the override `getResources` and `getResoruceAsStream` as well. Also, remove `findResource` and `findResources`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/LocalDatasetTypeClassLoaderFactory.java
```diff
@@ -16,24 +16,57 @@
...
-    if (moduleMeta.getJarLocation() == null) {
+    URI jarLocation = moduleMeta.getJarLocation();
+    if (jarLocation == null) {
       return parentClassLoader;
     }
 
-    return new URLClassLoader(new URL[]{moduleMeta.getJarLocation().toURL()}, parentClassLoader);
+    File jarFile = new File(jarLocation);
+    String dirName = String.format("%s.%s", moduleMeta.getName(), Files.hash(jarFile, Hashing.md5()).toString());
+    File expandDir = new File(jarFile.getParent(), dirName);
```
chtyim: So this will never get deleted?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/LocalDatasetTypeClassLoaderFactory.java
```diff
@@ -16,24 +16,57 @@
...
+          if (tempDir.exists()) {
+            DirUtils.deleteDirectoryContents(tempDir);
+          }
+        } catch (IOException e) {
+          // Just log, no propagate
+          LOG.warn("Failed to delete temp directory {}", tempDir, e);
+        }
+      }
+    }
+    return new DirectoryClassLoader(expandDir, parentClassLoader, "lib");
```
chtyim: Is this ClassLoader cached somewhere? Otherwise every call to this method will create a new one and eventually we'll run out of file descriptors. Maybe you can refer to `PluginInstantiator` on how it manages and deletes directories.

albertshau: we'll also use different classloaders for the same dataset. I think caching it correctly will fix https://issues.cask.co/browse/CDAP-23

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/LocalDatasetTypeClassLoaderFactory.java
albertshau: deleting this correctly will require some refactoring up the stack to make sure things are closed properly. Opened https://issues.cask.co/browse/CDAP-2564 to do this.

albertshau:
@chtyim what do you think of merging this first then handling the caching and cleanup separately? They require refactoring, and are not issues introduced by this bug fix.

albertshau:
build running at http://builds.cask.co/browse/CDAP-DUT1892

chtyim:
@albertshau Is it possible to do the refactoring first (in separate PR) as that's the prerequisite to have this fix works correctly in standalone. We should merge workable code only.

albertshau:
http://builds.cask.co/browse/CDAP-DUT2141

albertshau:
@chtyim addressed comments and rebased this on top of the dataset cleanup and cache changes. Can you take another look?

chtyim:
LGTM

PR #2752, opened 2015-05-22
CDAP-2422: Remove unnecessary catch clauses in StreamHandler

jwang47:
http://builds.cask.co/browse/CDAP-DUT1895-4

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/ConcurrentStreamWriter.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/ConcurrentStreamWriter.java
@@ -17,6 +17,7 @@ package co.cask.cdap.data.stream.service;
 
 import co.cask.cdap.api.flow.flowlet.StreamEvent;
 import co.cask.cdap.api.stream.StreamEventData;
+import co.cask.cdap.common.exception.NotFoundException;
 import co.cask.cdap.data.file.FileWriter;
 import co.cask.cdap.data.file.FileWriters;
 import co.cask.cdap.data.stream.StreamCoordinatorClient;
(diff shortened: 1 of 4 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/ConcurrentStreamWriter.java
```diff
@@ -234,7 +237,7 @@ private EventQueue getEventQueue(Id.Stream streamId) throws IOException {
       return eventQueue;
 
     } catch (Exception e) {
-      Throwables.propagateIfInstanceOf(e, IllegalArgumentException.class);
+      Throwables.propagateIfInstanceOf(e, NotFoundException.class);
```
chtyim: In Java 7, you can have this as:

```java
try {
...
} catch (NotFoundException | IOException e) {
  throw e;
} catch (Exception e) {
  throw new IOException(e);
}
```

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamHandler.java
```diff
@@ -186,28 +187,18 @@ public void getInfo(HttpRequest request, HttpResponder responder,
...
-      LOG.error("Failed to create stream {}", stream, e);
-      responder.sendString(HttpResponseStatus.BAD_REQUEST, e.getMessage());
-    } catch (Exception e) {
-      LOG.error("Failed to create stream {}", stream, e);
-      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
-    }
+    // TODO: Modify the REST API to support custom configurations.
+    streamAdmin.create(streamId);
+
+    // TODO: For create successful, 201 Created should be returned instead of 200.
```
bdmogal: I think we decided to not do this, so this TODO is no longer valid

bdmogal:
:+1:

PR #2755, opened 2015-05-22
Fix Documentation Build in Develop

awholegunch:
The develop documentation build has been broken because it wasn't correctly determining that its branch was...develop. Or master, for that matter. This fixes that.

It also corrects the developers-manual use of the including repos that it was downloading.

Fixes https://issues.cask.co/browse/CDAP-2562


cdap-docs/_common/common-build.sh
```diff
@@ -327,10 +328,12 @@ function version() {
...
   PROJECT_LONG_VERSION=`expr "${PROJECT_VERSION}" : '\([0-9]*\.[0-9]*\.[0-9]*\)'`
   PROJECT_SHORT_VERSION=`expr "${PROJECT_VERSION}" : '\([0-9]*\.[0-9]*\)'`
-  IFS=/ read -a branch <<< "`git rev-parse --abbrev-ref HEAD`"
-  # Determine branch and branch type: one of develop, master, release, develop-feature, release-feature
+  local full_branch=`git rev-parse --abbrev-ref HEAD`
+  IFS=/ read -a branch <<< "${full_branch}"
   GIT_BRANCH="${branch[1]}"
-  if [ "x${GIT_BRANCH}" == "xdevelop" -o  "x${GIT_BRANCH}" == "xmaster" ]; then
+  # Determine branch and branch type: one of develop, master, release, develop-feature, release-feature
+  if [ "x${full_branch}" == "xdevelop" -o  "x${full_branch}" == "xmaster" ]; then
```
wolf31o2: This is a horrible practice. Your conditional is checking for an exact string. This x${var} stuff is completely unnecessary. Just check the contents. Using this syntax is only appropriate when attempting to tell the difference between "empty" and "not set" in shell, not when doing string comparison. Even then, there are better ways of doing it.

wolf31o2:
:+1: LGTM

PR #2758, opened 2015-05-24
fix handling of null values in ObjectMappedTable

anew:
- add a unit test for ObjectMappedTable that uses a null field and fails without the fix
- add a unit test for ReflectionPutWriter (used by ObjectMappedTable) that covers more cases
- fix ReflectionPutWriter not to skip null fields 
- (add some generated files in client tests to git ignore)


albertshau:
lgtm, thanks for fixing this

PR #2761, opened 2015-05-26
Improve error handling in Standalone Main

awholegunch:
This adds better error handling to Standalone Main; if args are passed, and they aren't asking for the help/usage statement, then the args are printed (in addition to usage) instead of just giving a somewhat-misleading error message. This flags what is being passed so it can be corrected.

It also edits some comments.


cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
```diff
@@ -289,7 +289,12 @@ public static void main(String[] args) {
         usage(false);
         return;
       } else {
-        usage(true);
+         PrintStream out = System.err;
```
bdmogal: Indentation seems to be off by 1 character for this entire block.

bdmogal: Indentation seems to be off by 1 character for this entire block.

bdmogal: Also, the variable name ``out`` for ``System.err`` is a little misleading. Should we instead use ``err``, if absolutely required, else stick with ``System.err``, since ``System.err`` is used after this block as well, on line 307.

chtyim:
Seems like the whole usage thing shouldn't be in StandaloneMain at all, but rather in the cdap.sh/cdap.bat. The StandaloneMain itself doesn't use any of the arguments.

awholegunch:
On-hold until we can build again and I can test and check this...

cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
```diff
@@ -250,52 +250,16 @@ private void cleanupTempDir() {
...
-    if (args.length > 0) {
-      if ("--help".equals(args[0]) || "-h".equals(args[0])) {
-        usage(false);
-        return;
-      } else {
-        usage(true);
-      }
-    }
-
     StandaloneMain main = null;
```
chtyim: No need to assign to null.

chtyim:
Just one comment, rest LGTM.

awholegunch:
Building...

cdap-standalone/src/main/java/co/cask/cdap/StandaloneMain.java
awholegunch: Except then I get a compilation failure: http://builds.cask.co/browse/CDAP-DUT1897-4/log

awholegunch:
@chtyim I think that main needs to be set to null before the try block because otherwise you can get into the catch without it having been initialized. Can you take another look?

awholegunch:
Building...http://builds.cask.co/browse/CDAP-DUT1897-6

chtyim:
LGTM

PR #2762, opened 2015-05-26
Remove README.md files

awholegunch:
This removes markdown-based README files. It converts them to reST, but also removes in them some embedded references to earlier (2.6.0!) CDAP versions.

It adjusts the pom.xml files for the archetypes so that they are consistent across the different archetypes.

Fixes https://issues.cask.co/browse/CDAP-683


chtyim:
LGTM

PR #2764, opened 2015-05-26
[CDAP-2525] Inconsistent Projection Transform properties in UI

elfenheart:
Change projection transform 'Rename' field to use KeyValue widget


ajainarayanan:
:+1: LGTM

PR #2765, opened 2015-05-26
go to the most current runs when a program start is clicked

elfenheart:
Fixing a problem where when a user is at /runs/:runid and they click start a new run button, it does not go to the most recent runs.

### DIFF
--- a/cdap-ui/app/directives/start-stop-button/start-stop-button.js
+++ b/cdap-ui/app/directives/start-stop-button/start-stop-button.js
@@ -44,7 +44,12 @@ angular.module(PKG.name + '.commons')
           }
           dataSrc.request(requestObj)
             .then(function() {
-              $state.go($state.current, $state.params, {reload: true});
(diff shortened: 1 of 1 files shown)


cdap-ui/app/directives/start-stop-button/start-stop-button.js
```diff
@@ -44,7 +44,12 @@ angular.module(PKG.name + '.commons')
           }
           dataSrc.request(requestObj)
             .then(function() {
-              $state.go($state.current, $state.params, {reload: true});
+              if ($state.params.runid && action === 'start') {
```
ajainarayanan: Talked to @elfenheart. Changing this to use ```$state.is``` to check if we are in /runs or /runs/:runid states.

ajainarayanan:
One minor comment. Otherwise :+1: LGTM

ajainarayanan:
:+1: LGTM

PR #2767, opened 2015-05-26
Feature/get program manager apis

anwar6953:
This introduces a `ApplicationManager#getXManager` (such as `applicationManager.getFlowManager` or `applicationManager.getWorkflowManager`), because sometimes the methods of a program manager are needed without actually starting the program, which is how the program managers were previously available.

Also, applicationManager.startWorkflow did not previously start the workflow, because the usage of it simply needed the manager (without starting it). This PR also fixes that.

Also moved all of the `startX` (`startFlow`, `startWorkflow`, etc) to `AbstractApplicationManager`.

http://builds.cask.co/browse/CDAP-DUT1906-4

Note: This also resolves an issue: https://issues.cask.co/browse/CDAP-799


cdap-test/src/main/java/co/cask/cdap/test/AbstractApplicationManager.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.Map;
+
+/**
+ * A base implementation of {@link ApplicationManager}.
+ */
+public abstract class AbstractApplicationManager implements ApplicationManager {
+
+  @Override
+  public FlowManager startFlow(final String flowName) {
```
chtyim: I think all the "startXXX" should go into XXXManager. E.g. FlowManager.start().

anwar6953: and deprecate these?

chtyim: yes. this one simply do "getFlowManager().start()"

anwar6953:
Ok, added `start` methods to program managers. Also, the `runningProcesses` had `programName` as the key of the map, so I made it a set instead, because the same programName can be used multiple times (across namespaces).

anwar6953:
Replaced usages of the now deprecated `startX` methods. It does make the code a bit lengthier.

cdap-examples/FileSetExample/src/test/java/co/cask/cdap/examples/fileset/FileSetWordCountTest.java
```diff
@@ -56,7 +56,8 @@ public void testWordCountOnFileSet() throws Exception {
     final String line2 = "b a b";
 
     // discover the file set service
-    ServiceManager serviceManager = applicationManager.startService("FileSetService");
+    ServiceManager serviceManager = applicationManager.getServiceManager("FileSetService");
+    serviceManager.start();
```
chtyim: You can have the `start()` method returns the manager so that you can chain calls.

```java
ServiceManager serviceManager = applicationManager.getServiceManager("FileSetService").start();
```

anwar6953: Also then, other APIs of the manager should also return the object itself?

chtyim: It depends on what they do. I would say at this point, we don't have to change all of them.

anwar6953:
@jwang47 @chtyim 
Does this seem appropriate:
https://github.com/caskdata/cdap/commit/149574bf2264bacd782bc573a8ee5adeed90f635

jwang47:
https://github.com/caskdata/cdap/commit/149574bf2264bacd782bc573a8ee5adeed90f635 looks alright to me. Doesn't look nice, but doubt there are any better ways to do it.

anwar6953:
@jwang47 `runningProcesses` was something being managed previously. (Maybe to avoid errors upon stopping a program. The AppFabricClient stop method is only called if the program is within `runningProcesses`, for instance).

If it's suggested to remove `runningProcesses` (since yes, we can just check state every time, and implement stopAll differently), then I prefer to do that change in a separate PR.

jwang47:
LGTM

cdap-test/src/main/java/co/cask/cdap/test/ProgramManager.java
```diff
@@ -18,13 +18,28 @@
...
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 /**
  * Instance for this class is for managing a {@link Id.Program}.
+ * @param <T> The type of ProgramManager
  */
-public interface ProgramManager {
+public interface ProgramManager<T> {
```
chtyim: `<T extends ProgramManager>`

cdap-test/src/main/java/co/cask/cdap/test/AbstractProgramManager.java
```diff
@@ -34,6 +37,19 @@ public AbstractProgramManager(Id.Program programId, ApplicationManager applicati
...
+  public T start() {
+    return start(ImmutableMap.<String, String>of());
+  }
+
+  @Override
+  public T start(Map<String, String> arguments) {
+    applicationManager.startProgram(programId, arguments);
+    // this cast is fine as long as the derived classes extend AbstractProgramManager with the
+    // template (<T>) declared as its own class
+    return (T) this;
```
chtyim: I think you don't need to cast if you have `<T extends ProgramManager>` in the `ProgramManager` interface.

cdap-test/src/main/java/co/cask/cdap/test/ProgramManager.java
anwar6953: done.

cdap-test/src/main/java/co/cask/cdap/test/AbstractProgramManager.java
anwar6953: I tried it; didn't seem to work.

chtyim: ok.

PR #2771, opened 2015-05-27
Fixes adapter create view on smaller screen (metadata was unusable on…

ajainarayanan:
… small screens)

### DIFF
--- a/cdap-ui/app/features/adapters/templates/create.html
+++ b/cdap-ui/app/features/adapters/templates/create.html
@@ -18,7 +18,7 @@
         </div>
         <div class="panel-body" ng-if="metadataTabOpen">
           <div class="row">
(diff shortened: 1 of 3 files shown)


cdap-ui/app/features/adapters/templates/create.html
```diff
@@ -43,7 +43,7 @@ <h2 class="pull-left" style="margin-top:0;">Create Adapter</h2>
                 </ui-select-choices>
               </ui-select>
             </div>
-            <div class="col-lg-5">
+            <div class="col-lg-5 col-md-5 col-xs-12">
```
elfenheart: you can remove the col-lg-5

elfenheart:
:+1: LGTM

PR #2725, opened 2015-05-20
Corrections to examples

awholegunch:
This PR updates the examples for changes made in CDAP 3.0 that didn't make in the release. In particular, it updates the instructions to handle the new CDAP UI, and updates screenshots.

Where possible, it replaces examples and scripts with use of the CDAP CLI. It removes duplication of commands for demonstrating Windows, instead relying on an initial introductory portion describing the difference. It tries to streamline the examples by adding additional headers, and make them consistent and easier to follow. They have all been tested with CDAP 3.0.0.

In cases where Java files are included, tests have been added to the build script to check for any changes in those files, to prevent issues such as incorrect inclusions being made.

Capitalization has been reduced or eliminated, to follow current thinking on documenting CDAP components.

Staged at: [CDAP](http://docs-staging.cask.co/cdap/3.0.0-feature-r3.0_fix_examples/en/index.html)

Pages of interest: [Examples](http://docs-staging.cask.co/cdap/3.0.0-feature-r3.0_fix_examples/en/examples-manual/examples/index.html)

PR #2772, opened 2015-05-27
Adding documentation about deleting streams.

awholegunch:
https://issues.cask.co/browse/CDAP-2554

### DIFF
--- a/cdap-docs/developers-manual/source/building-blocks/streams.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/streams.rst
@@ -4,9 +4,9 @@
 
 .. _streams:
 
-============================================
+=======
 Streams
-============================================
+=======
 
 **Streams** are the primary means of bringing data from external systems into the CDAP in realtime.
 They are ordered, time-partitioned sequences of data, usable for realtime collection and consumption of data.
@@ -15,7 +15,7 @@ They can be created programmatically within your application, using the
 :ref:`http-restful-api-stream`, the :ref:`stream-client` of the :ref:`client-api`, or by
 using the :ref:`CDAP Command Line Interface. <cli>` 
 
-Data written to a Stream can be consumed in real-time by :ref:`Flows <flows>` or in batch
+Data written to a stream can be consumed in real-time by :ref:`Flows <flows>` or in batch
 by :ref:`MapReduce programs. <mapreduce>`.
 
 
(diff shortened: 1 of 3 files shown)


anwar6953:
Also modify the text in question here:
https://github.com/caskdata/cdap/pull/2725#discussion_r30744853

awholegunch:
Good catch! Thanks, I was looking for that one...

cdap-docs/developers-manual/source/building-blocks/streams.rst
```diff
@@ -63,12 +63,16 @@ changed, using the :ref:`http-restful-api-stream`, the :ref:`stream-client` of t
 :ref:`client-api`, or by using the :ref:`Command Line Interface. <cli>`
 
 
-.. rubric::  Truncating a Stream
+.. rubric::  Truncating and Deleting a Stream
 
 Streams can be truncated, which means deleting all events that were ever written to the
 Stream. This is permanent and cannot be undone. They can be truncated through the using
```
jwang47: lowercase

awholegunch: I would prefer, since this is a keyword in a title, that this remain Title Case. Would you agree?

jwang47: I don't think you're supposed to capitalize words in a paragraph just because they appear in the title of the paragraph. If that was the case, then why don't we capitalize "truncating" and "deleting"?

cdap-docs/reference-manual/source/http-restful-api/stream.rst
```diff
@@ -294,6 +294,50 @@ A Stream can be truncated with an HTTP POST method to the URL::
    * - Description
      - Delete all events in the Stream named *mystream* in the namespace *default*
 
+Deleting a Stream
+-----------------
+Deleting a stream means both deleting all events that were ever written to the stream and
+the stream endpoint itself. This is permanent and cannot be undone. If another stream is
+created with the same name, it will not return any of the previous stream's events.
+
+A Stream can be deleted with an HTTP DELETE method to the URL::
```
jwang47: stream

cdap-docs/reference-manual/source/http-restful-api/stream.rst
```diff
@@ -294,6 +294,50 @@ A Stream can be truncated with an HTTP POST method to the URL::
...
+.. list-table::
+   :widths: 20 80
+   :header-rows: 1
+
+   * - Parameter
+     - Description
+   * - ``<namespace>``
+     - Namespace ID
+   * - ``<stream-id>``
+     - Name of an existing Stream
```
jwang47: stream

cdap-docs/developers-manual/source/building-blocks/streams.rst
awholegunch: I think I'm mis-reading this. (The way Github had highlighted words...) I was talking about "Truncating and Deleting a Stream"; I suspect you were talking about "...that were ever written to the Stream." Correct? 

Yes, "Stream" should be "stream", but I was going to do a big pass over all the docs and make all those changes at once. I know you have started to do this, but I was putting it off...Was that what you were thinking?

jwang47: Yup, looks like we're in agreement.

jwang47:
LGTM

PR #2773, opened 2015-05-27
Fixes js error + cyclic dependency between MyDataSource,  and  modules

ajainarayanan:
TODO:
  Need to replace auto refresh on backend start which involves one single API from backend to query for backend status.


elfenheart:
:+1: LGTM

PR #2774, opened 2015-05-27
(CDAP-2569) Retry master services on start failure.

chtyim:
- Use RetryOnStartFailureService to wrap around starting and stopping of master services

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
@@ -60,6 +60,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.io.Closeables;
+import com.google.common.util.concurrent.AbstractIdleService;
 import com.google.common.util.concurrent.Service;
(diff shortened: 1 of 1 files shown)


cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
```diff
@@ -150,6 +152,43 @@ public MasterServiceMain() {
     this.serviceStore = injector.getInstance(ServiceStore.class);
     this.secureStoreUpdater = baseInjector.getInstance(TokenSecureStoreUpdater.class);
     this.leaderElection = createLeaderElection();
+
+    this.masterServices = new RetryOnStartFailureService(new Supplier<Service>() {
+      @Override
+      public Service get() {
+        return new AbstractIdleService() {
+          @Override
+          protected void startUp() throws Exception {
```
jwang47: So all of the operations in `startUp` can be executed multiple times without a corresponding call to `shutDown`?

chtyim: Still trying it out. The current way Services start/stop is not very good.

chtyim:
It's not a very good fix. Still working on it.

PR #2775, opened 2015-05-27
(CDAP-2548) Make UsageDataset backward compatible for the type change.

chtyim:


cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/UsageRegistry.java
```diff
@@ -55,11 +56,15 @@ public UsageRegistry(TransactionExecutorFactory txExecutorFactory, final Dataset
       @Override
       public UsageDatasetIterable get() {
         try {
-          UsageDataset usageDataset = DatasetsUtil.getOrCreateDataset(datasetFramework,
+          Object usageDataset = DatasetsUtil.getOrCreateDataset(datasetFramework,
```
bdmogal: Indentation seems to be off on the 3 lines below.

bdmogal:
minor comment. Rest :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT294-1

PR #2779, opened 2015-05-27
Fixes empty responses from backend to be handled by datasource + fixe…

ajainarayanan:
…s custom json propery in requests


elfenheart:
:+1: LGTM

PR #2780, opened 2015-05-27
CDAP-2557: Move out non-handler logic from DatasetInstanceHandler

jwang47:
http://builds.cask.co/browse/CDAP-DUT1915-11

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
@@ -18,14 +18,8 @@ package co.cask.cdap.data2.datafabric.dataset.service;
 
 import co.cask.cdap.api.dataset.DatasetSpecification;
 import co.cask.cdap.api.dataset.table.Table;
-import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.exception.NotFoundException;
-import co.cask.cdap.data2.datafabric.dataset.instance.DatasetInstanceManager;
-import co.cask.cdap.data2.datafabric.dataset.service.executor.DatasetOpExecutor;
-import co.cask.cdap.data2.datafabric.dataset.type.DatasetTypeManager;
-import co.cask.cdap.data2.registry.UsageRegistry;
-import co.cask.cdap.explore.client.ExploreFacade;
 import co.cask.cdap.proto.DatasetInstanceConfiguration;
 import co.cask.cdap.proto.DatasetMeta;
 import co.cask.cdap.proto.DatasetSpecificationSummary;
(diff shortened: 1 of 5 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceService.java
```diff
@@ -0,0 +1,302 @@
...
+                                                          .addAll(props.getProperties())
+                                                          .build());
+    instanceManager.add(namespace, spec);
+
+    // Enable explore
+    enableExplore(newInstance, props);
+  }
+
+  /**
+   * Updates an existing Dataset specification properties  {@link DatasetInstanceConfiguration}
```
awholegunch: Missing a period. Should be "Updates an existing Dataset specification properties."

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceService.java
```diff
@@ -0,0 +1,302 @@
...
+   * is constructed based on request and the Dataset instance is updated.
+   *
+   * @throws DatasetNotFoundException if the dataset was not found
```
awholegunch: missing parameters:
```
@param instance the dataset instance
@param  properties the dataset properties to be updated
```

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceService.java
```diff
@@ -0,0 +1,302 @@
...
+      // Type not found in the instance's namespace. Now try finding it in the system namespace
+      Id.DatasetType systemDatasetTypeId = Id.DatasetType.from(Constants.SYSTEM_NAMESPACE_ID, typeName);
+      typeMeta = implManager.getTypeInfo(systemDatasetTypeId);
+    }
+    return typeMeta;
+  }
+
+  /**
+   * Drops a dataset.
+   * @param spec specification of dataset to be dropped.
```
awholegunch: Insert a blank line before the @param, and remove the trailing period on the parameter.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
```diff
@@ -82,30 +75,17 @@
...
-    this.implManager = implManager;
-    this.instanceManager = instanceManager;
-    this.exploreFacade = exploreFacade;
-    this.usageRegistry = usageRegistry;
-    this.allowDatasetUncheckedUpgrade = conf.getBoolean(Constants.Dataset.DATASET_UNCHECKED_UPGRADE);
+  public DatasetInstanceHandler(DatasetInstanceService instanceService) {
+    this.instanceService = instanceService;
   }
 
   @GET
```
anew: May I kindly ask that this class gets some Javadocs. I think it is important to document, for each handler method, what its path and query parameters mean.

anew:
Overall I have quite a few concerns with this code. However, that is not subject of this PR, which appears to only move some logic from the HTTP handler into a new service. Which is cleaner and therefore better. What I would like to see is Javadocs that explain the non-obvious @PathParam and @QueryParam parameters of the handler methods. (For example the "owner" - what is that and what form does it have to have?)

Other than that LGTM.

PR #2781, opened 2015-05-27
Adding Streams API Factory

elfenheart:


cdap-ui/app/features/admin/controllers/AdminOverviewController.js
```diff
@@ -48,8 +46,14 @@ angular.module(PKG.name + '.feature.admin')
...
 
     function getStreams (namespace) {
-      return dataSrc.request({
-        _cdapPath: '/namespaces/' + namespace.name + '/streams'
-      });
+      var params = {
+        namespace: namespace.name,
+        scope: $scope
+      };
+      myStreamApi.list(params)
```
ajainarayanan: Just mentioning it,
  We can still return ```myStreamApi.list(params).$promise``` and still have the   ```then``` handler function just like we used to have. This is fine too but noticed we used ```getDatasets``` which returned a promise and ```getStreams``` which also returns a promise but we have the then handler inside getStreams. Some minor things that we might want to consider for consistency.

ajainarayanan:
:+1: LGTM

PR #2783, opened 2015-05-28
CDAP-2254 Fixed the error message returned by handler

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-2254
Build: http://builds.cask.co/browse/CDAP-DUT1918-1

Fix the following bug: when one tries to suspend or resume a stopped/completed/failed workflow, he will get incorrect message "Unable to retrieve runtime information for %s (Application: %s)"

How to fix:
For the two methods suspendWorkflowRun and resumeWorkflowRun in WorkflowHttpHandler.java, once I found the run-id is stopped/completed/failed, I sent an error message at once.
Since sendInvalidResponse method is not used any more, delete it.

Test case:
1. Tested fix manually on standalone as well as on distributed mode.
2. Added a unit test case to the corresponding test file, to test I get correct code when I try to suspend/resume after a workflow is completed.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
@@ -54,6 +54,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -89,7 +89,8 @@ public void suspendWorkflowRun(HttpRequest request, final HttpResponder responde
       Id.Program id = Id.Program.from(namespaceId, appId, ProgramType.WORKFLOW, workflowName);
       ProgramRuntimeService.RuntimeInfo runtimeInfo = runtimeService.list(id).get(RunIds.fromString(runId));
       if (runtimeInfo == null) {
-        sendInvalidResponse(responder, id);
+        responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Runtime information for run-id %s not found",
```
chtyim: I think the new standard for returning not found is to throw NotFoundException. Check with @jwang47

henry-cask: We should use "run id" instead of "run-id" to be consistent

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -118,7 +119,8 @@ public void resumeWorkflowRun(HttpRequest request, final HttpResponder responder
...
```
henry-cask: We should use "run id" instead of "run-id" to be consistent

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
jwang47: Yeah, we should throw NotFoundException instead. HttpExceptionHandler catches the exception and produces a response: https://github.com/caskdata/cdap/blob/develop/cdap-common/src/main/java/co/cask/cdap/common/exception/HttpExceptionHandler.java#L35

yaojiefeng: Now I am using NotFoundException.

yaojiefeng:
@chtyim @jwang47 
Please take a look again.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -100,8 +100,11 @@ public void suspendWorkflowRun(HttpRequest request, final HttpResponder responde
       }
       controller.suspend().get();
       responder.sendString(HttpResponseStatus.OK, "Program run suspended.");
-    }  catch (SecurityException e) {
+    } catch (SecurityException e) {
```
jwang47: You can remove this catch-block since HttpExceptionHandler is handling it.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -100,8 +100,11 @@ public void suspendWorkflowRun(HttpRequest request, final HttpResponder responde
...
       responder.sendStatus(HttpResponseStatus.UNAUTHORIZED);
+    } catch (NotFoundException e) {
```
jwang47: You can remove this catch-block since HttpExceptionHandler is handling it.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -129,30 +131,11 @@ public void resumeWorkflowRun(HttpRequest request, final HttpResponder responder
...
-        status = AppFabricServiceStatus.PROGRAM_NOT_FOUND;
-      } else if (ProgramController.State.COMPLETED.toString().equals(programStatus.getStatus())
-          || ProgramController.State.KILLED.toString().equals(programStatus.getStatus())
-          || ProgramController.State.ERROR.toString().equals(programStatus.getStatus())) {
-        status = AppFabricServiceStatus.PROGRAM_ALREADY_STOPPED;
-      } else {
-        status = AppFabricServiceStatus.RUNTIME_INFO_NOT_FOUND;
-      }
-      responder.sendString(status.getCode(), status.getMessage());
     } catch (SecurityException e) {
```
jwang47: You can remove this catch-block since HttpExceptionHandler is handling it.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -129,30 +131,11 @@ public void resumeWorkflowRun(HttpRequest request, final HttpResponder responder
...
       responder.sendStatus(HttpResponseStatus.UNAUTHORIZED);
+    } catch (NotFoundException e) {
```
jwang47: You can remove this catch-block since HttpExceptionHandler is handling it.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -100,8 +100,11 @@ public void suspendWorkflowRun(HttpRequest request, final HttpResponder responde
       }
       controller.suspend().get();
       responder.sendString(HttpResponseStatus.OK, "Program run suspended.");
-    }  catch (SecurityException e) {
+    } catch (SecurityException e) {
       responder.sendStatus(HttpResponseStatus.UNAUTHORIZED);
+    } catch (NotFoundException e) {
+      LOG.warn("NotFoundException while runtime information for run id is not found.", e);
+      responder.sendString(HttpResponseStatus.NOT_FOUND, e.getMessage());
     } catch (Throwable e) {
```
jwang47: You can remove these catch-blocks (SecurityException, NotFoundException, Throwable) since HttpExceptionHandler is handling them already.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -129,30 +131,11 @@ public void resumeWorkflowRun(HttpRequest request, final HttpResponder responder
...
-      } else {
-        status = AppFabricServiceStatus.RUNTIME_INFO_NOT_FOUND;
-      }
-      responder.sendString(status.getCode(), status.getMessage());
     } catch (SecurityException e) {
       responder.sendStatus(HttpResponseStatus.UNAUTHORIZED);
+    } catch (NotFoundException e) {
+      LOG.warn("NotFoundException while runtime information for run id is not found.", e);
+      responder.sendString(HttpResponseStatus.NOT_FOUND, e.getMessage());
     } catch (Throwable e) {
```
jwang47: You can remove these catch-blocks (SecurityException, NotFoundException, Throwable) since HttpExceptionHandler is handling them already.

jwang47: Looks like the issue is that in the unit test, it's calling the handler methods directly, so `HttpExceptionHandler.handle()` is never invoked. Can you fix that?

yaojiefeng: The try catch blocks are deleted under your help.

yaojiefeng:
@jwang47 
Third commit, please take a look again, thank you.

jwang47:
I don't think `WorkflowHttpHandlerTest` will pass now. Can you check?

yaojiefeng:
@jwang47 Yes the test case passes.

jwang47:
LGTM then.

yaojiefeng:
Thanks @jwang47 @chtyim @henry-cask  for reviewing the PR. New build is running here - http://builds.cask.co/browse/CDAP-DUT1918-3. Will merge once the build passes.

yaojiefeng:
Build passed, merging

PR #2785, opened 2015-05-28
Convenience method in MonitorClient to check if all system services a…

anwar6953:
Convenience method in MonitorClient to check if all system services are 'OK' or not.
https://builds.cask.co/browse/CDAP-DUT1920-1


jwang47:
LGTM

PR #2786, opened 2015-05-28
CDAP-2580 Feature/ui namespace delete

elfenheart:
Adding ability to delete namespace

screen shot 2015-05-28 at 12 06 57 pm


cdap-ui/app/features/admin/controllers/NamespaceSettingsController.js
```diff
@@ -0,0 +1,51 @@
...
+        _cdapPath: '/namespaces/' + $state.params.nsadmin + '/properties',
+        method: 'PUT',
+        body: {
+          'description': $scope.description
+        }
+      })
+      .then(function () {
+        $alert({
+          type: 'success',
+          content: 'Namespace have successfully updated'
```
anwar6953: This message seems grammatically incorrect.
Either `Namespace` should have a `s` at the end, or the `have` should be replaced by a `has`.

bdmogal: actually - just remove the ``have``

ajainarayanan:
:+1: LGTM

PR #2787, opened 2015-05-28
CDAP-2466: CLI for metrics search and query

jwang47:
https://issues.cask.co/browse/CDAP-2466
http://builds.cask.co/browse/CDAP-DUT1923-4

```
cdap (http://localhost:10000/default)> search metric tags "namespace=default"
app=CountRandom
app=HelloWorld
app=Netlens
app=PurchaseHistory
app=TwitterSentiment
app=Wise
app=WordCount
app=etlBatch
component=gateway
dataset=TimeSeriesTable
dataset=TotalCounts
dataset=anomalies
dataset=anomalyCounters
dataset=bounceCountStore
dataset=cloudera
dataset=converted_stream
dataset=counters
dataset=frequentCustomers
dataset=history
dataset=pageViewStore
dataset=purchases
dataset=randomTable
dataset=system.queue
dataset=topN
dataset=trafficCounters
dataset=uniqueCount
dataset=uniqueCounters
dataset=userProfiles
dataset=whom
dataset=wordAssocs
dataset=wordCounts
dataset=wordStats
stream=boo
stream=cloudwick
stream=foo
stream=intuit
stream=logEventStream
stream=logs
stream=mark
stream=myStream
stream=nithin
stream=purchaseStream
stream=searslog
stream=test
stream=who
stream=wordStream
stream=xyz

cdap (http://localhost:10000/default)> search metric tags "namespace=default stream=test"

cdap (http://localhost:10000/default)> search metric names "namespace=default stream=test"
system.collect.bytes
system.collect.events

cdap (http://localhost:10000/default)> get metric value system.collect.events "namespace=default stream=test" start now-60m end now
Start time: 1432838322
End time: 1432841922

Series: system.collect.events
+====================+
| timestamp  | value |
+====================+
| 1432841894 | 1     |
| 1432841901 | 1     |
| 1432841916 | 3     |
| 1432841917 | 4     |
| 1432841918 | 3     |
| 1432841919 | 1     |
| 1432841920 | 1     |
+====================+

cdap (http://localhost:10000/default)> get metric value system.collect.events "namespace=default stream=test"
Start time: 0
End time: 1432841934

Series: system.collect.events
+===================+
| timestamp | value |
+===================+
| 0         | 10014 |
+===================+
```

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
@@ -86,6 +86,6 @@ public class GetMetricCommand extends AbstractAuthCommand {
 
   @Override
   public String getDescription() {
-    return "Gets the value of a metric. Provide <tags> as a map in the format 'a=b, c=d'.";
+    return "Gets the value of a metric. Provide <tags> as a map in the format 'tag1=value1 tag2=value2'.";
   }
 }
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
@@ -54,7 +54,7 @@ public class GetMetricCommand extends AbstractAuthCommand {
     String end = arguments.get("end", "");
 
     MetricQueryResult result = client.query(
-      ImmutableList.of(metric), ImmutableList.<String>of(), tags,
+      tags, ImmutableList.of(metric), ImmutableList.<String>of(),
       start.isEmpty() ? null : start,
       end.isEmpty() ? null : end);
 
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.cli.command.metrics;
+
+import co.cask.cdap.cli.CLIConfig;
+import co.cask.cdap.cli.util.AbstractAuthCommand;
+import co.cask.cdap.cli.util.ArgumentParser;
+import co.cask.cdap.cli.util.RowMaker;
+import co.cask.cdap.cli.util.table.Table;
+import co.cask.cdap.client.MetricsClient;
+import co.cask.cdap.proto.MetricQueryResult;
+import co.cask.common.cli.Arguments;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+import com.google.inject.Inject;
+
+import java.io.PrintStream;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Gets the value of a metric.
+ */
+public class GetMetricCommand extends AbstractAuthCommand {
+
+  private final MetricsClient client;
+
+  @Inject
+  public GetMetricCommand(MetricsClient client, CLIConfig cliConfig) {
+    super(cliConfig);
+    this.client = client;
+  }
+
+  @Override
+  public void perform(Arguments arguments, PrintStream output) throws Exception {
+    String metric = arguments.get("metric-name");
+    Map<String, String> tags = ArgumentParser.parseMap(arguments.get("tags", ""));
+    String start = arguments.get("start", "");
+    String end = arguments.get("end", "");
+
+    MetricQueryResult result = client.query(
+      ImmutableList.of(metric), ImmutableList.<String>of(), tags,
+      start.isEmpty() ? null : start,
+      end.isEmpty() ? null : end);
+
+    output.printf("Start time: %d\n", result.getStartTime());
+    output.printf("End time: %d\n", result.getEndTime());
+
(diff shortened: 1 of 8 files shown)


cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
```diff
@@ -0,0 +1,91 @@
...
+  public GetMetricCommand(MetricsClient client, CLIConfig cliConfig) {
+    super(cliConfig);
+    this.client = client;
+  }
+
+  @Override
+  public void perform(Arguments arguments, PrintStream output) throws Exception {
+    String metric = arguments.get("metric-name");
+    Map<String, String> tags = ArgumentParser.parseMap(arguments.get("tags", ""));
+    String start = arguments.get("start", "");
```
anwar6953: can you pass `null` as the second argument?

cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  @Override
+  public String getPattern() {
+    return "get metric value <metric-name> [<tags>] [start <start>] [end <end>]";
+  }
+
+  @Override
+  public String getDescription() {
```
anwar6953: how come in the github comment, there's no comma separator for the tags, but in this description there is?

cdap-docs/reference-manual/source/cli-api.rst
```diff
@@ -285,6 +285,10 @@ These are the available commands:
    ``truncate dataset instance <dataset-name>``,"Truncates a Dataset."
    **Explore**
    ``execute <query> [<timeout>]``,"Executes a Query with optional <timeout> in minutes (default is no timeout)."
+   **Metrics**
+   ``get metric value <metric-name> [<tags>] [start <start>] [end <end>]``,"Gets the value of a metric. Provide <tags> as a map in the format 'a=b, c=d'."
```
anwar6953: Why is the <tags> map format different for this and the following two lines? (same inconsistency in the string the java code returns).

cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
jwang47: Oh, this is a typo..

cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
jwang47: Not right now. Need to change `Arguments` to support that, but it's in a separate repo.

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -133,6 +133,23 @@ public MetricQueryResult query(List<String> metrics, List<String> groupBys, Map<
    * @throws IOException if a network error occurred
    * @throws UnauthorizedException if the request is not authorized successfully in the gateway server
    */
+  public MetricQueryResult query(List<String> metrics, List<String> groupBys, Map<String, String> tags,
```
anwar6953: btw, maybe in separate PR, but I recommend have `tags` parameter before `groupBys`, primarily because tags are used much more frequently than groupBys. groupBys are often omitted.

anwar6953: Also, it seems weird to always require user to pass in a list of metrics (even though the user can simply do `ImmutableList.of(theirMetric)`.
I think its quite common to simply query for a single metric.

jwang47: Created JIRA https://issues.cask.co/browse/CDAP-2588

anwar6953: sounds good..

cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
anwar6953: ok

cdap-cli/src/main/java/co/cask/cdap/cli/command/metrics/GetMetricCommand.java
anwar6953: fixed in https://github.com/caskdata/cdap/commit/677ceb8150c2804426e91954d05d9eedcc567f1c

abaranau:
> namespace=default

would namespace:default be better? Or its against CLI style?

abaranau:
> search metric tags "namespace=default stream=test"

would comma be better separator than space? Or it is agains CLI style?

abaranau:
> get metric value system.collect.events "namespace=default stream=test" start now-60m end now

Would it be in more readable style with:

```
get metric value system.collect.events "namespace=default stream=test" from now-60m to now
```

abaranau:
Other than above, the format looks fine

anwar6953:
I've found that the pattern (for instance, in preferences and runtime arguments), is with `=` symbol and spare separator.
As for the `from now-60m to now` vs `start now-60m end now`, the latter seems more consistent with the parameters used in the REST endpoint.

abaranau:
I'm good for the syntax, if it makes it consistent with other commands, apart from
> As for the from now-60m to now vs start now-60m end now, the latter seems more consistent with the parameters used in the REST endpoint.

Do we try to have same names in CLI as in REST? Didn't know we do

anwar6953:
"Do we try to have same names in CLI as in REST? Didn't know we do"

I don't know if there's a conscious effort to do so, but why not?
It seems like, in the url, the query parameter names are chosen based upon their meaning - `start`, `end`, etc. - and so I see no reason not to use those terms in the CLI.

abaranau:
> I don't know if there's a conscious effort to do so, but why not?

We won't be able to do it everywhere - that can get confusing. In general.

In this case - I'm ok with param names. Let's merge it!

anew:
We do want to be consistent between different commands in the CLI. What is the syntax used for reading a time range from a stream? Looks like that is just "get stream <start> <end>" :(

anwar6953:
Hmm, one of the reasons I can see for metrics query to explicitly have to specify `start` and `end` keywords is because there are more arguments in the metrics query command.
The stream command, on the other hand, has much fewer arguments.
Though maybe, if the `start` and `end` are autocompleted (suggested), then they should also be in the stream command, as it would be useful to a user.

anew:
Agreed. I guess this will need an extra PR since this one is already merged.

PR #2789, opened 2015-05-28
CDAP-2560: Exclude append-increments from tx changeset

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-RBT295-1

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
```diff
@@ -249,46 +250,39 @@ public void startTx(Transaction tx) {
...
-    // we resolve conflicts on row level of individual table
-    List<byte[]> changes = new ArrayList<byte[]>(buff.size());
-    for (byte[] changedRow : buff.keySet()) {
-      changes.add(Bytes.add(getNameAsTxChangePrefix(), changedRow));
+    Collection<byte[]> changes = new TreeSet<byte[]>(UnsignedBytes.lexicographicalComparator());
+    for (Map.Entry<byte[], NavigableMap<byte[], Update>> rowChange : buff.entrySet()) {
+      for (Map.Entry<byte[], Update> updateEntry : rowChange.getValue().entrySet()) {
+        byte[] changeKey = getChangeKey(rowChange.getKey(), updateEntry.getKey(), updateEntry.getValue());
+        if (changeKey != null) {
+          changes.add(Bytes.add(getNameAsTxChangePrefix(), changeKey));
```
anew: This incurs unnecessary overhead of byte copying etc. when conflict detection level is row. Because you need to do the Bytes.add() repeatedly for columns in the same row, then add to the set that already has that key. In case of ROW, you should only generate and add the change key once. I believe the old code was doing that. I also find the old code more intuitive.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
```diff
@@ -249,46 +250,39 @@ public void startTx(Transaction tx) {
...
     }
     return changes;
   }
 
-  private Collection<byte[]> getColumnChanges() {
-    // we resolve conflicts on row level of individual table
-    List<byte[]> changes = new ArrayList<byte[]>(buff.size());
-    for (Map.Entry<byte[], NavigableMap<byte[], Update>> rowChange : buff.entrySet()) {
-      if (rowChange.getValue() == null) {
-        // NOTE: as of now we cannot detect conflict between delete whole row and row's column value change.
```
anew: Interesting note gets deleted... Tephra recently added support for row deletes (are we using that yet?). How does conflict detection work in that case? When conflict detection is at column level but the delete is at row level, how do we detect conflicts?

anew:
@abaranau please see my two comments.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
abaranau: we are not doing row deletes

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
abaranau: > I also find the old code more intuitive.

but not correct :)

abaranau:
Please see changes

anew:
Looks much better now. I still don't quite understand how increment and deletes interact. If one transactions increments a cell, and another deletes the same cell, aren't they in conflict? With this implementation, the conflict would not be detected,

abaranau:
Heh. It looks like it may not be good to remove increments from conflict detection. The level of conflict detection that we provide is "if two concurrent tx modified the same item only one will succeed". App logic may rely on this to e.g. make sure some invariant remains after ops are performed: e.g. account balance will not go over the credit limit. If app logic does withdraw using "readless increments" it expects that the invariant will not be broken as the logic checks the number and then does increment.

So, we can't actually exclude increments from tx conflict detection without changing the guarantees we provide (level of conflict detection). Instead, what we can say is you can safely use ConflictLevel.NONE with readless increments - it will not cause an error in counting (even if same counter is changed by two concurrent transactions).

I'll think more and this will probably result in docs change.

anew:
I do think that it makes sense to optimize conflict detection for readless increments. One increments cannot conflict with another increment in another transaction, because they are commutatitve and associative, the order in which they are applied does not matter. 

Perhaps this will require a paradigm shift in Tephra, Instead of simply listing a set of changes, a transaction could list different types of changes: ones that are in conflict with anything else, or ones that can co-occur with the same type of change but not with other types. Similar to the multi-read/single-wrote locks in RDBMS.

abaranau:
@anew It's not about whether the order matters or not, it is about relying on "if two concurrent tx modified the same data only one will succeed" guarantee, as I mentioned before.

ghelmling:
On the credit limit example, this is a case where you clearly do care about the current value of the field, and need to read it to enforce the invariant, so it doesn't seem like readless increments are a good match for this usage.  But having a situation where conflict detection behaves differently for readless increment + get operations vs. for normal increment operations seems very confusing from a user standpoint.

On read/write locks, agree that this could be useful for certain scenarios.  I brought up the same idea recently in relation to a Phoenix use case, where a metadata table stores the configured schema which writers must first read then use to write to a data table.  In this case as well, it would be useful for the schema reads to not conflict with each other, but to conflict with any concurrent writes to the metadata table for schema changes.  But we need to consider carefully if going in this direction is worth the additional cost it will impose on conflict detection.

So it makes sense to me to close this PR and add some documentation on when it is appropriate to use a NONE conflict level (assuming we actually expose this in CDAP).

anew:
agreed. Let's close this and write the documentation.

PR #2790, opened 2015-05-28
Revisions to build scripts:

awholegunch:
- cleaned up exit codes on main case statements
- added a message handling facility that writes messages to a tmp file and then plays them back again, at both the end of building a manual, and at the end of the entire run.
- returns 1 if there are any warning messages to indicate a failed run.

- [Example of failed run](http://builds.cask.co/browse/CDAP-DDBD151-1/log)
- [Example of successful run, after fixing MD5 changes](http://builds.cask.co/browse/CDAP-DDBD151-8/log)

https://issues.cask.co/browse/CDAP-2552

Update: it now outputs the Sphinx warnings to a tmp file, and then appends them to the message tmp file. That way, all messages are now visible at the end of the log file...

### DIFF
--- a/cdap-docs/_common/common-build.sh
+++ b/cdap-docs/_common/common-build.sh
@@ -38,6 +38,7 @@ PROJECT="cdap"
 PROJECT_CAPS="CDAP"
 REFERENCE="reference-manual"
 SOURCE="source"
+SPHINX_MESSAGES="build/warnings.txt"
 
 FALSE="false"
 TRUE="true"
@@ -153,7 +154,7 @@ function build_docs() {
   clean
   cd ${SCRIPT_PATH}
   check_includes
-  sphinx-build -b html -d build/doctrees source build/html
+  sphinx-build -w ${SPHINX_MESSAGES} -b html -d build/doctrees source build/html
   display_any_messages
 }
 
@@ -161,7 +162,7 @@ function build_docs_google() {
   clean
   cd ${SCRIPT_PATH}
   check_includes
-  sphinx-build -D googleanalytics_id=$1 -D googleanalytics_enabled=1 -b html -d build/doctrees source build/html
+  sphinx-build -w ${SPHINX_MESSAGES} -D googleanalytics_id=$1 -D googleanalytics_enabled=1 -b html -d build/doctrees source build/html
   display_any_messages
 }
 
@@ -410,19 +411,32 @@ function set_message() {
 }
 
 function display_any_messages() {
-  if [ "x${MESSAGES}" != "x" ]; then
-    echo ""
-    echo_red_bold "Warning Messages for \"${MANUAL}\":"
-    echo ""
-    echo -e "${MESSAGES}"
-    echo ""
+  if [[ "x${MESSAGES}" != "x" || -s ${SPHINX_MESSAGES} ]]; then
+    local m="Warning Messages for \"${MANUAL}\":"
+    if [ "x${MESSAGES}" != "x" ]; then
+      echo ""
+      echo_red_bold "${m}"
+      echo ""
+      echo -e "${MESSAGES}"
+    fi
+    if [ -s ${SPHINX_MESSAGES} ]; then
+      m="Sphinx ${m}"
+      echo ""
+      echo_red_bold "${m}"
+      echo ${m} >> ${TMP_MESSAGES_FILE}
+      cat ${SPHINX_MESSAGES} | while read line
+      do
+        echo ${line}
+        echo ${line} >> ${TMP_MESSAGES_FILE}
+      done
+    fi
   fi
 }
 
 function display_messages_file() {
   if [[ "x${TMP_MESSAGES_FILE}" != "x" && -a ${TMP_MESSAGES_FILE} ]]; then
     echo ""
-    echo "Warning Messages:"
+    echo "Warning Messages: ${TMP_MESSAGES_FILE}"
     echo ""
     cat ${TMP_MESSAGES_FILE} | while read line
     do
--- a/cdap-docs/_common/common-build.sh
+++ b/cdap-docs/_common/common-build.sh
@@ -396,12 +396,12 @@ function clear_messages() {
 
 function set_message() {
   if [ "x${MESSAGES}" == "x" ]; then
-    MESSAGES="${1}"
+    MESSAGES=${*}
   else
-    MESSAGES="${MESSAGES}\n\n${1}"
+    MESSAGES="${MESSAGES}\n\n${*}"
   fi
   if [ "x${TMP_MESSAGES_FILE}" != "x" ]; then
-    local clean_m=`echo_clean_colors "${1}"`
+    local clean_m=`echo_clean_colors "${*}"`
     if [ -e ${TMP_MESSAGES_FILE} ]; then
       echo >> ${TMP_MESSAGES_FILE}
     fi
--- a/cdap-docs/_common/common-build.sh
+++ b/cdap-docs/_common/common-build.sh
@@ -93,6 +93,13 @@ function echo_red_bold() {
   echo -e "${RED}${BOLD}${1}${NC}${2}"
 }
 
+function echo_clean_colors() {
+  local c="${1//${RED}/}"
+  c="${c//${BOLD}/}"
+  c="${c//${NC}/}"
+  echo -e "${c}"
+}
+
 # Hash of file with "Not Found"; returned by GitHub
 NOT_FOUND_HASH="9d1ead73e678fa2f51a70a933b0bf017"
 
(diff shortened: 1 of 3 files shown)


cdap-docs/_common/common-build.sh
```diff
@@ -358,7 +367,57 @@ function display_version() {
   echo "PROJECT_SHORT_VERSION: ${PROJECT_SHORT_VERSION}"
   echo "GIT_BRANCH: ${GIT_BRANCH}"
   echo "GIT_BRANCH_TYPE: ${GIT_BRANCH_TYPE}"
-  echo ""
+}
+
+function set_message_file() {
+  TMP_MESSAGE_FILE="/tmp/$(basename $0).$$.tmp"
+  export TMP_MESSAGE_FILE
+}
```
wolf31o2: If you're going to do this, you'll want some form of `cleanup_tmpfiles` function, which gets called unconditionally, before exit, after messages are replayed.

cdap-docs/_common/common-build.sh
```diff
@@ -358,7 +367,57 @@ function display_version() {
...
+function clear_message() {
+  MESSAGE=""
+  set_message_file
+}
+
+function set_message() {
+  if [ "${MESSAGE}" == "" ]; then
+    MESSAGE="${1}"
+  else
+    MESSAGE="${MESSAGE}\n\n${1}"
```
wolf31o2: Change these to `${*}` to print any arguments, not just the first one.

cdap-docs/_common/common-build.sh
```diff
@@ -291,16 +294,22 @@ function test_an_include() {
     new_md5_hash=`md5sum ${target} | awk '{print $1}'`
   fi
   
+  local result_message=""
```
wolf31o2: Personally, I would use an array for `result_message` so you don't constantly have to prepend additional lines with the previous contents. You simply add a new element to the array. When you set_message, you just send the entire array. That, plus printing all arguments to the functions rather than just the first should give you the same functionality, but a little cleaner.

cdap-docs/_common/common-build.sh
```diff
@@ -291,16 +294,22 @@ function test_an_include() {
...
-    echo -e "Old MD5 Hash: ${md5_hash} New MD5 Hash: ${RED}${BOLD}${new_md5_hash}${NC}" 
+    result_message="${WARNING} ${RED}${BOLD}${file_name} not found!${NC}"  
+    result_message="${result_message}\nfile: ${target}"  
+  elif [[ "${new_md5_hash}" != "${md5_hash}" ]]; then
+    result_message="${WARNING} ${RED}${BOLD}${file_name} has changed! Compare files and update hash!${NC}"   
+    result_message="${result_message}\nfile: ${target}"   
+    result_message="${result_message}\nOld MD5 Hash: ${md5_hash} New MD5 Hash: ${RED}${BOLD}${new_md5_hash}${NC}"   
+  fi
+  if [ "${result_message}" != "" ]; then 
+    set_message "${result_message}"
```
wolf31o2: This would be `set_message ${result_message[*]}`

cdap-docs/_common/common-build.sh
wolf31o2: Actually, it may be better to just keep the input as an array and append it to the previous array, assuming you make `MESSAGE` an array, versus a string as well as `result_message` above.

cdap-docs/_common/common-build.sh
```diff
@@ -358,7 +367,57 @@ function display_version() {
...
+    if [ -a ${TMP_MESSAGE_FILE} ]; then
+      echo -e "\n${warning_message}" >> ${TMP_MESSAGE_FILE}
+    else
+      echo -e "${warning_message}" >> ${TMP_MESSAGE_FILE}
+    fi
+  fi
+}
+
+function display_any_messages() {
+  if [ "${MESSAGE}" != "" ]; then
```
wolf31o2: Use `-n` if testing a string, or `${#MESSAGE[@]} -ne 0` if an array

cdap-docs/_common/common-build.sh
awholegunch: Added.

cdap-docs/_common/common-build.sh
awholegunch: Problem is that this not really an array, but a series of messages, with different line spacing between them. They are grouped together, with extra lines between groups. If I switch to an array, I loose the ease of control over that, which to me is more important since the whole point of this is a log file that's easy to read.

cdap-docs/_common/common-build.sh
```diff
@@ -358,7 +377,63 @@ function display_version() {
...
+
+function set_messages_file() {
+  TMP_MESSAGES_FILE="/tmp/$(basename $0).$$.tmp"
+  export TMP_MESSAGES_FILE
+}
+
+function cleanup_messages_file() {
+  if [[ "x${TMP_MESSAGES_FILE}" != "x" && -a ${TMP_MESSAGES_FILE} ]]; then
+    rm -f TMP_MESSAGES_FILE
+  fi
```
wolf31o2: Conditional is completely unnecessary as `rm -f` doesn't error on failure. Also, your variable isn't prepended with a `$` or in braces.

cdap-docs/_common/common-build.sh
```diff
@@ -60,6 +60,7 @@ EOF`
 
 SCRIPT=`basename ${0}`
 SCRIPT_PATH=`pwd`
```
wolf31o2: Just noticed this expects the script is always run from the correct working directory, no?

cdap-docs/_common/common-build.sh
```diff
@@ -358,7 +377,63 @@ function display_version() {
...
+
+function set_message() {
+  if [ "x${MESSAGES}" == "x" ]; then
+    MESSAGES="${1}"
+  else
+    MESSAGES="${MESSAGES}\n\n${1}"
+  fi
+  if [ "x${TMP_MESSAGES_FILE}" != "x" ]; then
+    local clean_m=`echo_clean_colors "${1}"`
+    if [ -a ${TMP_MESSAGES_FILE} ]; then
```
wolf31o2: test `-a` isn't portable

cdap-docs/_common/common-build.sh
```diff
@@ -358,7 +377,63 @@ function display_version() {
...
+      echo -e "\n" >> ${TMP_MESSAGES_FILE}
```
wolf31o2: Why not just `echo >> ${TMP_MESSAGES_FILE}` or even `>> ${TMP_MESSAGES_FILE}`

cdap-docs/_common/common-build.sh
```diff
@@ -358,7 +377,63 @@ function display_version() {
...
+}
+
+function clear_messages() {
+  MESSAGES=""
+  set_messages_file
+}
+
+function set_message() {
+  if [ "x${MESSAGES}" == "x" ]; then
+    MESSAGES="${1}"
```
wolf31o2: There's no reason to force everything into a single positional parameter. Using `${*}` would be my recommendation. Quotes are for strings.

cdap-docs/_common/common-build.sh
```diff
@@ -358,7 +377,63 @@ function display_version() {
...
+}
+
+function cleanup_messages_file() {
+  if [[ "x${TMP_MESSAGES_FILE}" != "x" && -a ${TMP_MESSAGES_FILE} ]]; then
+    rm -f TMP_MESSAGES_FILE
+  fi
+}
+
+function clear_messages() {
+  MESSAGES=""
```
wolf31o2: `MESSAGES=` is sufficient

cdap-docs/_common/common-build.sh
wolf31o2: I don't see the spacing you're referring to in the code. Where is it?

cdap-docs/_common/common-build.sh
wolf31o2: Do you care if MESSAGE has contents or if it's set? If only contents, use `-n` here.

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh
awholegunch: This is true. I keep forgetting to address this.

cdap-docs/_common/common-build.sh
awholegunch: Excuse my ignorance, but I don't understand what you mean here. Can you elaborate?

cdap-docs/_common/common-build.sh
awholegunch: Added a *check_starting_directory* function to **build.sh** to check and see that it is started from the correct directory.
Lower-level build scripts already fail if they are not started from the correct directory.

cdap-docs/_common/common-build.sh
wolf31o2: Sure. Change this line to `MESSAGES=${*}` and it'll take all arguments to `set_message`, not just the first one. You'll need to do the same 2 lines below.

Now, you can `set_message "some message"` **or** `set_message some message` and they'll be treated the same. Otherwise, you **must** enclose the arguments to `set_message` in quotes, when it's not really necessary.

awholegunch: Thanks for the explanation. Fixed.

cdap-docs/_common/common-build.sh
```diff
@@ -37,6 +38,7 @@ PROJECT="cdap"
 PROJECT_CAPS="CDAP"
 REFERENCE="reference-manual"
 SOURCE="source"
+SPHINX_MESSAGES="build/warnings.txt"
```
wolf31o2: When is this file cleared out? Won't it continually grow?

awholegunch: As it is written to the build directory, it gets removed at the start of each run. See the function *clean*:
```
rm -rf ${SCRIPT_PATH}/${BUILD}
```
(I figured that was a good place to write it, as it always gets deleted at the start.)

wolf31o2:
:+1: LGTM

PR #2792, opened 2015-05-28
fix truncate and json false

elfenheart:
- truncate having errors on datasets and streams
- providing options to specify json or interval on a $resource request


ajainarayanan:
:+1: LGTM

PR #2793, opened 2015-05-28
CDAP-2355: Fix adaper CLI descriptions

jwang47:
https://issues.cask.co/browse/CDAP-2355

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/DescribeAdapterCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/DescribeAdapterCommand.java
@@ -19,6 +19,8 @@ package co.cask.cdap.cli.command.adapter;
 import co.cask.cdap.cli.ArgumentName;
 import co.cask.cdap.cli.CLIConfig;
 import co.cask.cdap.cli.ElementType;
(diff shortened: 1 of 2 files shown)


cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/DescribeAdapterCommand.java
```diff
@@ -81,6 +83,7 @@ public String getPattern() {
 
   @Override
   public String getDescription() {
-    return String.format("Lists all %s.", ElementType.ADAPTER.getTitleNamePlural());
+    return String.format("Shows information about %s.",
+                         Fragment.of(Article.A, ElementType.ADAPTER.getTitleNamePlural()));
```
bdmogal: this shouldn't be plural, right? or am I missing something?

bdmogal:
1 comment. LGTM otherwise

cdap-cli/src/main/java/co/cask/cdap/cli/command/adapter/DescribeAdapterCommand.java
jwang47: Oops. Will fix.

PR #2794, opened 2015-05-29
Fix/ui adapter select onblur

ajainarayanan:
- Adds the ability to type in adapter name textbox and avoids the necessary to press TAB or ENTER to save the value
- Refactor adapter create controller to handle any scenario (create/edit an adapter/draft, switch between drafts, publish a new adapter, publish an already created draft)


elfenheart:
:+1: LGTM

ajainarayanan:
Fixing it in another PR with proper fix.

PR #2798, opened 2015-05-29
Simplify ProgramClient methods (remove code duplication) and Integrat…

anwar6953:
…ionTestBase methods.

http://builds.cask.co/browse/CDAP-DUT1931-3

### DIFF
--- a/cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
+++ b/cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
@@ -30,6 +30,7 @@ import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.Instances;
 import co.cask.cdap.proto.ProgramLiveInfo;
 import co.cask.cdap.proto.ProgramRecord;
+import co.cask.cdap.proto.ProgramRunStatus;
 import co.cask.cdap.proto.ProgramStatus;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.RunRecord;
@@ -76,9 +77,11 @@ public class ProgramClient {
   }
 
   public ProgramClient(ClientConfig config) {
-    this.config = config;
-    this.restClient = new RESTClient(config);
-    this.applicationClient = new ApplicationClient(config, restClient);
+    this(config, new RESTClient(config));
+  }
+
+  public ProgramClient(ClientConfig config, RESTClient restClient) {
+    this(config, restClient, new ApplicationClient(config, restClient));
   }
 
   /**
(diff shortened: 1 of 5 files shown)


cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
```diff
@@ -416,42 +419,6 @@ public void setServiceInstances(String appId, String serviceId, int instances)
...
-   * @param appId ID of the application that the program belongs to
-   * @param programType type of the program
-   * @param programId ID of the program
-   * @param state - filter by status of the program
-   * @return the run records of the program
-   * @throws IOException if a network error occurred
-   * @throws NotFoundException if the application or program could not be found
-   * @throws UnauthorizedException if the request is not authorized successfully in the gateway server
-   */
-  public List<RunRecord> getProgramRuns(String appId, ProgramType programType, String programId, String state,
```
anwar6953: This method is simplified on line 496.

cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
```diff
@@ -75,10 +76,12 @@ public ProgramClient(ClientConfig config, RESTClient restClient, ApplicationClie
     this.applicationClient = applicationClient;
   }
 
+  public ProgramClient(ClientConfig config, RESTClient restClient) {
+    this(config, restClient, new ApplicationClient(config, restClient));
+  }
+
   public ProgramClient(ClientConfig config) {
```
bdmogal: nit: can we make this the first constructor, the one on line 79 the second, and the first one here the third. Just makes it easier to read.

cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
```diff
@@ -485,24 +452,25 @@ public void setServiceInstances(String appId, String serviceId, int instances)
...
    * @param programId ID of the program
+   * @param state - filter by status of the program
    * @return the run records of the program
    * @throws IOException if a network error occurred
    * @throws NotFoundException if the application or program could not be found
    * @throws UnauthorizedException if the request is not authorized successfully in the gateway server
    */
-  public List<RunRecord> getAllProgramRuns(String appId, ProgramType programType, String programId,
-                                           long startTime, long endTime, int limit)
+  public List<RunRecord> getProgramRuns(String appId, ProgramType programType, String programId, String state,
```
bdmogal: Should perhaps use ``ProgramRunStatus`` instead of ``String`` for ``state``, but since you brought up that that would be an API change, maybe later.

Similar comment about using ``Id.Program`` instead of the first 3 parameters.

bdmogal:
just a couple comments.

bdmogal:
LGTM otherwise

cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
anwar6953: Hmm, I swapped the order of the bottom two, but I left the top one on the top, to remain consistent with the other clients' inject method being the topmost.

cdap-client/src/main/java/co/cask/cdap/client/ProgramClient.java
anwar6953: Opened this for now.
https://issues.cask.co/browse/CDAP-2620

PR #2801, opened 2015-06-01
[CDAP-1926] streams endpoint accepts now-xs format

jayfjin:
Build -http://builds.cask.co/browse/CDAP-DUT1934-2

JIRA - https://issues.cask.co/browse/CDAP-1926

Streams endpoint now accepts times formatted as "now-xs".

### DIFF
--- a/cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
+++ b/cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
@@ -87,7 +87,7 @@ public class StreamClient {
    * Gets the configuration of a stream.
    *
    * @param streamId ID of the stream
-   * @throws IOException if a network error occurred
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream was not found
    */
   public StreamProperties getConfig(String streamId)
@@ -106,11 +106,11 @@ public class StreamClient {
   /**
    * Sets properties of a stream.
    *
-   * @param streamId ID of the stream
+   * @param streamId   ID of the stream
    * @param properties properties to set
-   * @throws IOException if a network error occurred
-   * @throws UnauthorizedException if the client is unauthorized
-   * @throws BadRequestException if the request is bad
+   * @throws IOException             if a network error occurred
+   * @throws UnauthorizedException   if the client is unauthorized
+   * @throws BadRequestException     if the request is bad
    * @throws StreamNotFoundException if the stream was not found
    */
   public void setStreamProperties(String streamId, StreamProperties properties)
@@ -135,7 +135,7 @@ public class StreamClient {
    * Creates a stream.
    *
    * @param newStreamId ID of the new stream to create
-   * @throws IOException if a network error occurred
+   * @throws IOException         if a network error occurred
    * @throws BadRequestException if the provided stream ID was invalid
    */
   public void create(String newStreamId) throws IOException, BadRequestException, UnauthorizedException {
@@ -151,12 +151,12 @@ public class StreamClient {
    * Sends an event to a stream.
    *
    * @param streamId ID of the stream
-   * @param event event to send to the stream
-   * @throws IOException if a network error occurred
+   * @param event    event to send to the stream
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream with the specified ID was not found
    */
   public void sendEvent(String streamId, String event) throws IOException,
-                                                              StreamNotFoundException, UnauthorizedException {
+    StreamNotFoundException, UnauthorizedException {
     writeEvent(config.resolveNamespacedURLV3(String.format("streams/%s", streamId)), streamId, event);
   }
 
@@ -165,22 +165,22 @@ public class StreamClient {
    * the event has been received by the server, but may not get persisted.
    *
    * @param streamId ID of the stream
-   * @param event event to send to the stream
-   * @throws IOException if a network error occurred
+   * @param event    event to send to the stream
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream with the specified ID was not found
    */
   public void asyncSendEvent(String streamId, String event) throws IOException,
-                                                                   StreamNotFoundException, UnauthorizedException {
+    StreamNotFoundException, UnauthorizedException {
     writeEvent(config.resolveNamespacedURLV3(String.format("streams/%s/async", streamId)), streamId, event);
   }
 
   /**
    * Sends a file of the given content type to a stream batch endpoint.
    *
-   * @param streamId ID of the stream
+   * @param streamId    ID of the stream
    * @param contentType content type of the file
-   * @param file the file to upload
-   * @throws IOException if a network error occurred
+   * @param file        the file to upload
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream with the specified ID was not found
    */
   public void sendFile(String streamId, String contentType,
@@ -191,10 +191,10 @@ public class StreamClient {
   /**
    * Sends a batch request to a stream batch endpoint.
    *
-   * @param streamId ID of the stream
-   * @param contentType content type of the data
+   * @param streamId      ID of the stream
+   * @param contentType   content type of the data
    * @param inputSupplier provides content for the batch request
-   * @throws IOException if a network error occurred
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream with the specified ID was not found
    */
   public void sendBatch(String streamId, String contentType,
@@ -216,7 +216,7 @@ public class StreamClient {
    * Truncates a stream, deleting all stream events belonging to the stream.
    *
    * @param streamId ID of the stream to truncate
-   * @throws IOException if a network error occurred
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream with the specified name was not found
    */
   public void truncate(String streamId) throws IOException, StreamNotFoundException, UnauthorizedException {
@@ -233,7 +233,7 @@ public class StreamClient {
    * Deletes a stream.
    *
    * @param streamId ID of the stream to truncate
-   * @throws IOException if a network error occurred
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream with the specified name was not found
    */
   public void delete(String streamId) throws IOException, StreamNotFoundException, UnauthorizedException {
@@ -249,9 +249,9 @@ public class StreamClient {
   /**
    * Sets the Time-to-Live (TTL) of a stream. TTL governs how long stream events are readable.
    *
-   * @param streamId ID of the stream
+   * @param streamId     ID of the stream
    * @param ttlInSeconds desired TTL, in seconds
-   * @throws IOException if a network error occurred
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream with the specified name was not found
    */
   public void setTTL(String streamId, long ttlInSeconds)
@@ -276,20 +276,21 @@ public class StreamClient {
   public List<StreamDetail> list() throws IOException, UnauthorizedException {
     URL url = config.resolveNamespacedURLV3("streams");
     HttpResponse response = restClient.execute(HttpMethod.GET, url, config.getAccessToken());
-    return ObjectResponse.fromJsonBody(response, new TypeToken<List<StreamDetail>>() { }).getResponseObject();
+    return ObjectResponse.fromJsonBody(response, new TypeToken<List<StreamDetail>>() {
+    }).getResponseObject();
   }
 
   /**
    * Reads events from a stream
    *
-   * @param streamId ID of the stream
+   * @param streamId  ID of the stream
    * @param startTime Timestamp in milliseconds to start reading event from (inclusive)
-   * @param endTime Timestamp in milliseconds for the last event to read (exclusive)
-   * @param limit Maximum number of events to read
-   * @param results Collection for storing the resulting stream events
-   * @param <T> Type of the collection for storing results
+   * @param endTime   Timestamp in milliseconds for the last event to read (exclusive)
+   * @param limit     Maximum number of events to read
+   * @param results   Collection for storing the resulting stream events
+   * @param <T>       Type of the collection for storing results
    * @return The same collection object as passed in the {@code results} parameter
-   * @throws IOException If fails to read from stream
+   * @throws IOException             If fails to read from stream
    * @throws StreamNotFoundException If the given stream does not exists
    */
   public <T extends Collection<? super StreamEvent>> T getEvents(String streamId, String startTime,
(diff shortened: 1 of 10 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamFetchHandler.java
```diff
@@ -107,9 +108,24 @@ public StreamFetchHandler(CConfiguration cConf, Authenticator authenticator,
...
+                    @QueryParam("start") String start,
+                    @QueryParam("end") @DefaultValue("9223372036854775807") String end,
                     @QueryParam("limit") @DefaultValue("2147483647") int limit) throws Exception {
+    long startTime;
+    long endTime;
+    try {
+      /* if string is numeric then value is in milliseconds */
+      startTime = Long.parseLong(start);
+    } catch (NumberFormatException nfe) {
+      /* else TimeMathParser returns a value in seconds and we must convert to milliseconds */
```
henry-cask: I think the comment is incorrect. Should indicate that it could be in now or now-Xs format

cdap-test/src/main/java/co/cask/cdap/test/StreamManager.java
```diff
@@ -108,10 +108,10 @@
...
    *
-   * @param startTime the start time
-   * @param endTime the end time
+   * @param startTime the start time in seconds or "now-xs" format
+   * @param endTime the end time in seconds or "now-xs" format
    * @param limit the maximum number of events to return
    * @return a list of stream events in the given time range
    */
-  List<StreamEvent> getEvents(long startTime, long endTime, int limit) throws IOException;
+  List<StreamEvent> getEvents(String startTime, String endTime, int limit) throws IOException;
```
bdmogal: This is an API change. We should probably deprecate the existing method for a couple of releases, not remove it instantly.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamFetchHandler.java
```diff
@@ -107,9 +108,24 @@ public StreamFetchHandler(CConfiguration cConf, Authenticator authenticator,
...
-                    @QueryParam("start") long startTime,
-                    @QueryParam("end") @DefaultValue("9223372036854775807") long endTime,
+                    @QueryParam("start") String start,
+                    @QueryParam("end") @DefaultValue("9223372036854775807") String end,
                     @QueryParam("limit") @DefaultValue("2147483647") int limit) throws Exception {
+    long startTime;
+    long endTime;
+    try {
+      /* if string is numeric then value is in milliseconds */
+      startTime = Long.parseLong(start);
```
anwar6953: I don't think this is necessary. Take a look at the other usage of the `TimeMathParser#parseTime` method.
I believe it internally takes care of whether its in milliseconds or if its a string like `now-50s`.
For instance, take a look at: TimeMathParser.java, line 93-95.

anwar6953: Or even like how you're doing it in `StreamManager`, simply relying on the method and not attempting to parse it as a number yourself.
https://github.com/caskdata/cdap/pull/2801/files#diff-4f60b022b87728f1ceafd570b288e68eR161

jayfjin: In TimerMathParser if it sees it's a timestamp that's numeric doesn't it assume that it it's in seconds (on line 93)?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamFetchHandler.java
```diff
@@ -107,9 +108,24 @@ public StreamFetchHandler(CConfiguration cConf, Authenticator authenticator,
...
+    } catch (NumberFormatException nfe) {
+      /* else TimeMathParser returns a value in seconds and we must convert to milliseconds */
+      startTime = TimeUnit.SECONDS.toMillis(TimeMathParser.parseTime(start));
```
chtyim: I think it's better to have this `try-catch` logic as part of `TimeMathParser.parseTime`. Also, it's better to have a overload one that takes a `TimeUnit` so that the return time is in that given unit (that is Java idiomatic way as well).

`long startTime = TimeMathParser.parseTime(start, TimeUnit.MILLISECONDS);`

cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -310,15 +312,45 @@ public Boolean apply(StreamEvent input) {
...
    *                 upon invocation, it will stops the reading
    * @throws IOException If fails to read from stream
    * @throws StreamNotFoundException If the given stream does not exists
    */
-  public void getEvents(String streamId, long startTime, long endTime, int limit,
+  public void getEvents(String streamId, String start, String end, int limit,
                         Function<? super StreamEvent, Boolean> callback)
     throws IOException, StreamNotFoundException, UnauthorizedException {
 
+    long startTime = TimeMathParser.parseTime(start, TimeUnit.MILLISECONDS);
```
anwar6953: is this time-parsing necessary in the client? (since its already being done in the handler).

jayfjin: Yes since before this if it was a numeric string it would be in milliseconds but parseTime would return a value in seconds if the string is in "now-xs" format

cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java
```diff
@@ -80,19 +81,27 @@ public static int resolutionInSeconds(String resolutionStr) {
...
 
     if (NOW.equals(timeStr.toUpperCase())) {
       return now;
     }
-    // if its a timestamp in seconds
+    // if its a numeric timestamp
     Matcher matcher = TIMESTAMP_PATTERN.matcher(timeStr);
     if (matcher.matches()) {
-      return Integer.parseInt(timeStr);
+      return Long.parseLong(timeStr);
```
chtyim: Why it ignores the time unit? Does it mean the string representation of the numeric value should be in the time unit as being passed in? Can you describe it in the javadoc of the method?

cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java
```diff
@@ -80,19 +81,27 @@ public static int resolutionInSeconds(String resolutionStr) {
   }
 
   public static long parseTime(String timeStr) {
```
chtyim: All public methods should have javadoc. I don't know why the existing code get that slipped. Anyway, when seeing it is missing, please add it.

cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -310,15 +312,45 @@ public Boolean apply(StreamEvent input) {
...
+   * @param <T> Type of the collection for storing results
+   * @return The same collection object as passed in the {@code results} parameter
+   * @throws IOException If fails to read from stream
+   * @throws StreamNotFoundException If the given stream does not exists
+   */
+  @Deprecated
+  public <T extends Collection<? super StreamEvent>> T getEvents(String streamId, long startTime,
+                                                                 long endTime, int limit, final T results)
+    throws IOException, StreamNotFoundException, UnauthorizedException {
+    getEvents(streamId, startTime, endTime, limit, new Function<StreamEvent, Boolean>() {
```
chtyim: Shouldn't this one just call the other `getEvents(String, String, String, int, T)` (the one above)?

cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java
```diff
@@ -79,20 +83,53 @@ public static int resolutionInSeconds(String resolutionStr) {
     return output;
   }
 
+  /** parses a time in String format into a long value, assuming the input is in seconds if numeric
```
chtyim: Capitalize the first character and you need a full-stop for the first sentence. Also we start javadoc on the second line:

```java
/**
  * Parses ... 
  *
  * @param ...
  */
```

cdap-test/src/main/java/co/cask/cdap/test/StreamManager.java
```diff
@@ -108,10 +108,21 @@
   /**
    * Get events from the specified stream in the specified interval
    *
+   * @param startTime the start time in seconds or "now-xs" format
+   * @param endTime the end time in seconds or "now-xs" format
+   * @param limit the maximum number of events to return
+   * @return a list of stream events in the given time range
+   */
+  List<StreamEvent> getEvents(String startTime, String endTime, int limit) throws IOException;
```
chtyim: The API will be easier to use if it is:

```<T> List<StreamEvent> getEvents(T startTime, T endTime, int limit)```

With the contract as using `T.toString()` as the value for startTime and endTime. This means someone can still call it with `getEvents(0, Long.MAX_VALUE, 10);` rather then `getEvents("0", String.valueOf(Long.MAX_VALUE), 10);`

File: cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java [hunk 11]
```diff
@@ -79,20 +82,57 @@ public static int resolutionInSeconds(String resolutionStr) {
...
+  /**
+   * Parses a time in String format into a long value
+   *
+   * @param now the present time in seconds
+   * @param timeStr the string to parse
+   * @param timeUnit the unit of time to return, if timeStr is numeric then it is assumed to be in the unit timeUnit
+   * @return the parsed time
+   * @throws IllegalArgumentException if the format of timeStr is bad
+   */
+  public static long parseTime(long now, String timeStr, TimeUnit timeUnit) {
```
gsps1: you are adding this method so you can get the time in milliseconds and other time units right?. in the usage i see its mostly used with mill seconds as the timeUnit, but the timeStr string's valid units are only {s | m | h | d}, would it help to allow the parser to include milliseconds ?  - the time math parser was used only with metrics before and that's why seconds is the lowest granularity, for stream we might want to include milli seconds. what do you think?

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteStreamManager.java
```diff
@@ -122,10 +122,10 @@ public void send(Map<String, String> headers, ByteBuffer buffer) throws IOExcept
   }
 
   @Override
-  public List<StreamEvent> getEvents(long startTime, long endTime, int limit) throws IOException {
+  public <T> List<StreamEvent> getEvents(T startTime, T endTime, int limit) throws IOException {
```
anew: this allows to pass in anything? Boolean, Date, Dataset... Why not require a string? Or overload the method for long and string?

chtyim: I suggested this. But agree that it's better to overload than using generics

jayfjin: so I should change this back to what I had before?

cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java
jayfjin: The reason this is necessary is that fetch requires the time be in milliseconds, where as parsing a string like now - 5s returns a time in seconds. Thus, we must either attempt to parse the long beforehand, (which terence was against), or move the logic into parseTime.

cdap-integration-test/src/main/java/co/cask/cdap/test/remote/RemoteStreamManager.java
chtyim: Yes and undeprecate the one that takes long

cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java
gsps1: my question is, would we want to allow request like (now - 300ms) till now ?

jayfjin: Oh good point, I can add that logic in.

cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java
```diff
@@ -49,8 +49,11 @@ private static long convertToSeconds(String op, long num, String unitStr) {
       seconds = TimeUnit.HOURS.toSeconds(num);
     } else if ("d".equals(unitStr)) {
       seconds = TimeUnit.DAYS.toSeconds(num);
+    } else if ("ms".equals(unitStr)) {
+      seconds = TimeUnit.MILLISECONDS.toSeconds(num);
```
gsps1: this is incorrect,  if a user specifies, (now - 500ms), we round it to the last one second and include additional stream events. you should convert everything to milliseconds resolution, for metrics handler the resolution is still seconds, while the stream handler its in milliseconds, you have to handle both of them appropriately.

gsps1:
can you add some test case coverage in TimeMathParserTest

cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -85,7 +87,7 @@ public StreamClient(ClientConfig config) {
    * Gets the configuration of a stream.
    *
    * @param streamId ID of the stream
-   * @throws IOException if a network error occurred
+   * @throws IOException             if a network error occurred
```
chtyim: Why all the whitespace changes in this class?

cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -163,22 +165,22 @@ public void sendEvent(String streamId, String event) throws IOException,
...
    * @param streamId ID of the stream
-   * @param event event to send to the stream
-   * @throws IOException if a network error occurred
+   * @param event    event to send to the stream
+   * @throws IOException             if a network error occurred
    * @throws StreamNotFoundException if the stream with the specified ID was not found
    */
   public void asyncSendEvent(String streamId, String event) throws IOException,
-                                                                   StreamNotFoundException, UnauthorizedException {
+    StreamNotFoundException, UnauthorizedException {
```
chtyim: This alignment is worse then the original one.

cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
jayfjin: I think when I refactored the code it aligned the descriptions up so that the description of IOException is aligned with the description of StreamNotFoundException

chtyim: That's not the convention we have in the code base. Can you revert those changes in this file?

cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -306,19 +309,41 @@ public Boolean apply(StreamEvent input) {
   /**
    * Reads events from a stream
    *
-   * @param streamId ID of the stream
+   * @param streamId  ID of the stream
```
gsps1: @chtyim  mentioned about removing unwanted whitespace additions like this.

cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java
```diff
@@ -74,25 +80,63 @@ public static int resolutionInSeconds(String resolutionStr) {
...
+
+  /**
+   * Parses a time in String format into a long value. If the input is numeric we assume the input is in seconds.
+   *
+   * @param now the present time in seconds
+   * @param timeStr the string to parse
+   * @return the parsed time in seconds
+   */
   public static long parseTime(long now, String timeStr) {
+    return parseTime(TimeUnit.MILLISECONDS.convert(now, TimeUnit.SECONDS), timeStr, TimeUnit.SECONDS);
```
gsps1: what if now is already in ms ? is that possible ?

jayfjin: the function previously expects seconds so I didn't want to change the API

gsps1: i see you mention its in seconds in api. might be useful to have a check too

cdap-common/src/main/java/co/cask/cdap/common/utils/TimeMathParser.java
```diff
@@ -74,25 +80,63 @@ public static int resolutionInSeconds(String resolutionStr) {
...
 
+  /**
+   * Parses a time in String format into a long value. If the input is numeric we assume the input is in seconds.
+   *
+   * @param timeStr the string to parse
+   * @return the parsed time in seconds
+   */
   public static long parseTime(String timeStr) {
-    return parseTime(nowInSeconds(), timeStr);
+    return parseTime(timeStr, TimeUnit.SECONDS);
```
gsps1: same here

gsps1:
just couple of minor comments, rest LGTM

gsps1:
LGTM :+1

PR #2805, opened 2015-06-01
Feature/cdap 1666 use id program instead of also programtype

anwar6953:
Simply refactor method signatures to resolve the following JIRA.

https://issues.cask.co/browse/CDAP-1666
http://builds.cask.co/browse/CDAP-DUT1938-4


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -386,174 +369,4 @@ protected boolean respondIfElementNotFound(Throwable t, HttpResponder responder)
     }
     return false;
   }
-
-  // TODO: refactor
-  protected final void dataList(HttpResponder responder, Store store, DatasetFramework dsFramework,
```
henry-cask: Just curious why this block of code dissapear? Not used anywhere else?

anwar6953: Correct. It was used before, but then this functionality is replaced by the Datasets Registry concept.
You can see usages of it being removed here, but the method itself was not removed.
https://github.com/caskdata/cdap/pull/2166/files#diff-2b14ed88e3eb39a9c4550dfd17f60b8aL80

henry-cask:
LGTM

+1

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -328,18 +325,16 @@ void addApplication(Id.Application id,
   /**
    * Deletes a schedules from a particular program
    * @param program defines program from which a schedule is being deleted
-   * @param programType defines the type of the program
    * @param scheduleName the name of the schedule to be removed from the program
    */
-  void deleteSchedule(Id.Program program, SchedulableProgramType programType, String scheduleName);
```
gsps1: question: why do we need SchedulableProgramType ? i noticed the programType includes all of the types in SchedulableProgramType

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/program/ProgramBundle.java
```diff
@@ -57,15 +56,13 @@ public boolean apply(JarEntry input) {
...
+  public static Location create(Id.Program programId, ArchiveBundler bundler, Location output,
+                                String className, ApplicationSpecification appSpec) throws IOException {
+    return create(programId, bundler, output, className, appSpec, null);
   }
 
-  public static Location create(Id.Application id, ArchiveBundler bundler, Location output, String programName,
-                               String className, ProgramType type, ApplicationSpecification appSpec,
-                               Manifest other) throws IOException {
+  public static Location create(Id.Program programId, ArchiveBundler bundler, Location output,
+                               String className, ApplicationSpecification appSpec, Manifest other) throws IOException {
```
gsps1: add nullable for manifest ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -245,8 +245,8 @@ private void deleteProgramLocations(Id.Application appId) throws IOException {
     try {
       Id.Program programId = Id.Program.from(appId.getNamespaceId(), appId.getId(),
                                              ProgramType.WEBAPP, ProgramType.WEBAPP.name().toLowerCase());
-      Location location = Programs.programLocation(namespacedLocationFactory, appFabricDir, programId,
-                                                   ProgramType.WEBAPP);
+      Location location = Programs.programLocation(namespacedLocationFactory, appFabricDir, programId
+      );
```
gsps1: formatting: move this to previous line

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
anwar6953: Not all `ProgramType`s can be passed to scheduler APIs, and by `SchedulableProgramType` being a subset of `ProgramType`, this helps restrict which `ProgramType`s are allowed to be scheduled/unscheduled.

gsps1: okay, just wanted to clarify.

gsps1:
LGTM :+1:

PR #2800, opened 2015-06-01
Adapter code re-organization.

ajainarayanan:
- Re-organizes code around adapter create controller
- Adopts controller-as syntax for more clarity on usage of controllers in template (nested controller)
- Modifies add/edit behavior of plugins (no '+' or 'edit' buttons)

PR #2806, opened 2015-06-01
adds cursor and clickability to name and edit icons

astewes:


elfenheart:
@astewes this will get rewritten with PR #2800

PR #2808, opened 2015-06-02
[CDAP-2537] Fixed exists() condition in FileStreamAdmin to also make …

bdmogal:
…sure that the stream exists in the metastore along with checking that the stream config exists on the filesystem

Issue: [CDAP-2537](https://issues.cask.co/browse/CDAP-2537)
Analysis:
- ``streamAdmin.dropAllInNamespace()`` leaves behind the ``config.json`` file on the filesystem
- ``FileStreamAdmin.exists()`` however, only checks that ``config.json`` exists and doesn't make sure that the stream exists in the metastore as well. This PR adds that check, along with some minor test cleanup.
- This has been fixed via #2266 on develop already

Build: http://builds.cask.co/browse/CDAP-RBT297

### DIFF
--- a/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/transaction/stream/StreamAdminTest.java
+++ b/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/transaction/stream/StreamAdminTest.java
@@ -65,6 +65,12 @@ public abstract class StreamAdminTest {
 
     streamAdmin.create(otherStreamId);
     Assert.assertTrue(streamAdmin.exists(otherStreamId));
+
+    streamAdmin.drop(streamId);
+    Assert.assertFalse(streamAdmin.exists(streamId));
+
+    streamAdmin.drop(otherStreamId);
+    Assert.assertFalse(streamAdmin.exists(otherStreamId));
   }
 
   @Test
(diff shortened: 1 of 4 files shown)


cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/transaction/stream/StreamAdminTest.java
```diff
@@ -65,6 +65,12 @@ public void testCreateExist() throws Exception {
 
     streamAdmin.create(otherStreamId);
     Assert.assertTrue(streamAdmin.exists(otherStreamId));
+
+    streamAdmin.drop(streamId);
+    Assert.assertFalse(streamAdmin.exists(streamId));
```
anwar6953: So just to clarify, this test case would be failing before the change to FileStreamAdmin in this PR?

chtyim:
Can you just backport the changes in #2266 ?

cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/transaction/stream/StreamAdminTest.java
bdmogal: Unfortunately, no. I couldn't reproduce this failure in a unit/integration test. I added this stuff here just for completeness.

bdmogal:
@chtyim I could, but I wasn't sure we want that entire delete stream feature in 3.0.1, just for this bugfix.

chtyim:
Why not? And It includes fixes for doing truncate and delete properly

bdmogal:
Hmmm, primarily because I thought this was a bugfix release and delete stream was a feature. But I could port it. Much easier. I'll close this then. :)

PR #2809, opened 2015-06-02
[CDAP-2629] Workflow Diagrams Improvements

elfenheart:
Action nodes now polling for the particular status:
- Running node will show Green outline
- Completed node will show Green background
- Failed node will show Red background

Modified the color of condition node

screen shot 2015-06-02 at 1 06 28 pm


ajainarayanan:
:+1: LGTM

PR #2814, opened 2015-06-02
Utilize the config in the json of the metadata upon namespacece create

anwar6953:
Utilize the config in the json of the metadata upon namespace create.
Also, ignore the `name` field of the metadata upon namespace properties update.

https://issues.cask.co/browse/CDAP-2635
http://builds.cask.co/browse/CDAP-DUT1948-6

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
@@ -167,6 +167,25 @@ public class NamespaceHttpHandlerTest extends AppFabricTestBase {
   }
 
   @Test
+  public void testCreateWithConfig() throws Exception {
+    // prepare - create namespace with config in its properties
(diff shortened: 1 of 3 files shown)


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
```diff
@@ -167,6 +167,23 @@ public void testCreateDuplicate() throws Exception {
   }
 
   @Test
+  public void testCreateWithConfig() throws Exception {
```
bdmogal: Should we also add a check that verifies that the name from the json meta is not used in the update properties API?

bdmogal:
LGTM

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
anwar6953: Added here:
https://github.com/caskdata/cdap/pull/2814/files#diff-02fb0e9e2123afe3920f7f5aa83492d9R424

PR #2815, opened 2015-06-02
Feature/metrics client api changes

gsps1:
changed metrics client api to support basic query (single metric in a context) and changed the api to accept map for time range parameters. increased test coverage.
JIRA : https://issues.cask.co/browse/CDAP-2588
Build : http://builds.cask.co/browse/CDAP-DUT1947

### DIFF
--- a/cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
+++ b/cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
@@ -71,8 +71,7 @@ public class MetricsClientTestRun extends ClientTestBase {
       String flowlet = FakeFlow.FLOWLET_NAME;
 
       MetricQueryResult result = metricsClient.query(
-        ImmutableList.of(Constants.Metrics.Name.Flow.FLOWLET_INPUT),
-        ImmutableList.<String>of(),
+        Constants.Metrics.Name.Flow.FLOWLET_INPUT,
         MetricsTags.flowlet(programId, flowlet));
       Assert.assertEquals(1, result.getSeries()[0].getData()[0].getValue());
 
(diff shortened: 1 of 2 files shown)


cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -135,9 +138,7 @@ public MetricQueryResult query(List<String> metrics, List<String> groupBys, Map<
    */
   // TODO: take in query object shared by MetricsHandler
   public MetricQueryResult query(List<String> metrics, List<String> groupBys, Map<String, String> tags,
```
anwar6953: can we swap position of `tags` and `groupBys`, simply because `tags` would be used more frequently?

anwar6953:
remainder lgtm

cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
```diff
@@ -70,10 +71,18 @@ public void testAll() throws Exception {
...
-        ImmutableList.<String>of(),
-        MetricsTags.flowlet(programId, flowlet));
+      MetricQueryResult result = metricsClient.query(MetricsTags.flowlet(programId, flowlet),
+                                                     Constants.Metrics.Name.Flow.FLOWLET_INPUT);
+      Assert.assertEquals(1, result.getSeries()[0].getData()[0].getValue());
+
+      result = metricsClient.query(MetricsTags.flowlet(programId, flowlet),
+                                   ImmutableList.of(Constants.Metrics.Name.Flow.FLOWLET_INPUT),
+                                   ImmutableList.<String>of(), ImmutableMap.of("aggregate", "true"));
+      Assert.assertEquals(1, result.getSeries()[0].getData()[0].getValue());
```
anwar6953: somewhere, we should have a utility like this:
https://github.com/caskdata/cdap-integration-tests/blob/develop/integration-test-remote/src/test/java/co/cask/cdap/apps/AudiTestBase.java#L136

Otherwise, accessing the aggregate value like this is messy.

anwar6953: I think it'd be fine to merge this PR, but leave this JIRA open, so we can consider this sort of utility as a part of the JIRA.
https://issues.cask.co/browse/CDAP-2588

anwar6953:
LGTM

PR #2817, opened 2015-06-03
Feature/cdap 2535 fix deploy integration test manager

rohanahata:
Feature makes sure that deploy application is now making use of namespace that is passed as function parameter
Build : http://builds.cask.co/browse/CDAP-DUT1944-2
JIRA : https://issues.cask.co/browse/CDAP-2535

### DIFF
--- a/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
+++ b/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
@@ -80,12 +80,23 @@ public class IntegrationTestBase {
     assertIsClear();
   }
 
-  protected TestManager getTestManager() {
+  protected ClientConfig createNamespacedClientConfig(ClientConfig initialClientConfig, Id.Namespace namespace) {
+    ClientConfig newClientConfig = new ClientConfig.Builder(initialClientConfig).build();
+    newClientConfig.setNamespace(namespace);
+    return newClientConfig;
+  }
+
+  protected TestManager createTestManager(Id.Namespace namespace) {
     try {
-      return new IntegrationTestManager(getClientConfig(), new LocalLocationFactory(TEMP_FOLDER.newFolder()));
+      return new IntegrationTestManager(createNamespacedClientConfig(getClientConfig(), namespace),
+                                        new LocalLocationFactory(TEMP_FOLDER.newFolder()));
     } catch (IOException e) {
       throw Throwables.propagate(e);
     }
+  }
+
+  protected TestManager getTestManager() {
+    return createTestManager(Id.Namespace.DEFAULT);
   }
 
   /**
@@ -154,10 +165,17 @@ public class IntegrationTestBase {
     return new DatasetClient(getClientConfig());
   }
 
+  protected Id.Namespace createNamespace(String name) throws Exception {
+    Id.Namespace namespace = new Id.Namespace(name);
+    NamespaceMeta namespaceMeta = new NamespaceMeta.Builder().setName(namespace).build();
+    getTestManager().createNamespace(namespaceMeta);
+    return namespace;
+  }
+
   protected ApplicationManager deployApplication(Id.Namespace namespace,
                                                  Class<? extends Application> applicationClz,
                                                  File...bundleEmbeddedJars) throws IOException {
-    return getTestManager().deployApplication(namespace, applicationClz, bundleEmbeddedJars);
+    return createTestManager(namespace).deployApplication(namespace, applicationClz, bundleEmbeddedJars);
   }
 
   protected ApplicationManager deployApplication(Class<? extends Application> applicationClz,
--- a/cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
+++ b/cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
@@ -17,6 +17,10 @@
 package co.cask.cdap.test;
 
 import co.cask.cdap.StandaloneTester;
+import co.cask.cdap.client.ApplicationClient;
+import co.cask.cdap.client.config.ClientConfig;
+import co.cask.cdap.proto.Id;
+import org.junit.Assert;
 import org.junit.ClassRule;
 import org.junit.Test;
 
@@ -40,4 +44,21 @@ public class IntegrationTestBaseTest extends IntegrationTestBase {
     flowManager.stop();
   }
 
+  @Test
+  public void testDeployApplicationInNamespace() throws Exception {
+    Id.Namespace namespace = createNamespace("Test1");
+    ClientConfig clientConfig = new ClientConfig.Builder(getClientConfig()).build();
+    clientConfig.setNamespace(namespace);
+    ApplicationManager applicationManager = deployApplication(namespace, TestApplication.class);
+
+    // Check the default namespaces applications to see whether the application wasnt made in the default namespace
+    ClientConfig defaultClientConfig = new ClientConfig.Builder(getClientConfig()).build();
+    Assert.assertEquals(0, new ApplicationClient(defaultClientConfig).list().size());
+
+    ApplicationClient applicationClient = new ApplicationClient(clientConfig);
+    Assert.assertEquals("TestApplication", applicationClient.list().get(0).getName());
+    applicationClient.delete("TestApplication");
+    Assert.assertEquals(0, new ApplicationClient(clientConfig).list().size());
+
+  }
 }


cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
```diff
@@ -17,14 +17,23 @@
...
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Test for {@link IntegrationTestBase}.
  */
 public class IntegrationTestBaseTest extends IntegrationTestBase {
 
+  private static final Logger LOG = LoggerFactory.getLogger(IntegrationTestBaseTest.class);
```
anwar6953: seems unused

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
```diff
@@ -40,4 +49,26 @@ public void testFlowManager() throws Exception {
...
+    Assert.assertEquals("TestApplication", applicationClient.list().get(0).getName());
+    applicationClient.delete("TestApplication");
+    Assert.assertEquals(0, new ApplicationClient(clientConfig).list().size());
+
+    // Check the default namespaces applications to see whether the application wasnt made in the default namespace
+    ClientConfig defaultClientConfig = new ClientConfig.Builder(getClientConfig()).build();
+    defaultClientConfig.setNamespace(Constants.DEFAULT_NAMESPACE_ID);
+    Assert.assertEquals(0, new ApplicationClient(defaultClientConfig).list().size());
+  }
+
```
gokulavasan: remove extra lines

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
```diff
@@ -97,6 +97,11 @@ public ApplicationManager deployApplication(Id.Namespace namespace,
     }
   }
 
+
+  protected ClientConfig getModifiableClientConfig() {
```
anwar6953: whats the purpose of this? Is it being used?

gsps1: unused method, remove

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -96,14 +97,25 @@ public void tearDown() throws Exception {
     // TODO: check metrics, streams, etc.
   }
 
-  protected TestManager getTestManager() {
+  protected ClientConfig getModifiableClientConfig(ClientConfig initialClientConfig, Id.Namespace namespace) {
```
gsps1: rename this to ``getNamespacedClientConfig`` ?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -121,6 +133,7 @@ protected String getAccessToken() {
   }
 
   protected ClientConfig getClientConfig() {
+
```
gsps1: remove unwanted space

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
```diff
@@ -97,6 +97,11 @@ public ApplicationManager deployApplication(Id.Namespace namespace,
     }
   }
 
+
```
gsps1: remove extra space

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
anwar6953: agreed.

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
gsps1: unused variable, remove ?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
bdmogal: perhaps ``createNamespacedClientConfig``, since its creating a new object and returning.

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
```diff
@@ -40,4 +49,26 @@ public void testFlowManager() throws Exception {
...
+
+
```
gsps1: remove unwanted extra space

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
anwar6953: and how about it internally calls `getClientConfig`, instead of requiring user to pass it in?
Maybe call it `getClientConfig(Id.Namespace)`

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
```diff
@@ -40,4 +49,26 @@ public void testFlowManager() throws Exception {
     flowManager.stop();
   }
 
+  @Test
+  public void testDeployApplicationInNamespace() throws Exception {
+    Id.Namespace namespace = new Id.Namespace ("Test1");
+    createNamespace(new NamespaceMeta.Builder().setName(namespace).build());
```
gsps1: would it be better to have createNamespace(String namespace) and have the logic of namespace meta builder construction there ?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
chtyim: So this method exists to replace a one liner?

`new ClientConfig.Builder(otherConfig).setNamespace(namespace).build()`

Seems a bit unnecessary to me.

chtyim: and do we introduce different methods for just calling a method in the builder?

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
```diff
@@ -40,4 +49,26 @@ public void testFlowManager() throws Exception {
...
+    FlowManager flowManager = applicationManager.getFlowManager(TestFlow.NAME).start();
+    flowManager.stop();
+    ApplicationClient applicationClient = new ApplicationClient(clientConfig);
+    Assert.assertEquals("TestApplication", applicationClient.list().get(0).getName());
+    applicationClient.delete("TestApplication");
+    Assert.assertEquals(0, new ApplicationClient(clientConfig).list().size());
+
+    // Check the default namespaces applications to see whether the application wasnt made in the default namespace
+    ClientConfig defaultClientConfig = new ClientConfig.Builder(getClientConfig()).build();
+    defaultClientConfig.setNamespace(Constants.DEFAULT_NAMESPACE_ID);
```
gsps1: you don't need this line, as getClientConfig uses ``default`` namespace by default.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -96,14 +97,25 @@ public void tearDown() throws Exception {
...
   }
 
-  protected TestManager getTestManager() {
+  protected ClientConfig getModifiableClientConfig(ClientConfig initialClientConfig, Id.Namespace namespace) {
+    ClientConfig newClientConfig = new ClientConfig.Builder(initialClientConfig).build();
+    newClientConfig.setNamespace(namespace);
+    return newClientConfig;
+  }
+
+  protected TestManager getTestManager(Id.Namespace namespace) {
```
chtyim: Call it `createTestManager`. For method that create new instance, call it `createXXX`.

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
```diff
@@ -40,4 +49,26 @@ public void testFlowManager() throws Exception {
...
+    Assert.assertEquals(0, new ApplicationClient(defaultClientConfig).list().size());
```
gsps1: i would move this check before you delete the application in ``test`` namespace

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
chtyim: because in general, getter returns the same instance all the time.

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
```diff
@@ -40,4 +49,26 @@ public void testFlowManager() throws Exception {
...
 
+  @Test
+  public void testDeployApplicationInNamespace() throws Exception {
+    Id.Namespace namespace = new Id.Namespace ("Test1");
+    createNamespace(new NamespaceMeta.Builder().setName(namespace).build());
+    ClientConfig clientConfig = new ClientConfig.Builder(getClientConfig()).build();
+    clientConfig.setNamespace(namespace);
+    ApplicationManager applicationManager = deployApplication(namespace, TestApplication.class);
+    FlowManager flowManager = applicationManager.getFlowManager(TestFlow.NAME).start();
+    flowManager.stop();
```
gsps1: no need to start and stop the flow.

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
chtyim: Probably unrelated to this PR. But why `createNamespace` cannot take `Id.Namespace` but requires a `NamespaceMeta` instead?

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
```diff
@@ -40,4 +49,26 @@ public void testFlowManager() throws Exception {
     flowManager.stop();
   }
 
+  @Test
+  public void testDeployApplicationInNamespace() throws Exception {
+    Id.Namespace namespace = new Id.Namespace ("Test1");
+    createNamespace(new NamespaceMeta.Builder().setName(namespace).build());
+    ClientConfig clientConfig = new ClientConfig.Builder(getClientConfig()).build();
+    clientConfig.setNamespace(namespace);
```
chtyim: Seems strange to me to allow a `setNamespace` method if you have the `createNamespacedClientConfig` method above. Shouldn't `ClientConfig` be an immutable class @anwar6953 ?

anwar6953: And then have a setNamespace in the builder of the ClientConfig, instead?

chtyim: yes, otherwise why have builder? The builder pattern is useful to help creating immutable object.

rohanahata: This is going to be addressed in a separate issue CDAP-2638

cdap-integration-test/src/test/java/co/cask/cdap/test/IntegrationTestBaseTest.java
bdmogal: That's so you can pass other fields like ``description`` or ``NamespaceConfig`` for the namespace you are creating. Perhaps we can also add an overloaded field that only takes ``Id.Namespace`` and uses defaults for ``description`` and ``NamespaceConfig``.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
rohanahata: CDAP-2638

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -154,10 +165,24 @@ protected DatasetClient getDatasetClient() {
     return new DatasetClient(getClientConfig());
   }
 
+  protected Id.Namespace createNamespace(String name) throws Exception {
+    Id.Namespace namespace = new Id.Namespace (name);
+    createNamespace(new NamespaceMeta.Builder().setName(namespace).build());
+    return namespace;
+  }
+
+  protected void createNamespace(NamespaceMeta namespaceMeta) throws Exception {
```
anwar6953: Instead of implementing these `createNamespace` and `deleteNamespace` method, I think its fine to simply expose them via the `TestManager`.
So in the test case, user can simply call `getTestManager().deleteNamespace`.

I think that `TestManager` class has so many methods, that having the equivalent (wrapper) methods in `IntegrationTestBase.java` would be redundant and messy.

chtyim:
LGTM

rohanahata:
Squashed commits.

PR #2821, opened 2015-06-03
Feature/pref cli command fix

Shilpaks:
http://builds.cask.co/browse/CDAP-DUT1949


chtyim:
In general it's better to have more descriptive commit message to describe what the changes/fixes are.

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -29,48 +30,93 @@
...
-        printSuccessMessage(printStream, type);
-        break;
-
-      default:
-        throw new IllegalArgumentException("Unrecognized Element Type for Preferences " + type.getTitleName());
+    private final PreferencesClient client;
+    private final ElementType type;
+
+    protected AbstractSetPreferencesCommand(ElementType type, PreferencesClient client, CLIConfig cliConfig) {
+        super(cliConfig);
```
chtyim: The indentation is wrong. It should be two spaces instead of four. The checkstyle should fail.

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -29,48 +30,93 @@
...
-        client.setProgramPreferences(parseProgramId(programIdParts, type.getProgramType()), args);
-        printSuccessMessage(printStream, type);
-        break;
-
-      default:
-        throw new IllegalArgumentException("Unrecognized Element Type for Preferences " + type.getTitleName());
+    private final PreferencesClient client;
+    private final ElementType type;
+
+    protected AbstractSetPreferencesCommand(ElementType type, PreferencesClient client, CLIConfig cliConfig) {
```
sreevatsanraman: Looks like the tab is 4 space instead of 2. Need to make sure your IDE has the right settings.

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -29,48 +30,93 @@
...
+
+            default:
+                throw new IllegalArgumentException("Unrecognized Element Type for Preferences " + type.getTitleName());
+        }
+    }
+
+
+
+    protected String determinePattern(String action) { // action can either be set or load
+        if (action.equals("set")) {
```
chtyim: When comparing string with string literal, do it the other way round as it handles null better.

```if ("sets".equals(action))``.

File: cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java [hunk 4]
```diff
@@ -29,48 +30,93 @@
...
+                case FLOW:
+                case MAPREDUCE:
+                case WORKFLOW:
+                case SERVICE:
+                case SPARK:
+                    return String.format("set preferences %s <%s> [<%s>]", type.getName(), ArgumentName.RUNTIME_ARGS,
+                            type.getArgumentName());
+            }
+        }
+        else if(action.equals("load")) {
```
chtyim: else should be right after closing brace '} else if (...) {`

cdap-cli/src/main/java/co/cask/cdap/cli/command/LoadPreferencesCommand.java
```diff
@@ -86,8 +86,7 @@ public void perform(Arguments arguments, PrintStream printStream) throws Excepti
 
   @Override
   public String getPattern() {
-    return String.format("load preferences %s <%s> <%s> [<%s>]", type.getName(), ArgumentName.LOCAL_FILE_PATH,
-                         ArgumentName.CONTENT_TYPE, type.getArgumentName());
+    return this.determinePattern("load");
```
chtyim: No need to have `this` when calling method in the same class.

cdap-cli/src/main/java/co/cask/cdap/cli/command/SetPreferencesCommand.java
```diff
@@ -60,8 +60,7 @@ public void perform(Arguments arguments, PrintStream printStream) throws Excepti
 
   @Override
   public String getPattern() {
-    return String.format("set preferences %s <%s> [<%s>]", type.getName(), ArgumentName.RUNTIME_ARGS,
-                         type.getArgumentName());
+    return this.determinePattern("set");
```
chtyim: Same, no need to have this.

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
chtyim: Also, in Java 7, you can use `switch` statement on string and is better for this case. Also, better to have private method that do specific task.

```java
switch (action) {
  case "set":
    handleSet(...);
  break;
  case "load":
    handleLoad(...);
  break;
}

File: cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java [hunk 7]
```diff
@@ -29,48 +30,93 @@
...
+                client.setProgramPreferences(parseProgramId(programIdParts, type.getProgramType()), args);
+                printSuccessMessage(printStream, type);
+                break;
+
+            default:
+                throw new IllegalArgumentException("Unrecognized Element Type for Preferences " + type.getTitleName());
+        }
+    }
+
+
```
chtyim: Remove multiple empty lines. Only need one empty line between methods.

gsps1:
its useful to post JIRA link in the PR.

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -29,48 +30,93 @@
...
+
+    protected String determinePattern(String action) { // action can either be set or load
+        if (action.equals("set")) {
+            switch (type) {
+                case INSTANCE:
+                    return String.format("set preferences %s <%s>", type.getName(), ArgumentName.RUNTIME_ARGS);
+                case NAMESPACE:
+                    return String.format("set preferences %s <%s>", type.getName(), ArgumentName.RUNTIME_ARGS);
```
awholegunch: Instance and Namespace cases can't be combined?

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -29,48 +30,93 @@
...
+                    return String.format("set preferences %s <%s> <%s>", type.getName(), ArgumentName.RUNTIME_ARGS,
+                            type.getArgumentName());
+                case FLOW:
+                case MAPREDUCE:
+                case WORKFLOW:
+                case SERVICE:
+                case SPARK:
+                    return String.format("set preferences %s <%s> [<%s>]", type.getName(), ArgumentName.RUNTIME_ARGS,
+                            type.getArgumentName());
+            }
```
awholegunch: reading this, I'd rather see one return statement, and the first cases set extraParameters = "", the next extraParameters = "<%s>" and the last extraParameters = "[<%s>]"...

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -29,48 +30,93 @@
...
+                                         type.getName(), ArgumentName.RUNTIME_ARGS);
+                case APP:
+                    return String.format("set preferences %s <%s> <%s>", type.getName(),
+                                         ArgumentName.RUNTIME_ARGS, type.getArgumentName());
+                case FLOW:
+                case MAPREDUCE:
+                case WORKFLOW:
+                case SERVICE:
+                case SPARK:
+                    return String.format("set preferences %s <%s> [<%s>]", type.getName(), ArgumentName.RUNTIME_ARGS,
```
gokulavasan: For programs, the last argument is not optional. That is, we need the program name.

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -73,4 +74,57 @@ protected void setPreferences(String[] programIdParts, PrintStream printStream,
...
+  }
+
+  private String determinePatternSetHelper() {
+    switch (type) {
+      case INSTANCE:
+      case NAMESPACE:
+        return String.format("set preferences %s <%s>",
+                             type.getName(), ArgumentName.RUNTIME_ARGS);
+      case APP:
+        return String.format("set preferences %s <%s> <%s>", type.getName(),
```
gokulavasan: Can't we combine programs and app case statements?

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -73,4 +74,57 @@ protected void setPreferences(String[] programIdParts, PrintStream printStream,
...
+    return "None";
+  }
+
+  private String determinePatternLoadHelper() {
+    switch (type) {
+      case INSTANCE:
+      case NAMESPACE:
+        return String.format("load preferences %s <%s> <%s>", type.getName(),
+                             ArgumentName.LOCAL_FILE_PATH, ArgumentName.CONTENT_TYPE);
+      case APP:
```
gokulavasan: Same here

cdap-cli/src/main/java/co/cask/cdap/cli/command/AbstractSetPreferencesCommand.java
```diff
@@ -73,4 +74,57 @@ protected void setPreferences(String[] programIdParts, PrintStream printStream,
         throw new IllegalArgumentException("Unrecognized element type for preferences " + type.getTitleName());
     }
   }
+
+  protected String determinePattern(String action) {
+
+    switch (action) {
+      case "set":
```
gokulavasan: Also looks like we can combine the two switch statements into one and use the 'action' string to form the format.

Shilpaks: It might not make sense to combine the two switch statements as there are some differences in the arguments used for formatting load and save preferences commands.

gokulavasan:
LGTM :+1:

PR #2822, opened 2015-06-03
CDAP-2597 Catch invalid Plugin configuration and return 400 instead of 500

gokulavasan:
Handle NumberFormatException (invalid Integer/Double values in the PluginConfiguration) which was previously propagated as RuntimeException and thus returned 500 instead of 400 Bad Request.

Build : http://builds.cask.co/browse/CDAP-RBT300

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
@@ -614,6 +614,10 @@ public class AdapterService extends AbstractIdleService {
       if (cause instanceof IllegalArgumentException) {
         throw (IllegalArgumentException) cause;
       }
+
+      if (cause instanceof InvalidPluginConfigException) {
+        throw (InvalidPluginConfigException) cause;
+      }
       throw new RuntimeException(e);
     } catch (Exception e) {
       throw new RuntimeException(e);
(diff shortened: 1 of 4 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -614,6 +614,10 @@ public void stop(Id.Namespace id, Id.Program programId, ProgramType type) throws
       if (cause instanceof IllegalArgumentException) {
         throw (IllegalArgumentException) cause;
       }
+
```
chtyim: seems better to check if the `cause` is of type `RuntimeException`. If that's the case no need to wrap. You can simply do:

```throw Throwables.propagate(e.getCause())```

File: cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java [hunk 2]
```diff
@@ -268,9 +269,17 @@ private Object convertValue(TypeToken<?> fieldType, String value) throws Excepti
         rawType = Primitives.wrap(rawType);
       }
 
-      if (Primitives.isWrapperType(rawType)) {
-        Method valueOf = rawType.getMethod("valueOf", String.class);
-        return valueOf.invoke(null, value);
+      try {
```
chtyim: I would prefer the `try-catch` block be more precise, meaning moving it inside the `if` block, just around the `invoke` call.

gokulavasan:
@chtyim Addressed comments. Please review. Thanks!

chtyim:
LGTM

PR #2823, opened 2015-06-03
Fix metric-picker issue where latest tag was replacing entire tags string

anwar6953:
Fix metric-picker issue where latest tag was replacing entire tags string.


ajainarayanan:
:+1: LGTM

PR #2824, opened 2015-06-03
Ability to stop the particular run of the program (MapReduce/Workflow).

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-2541
Build: http://builds.cask.co/browse/CDAP-DUT1951-1

This PR contains following changes:
1. Configuration property `concurrentRunsDisabled` is used to control whether the concurrent runs of the MapReduce and Workflow programs should be allowed. By default the property is set to `false`, which means we will allow the concurrent runs of the MapReduce and Workflow programs by default.
2. Added endpoint  `/apps/{app-id}/{type}/{id}/runs/{run-id}/stop` to stop the particular run of the program.


anew:
Why would you want to disable concurrent runs? And what does it mean for a schedule that fires before the current run is finished?

sagarkapare:
We are enabling the concurrent runs by default for MapReduce and Workflows through schedules as well as through UI(rest end points). If user wants to disable the concurrent run, he can do it so by setting the property `concurrentRunsDisabled ` to `true`. Default value of this property is `false`.

anew:
I amassing g for a use case to disable it,and if so, what happens to programs that cannot run because if that. Are they simply skipped? Or will they run ax soon as the current run is done?

sagarkapare:
If the program cannot run because of the the concurrent runs are disabled, that particular schedule run will be skipped. It will run next time when schedule triggers. Should we remove the property `concurrentRunsDisabled ` and always have concurrent runs of the MapReduce and Workflow programs?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1367,11 +1393,34 @@ private boolean isRunning(Id.Program id) {
...
+  private AppFabricServiceStatus stop(Id.Program id) {
+    return stop(id, null);
+  }
+
   /**
    * Stops a Program.
    */
-  private AppFabricServiceStatus stop(Id.Program identifier) {
-    ProgramRuntimeService.RuntimeInfo runtimeInfo = findRuntimeInfo(identifier);
+  private AppFabricServiceStatus stop(Id.Program identifier, String runId) {
```
anwar6953: annotate `runId` param as `@nullable`, since it seems to be so.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1367,11 +1393,34 @@ private boolean isRunning(Id.Program id) {
...
+    ProgramRuntimeService.RuntimeInfo runtimeInfo;
```
anwar6953: I think it makes sense to abstract the next 6-7 lines into a method like `findRuntimeInfo(identifier, runId)`, since it'd be useful also when getting status by runId.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Good point! Made the changes.

sagarkapare:
@anew @anwar6953 Can you please take a look again? Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1340,18 +1374,17 @@ private synchronized void startStopProgram(HttpRequest request, HttpResponder re
...
-        return AppFabricServiceStatus.PROGRAM_ALREADY_RUNNING;
-      }
-
       Map<String, String> sysArgs = propertiesResolver.getSystemProperties(id);
       Map<String, String> userArgs = propertiesResolver.getUserProperties(id);
       if (overrides != null) {
         userArgs.putAll(overrides);
       }
 
+      if (isRunning(id) && isConcurrentRunsDisabled(id, userArgs)) {
```
anew: I am not sure I understand this correctly: What are the different ways that the CONCURRENT_RUNS_DISABLED can be set? Is it a property of the program? Or does it have to be set in the preferences store? Or is it a runtime argument? Here it appears that it can only be from preferences, but overridden with runtime arguments? 

I don't know what is the right way to disable/enable it, because I am not sure what the use case is. What would be reason to disable it? And if it is disabled, shouldn't that be a property of the program itself? Or a property of the schedule that triggers it?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -318,6 +319,32 @@ private void getScheduleStatus(HttpResponder responder, String appId, String nam
...
+      if (!isRunLevelActionAllowed(programType)) {
+        responder.sendString(HttpResponseStatus.FORBIDDEN,
+                             "Run level operation is allowed only for MapReduce and Workflow program.");
+      }
+
+      Id.Program program = Id.Program.from(namespaceId, appId, programType, id);
+      AppFabricServiceStatus status = stop(program, runId);
+      responder.sendString(status.getCode(), status.getMessage());
+     } catch (IllegalArgumentException e) {
+      responder.sendString(HttpResponseStatus.BAD_REQUEST, String.format("Invalid program type %s", type));
```
chtyim: Throw the `BadRequestException` instead. It's the new way to have one place to handle exceptions from handlers and responses properly.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -318,6 +319,32 @@ private void getScheduleStatus(HttpResponder responder, String appId, String nam
...
+  public void performRunLevelStop(HttpRequest request, HttpResponder responder,
+                                  @PathParam("namespace-id") String namespaceId,
+                                  @PathParam("app-id") String appId,
+                                  @PathParam("type") String type,
+                                  @PathParam("id") String id,
+                                  @PathParam("run-id") String runId) {
+    try {
+      ProgramType programType = ProgramType.valueOfCategoryName(type);
+      if (!isRunLevelActionAllowed(programType)) {
+        responder.sendString(HttpResponseStatus.FORBIDDEN,
```
chtyim: The method should return after this. It's better to throw an exception so that the `HttpExceptionHandler` can handle the response. Currently we don't have an exception representing the `forbidden` request. You can add a new exception and add handling logic to `HttpExceptionHandler` as well.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -318,6 +319,32 @@ private void getScheduleStatus(HttpResponder responder, String appId, String nam
...
+  @Path("/apps/{app-id}/{type}/{id}/runs/{run-id}/stop")
+  public void performRunLevelStop(HttpRequest request, HttpResponder responder,
+                                  @PathParam("namespace-id") String namespaceId,
+                                  @PathParam("app-id") String appId,
+                                  @PathParam("type") String type,
+                                  @PathParam("id") String id,
+                                  @PathParam("run-id") String runId) {
+    try {
+      ProgramType programType = ProgramType.valueOfCategoryName(type);
+      if (!isRunLevelActionAllowed(programType)) {
```
chtyim: BTW, why need this check? What happen if applying the stop to other program types?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1244,10 +1271,17 @@ protected ProgramStatus getProgramStatus(Id.Program id) {
...
-                               identifier.getNamespaceId(), identifier.getApplicationId());
-    for (ProgramRuntimeService.RuntimeInfo info : runtimeInfos) {
+    return findRuntimeInfo(identifier, null);
+  }
+
+  protected ProgramRuntimeService.RuntimeInfo findRuntimeInfo(Id.Program identifier, @Nullable String runId) {
+    Map<RunId, ProgramRuntimeService.RuntimeInfo> runtimeInfos = runtimeService.list(identifier.getType());
+
+    if (runId != null) {
+      return runtimeInfos.get(RunIds.fromString(runId));
```
chtyim: So this method can have `null` result. Annotate the method with `@Nullable`.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: As discussed, removed the property `CONCURRENT_RUNS_DISABLED`

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Updated the code.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: Throwing `ForbiddenException` now.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: This endpoint is only for the MapReduce and Workflow programs. For them we can have multiple runs concurrently. So we are providing the ability to select the particular run for stopping.

The endpoint `/apps/{app-id}/{type}/{id}/{action}` handles the stop requests for other types of programs.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java

sagarkapare:
@chtyim @anew Addressed comments. Please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
bdmogal: @chtyim @sagarkapare sorry, I missed this earlier. There are exceptions like ``CannotBeDeletedException``, ``CannotBeCreatedException``, etc in the codebase already. We respond with ``FORBIDDEN`` for these exceptions too.

I think it is ok to add ``ForbiddenException``, but lets also then make ``CannotBeDeletedException`` and ``CannotBeCreatedException`` subclasses of ``ForbiddenException``.

Also, I'm not sure, but should we rename it to ``CannotPerformOperationException`` to be consistent with existing names?

bdmogal: On second thoughts, this may be unrelated to this PR, so I think it is ok to report a jira and handle in a separate PR. Even more so, because I think these exceptions do not have a standard HttpResponse currently. They either result in ``FORBIDDEN`` or ``CONFLICT``. Definitely a separate issue from this one.

chtyim:
LGTM

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT1951-4. Merging.

PR #2828, opened 2015-06-03
CDAP-2587 During namespace create, delete the namespace directory on the filesystem if it already exists

bdmogal:
1st commit deletes the namespace directory if it exists during namespace creation. It also switches the order of deletion during namespace delete - Removes from MDS first, then from the storage providers. This ordering will help us make this process more fool-proof in future.

2nd commit is a mere rename of all ``Underlying**`` classes to ``StorageProvider**``.

Jira: [CDAP-2587](https://issues.cask.co/browse/CDAP-2587)
Build: http://builds.cask.co/browse/CDAP-RBT301


anwar6953:
LGTM

bdmogal:

PR #2830, opened 2015-06-03
Fixes going from app detailed status view to home page

ajainarayanan:
backbutton-programstatus

### DIFF
--- a/cdap-ui/app/features/apps/routes.js
+++ b/cdap-ui/app/features/apps/routes.js
@@ -34,18 +34,28 @@ angular.module(PKG.name+'.feature.apps')
       })
         .state('apps.detail.overview', {
           url: '/overview',
-          parent: 'apps.detail',
+          // parent: 'apps.detail',
           templateUrl: '/assets/features/apps/templates/detail.html',
           ncyBreadcrumb: {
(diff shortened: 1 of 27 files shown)


cdap-ui/app/features/apps/routes.js
```diff
@@ -34,18 +34,28 @@ angular.module(PKG.name+'.feature.apps')
       })
         .state('apps.detail.overview', {
           url: '/overview',
-          parent: 'apps.detail',
+          // parent: 'apps.detail',
```
anwar6953: just remove if not needed?

elfenheart:
This change actually breaks everything related to apps view...

elfenheart:
There is also the back button from dataset and streams overview page...

ajainarayanan:
@elfenheart Fixed all references of apps and fixed the same issue with datasets and streams features.

For datasets the implementation is a little different where if it is explorable then we show all tabs and if its not we show only programs tab. All other references to a dataset will still have status but the 'datasets.details' state determines if its explorable or not and shows appropriate things.
backbutton

elfenheart:
:+1: LGTM

PR #2832, opened 2015-06-03
Remove @beta from Filesets

awholegunch:
Remove beta annotation and import, and remove "experimental" description in documentation.


duddun:
LGTM

PR #2834, opened 2015-06-03
CDAP-2612 Add Flume agent to VM

wolf31o2:
Pull in the `hadoop` [cookbook](https://github.com/caskdata/hadoop_cookbook) from Chef Supermarket and install the `hadoop::flume_agent` recipe, from the default distribution.


duddun:
:+1:

PR #2836, opened 2015-06-03
CDAP-2640 Remove persistent network rules on VM

wolf31o2:
This removes the MAC address -> interface mapping from udev, so the first ethernet device will become `eth0` whether the MAC is reinitialized or not.


sreevatsanraman:
LGTM :+1:

PR #2837, opened 2015-06-03
Fix/ui paginate

ajainarayanan:
- Fixes pagination to work across pages.
- Fixes minor typo in view to call callback in controller
- Fixes pagination based on search results


elfenheart:
:+1: LGTM

PR #2839, opened 2015-06-03
CDAP-2601 Plugin config desc fix

gokulavasan:
Fix plugin config properties description and remove the redundant DRIVER_CLASS property in DB Source/Sink (which can be obtained using the Plugin Type and Plugin Name).

Build :


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -156,15 +160,16 @@ private void setResultSetMetadata() throws Exception {
...
+  private void ensureJDBCDriverIsAvailable(BatchSinkContext context) throws Exception {
     try {
       DriverManager.getDriver(dbSinkConfig.connectionString);
     } catch (SQLException e) {
       // Driver not found. We will try to register it with the DriverManager.
-      LOG.debug("Driver {} not found. Registering JDBC driver via shim {} ",
-                dbSinkConfig.driverClass, JDBCDriverShim.class.getName());
-      ClassLoader classLoader = conf.getClassLoader();
-      Class<?> driverClass = classLoader.loadClass(dbSinkConfig.driverClass);
+      LOG.debug("Plugin Type : {} and Plugin Name : {} not found. Registering JDBC driver via shim {} ",
```
bdmogal: should probably also log the driver class name

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/JmsSource.java
```diff
@@ -176,8 +176,7 @@ private void initializeJMSConnection(Hashtable<String, String> envVars, String d
           LOG.warn("Exception when closing connection", ex2);
         }
       }
-      throw new RuntimeException("JMSException thrown when trying to initialize connection: " + ex.getMessage(),
-                                 ex);
+      throw new RuntimeException("JMSException thrown when trying to initialize connection: " + ex.getMessage(), ex);
```
bdmogal: Not your PR, but probably don't need the ``ex.getMessage()`` when you're already logging with the Exception parameter.

bdmogal:
A couple of minor comments about log messages, otherwise LGTM. Also, we should remove this property from the ETL guides.

bdmogal:
:+1:

PR #2849, opened 2015-06-03
Fix the build log message.

awholegunch:
The doc log was truncating the last line of any collected messages. This fixes that by adding an extra line at the end.


duddun:
:+1:

PR #2850, opened 2015-06-04
Feature/cdap 2638 fix immutability of client configs

rohanahata:
This pull request aims to make clientconfigs immutable. All dependent calls have been rerouted through the builder.


anwar6953:
Just to clarify, this will have conflicts with: https://github.com/caskdata/cdap/pull/2817
So once one is merged, the other will need additional changes, right?

rohanahata:
@anwar6953 Yes. Both are standalone atm... once one gets modified the other will have to be changed accordingly

anwar6953:
And just to clarify:
Doing the following:
```java
new ClientConfig.Builder().setNamespace(myNamespace). setConnectionConfig(myConnectionConfig).build()
```
will result in a ClientConfig in which the `myNamespace` being set is not necessarily there, because the `myConnectionConfig` overrode it.

I'm not sure if that's desirable or not, but definitely wanted to point it out.

rohanahata:
@anwar6953 Addressed the issue that you brought up

rohanahata:
CLI works on the premise that there is one ClientConfig and mutating that affects everything. If we were to implement ClientConfigs immutably, a major chunk of CLIMain would have to be refactored. After discussing with @chtyim and @gsps1 we came to the conclusion to close this PR.

PR #2855, opened 2015-06-04
Use Java 7 diamond operator inference.

anwar6953:
Intellij and the docs linked below say that in Java 7, the explicit type declaration for generic classes is unnecessary as long as the compiler can infer them, so removed them.
I don't see any drawbacks to this.

http://docs.oracle.com/javase/7/docs/technotes/guides/language/type-inference-generic-instance-creation.html

https://builds.cask.co/browse/CDAP-DUT1972-1

These updates were done by IntelliJ's Run Inspection > Fix utility.


chtyim:
LGTM

PR #2858, opened 2015-06-04
[CDAP-2646] Fix/ui namespace list dropdown

elfenheart:
2 fixes:
- updating namespace dropdown after namespace delete
- handle bug when deleting current active namespace


ajainarayanan:
:+1: LGTM

PR #2860, opened 2015-06-04
Fix broken doc build.

awholegunch:
Fixes the doc build. The break was due to changes in Java files that are included in the examples and are incorporated as part of the build. As these files are referenced by line number, any changes to the files can potentially require a change in the code that includes fragments of the Java code. In this case, that was so, as lines had been deleted from the Java code, and so the offsets needed to be adjusted in the documentation to accommodate.

### DIFF
--- a/cdap-docs/examples-manual/build.sh
+++ b/cdap-docs/examples-manual/build.sh
@@ -154,9 +154,9 @@ function download_includes() {
   test_an_include 2176f426a50b2d68817cd7da7b7faffb ../../cdap-examples/StreamConversion/src/main/java/co/cask/cdap/examples/streamconversion/StreamConversionApp.java
   test_an_include b3ad7dc7ddbb5b905145dbf15a5c9b93 ../../cdap-examples/StreamConversion/src/main/java/co/cask/cdap/examples/streamconversion/StreamConversionMapReduce.java
   
-  test_an_include dc6519d102875a83621f262c33c61894 ../../cdap-examples/UserProfiles/src/main/java/co/cask/cdap/examples/profiles/UserProfiles.java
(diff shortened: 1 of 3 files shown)


cdap-docs/examples-manual/build.sh
```diff
@@ -154,9 +154,9 @@ function download_includes() {
   test_an_include 2176f426a50b2d68817cd7da7b7faffb ../../cdap-examples/StreamConversion/src/main/java/co/cask/cdap/examples/streamconversion/StreamConversionApp.java
   test_an_include b3ad7dc7ddbb5b905145dbf15a5c9b93 ../../cdap-examples/StreamConversion/src/main/java/co/cask/cdap/examples/streamconversion/StreamConversionMapReduce.java
   
-  test_an_include dc6519d102875a83621f262c33c61894 ../../cdap-examples/UserProfiles/src/main/java/co/cask/cdap/examples/profiles/UserProfiles.java
+  test_an_include 2260781c7cef2938aa36c946f3a34341 ../../cdap-examples/UserProfiles/src/main/java/co/cask/cdap/examples/profiles/UserProfiles.java
```
henry-cask: Always curious what does this number represent?

anew: So this tests whether the included file's MD5 matches that number? And the build fails if it is not the same? So every time someone changes a file, the docs build fails and you need to change the expected MD5 sums? Sounds like a lot of manual work, but I like it better than undetected broken literal includes. +1

anew:
LGTM

cdap-docs/examples-manual/build.sh
awholegunch: @anew 's explanation is correct. Sometimes the file has changed but it doesn't affect the include, and so I just update the MD5 hash. With this change, the files actually did change such that the include needed to be altered. It is manual work, but I haven't figured out a better solution.

PR #2862, opened 2015-06-04
Install Subversion into VM

wolf31o2:
The `motd` mentions Subversion and it's still very, very common. Adding it to the VM.


duddun:
:+1:

PR #2867, opened 2015-06-04
Feature/ui add adapter type

astewes:


cdap-ui/app/features/adapters/controllers/tabs/runs/tabs/status-ctrl.js
```diff
@@ -22,6 +25,6 @@ angular.module(PKG.name + '.feature.adapters')
         $scope.source = res.config.source;
         $scope.sink = res.config.sink;
         $scope.transforms = res.config.transforms || [];
+        $scope.template = res.template;
```
ajainarayanan: We don't need template to be added ```$scope``` right?

cdap-ui/app/features/adapters/controllers/tabs/runs/tabs/status-ctrl.js
```diff
@@ -4,15 +4,18 @@ angular.module(PKG.name + '.feature.adapters')
 
     $scope.transforms = [{
       name: '',
-      properties: {}
+      properties: {},
```
ajainarayanan: Can you add a TODO comment on top of this controller that we should use ```rAdapterDetail``` resolved variable. Right now we are making two different calls to backend for the same data. 
It could be something like this,
```
//TODO: We should use rAdapterDetail here since this data is already resolved at adapter.detail state
```

ajainarayanan:
:+1: LGTM

PR #2869, opened 2015-06-05
Add Application Template to Javadocs

awholegunch:
This PR adds the building of Javadocs for the Application Template packages. It changes the output of the Javadocs, in that instead of one long list of packages, it divides the list into two groups: 

- Core Packages
- Application Template Packages

This can be seen at: [CDAP API 3.0.1 Javadocs](http://docs-staging.cask.co/cdap/3.0.1-feature-r3.0.1_add_javadocs/en/reference-manual/javadocs/index.html).

The Javadoc build has ~~a large number of errors: over 200. Some are due to configuration issues (which I'll ask Terence to help on) and some are due to errors in the Javadoc comments in the source files (which I'll ask Gokul to help on). The errors can be seen in the [documentation build output log](http://builds.cask.co/download/CDAP-DDBD220-JOB1/build_logs/CDAP-DDBD220-JOB1-1.log).~~ one error, which we are going to put up with.

This fixes https://issues.cask.co/browse/CDAP-2643.


pom.xml
```diff
@@ -1498,14 +1499,40 @@
...
+                  </group>
+                </groups>
+                <links>
+                  <link>http://download.oracle.com/javase/${project.java.version}/docs/api/</link>
+                </links>
+                <maven.javadoc.failOnError>false</maven.javadoc.failOnError>
+                <quiet>true</quiet>
+                <!-- sourcepath needs to be specified to use excludePackageNames -->
+                <sourcepath>
+                  ${project.parent.basedir}/cdap-api/src/main/java:
```
gokulavasan: We have to explicitly specify this? And why is cdap-api here? I thought we had docs for it already.

gokulavasan:
One comment. Rest LGTM :+1:

pom.xml
awholegunch: There is a bug in the maven javadoc plugin: if you don't specify "sourcepath", the "excludePackageNames" directive won't work. Also, this is to specify all the javadocs being built: not just the App Template packages. Does that answer your question? (I wasn't quite sure what you meant by "I thought we had docs for it already."...)

awholegunch:
# Javadocs for co.cask.cdap.internal have been removed

One thing I forgot to point out: the change from earlier Javadocs is that the **Javadocs for co.cask.cdap.internal have been removed**. Compare the [new 3.0.1 Javadocs](http://docs-staging.cask.co/cdap/3.0.1-feature-r3.0.1_add_javadocs/en/reference-manual/javadocs/index.html) with the [current 3.0.0 Javadocs](http://docs.cask.co/cdap/current/en/reference-manual/javadocs/index.html) to see the difference. I though that we would like the internal packages not included as they are not intended for people to use. Correct me if I am wrong.

pom.xml
gokulavasan: Well we have javadocs for some packages under cdap-api module isn't it? So I was wondering why is cdap-api/src/main/java being included here. Basically if it was working earlier, why was this not present already.

gokulavasan:
@sreevatsanraman Can you confirm whether removal of co.cask.cdap.internal docs is ok.

Other changes LGTM.

pom.xml
awholegunch: To get this to work ("excludePackageNames", and the use of "sub packages", to restrict the use to just the two groups we'd like), "sourcepath" needs to be specified, and one needs to specify all the  source locations. The way this is being called (on the command line; see the changes to *BUILD.rst* and *common_build.sh*) has changed, and so a number of properties has changed to reflect that.

sreevatsanraman:
@gokulavasan yes, that's fine. LGTM :+1:

PR #2870, opened 2015-06-05
CDAP-2656 fix the time field for tweets to just use the timestamp

albertshau:
of the tweet instead of calculating difference in ingest time
versus tweet time.


gokulavasan:
LGTM :+1:

PR #2875, opened 2015-06-05
CDAP-2613 Cache dependencies to build CDAP examples

wolf31o2:
This is the same as #2871 cherry-picked for develop


duddun:
LGTM

PR #2876, opened 2015-06-05
Use IntegrationTestBase#getRestClient for all clients utilized in Int…

anwar6953:
Use `IntegrationTestBase#getRestClient` for all clients utilized in Integration testing framework

This makes it so that `getRestClient` can be overridden to return a custom `RESTClient` object, that is always used for all clients in the integration testing framework.
One example [use case](https://github.com/caskdata/cdap-integration-tests/blob/develop/integration-test-remote/src/test/java/co/cask/cdap/apps/AudiTestBase.java#L125) for why a user would want to do this is to create a `RESTClient` with a `RestClient.Listener`, for custom logging of all http requests made via the clients.

http://builds.cask.co/browse/CDAP-DUT1985-1


chtyim:
LGTM

PR #2882, opened 2015-06-05
Simplify AbstractAppFabricHttpHandler.java

anwar6953:
We removed quite a few v2 handlers just prior 3.0 release, but did not simplify the helper methods in AppFabric.
This cleans it up a bit:
`programList` no longer accepts an `applicationId`, since there are no usages that pass one.
Same for the `getAppRecords` method.

http://builds.cask.co/browse/CDAP-DUT1984-4

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
@@ -48,9 +48,7 @@ import org.slf4j.LoggerFactory;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.lang.reflect.Type;
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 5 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -168,36 +167,18 @@ protected int getInstances(HttpRequest request) throws IllegalArgumentException,
...
-        specList = new ArrayList<ApplicationSpecification>(store.getAllApplications(accId));
-      } else {
-        ApplicationSpecification appSpec = store.getApplication(new Id.Application(accId, appId));
-        if (appSpec == null) {
-          responder.sendStatus(HttpResponseStatus.NOT_FOUND);
-          return;
-        }
-        specList = Collections.singletonList(store.getApplication(new Id.Application(accId, appId)));
-      }
+      specList = new ArrayList<>(store.getAllApplications(namespace));
```
bdmogal: can be merged with the declaration.

bdmogal: In fact you should just be able to do the ``store.getAllApplications(namespace)`` in the for loop below

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/util/AbstractAppFabricHttpHandler.java
```diff
@@ -207,22 +188,11 @@ protected final void getAppRecords(HttpResponder responder, Store store, String
...
       List<ProgramRecord> programRecords;
-      if (applicationId == null) {
-        Id.Namespace accId = Id.Namespace.from(namespaceId);
-        programRecords = listPrograms(accId, type, store);
-      } else {
-        Id.Application appId = Id.Application.from(namespaceId, applicationId);
-        programRecords = listProgramsByApp(appId, type, store);
-      }
+      Id.Namespace namespace = Id.Namespace.from(namespaceId);
+      programRecords = listPrograms(namespace, type, store);
```
bdmogal: Merge with declaration?

bdmogal:
Only style comments. Intellij should warn already? 

LGTM otherwise

PR #2885, opened 2015-06-05
Skips jumping to admin view when uploading app from dev section

ajainarayanan:
appupload


ajainarayanan:
Cluster: http://add-app1697-1000.dev.continuuity.net:9999/ns/default

elfenheart:
Needs to be rebased, but :+1: LGTM

PR #2888, opened 2015-06-05
[CDAP-2663] fix dataset name with dot

elfenheart:
Cluster being created: edwin1698-1000.dev.continuuity.net


ajainarayanan:
:+1: LGTM

elfenheart:
http://edwin1698-1000.dev.continuuity.net:9999/

verified in cluster

PR #2893, opened 2015-06-06
Correct CDAP parent version.

awholegunch:
Shouldn't be SNAPSHOT, nor 3.1.0...


sreevatsanraman:
LGTM :+1: 

Although this is fixed in another feature branch by Ajai. We should merge this.

PR #2899, opened 2015-06-08
Fix Documentation Build in Develop

awholegunch:
Changes made to guarded files; the changes don't affect the HTML output, so merely an update of the MD5 hashes.

[Built successfully.](http://builds.cask.co/browse/CDAP-DDBD228-2/log)


henry-cask:
+1

LGTM

PR #2900, opened 2015-06-08
CDAP-2678 include record formats in the explore classpath so

albertshau:
that queries over streams will have access to the classes.


cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
```diff
@@ -273,6 +274,8 @@ private static void setupClasspath() throws IOException {
                                                                      bootstrapClassPaths, null));
     orderedDependencies.addAll(ExploreServiceUtils.traceDependencies(DatasetStorageHandler.class.getCanonicalName(),
                                                                      bootstrapClassPaths, null));
+    orderedDependencies.addAll(ExploreServiceUtils.traceDependencies(RecordFormats.class.getCanonicalName(),
```
chtyim: Use `getName` always.

chtyim:
Just one comment, rest LGTM.

albertshau:
http://builds.cask.co/browse/CDAP-RBT304

PR #2901, opened 2015-06-08
Propagate NaN values from JobClient to responder#sendJson, because ot…

anwar6953:
Propagate NaN values from JobClient to responder#sendJson, because otherwise GSON fails during serialization of NaN values.

Build: http://builds.cask.co/browse/CDAP-DUT1999-1
Related JIRA: https://issues.cask.co/browse/CDAP-2433


chtyim:
LGTM

PR #2903, opened 2015-06-08
[CDAP-2435] Workflow Nodes click handler 

elfenheart:
Preventing user to go to the mapreduce job when the action run has not started.


ajainarayanan:
:+1: LGTM

PR #2904, opened 2015-06-08
CDAP-2699 set y-axis 0 value to base of chart (since there are no neg…

anwar6953:
CDAP-2699 set y-axis 0 value to base of chart (since there are no negative metric values).

https://issues.cask.co/browse/CDAP-2699

image


elfenheart:
mentioned to @anwar6953 about queue.pending (flowlet) metric as timeseries can return negative value

anwar6953:
@elfenheart Just tested. Negative metric values work just fine with zero padding.

image

elfenheart:
:+1: LGTM

PR #2906, opened 2015-06-08
CDAP-2678 include record formats in the explore classpath so

albertshau:
that queries over streams will have access to the classes.


albertshau:
cherry-picking fix that went into 3.0 release branch

PR #2909, opened 2015-06-08
CDAP-2719 fix projection transform's conversion of null fields.

albertshau:


gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-RT37

PR #2910, opened 2015-06-09
Commit aims to get resolutions in the JSON response for Metrics. CDAP-2158

rohanahata:
Commit aims to get resolutions in the JSON response. Also test case coverage.

### DIFF
--- a/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
+++ b/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
@@ -277,8 +277,8 @@ public class MetricsHandlerTestRun extends MetricsSuiteTestBase {
 
     expected = ImmutableMap.of("testQuery3",
                                new QueryResult(ImmutableList.of(
-                                 new TimeSeriesSummary( ImmutableMap.<String, String>of(), "system.reads", 1, 4))
-                                 ,resolution),
+                                 new TimeSeriesSummary(ImmutableMap.<String, String>of(), "system.reads", 1, 4))
+                                 , resolution),
                                "testQuery4",
                                new QueryResult(ImmutableList.of(
                                  new TimeSeriesSummary(ImmutableMap.<String, String>of(), "system.reads", 1, 2),
--- a/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
+++ b/cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
@@ -544,20 +544,6 @@ public class MetricsHandlerTestRun extends MetricsSuiteTestBase {
   @Test
   public void testResolutionInResponse() throws Exception {
     long start = 1;
-    Map<String, String> sliceBy = getFlowletContext("resolutions", "WordCount1", "WordCounter", "run1", "splitter");
-
-    // 1 second
-    metricStore.add(new MetricValues(sliceBy, "reads", start, 1, MetricType.COUNTER));
-    // 30 second
-    metricStore.add(new MetricValues(sliceBy, "reads", start + 30, 1, MetricType.COUNTER));
-    // 1 minute
-    metricStore.add(new MetricValues(sliceBy, "reads", start + 60, 1, MetricType.COUNTER));
-    // 10 minutes
-    metricStore.add(new MetricValues(sliceBy, "reads", start + 600, 1, MetricType.COUNTER));
-    // 1 hour
-    metricStore.add(new MetricValues(sliceBy, "reads", start + 3600, 1, MetricType.COUNTER));
-    // 10 hours
-    metricStore.add(new MetricValues(sliceBy, "reads", start + 36000, 1, MetricType.COUNTER));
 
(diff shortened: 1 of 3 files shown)


chtyim:
It would be good to prefix the commit message with the JIRA issue. Something like:

```
(CDAP-2158) Return metrics resolutions in the JSON response.
 - Some more detail here about the changes.
```

cdap-proto/src/main/java/co/cask/cdap/proto/MetricQueryResult.java
```diff
@@ -44,6 +46,8 @@ public long getEndTime() {
     return series;
   }
 
+  public int getResolution() { return resolution; }
```
bdmogal: this should not pass checkstyle. there should be a newline before and after the ``return`` statement.

Btw, you may want to check if your IDE settings are up-to-date. You should get a checkstyle error in IDEA.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -471,7 +471,8 @@ private String humanToTagName(String humanTagName) {
     return Lists.newArrayList(Iterables.filter(metricNames, Predicates.notNull()));
   }
 
-  private MetricQueryResult decorate(Collection<MetricTimeSeries> series, long startTs, long endTs) {
+  private MetricQueryResult decorate(Collection<MetricTimeSeries> series, long startTs, long endTs,
+                                     Integer resolution) {
```
chtyim: Why takes an `Integer` not `int`?

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
```diff
@@ -541,6 +541,24 @@ public void testInterpolate() throws Exception {
...
 
+  @Test
+  public void testResolutionInResponse() throws Exception {
+    long start = 1;
+
+    String url = "/v3/metrics/query?" + getTags("resolutions", "WordCount1", "WordCounter", "splitter") +
+      "&metric=system.reads&resolution=auto&start=" + (start - 1) + "&end="
+      + (start + 36000);
+    MetricQueryResult queryResult = post(url, MetricQueryResult.class);
+    Assert.assertEquals(3600, queryResult.getResolution());
```
gsps1: should return 3600s

gsps1: also add a test case, where you skip resolution. then check for resolution in batched request as well. you can edit the existing test case if you may for that.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
```diff
@@ -352,7 +352,7 @@ private MetricQueryResult executeQuery(MetricQueryRequest queryRequest) throws E
       endTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis());
     }
 
-    return decorate(queryResult, timeRange.getStart(), endTime);
+    return decorate(queryResult, timeRange.getStart(), endTime, timeRange.getResolutionInSeconds().intValue());
```
gsps1: just do, timeRange.gerResolutionInSeconds(), you need to add  the unit - "s" as we talked before.

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
rohanahata: Needed a little clarity on this

rohanahata:
https://builds.cask.co/browse/CDAP-DUT2005/

cdap-proto/src/main/java/co/cask/cdap/proto/MetricQueryResult.java
```diff
@@ -25,11 +25,13 @@
...
   private final long endTime;
   private final TimeSeries[] series;
+  private final String resolution;
 
-  public MetricQueryResult(long startTime, long endTime, TimeSeries[] series) {
+  public MetricQueryResult(long startTime, long endTime, TimeSeries[] series, int resolution) {
     this.startTime = startTime;
     this.endTime = endTime;
     this.series = series;
+    this.resolution = resolution + "s";
```
gsps1: String.valueOf(resolution) + "s"

cdap-watchdog/src/main/java/co/cask/cdap/metrics/query/MetricsHandler.java
gsps1: can remove intValue()

cdap-gateway/src/test/java/co/cask/cdap/gateway/handlers/metrics/MetricsHandlerTestRun.java
```diff
@@ -329,7 +335,27 @@ public void testInvalidRequest() throws Exception {
       this.metrics = metrics;
       this.groupBy = groupBy;
       this.timeRange = timeRange;
+      this.resolution = resolution;
```
gsps1: do you need this ? i don't see resolution in the constructor params ?

gsps1:
LGTM

PR #2912, opened 2015-06-09
hide loading icon when there is an error in delete function

elfenheart:
hiding loading icon when an error occurs during delete operation of dataset, stream, or namespace


ajainarayanan:
:+1: LGMT

PR #2913, opened 2015-06-09
CDAP-2722 UI specific descriptions for projection, since the

albertshau:
backend describes the expected format of the values while the UI
hides that from users.


gokulavasan:
LGTM :+1:

PR #2916, opened 2015-06-09
Fix/ui adapter name

ajainarayanan:
- Fixes adapter create view to handle long adapter names.
- Adds a loading animation creating an adapter
adaptercreate


ajainarayanan:
adaptercreate

@elfenheart Have updated PR with a better implementation.

elfenheart:
:+1: LGTM

PR #2919, opened 2015-06-09
Feature/ui datasets controller as syntax

ajainarayanan:
- Modifies dataset controllers to use controller-as syntax
- Removes un-used files (templates + controllers)


cdap-ui/app/features/datasets/routes.js
```diff
@@ -15,7 +15,8 @@ angular.module(PKG.name + '.feature.datasets')
       .state('datasets.list', {
```
elfenheart: I don't think this state is applicable anymore?

ajainarayanan:
dataset

elfenheart:
:+1: LGTM

PR #2920, opened 2015-06-09
[CDAP-2595] Feature/ui flows data modelling

elfenheart:
Flows data modelling and code refactoring for readability

PS: don't merge yet, creating a cluster after everything looks good


cdap-ui/app/features/flows/controllers/tabs/runs/flowlets/input-ctrl.js
```diff
@@ -25,152 +28,159 @@ angular.module(PKG.name + '.feature.flows')
...
+        dataSrc.request({
+          _cdapPath: path + '&start=now-60s&end=now&aggregate=true',
+          method: 'POST'
+        }).then(function(res) {
+
+          // Get initial aggregate
+          aggregate = res.series[0] ? res.series[0].data[0].value : 0;
+
+        }).then(function() {
+
```
ajainarayanan: Since we are returning any promise in the previous ```then``` handler would this then function even be executed? The idea of chaining promises is to return back a promise from each then clause so that the next subsequent then function will be executed right?

cdap-ui/app/features/flows/controllers/tabs/runs/flowlets/input-ctrl.js
```diff
@@ -25,152 +28,159 @@ angular.module(PKG.name + '.feature.flows')
...
+            _cdapPath: path + '&start=now-60s&end=now',
+            method: 'POST'
+          }).then(function(initial) {
+            if (initial.series[0]) {
+              formatInitialTimeseries(aggregate, initial, input);
+            } else {
+              formatEmptyGraph(input);
+            }
+
+          }).then(function() {
```
ajainarayanan: We could make these ```then``` handlers in one level. If we return 

``` dataSrc.request({..}) ``` 

which is a promise, I believe the ```then``` handler can be in the same level as line 156. right?

File: cdap-ui/app/features/flows/controllers/tabs/runs/flowlets/input-ctrl.js [hunk 3]
```diff
@@ -25,152 +28,159 @@ angular.module(PKG.name + '.feature.flows')
...
+            // start polling aggregate
+            pollAggregateQueue(path, input);
+          });
+
+        });
```
ajainarayanan: Also we have many then handlers would be nice to have a ```catch``` if at all any step has any errors..

cdap-ui/app/features/flows/controllers/tabs/runs/flowlets/output-ctrl.js
```diff
@@ -39,67 +43,66 @@ angular.module(PKG.name + '.feature.flows')
...
-                  var val = [];
+                var val = [];
 
-                  for (var i = 60; i > 0; i--) {
-                    val.push({
-                      time: Math.floor((new Date()).getTime()/1000 - (i)),
-                      y: 0
-                    });
-                  }
+                for (var i = 60; i > 0; i--) {
```
ajainarayanan: Except for this ```for``` loop the ```updateOutput``` function and this function are the same right? Is it possible to have some code re-use here?

ajainarayanan:
:+1: LGTM

PR #2921, opened 2015-06-10
Feature/ui streams controller as syntax

ajainarayanan:
- Modifies Streams controllers to use controller-as syntax
- Removes un-used files.


cdap-ui/app/features/streams/controllers/tabs/status-ctrl.js
```diff
@@ -1,7 +1,8 @@
 angular.module(PKG.name + '.feature.streams')
-  .controller('StreamsDetailStatusController', function($scope, $state, myHelpers, MyDataSource, myStreamApi) {
+  .controller('StreamDetailStatusController', function($scope, $state, myHelpers, MyDataSource, myStreamApi) {
     var dataSrc = new MyDataSource($scope);
-
+    this.storage = null;
+    this.events = null;
```
elfenheart: should we change this default value to 0?

cdap-ui/app/features/streams/controllers/tabs/status-ctrl.js
```diff
@@ -34,7 +35,8 @@ angular.module(PKG.name + '.feature.streams')
         method: 'POST'
       }, function(metricData) {
           var data = myHelpers.objectQuery(metricData, 'series', 0, 'data', 0, 'value');
-          $scope[metric.scopeProperty] = data;
-        });
+          console.info(metric.scopeProperty, data);
```
elfenheart: remove?

cdap-ui/app/features/streams/routes.js
```diff
@@ -15,7 +15,8 @@ angular.module(PKG.name + '.feature.streams')
       .state('streams.list', {
```
elfenheart: same comment as dataset, list state is not applicable anymore

cdap-ui/app/features/streams/controllers/tabs/status-ctrl.js
ajainarayanan: The default value is not used anywhere. In the template we check for null value and use a &mdash; to indicate that we don't have any value for it. Thought null would make sense since if we don't get any metrics for it its default is null (which means we don't have any metric for it). Not sure if backend would return a value of 0 in which case we should display 0 instead of &mdash;

ajainarayanan:
streams

elfenheart:
:+1: LGTM

PR #2927, opened 2015-06-10
remove catch function in promise chain

elfenheart:
catch() is not a function defined in myPromise.

cluster: http://developtest1732-1000.dev.continuuity.net:9999/


ajainarayanan:
:+1: LGTM

PR #2929, opened 2015-06-11
Fix StreamHandlerTest failure

jayfjin:
Null start value


gsps1:
build running : http://builds.cask.co/browse/CDAP-DUT2021

gsps1:
LGTM

PR #2931, opened 2015-06-11
Feature/ui programs link directive

elfenheart:
Refactor links to programs into a directive. If we add more program types, we can just add the link to the new program type within the directive instead of many different places.

### DIFF
--- a/cdap-ui/app/directives/program-link/program-link.less
+++ b/cdap-ui/app/directives/program-link/program-link.less
@@ -1,0 +1,5 @@
+my-program-link {
+  .text-bold {
+    font-weight: bold;
+  }
+}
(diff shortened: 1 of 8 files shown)


cdap-ui/app/directives/program-link/program-link.less
```diff
@@ -0,0 +1,5 @@
+my-program-link {
+  .text-bold {
+    font-weight: bold;
+  }
+}
```
anwar6953: missing newline

cdap-ui/app/features/apps/templates/tabs/status.html
```diff
@@ -23,24 +23,15 @@
...
-                </a>
-                <a ng-if="program.type === 'Spark'" ui-sref="spark.detail.runs({programId: program.name})">
-                  <strong> {{program.name}} </strong>
-                </a>
-                <a ng-if="program.type === 'Worker'" ui-sref="worker.detail.runs({programId: program.name})">
-                  <strong> {{program.name}} </strong>
-                </a>
+
+                <my-program-link
+                  type="program.type"
```
anwar6953: why not just pass in the `program` object, and extract the `type`, `name`, and `appId` on the other (receiving) end?

elfenheart: because on some places the program object might not include everything, such as in admin

cdap-ui/app/directives/program-link/program-link.less
elfenheart: resolved

ajainarayanan:
:+1: LGTM

PR #2932, opened 2015-06-11
(CDAP-2569) Always try to create the root ZK node on Master service startup

chtyim:
- The “root” node can be namespace by the zk connect string, hence may not exist yet.
  - E.g. host:2181/cdap
  - By creating “/“ with the given connection string, it make sure the relative root exist.

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
@@ -60,6 +60,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.io.Closeables;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.Service;
 import com.google.inject.Guice;
 import com.google.inject.Injector;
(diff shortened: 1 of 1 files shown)


henry-cask:
Question, so shouldnt the root already exist even if it namespaced the ZK string as "host:2181/cdap" ?

chtyim:
Apparently not the case for a newly installed cluster.

dereklwood:
the default cdap zk quorum from cdap-default.xml is ``127.0.0.1:2181/cdap``.  Nothing guarantees that the ``/cdap`` zknode will already exist, and we never create that znode externally in chef-land

cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
```diff
@@ -174,6 +178,10 @@ public void start() {
     metricsCollectionService.startAndWait();
     serviceStore.startAndWait();
     leaderElection.startAndWait();
+
+    // Tries to create the ZK root node (which can be namespaced through the zk connection string)
+    Futures.getUnchecked(ZKOperations.ignoreError(zkClient.create("/", null, CreateMode.PERSISTENT),
```
henry-cask: Probably we should check and add log the existence of the cdap root directory from the ZK connect (I think we could get it from config?) to make sure it is recorded we try to create root node.

chtyim: Hum, you mean log a message prior to this statement?

henry-cask: Yes, I am thinking it would help tracking additional things the code need to do when the environment is not set "properly".

chtyim: It's not exactly a requirement that the env needs to have the namespace setup.

PR #2933, opened 2015-06-11
Use solid neutral grey background

wolf31o2:
Screenshot:
screen shot 2015-06-11 at 12 48 48 pm


duddun:
:+1:

cdap-distributions/src/packer/scripts/lxde.sh
```diff
@@ -86,5 +86,11 @@ for i in cdap-ui cdap-docs eclipse idea lxterminal ; do
 done
 echo '@xterm -e "cat /etc/welcome.txt; bash -l"' >> /etc/xdg/lxsession/LXDE/autostart
 
+# Customize look and feel
+sed -i \
+    -e 's/wallpaper_mode=.*/wallpaper_mode=0/' \
+    -e 's/desktop_bg=.*/desktop_bg=#7f7f7f/' \
```
awholegunch: I'd put the color in a variable so named so that it easy to find and change at a later date. Which is for certain going to happen. We always want to change colors.

PR #2935, opened 2015-06-11
Fix SDK build and javadocs

awholegunch:
Removes the HTML Javadocs included in the SDK zip file, and cdap-api files no longer required as offline compilation is no longer supported.

Built and tested locally.

Fix for https://issues.cask.co/browse/CDAP-2730


sreevatsanraman:
LGTM :+1:

PR #2936, opened 2015-06-11
consolidate $destroy callback to delete binding instances and stop poll

elfenheart:
- improving apparent performance by making the stopPoll operation an asynchronous operation.
- consolidate the $destroy callback to both delete the instance binding and stop poll.


ajainarayanan:
:+1: LGTM

PR #2937, opened 2015-06-11
Refactor and fix start-debug command of CDAP CLI.

anwar6953:
The `StartDebugProgramCommand` seemed to be nearly identical to `StartProgramCommand`, with only minor differences, so refactored it.

Also, the `StartDebugProgramCommand` was not being added to the `StartProgramCommandSet`, which is fixed in this PR.

Build: http://builds.cask.co/browse/CDAP-DUT2028-1

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
@@ -21,57 +21,16 @@ import co.cask.cdap.cli.CLIConfig;
 import co.cask.cdap.cli.ElementType;
 import co.cask.cdap.cli.english.Article;
 import co.cask.cdap.cli.english.Fragment;
-import co.cask.cdap.cli.exception.CommandInputError;
-import co.cask.cdap.cli.util.AbstractAuthCommand;
-import co.cask.cdap.cli.util.ArgumentParser;
 import co.cask.cdap.client.ProgramClient;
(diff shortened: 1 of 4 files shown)


cdap-cli/src/main/java/co/cask/cdap/cli/command/StartDebugProgramCommand.java
```diff
@@ -21,57 +21,16 @@
...
-    } else {
-      // run with user-provided runtime args
-      Map<String, String> runtimeArgs = ArgumentParser.parseMap(runtimeArgsString);
-      programClient.start(appId, elementType.getProgramType(), programId, true, runtimeArgs);
-      output.printf("Successfully started %s '%s' of application '%s' with provided runtime arguments '%s'\n",
-                    elementType.getTitleName(), programId, appId, runtimeArgsString);
-    }
-
+    super(elementType, programClient, cliConfig);
+    this.isDebug = true;
```
gsps1: can you move isDebug to the constructor of the parent class ?

anwar6953: I just didn't think that's necessary, because then it would be possible to directly instantiate the parent class with either value (true or false), which would then be inconsistent with the `getPattern` and `getDescription` of that class.

Alternative is just to merge the two patterns/descriptions into one class, if that's what you mean?

gsps1:
lgtm :+1:

PR #2939, opened 2015-06-12
Fixes the scope of a function when no config is sent from node proxy …

ajainarayanan:
…(show text boxes for all properties)


elfenheart:
:+1: LGTM

PR #2945, opened 2015-06-12
WorkflowToken api changes

sagarkapare:
Initial API changes for the WorkflowToken. Opening PR to get early feedback.


sagarkapare:
@anew @chtyim Please help review the PR. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -20,9 +20,47 @@
...
  */
 public interface WorkflowToken {
+
+  /**
+   * Put the specified key-value entry in the {@link WorkflowToken}.
+   * The value is stored against the node at which it is being added.
+   * @param key   the key representing the entry
+   * @param value the value for the key
+   */
+  void putValue(String key, String value);
```
anew: I would call this setValue, because below you have a method getLastSetter(). Or call that one getLastPutter()? (that sounds strange)

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -20,9 +20,47 @@
...
+
+  /**
+   * Get the name of the node which most recently set the value for the key.
+   * Multiple nodes in the Workflow can add the same key. In this case, the name of the node that most
+   * recently added the key to the {@link WorkflowToken} will be returned.
+   * @param key the key to be searched
+   * @return the name of the node that most recently set the value for the key
+   */
+  @Nullable
+  String getLastSetterNode(String key);
```
anew: Call this getLastSetter()?

cdap-api/src/main/java/co/cask/cdap/api/workflow/TokenValueWithTimestamp.java
```diff
@@ -0,0 +1,39 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.workflow;
+
+/**
+ * Class representing the value for the key along with the timestamp inside the {@link WorkflowToken}.
+ */
+public final class TokenValueWithTimestamp {
```
anew: It is not the value of the token (token holds many key/values...). I think it is fine to name this just ValueWithTime

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Renamed to `setValue`.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Renamed to `getLastSetter`

cdap-api/src/main/java/co/cask/cdap/api/workflow/TokenValueWithTimestamp.java
sagarkapare: Renamed to `ValueWithTime`.

sagarkapare:
@anew Addressed comments. Please take a look again. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -20,9 +20,47 @@
...
+  @Nullable
+  String getValue(String key);
+
+  /**
+   * Same key can be added to the WorkflowToken by multiple nodes.
+   * This method returns the map of node name to the value which was set by the node for this key.
+   * @param key the key to be searched
+   * @return the map of node name to the value
+   */
+  @Nullable
```
chtyim: Should it be returning an empty map instead?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -28,6 +31,54 @@
...
+  }
+
+  @Override
+  public void setValue(String key, String value) {
+    Preconditions.checkNotNull(nodeName, "Node name cannot be null.");
+    WorkflowTokenValue tokenValue = keyTokenValue.get(key);
+    if (tokenValue == null) {
+      tokenValue = new WorkflowTokenValue();
+    }
+    tokenValue.putValue(nodeName, value);
```
chtyim: so you don't need to put the `tokenValue` back to the `keyTokenValue` map?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -28,6 +31,54 @@
 public class BasicWorkflowToken implements WorkflowToken {
   private Map<String, Map<String, Long>> mapReduceCounters;
 
+  private Map<String, WorkflowTokenValue> keyTokenValue = Maps.newHashMap();
```
chtyim: better call it `tokenValues` or `tokenValueMap`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowTokenValue.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Class representing the value for the key in {@link WorkflowToken}.
+ */
+final class WorkflowTokenValue {
+
+  private final Map<String, ValueWithTime> nodeTokenValue = Maps.newHashMap();
+
+  private String lastSetterNode = null;
```
chtyim: No need to assign to null.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowTokenValue.java
```diff
@@ -0,0 +1,57 @@
...
+    nodeTokenValue.put(nodeName, new ValueWithTime(value, System.currentTimeMillis()));
+    lastSetterNode = nodeName;
+  }
+
+  @Nullable
+  String getLastSetter() {
+    return lastSetterNode;
+  }
+
+  @Nullable
```
chtyim: How to distinguish the case where a key is missing vs key is there but with `null` value?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowTokenValue.java
```diff
@@ -0,0 +1,57 @@
...
+  String getValue() {
+    if (!nodeTokenValue.containsKey(lastSetterNode)) {
+      return null;
+    }
+    return nodeTokenValue.get(lastSetterNode).getValue();
+  }
+
+  Map<String, ValueWithTime> getAllValues() {
+    return nodeTokenValue;
```
chtyim: Shouldn't return the underlying map directly, as it can get modified outside of this class.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -20,9 +20,47 @@
...
 
 /**
- * Interface to represent the data that is transferred from one node to the next node in the {@link Workflow}
+ * Interface to represent the data that is transferred from one node to the next node in the {@link Workflow}.
  */
 public interface WorkflowToken {
+
+  /**
+   * Put the specified key-value entry in the {@link WorkflowToken}.
+   * The value is stored against the node at which it is being added.
```
anew: The token may store additional information about the context in which this key is being set, for example, the unique name of the workflow node.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Yes. Empty map makes client code easier. updated to return empty map.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Renamed to `tokenValueMap`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowTokenValue.java
sagarkapare: Removed the null assignment.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowTokenValue.java
sagarkapare: Returning immutable copy of the map now.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Fixed the documentation.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowTokenValue.java
sagarkapare: We can add another version of the `getValue` method as `getValueIfExist`. However not sure what will be the use case to distinguish between the missing key and `null` value.

sagarkapare:
@chtyim @anew Addressed review comments. Please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowTokenValue.java
sagarkapare: Added new method `containsKey` to the `WorkflowToken` API to test the presence of the key in the `WorkflowToken`

sagarkapare:
Design document for the WorkflowToken API -https://wiki.cask.co/display/CE/WorkflowToken+Enhancements

Jira: https://issues.cask.co/browse/CDAP-2709

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -16,18 +16,113 @@
...
+   * Assume that the branch containing node 5 finishes the execution first.
+   *
+   * Now the method invocation getValues("myKey") will return the list
+   * in which the keys will be ordered as 0-1-2-5-6-3-4-7-8.
+   *
+   * @param key the key to be searched
+   * @return the list of {@link NodeValueEntry} from node name to the value that node
+   * added for the input key
+   */
+  List<NodeValueEntry> getAll(String key);
```
sagarkapare: For now keeping the return type as `NodeValueEntry` instead of `Map<Entry, Entry>`.  `NodeValueEntry` has an advantage as by calling `NodeValueEntry.getNodeName()` we are making it explicit that the `nodeName` would be return, however in case of `Map<Entry, Entry>` in order to get `nodeName`, we will need to call `entry.getKey()`.  Using `Map<Entry, Entry>` may look more natural in the API. No strong opinion. Please let me know which one should be used.

sagarkapare:
@anew @chtyim Please help review.

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
```diff
@@ -123,4 +126,12 @@
    * @param resources Resources that each reducer should use.
    */
   void setReducerResources(Resources resources);
+
+  /**
+   * @return the {@link WorkflowToken} associated with the current {@link Workflow},
+   * if the {@link MapReduce} program is executed as a part of the Workflow.
+   *
```
anwar6953: extra newline

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -16,18 +16,113 @@
...
+   * @param key the key to be searched
+   * @return the value for the key
+   */
+  @Nullable
+  String get(String key);
+
+  /**
+   * This method is deprecated as of release 3.1. Instead to get the
+   * MapReduce counters from the WorkflowToken, use the flatten key prefixed
+   * by 'mr.counters.'.
```
anwar6953: Can you build the javadocs and make sure its properly formatted?
`mvn clean package javadoc:javadoc -pl cdap-api -am -DskipTests -P release`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -80,6 +82,7 @@ public BasicMapReduceContext(Program program,
                                MapReduceSpecification spec,
                                long logicalStartTime,
                                String workflowBatch,
+                               WorkflowToken workflowToken,
```
anwar6953: mark as `@Nullable`?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/DynamicMapReduceContext.java
```diff
@@ -74,6 +75,7 @@ public DynamicMapReduceContext(Program program,
                                  Arguments runtimeArguments,
                                  MapReduceSpecification spec,
                                  long logicalStartTime, String workflowBatch,
+                                 WorkflowToken workflowToken,
```
anwar6953: mark as `@Nullable`?

anwar6953: Hmm, other parameters of this constructor aren't appropriately marked. Maybe the inconsistency would be more confusing. Unsure.

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
```diff
@@ -0,0 +1,70 @@
...
+  public String toString() {
+    StringBuilder sb = new StringBuilder("NodeValueEntry {");
+    sb.append("nodeName=").append(nodeName);
+    sb.append(", value=").append(value);
+    sb.append('}');
+    return sb.toString();
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
bdmogal: Can we start using Java 7 ``Objects`` class for all ``equals``, ``hashCode`` and ``toString`` now?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
```diff
@@ -109,8 +112,8 @@ public BasicMapReduceContext build(MapReduceMetrics.TaskType type,
     MapReduceSpecification spec = program.getApplicationSpecification().getMapReduce().get(program.getName());
     BasicMapReduceContext context =
       new BasicMapReduceContext(program, type, RunIds.fromString(runId), taskId, runtimeArguments, datasets, spec,
-                                logicalStartTime, workflowBatch, discoveryServiceClient, metricsCollectionService,
-                                datasetFramework, adapterSpec, pluginInstantiator);
+                                logicalStartTime, workflowBatch, workflowToken, discoveryServiceClient,
```
bdmogal: Does ``WorkflowContext``or anything similar contain both ``WorkflowToken`` and the name of the workflow? If so, can we use that here instead of two distinct parameters representing a workflow?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
```diff
@@ -65,6 +67,7 @@ public BasicMapReduceContext build(MapReduceMetrics.TaskType type,
                                      String taskId,
                                      long logicalStartTime,
                                      String workflowBatch,
```
bdmogal: I think ``workflowBatch`` should also be ``@Nullable``. Although please see https://github.com/caskdata/cdap/pull/2945/files#r33312959

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
```diff
@@ -122,6 +126,20 @@ public String getWorkflowBatch() {
     return hConf.get(HCONF_ATTR_WORKFLOW_BATCH);
   }
 
+  private void setWorkflowToken(@Nullable WorkflowToken workflowToken) {
+    if (workflowToken != null) {
+      hConf.set(HCONF_ATTR_WORKFLOW_TOKEN, GSON.toJson(workflowToken));
+    }
+  }
+
+  public WorkflowToken getWorkflowToken() {
```
bdmogal: ``@Nullable``

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -137,6 +139,13 @@ public ProgramController run(final Program program, ProgramOptions options) {
                                 : System.currentTimeMillis();
 
     String workflowBatch = arguments.getOption(ProgramOptionConstants.WORKFLOW_BATCH);
+
+    WorkflowToken workflowToken = null;
```
bdmogal: need to initialize to ``null``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
```diff
@@ -54,13 +55,16 @@
  * The {@link RuntimeContext} is blocked until completion of the associated program.
  */
 public abstract class AbstractProgramWorkflowRunner implements ProgramWorkflowRunner {
+  private static final Gson GSON = new Gson();
+
+  private final Arguments userArguments;
+  private final Arguments systemArguments;
+  private final String nodeId;
+
```
bdmogal: remove empty line

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -28,18 +34,137 @@
 public class BasicWorkflowToken implements WorkflowToken {
   private Map<String, Map<String, Long>> mapReduceCounters;
 
+  private Map<String, List<NodeValueEntry>> tokenValueMap = Maps.newHashMap();
+
```
bdmogal: remove newlines between fields

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -28,18 +34,137 @@
...
+      }
+    }
+    if (other.getMapReduceCounters() != null) {
+      setMapReduceCounters(other.getMapReduceCounters());
+    }
+  }
+
+  @Override
+  public void put(String key, String value) {
+    Preconditions.checkNotNull(nodeName, "Node name cannot be null.");
```
bdmogal: This throws ``NullPointerException`` if the precondition is not satisfied. Is that ok? Or do you think it should be an ``IllegalArgumentException``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -28,18 +34,137 @@
...
+   * with the WorkflowToken on which the method is invoked.
+   * @param other the other WorkflowToken to be merged
+   */
+  void mergeToken(BasicWorkflowToken other) {
+    Map<String, List<NodeValueEntry>> otherTokenValueMap = other.tokenValueMap;
+
+    for (Map.Entry<String, List<NodeValueEntry>> entry : otherTokenValueMap.entrySet()) {
+      if (!tokenValueMap.containsKey(entry.getKey())) {
+        // Key is newly added to the other WorkflowToken
+        tokenValueMap.put(entry.getKey(), Lists.<NodeValueEntry>newArrayList());
```
bdmogal: hmm, am I missing something? shouldn't this be ``entry.getValue()`` instead of empty arraylist?

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
sagarkapare: Removed.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Updated the javadoc and verified the formatting.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/DynamicMapReduceContext.java
sagarkapare: Marked them as `@Nullable`

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
sagarkapare: Not sure how the `Objects` class would be useful here for the class which only contain primitive types.
For example, current `equals` method return `nodeName.equals(that.nodeName) && value.equals(that.value)`. Changing it to `Objects.equals(this.nodeName, that.nodeName) && Objects.equals(this.value, that.value)` would be more verbose.. no?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
sagarkapare: `workflowBatch` carries the name of the current program which Workflow wants to run, which is not included in the `WorkflowContext`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
sagarkapare: We need to initialize the local variables.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/AbstractProgramWorkflowRunner.java
sagarkapare: Removed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Removed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Good point! I think `IllegalStateException` would be correct to throw here. Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: All the values from the `entry.getValue()` will get added to the `tokenValueMap` in the for loop following on line 58.

sagarkapare:
@bdmogal @anwar6953 Thanks for the review comments. Updated the code per feedback. Please take a look again.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
bdmogal: correct, although it might be better to use ``entry.getValue()`` here followed by ``continue``? This logic is slightly complex. Trying to see how we can remove some complexity.

sagarkapare: ohh ok. yeah we can do that. updated the code.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
chtyim: What is `Map<Entry, Entry>`? Do you mean `List<Map.Entry<String, String>>`? Or something else?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -16,18 +16,136 @@
...
+   * Get the {@link Map} of key-values that were added to the {@link WorkflowToken}
+   * by specific node.
+   * <p>
+   * This method also accepts the optional prefix parameter. When
+   * supplied, the returned map would be filtered by the keys prefixed by the input prefix.
+   * @param nodeName the unique name of the node
+   * @param prefix optional prefix to filter the keys
+   * @return the map of key-values that were added by the specified node
+   */
+  Map<String, String> getAllFromNode(String nodeName, @Nullable String prefix);
```
chtyim: Is the prefix a string prefix or dot terminated prefix? For example, if prefix = "a.b", does it match with "a.b*" or only match with "a.b.*"?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
chtyim: The name of the current program? So it's not the unique name (node name)?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -27,19 +32,139 @@
  */
 public class BasicWorkflowToken implements WorkflowToken {
   private Map<String, Map<String, Long>> mapReduceCounters;
```
chtyim: Why need to keep it separate? Can it be just a filtered map view on the `tokenValueMap` for the `getMapReduceCounters` method?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -27,19 +32,139 @@
...
+  private Map<String, List<NodeValueEntry>> tokenValueMap = Maps.newHashMap();
```
chtyim: Can this be final? Instead of the setter method, you can have a replace method that will clear the map and replace the content.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -27,19 +32,139 @@
...
+   * with the WorkflowToken on which the method is invoked.
+   * @param other the other WorkflowToken to be merged
+   */
+  void mergeToken(BasicWorkflowToken other) {
+    Map<String, List<NodeValueEntry>> otherTokenValueMap = other.tokenValueMap;
+
+    for (Map.Entry<String, List<NodeValueEntry>> entry : otherTokenValueMap.entrySet()) {
+      if (!tokenValueMap.containsKey(entry.getKey())) {
+        // Key is newly added to the other WorkflowToken
+        tokenValueMap.put(entry.getKey(), entry.getValue());
```
chtyim: The other list may not be immutable. It's not safe to just use the List returned by getValue. Also, after the merge, you may append new entries to the List right? Meaning it will modify the List inside another object?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -27,19 +32,139 @@
...
+    }
+    nodeValueList.add(new NodeValueEntry(nodeName, value));
+  }
+
+  @Override
+  public List<NodeValueEntry> getAll(String key) {
+    if (containsKey(key)) {
+      return ImmutableList.copyOf(tokenValueMap.get(key));
+    }
+    return Lists.newArrayList();
```
chtyim: Better be `ImmutableList.of();`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
chtyim: Another question not related to the field signature. I think it's better for the actual instance of the List be immutable.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -27,19 +32,139 @@
...
+  }
 
   @Nullable
   @Override
+  public String get(String key) {
+    if (!containsKey(key)) {
+      return null;
+    }
+    List<NodeValueEntry> nodeValueList = tokenValueMap.get(key);
+    if (nodeValueList.size() == 0) {
```
chtyim: Use `isEmpty()` instead.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Yes. Keeping the return type of `List<NodeValueEntry>` instead of `List<Map.Entry<String, String>`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
sagarkapare: `workflowBatch` is the part of existing code. We pass the name of the program in it, so that we can correctly initialize the `Program` from its specifications. It won't be the unique name of the node.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Its a string prefix. so if prefix = "a.b", it will match with "a.b", "a.b." and "a.bc" etc. We find the key matching prefix if `key.startsWith(prefix)` is true.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Since the `tokenValueMap` maintains flatten counter names as `mr.counters.groupname.countername`, we won't be able to reconstruct the group name and counter name from it. I thought of maintaining the map of all counter group names and counter names and then use them to recreate the `mapReduceCounters`, however it will not work if the user has added the custom counter.

chtyim: but the mapping is known to us when the raw MapReduce counters are provided this class right? However, it's not a very big deal to keep it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Yes. Agree. Changed the code to create the new list and adding the elements to it explicitly on the line 58

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Made the `tokenValueMap` final. 
Can you please explain the comment 
 >> I think it's better for the actual instance of the List be immutable.
Since we keep adding the `NodeValueEntry` to the list if the `put` on the same key is made by subsequent nodes, not sure how making the List immutable will work.

sagarkapare:

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
bdmogal: Maybe not for the equals method. How about ``hashCode`` and ``toString``?

sagarkapare: Since the class `NodeValueEntry` itself only contains 2 members of type `String`, not sure how the `Objects.toString` would be better either. 

For example, current `toString` directly uses the `nodeName` and `value` instances. We should not use `Objects.toString(nodeName)` and `Objects.toString(value)` right? 

Had `nodeName` and `value` been the instances of another object, then yes you are right we would require `Objects.toString`.

Same with the `hashCode` method.

Please let me know if I am missing anything here.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -16,18 +16,136 @@
...
 import javax.annotation.Nullable;
 
 /**
- * Interface to represent the data that is transferred from one node to the next node in the {@link Workflow}
+ * Interface to represent the data that is transferred from one node to the next nodes in the {@link Workflow}.
  */
 public interface WorkflowToken {
+
+  /**
+   * Put the specified key-value entry in the {@link WorkflowToken}.
```
anew: in/into

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -16,18 +16,136 @@
...
-   * if the counters are not set.
+   * Same key can be added to the WorkflowToken by multiple nodes.
+   * This method returns the {@link List} of {@link NodeValueEntry}, where
+   * each entry represents the unique node name and the value that it set
+   * for the specified key.
+   * <p>
+   * The list maintains the order in which the values were
+   * inserted in the WorkflowToken for a specific key except in the case of fork
+   * and join. In case of fork in the Workflow, copies of the WorkflowToken are made
+   * and passed it along the each branch. At the join, all copies of the
```
anew: remove "it"

anew: I would not go into that much detail, as it limits us as to how we can change this in the future. All I would say here is that order is guaranteed for nodes within the same branch, but not across different branches.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -16,18 +16,136 @@
...
+   */
+  Map<String, String> getAllFromNode(String nodeName, @Nullable String prefix);
+
+  /**
+   * Get the value set for the specified key by the specified node.
+   * @param key the key to be searched
+   * @param nodeName the name of the node
+   * @return the value set for the key by nodeName
+   */
+  @Nullable
```
anew: How about moving this up a little, to come right after put().

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
anew: I like Terence's idea of saying it is along dot boundaries. So prefix a.b will match a.b and a.b.c, but not a.bc

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -16,18 +16,136 @@
...
+   * added for the input key
+   */
+  List<NodeValueEntry> getAll(String key);
+
+  /**
+   * Get the {@link Map} of key-values that were added to the {@link WorkflowToken}
+   * by specific node.
+   * <p>
+   * This method also accepts the optional prefix parameter. When
+   * supplied, the returned map would be filtered by the keys prefixed by the input prefix.
```
anew: would be/is

sagarkapare: Updated.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Moved up.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Yes. I agree. Removed the extra information.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Changed.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Sounds good! Updated the code. 
So Prefix "a.b" will match with the key "a.b" or any key starting with the "a.b.", however it will not match with the key "a.bc". To retrieve the MapReduce counters from the particular node use the prefix "mr.counters".

sagarkapare:
@anew Addressed the comments. Please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
chtyim: So what is this name? Do you have an example?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
chtyim: Regarding my previous comment about immutable list, currently you store the list as ArrayList (hence mutable). My question was, would it be better to make it as an ImmutableList / UnmodifiableList (the logic of constructing the map would be slight different)? I feel more comfortable with immutable collections and I don't see the need of mutability when construction is done.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -27,19 +32,138 @@
...
+  public void put(String key, String value) {
+    if (nodeName == null) {
+      throw new IllegalStateException("Node name cannot be null.");
+    }
+    List<NodeValueEntry> nodeValueList = tokenValueMap.get(key);
+    if (nodeValueList == null) {
+      nodeValueList = Lists.newArrayList();
+      tokenValueMap.put(key, nodeValueList);
+    }
+    nodeValueList.add(new NodeValueEntry(nodeName, value));
```
chtyim: So with this logic, it means a node can set with the same key twice with different values and both will be retained, but then the `getAllFromNode` method will only return the latest value, right? I don't find it stated clearly about this behavior in the javadoc.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -27,19 +32,138 @@
...
+        return nodeValue.getValue();
+      }
+    }
+    return null;
+  }
+
+  @Nullable
+  @Override
+  public String get(String key) {
+    if (!containsKey(key)) {
```
chtyim: you actually don't need this check. Just do a get from the map and it returns `null`, then return `null`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/AbstractMapReduceContextBuilder.java
sagarkapare: So for example when we run the `PurchaseHistoryWorkflow`, workflowBatch will be the `PurchaseHistoryBuilder`, the MapReduce program run by the Workflow. In `AbstractMapReduceContextBuilder`, while building the MapReduce context, if workflowBatch is not null, then we create the `Program` by fetching the specification for the MapReduce program from application specifications.
Corresponding code snippet in `AbstractMapReduceContextBuilder`:
```java
    // Initializing Program
    Program program = mrProgram;

    // See if it was launched from Workflow; if it was, change the Program.
    if (workflowBatch != null) {
      MapReduceSpecification mapReduceSpec = program.getApplicationSpecification().getMapReduce().get(workflowBatch);
      Preconditions.checkArgument(mapReduceSpec != null, "Cannot find MapReduceSpecification for %s", workflowBatch);
      program = new WorkflowMapReduceProgram(program, mapReduceSpec);
    }
```

chtyim: So it's the program Id (the string part) of the MapReduce program, right? The variable name seems really bad.

sagarkapare: yes. will change the name to something similar to `programNameInWorkflow`?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: oh ok. I think while inserting into the List, if the entry for the nodeName already exists for the specific key, that entry should be replaced by the new one. So that for any key, we will always have only one entry for the specific node. I will update the code accordingly.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: done.

sagarkapare:
@chtyim @bdmogal Please take a look again. 
Fixed following things - 
1. Renamed the variable `workflowBatch` to the `programNameInWorkflow`.
2. Created the constructor for the `BasicWorkflowToken` which accepts the individual maps.
3. Removed prefix from the api and filed jira - https://issues.cask.co/browse/CDAP-2894

New build is running here - http://builds.cask.co/browse/CDAP-DUT2031-6

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
bdmogal: I see. ignore.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -132,8 +136,17 @@ public long getLogicalStartTime() {
   /**
    * Returns the name of the Batch job when running inside workflow. Otherwise, return null.
    */
-  public String getWorkflowBatch() {
-    return workflowBatch;
+  public String getProgramNameInWorkflow() {
+    return programNameInWorkflow;
```
bdmogal: is this ``@Nullable``? If so, mark both the constructor param and this method.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -27,27 +32,165 @@
...
+      // Only add those NodeValueEntry to the merged token which already do not exist.
+      for (NodeValueEntry otherNodeValueEntry : otherTokenValueMap.get(entry.getKey())) {
+        boolean otherNodeValueEntryExist = false;
+        for (NodeValueEntry thisNodeValueEntry : tokenValueMap.get(entry.getKey())) {
+          if (thisNodeValueEntry.equals(otherNodeValueEntry)) {
+            otherNodeValueEntryExist = true;
+            break;
+          }
+        }
+        if (!otherNodeValueEntryExist) {
```
bdmogal: will this condition ever be ``false``? Please check. Else remove.

cdap-app-fabric/src/test/java/co/cask/cdap/ConditionalWorkflowApp.java
```diff
@@ -116,6 +128,31 @@ public void beforeSubmit(MapReduceContext context) throws Exception {
...
+    // TODO: CDAP-2894 MapReduce counters should be available implicitly from the WorkflowToken.
+    // We yet to figure out the scheme.
+    @Override
+    public void onFinish(boolean succeeded, MapReduceContext context) throws Exception {
+      if (!succeeded) {
+        return;
+      }
+
+      WorkflowToken workflowToken = context.getWorkflowToken();
+      if (workflowToken == null) {
```
bdmogal: nit: this is kind of ugly. apps should ideally not have to do this check. But unsure how we can fix it (if we can). Maybe lets report a jira?

bdmogal:
couple of minor nits. Other than that, LGTM. Presume that you have spoken with @chtyim about a couple of his comments that seem unresolved.

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT2031-8. Merging.

PR #2946, opened 2015-06-12
Fixes view of non-explorable dataset (makes it static instead of dynamic states)

ajainarayanan:
Fixes navigating to non-explorable datasets. Instead of mixing up states we just go to a static tab (programs) for non-explorable datasets.


elfenheart:
:+1: LGTM

PR #2947, opened 2015-06-12
Use POST in RanksServiceHandler for better UI integration

rohitsinha54:
Build: http://builds.cask.co/browse/CDAP-RBT305-1


henry-cask:
Why is it actually change to POST? To avoid the client cache?

rohitsinha54:
@henry-cask : Earlier we were doing with query params which does not have support through our current UI. So we changed it to use POST as the data is an url which will have / etc.

henry-cask:
+1

LGTM

rohitsinha54:
Will merge as soon as the build passes.

rohitsinha54:
Build : http://builds.cask.co/browse/CDAP-RBT305-2

sreevatsanraman:
LGTM :+1:

PR #2940, opened 2015-06-12
Add CDAP-UI Dependencies to license files

awholegunch:
As 3.0.1 has now been merged back to develop, this PR now includes just the 3.1.0 versions of the PDFs. However, there are improvements to the Python script requested. I will make them in this PR.

Staged at: 

- [Trademarks, Licenses, and Dependencies](http://docs-staging.cask.co/cdap/3.1.0-SNAPSHOT-feature-3.1_doc_ui_licenses/en/reference-manual/licenses/index.html)
- [Cask Data Application Platform UI Dependencies](http://docs-staging.cask.co/cdap/3.1.0-SNAPSHOT-feature-3.1_doc_ui_licenses/en/reference-manual/licenses/cdap-ui-dependencies.html)

Fix for https://issues.cask.co/browse/CDAP-2582

PR #2948, opened 2015-06-12
Adding CDAP-UI Dependencies to license files.

awholegunch:
A copy of https://github.com/caskdata/cdap/pull/2940, but for release/3.0.

Adds processing of the cdap-ui dependencies, using the checked-in ``bower`` and ``npm`` references, ``bower.json`` and ``package.json``

Staged at: 

- [Trademarks, Licenses, and Dependencies](http://docs-staging.cask.co/cdap/3.0.1-feature-r3.0.1_doc_ui_licenses/en/reference-manual/licenses/index.html)
- [Cask Data Application Platform UI Dependencies](http://docs-staging.cask.co/cdap/3.0.1-feature-r3.0.1_doc_ui_licenses/en/reference-manual/licenses/cdap-ui-dependencies.html)

Fix for https://issues.cask.co/browse/CDAP-2582.


cdap-docs/tools/licenses/licenses.py
```diff
@@ -207,13 +256,118 @@ def master_print():
...
+        if CDAP_UI_DEPENDENCIES_KEY in data.keys():
+            for dependency in data[CDAP_UI_DEPENDENCIES_KEY]:
+                if not dependency.startswith(CDAP_UI_CASK_DEPENDENCIES):
+                    version = data[CDAP_UI_DEPENDENCIES_KEY][dependency]
+                    if master_libs_dict.has_key(dependency):
+                        # Look up reference in dictionary
+                        cdap_ui_dict[dependency] = master_libs_dict[dependency]
+                    else:
+                        missing_libs_dict[dependency] = (type, version)
+
```
duddun: What happens in the code below, if "if CDAP_UI_DEPENDENCIES_KEY ..." is false?

cdap-docs/tools/licenses/licenses.py
```diff
@@ -207,13 +256,118 @@ def master_print():
...
+    
+    for type in CDAP_UI_SOURCES.keys():
+        
+        source = CDAP_UI_SOURCES[type][0]
+        json_path = os.path.join(SCRIPT_DIR_PATH, source)
+        print "Reading '%s' dependencies file:\n%s" % (type, json_path)
+        with open(json_path) as data_file:    
+            data = json.load(data_file)
+            
+        if CDAP_UI_DEPENDENCIES_KEY in data.keys():
```
duddun: What happens after line 305, if this returns false?

cdap-docs/tools/licenses/licenses.py
```diff
@@ -207,13 +256,118 @@ def master_print():
...
+            for dependency in data[CDAP_UI_DEPENDENCIES_KEY]:
+                if not dependency.startswith(CDAP_UI_CASK_DEPENDENCIES):
```
duddun: Same as comment above (what happens if it starts with ...)

cdap-docs/tools/licenses/licenses.py
awholegunch: The missing_libs_dict will remain empty. The count will be zero. The block beginning at L308 will be skipped. And an empty array will be returned. All expected behavior.

cdap-docs/tools/licenses/licenses.py
awholegunch: If the dependency doesn't start with the CDAP_UI_CASK_DEPENDENCIES ("cdap-"), then it gets processed by the code. Else wise, it gets skipped. It ignores all the cdap- dependencies.

awholegunch: Do you think that these static variables need better comments to explain them and the code, and perhaps add Python doc comments? (Not that we build Python documentation...)

duddun:
LGTM

PR #2949, opened 2015-06-12
CDAP-2734 Partition-level metadata support for PartitionedFileSet and…

anwar6953:
CDAP-2734 Partition-level metadata support for PartitionedFileSet and TimePartitionedFileSet.

https://issues.cask.co/browse/CDAP-2734
http://builds.cask.co/browse/CDAP-DUT2017-19

A couple of comments by Andreas on a previous commit:
https://github.com/caskdata/cdap/commit/df1476486a0cd5e6d42703fff4ffef49d7299bb5#commitcomment-11645773 - resolved.
https://github.com/caskdata/cdap/commit/df1476486a0cd5e6d42703fff4ffef49d7299bb5#commitcomment-11645814 - resolved.
https://github.com/caskdata/cdap/commit/df1476486a0cd5e6d42703fff4ffef49d7299bb5#commitcomment-11645825 - resolved.
https://github.com/caskdata/cdap/commit/df1476486a0cd5e6d42703fff4ffef49d7299bb5#commitcomment-11645838 - resolved.
https://github.com/caskdata/cdap/commit/df1476486a0cd5e6d42703fff4ffef49d7299bb5#commitcomment-11645848 - opened JIRA https://issues.cask.co/browse/CDAP-2784

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.dataset.lib;
+
+import com.google.common.base.Objects;
+
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * Contains metadata associated with a particular {@link Partition} of a {@link @PartitionedFileSet}
+ */
+public class PartitionMetadata implements Iterable<Map.Entry<String, String>> {
+  private final Map<String, String> metadata;
+
+  public PartitionMetadata(Map<String, String> metadata) {
+    this.metadata = metadata;
+  }
+
+  public String get(String key) {
+    return metadata.get(key);
+  }
+
+  public Map<String, String> getAsMap() {
+    return metadata;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    PartitionMetadata that = (PartitionMetadata) o;
+
+    return Objects.equal(this.metadata, that.metadata);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(metadata);
+  }
+
+  @Override
+  public Iterator<Map.Entry<String, String>> iterator() {
+    return metadata.entrySet().iterator();
+  }
+}
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
@@ -18,6 +18,8 @@ package co.cask.cdap.api.dataset.lib;
 
 import com.google.common.base.Objects;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 
@@ -28,14 +30,14 @@ public class PartitionMetadata implements Iterable<Map.Entry<String, String>> {
   private final Map<String, String> metadata;
 
   public PartitionMetadata(Map<String, String> metadata) {
-    this.metadata = metadata;
+    this.metadata = Collections.unmodifiableMap(new HashMap<>(metadata));
   }
 
   public String get(String key) {
     return metadata.get(key);
   }
 
-  public Map<String, String> getAsMap() {
+  public Map<String, String> asMap() {
     return metadata;
   }
 
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
@@ -16,12 +16,11 @@
 
 package co.cask.cdap.api.dataset.lib;
 
-import com.google.common.base.Objects;
-
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 /**
  * Contains metadata associated with a particular {@link Partition} of a {@link @PartitionedFileSet}
@@ -33,10 +32,19 @@ public class PartitionMetadata implements Iterable<Map.Entry<String, String>> {
     this.metadata = Collections.unmodifiableMap(new HashMap<>(metadata));
   }
 
+  /**
+   * Perform a get on the underlying map of user-defined metadata.
+   * @param key the key to use in the lookup on the underlying map.
+   * @return the value returned by the underlying map.
+   */
+  @Nullable
   public String get(String key) {
     return metadata.get(key);
   }
 
+  /**
+   * @return the underlying map of user-defined metadata.
+   */
   public Map<String, String> asMap() {
     return metadata;
   }
@@ -52,12 +60,12 @@ public class PartitionMetadata implements Iterable<Map.Entry<String, String>> {
 
     PartitionMetadata that = (PartitionMetadata) o;
 
-    return Objects.equal(this.metadata, that.metadata);
+    return this.metadata.equals(that.metadata);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(metadata);
+    return metadata.hashCode();
   }
 
   @Override
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionOutput.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionOutput.java
@@ -16,6 +16,8 @@
 
 package co.cask.cdap.api.dataset.lib;
 
+import java.util.Map;
+
 /**
  * Represents a partition of a partitioned file set for writing.
  */
@@ -29,5 +31,5 @@ public interface PartitionOutput extends Partition {
   /**
    * Sets the metadata of a partition, when creating a new partition.
    */
-  void setMetadata(PartitionMetadata metadata);
+  void setMetadata(Map<String, String> metadata);
 }
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionOutput.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionOutput.java
@@ -27,7 +27,7 @@ public interface PartitionOutput extends Partition {
   void addPartition();
 
   /**
-   * Sets the metadata of a partition.
+   * Sets the metadata of a partition, when creating a new partition.
    */
   void setMetadata(PartitionMetadata metadata);
 }
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionOutput.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionOutput.java
@@ -25,4 +25,9 @@ public interface PartitionOutput extends Partition {
    * Add the partition to the partitioned file set.
    */
   void addPartition();
+
+  /**
+   * Sets the metadata of a partition.
+   */
+  void setMetadata(PartitionMetadata metadata);
 }
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
@@ -52,12 +52,19 @@ public interface PartitionedFileSet extends Dataset, InputFormatProvider, Output
    * Add a partition for a given partition key, stored at a given path (relative to the file set's base path),
    * with the given metadata.
    */
-  void addPartition(PartitionKey key, String path, PartitionMetadata metadata);
+  void addPartition(PartitionKey key, String path, Map<String, String> metadata);
 
   /**
-   * Updates the metadata for a particular partition.
+   * Adds a new metadata entry for a particular partition.
+   * Note that existing entries can not be updated; {@link IllegalArgumentException} will be thrown in that case.
    */
-  void updateMetadata(PartitionKey key, PartitionMetadata metadata);
+  void addMetadata(PartitionKey key, String metadataKey, String metadataValue);
+
+  /**
+   * Adds a set of new metadata entries for a particular partition
+   * Note that existing entries can not be updated; {@link IllegalArgumentException} will be thrown in that case.
+   */
+  void addMetadata(PartitionKey key, Map<String, String> metadata);
 
   /**
    * Remove a partition for a given partition key.
@@ -68,14 +75,14 @@ public interface PartitionedFileSet extends Dataset, InputFormatProvider, Output
    * Return the partition for a specific partition key.
    */
   @Nullable
-  Partition getPartition(PartitionKey key);
+  PartitionDetail getPartition(PartitionKey key);
 
   /**
    * Return all partitions matching the partition filter.
    * @param filter If non null, only partitions that match this filter are returned. If null,
    *               all partitions are returned.
    */
-  Set<Partition> getPartitions(@Nullable PartitionFilter filter);
+  Set<PartitionDetail> getPartitions(@Nullable PartitionFilter filter);
 
   /**
    * Return a partition output for a specific partition key, in preparation for creating a new partition.
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
@@ -56,13 +56,15 @@ public interface PartitionedFileSet extends Dataset, InputFormatProvider, Output
 
   /**
    * Adds a new metadata entry for a particular partition.
-   * Note that existing entries can not be updated; {@link IllegalArgumentException} will be thrown in that case.
+   * Note that existing entries can not be updated.
+   * @throws IllegalArgumentException in case an attempt is made to update existing entries.
    */
   void addMetadata(PartitionKey key, String metadataKey, String metadataValue);
 
   /**
    * Adds a set of new metadata entries for a particular partition
-   * Note that existing entries can not be updated; {@link IllegalArgumentException} will be thrown in that case.
+   * Note that existing entries can not be updated.
+   * @throws IllegalArgumentException in case an attempt is made to update existing entries.
    */
   void addMetadata(PartitionKey key, Map<String, String> metadata);
 
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
@@ -49,6 +49,17 @@ public interface PartitionedFileSet extends Dataset, InputFormatProvider, Output
   void addPartition(PartitionKey key, String path);
 
   /**
+   * Add a partition for a given partition key, stored at a given path (relative to the file set's base path),
+   * with the given metadata.
+   */
+  void addPartition(PartitionKey key, String path, PartitionMetadata metadata);
+
+  /**
+   * Updates the metadata for a particular partition.
+   */
+  void updateMetadata(PartitionKey key, PartitionMetadata metadata);
+
+  /**
    * Remove a partition for a given partition key.
    */
   void dropPartition(PartitionKey key);
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
@@ -19,6 +19,7 @@ package co.cask.cdap.api.dataset.lib;
 import co.cask.cdap.api.annotation.Beta;
 import co.cask.cdap.api.data.batch.InputFormatProvider;
 import co.cask.cdap.api.data.batch.OutputFormatProvider;
+import co.cask.cdap.api.dataset.DataSetException;
 import co.cask.cdap.api.dataset.Dataset;
 
 import java.util.Map;
@@ -57,14 +58,14 @@ public interface PartitionedFileSet extends Dataset, InputFormatProvider, Output
   /**
    * Adds a new metadata entry for a particular partition.
    * Note that existing entries can not be updated.
-   * @throws IllegalArgumentException in case an attempt is made to update existing entries.
+   * @throws DataSetException in case an attempt is made to update existing entries.
    */
   void addMetadata(PartitionKey key, String metadataKey, String metadataValue);
 
   /**
    * Adds a set of new metadata entries for a particular partition
    * Note that existing entries can not be updated.
-   * @throws IllegalArgumentException in case an attempt is made to update existing entries.
+   * @throws DataSetException in case an attempt is made to update existing entries.
    */
   void addMetadata(PartitionKey key, Map<String, String> metadata);
 
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
@@ -59,12 +59,14 @@ public interface TimePartitionedFileSet extends PartitionedFileSet {
   /**
    * Adds a new metadata entry for a particular partition.
    * Note that existing entries can not be updated.
+   * @throws IllegalArgumentException in case an attempt is made to update existing entries.
    */
   void addMetadata(long time, String metadataKey, String metadataValue);
 
   /**
    * Adds a set of new metadata entries for a particular partition
    * Note that existing entries can not be updated.
+   * * @throws IllegalArgumentException in case an attempt is made to update existing entries.
    */
   void addMetadata(long time, Map<String, String> metadata);
 
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
@@ -50,6 +50,19 @@ public interface TimePartitionedFileSet extends PartitionedFileSet {
    */
   void addPartition(long time, String path);
 
+  // Why are some of these methods methodNameByTime, while others are just methodName?
+  // (why are some suffixed by 'ByTime')
+  /**
+   * Add a partition for a given time, stored at a given path (relative to the file set's base path),
+   * with given metadata.
+   */
+  void addPartition(long time, String path, PartitionMetadata metadata);
+
+  /**
+   * Updates the metadata for a particular partition.
+   */
+  void updateMetadata(long time, PartitionMetadata metadata);
+
   /**
    * Remove a partition for a given time.
    */
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
@@ -17,6 +17,7 @@
 package co.cask.cdap.api.dataset.lib;
 
 import co.cask.cdap.api.annotation.Beta;
+import co.cask.cdap.api.dataset.DataSetException;
 
 import java.util.Collection;
 import java.util.Map;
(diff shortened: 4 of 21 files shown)


anwar6953:
@anew Regarding this comment: https://github.com/caskdata/cdap/commit/df1476486a0cd5e6d42703fff4ffef49d7299bb5#commitcomment-11645814
You think it'd be ok to remove those checks, or turn them into assertions, since each time we write to a rowKey, we always write the path. So it really is an error if its missing (as opposed to simply returning null).

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Map;
+
+/**
+ * Contains metadata associated with a particular {@link Partition} of a {@link @PartitionedFileSet}
+ */
+public class PartitionMetadata implements Iterable<Map.Entry<String, String>> {
+  private final Map<String, String> metadata;
+
+  public PartitionMetadata(Map<String, String> metadata) {
+    this.metadata = metadata;
```
chtyim: Better do a defensive copy of the map:

```java
this.metadata = new HashMap<>(metadata);
```

chtyim: maybe wrap it as unmodifiable as well:

```java
this.metadata = Collections.unmodifiableMap(new HashMap<>(metadata));
```

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public String get(String key) {
+    return metadata.get(key);
+  }
+
+  public Map<String, String> getAsMap() {
```
chtyim: Just `asMap`.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionOutput.java
```diff
@@ -25,4 +25,9 @@
    * Add the partition to the partitioned file set.
    */
   void addPartition();
+
+  /**
+   * Sets the metadata of a partition.
+   */
+  void setMetadata(PartitionMetadata metadata);
```
chtyim: What is it for? When it will get called and who calls it?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -113,10 +117,41 @@ public Partitioning getPartitioning() {
...
 
-  protected void addPartition(PartitionKey key, String path, boolean addToExplore) {
+  @Override
+  public void addPartition(PartitionKey key, String path, PartitionMetadata metadata) {
+    addPartition(key, path, true, metadata);
+  }
+
+  @Override
+  public void updateMetadata(PartitionKey key, PartitionMetadata metadata) {
+    final byte[] rowKey = generateRowKey(key, partitioning);
```
chtyim: why need `final`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -113,10 +117,41 @@ public Partitioning getPartitioning() {
...
+    Row row = partitionsTable.get(rowKey);
+    if (row == null) {
+      throw new DataSetException(String.format("Dataset '%s' does not have a partition for key: %s", getName(), key));
+    }
+
+    // delete all the old metadata and then write the new metadata
+    Delete delete = new Delete(rowKey);
+    for (Map.Entry<byte[], byte[]> entry : row.getColumns().entrySet()) {
```
chtyim: To be consistent with the usage of "addMetadataToPut" method, I think this for loop should also be abstracted out in a method.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: I guess, I just took this from one of the other methods, all of which had their `byte[]` final.

chtyim: remove it if not needed.

anwar6953: done

anwar6953: Also, why do you feel that it is not a good idea, but it is a good idea in fields of a class?
Wouldn't it be helpful to convey to other readers that this variable won't ever change its reference?

anew:
@ali regarding your question: yes I think that would be ok.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anew: normally we have to make them final if they are referenced from an anonymous class. Possibly where you copied it from had it final for that reason?

anew: My personal opinion is that it is ok to declare a method variable final even if it is not strictly needed to be final. I a not sure whether we have a rule about that in the coding standards.

anew:
I feel like we want to be able to set certain meta data by the system (such as, time of creation, lineage, etc.) Such fields should never be settable by the user, or changeable by the user by setting new meta data. So perhaps we only allow the user to add properties, but the Metadata object has additional methods to get creation time etc. This would still allow the user to tag partitions after they were created.

anwar6953:
@anew 
Were you thinking that for system metadata entries, such as creation time, that column could be recorded when the row (for the partition) is created?
And how do you expect lineage and other system metadata to be recorded?

anew:
@anwar6953 yes, the creation time should be set when the partition is first added. I opened https://issues.cask.co/browse/CDAP-2746 for that

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/BasePartition.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package co.cask.cdap.api.dataset.lib;
+
+import org.apache.twill.filesystem.Location;
+
+/**
+ * Represents a partition of a partitioned file set.
+ */
+public interface BasePartition {
```
chtyim: I think it's better for this as `Partition` and the one with metadata (extends from this one) as `PartitionDetails`

chtyim: or `PartitionDetail`.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/BaseTimePartition.java
```diff
@@ -0,0 +1,28 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.dataset.lib;
+
+/**
+ * Represents a partition of a time partitioned file set.
+ */
+public interface BaseTimePartition extends BasePartition {
```
chtyim: Why need this? The `TimePartition` is not good enough?

anwar6953: There is `TimePartitionOutput` also, which doesn't have meta, same as the non-Time versions.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
```diff
@@ -0,0 +1,67 @@
...
+ * Contains metadata associated with a particular {@link Partition} of a {@link @PartitionedFileSet}
+ */
+public class PartitionMetadata implements Iterable<Map.Entry<String, String>> {
+  private final Map<String, String> metadata;
+
+  public PartitionMetadata(Map<String, String> metadata) {
+    this.metadata = Collections.unmodifiableMap(new HashMap<>(metadata));
+  }
+
+  public String get(String key) {
```
chtyim: This method should be `@Nullable`. Also missing javadoc.

Also, do we allow `null` value in the map? If not, then the corresponding set/add/put method should check and bounce on null value.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
```diff
@@ -0,0 +1,67 @@
...
+    return metadata.get(key);
+  }
+
+  public Map<String, String> asMap() {
```
chtyim: javadoc.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
```diff
@@ -49,6 +49,24 @@
   void addPartition(PartitionKey key, String path);
 
   /**
+   * Add a partition for a given partition key, stored at a given path (relative to the file set's base path),
+   * with the given metadata.
+   */
+  void addPartition(PartitionKey key, String path, Map<String, String> metadataEntries);
```
chtyim: Name the parameter `metadata` to be consistent with the `PartitionOuput.setMetadata`.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
```diff
@@ -49,6 +49,24 @@
...
+
+  /**
+   * Adds a new metadata entry for a particular partition.
+   * Note that existing entries can not be updated.
+   */
+  void addMetadataEntry(PartitionKey key, String metadataKey, String metadataValue);
```
chtyim: I would say just call it `addMetadata` is fine.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
anwar6953: Did you have a preference on `metadata` or `metadataEntries`?

chtyim: Metadata

anwar6953:
Addressed comments.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
```diff
@@ -0,0 +1,77 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.dataset.lib;
+
+import com.google.common.base.Objects;
```
chtyim: no guava usage for new api class. Sorry missed that earlier

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    PartitionMetadata that = (PartitionMetadata) o;
+
+    return Objects.equal(this.metadata, that.metadata);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(metadata);
```
chtyim: just return `metadata.hashCode()`.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionMetadata.java
```diff
@@ -0,0 +1,77 @@
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    PartitionMetadata that = (PartitionMetadata) o;
+
+    return Objects.equal(this.metadata, that.metadata);
```
chtyim: Since metadata cannot be null, simply `return metadata.equals(that.metadata);`

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
```diff
@@ -49,6 +49,24 @@
...
+
+  /**
+   * Adds a new metadata entry for a particular partition.
+   * Note that existing entries can not be updated.
+   */
+  void addMetadata(PartitionKey key, String metadataKey, String metadataValue);
+
+  /**
+   * Adds a set of new metadata entries for a particular partition
+   * Note that existing entries can not be updated.
```
chtyim: Does that mean it will throw exception from this method or get ignored silently?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
```diff
@@ -50,6 +50,26 @@
    */
   void addPartition(long time, String path);
 
+  // Why are some of these methods methodNameByTime, while others are just methodName?
```
chtyim: do you get the answer? We should keep this comment.

anwar6953: I did not get an answer, but my assumption is that the consistency of the methods was not a concern at the time they were named.
All of the methods act with `long time` as the first parameter, but the suffix is not consistent.
What do you mean keep this comment, though? You mean with an answer to the question?

chtyim: I mean we shouldn't keep this comment

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/BasicPartition.java
```diff
@@ -0,0 +1,69 @@
...
+import co.cask.cdap.api.dataset.lib.Partition;
+import co.cask.cdap.api.dataset.lib.PartitionKey;
+import com.google.common.base.Objects;
+import org.apache.twill.filesystem.Location;
+
+/**
+ * Simple Implementation of Partition.
+ */
+class BasicPartition implements Partition {
+  private PartitionedFileSetDataset partitionedFileSetDataset;
```
chtyim: final

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/BasicPartition.java
```diff
@@ -0,0 +1,69 @@
...
+
+  protected BasicPartition(PartitionedFileSetDataset partitionedFileSetDataset, String relativePath, PartitionKey key) {
+    this.partitionedFileSetDataset = partitionedFileSetDataset;
+    this.relativePath = relativePath;
+    this.key = key;
+  }
+
+  @Override
+  public Location getLocation() {
+    return partitionedFileSetDataset.files.getLocation(relativePath);
```
chtyim: Better not to reference field across classes. Use a getter instead.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/BasicPartitionDetail.java
```diff
@@ -0,0 +1,58 @@
...
+import co.cask.cdap.api.dataset.lib.PartitionDetail;
+import co.cask.cdap.api.dataset.lib.PartitionKey;
+import co.cask.cdap.api.dataset.lib.PartitionMetadata;
+import com.google.common.base.Objects;
+
+/**
+ * Implementation of Partition, with associated metadata.
+ */
+class BasicPartitionDetail extends BasicPartition implements PartitionDetail {
+  protected PartitionMetadata metadata;
```
chtyim: final

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -113,11 +116,16 @@ public Partitioning getPartitioning() {
...
   }
 
-  protected void addPartition(PartitionKey key, String path, boolean addToExplore) {
-    final byte[] rowKey = generateRowKey(key, partitioning);
+  @Override
+  public void addPartition(PartitionKey key, String path, Map<String, String> metadata) {
+    addPartition(key, path, true, metadata);
+  }
+
+  protected void addPartition(PartitionKey key, String path, boolean addToExplore, Map<String, String> metadata) {
```
chtyim: javadoc.

chtyim: Better call `explorable` instead of `addToExplore`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -136,14 +144,50 @@ protected void addPartition(PartitionKey key, String path, boolean addToExplore)
...
+  @Override
+  public void addMetadata(PartitionKey key, Map<String, String> metadata) {
+    final byte[] rowKey = generateRowKey(key, partitioning);
+    Row row = partitionsTable.get(rowKey);
+    if (row == null) {
+      throw new DataSetException(String.format("Dataset '%s' does not have a partition for key: %s", getName(), key));
+    }
+
+    // ensure that none of the entries already exist in the metadata
+    for (Map.Entry<String, String> metadataEntry : metadata.entrySet()) {
```
chtyim: You can merge then checking and add to put in one loop.

chtyim: Basically just move the check into the `addMetadataToPut` method.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -217,16 +262,23 @@ public void consume(PartitionKey key, String path) {
...
   protected void getPartitions(@Nullable PartitionFilter filter, PartitionConsumer consumer) {
-    final byte[] startKey = generateStartKey(filter);
-    final byte[] endKey = generateStopKey(filter);
+    // by default, parse the metadata from the rows
+    getPartitions(filter, consumer, true);
+  }
+
+  // if parseMetadata is false, null is passed as the PartitionMetadata to the PartitionConsumer,
+  // for efficiency reasons, since the metadata is not always needed
+  protected void getPartitions(@Nullable PartitionFilter filter, PartitionConsumer consumer, boolean parseMetadata) {
```
chtyim: It's more like `decode` than `parse`. Call it `decodeMetadata`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: the check isn't needed when first adding the partition, though, so the check on the Row shouldn't be moved into the `addMetadataToPut` method.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
```diff
@@ -49,6 +49,24 @@
...
 
   /**
+   * Add a partition for a given partition key, stored at a given path (relative to the file set's base path),
+   * with the given metadata.
+   */
+  void addPartition(PartitionKey key, String path, Map<String, String> metadata);
+
+  /**
+   * Adds a new metadata entry for a particular partition.
+   * Note that existing entries can not be updated; {@link IllegalArgumentException} will be thrown in that case.
```
anew: Better add a @throws to the Javadoc

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSet.java
```diff
@@ -49,6 +49,24 @@
...
+   */
+  void addMetadata(PartitionKey key, String metadataKey, String metadataValue);
+
+  /**
+   * Adds a set of new metadata entries for a particular partition
+   * Note that existing entries can not be updated; {@link IllegalArgumentException} will be thrown in that case.
```
anew: same here

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
```diff
@@ -51,6 +51,24 @@
...
   /**
+   * Add a partition for a given time, stored at a given path (relative to the file set's base path),
+   * with given metadata.
+   */
+  void addPartition(long time, String path, Map<String, String> metadata);
+
+  /**
+   * Adds a new metadata entry for a particular partition.
+   * Note that existing entries can not be updated.
+   */
```
anew: add a @throws IllegalArgumentException?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSet.java
```diff
@@ -51,6 +51,24 @@
...
+  void addMetadata(long time, String metadataKey, String metadataValue);
+
+  /**
+   * Adds a set of new metadata entries for a particular partition
+   * Note that existing entries can not be updated.
+   */
```
anew: same here

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/batch/MapReduceWithPartitionedTest.java
```diff
@@ -169,9 +169,9 @@ public void apply() {
       new TransactionExecutor.Subroutine() {
         @Override
         public void apply() {
-          Partition partition = tpfs.getPartitionByTime(time);
-          Assert.assertNotNull(partition);
-          String path = partition.getRelativePath();
```
anew: this change is actually not needed, because you never call getMetadata(), it should be fine to retrieve it as a Partition

anwar6953: Right. Changed back to use the interfaces.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -136,14 +153,50 @@ protected void addPartition(PartitionKey key, String path, boolean addToExplore)
...
+    Row row = partitionsTable.get(rowKey);
+    if (row == null) {
+      throw new DataSetException(String.format("Dataset '%s' does not have a partition for key: %s", getName(), key));
+    }
+
+    // ensure that none of the entries already exist in the metadata
+    for (Map.Entry<String, String> metadataEntry : metadata.entrySet()) {
+      String metadataKey = metadataEntry.getKey();
+      byte[] columnKey = columnKeyFromMetadataKey(metadataKey);
+      Preconditions.checkArgument(row.get(columnKey) == null, "Entry already exists for metadata key: {}", metadataKey);
```
anew: I am not sure whether this is Precondition...

anew: Actually, when you add a partition with a partition key that already exists, it throws a DatasetException. For consistency (since this is a similar situation), it should throw the same exception, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -186,29 +239,30 @@ public PartitionOutput getPartitionOutput(PartitionKey key) {
...
-  public Set<Partition> getPartitions(@Nullable PartitionFilter filter) {
-    final Set<Partition> partitions = Sets.newHashSet();
+  public Set<PartitionDetail> getPartitions(@Nullable PartitionFilter filter) {
+    final Set<PartitionDetail> partitionDetails = Sets.newHashSet();
     getPartitions(filter, new PartitionConsumer() {
       @Override
-      public void consume(PartitionKey key, String path) {
-        partitions.add(new BasicPartition(path, key));
+      public void consume(PartitionKey key, String path, @Nullable PartitionMetadata metadata) {
+        partitionDetails.add(new BasicPartitionDetail(PartitionedFileSetDataset.this, path, key, metadata));
```
anew: PartitionDetail assumes that Metadata is never null, right? But here metadata is @Nullable. I guess if it's null, you would want to use an empty meta data object?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -271,11 +351,15 @@ static String getOutputPath(Partitioning partitioning, PartitionKey key) {
     return builder.toString();
   }
 
+  FileSet getFiles() {
```
anew: Doesn't this already have a method getEmbeddedFileSet()

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -527,62 +612,24 @@ static PartitionKey parseRowKey(byte[] rowKey, Partitioning partitioning) {
...
-    }
-  }
-
-  /**
    * Simple Implementation of PartitionOutput.
    */
   protected class BasicPartitionOutput extends BasicPartition implements PartitionOutput {
+    private Map<String, String> metadata;
 
     protected BasicPartitionOutput(String relativePath, PartitionKey key) {
```
anew: I think this class can be static now

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/TimePartitionedFileSetDataset.java
```diff
@@ -101,11 +103,26 @@ public TimePartitionedFileSetDataset(DatasetContext datasetContext, String name,
 
   @Override
   public void addPartition(long time, String path) {
+    addPartition(time, path, ImmutableMap.<String, String>of());
```
anew: You can use Collections.emptyMap(), instead of creating a new immutable map every time it is called

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTest.java
```diff
@@ -116,6 +120,63 @@ public void testDecodeIncomplete() {
   }
 
   @Test
+  public void testPartitionMetadata() throws Exception {
```
anew: I think these tests should all use transactions.

anew:
@anwar6953 please see my remaining comments, nothing major, but please address.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: I will change it to that, though unsure what all situations that DatasetException is intended for.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: right; I missed that.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTest.java
anwar6953: Were you thinking to wrap the different actions (read, scan, get, etc.) into separate transactions, or just one transaction for the test case?

anwar6953: ok, well, for now, updated by wrapping the actions of a test case within just 1 transaction, and pushed.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -136,14 +153,52 @@ protected void addPartition(PartitionKey key, String path, boolean addToExplore)
...
+  @Override
+  public void addMetadata(PartitionKey key, String metadataKey, String metadataValue) {
+    addMetadata(key, ImmutableMap.of(metadataKey, metadataValue));
+  }
+
+  @Override
+  public void addMetadata(PartitionKey key, Map<String, String> metadata) {
+    final byte[] rowKey = generateRowKey(key, partitioning);
+    Row row = partitionsTable.get(rowKey);
+    if (row == null) {
```
anew: Note that get() return an empty Row if the row does not exist.

anew:
I have one one (new) comment, sorry, that slipped through previously. All else looks good.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: Right, I either missed that, or just followed the pattern in getPartition method, which also checked for `row == null`.
I changed both of these places and added a relevant test case.

anew:
LGTM

PR #2951, opened 2015-06-12
Documentation Build Script changes

awholegunch:
Fixes the build script used to generate the SDK for testing used by docs. Adds a guard so if BUILD.rst is changed, it is flagged when the script runs something that has been copied from BUILD.rst.

Adds MD5 hash for BUILD.rst.

Fixes minor formatting errors that were causing problems with source code coloring in the text editor "TextWrangler".


duddun:
LGTM

duddun:
LGTM again

awholegunch:
Building...http://builds.cask.co/browse/CDAP-DDBD319-1/log

PR #2952, opened 2015-06-12
increase polling frequency

elfenheart:
Increase polling frequency for metrics so that it will show up in the status page.
Spark program finished very quickly so the metric changed rapidly.


ajainarayanan:
:+1: LGTM

PR #2953, opened 2015-06-12
Make DiscoveryServiceClient non-static

bdmogal:
1. Made ``DiscoveryServiceClient`` non-static in ``SerializableServiceDiscoverer``
2. Added an ``InMemorySerializableServiceDiscoverer`` for use in in-memory mode
3. Changed the constructor in ``AbstractServiceDiscoverer`` to accept just an ``Id.Application`` instead of the entire ``Program`` object which is not needed.

Only works in in-memory mode. Testing on a cluster with more changes. Will create a separate PR for distributed once that works.

Build: https://builds.cask.co/browse/CDAP-DUT2032/


bdmogal:
We may not need this, with the CDAP Spark code refactor.

chtyim:
We don't need this change.

PR #2954, opened 2015-06-12
Feature/ui apps home overview controlleras

ajainarayanan:
Modifies Home, Overview and Apps features to use controller-as syntax.


elfenheart:
:+1: LGTM

PR #2955, opened 2015-06-13
refactor services program type to use controller as syntax

elfenheart:


cdap-ui/app/features/services/templates/tabs/runs.html
```diff
@@ -6,11 +6,11 @@
       <button type="button"
               class="btn btn-default dropdown-toggle"
               dropdown-toggle ng-disabled="runs.length === 0">
```
ajainarayanan: ```RunsController.runs.length```?

ajainarayanan:
:+1: LGTM

PR #2958, opened 2015-06-15
CDAP-2563 refactoring RemoteDatasetFramework so that it no longer

albertshau:
unpacks dataset jars by default each time it instantiates a
dataset or dataset type. This is because in most contexts, the
classloader passed in will be a ProgramClassLoader that already
has access to the dataset.

For calls that are made in the CDAP system context, added a
separate method to DatasetFramework that takes a
DatasetClassLoaderProvider that can unpack
dataset jars if it needs to, but which is closeable to allow
cleanup of those unpacked jars. This method is used by the dataset
admin op handler and by explore to instantiate datasets.

Also refactoring the way datasets are instantiated in explore,
since there is no longer any need to cache dataset classloaders,
but need to make sure to clean up unpacked directories. To do this,
adding a SystemDatasetInstantiator class that is intended to be
used in CDAP system contexts, which will delete relevant
local directories when it is closed. This replaces most direct
interactions with DatasetFramework.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2041

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/DatasetClassLoaderProvider.java
```diff
@@ -18,13 +18,28 @@
...
+   */
+  ClassLoader get(DatasetModuleMeta moduleMeta) throws IOException;
 
+  /**
+   * Get the parent classloader for module classloaders
+   *
+   * @return the parent classloader for module classloaders
+   * @throws IOException
+   */
+  ClassLoader getParent();
```
chtyim: This seems a bit awkward to have a method that return a ClassLoader that claims to be "parent". Usually parent classloader is get from a ClassLoader instance. How is this method being used?

chtyim: Also, can it return null?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/DirectoryClassLoaderProvider.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @Override
+  public ClassLoader getParent() {
+    return parentClassLoader;
+  }
+
+  @Override
+  public void close() throws IOException {
+    // Cleanup the ClassLoader cache and the directory used for expanded dataset jars
+    classLoaders.invalidateAll();
```
chtyim: Need to have a removal listener to close the ClassLoaders as well (if the ClassLoader is closable, which usually is). See https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java#L82

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/DatasetClassLoaderProvider.java
albertshau: hm yeah this is a bad name. It is really the classloader to use if a dataset doesn't use any dataset modules.  Which shouldn't be possible, but wasn't clear to me when I was first looking at the code.  Will remove this method.

albertshau:
@chtyim addressed comments, can you take another look?

cdap-explore/src/main/java/co/cask/cdap/hive/datasets/DatasetSerDe.java
```diff
@@ -72,26 +83,67 @@ public void initialize(Configuration entries, Properties properties) throws SerD
...
+
+    // Hive may call initialize a bunch of times... so remember the schema so we don't instantiate the dataset
+    // a bunch of times.
+    if (schema == null) {
+      Id.DatasetInstance datasetId = Id.DatasetInstance.from(namespace, datasetName);
+      getDatasetSchema(conf, datasetId);
     }
+
+    this.deserializer = new ObjectDeserializer(properties, schema);
+    ArrayList<String> columnNames = Lists.newArrayList(StringUtils.split(properties.getProperty("columns"), ","));
```
chtyim: `List`.

cdap-data-fabric/src/main/java/co/cask/cdap/data/dataset/DatasetInstantiator.java
```diff
@@ -39,24 +39,20 @@
  */
 public class DatasetInstantiator implements DatasetContext {
 
-  private final DatasetFramework datasetFramework;
-  // the class loader to use for data set classes
-  private final ClassLoader classLoader;
-  private final Set<TransactionAware> txAware = Sets.newIdentityHashSet();
-  // in this collection we have only datasets initialized with getDataset() which is OK for now...
-  private final Map<TransactionAware, String> txAwareToMetricNames = Maps.newIdentityHashMap();
```
anew: is this not needed any longer?

cdap-data-fabric/src/main/java/co/cask/cdap/data/dataset/SystemDatasetInstantiator.java
```diff
@@ -0,0 +1,90 @@
...
+    throws DatasetInstantiationException {
+
+    T dataset;
+    try {
+      if (!datasetFramework.hasInstance(datasetId)) {
+        throw new DatasetInstantiationException("Trying to access dataset that does not exist: " + datasetId);
+      }
+
+      dataset = datasetFramework.getDataset(datasetId, arguments, parentClassLoader, classLoaderProvider, owners);
+      if (dataset == null) {
```
anew: can this return null? when would that happen? Shouldn't it rather throw an exception with the reason for the failure?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -317,7 +326,7 @@ private static Location createDeploymentJar(Class<?> clz, Location destination)
 
       Class<?> moduleClass;
 
-      // try program class loader then cdap class loader
+      // try program class loader then cdap class loader0
```
anew: extra 0?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/executor/DatasetAdminOpHTTPHandler.java
```diff
@@ -93,19 +102,22 @@ public void create(HttpRequest request, HttpResponder responder,
...
     LOG.info("Creating dataset instance {}.{}, type meta: {}, props: {}", namespaceId, name, typeMeta, props);
-    DatasetType type = dsFramework.getDatasetType(typeMeta, null);
-
-    if (type == null) {
-      String msg = String.format("Cannot instantiate dataset type using provided type meta: %s", typeMeta);
-      LOG.error(msg);
-      responder.sendString(HttpResponseStatus.BAD_REQUEST, msg);
-      return;
+    try (DatasetClassLoaderProvider classLoaderProvider =
+           new DirectoryClassLoaderProvider(cConf, locationFactory)) {
```
anew: this is nice.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -295,10 +304,10 @@ private static Location createDeploymentJar(Class<?> clz, Location destination)
 
   // can be used directly if DatasetTypeMeta is known, like in create dataset by dataset ops executor service
   public <T extends DatasetType> T getDatasetType(DatasetTypeMeta implementationInfo,
```
anew: this should have a Javadoc. 
Also, why is this not a method in the interface? Is it only useful for remote dataset frameworks?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/ConstantClassLoaderProvider.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  public ClassLoader get(DatasetModuleMeta moduleMeta, ClassLoader parentClassLoader) throws IOException {
+    return classLoader;
+  }
+
+  @Override
+  public void close() throws IOException {
+    // no-op
```
anew: if the class loader passed in to the constructor is closeable, should it be closed here?

cdap-explore/src/main/java/co/cask/cdap/explore/executor/ExploreExecutorHttpHandler.java
```diff
@@ -216,15 +219,16 @@ public void addPartition(HttpRequest request, HttpResponder responder,
...
-      responder.sendJson(HttpResponseStatus.OK, json);
+    try (SystemDatasetInstantiator datasetInstantiator = datasetInstantiatorFactory.create()) {
+      dataset = datasetInstantiator.getDataset(datasetInstanceId);
+
+      if (dataset == null) {
+        responder.sendString(HttpResponseStatus.NOT_FOUND, "Cannot load dataset " + datasetInstanceId);
+        return;
+      }
+    } catch (IOException e) {
+      LOG.error("Exception instantiating dataset {}.", datasetInstanceId, e);
```
anew: this changes the behavior. The old code returned OK.

cdap-explore/src/main/java/co/cask/cdap/explore/executor/ExploreExecutorHttpHandler.java
```diff
@@ -274,15 +278,17 @@ public void dropPartition(HttpRequest request, HttpResponder responder,
...
+    try (SystemDatasetInstantiator datasetInstantiator = datasetInstantiatorFactory.create()) {
+
+      dataset = datasetInstantiator.getDataset(datasetInstanceId);
+      if (dataset == null) {
+        responder.sendString(HttpResponseStatus.NOT_FOUND, "Cannot load dataset " + datasetInstanceId);
+        return;
+      }
+    } catch (IOException e) {
+      LOG.error("Exception instantiating dataset {}.", datasetInstanceId, e);
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR,
```
anew: same here

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -1049,20 +1050,24 @@ private void upgradeFilesetTable(TableInfo tableInfo) throws Exception {
...
-        QueryHandle handle = exploreTableManager.addPartitions(datasetID, partitionDetails);
-        QueryStatus status = waitForCompletion(handle);
-        // if add partitions failed, stop
-        if (status.getStatus() != QueryStatus.OpStatus.FINISHED) {
-          throw new ExploreException("Failed to add all partitions to dataset " + datasetID);
+    try (SystemDatasetInstantiator datasetInstantiator = datasetInstantiatorFactory.create()) {
+      Dataset dataset = datasetInstantiator.getDataset(datasetID);
+
+      // if this is a time partitioned file set, we need to add all partitions
+      if (dataset instanceof TimePartitionedFileSet) {
```
anew: what about PartitionedFileSets?

cdap-explore/src/main/java/co/cask/cdap/hive/datasets/DatasetAccessor.java
```diff
@@ -16,265 +16,68 @@
...
- * Helps in instantiating a dataset.
+ * Instantiates the dataset used during runtime of a Hive query. This means it's used in mappers and reducers,
+ * and must use the Hadoop configuration to look up what dataset to instantiate. Should not be closed until the
+ * dataset has been closed. Assumes the dataset name and namespace are settings in the configuration.
+ * It may seem like this would not work if multiple datasets are used in a single query, but that is not the case.
+ * It is not obvious, but dataset name and namespace are added as job properties in DatasetStorageHandler. This tells
+ * Hive to add those properties to the Configuration object before passing it in to the methods of an InputFormat
+ * or OutputFormat. So even if multiple datasets are used in the same query (a join query for example), dataset name
+ * will not get clobbered.
  */
```
anew: Thanks for this comment, very helpful.

anew:
This is a huge change. I am not 100% confident that this is 100% correct, but it looks pretty good. I had a few comments, though.

cdap-data-fabric/src/main/java/co/cask/cdap/data/dataset/DatasetInstantiator.java
albertshau: this map was being populated, but never used in the old implementation so i just removed it.

cdap-data-fabric/src/main/java/co/cask/cdap/data/dataset/SystemDatasetInstantiator.java
albertshau: this is a move of code that was in DatasetInstantiator. DatasetFramework returns null if the dataset doesn't exist... but that check is made above.  Should be safe to remove the hasInstance check and replace this exception with that one.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
albertshau: I am not certain, but I think its because DatasetTypeMeta is not available in all contexts.  I'll add a javadoc though.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
albertshau: this was upgrade code for the namespace release. I think PartitionedFileSets came after so there was no handling for that.  I'll add some comments

cdap-explore/src/main/java/co/cask/cdap/explore/executor/ExploreExecutorHttpHandler.java
albertshau: I'll add back the handling for ClassNotFound

cdap-explore/src/main/java/co/cask/cdap/hive/datasets/DatasetSerDe.java
albertshau: this has to be an ArrayList because Hive's ```StructTypeInfo.setAllStructFieldNames()``` method takes ArrayList

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/ConstantClassLoaderProvider.java
albertshau: the reason I didn't close it was because the classloader is not owned by this class but is passed in. Currently, it's a ProgramClassLoader that is passed in, and I didn't want to close it here since the program may still need to continue and load classes.  I'll at least add a comment

cdap-data-fabric/src/main/java/co/cask/cdap/data/dataset/DatasetInstantiator.java
anew: makes sense

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/ConstantClassLoaderProvider.java
anew: ok, it makes sense, and so does a comment. thanks

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
anew: ok, I see.  I missed that

anew:
Nice work. LGTM.

PR #2960, opened 2015-06-15
CDAP-2585: Use Id in cdap-client

jwang47:
https://issues.cask.co/browse/CDAP-2585
http://builds.cask.co/browse/CDAP-DUT2047


anwar6953:
Was there a reason though that a lot of the `create` commands were still in the old pattern?

jwang47:
Not really, just thought it might make more sense.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -127,25 +127,28 @@ public TableRenderer getTableRenderer() {
   }
 
   public Id.Namespace getCurrentNamespace() {
-    return clientConfig.getNamespace();
+    return connectionConfig != null ? connectionConfig.getNamespace() : null;
```
anwar6953: What does it mean for a CLIConfig to operate without a ConnectionConfig? Does it just mean that it will use the ClientConfig's internal ConnectionConfig?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
```diff
@@ -127,25 +127,28 @@ public TableRenderer getTableRenderer() {
   }
 
   public Id.Namespace getCurrentNamespace() {
```
anwar6953: add `@Nullable`?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
anwar6953: And it seems like a lot of the usages in the CLI use this method without expecting it to be able to return null.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -118,15 +108,18 @@ protected String getAccessToken() {
   }
 
   private void assertIsClear() throws Exception {
+    // TODO: support multiple namespaces
```
anwar6953: support multiple namespaces? Isn't it sufficient to assert that the default namespace is the only one existing?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -211,42 +194,58 @@ private boolean isUserDataset(DatasetSpecificationSummary specification) {
     return !dsNamespace.contains(specification.getName(), Constants.SYSTEM_NAMESPACE);
   }
 
-  private void assertNoUserDatasets() throws Exception {
+  private void truncateAllStreams(Id.Namespace namespace) throws IOException, UnauthorizedException {
+    // TODO: check no streams once streams can be deleted instead of truncating all streams
```
anwar6953: Streams can be deleted now.
https://github.com/caskdata/cdap/blob/feature/pfs-metadata/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/StreamHandler.java#L263-263

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConfig.java
jwang47: Not having a ConnectionConfig means the CLI isn't connected to any cluster (you'll have the prompt `cdap (DISCONNECTED)>`). I'll change it to not allow returning null though.

anwar6953:
No further comments from me at this time, though I think it's a good idea if someone else also takes a look at this PR, because of its size.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConnectionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * {@link ConnectionConfig} that stores a username when available. Helps keep connectionConfig thread safe.
+ */
+public class CLIConnectionConfig extends ConnectionConfig {
+
+  private final Id.Namespace namespace;
+  private final String username;
+
+  public CLIConnectionConfig(ConnectionConfig connectionConfig, Id.Namespace namespace, @Nullable String username) {
+    super(connectionConfig.getHostname(), connectionConfig.getPort(), connectionConfig.isSSLEnabled());
```
bdmogal: should we just add a ``ConnectionConfig`` constructor that accepts a ``ConnectionConfig``? would make this a little cleaner, perhaps?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConnectionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+    return namespace;
+  }
+
+  public String getUsername() {
+    return username;
+  }
+
+  @Override
+  public URI getURI() {
+    return URI.create(String.format("%s://%s%s:%d/%s", super.isSSLEnabled() ? "https" : "http",
```
bdmogal: if you have all these components, then it may be better to use the [constructor](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#URI(java.lang.String,%20java.lang.String,%20java.lang.String,%20int,%20java.lang.String,%20java.lang.String,%20java.lang.String)) instead of formatting a String. However, you may then have to handle exceptions.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -206,15 +207,14 @@ private String limit(String string, int maxLength) {
...
+    cli.getReader().setPrompt(getPrompt(config));
   }
 
-  public String getPrompt(ClientConfig clientConfig) {
+  public String getPrompt(CLIConnectionConfig config) {
     try {
-      ConnectionConfig connectionConfig = clientConfig.getConnectionConfig();
-      URI baseURI = connectionConfig.getURI();
-      URI uri = baseURI.resolve("/" + connectionConfig.getNamespace());
+      URI baseURI = config.getURI();
```
bdmogal: nit: remove this variable. Just use ``URI uri = config.getURI().resolve...``?

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConnectionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+import co.cask.cdap.client.config.ConnectionConfig;
+import co.cask.cdap.proto.Id;
+
+import java.net.URI;
+import javax.annotation.Nullable;
+
+/**
+ * {@link ConnectionConfig} that stores a username when available. Helps keep connectionConfig thread safe.
+ */
+public class CLIConnectionConfig extends ConnectionConfig {
```
bdmogal: Should this class extend ``NamespacedConnectionConfig`` instead?

cdap-cli/src/main/java/co/cask/cdap/cli/command/DeleteDatasetInstanceCommand.java
```diff
@@ -43,10 +44,11 @@ public DeleteDatasetInstanceCommand(DatasetClient datasetClient, CLIConfig cliCo
...
   @Override
   public void perform(Arguments arguments, PrintStream output) throws Exception {
-    String datasetName = arguments.get(ArgumentName.DATASET.toString());
+    Id.DatasetInstance dataset = Id.DatasetInstance.from(cliConfig.getCurrentNamespace(),
+                                                         arguments.get(ArgumentName.DATASET.toString()));
 
-    datasetClient.delete(datasetName);
-    output.printf("Successfully deleted dataset named '%s'\n", datasetName);
+    datasetClient.delete(dataset);
+    output.printf("Successfully deleted %s\n", dataset);
```
bdmogal: we don't have to worry that the output format of some CLI commands now changes, right? Do we worry about compatibility there?

cdap-cli/src/main/java/co/cask/cdap/cli/commandset/ScheduleCommands.java
```diff
@@ -121,9 +124,11 @@ public void perform(Arguments arguments, PrintStream printStream) throws Excepti
       }
 
       String appId = programIdParts[0];
-      String scheduleId = programIdParts[1];
-      scheduleClient.suspend(appId, scheduleId);
-      printStream.printf("Successfully suspended schedule '%s' in app '%s'", scheduleId, appId);
+      String scheduleName = programIdParts[1];
```
bdmogal: no need for this variable?

cdap-cli/src/main/java/co/cask/cdap/cli/commandset/ScheduleCommands.java
```diff
@@ -158,9 +163,11 @@ public void perform(Arguments arguments, PrintStream printStream) throws Excepti
       }
 
       String appId = programIdParts[0];
-      String scheduleId = programIdParts[1];
-      scheduleClient.resume(appId, scheduleId);
-      printStream.printf("Successfully resumed schedule '%s' in app '%s'", scheduleId, appId);
+      String scheduleName = programIdParts[1];
```
bdmogal: same here

cdap-cli/src/main/java/co/cask/cdap/cli/commandset/ScheduleCommands.java
```diff
@@ -195,8 +202,10 @@ public void perform(Arguments arguments, PrintStream output) throws Exception {
       }
 
       final String appId = programIdParts[0];
-      String scheduleId = programIdParts[1];
-      List<ScheduleSpecification> list = scheduleClient.list(appId, scheduleId);
+      String workflowName = programIdParts[1];
```
bdmogal: same here

cdap-client-tests/src/test/java/co/cask/cdap/client/AdapterClientTestRun.java
```diff
@@ -56,58 +56,59 @@
...
     super.setUp();
-    clientConfig.setNamespace(Constants.DEFAULT_NAMESPACE_ID);
     adapterClient = new AdapterClient(clientConfig);
     appTemplateClient = new ApplicationTemplateClient(clientConfig);
   }
 
   @Test
   public void testAdapters() throws Exception {
-    List<AdapterDetail> initialList = adapterClient.list();
+    Id.Namespace namespace = Id.Namespace.DEFAULT;
```
bdmogal: can we use ``Id.Namespace.DEFAULT``, instead of this extra variable?

Also, could you please create a jira to unify ``Id.Namespace.DEFAULT`` and ``Constants.DEFAULT_NAMESPACE_ID``? Shouldn't have the latter anymore.

cdap-client-tests/src/test/java/co/cask/cdap/client/ApplicationClientTestRun.java
```diff
@@ -62,14 +63,16 @@ public void cleanup() throws Throwable {
     // Delete FakeApp's dataset and module so that DatasetClientTestRun works when running both inside a test suite
     // This is due to DatasetClientTestRun assuming that it is using a blank CDAP instance
 
+    Id.Namespace namespace = Id.Namespace.DEFAULT;
```
bdmogal: can we remove this?

cdap-client-tests/src/test/java/co/cask/cdap/client/ApplicationClientTestRun.java
```diff
@@ -77,64 +80,71 @@ public void cleanup() throws Throwable {
...
+      appClient.delete(app);
+      appClient.waitForDeleted(app, 30, TimeUnit.SECONDS);
+      Assert.assertEquals(0, appClient.list(namespace).size());
     }
   }
 
   @Test
   public void testDeleteAll() throws Exception {
-    Assert.assertEquals(0, appClient.list().size());
+    Id.Namespace namespace = Id.Namespace.DEFAULT;
```
bdmogal: same here. Although this, like the previous comments about similar things are nits. I just think assigning a variable to a constant is unnecessary. Although I can see that doing it this way makes the variable name shorter.

cdap-client-tests/src/test/java/co/cask/cdap/client/DatasetClientTestRun.java
```diff
@@ -82,113 +81,121 @@ public void tearDown() throws Exception {
 
   @Test
   public void testAll() throws Exception {
-    int numBaseModules = moduleClient.list().size();
-    int numBaseTypes = typeClient.list().size();
+    Id.Namespace namespace = TEST_NAMESPACE;
```
bdmogal: same here

cdap-client-tests/src/test/java/co/cask/cdap/client/MetricsClientTestRun.java
```diff
@@ -58,11 +58,16 @@ public void setUp() throws Throwable {
 
   @Test
   public void testAll() throws Exception {
-    appClient.deploy(createAppJarFile(FakeApp.class));
+    Id.Namespace namespace = Id.Namespace.DEFAULT;
```
bdmogal: same here

cdap-client-tests/src/test/java/co/cask/cdap/client/PreferencesClientTestRun.java
```diff
@@ -50,8 +50,10 @@
  */
 @Category(XSlowTests.class)
 public class PreferencesClientTestRun extends ClientTestBase {
+
   private static final Gson GSON = new Gson();
-  private static final Id.Application FAKE_APP_ID = Id.Application.from(Constants.DEFAULT_NAMESPACE_ID, FakeApp.NAME);
+  private static final Id.Namespace NAMESPACE = Id.Namespace.DEFAULT;
```
bdmogal: should we use ``Id.Namespace.DEFAULT`` directly?

cdap-client-tests/src/test/java/co/cask/cdap/client/AdapterClientTestRun.java
jwang47: Created JIRA https://issues.cask.co/browse/CDAP-2917

bdmogal:
@jwang47 code-wise the PR is good, just a few comments. A couple of concerns though:
1. What is the motivation for doing this? It makes some part of the code cleaner (users of ``cdap-client`` can reuse Client objects across namespaces). However, the cdap-clients themselves have namespaces in every method now, making them a little less cleaner than before?
2. What compatibility guarantees do we have for cdap-clients? This is an incompatible change in them, right? Is that ok?

jwang47:
@bdmogal 
1. After using cdap-client in writing some integration tests, it was clear that using Id.Namespace + Strings would be much more verbose and error-prone than simply using Id.*. See https://github.com/caskdata/cdap-integration-tests/blob/develop/integration-test-remote/src/test/java/co/cask/cdap/apps/purchase/PurchaseAudiTest.java#L73-L83 for an example.
2. No compatibility guarantees. Should've marked the methods/classes as @Beta though, but it's too late now.

cdap-cli/src/main/java/co/cask/cdap/cli/CLIConnectionConfig.java
jwang47: This class replaces `NamespacedConnectionConfig`.

cdap-cli/src/main/java/co/cask/cdap/cli/command/DeleteDatasetInstanceCommand.java
jwang47: Nope.

cdap-cli/src/main/java/co/cask/cdap/cli/commandset/ScheduleCommands.java
jwang47: Not necessary, but it makes it understandable what `programIdParts[1]` is without having to look up javadocs for `ScheduleClient`.

cdap-client-tests/src/test/java/co/cask/cdap/client/ApplicationClientTestRun.java
jwang47: The idea was to make it easy to change the namespace that we use, since default namespace behaves differently from non-default namespaces.

bdmogal:
I think 1. is ok. But there is a problem with 2. I think the question is not if we provide compatibility guarantees. Its if we recommend users to use ``cdap-clients`` to interact with CDAP. Just trying to find out if these are *internal* clients (generally done by restricting scope or having *internal* in the package name) only meant to be used by ``cdap-cli`` or if users are free to use them in integration tests for their apps, for example. If the latter, then I think we have a problem here, because users who have written tests for their apps using CDAP 3.0 will have to modify/recompile their tests to work with 3.1. Thoughts?

bdmogal:
For the tests that I mentioned in the previous comment, I think we would still be ok if the integration test framework abstracts ``cdap-clients``, so users never have to use ``cdap-clients``, they only rely on the integration test framework.

jwang47:
True, users will need to recompile their tests to work with 3.1. Alternatively, we can have the old methods deprecated and add the new methods in 3.1 and remove the old methods in 3.2.

The integration test framework wraps cdap-client for all the "unit test" functionality available via `TestManager`, but also exposes cdap-client since `TestManager` doesn't have all the functionality that cdap-client has.

sreevatsanraman:
@jwang47 - we cannot remove the APIs we have introduced. We will have customers who have written tests based on these API. Right approach is to mark the APIs we are going to remove as deprecated and then introduce the new APIs. We should remove the deprecated APIs in the next rev. 

Can we make sure that this happens?

jwang47:
Added back the deprecated methods.

anwar6953:
I think it would be a good idea to create a JIRA with a fixVersion for whenever we plan to remove the deprecated methods (3.2? 3.3? 4.0?), linking to this PR, to revert the changes of the last ten commits (so that we don't lose track of it over the months).

jwang47:
Created JIRA here: https://issues.cask.co/browse/CDAP-2951

anwar6953:
LGTM

sreevatsanraman:
Ali gave LGTM and the build passed http://builds.cask.co/browse/CDAP-DUT2047-14 
Merging.

PR #2961, opened 2015-06-16
CDAP-2760: collect metrics processed stats at Metrics Processor

abaranau:
NOTE: it was instrumented incorrectly before: the metric was collected on client side.


abaranau:
http://builds.cask.co/browse/CDAP-DUT2044-1

gsps1:
do we want to keep the client level meta metrics as well ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
```diff
@@ -41,6 +43,8 @@
 public final class MetricsMessageCallback implements KafkaConsumer.MessageCallback {
 
   private static final Logger LOG = LoggerFactory.getLogger(MetricsMessageCallback.class);
+  private static final ImmutableMap<String, String> METRICS_PROCESSOR_CONTEXT =
+    ImmutableMap.of(Constants.Metrics.Tag.NAMESPACE, "system", Constants.Metrics.Tag.COMPONENT, "metrics.processor");
```
gsps1: better to Constants.SYSTEM_NAMESPACE and another constant for metrics-processor. would be easier to track.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
```diff
@@ -59,18 +63,17 @@ public MetricsMessageCallback(DatumReader<MetricValues> recordReader,
...
-          LOG.info("Failed to decode message to MetricValue. Skipped. {}", e.getMessage());
-          return null;
-        }
+    List<MetricValues> records = Lists.newArrayList();
+
+    while (messages.hasNext()) {
+      FetchedMessage input = messages.next();
+      try {
+        MetricValues metricValues = recordReader.read(new BinaryDecoder(is.reset(input.getPayload())), recordSchema);
+        records.add(metricValues);
```
gsps1: can metricValues be null? if so , we might want to skip that.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
```diff
@@ -93,6 +97,16 @@ public MetricValues apply(FetchedMessage input) {
     }
   }
 
+  private void addProcessingStats(List<MetricValues> records) {
+    int count = records.size();
+    long now = System.currentTimeMillis();
+    long delay = now - records.get(records.size() - 1).getTimestamp();
```
gsps1: check if records is not empty. the list could be empty if we can't decode the messages we received right ?

gsps1: also getTimeStamp() is in seconds, while now is in milliseconds. you might want to convert appropriately.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
```diff
@@ -93,6 +97,16 @@ public MetricValues apply(FetchedMessage input) {
...
+    records.add(
+      new MetricValues(METRICS_PROCESSOR_CONTEXT, now / 1000,
+                       ImmutableList.of(new MetricValue("metrics.process.count", MetricType.COUNTER, count),
+                                        new MetricValue("metrics.process.delay", MetricType.GAUGE, delay))));
```
gsps1: add unit?, delay.seconds or delay.ms ?

gsps1:
minor comments, lgtm otherwise

abaranau:
> do we want to keep the client level meta metrics as well ?

we don't aggregate them yet. We might add separately, correctly.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
abaranau: > check if records is not empty. the list could be empty if we can't decode the messages we received right ?

it is only called if not empty

abaranau: > also getTimeStamp() is in seconds, while now is in milliseconds. you might want to convert appropriately.

it is converted later

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
abaranau: done

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
abaranau: no, it cant..

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
abaranau: done

abaranau: done anyways

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
abaranau: done

PR #2962, opened 2015-06-16
Flattens node module dependencies with shrinkwrap

ajainarayanan:
Right now there is no way to freeze on dependencies for node modules in node proxy. shrinkwrap allows us to lock onto the specific versions. So any subsequent npm install would install only those specific versions guaranteeing a stable version of node server based on current dependencies.

NOTE:
  I have manually gone through all 1st level dependencies' package.json files to check what is the min node version that it needs to run stable. Based on that I have modified the min version of node engine to be 0.10.0. This ensures that if the user is using a node version that is less than that then npm install would fail, indicating that they need a newer version of node.

### DIFF
--- a/cdap-ui/npm-shrinkwrap.json
+++ b/cdap-ui/npm-shrinkwrap.json
@@ -1,0 +1,862 @@
+{
+  "name": "cdap-ui",
+  "version": "3.0.1",
+  "dependencies": {
+    "body-parser": {
+      "version": "1.13.0",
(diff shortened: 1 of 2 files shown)


ajainarayanan:
This change still needs to be integrated with beamer (when I do beamer software install cdap-ui) it doesn't beam npm-shrinkwrap.json file. Once that is done it should be available for cluster. Right now I have this cluster (http://104.154.39.178:9999/ns/default) that has these minor node changes.

PR #2963, opened 2015-06-16
CDAP-2761: separate MetricsCollector from the MetricsContext to allow…

abaranau:
… exposing in MetricsCollector for custom datasets.

We want to unify the metrics collection API. We also want to separate it from MetricsContext - so that we can avoid exposing the latter one to the custom user code.

In this PR:
* MetricsCollector is split into two: MetricsCollector and MetricsContext that extends from it. 

Nothing else has been changed: all changes across files are automatic refactoring.

### DIFF
--- a/cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricsCollectionService.java
+++ b/cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricsCollectionService.java
@@ -28,7 +28,7 @@ public interface MetricsCollectionService extends Service {
    * Returns the metric collector for the given context.
    *
    * @param context The tags that define the metrics context.
-   * @return A {@link MetricsCollector} for emitting metrics.
+   * @return A {@link MetricsContext} for emitting metrics.
    */
-  MetricsCollector getCollector(Map<String, String> context);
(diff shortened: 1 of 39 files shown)


cdap-watchdog-api/src/main/java/co/cask/cdap/api/metrics/MetricsCollectionService.java
```diff
@@ -28,7 +28,7 @@
    * Returns the metric collector for the given context.
    *
    * @param context The tags that define the metrics context.
-   * @return A {@link MetricsCollector} for emitting metrics.
+   * @return A {@link MetricsContext} for emitting metrics.
    */
-  MetricsCollector getCollector(Map<String, String> context);
+  MetricsContext getCollector(Map<String, String> context);
```
chtyim: Maybe rename the method as well? Something like `getContext`?

abaranau: fixed

abaranau:
was GREEN. Starting new

abaranau:
http://builds.cask.co/browse/CDAP-DUT2046-3

chtyim:
LGTM

PR #2964, opened 2015-06-16
CDAP-2769 Use co.cask.cdap.common.app.RunIds for InMemoryServiceProgramRunner

henry-cask:
Use co.cask.cdap.common.app.RunIds instead of org.apache.twill.internal.RunIds for InMemoryServiceProgramRunner.

This issue is found when the RunId cleaner try to filter the RUNNING run records with check to
the RunId hash code which is from UUID for CDAP common and String for Twill.


gokulavasan:
LGTM :+1:

henry-cask:
Thx, merging ...

PR #2965, opened 2015-06-16
Fix/ui authenticated nodeproxy

ajainarayanan:
- Fixes Authentication in UI
- Handles re-directs when auth token is invalid
- Handles /backendstatus and /version end point calls to be made only after authentication (if security is enabled).
- Streams response back from backend for /v3/namespace end point instead of sending out a 404.

Security Enabled cluster having this code: http://securecluster1762-1000.dev.continuuity.net:9999
Security Disabled cluster having the same code: http://nodeproxy1736-1000.dev.continuuity.net:9999

### DIFF
--- a/cdap-ui/app/features/overview/controllers/overview-ctrl.js
+++ b/cdap-ui/app/features/overview/controllers/overview-ctrl.js
@@ -3,7 +3,7 @@
  */
 
 angular.module(PKG.name+'.feature.overview').controller('OverviewCtrl',
-function ($scope, $state, myLocalStorage, MY_CONFIG, Widget, MyMetricsQueryHelper, myHelpers, MyChartHelpers, MyDataSource) {
+function ($scope, $state, myLocalStorage, MY_CONFIG, Widget, MyMetricsQueryHelper, myHelpers, MyChartHelpers, MyDataSource, EventPipe) {
 
   var dataSrc = new MyDataSource($scope);
   if(!$state.params.namespace) {
@@ -33,7 +33,8 @@ function ($scope, $state, myLocalStorage, MY_CONFIG, Widget, MyMetricsQueryHelpe
   dataSrc.poll({
     _cdapPath: '/system/services/status',
     interval: 10000
-  }, function(res) {
+  },
+  function success(res) {
     var serviceStatuses = Object.keys(res).map(function(value) {
       return res[value];
     });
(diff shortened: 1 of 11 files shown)


cdap-ui/app/main.js
```diff
@@ -141,14 +141,14 @@ angular
...
         'request': function(config) {
           if ($rootScope.currentUser) {
-            angular.extend({
+            angular.extend(config, {
               user: $rootScope.currentUser || null,
               headers: {
                 // Accessing stuff from $rootScope is bad. This is done as to resolve circular dependency.
                 // $http <- myAuthPromise <- myAuth <- $http <- $templateFactory <- $view <- $state
-                authorization: ($rootScope.currentUser.token ? 'Bearer ' + $rootScope.currentUser.token: null)
+                Authorization: ($rootScope.currentUser.token ? 'Bearer ' + $rootScope.currentUser.token: null)
```
anew: really?

cdap-ui/app/services/data/datasource.js
```diff
@@ -69,6 +69,15 @@ angular.module(PKG.name+'.services')
         }
       }
 
+      // FIXME: There is a circular dependency and that is why
+      // myAuth.isAuthenticated is not used. There should be a better way to do this.
+      if ($rootScope.currentUser && $rootScope.currentUser.token) {
+        re.headers = {
+          Authorization: 'Bearer '+ $rootScope.currentUser.token,
+          'X-ApiKey': ''
```
anew: Is X-ApiKey needed? I thought that is from long ago and obsolete?

anew:
Code changes look OK to me. But I am not a js expert... 
I tried both of the test clusters (secure and insecure) and they appear to work fine.

cdap-ui/app/main.js
ajainarayanan: :) yeah.. Angular has Dependency Injection and it will resolve for circular dependencies at runtime.

cdap-ui/app/services/data/datasource.js
ajainarayanan: Removed.

cdap-ui/app/features/overview/controllers/overview-ctrl.js
```diff
@@ -46,7 +47,15 @@ function ($scope, $state, myLocalStorage, MY_CONFIG, Widget, MyMetricsQueryHelpe
     if (serviceStatuses.indexOf('NOTOK') === -1) {
       this.systemStatus = 'green';
     }
-  }.bind(this));
+  }.bind(this),
+  function error(res) {
+    if (res.auth_uri) {
+      EventPipe.emit('backendDown', 'Invalid Auth Token. Please refresh to re-authenticate.');
```
elfenheart: If Auth Token is invalid, do we want to disable the entire UI or just display a message? We can reroute to the login screen and display an error message right?

cdap-ui/app/services/namespace.js
```diff
@@ -58,21 +58,26 @@ angular.module(PKG.name + '.services')
 
       _.debounce(function() {
         $http.get('http://' + window.location.host + '/backendstatus', {ignoreLoadingBar: true})
-                .success(success).error(error);
-              }, 2000)();
+             .success(success)
```
elfenheart: Should we move this entire section somewhere else other than myNamespace?

ajainarayanan:
The nodeproxy seems to be down with Dataset services not starting up. GIF of how the change behaves in a non-secure standalone 
authnonsecure

elfenheart:
:+1: LGTM

PR #2967, opened 2015-06-16
adds search and pagination filtering to apps list

astewes:
Search input has been positioned to the right of the Actions dropdown since it requires more space overall. This will allow it to kick to a full row column on smaller devices. 
rwd-app-list


elfenheart:
:+1: LGTM

PR #2968, opened 2015-06-16
[CDAP-2683] [CDAP-2684] [CDAP-2681] Flow Graph Improvements

elfenheart:
- Remove Fork and Join node (turn it into invisible node
- Workflow becomes directed graph
- Disabled mousescroll zoom and add zoom in/zoom out & fit to screen buttons

Cluster: http://flowgraph1765-1000.dev.continuuity.net:9999/

workflow


ajainarayanan:
:+1: LGTM

PR #2971, opened 2015-06-17
Improvements to CDAP timeline

awholegunch:
These changes support modifications to the CDAP timeline, displayed at http://docs.cask.co/cdap/index.html. The changes modify the json-versions.js so that it results in a condensed timeline, and with improved formatting of the list of "Older" releases.

- Added targets to build and display the JSON file (json-versions.js) to the build script.
- Moved scripts that manipulate the JSON file from the common configuration to the high-level configuration file, so that the common conf.py is more just configuration rather than scripting.


duddun:
LGTM

PR #2973, opened 2015-06-17
[CDAP-2781] fix workflow diagram not consistently getting nodes status

elfenheart:
Fixing the problem where the list of nodes have not been resolved yet, but the polling of status for each nodes have started.


ajainarayanan:
:+1: LGTM

PR #2976, opened 2015-06-17
Refactor to move check program type early for run records cleaner

henry-cask:
Simple refactor to move checking program type early for invalid run records. 
Add DEBUG log level for any exception thrown in the cleaner thread.


henry-cask:
http://builds.cask.co/browse/CDAP-DUT2055-1

henry-cask:
Build passes

PR #2980, opened 2015-06-17
Fix Installation Guide instructions for SDK zip

awholegunch:
Add to the standalone/zip (which is linked on the Cask website as the "Installation Guide" for the SDK zip) instructions on the dependency requirements.

Staged at: http://docs-staging.cask.co/cdap/3.0.1-feature-r3.0.1_download_instructions/en/developers-manual/getting-started/standalone/zip.html
Current: http://docs.cask.co/cdap/3.0.0/en/developers-manual/getting-started/standalone/zip.html


duddun:
LGTM

PR #2981, opened 2015-06-17
Feature/cdap 1988 fix

ataylor4:
The fix involves creating an empty map if properties is null.

Build: http://builds.cask.co/browse/CDAP-DUT2057-1

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
@@ -231,9 +231,6 @@ public class DatasetInstanceHandler extends AbstractHttpHandler {
   private DatasetInstanceConfiguration  getInstanceConfiguration(HttpRequest request) {
     Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
     DatasetInstanceConfiguration creationProperties = GSON.fromJson(reader, DatasetInstanceConfiguration.class);
-    if (creationProperties.getProperties() == null) {
-      creationProperties = new DatasetInstanceConfiguration(creationProperties.getTypeName());
-    }
     fixProperties(creationProperties.getProperties());
     return creationProperties;
   }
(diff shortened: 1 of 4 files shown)


cdap-proto/src/main/java/co/cask/cdap/proto/DatasetInstanceConfiguration.java
```diff
@@ -16,6 +16,8 @@
 
 package co.cask.cdap.proto;
 
+import com.google.gson.internal.LinkedTreeMap;
```
jwang47: Avoid using internal classes.

cdap-proto/src/main/java/co/cask/cdap/proto/DatasetInstanceConfiguration.java
```diff
@@ -25,6 +27,11 @@
   private final String typeName;
   private final Map<String, String> properties;
 
+  public DatasetInstanceConfiguration(String typeName) {
+    this.typeName = typeName;
+    this.properties = new LinkedTreeMap<>();
```
jwang47: Use `ImmutableMap.of()` instead.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
```diff
@@ -231,6 +231,9 @@ public void update(HttpRequest request, HttpResponder responder, @PathParam("nam
   private DatasetInstanceConfiguration  getInstanceConfiguration(HttpRequest request) {
     Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
     DatasetInstanceConfiguration creationProperties = GSON.fromJson(reader, DatasetInstanceConfiguration.class);
+    if (creationProperties.getProperties() == null) {
```
jwang47: I don't like having extra logic modifying the deserialized object after a `GSON.fromJson` since if we want to deserialize `DatasetInstanceConfiguration` elsewhere, we may have to copy paste this logic there. Even if not, it would be confusing since there's different logic in deserializing `DatasetInstanceConfiguration` in different places, which would make things messy.

Instead, if you want to have GSON use a default value in case a field is missing, you can probably initialize default fields in the default constructor in `DatasetInstanceConfiguration`. Alternatively, you can use something like `FlowletSpecificationCodec` then configure your GSON instance to use it.

jwang47: Instead of having a default value for properties, could we just handle null properties somewhere in `DatasetInstanceHandler`? I feel like that would be the most simple way.

chtyim: You can simply have `DatasetInstanceConfiguration.getProperties()` implemented as:

```java
public Map<String, String> getProperties() {
  return properties == null ?? ImmutableMap.<String, String>of() : properties;
}
```

ataylor4:
The test passes, and it works running the curl command

ataylor4:
And properties can't be null - the constructor specifically makes it non-null

jwang47:
LGTM

chtyim:
But when doing GSON decode, the specific constructor won't get called, hence null not handled.

chtyim:
LGTM

PR #2983, opened 2015-06-17
Remove the extra create ZK node, resulting from merging from release/…

chtyim:
…3.0 branch.

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
@@ -180,10 +180,6 @@ public class MasterServiceMain extends DaemonMain {
     metricsCollectionService.startAndWait();
     serviceStore.startAndWait();
     leaderElection.startAndWait();
-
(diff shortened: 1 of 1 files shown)


cdap-master/src/main/java/co/cask/cdap/data/runtime/main/MasterServiceMain.java
```diff
@@ -180,10 +180,6 @@ public void start() {
     metricsCollectionService.startAndWait();
     serviceStore.startAndWait();
     leaderElection.startAndWait();
-
-    // Tries to create the ZK root node (which can be namespaced through the zk connection string)
```
chtyim: It's done in line 176 already.

anwar6953: sorry, I don't understand how the merge to develop didn't remove it from here?

chtyim: Because there was a pick from develop earlier.

anwar6953:
lgtm

PR #2984, opened 2015-06-17
Fix including of Spark example.

awholegunch:
Staged at: http://docs-staging.cask.co/cdap/3.0.1-feature-r3.0.1_fix_example/en/examples-manual/examples/spark-page-rank.html


jwang47:
LGTM

PR #2985, opened 2015-06-18
CDAP-2796, CDAP-2803, CDAP-2804, CDAP-2810: dataset metric fixes

abaranau:


chtyim:
So that is kind of backward incompatible, right?

abaranau:
UI uses dataset.store.XXX metrics. Docs had incorrectly stating store.XXX, though in example of metrics queries it showed correct metric.

abaranau:
http://builds.cask.co/browse/CDAP-DUT2061-1

abaranau:
@chtyim so potentially we may break some program (outside CDAP) that relies on that old metric name. I wouldn't worry about that though. We can add in release notes which metric to use instead

gsps1:
how about these metrics ? https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/DataSetRecordReader.java#L63

where store.ops is in program context and dataset.store.ops is in dataset context

abaranau:
hm. That thing looks wrong!

abaranau:
should not be there

awholegunch:
What about this file?
https://github.com/caskdata/cdap/blob/develop/cdap-watchdog/src/test/java/co/cask/cdap/metrics/query/MetricQueryParserTest.java#L371

abaranau:
@gsps1 this is wrong - what you found, the code is wrong. Created https://issues.cask.co/browse/CDAP-2803 https://issues.cask.co/browse/CDAP-2804. Will see if it makes sense to fix both with this PR

abaranau:
and https://issues.cask.co/browse/CDAP-2810

abaranau:
addressed the above three jiras as well, http://builds.cask.co/browse/CDAP-DUT2061-2

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
```diff
@@ -181,13 +181,13 @@ These metrics are available in a flowlet context:
...
    * - ``system.process.tuples.read``
      - Number of tuples read by the flowlet
-   * - ``system.store.bytes``
+   * - ``system.dataset.store.bytes``
      - Number of bytes written to datasets
-   * - ``system.store.ops``
+   * - ``system.dataset.store.ops``
      - Operations (writes and read) performed on datasets
-   * - ``system.store.reads``
+   * - ``system.dataset.store.reads``
```
awholegunch: Is there an error in this doc now, because the only instance of ``system.store.reads`` is in the example under *Search for Metrics*:
```
   * - HTTP Method
     - ``POST '<base-url>/metrics/search?target=metric&tag=namespace:default&tag=app:HelloWorld&tag=service:Greeting'``
   * - Returns
     - | ``["system.dataset.store.ops","system.dataset.store.reads","system.requests.count",``
       | `` "system.response.successful.count","system.store.ops","system.store.reads",``
```
yet, this does not appear under the service context, nor anywhere else.

awholegunch:
And, what about https://github.com/caskdata/cdap/blob/develop/cdap-watchdog/src/test/java/co/cask/cdap/metrics/query/MetricQueryParserTest.java#L371 ? Is that file correct?

cdap-docs/reference-manual/source/http-restful-api/metrics.rst
abaranau: fixed

abaranau:
MetricQueryParserTest is fine - it is a test name for the metric provided couple lines before that

abaranau:
build is GREEN

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
```diff
@@ -395,6 +397,12 @@ public Row get(byte[] row, byte[] startColumn, byte[] stopColumn, int limit) {
    */
   @Override
   public void put(byte[] row, byte[][] columns, byte[][] values) {
+    putInternal(row, columns, values);
+    // report metrics _after_ write was performed
+    reportWrite(1, getSize(row) + getSize(columns) + getSize(values));
```
chtyim: So this is reported after written to the in memory buffer?

abaranau: yes. We report here ops of the dataset.. We should also report datastore ops - in addition.

chtyim:
LGTM

PR #2989, opened 2015-06-18
CDAP-2802: Stop using HiveConf.ConfVars.defaultValue, to support Hive >0.13

jwang47:
https://issues.cask.co/browse/CDAP-2802
http://builds.cask.co/browse/CDAP-DUT2067

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
@@ -231,8 +231,10 @@ public class ExploreRuntimeModule extends RuntimeModule {
         setupClasspath();
 
         // Set local tmp dir to an absolute location in the twill runnable otherwise Hive complains
+        String localScratchPath = System.getProperty("java.io.tmpdir") + File.separator +
+          "hive-" + System.getProperty("user.name");
         System.setProperty(HiveConf.ConfVars.LOCALSCRATCHDIR.toString(),
-                           new File(HiveConf.ConfVars.LOCALSCRATCHDIR.defaultVal).getAbsolutePath());
+                           new File(localScratchPath).getAbsolutePath());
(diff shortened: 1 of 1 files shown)


cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
```diff
@@ -231,8 +231,10 @@ protected void configure() {
         setupClasspath();
 
         // Set local tmp dir to an absolute location in the twill runnable otherwise Hive complains
+        String localScratchPath = System.getProperty("java.io.tmpdir") + File.separator +
```
chtyim: Is this within container? How about in standalone?

jwang47: This is the same as the current behavior for Hive 0.13, 0.14, and 1.1 (haven't checked other ones).

chtyim: Preserving the current behavior may not be the right thing. Can you check if runs in standalone will it pollute the "/tmp"? (for distributed mode, it's probably ok, since "java.io.tmpdir" is set to container local directory).

jwang47: Read the code some more. This part is only for distributed mode. In standalone, we have the following:

```

        File hiveDataDir = new File(cConf.get(Constants.Explore.LOCAL_DATA_DIR));

        // The properties set using setProperty will be included to any new HiveConf object created,
        // at the condition that the configuration is known by Hive, and so is one of the HiveConf.ConfVars
        // variables.

        System.setProperty(HiveConf.ConfVars.SCRATCHDIR.toString(),
                           new File(hiveDataDir, cConf.get(Constants.AppFabric.TEMP_DIR)).getAbsolutePath());
```

so it's just using the data dir.

jwang47:
Reviewed in https://github.com/caskdata/cdap/pull/3004

PR #2994, opened 2015-06-18
Fixes non-explorable secure environments to show datasets even though its not explorable

ajainarayanan:
If explorer service is not enabled the UI shouldn't depend on it to show anything. Right now UI uses /explore/tables end point to view any dataset (to check if its explorable or not). If this call fails the UI should go to a default detailed view of a dataset.


ajainarayanan:
Closing this PR created against develop as release/3.0 will eventually be merged to develop.

PR #2996, opened 2015-06-18
CDAP-2811: Improve Hive version detection

jwang47:
https://issues.cask.co/browse/CDAP-2811
http://builds.cask.co/browse/CDAP-DUT2070

* Works in standalone and distributed

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
@@ -43,6 +43,7 @@ import com.google.common.io.Files;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.mapreduce.Job;
 import org.apache.hadoop.util.VersionInfo;
+import org.apache.hive.common.util.HiveVersionInfo;
 import org.apache.twill.internal.utils.Dependencies;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
```diff
@@ -191,26 +179,16 @@ public static HiveSupport checkHiveSupport(ClassLoader hiveClassLoader) {
...
+      } else if (hiveVersion.startsWith("0.13.")) {
+        return HiveSupport.HIVE_13;
       }
+
       throw new RuntimeException("Hive distribution not supported. Set the configuration '" +
-                                 Constants.Explore.EXPLORE_ENABLED +
-                                 "' to false to start up without Explore.");
+                                   Constants.Explore.EXPLORE_ENABLED +
+                                   "' to false to start up without Explore.");
     } catch (RuntimeException e) {
```
chtyim: does this still applies?

jwang47: I think so, since we don't support every version of Hive.

chtyim: but catching a `RuntimeException` and throwing it is meaningless. Why catch and rethrow?

chtyim: In fact, if the `HiveVersionInfo.getVersion()` doesn't throw checked exception, there is no need for a try block here.

jwang47: True, I misread and thought your comment was on the throw line. Fixed.

jwang47:
Reviewed in https://github.com/caskdata/cdap/pull/3004

PR #2997, opened 2015-06-18
[CDAP-2739] Include S3 as a batch source for ETL

jayfjin:
Build - http://builds.cask.co/browse/CDAP-DUT2023-4
JIRA - https://issues.cask.co/browse/CDAP-2739

PR to include S3 as a batch source for ETL. Has already been tested with the UI and works with statefulness.


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+
+  private static KeyValueTable table;
+  private static Date prevMinute;
+  private static String currentTime;
+
+  private static final String REGEX_DESCRIPTION = "Regex to filter out filenames in the path. " +
+    "To use the TimeFilter, input \"timefilter\". The TimeFilter assumes that we " +
+    "are reading in files with the S3 log naming convention of YYYY-MM-DD-HH-mm-SS-Tag. The TimeFilter " +
+    "reads in files from the previous hour if the timeTable field is left blank. So if it's currently " +
+    "2015-06-16-15, (June 16th 2015, 3pm), it will read in files that contain 2015-06-16-14 in the filename. " +
```
awholegunch: 2015-06-16-15 (June 16th 2015, 3pm) : delete first comma.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/org/apache/hadoop/fs/FileSystemDummy.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.io.IOException;
+import java.net.URI;
+
+/**
+ * Dummy class to use a private static method in FileSystem for testing purposes.
+ */
+public class FileSystemDummy {
+  public static void addFileSystemForTesting(URI uri, Configuration conf,
+                                      FileSystem fs) throws IOException {
```
anwar6953: `FileSystem` on the second line should align with `URI`

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/ETLStreamConversionTest.java
```diff
@@ -104,7 +104,7 @@ private ETLBatchConfig constructETLBatchConfig(String fileSetName) {
     return new ETLBatchConfig("* * * * *", source, sink, Lists.newArrayList(transform));
   }
 
-  private List<GenericRecord> readOutput(TimePartitionedFileSet fileSet, Schema schema) throws IOException {
+  public static List<GenericRecord> readOutput(TimePartitionedFileSet fileSet, Schema schema) throws IOException {
```
chtyim: why ? static is bad. public static is really bad.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/ETLMapReduceTest.java
```diff
@@ -170,4 +178,54 @@ public void testTableToTable() throws Exception {
...
+
+    manager.start();
+    manager.waitForOneRunToFinish(2, TimeUnit.MINUTES);
+    manager.stop();
+
+    DataSetManager<TimePartitionedFileSet> fileSetManager = getDataset("TPFSsink");
+    TimePartitionedFileSet fileSet = fileSetManager.get();
+    List<GenericRecord> records = ETLStreamConversionTest.readOutput(fileSet, defaultSchema);
+    Assert.assertEquals(1, records.size());
+    System.out.println(records.get(0).get("body"));
```
anwar6953: remove println

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+    "If the field timeTable is present, then it will read files in that haven't been read yet.";
+  private static final String NAME_DESCRIPTION = "Name of the S3 source.";
+  private static final String ACCESS_ID_DESCRIPTION = "AccessID of the S3 source.";
+  private static final String ACCESS_KEY_DESCRIPTION = "AccessKey of the S3 source.";
+  private static final String PATH_DESCRIPTION = "Path to file(s) to be read. If a directory is specified," +
+    " terminate the path name with a \'/\'";
```
awholegunch: Why not be consistent, and terminate all previous lines with a space, so that following lines always start the same, tight to the double-quote?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/org/apache/hadoop/fs/FileSystemDummy.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.hadoop.conf.Configuration;
+
+import java.io.IOException;
+import java.net.URI;
+
+/**
+ * Dummy class to use a private static method in FileSystem for testing purposes.
+ */
+public class FileSystemDummy {
```
chtyim: just use reflection to call the method. No need to have a dummy class.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+
+  //length of 'YYYY-MM-dd-HH-mm"
+  private static final int DATE_LENGTH = 16;
+  private static final Logger LOG = LoggerFactory.getLogger(S3BatchSource.class);
+  private S3BatchConfig config;
+
+  private static final Schema DEFAULT_SCHEMA = Schema.recordOf(
+    "event",
+    Schema.Field.of("ts", Schema.of(Schema.Type.LONG)),
+   Schema.Field.of("body", Schema.of(Schema.Type.STRING))
```
anwar6953: align vertically with previous parameter

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+@Name("S3")
+@Description("Batch source for S3")
+public class S3BatchSource extends BatchSource<LongWritable, Text, StructuredRecord> {
+
+  private static KeyValueTable table;
+  private static Date prevMinute;
+  private static String currentTime;
+
+  private static final String REGEX_DESCRIPTION = "Regex to filter out filenames in the path. " +
+    "To use the TimeFilter, input \"timefilter\". The TimeFilter assumes that we " +
```
awholegunch: The TimeFilter assumes that it " + "is reading in files with the S3 log... \[usually don't use "we"]

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
awholegunch: But where is the 3pm in the example?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
awholegunch: inconsistent indentation

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+    if (config.timeTable != null) {
+      pipelineConfigurer.createDataset(config.timeTable, KeyValueTable.class, DatasetProperties.EMPTY);
+    }
+  }
+
+  @Override
+  public void prepareRun(BatchSourceContext context) throws Exception {
+    //calculate date one minute ago, rounded down to the nearest minute
+    prevMinute = new Date();
+    prevMinute.setTime(System.currentTimeMillis() - TimeUnit.MINUTES.toMillis(1));
```
anwar6953: this can just be constructor parameter (in previous line), right?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+   Schema.Field.of("body", Schema.of(Schema.Type.STRING))
+  );
+
+  public S3BatchSource (S3BatchConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    //will create the table if it doesn't exist already
```
awholegunch: Javadocs should be added to this class, and this comment should be in it.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+    if (config.timeTable != null) {
+      pipelineConfigurer.createDataset(config.timeTable, KeyValueTable.class, DatasetProperties.EMPTY);
+    }
+  }
+
+  @Override
+  public void prepareRun(BatchSourceContext context) throws Exception {
+    //calculate date one minute ago, rounded down to the nearest minute
```
awholegunch: This should be in a Javadoc comment for this class.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
anwar6953: Since it's an overriding method, the method already has a javadoc, right

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+          String currentTime = new SimpleDateFormat("yyyy-MM-dd-HH").format(prevHour);
+
+          if (filename.contains(currentTime)) {
+            return true;
+          }
+          return false;
+        }
+      }
+
+      //use regex
```
chtyim: This comment gives no value. Maybe better as "filter by file name using regex from configuration".

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+    StructuredRecord output = StructuredRecord.builder(DEFAULT_SCHEMA)
+      .set("ts", System.currentTimeMillis())
+      .set("body", input.getValue().toString())
+      .build();
+    emitter.emit(output);
+  }
+
+  @Override
+  public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    if (succeeded && table != null && config.fileRegex != null && config.fileRegex.equals("timefilter")) {
```
anwar6953: Is it a good idea to simplify:
replace
```java
config.fileRegex != null && config.fileRegex.equals("timefilter")
```
with
```java
"timefilter".equals(config.fileRegex)
```

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+   */
+  public static class S3Filter extends Configured implements PathFilter {
+    private boolean useTimeFilter;
+    private Pattern regex;
+    private String pathName;
+
+    @Override
+    public boolean accept(Path path) {
+      String filename = path.toString();
+      System.out.println("attempt :" + filename);
```
anwar6953: remove println

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+        }
+      }
+
+      //use regex
+      Matcher matcher = regex.matcher(filename);
+      return matcher.matches();
+    }
+
+    public void setConf(Configuration conf) {
+      if (conf != null) {
```
chtyim: Prefer inverted logic to exit early.

```java
if (conf == null) {
  return;
}
// normal logic
```

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,271 @@
...
+            return true;
+          }
+          return false;
+        } else {
+          //use hourly time filter
+          Date prevHour = new Date();
+          prevHour.setTime(System.currentTimeMillis() - TimeUnit.HOURS.toMillis(1));
+          String currentTime = new SimpleDateFormat("yyyy-MM-dd-HH").format(prevHour);
+
+          if (filename.contains(currentTime)) {
```
anwar6953: replace this line and following t3 lines with the following? :
```java
return filename.contains(currentTime)
```

awholegunch:
Will there be some modifications to the docs to add this? (http://docs-staging.cask.co/cdap/3.1.0-SNAPSHOT/en/application-templates/etl/index.html)?

awholegunch:
Also, when you create a PR, please add appropriate labels (using the label menu on the right-hand side; ask someone to show you if you can't find it.) Merges to develop should be labelled "3.1".

awholegunch:
You can also assign it to @bdmogal :)

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,263 @@
...
+
+    @Override
+    public boolean accept(Path path) {
+      String filename = path.toString();
+      //InputPathFilter will first check the directory if a directory is given
+      if (filename.equals(pathName) || filename.equals(pathName + "/")) {
+        return true;
+      }
+
+      if (useTimeFilter) {
```
anwar6953: I think if you invert this condition, you can then put lines 197-198 in the if block.
And then lines 163-193 can be unindented. Confirm that it would be equivalent.
In my opinion, at least, the less indentation/nesting, the better.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,263 @@
...
+          Date fileDate;
+          try {
+            fileDate = sdf.parse(path.getName().substring(0, DATE_LENGTH));
+          } catch (ParseException pe) {
+            //this should never happen
+            LOG.warn("Couldn't parse file: " + path.getName());
+            fileDate = prevMinute;
+          }
+
+          if (fileDate.compareTo(dateLastRead) > 0 && fileDate.compareTo(prevMinute) <= 0) {
```
anwar6953: Also, should this be replaced by
```java
return fileDate.compareTo(dateLastRead) > 0 && fileDate.compareTo(prevMinute) <= 0;
```
or do you think that this is better for clarity?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,263 @@
...
+    @Description(REGEX_DESCRIPTION)
+    private String fileRegex;
+
+    @Name("timeTable")
+    @Nullable
+    @Description(TABLE_DESCRIPTION)
+    private String timeTable;
+
+    public S3BatchConfig(String name, String accessID, String accessKey, String path, String regex,
+                         String timeTable) {
```
anwar6953: Annotate the parameters in the constructor as `@Nullable`, if they are.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,263 @@
...
+
+    public void setConf(Configuration conf) {
+      if (conf == null) {
+        return;
+      }
+      pathName = conf.get("input.path.name", "/");
+
+      //path is a directory so remove trailing '/'
+      if (pathName.endsWith("/")) {
+        pathName = pathName.substring(0, pathName.length() - 1);
```
anwar6953: In case the `conf` is missing the "input.path.name" field, then this would still be true, and pathName would then be empty string, right?

jayfjin: Yes, and path is not a nullable parameter, so input.path.name should never be missing anyways.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BaseETLBatchTest.java
```diff
@@ -70,4 +78,29 @@ public static void setupTest() throws IOException {
       PipelineConfigurable.class.getPackage().getName(),
       BatchSource.class.getPackage().getName());
   }
+
+  protected List<GenericRecord> readOutput(TimePartitionedFileSet fileSet, Schema schema) throws IOException {
+    org.apache.avro.Schema avroSchema = new org.apache.avro.Schema.Parser().parse(schema.toString());
+    DatumReader<GenericRecord> datumReader = new GenericDatumReader<>(avroSchema);
+    List<GenericRecord> records = com.google.common.collect.Lists.newArrayList();
```
chtyim: Why full class name?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BaseETLBatchTest.java
```diff
@@ -70,4 +78,29 @@ public static void setupTest() throws IOException {
...
+      for (Location timeLoc : dayLoc.list()) {
+        // this level should be the time (ex: 21-23.1234567890000)
+        for (Location file : timeLoc.list()) {
+          // this level should be the actual mapred output
+          String locName = file.getName();
+
+          if (locName.endsWith(".avro")) {
+            DataFileStream<GenericRecord> fileStream =
+              new DataFileStream<>(file.getInputStream(), datumReader);
+            while (fileStream.hasNext()) {
```
chtyim: `DataFileStream` implements `Iterable`. You can simply do `Iterables.addAll(records, fileStream)`. Also, you need to close the stream.

```java
try (DataFileStream<GenericRecord> fs = new ...) {
  Iterables.addAll(records, fs);
}
```

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/ETLMapReduceTest.java
```diff
@@ -170,4 +179,56 @@ public void testTableToTable() throws Exception {
...
+
+    manager.start();
+    manager.waitForOneRunToFinish(2, TimeUnit.MINUTES);
+    manager.stop();
+
+    DataSetManager<TimePartitionedFileSet> fileSetManager = getDataset("TPFSsink");
+    TimePartitionedFileSet fileSet = fileSetManager.get();
+    List<GenericRecord> records = readOutput(fileSet, defaultSchema);
+    Assert.assertEquals(1, records.size());
+    Assert.assertEquals(records.get(0).get("body").toString(), testData);
```
chtyim: The expected value should be the first parameter.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * A {@link BatchSource} to use S3 as a Source.
+ */
+@SuppressWarnings("unused")
+@Plugin(type = "source")
+@Name("S3")
+@Description("Batch source for S3")
+public class S3BatchSource extends BatchSource<LongWritable, Text, StructuredRecord> {
+
+  private static KeyValueTable table;
```
chtyim: Why static?

chtyim: We shouldn't use non final static field (meaning static field only for constants).

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+    //will create the table if it doesn't exist already
+    if (config.timeTable != null) {
+      pipelineConfigurer.createDataset(config.timeTable, KeyValueTable.class, DatasetProperties.EMPTY);
+    }
+  }
+
+  @Override
+  public void prepareRun(BatchSourceContext context) throws Exception {
+    //calculate date one minute ago, rounded down to the nearest minute
+    prevMinute = new Date(System.currentTimeMillis() - TimeUnit.MINUTES.toMillis(1));
```
chtyim: Why use the current time? Shouldn't we use logical start time?

chtyim: Also, why always one minute ago? Shouldn't it based on the schedule?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+        dateLastRead.setTime(0);
+      }
+
+      Date fileDate;
+      try {
+        fileDate = sdf.parse(path.getName().substring(0, DATE_LENGTH));
+      } catch (ParseException pe) {
+        //this should never happen
+        LOG.warn("Couldn't parse file: " + path.getName());
+        fileDate = prevMinute;
```
chtyim: You cannot use `prevMinute` like this. An instance of class is created per Mapper process and the prepareRun of the outer class won't get called. In fact, you should treat this class as a top level class.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+    if (config.timeTable != null) {
+      table = context.getDataset(config.timeTable);
+    }
+
+    FileInputFormat.setInputPathFilter(job, S3Filter.class);
+    FileInputFormat.addInputPath(job, new Path(config.path));
+  }
+
+  @Override
+  public void transform(KeyValue<LongWritable, Text> input, Emitter<StructuredRecord> emitter) throws Exception {
```
chtyim: So this S3 batch source can only support text file format?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
jayfjin: that's the point, I want prevMinute to be updated once during prepareRun so that it remains constant across all calls of the filter class.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+  public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    if (succeeded && table != null && "timefilter".equals(config.fileRegex)) {
+      table.write("lastTimeRead", new SimpleDateFormat("yyyy-MM-dd-HH-mm").format(prevMinute));
+    }
+  }
+
+  /**
+   * Filter class to filter out filenames in the input path.
+   */
+  public static class S3Filter extends Configured implements PathFilter {
```
chtyim: Seems to me this class is pretty generic. In fact, shouldn't we have something like `FileBatchSource` that allow any `InputFormat` class to be used, along with set of properties to set to the job conf (in this particular case, S3 parameters)?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
chtyim: It won't work. Each Mapper process will have a new instance of this filter class, which the `prevMinute` static field in the outer class not set (because prepareRun will not be called).

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+    @Description(REGEX_DESCRIPTION)
+    private String fileRegex;
+
+    @Name("timeTable")
+    @Nullable
+    @Description(TABLE_DESCRIPTION)
+    private String timeTable;
+
+
+    public S3BatchConfig(String name, String accessID, String accessKey, String path, @Nullable String regex,
```
chtyim: Why have this constructor? Who is calling it?

cdap-app-templates/cdap-etl/cdap-etl-batch/pom.xml
```diff
@@ -62,6 +62,12 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>org.apache.hadoop</groupId>
```
bdmogal: should add a note here to switch to {{hadoop-aws}} when we upgrade to 2.6.0

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/org/apache/hadoop/fs/s3native/S3NInMemoryFileSystem.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * NOTE: DELETE THIS CLASS ONCE HADOOP DEPENDENCY IS UPDATED TO 2.6.0
+ * Used for testing S3 without actually connection to an S3 instance.
+ */
+public class S3NInMemoryFileSystem extends NativeS3FileSystem {
+  public S3NInMemoryFileSystem() {
+    super(new InMemoryNativeFileSystemStore());
+  }
+}
+
```
bdmogal: extra blank line

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+import java.util.Date;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/**
+ * A {@link BatchSource} to use S3 as a Source.
+ */
+@SuppressWarnings("unused")
```
bdmogal: Better to add this only where required. not for the entire class.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+  private static final String ACCESS_KEY_DESCRIPTION = "AccessKey of the S3 source.";
+  private static final String PATH_DESCRIPTION = "Path to file(s) to be read. If a directory is specified, " +
+    "terminate the path name with a \'/\'";
+  private static final String TABLE_DESCRIPTION = "Name of the Table that keeps track of the last time files " +
+    "were read in.";
+
+  //length of 'YYYY-MM-dd-HH-mm"
+  private static final int DATE_LENGTH = 16;
+  private static final Logger LOG = LoggerFactory.getLogger(S3BatchSource.class);
+  private S3BatchConfig config;
```
bdmogal: final

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
bdmogal: Is this comment adding much value though? Should probably remove it.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/S3BatchSource.java
```diff
@@ -0,0 +1,257 @@
...
+  );
+
+  public S3BatchSource (S3BatchConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    //will create the table if it doesn't exist already
+    if (config.timeTable != null) {
```
bdmogal: What happens when timeTable is null?

jayfjin: the table isn't created/loaded, and information won't be passed to the file filter class

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/ETLMapReduceTest.java
```diff
@@ -170,4 +179,56 @@ public void testTableToTable() throws Exception {
     Assert.assertEquals("island", row.getString("item"));
   }
 
+  @Test
+  public void testS3toTPFS() throws Exception {
+    String testPath = "s3n://test/2015-06-17-00-00-00.txt";
```
bdmogal: I'm not sure how s3 works, but is this a valid URI? the component between the second and third slashes is supposed to be an authority (host:port). I guess its ok coz you're using an in-memory s3?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/org/apache/hadoop/fs/s3native/S3NInMemoryFileSystem.java
```diff
@@ -0,0 +1,27 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.hadoop.fs.s3native;
+
+/**
+ * NOTE: DELETE THIS CLASS ONCE HADOOP DEPENDENCY IS UPDATED TO 2.6.0
+ * Used for testing S3 without actually connection to an S3 instance.
```
bdmogal: typo: 'connecting'

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/org/apache/hadoop/fs/s3native/S3NInMemoryFileSystem.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.hadoop.fs.s3native;
+
+/**
+ * NOTE: DELETE THIS CLASS ONCE HADOOP DEPENDENCY IS UPDATED TO 2.6.0
```
bdmogal: Lets create a Jira, then convert this note to a ``TODO:`` with the Jira in it. Also, lets remove the upper-casing.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    "were read in.";
+  private static final String INPUT_FORMAT_CLASS_DESCRIPTION = "Name of the input format class. Defaults to " +
+    "textInputFormat";
+  private static final String FILE_SYSTEM_DESCRIPTION = "Distributed file system to read in from. Currently " +
+    "only supports S3.";
+
+  private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
+  private S3BatchConfig config;
+
+  private static final Schema DEFAULT_SCHEMA = Schema.recordOf(
```
bdmogal: Looks like we're defining ``DEFAULT_SCHEMA`` at multiple places in this PR itself. Can you check all the instances of this definition and see if we can pull it out to some common place? If there's no better place, we should at least move it to ``Constants.java``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+      String lastTimeRead = Bytes.toString(table.read("lastTimeRead"));
+      if (lastTimeRead == null) {
+        lastTimeRead = "0";
+      }
+      job.getConfiguration().set("last.time.read", lastTimeRead);
+    }
+
+    job.getConfiguration().set("cutoff.read.time", new SimpleDateFormat("yyyy-MM-dd-HH-mm").format(prevMinute));
+    if (config.inputFormatClass != null) {
+      Class<?> classType = Class.forName(config.inputFormatClass);
```
bdmogal: This is likely not right. Does this work on standalone/on a cluster? I think we should instead either do:
```
ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
Class<? extends FileInputFormat> classType = classLoader.loadClass(config.inputFormatClass)
```

or

```
Configuration conf = job.getConfiguration();
ClassLoader classLoader = conf.getClassLoader();
Class<? extends FileInputFormat> = classLoader.loadClass(config.inputFormatClass)
```

bdmogal: Also note that you should use ``Class<? extends FileInputFormat>`` instead of ``Class<?>`` to make it stricter.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    "2015-06-16-15 (June 16th 2015, 3pm), it will read in files that contain 2015-06-16-14 in the filename. " +
+    "If the field timeTable is present, then it will read files in that haven't been read yet.";
+  private static final String ACCESS_ID_DESCRIPTION = "AccessID of the S3 source.";
+  private static final String ACCESS_KEY_DESCRIPTION = "AccessKey of the S3 source.";
+  private static final String PATH_DESCRIPTION = "Path to file(s) to be read. If a directory is specified, " +
+    "terminate the path name with a \'/\'";
+  private static final String TABLE_DESCRIPTION = "Name of the Table that keeps track of the last time files " +
+    "were read in.";
+  private static final String INPUT_FORMAT_CLASS_DESCRIPTION = "Name of the input format class. Defaults to " +
+    "textInputFormat";
```
bdmogal: *T*extInputFormat

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    "reads in files from the previous hour if the timeTable field is left blank. So if it's currently " +
+    "2015-06-16-15 (June 16th 2015, 3pm), it will read in files that contain 2015-06-16-14 in the filename. " +
+    "If the field timeTable is present, then it will read files in that haven't been read yet.";
+  private static final String ACCESS_ID_DESCRIPTION = "AccessID of the S3 source.";
+  private static final String ACCESS_KEY_DESCRIPTION = "AccessKey of the S3 source.";
+  private static final String PATH_DESCRIPTION = "Path to file(s) to be read. If a directory is specified, " +
+    "terminate the path name with a \'/\'";
+  private static final String TABLE_DESCRIPTION = "Name of the Table that keeps track of the last time files " +
+    "were read in.";
+  private static final String INPUT_FORMAT_CLASS_DESCRIPTION = "Name of the input format class. Defaults to " +
```
bdmogal: Mention in the description that this class should be a subclass of ``FileInputFormat``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+@Name("S3")
+@Description("Batch source for S3")
+public class FileBatchSource extends BatchSource<LongWritable, Object, StructuredRecord> {
+
+  private KeyValueTable table;
+  private Date prevMinute;
+
+  private static final String REGEX_DESCRIPTION = "Regex to filter out filenames in the path. " +
+    "To use the TimeFilter, input \"timefilter\". The TimeFilter assumes that it " +
+    "are reading in files with the S3 log naming convention of YYYY-MM-DD-HH-mm-SS-Tag. The TimeFilter " +
```
bdmogal: Typo: **is** reading. 

Also, the S3 stuff may not be valid anymore?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    "reads in files from the previous hour if the timeTable field is left blank. So if it's currently " +
+    "2015-06-16-15 (June 16th 2015, 3pm), it will read in files that contain 2015-06-16-14 in the filename. " +
+    "If the field timeTable is present, then it will read files in that haven't been read yet.";
+  private static final String ACCESS_ID_DESCRIPTION = "AccessID of the S3 source.";
```
bdmogal: Hmmm, seems strange that we have s3 stuff in this class now. To make this class truly generic, we shouldn't have any  s3 properties at all. This class should be able to get a json string perhaps of a list/map of properties for a Batch File Source. Then you should iterate over that json to set all those properties in ``job.getConfiguration()``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+  private static final String ACCESS_KEY_DESCRIPTION = "AccessKey of the S3 source.";
```
bdmogal: Same comment as above

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+  public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    if (succeeded && table != null && "timefilter".equals(config.fileRegex)) {
+      table.write("lastTimeRead", new SimpleDateFormat("yyyy-MM-dd-HH-mm").format(prevMinute));
+    }
+  }
+
+  /**
+   * Config class that contains all the properties needed for the S3 source.
+   */
+  public static class S3BatchConfig extends PluginConfig {
```
bdmogal: Should be called ``FileBatchConfig``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+
+    @Name("fileSystem")
+    @Description(FILE_SYSTEM_DESCRIPTION)
+    private String fileSystem;
+
+    @Name("accessID")
+    @Nullable
+    @Description(ACCESS_ID_DESCRIPTION)
+    private String accessID;
```
bdmogal: See my comment here for this and the next field - https://github.com/caskdata/cdap/pull/2997/files#r32881761

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    prevMinute = new Date(context.getLogicalStartTime() - TimeUnit.MINUTES.toMillis(1));
+    Calendar cal = Calendar.getInstance();
+    cal.setTime(prevMinute);
+    cal.set(Calendar.SECOND, 0);
+    cal.set(Calendar.MILLISECOND, 0);
+    prevMinute = cal.getTime();
+
+    Job job = context.getHadoopJob();
+    if (!config.fileSystem.equals("S3")) {
+      LOG.warn("File system other than S3 not currently supported, using S3");
```
bdmogal: Hmmm, if it is not yet supported, maybe we should throw an ``InvalidConfigException`` or something and exit from here.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+    job.getConfiguration().set("fs.s3n.awsAccessKeyId", config.accessID);
```
bdmogal: So these two lines should replace the loop/iteration I mentioned in https://github.com/caskdata/cdap/pull/2997/files#r32881761

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    job.getConfiguration().set("fs.s3n.awsSecretAccessKey", config.accessKey);
+
+    if (config.fileRegex != null) {
+      job.getConfiguration().set("input.path.regex", config.fileRegex);
+    }
+    job.getConfiguration().set("input.path.name", config.path);
```
bdmogal: Should we extract these strings to constants somewhere?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+      .set("ts", System.currentTimeMillis())
+      .set("body", input.getValue().toString())
+      .build();
+    emitter.emit(output);
+  }
+
+  @Override
+  public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    if (succeeded && table != null && "timefilter".equals(config.fileRegex)) {
+      table.write("lastTimeRead", new SimpleDateFormat("yyyy-MM-dd-HH-mm").format(prevMinute));
```
bdmogal: Should extract this ``SimpleDataFormat`` as a constant, since it will never change

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+  }
+
+  /**
+   * Config class that contains all the properties needed for the S3 source.
```
bdmogal: s/S3/File/

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+   */
+  public static class S3BatchConfig extends PluginConfig {
+
+    @Name("fileSystem")
+    @Description(FILE_SYSTEM_DESCRIPTION)
+    private String fileSystem;
```
bdmogal: So this is only used to check if the filesystem in question is supported, right? I'm not sure it provides much value, unless you use it to derive some other configuration for the filesystem. What do you think?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,201 @@
...
+    @Nullable
+    @Description(TABLE_DESCRIPTION)
+    private String timeTable;
+
+    @Name("inputFormatClass")
+    @Nullable
+    @Description(INPUT_FORMAT_CLASS_DESCRIPTION)
+    private String inputFormatClass;
+
+    public S3BatchConfig(String fileSystem, @Nullable String accessID, @Nullable String accessKey, String path,
```
bdmogal: Better to have all the ``@Nullable`` params at the end

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/StreamBatchSource.java
```diff
@@ -55,7 +56,7 @@
 @Plugin(type = "source")
 @Name("Stream")
 @Description("Batch source for a Stream")
-public class StreamBatchSource extends BatchSource<LongWritable, Object, StructuredRecord> {
+public class StreamBatchSource extends BatchSource<LongWritable, Writable, StructuredRecord> {
```
bdmogal: is this change needed?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Filter class to filter out filenames in the input path.
+ */
+public class BatchFileFilter extends Configured implements PathFilter {
+  private boolean useTimeFilter;
+  private Pattern regex;
+  private String pathName;
+  private String lastRead;
+  private Date prevMinute;
+  private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
```
bdmogal: we generally have all the ``private static final`` fields at the beginning

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/Properties.java
```diff
@@ -37,6 +37,19 @@
   }
 
   /**
+   * Class to hold properties for S3BatchSource
+   */
+  public static class S3 {
```
bdmogal: rename to ``File``?

cdap-app-templates/cdap-etl/pom.xml
```diff
@@ -156,6 +156,43 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
```
bdmogal: Also add a TODO here to use ``hadoop-aws`` when we move to 2.6.0

bdmogal:
@jayfjin some comments about making this more generic. Also, we should add the test you have for local filesystem as well, just so we verify it works for more than S3.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/FileBatchProperties.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package co.cask.cdap.template.etl.common;
+
+import java.util.Map;
+
+/**
+ * Class used to parse the JSON passed into {@link co.cask.cdap.template.etl.batch.source.FileBatchSource}
+ */
+public class FileBatchProperties {
```
bdmogal: Shouldn't need this class. Just deserialize it to a ``Map<String, String>``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,211 @@
...
+    "terminate the path name with a \'/\'";
+  private static final String TABLE_DESCRIPTION = "Name of the Table that keeps track of the last time files " +
+    "were read in.";
+  private static final String INPUT_FORMAT_CLASS_DESCRIPTION = "Name of the input format class, which must be a " +
+    "subclass of FileInputFormat. Defaults to TextInputFormat";
+  private static final String FILESYSTEM_DESCRIPTION = "Distributed file system to read in from.";
+  private static final Gson GSON = new Gson();
+  private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
+  private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
+  private static final Schema DEFAULT_SCHEMA = Schema.recordOf(
```
bdmogal: Let's move this to ``ETLUtils.java`` for now, and re-use it in this class and in the ``StreamBatchSource`` (or wherever else it is used)

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,211 @@
...
+      String lastTimeRead = Bytes.toString(table.read("lastTimeRead"));
+      if (lastTimeRead == null) {
+        lastTimeRead = "0";
+      }
+      job.getConfiguration().set(LAST_TIME_READ, lastTimeRead);
+    }
+
+    job.getConfiguration().set(CUTOFF_READ_TIME, DATE_FORMAT.format(prevMinute));
+    if (config.inputFormatClass != null) {
+      ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
```
bdmogal: Can you confirm this with @chtyim

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,211 @@
...
+      Class<?> classType = classLoader.loadClass(config.inputFormatClass);
```
bdmogal: I think you missed my previous comment here. Make this type ``Class<? extends FileInputFormat>``, so you don't need lines 142-147 below.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,211 @@
...
+    @Nullable
+    @Description(TABLE_DESCRIPTION)
+    private String timeTable;
+
+    @Name("inputFormatClass")
+    @Nullable
+    @Description(INPUT_FORMAT_CLASS_DESCRIPTION)
+    private String inputFormatClass;
+
+    public FileBatchConfig(String fileSystem, @Nullable String fileSystemProperties, String path,
```
bdmogal: nit: Move ``@Nullable`` fields to the end

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -0,0 +1,117 @@
...
+
+  private Pattern regex;
+  private String pathName;
+  private String lastRead;
+  private Date prevMinute;
+
+  @Override
+  public boolean accept(Path path) {
+    String filename = path.toString();
+    //InputPathFilter will first check the directory if a directory is given
```
bdmogal: What is ``InputPathFilter``?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,210 @@
...
+  @Override
+  public void prepareRun(BatchSourceContext context) throws Exception {
+    //calculate date one minute ago, rounded down to the nearest minute
+    prevMinute = new Date(context.getLogicalStartTime() - TimeUnit.MINUTES.toMillis(1));
+    Calendar cal = Calendar.getInstance();
+    cal.setTime(prevMinute);
+    cal.set(Calendar.SECOND, 0);
+    cal.set(Calendar.MILLISECOND, 0);
+    prevMinute = cal.getTime();
+    System.out.print(FILESYSTEM_PROPERTIES_DESCRIPTION);
```
bdmogal: Shouldn't have a sysout statement

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,210 @@
...
+    Job job = context.getHadoopJob();
+    if (config.fileSystemProperties != null) {
+      Map<String, String> properties = GSON.fromJson(config.fileSystemProperties, MAP_STRING_STRING_TYPE);
+      for (Map.Entry<String, String> entry : properties.entrySet()) {
+        job.getConfiguration().set(entry.getKey(), entry.getValue());
```
bdmogal: nit: To make the code more readable, is it better to extract ``job.getConfiguration()`` out to a variable

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    job.getConfiguration().set(CUTOFF_READ_TIME, DATE_FORMAT.format(prevMinute));
+    if (config.inputFormatClass != null) {
+      ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+      Class<? extends FileInputFormat> classType = (Class<? extends FileInputFormat>)
+        classLoader.loadClass(config.inputFormatClass);
+      job.setInputFormatClass(classType);
+    }
+    FileInputFormat.setInputPathFilter(job, BatchFileFilter.class);
```
bdmogal: Eventually, I think we need to allow users to specify their own Filters. Should probably add a jira to allow this to be configured as well, in a future PR. What do you think?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -0,0 +1,117 @@
...
+      fileDate = sdf.parse(path.getName().substring(0, DATE_LENGTH));
+    } catch (ParseException pe) {
+      //this should never happen
+      LOG.warn("Couldn't parse file: " + path.getName());
+      fileDate = prevMinute;
+    }
+    return fileDate.compareTo(dateLastRead) > 0 && fileDate.compareTo(prevMinute) <= 0;
+  }
+
+  public void setConf(Configuration conf) {
```
bdmogal: Annotate with ``@Override``

bdmogal:
LGTM

PR #3001, opened 2015-06-19
Fixes repeated alerts and confusion about opened tabs in adapter create view

ajainarayanan:
- Should be able to add the same source/sink any number of times the user want. The UI should still create only object (no overwrites should happen)
- When plugin's property is already being edited (opened in a new tab) and if the plugin is either deleted or overwritten then the tab should be removed to avoid any confusion.
- When publish button is clicked and when the UI does client side validations only one alert message should be opened at a time. No repeated multiple alert messages.

Secure Cluster: http://securecluster1762-1000.dev.continuuity.net
Non-Secure Cluster: http://nodeproxy1736-1000.dev.continuuity.net:9999/

multiplealerts


elfenheart:
:+1: LGTM

PR #3004, opened 2015-06-19
CDAP-2798: Hive 0.14 support

jwang47:
https://issues.cask.co/browse/CDAP-2798
http://builds.cask.co/browse/CDAP-DUT2074

Depends on https://github.com/caskdata/cdap/pull/2996 and https://github.com/caskdata/cdap/pull/2989


chtyim:
LGTM

PR #3005, opened 2015-06-19
Refactored workers, custom actions. Custom actions now have dataset, service, stream support

dtzeng:
Issue: https://issues.cask.co/browse/CDAP-2752
Build: http://builds.cask.co/browse/CDAP-DUT2075-1

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionSpecification.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionSpecification.java
@@ -16,19 +16,15 @@
 
 package co.cask.cdap.api.workflow;
 
-import co.cask.cdap.api.builder.Creator;
-import co.cask.cdap.api.builder.DescriptionSetter;
-import co.cask.cdap.api.builder.NameSetter;
-import co.cask.cdap.api.builder.OptionsSetter;
 import co.cask.cdap.api.common.PropertyProvider;
-import co.cask.cdap.api.mapreduce.MapReduceSpecification;
-import co.cask.cdap.internal.builder.BaseBuilder;
-import co.cask.cdap.internal.builder.SimpleDescriptionSetter;
-import co.cask.cdap.internal.builder.SimpleNameSetter;
+import co.cask.cdap.api.dataset.Dataset;
 import co.cask.cdap.internal.workflow.DefaultWorkflowActionSpecification;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Specification for a {@link WorkflowAction}.
@@ -51,38 +47,75 @@ public interface WorkflowActionSpecification extends PropertyProvider {
   String getDescription();
 
   /**
-   * Builder interface for the last stage of building a {@link WorkflowActionSpecification}.
+   * @return an immutable set of {@link Dataset} name that are used by the {@link WorkflowAction}
    */
-  interface SpecificationCreator extends Creator<WorkflowActionSpecification>,
-                                         OptionsSetter<Creator<WorkflowActionSpecification>> { }
+  Set<String> getDatasets();
 
   /**
    * Builder class for building the {@link WorkflowActionSpecification}.
    */
-  final class Builder extends BaseBuilder<WorkflowActionSpecification> implements SpecificationCreator {
+  static final class Builder {
+    private String name;
+    private String description;
+    private Map<String, String> options = Maps.newHashMap();
+    private final ImmutableSet.Builder<String> datasets = ImmutableSet.builder();
 
-    private final Map<String, MapReduceSpecification> mapReduces = Maps.newHashMap();
-    private final Map<String, String> options = Maps.newHashMap();
-
-    public static NameSetter<DescriptionSetter<SpecificationCreator>> with() {
-      Builder builder = new Builder();
-      return SimpleNameSetter.create(
-        getNameSetter(builder), SimpleDescriptionSetter.create(
-        getDescriptionSetter(builder), (SpecificationCreator) builder
-      ));
+    public static NameSetter with() {
+      return new Builder().new NameSetter();
     }
 
-    @Override
-    public Creator<WorkflowActionSpecification> withOptions(Map<String, String> options) {
-      this.options.putAll(options);
-      return this;
+    public final class NameSetter {
+      /**
+       * Sets the name of a action.
+       * @param name Name of the action.
+       * @return An instance of {@link DescriptionSetter}
+       */
+      public DescriptionSetter setName(String name) {
+        Preconditions.checkArgument(name != null, "Name cannot be null.");
+        Builder.this.name = name;
+        return new DescriptionSetter();
+      }
     }
 
-    @Override
-    public WorkflowActionSpecification build() {
-      return new DefaultWorkflowActionSpecification(name, description, options);
+    /**
+     * Class defining the description setter that is used as part of the builder.
+     */
+    public final class DescriptionSetter {
+      /**
+       * Sets the description of the action.
+       * @param description Description to be associated with action.
+       * @return An instance of what needs to be done after description {@link AfterDescription}
+       */
+      public AfterDescription setDescription(String description) {
+        Preconditions.checkArgument(description != null, "Description cannot be null.");
+        Builder.this.description = description;
+        return new AfterDescription();
+      }
     }
 
+    /**
+     * Class defining the action after defining the description for a action.
+     */
+    public final class AfterDescription {
+
+      public AfterDescription withOptions(Map<String, String> options) {
+        Builder.this.options.putAll(options);
+        return this;
+      }
+
+      public AfterDescription useDataset(String dataset, String...moreDatasets) {
+        Builder.this.datasets.add(dataset).add(moreDatasets);
+        return this;
+      }
+
+      public WorkflowActionSpecification build() {
+        return new DefaultWorkflowActionSpecification(name, description, options, datasets.build());
+      }
+    }
+
+    /**
+     * Private builder to maintain builder contract.
+     */
     private Builder() {
     }
   }
--- a/cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
@@ -16,13 +16,16 @@
 package co.cask.cdap.api.workflow;
 
 import co.cask.cdap.api.Predicate;
+import co.cask.cdap.api.RuntimeContext;
+import co.cask.cdap.api.ServiceDiscoverer;
+import co.cask.cdap.api.data.DatasetContext;
 
 import java.util.Map;
 
 /**
  * Represents the runtime context of a {@link WorkflowAction}.
  */
-public interface WorkflowContext {
+public interface WorkflowContext extends RuntimeContext, ServiceDiscoverer, DatasetContext {
 
   WorkflowSpecification getWorkflowSpecification();
 
--- a/cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
+++ b/cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
@@ -18,16 +18,20 @@ package co.cask.cdap.internal.workflow;
 import co.cask.cdap.api.workflow.WorkflowAction;
 import co.cask.cdap.api.workflow.WorkflowActionSpecification;
 import co.cask.cdap.internal.lang.Reflections;
+import co.cask.cdap.internal.specification.DataSetFieldExtractor;
 import co.cask.cdap.internal.specification.PropertyFieldExtractor;
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.google.common.reflect.TypeToken;
 
 import java.util.Map;
+import java.util.Set;
 
 /**
- *
+ * The default implementation for a {@link WorkflowActionSpecification}.
  */
 public class DefaultWorkflowActionSpecification implements WorkflowActionSpecification {
 
@@ -35,30 +39,43 @@ public class DefaultWorkflowActionSpecification implements WorkflowActionSpecifi
   private final String name;
   private final String description;
   private final Map<String, String> properties;
+  private final Set<String> datasets;
 
-  public DefaultWorkflowActionSpecification(String name, String description, Map<String, String> properties) {
-    this(null, name, description, properties);
+  public DefaultWorkflowActionSpecification(String name, String description,
+                                            Map<String, String> properties, Set<String> datasets) {
+    this.className = null;
+    this.name = name;
+    this.description = description;
+    this.properties = ImmutableMap.copyOf(properties);
+    this.datasets = ImmutableSet.copyOf(datasets);
   }
 
   public DefaultWorkflowActionSpecification(WorkflowAction action) {
     WorkflowActionSpecification spec = action.configure();
 
     Map<String, String> properties = Maps.newHashMap(spec.getProperties());
+    Set<String> datasets = Sets.newHashSet();
     Reflections.visit(action, TypeToken.of(action.getClass()),
+                      new DataSetFieldExtractor(datasets),
                       new PropertyFieldExtractor(properties));
+
+    // Add datasets that are specified in overriding configure with useDataset method
+    datasets.addAll(spec.getDatasets());
 
     this.className = action.getClass().getName();
     this.name = spec.getName();
     this.description = spec.getDescription();
     this.properties = ImmutableMap.copyOf(properties);
+    this.datasets = ImmutableSet.copyOf(datasets);
   }
 
-  public DefaultWorkflowActionSpecification(String className, String name,
-                                            String description, Map<String, String> properties) {
+  public DefaultWorkflowActionSpecification(String className, String name, String description,
+                                            Map<String, String> properties, Set<String> datasets) {
     this.className = className;
     this.name = name;
     this.description = description;
     this.properties = ImmutableMap.copyOf(properties);
+    this.datasets = ImmutableSet.copyOf(datasets);
   }
 
   @Override
@@ -92,6 +109,12 @@ public class DefaultWorkflowActionSpecification implements WorkflowActionSpecifi
       .add("name", name)
       .add("class", className)
       .add("options", properties)
+      .add("datasets", datasets)
       .toString();
+  }
+
+  @Override
+  public Set<String> getDatasets() {
+    return datasets;
   }
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
@@ -15,19 +15,31 @@
  */
 package co.cask.cdap.internal.app.runtime.workflow;
 
+import co.cask.cdap.api.metrics.Metrics;
+import co.cask.cdap.api.metrics.MetricsCollectionService;
+import co.cask.cdap.api.metrics.MetricsContext;
 import co.cask.cdap.api.workflow.WorkflowActionSpecification;
 import co.cask.cdap.api.workflow.WorkflowContext;
 import co.cask.cdap.api.workflow.WorkflowSpecification;
 import co.cask.cdap.api.workflow.WorkflowToken;
+import co.cask.cdap.app.metrics.ProgramUserMetrics;
+import co.cask.cdap.app.program.Program;
+import co.cask.cdap.data2.dataset2.DatasetFramework;
+import co.cask.cdap.internal.app.runtime.AbstractContext;
+import co.cask.cdap.internal.app.runtime.BasicArguments;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+import org.apache.twill.api.RunId;
+import org.apache.twill.discovery.DiscoveryServiceClient;
 
+import java.util.HashSet;
 import java.util.Map;
 import javax.annotation.Nullable;
 
 /**
- *
+ * Default implementation of a {@link WorkflowContext}.
  */
-final class BasicWorkflowContext implements WorkflowContext {
+final class BasicWorkflowContext extends AbstractContext implements WorkflowContext {
 
   private final WorkflowSpecification workflowSpec;
   private final WorkflowActionSpecification specification;
@@ -35,16 +47,38 @@ final class BasicWorkflowContext implements WorkflowContext {
   private final ProgramWorkflowRunner programWorkflowRunner;
   private final Map<String, String> runtimeArgs;
   private final WorkflowToken token;
+  private final Metrics userMetrics;
 
-  BasicWorkflowContext(WorkflowSpecification workflowSpec, @Nullable WorkflowActionSpecification specification,
+  BasicWorkflowContext(WorkflowSpecification workflowSpec, @Nullable WorkflowActionSpecification spec,
                        long logicalStartTime, @Nullable ProgramWorkflowRunner programWorkflowRunner,
-                       Map<String, String> runtimeArgs, WorkflowToken token) {
+                       Map<String, String> runtimeArgs, WorkflowToken token, Program program, RunId runId,
+                       MetricsCollectionService metricsCollectionService,
+                       DatasetFramework datasetFramework, DiscoveryServiceClient discoveryServiceClient) {
+    super(program, runId, new BasicArguments(runtimeArgs), (spec == null) ? new HashSet<String>() : spec.getDatasets(),
+          getMetricCollector(program, runId.getId(), metricsCollectionService),
+          datasetFramework, discoveryServiceClient, null, null);
     this.workflowSpec = workflowSpec;
-    this.specification = specification;
+    this.specification = spec;
     this.logicalStartTime = logicalStartTime;
     this.programWorkflowRunner = programWorkflowRunner;
     this.runtimeArgs = ImmutableMap.copyOf(runtimeArgs);
     this.token = token;
+    if (metricsCollectionService != null) {
+      this.userMetrics = new ProgramUserMetrics(getProgramMetrics());
+    } else {
+      this.userMetrics = null;
+    }
+  }
+
+  @Nullable
+  private static MetricsContext getMetricCollector(Program program, String runId,
+                                                   @Nullable MetricsCollectionService service) {
+    if (service == null) {
+      return null;
+    }
+    Map<String, String> tags = Maps.newHashMap(getMetricsContext(program, runId));
+
+    return service.getContext(tags);
   }
 
   @Override
@@ -71,6 +105,11 @@ final class BasicWorkflowContext implements WorkflowContext {
       throw new UnsupportedOperationException("Operation not allowed.");
     }
     return programWorkflowRunner.create(name);
+  }
+
+  @Override
+  public Metrics getMetrics() {
+    return userMetrics;
   }
 
   @Override
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/ProgramWorkflowRunnerFactory.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/ProgramWorkflowRunnerFactory.java
@@ -63,7 +63,6 @@ public class ProgramWorkflowRunnerFactory {
   public ProgramWorkflowRunner getProgramWorkflowRunner(WorkflowActionSpecification actionSpec, WorkflowToken token,
                                                         String nodeId) {
 
-
     if (actionSpec.getProperties().containsKey(ProgramWorkflowAction.PROGRAM_TYPE)) {
       switch (SchedulableProgramType.valueOf(actionSpec.getProperties().get(ProgramWorkflowAction.PROGRAM_TYPE))) {
         case MAPREDUCE:
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
@@ -16,7 +16,9 @@
 package co.cask.cdap.internal.app.runtime.workflow;
 
 import co.cask.cdap.api.Predicate;
+import co.cask.cdap.api.dataset.Dataset;
 import co.cask.cdap.api.mapreduce.MapReduceSpecification;
+import co.cask.cdap.api.metrics.MetricsCollectionService;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
 import co.cask.cdap.api.spark.SparkSpecification;
 import co.cask.cdap.api.workflow.ScheduleProgramInfo;
@@ -34,17 +36,27 @@ import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.program.Program;
 import co.cask.cdap.app.runtime.Arguments;
 import co.cask.cdap.app.runtime.ProgramOptions;
+import co.cask.cdap.common.app.RunIds;
 import co.cask.cdap.common.lang.ClassLoaders;
 import co.cask.cdap.common.lang.InstantiatorFactory;
+import co.cask.cdap.common.lang.PropertyFieldSetter;
 import co.cask.cdap.common.logging.LoggingContext;
 import co.cask.cdap.common.logging.LoggingContextAccessor;
+import co.cask.cdap.data2.dataset2.DatasetFramework;
+import co.cask.cdap.internal.app.runtime.DataSetFieldSetter;
+import co.cask.cdap.internal.app.runtime.MetricsFieldSetter;
 import co.cask.cdap.internal.app.runtime.ProgramOptionConstants;
 import co.cask.cdap.internal.app.runtime.ProgramRunnerFactory;
+import co.cask.cdap.internal.lang.Reflections;
 import co.cask.cdap.internal.workflow.DefaultWorkflowActionSpecification;
 import co.cask.cdap.internal.workflow.ProgramWorkflowAction;
 import co.cask.cdap.logging.context.WorkflowLoggingContext;
 import co.cask.cdap.templates.AdapterDefinition;
 import co.cask.http.NettyHttpService;
+import co.cask.tephra.TransactionAware;
+import co.cask.tephra.TransactionContext;
+import co.cask.tephra.TransactionFailureException;
+import co.cask.tephra.TransactionSystemClient;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.base.Throwables;
@@ -56,11 +68,14 @@ import com.google.common.reflect.TypeToken;
 import com.google.common.util.concurrent.AbstractExecutionThreadService;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.gson.Gson;
+import org.apache.twill.api.RunId;
+import org.apache.twill.discovery.DiscoveryServiceClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -98,9 +113,16 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
   private boolean suspended;
   private Lock lock;
   private Condition condition;
+  private final RunId runId;
+  private final MetricsCollectionService metricsCollectionService;
+  private final DatasetFramework datasetFramework;
+  private final DiscoveryServiceClient discoveryServiceClient;
+  private final TransactionSystemClient txClient;
 
   WorkflowDriver(Program program, ProgramOptions options, InetAddress hostname,
-                 WorkflowSpecification workflowSpec, ProgramRunnerFactory programRunnerFactory) {
+                 WorkflowSpecification workflowSpec, ProgramRunnerFactory programRunnerFactory,
+                 MetricsCollectionService metricsCollectionService, DatasetFramework datasetFramework,
+                 DiscoveryServiceClient discoveryServiceClient, TransactionSystemClient txClient) {
     this.program = program;
     this.hostname = hostname;
     this.runtimeArgs = createRuntimeArgs(options.getUserArguments());
@@ -121,6 +143,11 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
     this.loggingContext = new WorkflowLoggingContext(program.getNamespaceId(), program.getApplicationId(),
                                                      program.getName(),
                                                      arguments.getOption(ProgramOptionConstants.RUN_ID), adapterName);
+    this.runId = RunIds.fromString(options.getArguments().getOption(ProgramOptionConstants.RUN_ID));
+    this.metricsCollectionService = metricsCollectionService;
+    this.datasetFramework = datasetFramework;
+    this.discoveryServiceClient = discoveryServiceClient;
+    this.txClient = txClient;
   }
 
   @Override
@@ -189,11 +216,11 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
   }
 
   private void executeAction(ApplicationSpecification appSpec, WorkflowActionNode node,
-                             InstantiatorFactory instantiator, ClassLoader classLoader,
+                             InstantiatorFactory instantiator, final ClassLoader classLoader,
                              WorkflowToken token) throws Exception {
 
     final WorkflowActionSpecification actionSpec;
-    ScheduleProgramInfo actionInfo = node.getProgram();
+    final ScheduleProgramInfo actionInfo = node.getProgram();
     switch (actionInfo.getProgramType()) {
       case MAPREDUCE:
         MapReduceSpecification mapReduceSpec = appSpec.getMapReduce().get(actionInfo.getProgramName());
@@ -218,7 +245,9 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
 
     status.put(node.getNodeId(), node);
 
-    final WorkflowAction action = initialize(actionSpec, classLoader, instantiator, token, node.getNodeId());
+    final BasicWorkflowContext workflowContext = createWorkflowContext(actionSpec, token, node.getNodeId());
+    final WorkflowAction action = initialize(actionSpec, classLoader, instantiator, workflowContext);
+
     ExecutorService executor = Executors.newSingleThreadExecutor(
       new ThreadFactoryBuilder().setNameFormat("workflow-executor-%d").build());
     try {
@@ -228,10 +257,21 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
         public void run() {
           ClassLoaders.setContextClassLoader(action.getClass().getClassLoader());
           try {
-            action.run();
+            if (actionInfo.getProgramType() == SchedulableProgramType.CUSTOM_ACTION) {
+              try {
+                runInTransaction(action, workflowContext);
+              } catch (TransactionFailureException e) {
+                throw Throwables.propagate(e);
+              }
+            } else {
+              action.run();
+            }
           } finally {
-            // Call action.destroy().
-            destroy(actionSpec, action);
+            try {
+              destroyInTransaction(action, actionSpec, workflowContext);
+            } catch (TransactionFailureException e) {
+              throw Throwables.propagate(e);
+            }
           }
         }
       });
@@ -244,6 +284,53 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
       executor.shutdownNow();
       status.remove(node.getNodeId());
     }
+  }
+
+  private TransactionContext getTransactionContextFromContext(WorkflowContext context) {
+    TransactionContext transactionContext = new TransactionContext(txClient);
+    Collection<Dataset> datasets = ((BasicWorkflowContext) context).getDatasets().values();
+    for (Dataset ds : datasets) {
+      transactionContext.addTransactionAware((TransactionAware) ds);
+    }
+    return transactionContext;
+  }
+
+  private void runInTransaction(WorkflowAction action,
+                                WorkflowContext workflowContext) throws TransactionFailureException {
+    TransactionContext transactionContext = getTransactionContextFromContext(workflowContext);
+    transactionContext.start();
+    try {
+      action.run();
+    } catch (Throwable e) {
+      LOG.error("Transaction failed to run: " + e.getMessage());
+      transactionContext.abort(new TransactionFailureException("Transaction function failure for transaction. ", e));
+    }
+    transactionContext.finish();
+  }
+
+  private void destroyInTransaction(WorkflowAction action, WorkflowActionSpecification spec,
+                                    WorkflowContext workflowContext) throws TransactionFailureException {
+    TransactionContext transactionContext = getTransactionContextFromContext(workflowContext);
+    transactionContext.start();
+    try {
+      destroy(spec, action);
+    } catch (Throwable e) {
+      LOG.error("Transaction failed to destroy: " + e.getMessage());
+      transactionContext.abort(new TransactionFailureException("Transaction function failure for transaction. ", e));
+    }
+    transactionContext.finish();
+  }
+
+  private void initializeInTransaction(WorkflowAction action, WorkflowContext workflowContext) throws Exception {
+    TransactionContext transactionContext = getTransactionContextFromContext(workflowContext);
+    transactionContext.start();
+    try {
+      action.initialize(workflowContext);
+    } catch (Throwable e) {
+      LOG.error("Transaction failed to initialize: " + e.getMessage());
+      transactionContext.abort(new TransactionFailureException("Transaction function failure for transaction. ", e));
+    }
+    transactionContext.finish();
   }
 
   private void executeFork(final ApplicationSpecification appSpec, WorkflowForkNode fork,
@@ -322,7 +409,9 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
     Predicate<WorkflowContext> predicate = instantiator.get(
       TypeToken.of((Class<? extends Predicate<WorkflowContext>>) clz)).create();
 
-    WorkflowContext context = new BasicWorkflowContext(workflowSpec, null, logicalStartTime, null, runtimeArgs, token);
+    WorkflowContext context = new BasicWorkflowContext(workflowSpec, null, logicalStartTime, null, runtimeArgs, token,
+                                                       program, runId, metricsCollectionService,
+                                                       datasetFramework, discoveryServiceClient);
     Iterator<WorkflowNode> iterator;
     if (predicate.apply(context)) {
       // execute the if branch
@@ -334,7 +423,7 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
     executeAll(iterator, appSpec, instantiator, classLoader, token);
   }
 
-    @Override
+  @Override
   protected void run() throws Exception {
     LOG.info("Start workflow execution for {}", workflowSpec);
     WorkflowToken token = new BasicWorkflowToken();
@@ -379,24 +468,33 @@ final class WorkflowDriver extends AbstractExecutionThreadService {
     return httpService.getBindAddress();
   }
 
+  private BasicWorkflowContext createWorkflowContext(WorkflowActionSpecification actionSpec,
+                                                     WorkflowToken token, String nodeId) {
+    return new BasicWorkflowContext(workflowSpec, actionSpec, logicalStartTime,
+                                    workflowProgramRunnerFactory.getProgramWorkflowRunner(actionSpec, token, nodeId),
+                                    runtimeArgs, token, program, runId, metricsCollectionService,
+                                    datasetFramework, discoveryServiceClient);
+  }
+
   /**
    * Instantiates and initialize a WorkflowAction.
    */
   @SuppressWarnings("unchecked")
   private WorkflowAction initialize(WorkflowActionSpecification actionSpec,
                                     ClassLoader classLoader, InstantiatorFactory instantiator,
-                                    WorkflowToken token, String nodeId) throws Exception {
+                                    BasicWorkflowContext workflowContext) throws Exception {
     Class<?> clz = Class.forName(actionSpec.getClassName(), true, classLoader);
     Preconditions.checkArgument(WorkflowAction.class.isAssignableFrom(clz), "%s is not a WorkflowAction.", clz);
     WorkflowAction action = instantiator.get(TypeToken.of((Class<? extends WorkflowAction>) clz)).create();
 
     ClassLoader oldClassLoader = ClassLoaders.setContextClassLoader(action.getClass().getClassLoader());
     try {
-      action.initialize(new BasicWorkflowContext(workflowSpec, actionSpec,
-                                                 logicalStartTime,
-                                                 workflowProgramRunnerFactory.getProgramWorkflowRunner(actionSpec,
-                                                                                                       token, nodeId),
-                                                 runtimeArgs, token));
+      Reflections.visit(action, TypeToken.of(action.getClass()),
+                        new PropertyFieldSetter(actionSpec.getProperties()),
+                        new DataSetFieldSetter(workflowContext),
+                        new MetricsFieldSetter(workflowContext.getMetrics()));
+
+      initializeInTransaction(action, workflowContext);
     } catch (Throwable t) {
       LOG.warn("Exception on WorkflowAction.initialize(), abort Workflow. {}", actionSpec, t);
       // this will always rethrow
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
@@ -16,6 +16,7 @@
 package co.cask.cdap.internal.app.runtime.workflow;
 
 import co.cask.cdap.api.metrics.MetricsCollectionService;
+import co.cask.cdap.api.workflow.Workflow;
 import co.cask.cdap.api.workflow.WorkflowSpecification;
 import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.program.Program;
@@ -24,36 +25,49 @@ import co.cask.cdap.app.runtime.ProgramOptions;
 import co.cask.cdap.app.runtime.ProgramRunner;
 import co.cask.cdap.common.app.RunIds;
 import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.data2.dataset2.DatasetFramework;
 import co.cask.cdap.internal.app.runtime.ProgramOptionConstants;
 import co.cask.cdap.internal.app.runtime.ProgramRunnerFactory;
 import co.cask.cdap.proto.ProgramType;
+import co.cask.tephra.TransactionSystemClient;
 import com.google.common.base.Preconditions;
+import com.google.gson.Gson;
 import com.google.inject.Inject;
 import com.google.inject.name.Named;
 import org.apache.twill.api.RunId;
 import org.apache.twill.api.ServiceAnnouncer;
+import org.apache.twill.discovery.DiscoveryServiceClient;
 
 import java.net.InetAddress;
 
 /**
- *
+ * A {@link ProgramRunner} that runs a {@link Workflow}.
  */
 public class WorkflowProgramRunner implements ProgramRunner {
+  private static final Gson GSON = new Gson();
 
   private final ProgramRunnerFactory programRunnerFactory;
   private final ServiceAnnouncer serviceAnnouncer;
   private final InetAddress hostname;
+
   private final MetricsCollectionService metricsCollectionService;
+  private final DatasetFramework datasetFramework;
+  private final DiscoveryServiceClient discoveryServiceClient;
+  private final TransactionSystemClient txClient;
 
   @Inject
   public WorkflowProgramRunner(ProgramRunnerFactory programRunnerFactory,
                                ServiceAnnouncer serviceAnnouncer,
                                @Named(Constants.AppFabric.SERVER_ADDRESS) InetAddress hostname,
-                               MetricsCollectionService metricsCollectionService) {
+                               MetricsCollectionService metricsCollectionService, DatasetFramework datasetFramework,
+                               DiscoveryServiceClient discoveryServiceClient, TransactionSystemClient txClient) {
     this.programRunnerFactory = programRunnerFactory;
     this.serviceAnnouncer = serviceAnnouncer;
     this.hostname = hostname;
     this.metricsCollectionService = metricsCollectionService;
+    this.datasetFramework = datasetFramework;
+    this.discoveryServiceClient = discoveryServiceClient;
+    this.txClient = txClient;
   }
 
   @Override
@@ -68,12 +82,15 @@ public class WorkflowProgramRunner implements ProgramRunner {
 
     WorkflowSpecification workflowSpec = appSpec.getWorkflows().get(program.getName());
     Preconditions.checkNotNull(workflowSpec, "Missing WorkflowSpecification for %s", program.getName());
-    
-    WorkflowDriver driver = new WorkflowDriver(program, options, hostname, workflowSpec, programRunnerFactory);
+
+    WorkflowDriver driver = new WorkflowDriver(program, options, hostname, workflowSpec, programRunnerFactory,
+                                               metricsCollectionService, datasetFramework,
+                                               discoveryServiceClient, txClient);
 
     // Controller needs to be created before starting the driver so that the state change of the driver
     // service can be fully captured by the controller.
     RunId runId = RunIds.fromString(options.getArguments().getOption(ProgramOptionConstants.RUN_ID));
+
     ProgramController controller = new WorkflowProgramController(program, driver, serviceAnnouncer, runId);
     driver.start();
 
--- a/cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetWithCustomActionApp.java
+++ b/cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetWithCustomActionApp.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright © 2014-2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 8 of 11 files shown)


cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
```diff
@@ -0,0 +1,42 @@
...
+
+import co.cask.cdap.api.data.stream.StreamWriter;
+import co.cask.cdap.api.templates.AdapterContext;
+
+/**
+ * Created by dtzeng on 6/17/15.
+ */
+public interface RunnableContext extends RuntimeContext, ServiceDiscoverer, StreamWriter, AdapterContext {
+
+//  /**
```
gokulavasan: Remove comments

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
```diff
@@ -0,0 +1,42 @@
...
+ * the License.
+ */
+
+package co.cask.cdap.api;
+
+import co.cask.cdap.api.data.stream.StreamWriter;
+import co.cask.cdap.api.templates.AdapterContext;
+
+/**
+ * Created by dtzeng on 6/17/15.
```
gokulavasan: We usually don't leave name/date. Replace it with meaningful javadoc.

dtzeng:
Refactoring design:
The main purpose of this is to extract the dataset/service functionality of Workers and refactor it so Custom Actions can also use this since they are both Runnable interfaces.

The context for this functionality is pulled out into a RunnableContext (implemented by BasicRunnableContext).
BasicWorkerContext will extend from BasicRunnableContext and the rest of the Worker implementation remains mostly the same.

Custom Actions are slightly different, because they are "contained" inside a workflow rather than having their own context.  So, I created a new WorkflowActionContext, which will be added to every WorkflowContext (see WorkflowProgramRunner).  If an action wants to access datasets/service, it will need to do get the action context from the workflow context: getContext().getWorkflowActionContext().{execute,write,getServiceURL}

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetUser.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.common;
+
+
```
bdmogal: extra newline

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetUser.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Set;
+
+/**
+ * Defines an interface for components which use Datasets.
+ */
+public interface DatasetUser {
+
+  /**
+   * @return An immutable set of {@link co.cask.cdap.api.dataset.Dataset DataSets} that are used
```
bdmogal: fully qualified class name?

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetUser.java
```diff
@@ -0,0 +1,31 @@
...
+
+package co.cask.cdap.api.common;
+
+
+import java.util.Set;
+
+/**
+ * Defines an interface for components which use Datasets.
+ */
+public interface DatasetUser {
```
bdmogal: we should find a better name for this class. 'User' might indicate an actual user. ``DatasetAccessor``? But looks like we already have it in ``cdap-explore``

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowAction.java
```diff
@@ -15,6 +15,7 @@
  */
 package co.cask.cdap.api.workflow;
 
+
```
bdmogal: extra newline

cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetWithCustomActionApp.java
```diff
@@ -0,0 +1,70 @@
...
+import co.cask.cdap.api.data.DatasetContext;
+import co.cask.cdap.api.dataset.lib.KeyValueTable;
+import co.cask.cdap.api.workflow.AbstractWorkflow;
+import co.cask.cdap.api.workflow.AbstractWorkflowAction;
+
+/**
+ * App which copies data from one KVTable to another using a Workflow Custom Action.
+ */
+public class DatasetWithCustomActionApp extends AbstractApplication {
+  public static final String INPUT_DS = "t1";
```
bdmogal: can these be package-private?

dtzeng:
cleaned up code

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
```diff
@@ -0,0 +1,37 @@
...
+
+package co.cask.cdap.api;
+
+import co.cask.cdap.api.data.stream.StreamWriter;
+import co.cask.cdap.api.templates.AdapterContext;
+
+/**
+ * Defines an interface for Runnable components which need access to datasets, services, streams.
+ */
+public interface RunnableContext extends RuntimeContext, ServiceDiscoverer, StreamWriter, AdapterContext {
```
bdmogal: How did you choose the name ``Runnable``? It has a history in CDAP, so wondering how :)

Either way, we may have to think of a better name.

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Execute a set of operations on datasets via a {@link TxRunnable} that are committed as a single transaction.
+   * @param runnable the runnable to be executed in the transaction
+   */
+  void execute(TxRunnable runnable);
```
bdmogal: Rename to ``txRunnable`` since this class itself talks about a different Runnable

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Defines an interface for components which use Datasets.
+ */
+public interface DatasetProvider {
+
+  /**
+   * @return An immutable set of {@link co.cask.cdap.api.dataset.Dataset DataSets} that are used
+   */
+  Set<String> getDatasets();
```
bdmogal: Is just a ``String`` sufficient? Or do we need more info? Do we need ``DatasetSpecification``, or would that be too much? Is there a similar class in API?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionContext.java
```diff
@@ -0,0 +1,31 @@
...
+ * Defines an interface for the context of Workflow Actions
+ */
+public interface WorkflowActionContext extends RunnableContext {
+
+  /**
+   * Returns the specification used to configure {@link WorkflowAction} bounded to this context.
+   */
+  WorkflowActionSpecification getSpecification();
+
+}
```
bdmogal: missing newline at the end?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionContext.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Defines an interface for the context of Workflow Actions
+ */
+public interface WorkflowActionContext extends RunnableContext {
+
+  /**
+   * Returns the specification used to configure {@link WorkflowAction} bounded to this context.
+   */
+  WorkflowActionSpecification getSpecification();
+
```
bdmogal: unnecessary emptyline

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * @return the instance id
+   */
+  int getInstanceId();
+}
```
bdmogal: missing newline at the end?

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetProvider.java
```diff
@@ -0,0 +1,30 @@
...
+}
```
bdmogal: missing newline at the end?

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
```diff
@@ -35,6 +37,8 @@
   private final String name;
   private final String description;
   private final Map<String, String> properties;
+  private final Set<String> datasets;
+
```
bdmogal: extra newline

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
gokulavasan: We should have the AdapterContext and StreamWriter interfaces only in WorkerContext.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
```diff
@@ -51,14 +55,16 @@ public DefaultWorkflowActionSpecification(WorkflowAction action) {
...
 
-  public DefaultWorkflowActionSpecification(String className, String name,
-                                            String description, Map<String, String> properties) {
+  public DefaultWorkflowActionSpecification(String className, String name, String description,
+                                            Map<String, String> properties) {
     this.className = className;
     this.name = name;
     this.description = description;
     this.properties = ImmutableMap.copyOf(properties);
+    this.datasets = new HashSet<>();
```
bdmogal: Use ``Sets.newHashSet()``

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
```diff
@@ -86,6 +92,7 @@ public String getProperty(String key) {
     return properties.get(key);
   }
 
+
```
bdmogal: extra newline

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
```diff
@@ -86,6 +92,7 @@ public String getProperty(String key) {
...
   @Override
   public String toString() {
```
bdmogal: Should we add the new property to the ``toString()``

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/BasicRunnableContext.java
```diff
@@ -0,0 +1,250 @@
...
+      .removalListener(new RemovalListener<Long, Map<DatasetCacheKey, Dataset>>() {
+        @Override
+        @ParametersAreNonnullByDefault
+        public void onRemoval(RemovalNotification<Long, Map<DatasetCacheKey, Dataset>> notification) {
+          if (notification.getValue() != null) {
+            for (Map.Entry<DatasetCacheKey, Dataset> entry : notification.getValue().entrySet()) {
+              try {
+                entry.getValue().close();
+              } catch (IOException e) {
+                LOG.error("Error closing dataset: {}", entry.getKey(), e);
```
bdmogal: Silent ignore?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
```diff
@@ -35,16 +36,25 @@
   private final ProgramWorkflowRunner programWorkflowRunner;
   private final Map<String, String> runtimeArgs;
   private final WorkflowToken token;
+  private final WorkflowActionContext context;
```
bdmogal: rename to ``workflowActionContext``

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -97,9 +97,11 @@
   private boolean suspended;
   private Lock lock;
   private Condition condition;
+  private final BasicWorkflowActionContext context;
```
bdmogal: rename to ``workflowActionContext``

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
```diff
@@ -68,15 +96,48 @@ public ProgramController run(Program program, ProgramOptions options) {
...
+
     ProgramController controller = new WorkflowProgramController(program, driver, serviceAnnouncer, runId);
     driver.start();
 
     return controller;
   }
+
+  @Nullable
+  private AdapterDefinition getAdapterSpecification(Arguments arguments) {
+    // TODO: Refactor ProgramRunner class hierarchy to have common logic moved to a common parent.
```
bdmogal: Jira?

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
dtzeng: Any suggestions for RunnableContext alternates?

Gokul, should Custom Actions be able to write to streams too?  If they can have dataset access, why not streams?

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetProvider.java
dtzeng: The original WorkerSpecification was a Set<String>.  I've also looked at other examples like FlowletDefinition, MapReduceSpecification and their getDatasets all return a Set<String> too.  Do I need to change this for Workers/Custom actions?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/BasicRunnableContext.java
dtzeng: What's this mean?

I didn't write this code, because this was directly copied over from the original worker implementation

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
dtzeng: This was here from before, not sure what it's about

dtzeng:
Made more fixes from comments

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetProvider.java
bdmogal: I think its ok to leave as is then. Didn't realize that this was existing code.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/BasicRunnableContext.java
bdmogal: I mean't this just catches ``Exception``, logs and moves forward, so was wondering if that's ok.

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * Execute a set of operations on datasets via a {@link TxRunnable} that are committed as a single transaction.
+   * @param txRunnable the runnable to be executed in the transaction
+   */
+  void execute(TxRunnable txRunnable);
+
+  /**
+   * @return the instance id
+   */
+  int getInstanceId();
```
gokulavasan: Can Workflow Action have multiple instances? If not, we do we want getInstanceId in runnableContext, shouldn't it be part of WorkerContext alone? 

If Workflow Action can have multiple instances, then getInstanceCount method should be in runnableContext and not just in WorkerContext.

sagarkapare: No WorkflowAction don't have multiple instances.

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
chtyim: Why this interface needs to extends from bunch of other unrelated interfaces? Generally it should only be the class who implements all these interfaces.

cdap-api/src/main/java/co/cask/cdap/api/RunnableContext.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Defines an interface for Runnable components which need access to datasets, services, streams.
+ */
+public interface RunnableContext extends RuntimeContext, ServiceDiscoverer, StreamWriter, AdapterContext {
+
+  /**
+   * Execute a set of operations on datasets via a {@link TxRunnable} that are committed as a single transaction.
+   * @param txRunnable the runnable to be executed in the transaction
+   */
+  void execute(TxRunnable txRunnable);
```
chtyim: Who needs this method besides Worker? For Workflow custom action, we can make the whole `run` method runs inside a transaction, which shouldn't be long running (it's unlike Worker.run(), which expected to be long running).

dtzeng:
decoupled custom actions from workers, dataset access can now use @UseDataset

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/WorkflowActionLoggingContext.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Constructs ApplicationLoggingContext.
+   *
+   * @param namespaceId   namespace id
+   * @param applicationId application id
+   * @param runId         run id of the application
+   */
+  public WorkflowActionLoggingContext(String namespaceId, String applicationId, String actionId, String runId,
+                                      String instanceId, @Nullable String adapterId) {
```
bdmogal: Shouldn't have ``adapterId``

cdap-api/pom.xml
```diff
@@ -52,6 +52,10 @@
       <artifactId>tephra-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>co.cask.tephra</groupId>
```
bdmogal: why do we need this dependency? seems like we should only be using tephra-api?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionContext.java
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * @return the instance id
+   */
+  int getInstanceId();
+
+  /**
+   * @return the datasets used by this workflow action
+   */
+  Map<String, Dataset> getDatasets();
+}
```
bdmogal: missing newline

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetProvider.java
bdmogal: please add this. also check your IDE settings. It should complain about missing newlines.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowCustomActionProgram.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public ProgramType getType() {
+    return ProgramType.CUSTOM_ACTION;
+  }
+
+  @Override
+  public ApplicationSpecification getApplicationSpecification() {
+    throw new IllegalStateException("Custom action does not need this method.");
+  }
+}
```
bdmogal: newline

cdap-unit-test/src/test/java/co/cask/cdap/test/app/DatasetWithCustomActionApp.java
```diff
@@ -0,0 +1,123 @@
...
+      @GET
+      @Path("{name}")
+      public void sayHi(HttpServiceRequest request, HttpServiceResponder responder,
+                        @PathParam("name") String name) {
+        table.write("hi", name);
+        responder.sendString(name);
+      }
+    }
+  }
+}
```
bdmogal: newline

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/WorkflowActionLoggingContext.java
```diff
@@ -0,0 +1,58 @@
...
+  public String getLogPartition() {
+    return String.format("%s:%s", super.getLogPartition(), getSystemTag(TAG_WORKFLOWACTION_ID));
+  }
+
+  @Override
+  public String getLogPathFragment(String logBaseDir) {
+    return String.format("%s/workflowaction-%s", super.getLogPathFragment(logBaseDir),
+                         getSystemTag(TAG_WORKFLOWACTION_ID));
+  }
+}
```
bdmogal: newline

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
```diff
@@ -68,15 +103,39 @@ public ProgramController run(Program program, ProgramOptions options) {
...
+    if (!arguments.hasOption(ProgramOptionConstants.ADAPTER_SPEC)) {
+      return null;
+    }
+    return GSON.fromJson(arguments.getOption(ProgramOptionConstants.ADAPTER_SPEC), AdapterDefinition.class);
+  }
+
+  @Nullable
+  private PluginInstantiator createPluginInstantiator(@Nullable AdapterDefinition adapterSpec,
```
bdmogal: I don't see this method being used. Please check and remove.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
```diff
@@ -68,15 +103,39 @@ public ProgramController run(Program program, ProgramOptions options) {
...
     WorkflowSpecification workflowSpec = appSpec.getWorkflows().get(program.getName());
     Preconditions.checkNotNull(workflowSpec, "Missing WorkflowSpecification for %s", program.getName());
-    
-    WorkflowDriver driver = new WorkflowDriver(program, options, hostname, workflowSpec, programRunnerFactory);
 
     // Controller needs to be created before starting the driver so that the state change of the driver
     // service can be fully captured by the controller.
     RunId runId = RunIds.fromString(options.getArguments().getOption(ProgramOptionConstants.RUN_ID));
+
+    AdapterDefinition adapterSpec = getAdapterSpecification(options.getArguments());
```
bdmogal: Doesn't seem to be used. Please remove. Also remove the ``getAdapterSpecification`` method if this is unused.

cdap-api/pom.xml
chtyim: shouldn't have this dependency.

cdap-api/src/main/java/co/cask/cdap/api/common/DatasetProvider.java
chtyim: Why need a separate interface?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionContext.java
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Returns the specification used to configure {@link WorkflowAction} bounded to this context.
+   */
+  WorkflowActionSpecification getSpecification();
+
+  /**
+   * Execute a set of operations on datasets via a {@link TxRunnable} that are committed as a single transaction.
+   * @param txRunnable the runnable to be executed in the transaction
+   */
+  void execute(TxRunnable txRunnable);
```
chtyim: Why have this method? Shouldn't have this. The `WorkflowAction.run()` should be called within a transaction, meaning no need for this method.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionContext.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * @return the instance id
+   */
+  int getInstanceId();
```
chtyim: What does it actually return? What is instance id of WorkflowAction?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionContext.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * @return the datasets used by this workflow action
+   */
+  Map<String, Dataset> getDatasets();
```
chtyim: What is this method for? Who need to use this?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
```diff
@@ -51,4 +52,10 @@
    * @return a {@link WorkflowToken}
    */
   WorkflowToken getToken();
+
+  /**
+   * @return the {@link WorkflowActionContext} for custom actions
+   */
+  @Nullable
+  WorkflowActionContext getWorkflowActionContext();
```
chtyim: Who call this method? Seems wrong to have this method.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
```diff
@@ -42,11 +50,16 @@
   private String programName;
   private Runnable programRunner;
   private SchedulableProgramType programType;
+  private Map<String, Dataset> datasets;
+  private TransactionSystemClient txClient;
 
-  public ProgramWorkflowAction(String name, String programName, SchedulableProgramType programType) {
+  public ProgramWorkflowAction(String name, String programName, SchedulableProgramType programType,
+                               @Nullable TransactionSystemClient txClient) {
```
chtyim: Why it is nullable?

cdap-api/pom.xml
dtzeng: when I don't have it, it says it cannot find TransactionContext, TransactionFailureException, TransactionSystemClient in ProgramWorkflowAction

chtyim: Meaning you shouldn't use those classes in the cdap-api module.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionContext.java
dtzeng: This method needs to be used in ProgramWorkflowAction to add the datasets the action uses into the transaction

chtyim: Since the execute() method is not needed, I don't think there is need for this method. Also, even we really need it, I don't think it should a public API.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
dtzeng: Datasets can be injected using @UseDataset, but for services you still need to get the context and get the service URL (like in the run method in DatasetWithCustomActionApp)

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
dtzeng: This is nullable because in WorkflowDriver's executeAction, we only need to pass in the txClient only if we are doing a custom action

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
chtyim: Only custom action have access to `WorkflowContext`, so can have the WorkflowContext extends from DatasetContext directly. No need to have extra layer of indirection.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
chtyim: ProgramWorkflowAction is purely for running another program (MR / Spark), it shouldn't be used for running custom action. @sagarkapare That should be the case, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
dtzeng: I don't need adapterSpec anymore, but I didn't write the getAdapterSpecification and createPluginInstantiator methods.  Should I still remove those methods?

cdap-api/pom.xml
dtzeng: I need those classes to wrap the run method in a transaction.  Is there a better way to do this?

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
sagarkapare: WorkflowDriver calls the `action.run` in order to execute the action. For programs (like MapReduce/Spark), it is the run method on the ProgramWorkflowAction which eventually runs the program, however for custom actions its the user defined `run` method that would get called. We wanted a wrapper on the custom action's `run` method as well, so that inside the wrapper's `run` method we can execute the custom action's `run` method within the transaction. 

For this reason, the ProgramWorkflowAction also used for running the custom actions run method.

dtzeng: We decided to make a custom action another "program", because otherwise we do not have access to the datasets in WorkflowDriver's executeAction.  The datasets are only instantianted in the initialize method when we create the WorkflowActionContext's, so only then can we pass the datasets into the ProgramWorkflowAction where they are added to the transaction.

cdap-watchdog/src/main/java/co/cask/cdap/logging/context/WorkflowActionLoggingContext.java
sagarkapare: We may not need this logging context right now, since the logs for the Workflow custom actions are collected with the WorkflowDriver's logging context only.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/CustomActionProgramWorkflowRunner.java
```diff
@@ -0,0 +1,54 @@
...
+import co.cask.cdap.api.workflow.WorkflowSpecification;
+import co.cask.cdap.api.workflow.WorkflowToken;
+import co.cask.cdap.app.program.Program;
+import co.cask.cdap.app.runtime.ProgramOptions;
+import co.cask.cdap.internal.app.runtime.ProgramRunnerFactory;
+
+/**
+ * A {@link ProgramWorkflowRunner} that creates {@link Runnable} for executing custom action from Workflow.
+ */
+final class CustomActionProgramWorkflowRunner extends AbstractProgramWorkflowRunner {
```
sagarkapare: Instead of extending AbstractProgramWorkflowRunner, this class can simply implement the ProgramWorkflowRunner and create method of the implementation can return the runnable.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/ProgramWorkflowAction.java
```diff
@@ -74,9 +87,29 @@ public void initialize(WorkflowContext context) throws Exception {
              programType != null ? programType.name() : null, programName);
   }
 
+  public void setTxClient(TransactionSystemClient txClient) {
```
sagarkapare: Do we need this method, since the txClient is already set in the constructor, right?

dtzeng: It is set in the constructor, but the problem is that in WorkflowDriver's initialize, we use reflection to get back a new instance of the class, so we need to set the txClient again explicitly

chtyim: It think we don't need TransactionSystemClient in this class at all. As I mentioned above, the workflow driver should be the one controlling the transaction.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowContext.java
dtzeng: I did it like this, because not all Workflow's will have a custom action.  In those cases, getWorkflowActionContext will just return null.  Otherwise, the WorkflowActionContext will have the datasets from WorkflowActionSpecification.getDatasets() that AbstractContext can instantiate.  If a Workflow does not have a custom action, we shouldn't need to pass in all the DatasetFramework, TransactionSystemClient, DiscoveryServiceClient etc.. right? (constructor of BasicWorkflowActionContext)

chtyim: It's better to pass in those, other than having them as null.

dtzeng:
Custom actions are not run as programs anymore, and all transactions are handled in WorkflowDriver now

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
sagarkapare: `getAdapterSpecification` and `createPluginInstantiator` methods used somewhere? if not we can remove them.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
```diff
@@ -16,44 +16,76 @@
...
+ * A {@link ProgramRunner} that runs a {@link Workflow}
  */
 public class WorkflowProgramRunner implements ProgramRunner {
+  private static final Gson GSON = new Gson();
 
   private final ProgramRunnerFactory programRunnerFactory;
   private final ServiceAnnouncer serviceAnnouncer;
   private final InetAddress hostname;
+
+  private final CConfiguration cConf;
```
sagarkapare: Not sure where the `cConf` is used. If not required remove it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -98,10 +115,20 @@
...
+  private final TransactionSystemClient txClient;
 
   WorkflowDriver(Program program, ProgramOptions options, InetAddress hostname,
-                 WorkflowSpecification workflowSpec, ProgramRunnerFactory programRunnerFactory) {
+                 WorkflowSpecification workflowSpec, ProgramRunnerFactory programRunnerFactory, RunId runId,
+                 CConfiguration cConf, MetricsCollectionService metricsCollectionService,
+                 DatasetFramework datasetFramework, DiscoveryServiceClient discoveryServiceClient,
+                 TransactionSystemClient txClient) {
     this.program = program;
+    this.programOptions = options;
```
sagarkapare: If we are not using, then no need to store `programOptions` in member variable.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -98,10 +115,20 @@
...
+  private final RunId runId;
+  private final CConfiguration cConf;
+  private final MetricsCollectionService metricsCollectionService;
+  private final DatasetFramework datasetFramework;
+  private final DiscoveryServiceClient discoveryServiceClient;
+  private final TransactionSystemClient txClient;
 
   WorkflowDriver(Program program, ProgramOptions options, InetAddress hostname,
-                 WorkflowSpecification workflowSpec, ProgramRunnerFactory programRunnerFactory) {
+                 WorkflowSpecification workflowSpec, ProgramRunnerFactory programRunnerFactory, RunId runId,
```
sagarkapare: We can get RunId from the options itself 
`RunId runId = RunIds.fromString(options.getArguments().getOption(ProgramOptionConstants.RUN_ID));`
No need to pass in the constructor.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -156,6 +189,7 @@ private void blockIfSuspended() {
 
   /**
    * Suspends the execution of the Workflow after the currently running actions complete.
+   *
```
sagarkapare: Remove the blank line.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -170,6 +204,7 @@ public void suspend() throws Exception {
 
   /**
    * Resumes the execution of the Workflow.
+   *
```
sagarkapare: Remove the blank line.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -194,21 +229,25 @@ private void executeAction(ApplicationSpecification appSpec, WorkflowActionNode
 
     final WorkflowActionSpecification actionSpec;
     ScheduleProgramInfo actionInfo = node.getProgram();
+    final boolean isCustomAction;
```
sagarkapare: No need for this variable. We can use `actionInfo.getProgramType` to determine whether the action is `CUSTOM_ACTION`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -218,7 +257,18 @@ private void executeAction(ApplicationSpecification appSpec, WorkflowActionNode
 
     status.put(node.getNodeId(), node);
 
+    // Need to pass in user implemented custom action to give ProgramWorkflowAction access to call its run method
```
sagarkapare: Seems this comment is no longer valid.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -228,16 +278,25 @@ private void executeAction(ApplicationSpecification appSpec, WorkflowActionNode
         public void run() {
           ClassLoaders.setContextClassLoader(action.getClass().getClassLoader());
           try {
+            if (isCustomAction) {
```
sagarkapare: it can be moved to other method I think, like
```java
if(customAction) {
  runInTransaction(action);
}
else { 
  action.run();
}
```
And then `runInTransaction` does the job of wrapping the custom action's run in transaction.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -228,16 +278,25 @@ private void executeAction(ApplicationSpecification appSpec, WorkflowActionNode
...
+              startTransaction(transactionContext);
+            }
             action.run();
+            if (isCustomAction) {
+              finishTransaction(transactionContext);
+            }
           } finally {
             // Call action.destroy().
+            startTransaction(transactionContext);
```
sagarkapare: similarly we can have another method `destroyInTransaction` ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -246,6 +305,33 @@ public void run() {
     }
   }
 
+  private void startTransaction(TransactionContext transactionContext) {
+    try {
```
sagarkapare: We will not need these methods if we have `runInTx` method for the custom actions.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -246,6 +305,33 @@ public void run() {
...
+      transactionContext.start();
+    } catch (TransactionFailureException e) {
+      LOG.error("Transaction start failure");
```
sagarkapare: We may want to log more information.

cdap-api/src/main/java/co/cask/cdap/api/worker/WorkerContext.java
```diff
@@ -34,9 +34,9 @@
 
   /**
    * Execute a set of operations on datasets via a {@link TxRunnable} that are committed as a single transaction.
-   * @param runnable the runnable to be executed in the transaction
+   * @param txRunnable the runnable to be executed in the transaction
```
sagarkapare: I think we can revert the changes in this file.

cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
```diff
@@ -21,7 +22,7 @@
 public abstract class AbstractWorkflowAction implements WorkflowAction {
 
   private final String name;
-  private WorkflowContext context;
+  private WorkflowContext workflowContext;
```
sagarkapare: No need to rename, since we are already in the WorkflowAction, context would be of WorkflowContext.

cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
```diff
@@ -49,8 +50,8 @@ public void destroy() {
     // No-op
   }
 
-  protected final WorkflowContext getContext() {
-    return context;
+  public WorkflowContext getContext() {
```
sagarkapare: why make this public?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionSpecification.java
```diff
@@ -16,24 +16,21 @@
...
 import com.google.common.collect.Maps;
 
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Specification for a {@link WorkflowAction}.
  */
-public interface WorkflowActionSpecification extends PropertyProvider {
+public interface WorkflowActionSpecification extends ProgramSpecification, PropertyProvider {
```
sagarkapare: Not sure why we need to extend `ProgramSpecification`. Can you please explain?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/ProgramWorkflowRunnerFactory.java
```diff
@@ -19,6 +19,7 @@
 import co.cask.cdap.api.mapreduce.MapReduce;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
 import co.cask.cdap.api.spark.Spark;
+import co.cask.cdap.api.workflow.WorkflowAction;
```
sagarkapare: Remove import if not required.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/ProgramWorkflowRunnerFactory.java
```diff
@@ -72,6 +73,7 @@ public ProgramWorkflowRunner getProgramWorkflowRunner(WorkflowActionSpecificatio
         case SPARK:
           return new SparkProgramWorkflowRunner(workflowSpec, programRunnerFactory, workflowProgram,
                                                 workflowProgramOptions, token, nodeId);
+
```
sagarkapare: Remove blank line.

cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
dtzeng: we need this public because the user needs to call getContext() in order to use getServiceURL for example

sagarkapare: but that will happen in the action that user writes, right? which would be extended from the `AbstractWorkflowAction`

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionSpecification.java
dtzeng: Not sure why I added it, as the interface already specifies the getClassName, getName and getDescription methods.  Would it be better to delete the "extend ProgramSpecification" or the 3 method declarations?

cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
dtzeng: I also need it in the executeAction of WorkflowDriver, when I cast the action back in order to get the datasets by using getContext.  That requires the method to be public

sagarkapare: But that information can be retrieved from the action specifications.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionSpecification.java
sagarkapare: I think its better to keep it as it was before 3 methods declaration. We may later add some functionality to the `ProgramSpecification` which may not be relevant to the CUSTOM actions though.

cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
dtzeng: The action specs only provide the String names of the datasets.  The initialize method will create the Datasets, so I use the action to get back the instantiated Datasets

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionSpecification.java
```diff
@@ -51,38 +48,75 @@
   String getDescription();
 
   /**
-   * Builder interface for the last stage of building a {@link WorkflowActionSpecification}.
+   * @return an immutable set of {@link Dataset} name that are used by the {@link WorkflowAction}
```
sagarkapare: Put `.` at the end of the comment.

sagarkapare: ohh its `@return`, so ignore it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
```diff
@@ -35,16 +47,46 @@
...
   private final Map<String, String> runtimeArgs;
   private final WorkflowToken token;
+  private final Metrics userMetrics;
 
   BasicWorkflowContext(WorkflowSpecification workflowSpec, @Nullable WorkflowActionSpecification specification,
                        long logicalStartTime, @Nullable ProgramWorkflowRunner programWorkflowRunner,
                        Map<String, String> runtimeArgs, WorkflowToken token) {
+    this(workflowSpec, specification, logicalStartTime, programWorkflowRunner,
+         ImmutableMap.copyOf(runtimeArgs), token, null, null, null, null, null, null);
+  }
```
sagarkapare: We may not need this constructor now. Always create `BasicWorkflowContext` with all values of the parameters.

sagarkapare: In `WorkflowDriver` when we execute the condition, we create the `BasicWorkflowContext` and pass this context to the predicate inside the condition. Since context now has `getDataset` method, I think its better to wrap the `predicate.apply(context)` inside transaction as well, so that conditions will also have access to the datasets and services.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -219,6 +248,16 @@ private void executeAction(ApplicationSpecification appSpec, WorkflowActionNode
     status.put(node.getNodeId(), node);
 
     final WorkflowAction action = initialize(actionSpec, classLoader, instantiator, token, node.getNodeId());
+
+    final TransactionContext transactionContext = new TransactionContext(txClient);
```
sagarkapare: should this be moved to `runInTransaction` method?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -246,6 +289,41 @@ public void run() {
     }
   }
 
+  private void runInTransaction(WorkflowAction action, TransactionContext transactionContext) {
```
sagarkapare: Its better to create the instance of `TransactionContext` in the method instead of passing it as a method parameter, so that we will have all transaction related code localized.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
dtzeng: Should I have every method each get all the datasets again and create a new TransactionContext?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
dtzeng: Oh just saw this, got it

sagarkapare: Also there is a code duplication to make the datasets transaction aware (before initialize and run). we can make it as a part of common method. Also the process of making datasets transaction aware need to happen for destroy as well.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
dtzeng: fixed

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
dtzeng: fixed

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
dtzeng: done

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
```diff
@@ -18,16 +18,20 @@
...
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.google.common.reflect.TypeToken;
 
 import java.util.Map;
+import java.util.Set;
 
 /**
- *
+ * The default implementation for a {@link WorkflowActionSpecification}
```
sagarkapare: `.` at the end of the comment.

dtzeng: added

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionSpecification.java
```diff
@@ -16,19 +16,16 @@
 
 package co.cask.cdap.api.workflow;
 
-import co.cask.cdap.api.builder.Creator;
-import co.cask.cdap.api.builder.DescriptionSetter;
-import co.cask.cdap.api.builder.NameSetter;
-import co.cask.cdap.api.builder.OptionsSetter;
+import co.cask.cdap.api.ProgramSpecification;
```
sagarkapare: Remove unused imports, in general from all files which are part of this PR.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
```diff
@@ -35,30 +39,43 @@
...
+  private final Set<String> datasets;
 
-  public DefaultWorkflowActionSpecification(String name, String description, Map<String, String> properties) {
-    this(null, name, description, properties);
+  public DefaultWorkflowActionSpecification(String name, String description,
+                                            Map<String, String> properties, Set<String> datasets) {
+    this.className = null;
+    this.name = name;
+    this.description = description;
+    this.properties = properties;
```
sagarkapare: Make a defensive copy of the properties.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
```diff
@@ -35,30 +39,43 @@
...
     WorkflowActionSpecification spec = action.configure();
 
     Map<String, String> properties = Maps.newHashMap(spec.getProperties());
+    Set<String> datasets = Sets.newHashSet();
     Reflections.visit(action, TypeToken.of(action.getClass()),
+                      new DataSetFieldExtractor(datasets),
                       new PropertyFieldExtractor(properties));
 
+    // Add datasets that are specified in overriding configure with useDataset method
+    datasets.addAll(spec.getDatasets());
```
sagarkapare: shouldn't the `datasets` be populated before the `Reflections.visit` ? - 
`Set<String> datasets = Sets.newHashSet(spec.getDatasets());`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
```diff
@@ -15,19 +15,31 @@
...
+import com.google.common.collect.Maps;
+import org.apache.twill.api.RunId;
+import org.apache.twill.discovery.DiscoveryServiceClient;
 
 import java.util.Map;
 import javax.annotation.Nullable;
 
 /**
- *
+ * Default implementation of a {@link WorkflowContext}
```
sagarkapare: Missing `.` at the end.

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
dtzeng: It doesn't matter which order it goes in because the Reflection adds the datasets which used \@UseDataset, and the addAll adds the datasets which were declared by overriding the configure

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionSpecification.java
dtzeng: fixed

cdap-api/src/main/java/co/cask/cdap/internal/workflow/DefaultWorkflowActionSpecification.java
dtzeng: done

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
dtzeng: added

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
```diff
@@ -68,12 +97,15 @@ public ProgramController run(Program program, ProgramOptions options) {
...
     WorkflowSpecification workflowSpec = appSpec.getWorkflows().get(program.getName());
     Preconditions.checkNotNull(workflowSpec, "Missing WorkflowSpecification for %s", program.getName());
-    
-    WorkflowDriver driver = new WorkflowDriver(program, options, hostname, workflowSpec, programRunnerFactory);
 
     // Controller needs to be created before starting the driver so that the state change of the driver
     // service can be fully captured by the controller.
     RunId runId = RunIds.fromString(options.getArguments().getOption(ProgramOptionConstants.RUN_ID));
+
+    WorkflowDriver driver = new WorkflowDriver(program, options, hostname, workflowSpec, programRunnerFactory,
```
sagarkapare: Its better to move the instantiation of the `driver` up before the `runId` as it was previously.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
```diff
@@ -16,44 +16,73 @@
...
 import org.apache.twill.api.RunId;
 import org.apache.twill.api.ServiceAnnouncer;
+import org.apache.twill.discovery.DiscoveryServiceClient;
 
 import java.net.InetAddress;
+import javax.annotation.Nullable;
 
 /**
- *
+ * A {@link ProgramRunner} that runs a {@link Workflow}
```
sagarkapare: Missing `.` at the end.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -392,11 +492,19 @@ private WorkflowAction initialize(WorkflowActionSpecification actionSpec,
...
     ClassLoader oldClassLoader = ClassLoaders.setContextClassLoader(action.getClass().getClassLoader());
     try {
-      action.initialize(new BasicWorkflowContext(workflowSpec, actionSpec,
-                                                 logicalStartTime,
-                                                 workflowProgramRunnerFactory.getProgramWorkflowRunner(actionSpec,
-                                                                                                       token, nodeId),
-                                                 runtimeArgs, token));
+      BasicWorkflowContext workflowContext =
+        new BasicWorkflowContext(workflowSpec, actionSpec,
+                                 logicalStartTime,
```
sagarkapare: This can be moved to previous line.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
dtzeng: done

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
dtzeng: done

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
dtzeng: done

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowProgramRunner.java
```diff
@@ -68,12 +82,16 @@ public ProgramController run(Program program, ProgramOptions options) {
...
+
+    WorkflowDriver driver = new WorkflowDriver(program, options, hostname, workflowSpec, programRunnerFactory,
+                                               metricsCollectionService, datasetFramework,
+                                               discoveryServiceClient, txClient);
 
     // Controller needs to be created before starting the driver so that the state change of the driver
     // service can be fully captured by the controller.
     RunId runId = RunIds.fromString(options.getArguments().getOption(ProgramOptionConstants.RUN_ID));
+
+
```
sagarkapare: extra lines.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -391,12 +495,20 @@ private WorkflowAction initialize(WorkflowActionSpecification actionSpec,
     WorkflowAction action = instantiator.get(TypeToken.of((Class<? extends WorkflowAction>) clz)).create();
 
     ClassLoader oldClassLoader = ClassLoaders.setContextClassLoader(action.getClass().getClassLoader());
+
+    BasicWorkflowContext workflowContext =
+      new BasicWorkflowContext(workflowSpec, actionSpec,
+                               logicalStartTime,
```
sagarkapare: `logicalStartTime` can go up in the same line as `workflowSpec` and `actionSpec`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -246,6 +279,75 @@ public void run() {
...
+      transactionContext.start();
+      action.initialize(workflowContext);
+      transactionContext.finish();
+    } catch (TransactionFailureException e) {
+      LOG.error("Transaction failed to initialize: " + e.getMessage());
+      throw Throwables.propagate(e);
+    }
+  }
+
+  private boolean applyPredicateInTransaction(Predicate predicate, WorkflowContext context) {
```
sagarkapare: Lets not do it in this PR. we can create another PR for fixing predicates and unit test case over there.

sagarkapare:
Couple of minor comments. Please have @chtyim also take a look at this.

dtzeng:
build green: http://builds.cask.co/browse/CDAP-DUT2075-9

cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
```diff
@@ -49,7 +50,7 @@ public void destroy() {
     // No-op
   }
 
-  protected final WorkflowContext getContext() {
+  public WorkflowContext getContext() {
```
chtyim: Why becomes public? It should never be. The Workflow driver is the one that pass in the context object to custom action, it shouldn't need to or rely on the action to return it back (since it's user code, it can return something else for you).

dtzeng: The WorkflowDriver doesn't pass in the context object, it's actually created only in the initialize method.  If I don't have it public, what's another way to get the context back? (so I can do this in getTransactionContextFromAction)

chtyim: ? Created in initialize? The `initialize()` method should be invoked from the WorkflowDriver, isn't it?

dtzeng: Fixed

cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
```diff
@@ -49,7 +50,7 @@ public void destroy() {
     // No-op
   }
 
-  protected final WorkflowContext getContext() {
+  protected WorkflowContext getContext() {
```
chtyim: This file should be unchanged. Revert all changes.

dtzeng: done

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
```diff
@@ -35,16 +47,38 @@
...
     this.token = token;
+    if (metricsCollectionService != null) {
+      this.userMetrics = new ProgramUserMetrics(getProgramMetrics());
+    } else {
+      this.userMetrics = null;
+    }
+  }
+
+  @Nullable
+  private static MetricsContext getMetricCollector(Program program, String runId,
```
chtyim: Better have this method returns `Metrics` so that you don't need to check for null for `MetricsCollectionService` in both constructor and this method (you just need to check here).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -246,6 +278,52 @@ public void run() {
...
+      transactionContext.addTransactionAware((TransactionAware) ds);
+    }
+    return transactionContext;
+  }
+
+  private void runInTransaction(WorkflowAction action, WorkflowContext workflowContext) {
+    TransactionContext transactionContext = getTransactionContextFromContext(workflowContext);
+    try {
+      transactionContext.start();
+      action.run();
```
chtyim: You need to abort the transaction context if the user method call throws exception. See how https://github.com/caskdata/tephra/blob/develop/tephra-core/src/main/java/co/cask/tephra/DefaultTransactionExecutor.java#L131 is implemented.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowContext.java
dtzeng: This method is used for the super to receive a MetricsContext, not for assigning userMetrics?

chtyim: Oh.. Collector. Nevermind, yet I blame the parent class. Anyways, it needs refactoring, but not in this PR.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
dtzeng: Does this fix work?

chtyim: Also, you can reduce the code duplication by have a method to execute transactionally.

```<V> V executeInTransaction(WorkflowContext workflowContext, Callable<V> callable) throws TransactionFailureException```

chtyim:
LGTM

PR #3010, opened 2015-06-21
Fix/ui storing token in cookie

ajainarayanan:
- Adds the ability to refresh UI when security is enabled, without the necessity to login repeatedly. 
- Fixes profile view to view auth-token when security is enabled.

Secure Cluster: http://securecluster1762-1000.dev.continuuity.net:9999/

authentication1
authentication


elfenheart:
:+1: LGTM

PR #3011, opened 2015-06-21
Removes live-reload + avoid logging request and response in secure mode

ajainarayanan:
- Removes live-reload script from index.html 
- Avoids logging-in request and response in debugger when in secure mode.


elfenheart:
:+1: LGTM

PR #3012, opened 2015-06-22
Fix broken documentation build

awholegunch:
Update MD5 hashes and update include of a file.

Documentation build was broken as of [build #205](https://builds.cask.co/browse/CDAP-DDBD-205/).

Clean Built at: 

- [CDAP Docs](http://docs-staging.cask.co/cdap/3.1.0-SNAPSHOT-feature-3.1_fix_examples_manual/en/index.html)
- [Log](http://builds.cask.co/browse/CDAP-DDBD385-1/log)

### DIFF
--- a/cdap-docs/examples-manual/build.sh
+++ b/cdap-docs/examples-manual/build.sh
@@ -148,8 +148,8 @@ function download_includes() {
   test_an_include 97e355542831d767a297b52ff09f5674 ../../cdap-examples/SparkPageRank/src/main/java/co/cask/cdap/examples/sparkpagerank/SparkPageRankApp.java
 
   test_an_include afe12d26b79607a846d3eaa58958ea5f ../../cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
-  test_an_include bbeed2893195fa12553e4eb28016306a ../../cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/UploadService.java
-  test_an_include 1089a092c1f0af1ecfedae646c9c6b99 ../../cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/ScoreCounter.java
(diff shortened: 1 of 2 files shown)


cdap-docs/examples-manual/build.sh
```diff
@@ -148,8 +148,8 @@ function download_includes() {
   test_an_include 97e355542831d767a297b52ff09f5674 ../../cdap-examples/SparkPageRank/src/main/java/co/cask/cdap/examples/sparkpagerank/SparkPageRankApp.java
 
   test_an_include afe12d26b79607a846d3eaa58958ea5f ../../cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/SportResults.java
-  test_an_include bbeed2893195fa12553e4eb28016306a ../../cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/UploadService.java
-  test_an_include 1089a092c1f0af1ecfedae646c9c6b99 ../../cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/ScoreCounter.java
+  test_an_include 2d85727db18c3261b60d4cb278846329 ../../cdap-examples/SportResults/src/main/java/co/cask/cdap/examples/sportresults/UploadService.java
```
sreevatsanraman: This doesn't seem to be a very scalable model. Each time we change anything in the examples we have to update the docs build. Created a JIRA to fix in 3.1.0: https://issues.cask.co/browse/CDAP-2834

sreevatsanraman:
While this change looks good, we should find a better solution. The docs will break each time we change an example. Let's find a proper fix. OK to merge this in the meantime.

cdap-docs/examples-manual/build.sh
wolf31o2: It's not scalable. However, with the way we're including parts of the code in the examples, it requires manual intervention if the code in question changes, as the description may no longer be accurate to the code being shown.

PR #2999, opened 2015-06-19
Feature/cdap 2733 add TPFS as a source for adapters

rohanahata:
https://issues.cask.co/browse/CDAP-2733
http://builds.cask.co/browse/CDAP-DUT2071
This pull request aims to add TimePartitionedFileSets as an Source for Adapters.

PR #3013, opened 2015-06-22
CDAP-2823 aims to introduce nested records conversion between Avro and Structured Types.

rohanahata:
There are tests and functions to support the conversions. Also, this is linked with CDAP-2733 #2999. 
https://issues.cask.co/browse/CDAP-2823


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      structuredSchema = Schema.parseJson(genericRecordSchema.toString());
+      schemaCache.put(hashCode, structuredSchema);
+    }
+
+    StructuredRecord.Builder builder = StructuredRecord.builder(structuredSchema);
+    for (Schema.Field field: structuredSchema.getFields()) {
+      String fieldName = field.getName();
+      Object object = genericRecord.get(fieldName);
+      if (fieldName.equals("record")) {
```
albertshau: you should be using the schema type, not the field name, to determine if a field is a record or not

albertshau: in addition, you need to do the recursive calls all the way through instead of only for records. Consider the case where the field is a list of records for example.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/StructuredToAvroTransformer.java
```diff
@@ -47,7 +51,112 @@ public GenericRecord transform(StructuredRecord structuredRecord) throws Excepti
...
+      case FLOAT:
+      case DOUBLE:
+      case BOOLEAN:
+        return field;
+      default:
+        throw new UnexpectedFormatException("field type " + fieldType + " is not supported.");
+    }
+  }
+
+  private Object getRecordField(Object record, String fieldName) {
```
albertshau: there is no need for this, a record is always a StructuredRecord

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/StructuredToAvroTransformer.java
```diff
@@ -47,7 +51,112 @@ public GenericRecord transform(StructuredRecord structuredRecord) throws Excepti
...
+        return convertRecord(field, fieldSchema);
+      case ARRAY:
+        return convertArray(field, fieldSchema.getElementType());
+      case MAP:
+        return convertMap((Map<String, Object>) field, fieldSchema.getValueType());
+      case UNION:
+        return convertUnion(field, fieldSchema.getTypes());
+      case NULL:
+        return null;
+      case STRING:
```
albertshau: for string, we should return field.toString() since avro doesn't always use a String to represent a String type

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+        StructuredRecord innerRecord = transform((GenericRecord) object);
+        builder.set(fieldName, innerRecord);
+      }
+      else {
```
chtyim: Wrong formatting. Should be:

```java
} else {
```

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/StructuredToAvroTransformer.java
```diff
@@ -47,7 +51,112 @@ public GenericRecord transform(StructuredRecord structuredRecord) throws Excepti
...
+        }
+      }
+    }
+    if (isNullable) {
+      return null;
+    }
+    throw new UnexpectedFormatException("unable to determine union type.");
+  }
+
+  private GenericRecord convertRecord(Object record, Schema avroSchema) {
```
albertshau: In this class, a field with type record will be a StructuredRecord

albertshau:
please also add tests for avro -> structured

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/StructuredToAvroTransformer.java
```diff
@@ -47,7 +51,112 @@ public GenericRecord transform(StructuredRecord structuredRecord) throws Excepti
...
+  private List<Object> convertArray(Object values, Schema elementSchema) {
+    List<Object> output;
+    if (values instanceof List) {
+      List<Object> valuesList = (List<Object>) values;
+      output = Lists.newArrayListWithCapacity(valuesList.size());
+      for (Object value : valuesList) {
+        output.add(convertField(value, elementSchema));
+      }
+    } else {
+      Object[] valuesArr = (Object[]) values;
```
chtyim: Simplify the if-else logic by wrapping the array into a List.

```java
List<Object> list;
if (values.getClass().isArray()) {
  list = Arrays.asList((Object[]) values);
} else if (values instance of List) {
  list = (List<Object>) values;
}

File: cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/AvroToStructured.java [hunk 7]
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ *
+ */
+public class AvroToStructured extends Converter<GenericRecord, StructuredRecord> {
+
+  private final Map<Integer, co.cask.cdap.api.data.schema.Schema> schemaCache = Maps.newHashMap();
+
+  @Override
+  public StructuredRecord convert(GenericRecord genericRecord) throws IOException {
```
albertshau: we don't need both this and convertRecord, there should just be one method

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,32 @@
...
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import org.apache.avro.generic.GenericRecord;
+
+import java.io.IOException;
+
+/**
+ * Creates StructuredRecords from GenericRecords.
+ */
+public class AvroToStructuredTransformer {
```
albertshau: this and AvroToStructured are redundant. Don't need this class

rohanahata:
StructuredtoAvroTest contains a test for AvroToStructured... Please let me know if the tests are sufficient or if they should be in a different file/with a changed file name

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/Converter.java
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Structured to Avro Conversions and vice versa
+ * @param <INPUT>
+ * @param <OUTPUT>
+ */
+public abstract class Converter<INPUT, OUTPUT> {
+
+  abstract OUTPUT transform(INPUT record) throws IOException;
+
+  protected Object getRecordField(Object record, String fieldName) {
```
albertshau: this should not be needed

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/Converter.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.avro.generic.GenericRecord;
+
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Structured to Avro Conversions and vice versa
+ * @param <INPUT>
```
albertshau: should have some description in the javadoc

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/Converter.java
```diff
@@ -0,0 +1,131 @@
...
+ * @param <OUTPUT>
+ */
+public abstract class Converter<INPUT, OUTPUT> {
```
albertshau: RecordConverter may be a better name

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/Converter.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Structured to Avro Conversions and vice versa
```
albertshau: the way it is written, it doesn't have to use StructuredRecord or GenericRecord. A more accurate description would be something like,  "converts an object with a schema into another type of object with the same schema. For example, child implementations could convert a StructuredRecord to a GenericRecord and vice versa"

albertshau:
a few comments, also please add some test cases for avro -> structured

albertshau:
o didn't see your earlier comment. It's better to put the test cases in separate classes since they are testing different classes

albertshau:
lgtm :+1:

PR #3015, opened 2015-06-22
Record creationTime of partitions

anwar6953:
https://issues.cask.co/browse/CDAP-2746
http://builds.cask.co/browse/CDAP-DUT2080-5

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
@@ -71,6 +71,7 @@ public class PartitionedFileSetDataset extends AbstractDataset implements Partit
   protected static final byte[] RELATIVE_PATH = { 'p' };
   protected static final byte[] FIELD_PREFIX = { 'f', '.' };
   protected static final byte[] METADATA_PREFIX = { 'm', '.' };
+  protected static final byte[] CREATION_TIME = { 'c' };
 
   protected final FileSet files;
   protected final Table partitionsTable;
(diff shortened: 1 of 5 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionDetail.java
```diff
@@ -25,4 +25,9 @@
    * Gets the metadata for the partition
    */
   PartitionMetadata getMetadata();
+
+  /**
+   * Gets the creation time of the partition, in milliseconds
+   */
+  long getCreationTime();
```
anew: Shouldn't it be part of the PartitionMetadata?

anwar6953: Updated; moved creationTime into `PartitionMetadata`

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -71,6 +71,8 @@
   protected static final byte[] RELATIVE_PATH = { 'p' };
   protected static final byte[] FIELD_PREFIX = { 'f', '.' };
   protected static final byte[] METADATA_PREFIX = { 'm', '.' };
+  // also used as the rowkey prefix of the creation time index
+  protected static final byte[] CREATION_TIME = { 'c', '.' };
```
anew: why the trailing dot?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -315,6 +322,7 @@ protected void getPartitions(@Nullable PartitionFilter filter, PartitionConsumer
         }
         byte[] pathBytes = row.get(RELATIVE_PATH);
         if (pathBytes != null) {
+          byte[] creationTime = row.get(CREATION_TIME);
```
anew: this does not seem to be used...

anwar6953: right, the time is now being read in `metadataFromRow` method. Removed this unnecessary

anew:
Only two comments from me. Otherwise LGTM.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: I was originally having another table, as the index on the creation time, but removed that from this PR.
So, removed the trailing dot, as its not necessary.

anwar6953:
Updated.

anew:
I assume the last commit contains all the previous ones and no further changes except the two small changes suggested in the last round. LGTM

PR #3018, opened 2015-06-22
Table#get is guaranteed not to return null.

anwar6953:
The javadoc on the `Row get(byte[] row)` method of `Table` class says that it never returns `null`.
I have checked the two implementations of this method to verify that, and then make use of this fact to update the calls to the method.

http://builds.cask.co/browse/CDAP-DUT2082-2 (GREEN)

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
@@ -201,14 +201,12 @@ public class IndexedTable extends AbstractDataset implements Table {
 
     // first read the existing indexed values to find which have changed and need to be updated
     Row existingRow = table.get(dataRow, colsToIndex.toArray(new byte[colsToIndex.size()][]));
-    if (existingRow != null) {
-      for (Map.Entry<byte[], byte[]> entry : existingRow.getColumns().entrySet()) {
-        if (!Arrays.equals(entry.getValue(), putColumns.get(entry.getKey()))) {
-          index.delete(createIndexKey(dataRow, entry.getKey(), entry.getValue()), IDX_COL);
-        } else {
-          // value already indexed
-          colsToIndex.remove(entry.getKey());
-        }
+    for (Map.Entry<byte[], byte[]> entry : existingRow.getColumns().entrySet()) {
+      if (!Arrays.equals(entry.getValue(), putColumns.get(entry.getKey()))) {
+        index.delete(createIndexKey(dataRow, entry.getKey(), entry.getValue()), IDX_COL);
+      } else {
+        // value already indexed
+        colsToIndex.remove(entry.getKey());
       }
     }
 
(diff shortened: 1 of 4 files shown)


anew:
LGTM. Thanks for taking action! 

One thing I am wondering about is whether this API is actually intuitive. Since several people in our own team (including myself) used it incorrectly... it might be more intuitive to return null? That is a different discussion though.

anwar6953:
I know that it should be implied that if `Nullable` is omitted, then that means the method doesn't return null.
However, the benefit of specifying `@Nonnull` is that then the IDE can produce warnings whenever someone uses `Table#get`, and does a check against null. So it might be good practice to utilize the `@Nonnull` annotation.

anew:
@chtyim, I would love to hear your opinion.

chtyim:
I like the convention "If not annotated with `@Nullable`, it must not be `null`" since it is generally yields cleaner and easier to understand code. For IDE, there is setting to use this convention to give correct warning without flooding the code base with `@Nonnull`.

anew:
And still you could argue that that is not the default setting in most IDEs. I agree that we don't want to flood the code with unneeded annotations. But we could do it only for the api packages. Those are public APIs used by non-platform developers, and I don't see a harm in making it explicit there.

chtyim:
I think in general good API should try to avoid returning `null` (with exception in `get` method in `Map`). Also, even just doing it only for the api module, it means all methods and all parameters need to be annotated? Maybe we can figure out if that's the default settings in common IDEs or not.

A side note: (http://help.eclipse.org/juno/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Ftasks%2Ftask-using_null_annotations.htm)

anew:
It is interesting that you mention the get method in a map. Isn't the intention of Table (and other datasets) to give a programmer an experience similar to Java collections? For example, KeyValueTable return null if the key does not exist. So does ObjectMappedTable.

chtyim:
Another point of reference is the guava library, which doesn't use `@Nonnull` at all, but using `@Nullable` only. One interesting thing I noticed from the guava library is it uses `@ParametersAreNonnullByDefault` in the `package-info.java` file

chtyim:
About the `Table` API, I guess it was following the HBase convention, which returns an empty `Result`. I don't have strong preference toward one over the other, although the empty `Row` may gives slightly simpler client code in certain manipulation (e.g. just loop over columns to do something).

anew:
It is a tricky question. Intuitively I would say, a Table is a map from row keys to rows, and hence the get should return null if the row key does not exist. But because we can also specify a set of columns in the get, it might be that the row exists but does not have any of the columns. How would a caller distinguish these two cases? Or perhaps it should not matter, because HBase is the same? Then again, not everything is right just because HBase does it that way. 

The initial reason why I brought this up is that we have quite a few places in our code where we did not use it correctly. Including one occurrence in my own code - even though I knew about the semantics. I was wondering whether it is more intuitive to return null. But changing an existing API is not a good idea. Adding an annotation, however, does not break any existing code.

anwar6953:
@chtyim @anew Any resolution on this topic?
I will probably merge this, since the topic (though related) is not affecting the PR.
Build is green.

anwar6953:
Merged PR. @chtyim @anew Any resolution on your guys' discussion?

anew:
I will open a Jira to track it.

PR #3021, opened 2015-06-22
CDAP-2844: MapReduce metrics collection to not use counters, report directly to Kafka

abaranau:
Tested in singlenode and in distributed.

Note: explicit snappy dependency is needed to include latest version to avoid errors at writing to Kafka from mr tasks:
```
2015-06-22 18:23:39,133 ERROR [metrics-collection] kafka.producer.async.DefaultEventHandler: Failed to send messages
java.lang.UnsatisfiedLinkError: org.xerial.snappy.SnappyNative.maxCompressedLength(I)I
        at org.xerial.snappy.SnappyNative.maxCompressedLength(Native Method)
        at org.xerial.snappy.Snappy.maxCompressedLength(Snappy.java:320)
        at org.xerial.snappy.SnappyOutputStream.<init>(SnappyOutputStream.java:79)
        at org.xerial.snappy.SnappyOutputStream.<init>(SnappyOutputStream.java:66)
        at kafka.message.CompressionFactory$.apply(Unknown Source)
        at kafka.message.ByteBufferMessageSet$.kafka$message$ByteBufferMessageSet$$create(Unknown Source)

```


abaranau:
http://builds.cask.co/browse/CDAP-DUT2086-1

anew:
Code looks good. 

One side effect of this is that it removes the CDAP metrics from the MapReduce counters. In workflow, we allow conditions based on the mapReduce counters of a previous action. Until now those counters would have included CDAP system and user metrics. Users will now have to emit the metrics of interest explicitly as counters. 

Do you think that should be documented? Or perhaps we should still emit the metrics as counters, in addition to sending them directly to Kafka?

abaranau:
> Or perhaps we should still emit the metrics as counters, in addition to sending them directly to Kafka?

please take a look at the JIRA issue.

> In workflow, we allow conditions based on the mapReduce counters of a previous action.

if that is what we say now and is documented - there's nothing to change in docs. We don't say we are using counters to collect metrics, moreover we should never allow users (devs) to rely on it and their format, as it may change overtime without notice.

abaranau:
> Code looks good.

is that LGTM? Assuming other Qs are resolved.

anew:
Why do you think I have not read the Jira? Perhaps I should make the comment in the Jira instead. It's just that there was no way to discuss this on Jira because the Jira and PR were created at the same time.

abaranau:
replied in JIRA

> Code looks good.

is that LGTM then?

anew:
LGTM if the build passes.

PR #3023, opened 2015-06-22
CDAP-2822 IndexedTable should support scans on the indexed field.

anwar6953:
https://issues.cask.co/browse/CDAP-2822
http://builds.cask.co/browse/CDAP-DUT2089-8

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
@@ -34,6 +34,8 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.primitives.Longs;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Type;
 import java.util.Arrays;
@@ -63,8 +65,9 @@ import javax.annotation.Nullable;
  * </p>
  *
  * <p>Index entries are created by storing additional rows in a second table.  These index rows are keyed by
- * column name, column value, and original row key, each field separated by a single null byte delimiter.  Values
- * containing the null byte are allowed, but may be read and filtered out when reading from the index.
+ * column name, column value, and original row key, each field separated by a single null byte delimiter.  It is not
+ * recommended that the name of an index column contains the null byte (this can cause a degradation in performance
+ * of index reads).
  * </p>
  *
  * <p>The columns to index can be configured in the {@link co.cask.cdap.api.dataset.DatasetProperties} used
@@ -92,18 +95,21 @@ import javax.annotation.Nullable;
  * @see co.cask.cdap.api.dataset.lib.IndexedTableDefinition#INDEX_COLUMNS_CONF_KEY
  */
 public class IndexedTable extends AbstractDataset implements Table {
+  private static final Logger LOG = LoggerFactory.getLogger(IndexedTable.class);
 
   /**
    * Column key used to store the existence of a row in the secondary index.
    */
   private static final byte[] IDX_COL = {'r'};
+  private static final byte DELIMITER_BYTE = 0;
+  private static final byte[] KEY_DELIMITER = new byte[] { DELIMITER_BYTE };
 
+  private final boolean hasColumnWithDelimiter;
   // the two underlying tables
   private Table table, index;
   // the secondary index column
   private SortedSet<byte[]> indexedColumns;
 
-  private byte[] keyDelimiter = new byte[] {0};
 
   /**
    * Configuration time constructor.
@@ -118,7 +124,34 @@ public class IndexedTable extends AbstractDataset implements Table {
     this.table = table;
     this.index = index;
     this.indexedColumns = Sets.newTreeSet(Bytes.BYTES_COMPARATOR);
+    this.hasColumnWithDelimiter = hasDelimiterByte(columnsToIndex);
     Collections.addAll(this.indexedColumns, columnsToIndex);
+  }
+
+  /**
+   * Checks if a set of columns contains the DELIMITER_BYTE.
+   * This is needed because only when a column has a null byte in it do we need to check for false positive in the scan
+   * on the index by checking against the column's value in the data table.
+   * For instance, a false positive can arise with an index on the following two rows:
+   * <ul>
+   *  <li>Row #1: column name: a\0b, column value: c</li>
+   *  <li>Row #2: column name: a, column value: b\0c</li>
+   * </ul>
+   * Because of how the rows are constructed in the index table, both will be prefixed (and therefore looked up by)
+   * the row key prefix a\0b\0c.
+   *
+   * @param columns the set of columns being checked
+   * @return true if any of the columns contain the DELIMITER_BYTE
+   */
+  private boolean hasDelimiterByte(byte[][] columns) {
+    for (byte[] column : columns) {
+      for (byte b : column) {
+        if (b == DELIMITER_BYTE) {
+          return true;
+        }
+      }
+    }
+    return false;
   }
 
   /**
@@ -167,10 +200,10 @@ public class IndexedTable extends AbstractDataset implements Table {
    */
   public Scanner readByIndex(byte[] column, byte[] value) {
     assertIndexedColumn(column);
-    byte[] rowKeyPrefix = Bytes.concat(column, keyDelimiter, value, keyDelimiter);
+    byte[] rowKeyPrefix = Bytes.concat(column, KEY_DELIMITER, value, KEY_DELIMITER);
     byte[] stopRow = Bytes.stopKeyForPrefix(rowKeyPrefix);
     Scanner indexScan = index.scan(rowKeyPrefix, stopRow);
-    return new IndexScanner(indexScan, rowKeyPrefix);
+    return new IndexScanner(indexScan, column, value);
   }
 
   /**
@@ -183,17 +216,19 @@ public class IndexedTable extends AbstractDataset implements Table {
    * @param endValue the exclusive end of the range for which rows must fall within to be returned in the scan
    *                 {@code null} means end with the last row of the table
    * @return a Scanner returning rows from the data table, whose stored value for the given column is within the the
-   * given range.
+   *         given range.
    * @throws java.lang.IllegalArgumentException if the given column is not configured for indexing.
    */
   public Scanner scanByIndex(byte[] column, @Nullable byte[] startValue, @Nullable byte[] endValue) {
     assertIndexedColumn(column);
-    // keyDelimiter is not used at the end of the rowKeys, because they are used for a range scan,
+    // KEY_DELIMITER is not used at the end of the rowKeys, because they are used for a range scan,
     // instead of a fixed-match lookup
-    byte[] startRow = startValue == null ? null : Bytes.concat(column, keyDelimiter, startValue);
-    byte[] stopRow = endValue == null ? null : Bytes.concat(column, keyDelimiter, endValue);
+    byte[] startRow = startValue == null ? Bytes.concat(column, KEY_DELIMITER) :
+      Bytes.concat(column, KEY_DELIMITER, startValue);
+    byte[] stopRow = endValue == null ? Bytes.stopKeyForPrefix(Bytes.concat(column, KEY_DELIMITER)) :
+      Bytes.concat(column, KEY_DELIMITER, endValue);
     Scanner indexScan = index.scan(startRow, stopRow);
-    return new IndexScanner(indexScan, null);
+    return new IndexRangeScanner(indexScan, column, startValue, endValue);
   }
 
   private void assertIndexedColumn(byte[] column) {
@@ -243,7 +278,7 @@ public class IndexedTable extends AbstractDataset implements Table {
   }
 
   private byte[] createIndexKey(byte[] row, byte[] column, byte[] value) {
-    return Bytes.concat(column, keyDelimiter, value, keyDelimiter, row);
+    return Bytes.concat(column, KEY_DELIMITER, value, KEY_DELIMITER, row);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -180,6 +174,33 @@ public Scanner readByIndex(byte[] column, byte[] value) {
...
+   * range, then a {@link co.cask.cdap.api.dataset.table.Scanner} with no results will be returned.
+   *
+   * @param column the column to use for the index lookup
+   * @param startValue the inclusive start of the range for which rows must fall within to be returned in the scan
+   * @param endValue the exclusive end of the range for which rows must fall within to be returned in the scan
+   * @return a Scanner returning rows from the data table, whose stored value for the given column is within the the
+   * given range.
+   * @throws java.lang.IllegalArgumentException if the given column is not configured for indexing.
+   */
+  public Scanner scanByIndex(byte[] column, byte[] startValue, byte[] endValue) {
```
ghelmling: When doing range scans like this, you will often want to pass a null for start value or end value, indicating "all values from the beginning" and "all values from this start to the end".

I think supporting nullable values for these two parameters would make for a much nicer API.

anwar6953: Right; updated.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -180,6 +174,35 @@ public Scanner readByIndex(byte[] column, byte[] value) {
...
 
   /**
+   * Reads table rows within the given secondary index key range. If no rows are indexed, falling within the given
+   * range, then a {@link co.cask.cdap.api.dataset.table.Scanner} with no results will be returned.
+   *
+   * @param column the column to use for the index lookup
+   * @param startValue the inclusive start of the range for which rows must fall within to be returned in the scan.
+   *                   {@code null} means start from first row of the table
+   * @param endValue the exclusive end of the range for which rows must fall within to be returned in the scan
+   *                 {@code null} means start from first row of the table
```
chtyim: Wrong javadoc. `null` means scan till then end of the last row in the table.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -532,22 +557,20 @@ public IndexScanner(Scanner baseScanner, byte[] rowKeyPrefix) {
...
-          return null;
-        }
+      Row indexRow;
+      while ((indexRow = baseScanner.next()) != null) {
         byte[] rowkey = indexRow.get(IDX_COL);
         // verify that datarow matches the expected row key to avoid issues with column name or value
         // containing the delimiter used
-        if (rowkey != null && Bytes.equals(indexRow.getRow(), Bytes.add(rowKeyPrefix, rowkey))) {
-          dataRow = table.get(rowkey);
+        if (rowkey != null) {
```
anew: this is incorrect. Suppose I have two indexed fields: a and a\0b. A row x has a=b\0c, and a row y has a\0b=c. When scanning on column a with the range b..c, I should only see row x. But I will see both rows. 

Actually this was incorrect in the existing IndexedTable. I think after fetching the actual row, we have to double-check that it is a correct result.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -180,6 +174,35 @@ public Scanner readByIndex(byte[] column, byte[] value) {
...
+   *                 {@code null} means end with the last row of the table
+   * @return a Scanner returning rows from the data table, whose stored value for the given column is within the the
+   * given range.
+   * @throws java.lang.IllegalArgumentException if the given column is not configured for indexing.
+   */
+  public Scanner scanByIndex(byte[] column, @Nullable byte[] startValue, @Nullable byte[] endValue) {
+    assertIndexedColumn(column);
+    // keyDelimiter is not used at the end of the rowKeys, because they are used for a range scan,
+    // instead of a fixed-match lookup
+    byte[] startRow = startValue == null ? null : Bytes.concat(column, keyDelimiter, startValue);
```
ghelmling: When start value is null, we don't want to scan from the start of the table, we want to scan from the start of the entries for this column.  So start row should be column + keyDelimiter.

As long as we don't make keyDelimiter configurable, end row can be column + (keyDelimiter +1).  Theoretically we could use filters to do a prefix match on the row keys, but the Table API does not support filters.

anwar6953: Ok, right. Also, for the end row, it seems like `stopKeyForPrefix(column + keyDelimiter)` is more suitable than `column + keyDelimiter + 1`, because we would want to include all rows that have the `column + keyDelimiter` as the prefix, right?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -120,9 +119,21 @@ public IndexedTable(String name, Table table, Table index, byte[][] columnsToInd
...
     this.index = index;
     this.indexedColumns = Sets.newTreeSet(Bytes.BYTES_COMPARATOR);
-    for (byte[] col : columnsToIndex) {
-      this.indexedColumns.add(col);
+    this.hasColumnWithZero = hasZeros(columnsToIndex);
+    Collections.addAll(this.indexedColumns, columnsToIndex);
+  }
+
+  private boolean hasZeros(byte[][] columns) {
+    for (byte[] column : columns) {
```
anew: It would be cleaner to check whether a column contains the keyDelimiter. I am not sure whether can be changed or configured, but since it is not a static final, I assume that it is not fixed as 0. Perhaps it also makes sense to make it a static final?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -120,9 +119,21 @@ public IndexedTable(String name, Table table, Table index, byte[][] columnsToInd
...
+      for (byte b : column) {
+        if (b == 0) {
+          // only when a column has a zero value in it do we need to check for false positive in the scan on the index
+          // by checking against the column's value in the data table
```
anew: this would better be a Javadoc for the method.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -518,36 +558,49 @@ public void write(byte[] bytes, Put put) {
...
-      Row dataRow = null;
       // keep going until we hit a non-null, non-empty data row, or we exhaust the index
-      while (dataRow == null) {
-        Row indexRow = baseScanner.next();
-        if (indexRow == null) {
-          // end of index
-          return null;
-        }
+      Row indexRow;
+      while ((indexRow = baseScanner.next()) != null) {
```
anew: I know it's convenient, but we try to avoid assignments inside conditions. Code style :)

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -518,36 +558,49 @@ public void write(byte[] bytes, Put put) {
...
         byte[] rowkey = indexRow.get(IDX_COL);
-        // verify that datarow matches the expected row key to avoid issues with column name or value
-        // containing the delimiter used
-        if (rowkey != null && Bytes.equals(indexRow.getRow(), Bytes.add(rowKeyPrefix, rowkey))) {
-          dataRow = table.get(rowkey);
+        if (rowkey == null) {
+          continue;
```
anew: what does it mean if the rowkey is null? Can that ever happen? I feel like that indicates data corruption. Perhaps issue a warning in this case?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -518,36 +558,49 @@ public void write(byte[] bytes, Put put) {
...
+        }
+        byte[] columnValue = Arrays.copyOfRange(indexRow.getRow(),
+                                                column.length + 1,
+                                                indexRow.getRow().length - rowkey.length - 1);
+        // Verify that datarow matches the expected row key to avoid issues with column name or value
+        // containing the delimiter used. This is a sufficient check, as long as columns don't have zero value in it.
+        if (matches(columnValue)) {
+          Row row = table.get(rowkey);
+          // If a column has zero (the key delimiter) in it, then we need to check against the data row's column
+          // to be sure this row isn't a rowkey isn't a false positive in the scan.
```
anew: remove "isn't a rowkey"

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: added a warning.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: ok, I tried putting it as a for-loop, though it does seem stranger.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -555,4 +608,37 @@ public void close() {
...
+    private final byte[] start;
+    private final byte[] end;
+
+    public IndexRangeScanner(Scanner baseScanner, byte[] column, byte[] start, byte[] end) {
+      super(baseScanner, column);
+      this.start = start;
+      this.end = end;
+    }
+
+    @Override
```
anew: I am wondering whether this check is always true. Since the scan is from <column>0<start> to <column>0<end>, and the extracted column value is always a prefix of what follows <column>0, can it ever happen that it is not in the range?

cdap-data-fabric/src/test/java/co/cask/cdap/api/dataset/lib/IndexedTableTest.java
```diff
@@ -264,6 +269,132 @@ public void apply() throws Exception {
...
+          } finally {
+            scanner.close();
+          }
+        }
+      });
+    } finally {
+      dsFrameworkUtil.deleteInstance(delimTabInstance);
+    }
+  }
+
```
anew: can you add a test case for filtering false positives in a scan?

anwar6953: updated.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: I didn't quite understand your concern here. Can you explain it a bit more thoroughly?
I don't think the condition is always true, though.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anew: I think those two are the same as long as keyDelimiter is not 0xFF

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -120,9 +123,27 @@ public IndexedTable(String name, Table table, Table index, byte[][] columnsToInd
...
-    for (byte[] col : columnsToIndex) {
-      this.indexedColumns.add(col);
+    this.hasColumnWithZero = hasZeros(columnsToIndex);
+    Collections.addAll(this.indexedColumns, columnsToIndex);
+  }
+
+  /**
+   * Checks if a set of columns contains the DELIMITER_BYTE.
+   * This is needed because only when a column has a zero value in it do we need to check for false positive in the scan
+   * on the index by checking against the column's value in the data table
```
anew: It would also be useful to explain why ]0 in an indexColumn can cause false positives. For example by giving an example of a=b\0c looks the same as a\0b=c. It is not obvious from the code that this would be a false positive...

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -59,8 +60,8 @@
  * </p>
  *
  * <p>The indexed values need not be unique.  When reading the data back by index value, a {@link Scanner} will be
- * returned, allowing the client to iterate through all matching rows. Only exact matches on indexed values
- * are currently supported.
+ * returned, allowing the client to iterate through all matching rows. Exact matches as well as range lookups on
+ * indexed values are supported.
```
anew: Can you add here that is not recommended to have \0 characters in an index column?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -518,36 +568,53 @@ public void write(byte[] bytes, Put put) {
...
-          return null;
-        }
+      for (Row indexRow = baseScanner.next(); indexRow != null; indexRow = baseScanner.next()) {
         byte[] rowkey = indexRow.get(IDX_COL);
-        // verify that datarow matches the expected row key to avoid issues with column name or value
-        // containing the delimiter used
-        if (rowkey != null && Bytes.equals(indexRow.getRow(), Bytes.add(rowKeyPrefix, rowkey))) {
-          dataRow = table.get(rowkey);
+        if (rowkey == null) {
+          LOG.warn("Row of Indexed table is missing index column. Row key: {}", indexRow.getRow());
```
anew: Can you add the name of the dataset to this message?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: added in the next paragraph, which discusses the row key construction.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anew: Thinking a little more about it, there are very rare conditions where the condition is false. Only two cases I can think of is value that is a prefix of the start key, or a value of which the end key is a prefix. So I guess both checks have to stay.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -120,9 +124,33 @@ public IndexedTable(String name, Table table, Table index, byte[][] columnsToInd
...
+  /**
+   * Checks if a set of columns contains the DELIMITER_BYTE.
+   * This is needed because only when a column has a zero value in it do we need to check for false positive in the scan
+   * on the index by checking against the column's value in the data table.
+   * For instance, a false positive can arise with an index on the following two rows:
+   *        Row #1: column name: a\0b, column value: c
+   *        Row #2: column name: a, column value: b\0c
+   * Because of how the rows are constructed in the index table, both will be prefixed (and therefore looked up by)
+   * the row key prefix a\0b\0c.
+   *
```
anew: remove the extra line?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -120,9 +124,33 @@ public IndexedTable(String name, Table table, Table index, byte[][] columnsToInd
...
-      this.indexedColumns.add(col);
+    this.hasColumnWithZero = hasZeros(columnsToIndex);
+    Collections.addAll(this.indexedColumns, columnsToIndex);
+  }
+
+  /**
+   * Checks if a set of columns contains the DELIMITER_BYTE.
+   * This is needed because only when a column has a zero value in it do we need to check for false positive in the scan
+   * on the index by checking against the column's value in the data table.
+   * For instance, a false positive can arise with an index on the following two rows:
```
anew: Not sure whether this formats well in a Javadoc...

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -59,13 +60,14 @@
...
  * <p>The indexed values need not be unique.  When reading the data back by index value, a {@link Scanner} will be
- * returned, allowing the client to iterate through all matching rows. Only exact matches on indexed values
- * are currently supported.
+ * returned, allowing the client to iterate through all matching rows. Exact matches as well as range lookups on
+ * indexed values are supported.
  * </p>
  *
  * <p>Index entries are created by storing additional rows in a second table.  These index rows are keyed by
  * column name, column value, and original row key, each field separated by a single null byte delimiter.  Values
- * containing the null byte are allowed, but may be read and filtered out when reading from the index.
```
anew: Thinking about this more I think the wording is not that clear. Perhaps this should just say: It is not recommended that the name of an index column contains the null byte (this can cause a degradation in performance of index reads).

anew:
I have only a few minor comments about the javadoc comments. Everything else looks good. Please fix these and merge.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: since it's a private method, it doesn't get shown in the javadocs

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: done.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: and, I updated the format to look better (using html tags) just in case.

anwar6953:
@anew Updated.

anew:
One last comment... Please remove that sentence and merge without another review. 
Nice work overall!

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: Hmm, I found 22 occurrences in our code base where we do something like:
```
while ((indexRow = baseScanner.next()) != null) {
```
I also think it looks cleaner/concise compared to the for loop that is in this PR now.

PR #3026, opened 2015-06-23
CDAP-2665 adding an ArtifactStore that can be used to persist

albertshau:
artifacts and their metadata. The store keeps the artifacts
on hdfs and metadata about the artifacts in a table. Artifact
metadata consists of information about the plugins contained
in the artifact. In the future, it will also contain info
about application classes in the artifact. The table also stores
a mapping from plugin to artifact to allow lookups by plugin
in addition to lookups by artifact id.

Not yet used anywhere, but the next step will be to use it as
the backing for plugins and plugin jars.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
@@ -17,6 +17,7 @@
 package co.cask.cdap.internal.app.runtime.artifact;
 
 import co.cask.cdap.api.templates.plugins.PluginClass;
+import com.google.common.collect.ImmutableList;
 
 import java.util.List;
 import java.util.Objects;
@@ -28,7 +29,7 @@ public class ArtifactMeta {
   private final List<PluginClass> plugins;
 
   public ArtifactMeta(List<PluginClass> plugins) {
-    this.plugins = plugins;
+    this.plugins = ImmutableList.copyOf(plugins);
   }
 
   public List<PluginClass> getPlugins() {
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.internal.app.runtime.artifact;
+
+import co.cask.cdap.api.templates.plugins.PluginClass;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Metadata about an artifact, such as what plugins are contained in the artifact.
+ */
+public class ArtifactMeta {
+  private final List<PluginClass> plugins;
+
+  public ArtifactMeta(List<PluginClass> plugins) {
+    this.plugins = plugins;
+  }
+
+  public List<PluginClass> getPlugins() {
+    return plugins;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    ArtifactMeta that = (ArtifactMeta) o;
+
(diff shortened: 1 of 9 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2090

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Objects;
+
+/**
+ * Metadata about an artifact, such as what plugins are contained in the artifact.
+ */
+public class ArtifactMeta {
+  private final List<PluginClass> plugins;
+
+  public ArtifactMeta(List<PluginClass> plugins) {
+    this.plugins = plugins;
```
gokulavasan: No need to create a copy of the array?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -0,0 +1,515 @@
...
+ * for each entity contained in the artifact.
+ *
+ * Artifact metadata is stored with r:{namespace}:{artifact-name} as the rowkey, {artifact-version} as the column,
+ * and ArtifactMeta as the value.
+ *
+ * PluginClass metadata is stored with p:{namespace}:{plugin-type}:{plugin-name} as the rowkey,
+ * {artifact-version}:{artifact-name} as the column, and PluginClass as the value
+ *
+ * TODO: (CDAP-2764) add this part when we have a better idea of what needs to be in AppClass.
+ * AppClass metadata is stored with a:{namespace}:{app-name} as the rowkey,
```
gokulavasan: Do we need this for applications separately?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -0,0 +1,515 @@
...
+
+  /**
+   * Clear all data in the given namespace. Not terribly efficient, can do more deletes than needed.
+   * This is purely for unit tests, so it's not a big concern.
+   *
+   * @param namespace the namespace to delete data in
+   * @throws IOException if there was some problem deleting the data
+   */
+  @VisibleForTesting
+  void clear(Id.Namespace namespace) throws IOException {
```
gokulavasan: Does this need to be invoked when a delete namespace request is received?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
gokulavasan: Oh since AppClass is not the same as PluginClass?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: thats a good point, it needs to be called.  I'll figure out how to expose it later when it is actually in use.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: yeah they are different because we need to be able to look up plugins on their own.  We may not need to be able to look up applications on their own just yet though, since currently an artifact contains a single application.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
albertshau: hm yeah i'll make an immutable copy of it

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStoreTest.java
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  @Test(expected = ArtifactAlreadyExistsException.class)
+  public void testImmutability() throws Exception {
+    Id.Artifact artifactId = Id.Artifact.from(Constants.DEFAULT_NAMESPACE_ID, "myplugins", "1.0.0");
+    ArtifactMeta artifactMeta = new ArtifactMeta(ImmutableList.<PluginClass>of());
+
+    String artifactContents = "abc123";
+    // this should work
+    artifactStore.write(artifactId, artifactMeta, new ByteArrayInputStream(Bytes.toBytes(artifactContents)));
```
gokulavasan: How do we know that this code didn't throw the ArtifactAlreadyExistsException?

gokulavasan:
Few comments. Otherwise LGTM :+1:

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStoreTest.java
albertshau: that's a good point. I'll wrap in a try-catch it for this first call and fail the test if it catches the exception.

albertshau:
@gokulavasan addressed comments, please take another look

gokulavasan:
LGTM :+1:

PR #3027, opened 2015-06-23
Feature/ui ns gear smaller devices

astewes:
In this PR:
* Namespace and Settings dropdown menus now visible across all device widths
* "one" class added to namespace dropdown via ng-class, allowing for a tooltip and special styling when there is only one namespace to display
* Improved organization of header styles

header


ajainarayanan:
:+1: LGTM

PR #3028, opened 2015-06-23
Use SDK examples to fill Maven cache, not develop

wolf31o2:
This uses the files from /opt/cdap/sdk/examples versus fetching develop from git.


duddun:
LGTM

PR #3030, opened 2015-06-23
[CDAP-2849] - FileBatchSource reads in twice

jayfjin:
JIRA -  https://issues.cask.co/browse/CDAP-2849
Build - http://builds.cask.co/browse/CDAP-DUT2093-1

FileBatchSource would read in twice if the mapreduce took longer than the duration of the workflow, since the last time read wouldn't be updated until onFinish(). Thus, we now update it in prepareRun() and revert it in onFinish() if the mapreduce failed.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
@@ -94,6 +94,7 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
   private final FileBatchConfig config;
   private KeyValueTable table;
   private Date prevMinute;
+  private String previousTimeRead;
 
   public FileBatchSource(FileBatchConfig config) {
     this.config = config;
@@ -132,10 +133,12 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
 
     if (config.timeTable != null) {
       table = context.getDataset(config.timeTable);
-      String lastTimeRead = Bytes.toString(table.read("lastTimeRead"));
+      String lastTimeRead = Bytes.toString(table.read(LAST_TIME_READ));
       if (lastTimeRead == null) {
         lastTimeRead = "0";
       }
+      previousTimeRead = lastTimeRead;
+      table.write(LAST_TIME_READ, DATE_FORMAT.format(prevMinute));
       conf.set(LAST_TIME_READ, lastTimeRead);
     }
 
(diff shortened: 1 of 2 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -136,6 +137,8 @@ public void prepareRun(BatchSourceContext context) throws Exception {
       if (lastTimeRead == null) {
         lastTimeRead = "0";
       }
+      previousTimeRead = lastTimeRead;
+      table.write("lastTimeRead", DATE_FORMAT.format(prevMinute));
```
bdmogal: pull out ``lastTimeRead`` as a constant

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -171,30 +174,30 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
    */
   public static class FileBatchConfig extends PluginConfig {
 
-    @Name("fileSystem")
+    @Name("Filesystem")
```
bdmogal: I think it is ok to keep these as camel case. To change how they appear in the UI, add a templates file in ``cdap-ui/templates/ETLBatch`` and specify the ``Label`` field appropriately. Sync up with @ajainarayanan or @elfenheart on how to add/use those files.

bdmogal:
LGTM. Trust that the template JSON works.

Also, please rebase with develop, and reword the commit message to include the jira number.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -161,8 +164,8 @@ public void transform(KeyValue<LongWritable, Object> input, Emitter<StructuredRe
 
   @Override
   public void onRunFinish(boolean succeeded, BatchSourceContext context) {
-    if (succeeded && table != null && USE_TIMEFILTER.equals(config.fileRegex)) {
-      table.write("lastTimeRead", DATE_FORMAT.format(prevMinute));
+    if (!succeeded && table != null && USE_TIMEFILTER.equals(config.fileRegex)) {
```
anwar6953: so this is a rollback?

jayfjin: yes

PR #3033, opened 2015-06-23
Feature/ui preference data modelling

elfenheart:
- Refactor Preferences to use Data Modelling.
- Adding ability to hit enter to add new preference value

preferences

Cluster: http://singlenodecluster21830-1000.dev.continuuity.net:9999/ns/default

There are 4 levels of preferences:
1. System
2. Namespace
3. Application
4. Program

1-3 are located under management
4 is located in program view (drop down next to start button)


ajainarayanan:
Few minor comments about variable naming (that I talked to @elfenheart offline) and if possible a cluster to demo the change could be nicer. Apart from that :+1: LGTM

ajainarayanan:
:+1: LGTM

PR #3036, opened 2015-06-23
Build script for creating GitHub pages index file.

awholegunch:


build.sh
```diff
@@ -0,0 +1,75 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+  
+# Builds README.rst file for GitHub pages
+
+SCRIPT=`basename ${0}`
+SCRIPT_PATH=`pwd`
```
wolf31o2: This will only ever give you the correct SCRIPT_PATH if you're executing the script as `./build.sh`

build.sh
```diff
@@ -0,0 +1,75 @@
...
+RETURN=$(printf '\n')
+CDAP="cdap"
```
wolf31o2: This is used exactly once. Is it really necessary?

build.sh
```diff
@@ -0,0 +1,75 @@
...
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+  
+# Builds README.rst file for GitHub pages
+
+SCRIPT=`basename ${0}`
```
wolf31o2: This is never used.

build.sh
```diff
@@ -0,0 +1,75 @@
...
+SCRIPT_PATH=`pwd`
+RETURN=$(printf '\n')
+CDAP="cdap"
+CDAP_PATH="${SCRIPT_PATH}/${CDAP}"
+README_FILE="README.rst"
```
wolf31o2: Another variable that's only used once, also in the variables section. It's a bit unnecessary.

build.sh
```diff
@@ -0,0 +1,75 @@
...
+=================
+CDAP GitHub Pages
+=================
+
+Cask Data Application Platform (CDAP) Documentation
+EOF`
+
+READ_ME_EARLIER=`cat <<EOF
+Earlier versions:
+EOF`
```
wolf31o2: Why this complex syntax for a single line? It gets echo'd by itself, anyway.

build.sh
```diff
@@ -0,0 +1,75 @@
...
+
+function get_current_version() {
+  for release in $(ls -d */ | sort --reverse)
+    do
+      if [[ "${release}" =~ ^[0-9].* ]]; then
+        CURRENT="${release%%/}"
+        return
+      fi
+    done
+}
```
wolf31o2: This function requires execution from a certain directory, yet does no checking that it's in the correct place. I would make this agnostic to its location by putting the full path in the `ls` and then stipping it back off before sorting.

build.sh
```diff
@@ -0,0 +1,75 @@
...
+  echo "- \`Version ${rel} <http://docs.cdap.io/cdap/${rel}>\`__" >> ${README_FILE_PATH}
+}
+
+
+echo "${READ_ME_BASE}" > ${README_FILE_PATH}
+
+cd ${CDAP_PATH}
+
+get_current_version
+echo "${RETURN}" >> ${README_FILE_PATH}
```
wolf31o2: Doesn't
```
echo >> ${README_FILE_PATH}
```
accomplish the same thing?

build.sh
```diff
@@ -0,0 +1,75 @@
...
+echo "\`Latest version: ${CURRENT} <http://docs.cdap.io/cdap/current>\`__" >> ${README_FILE_PATH}
+echo "${RETURN}" >> ${README_FILE_PATH}
+write_version ${CURRENT}
+echo "${RETURN}" >> ${README_FILE_PATH}
+
+echo "${READ_ME_EARLIER}" >> ${README_FILE_PATH}
+echo "${RETURN}" >> ${README_FILE_PATH}
+
+for release in $(ls -d */ | sort --reverse)
```
wolf31o2: Same comment as previous `ls` invocation.

build.sh
awholegunch: Removed.

build.sh
awholegunch: Removed.

build.sh
awholegunch: Removed.

build.sh
awholegunch: Because it had other lines around it, since removed...removed.

build.sh
awholegunch: Fixed.

build.sh
awholegunch: fixed.

build.sh
awholegunch: fixed.

build.sh
awholegunch: Fixed.

awholegunch:
Fixed as per comments. Please re-review!

build.sh
```diff
@@ -0,0 +1,71 @@
...
+      if [[ "${rel}" =~ ^[0-9].* ]]; then
+        CURRENT="${rel%%/}"
+        return
+      fi
+    done
+}
+
+function write_version() {
+  local rel="${1%%/}"
+  echo "- \`Version ${rel} <http://docs.cdap.io/cdap/${rel}>\`__" >> ${README_FILE_PATH}
```
wolf31o2: I would prefer if we left the redirect outside this function. If this function just output the line, and redirect when done when calling, I think it would be a bit "safer" for any potential future edits to ensure the file remains the way you want.

build.sh
```diff
@@ -0,0 +1,71 @@
...
+function get_current_version() {
+  for release in $(ls -d ${CDAP_PATH}/*/ | sort --reverse )
+    do
+      rel="$(basename $release)"
+      if [[ "${rel}" =~ ^[0-9].* ]]; then
+        CURRENT="${rel%%/}"
+        return
+      fi
+    done
+}
```
wolf31o2: You should return non-zero outside the `for` loop, so this function triggers a failure if it's unable to determine the version.

build.sh
```diff
@@ -0,0 +1,71 @@
...
+  local rel="${1%%/}"
+  echo "- \`Version ${rel} <http://docs.cdap.io/cdap/${rel}>\`__" >> ${README_FILE_PATH}
+}
+
+
+echo "${READ_ME_BASE}" > ${README_FILE_PATH}
+
+cd ${CDAP_PATH}
+
+get_current_version
```
wolf31o2: After this function returns non-zero on error, you should `|| echo "Failed to get version" && exit 1`

build.sh
```diff
@@ -0,0 +1,71 @@
...
+
+echo >> ${README_FILE_PATH}
+echo "\`Latest version: ${CURRENT} <http://docs.cdap.io/cdap/current>\`__" >> ${README_FILE_PATH}
+echo >> ${README_FILE_PATH}
+write_version ${CURRENT}
```
wolf31o2: Assuming this is made to only output to stdout, `>> ${README_FILE_PATH}` after this and others below.

build.sh
```diff
@@ -0,0 +1,71 @@
...
+echo >> ${README_FILE_PATH}
+
+for release in $(ls -d ${CDAP_PATH}/*/ | sort --reverse )
+do
+  rel="$(basename $release)"
+  rel="${rel%%/}"
+  if [[ "$rel" != "${CURRENT}" && "$rel" != "current" ]]; then
+    write_version ${rel}
+  fi
+done
```
wolf31o2: Explicitly `exit 0` at the end, for CI.

build.sh
awholegunch: Done.

build.sh
awholegunch: Added.

build.sh
awholegunch: Done.

build.sh
awholegunch: Added.

build.sh
awholegunch: Fixed.

awholegunch:
More fixes per comments. Please take another look.

wolf31o2:
:+1: LGTM

PR #3037, opened 2015-06-24
CDAP-2855 Fail on errors in scripts for Docker/VM

wolf31o2:
This error checks cases where we could potentially fail to produce the desired output. Specifically, this will fail the build if the Maven cache cannot be filled.

### DIFF
--- a/cdap-distributions/src/packer/scripts/cookbook-setup.sh
+++ b/cdap-distributions/src/packer/scripts/cookbook-setup.sh
@@ -20,8 +20,10 @@
 
 # Grab cookbooks using knife
 for cb in hadoop idea maven nodejs cdap ; do
-  knife cookbook site install $cb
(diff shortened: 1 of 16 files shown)


duddun:
:+1:

cdap-distributions/src/packer/scripts/cookbook-setup.sh
```diff
@@ -20,8 +20,10 @@
 
 # Grab cookbooks using knife
 for cb in hadoop idea maven nodejs cdap ; do
-  knife cookbook site install $cb
+  knife cookbook site install $cb || echo "Cannot fetch cookbook $cb" && exit 1
```
duddun: How did I miss these? ;-)
precedence order for both bitwise and logical tests is always NOT, AND, OR

PR #3038, opened 2015-06-24
Feature/cdap 2470 fix

ataylor4:
This fix will check the validity of the table name for a DBSink and the validity of a connection for DBSource. It also removes the requirement that the user input a table name for DBSource. We chose to implement DBSource in this way because multiple table names could be involved in the query and forcing the user to enter all of them felt overly restrictive (since they have to enter the table names anyway in the query).

http://builds.cask.co/browse/CDAP-DT98-4


gokulavasan:
@ataylor4 Is https://issues.cask.co/browse/CDAP-2735 also addressed in this PR?

ataylor4:
Yes it is.

On Jun 24, 2015, at 12:00 PM, Gokul <notifications@github.com> wrote:

@ataylor4 <https://github.com/ataylor4> Is
https://issues.cask.co/browse/CDAP-2735 also addressed in this PR?

—
Reply to this email directly or view it on GitHub
<https://github.com/caskdata/cdap/pull/3038#issuecomment-114980718>.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -268,6 +267,81 @@ public void testDBSource() throws Exception {
...
+      Id.Adapter adapterIdBadConn = Id.Adapter.from(NAMESPACE, "dbSourceTest");
+      AdapterManager managerBadConn = createAdapter(adapterIdBadConn, adapterConfigBadConn);
+
+      managerBadConn.start();
+      managerBadConn.waitForOneRunToFinish(5, TimeUnit.MINUTES);
+      managerBadConn.stop();
+    } catch (Exception e) {
+      badConn = true;
+    }
+    Assert.assertTrue(badConn);
```
chtyim: Move this to the end of the try block as `Assert.fail();` with a no-op in the catch block and get rid of the `badConn` boolean.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -268,6 +267,81 @@ public void testDBSource() throws Exception {
...
+      AdapterManager managerBadSinkName = createAdapter(adapterIdBadSinkName, adapterConfigBadSinkName);
+
+      createInputData();
+      managerBadSinkName.start();
+      managerBadSinkName.waitForOneRunToFinish(5, TimeUnit.MINUTES);
+      managerBadSinkName.stop();
+    } catch (Exception e) {
+      badSinkTableNameFailed = true;
+    }
+    Assert.assertTrue(badSinkTableNameFailed);
```
chtyim: Same as above.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -176,17 +178,47 @@ private void ensureJDBCDriverIsAvailable() throws Exception {
       LOG.debug("Plugin Type: {} and Plugin Name: {}; Driver Class: {} not found. Registering JDBC driver via shim {} ",
                 dbSinkConfig.jdbcPluginType, dbSinkConfig.jdbcPluginName, driverClass.getName(),
                 JDBCDriverShim.class.getName());
-      driverShim = new JDBCDriverShim(driverClass.newInstance());
+      if (driverShim == null) {
+        driverShim = new JDBCDriverShim(driverClass.newInstance());
```
chtyim: `driverClass` is null if this method is called from `tableNameExists()` (which in turn call from `configurePipeline()`, which doesn't set any value to the `drvierClass` field).

I prefer not to have `driverClass` as a field, but rather just a parameter passed to this method.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -176,17 +178,47 @@ private void ensureJDBCDriverIsAvailable() throws Exception {
...
       DriverManager.registerDriver(driverShim);
     }
   }
 
+  private boolean tableNameExists() {
+    try {
+      try {
+        ensureJDBCDriverIsAvailable();
+      } catch (Exception e) {
+        Preconditions.checkArgument(false, "Driver must be able to connect");
```
chtyim: Why not just propagate the exception? You can use `throw Throwables.propagate(e)`.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -176,17 +178,47 @@ private void ensureJDBCDriverIsAvailable() throws Exception {
...
-      driverShim = new JDBCDriverShim(driverClass.newInstance());
+      if (driverShim == null) {
+        driverShim = new JDBCDriverShim(driverClass.newInstance());
+      }
       DBUtils.deregisterAllDrivers(driverClass);
       DriverManager.registerDriver(driverShim);
     }
   }
 
+  private boolean tableNameExists() {
```
bdmogal: This should be ``tableExists``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -176,17 +178,47 @@ private void ensureJDBCDriverIsAvailable() throws Exception {
...
+        connection = DriverManager.getConnection(dbSinkConfig.connectionString);
+      } else {
+        connection = DriverManager.getConnection(dbSinkConfig.connectionString,
+                                                 dbSinkConfig.user, dbSinkConfig.password);
+      }
+
+      DatabaseMetaData metadata = connection.getMetaData();
+      ResultSet rs = metadata.getTables(null, null, dbSinkConfig.tableName, null);
+
+      return rs.next();
```
chtyim: You need to close the `ResultSet` and the `Connection`, otherwise there will be leakage of file descriptors. Do something like this:

```java
try (
  Connection connection = createConnection(...);
  ResultSet rs = connection.getMetaData().getTables(.....);
) {
  return rs.next();
}
```

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -176,17 +178,47 @@ private void ensureJDBCDriverIsAvailable() throws Exception {
...
+    } catch (java.sql.SQLException e) {
+      return false;
```
chtyim: Why return `false`? I think it's better to throw exception so that more appropriate error/exception can be thrown from the configurePipeline method to provide more context about the error.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -176,17 +178,47 @@ private void ensureJDBCDriverIsAvailable() throws Exception {
...
+  }
+
   /**
    * {@link PluginConfig} for {@link DBSink}
    */
   public static class DBSinkConfig extends DBConfig {
     @Description(COLUMNS_DESCRIPTION)
     String columns;
+
+    @Description("Table name to export to.")
```
chtyim: Make the string as a constant to make it consistent with all other plugins.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -76,18 +80,19 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
                                   "authentication. If not, please remove dbUser and retry.");
     String jdbcPluginId = String.format("%s.%s.%s", "source", dbSourceConfig.jdbcPluginType,
                                         dbSourceConfig.jdbcPluginName);
-    Class<? extends Driver> jdbcDriverClass = pipelineConfigurer.usePluginClass(dbSourceConfig.jdbcPluginType,
+    driverClass = pipelineConfigurer.usePluginClass(dbSourceConfig.jdbcPluginType,
```
chtyim: Better for the `ensureValidConnection()` method takes a driver class as parameter. Also, you need to call `DBUtils.cleanup(driverClass);` with a finally block. The destroy method will not be called during configure phase.

chtyim: The same cleanup needs to be done for the DBSink above.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -124,6 +129,47 @@ private String getJDBCPluginId() {
...
+  private boolean ensureValidConnection() {
+    try {
+      try {
+        ensureJDBCDriverIsAvailable();
+      } catch (Exception e) {
+        Preconditions.checkArgument(false, "Driver must be able to connect");
+      }
+
+      if (dbSourceConfig.user == null) {
+        DriverManager.getConnection(dbSourceConfig.connectionString);
```
chtyim: Need to free up resources as well. See comments in DBSink above.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -124,6 +129,47 @@ private String getJDBCPluginId() {
     return String.format("%s.%s.%s", "source", dbSourceConfig.jdbcPluginType, dbSourceConfig.jdbcPluginName);
   }
 
+  private boolean ensureValidConnection() {
```
chtyim: Better have this method throw exception than returning a boolean.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -268,6 +267,81 @@ public void testDBSource() throws Exception {
     Assert.assertEquals("user2", Bytes.toString(row2.get("BLOB_COL"), 0, 5));
     Assert.assertEquals(clobData, Bytes.toString(row1.get("CLOB_COL"), 0, clobData.length()));
     Assert.assertEquals(clobData, Bytes.toString(row2.get("CLOB_COL"), 0, clobData.length()));
+
```
bdmogal: remove extra newline

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -124,6 +129,47 @@ private String getJDBCPluginId() {
...
+      }
+      return true;
+    } catch (SQLException e) {
+      LOG.error("SQL Exception thrown when trying to connect to driver", e);
+      return false;
+    }
+  }
+
+  /*
+   *Throws InstantiationException or IllegalAccessException if it can't get access to the jdbcDriverClass
```
bdmogal: nit: space after the ``*``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -168,25 +170,54 @@ private void setResultSetMetadata() throws Exception {
...
+      ensureJDBCDriverIsAvailable(jdbcDriverClass);
+    } catch (Exception e) {
+      throw Throwables.propagate(e);
+    }
+    try (Connection connection = createConnection();
+         ResultSet rs = connection.getMetaData().getTables(null, null, dbSinkConfig.tableName, null)
+    ) {
+      return rs.next();
+    } catch (java.sql.SQLException e) {
+      LOG.error("SQL Exception thrown when trying to connect to driver", e);
```
chtyim: Shouldn't log if you are re-throwing the exception.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -168,25 +170,54 @@ private void setResultSetMetadata() throws Exception {
...
+      throw Throwables.propagate(e);
```
chtyim: If you propagate, it get wrapped as RuntimeException, which will result in 500 response to the REST call. Is that what you want?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -91,6 +92,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
                                                                                 getJDBCPluginId(),
                                                                                 PluginProperties.builder().build());
     Preconditions.checkArgument(jdbcDriverClass != null, "JDBC Driver class must be found.");
+    Preconditions.checkArgument(tableExists(jdbcDriverClass), "Invalid table name %s", dbSinkConfig.tableName);
```
chtyim: I believe you need to call `DBUtils.deregisterAllDrivers` as well after calling the `tableExists` in here.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -168,25 +170,53 @@ private void setResultSetMetadata() throws Exception {
...
+      throw new IllegalArgumentException("JDBCDriver is not available", e);
+    }
+    try (Connection connection = createConnection();
+         ResultSet rs = connection.getMetaData().getTables(null, null, dbSinkConfig.tableName, null)
+    ) {
+      return rs.next();
+    } catch (java.sql.SQLException e) {
+      throw new IllegalArgumentException("SQL Exception thrown when trying to connect to driver", e);
+    } finally {
+      DBUtils.cleanup(jdbcDriverClass);
```
chtyim: Since the `jdbcDrvierClass` is passed to this class, the cleanup should be done from the caller side.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -81,13 +87,14 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
                                                                                 jdbcPluginId,
                                                                                 PluginProperties.builder().build());
     Preconditions.checkArgument(jdbcDriverClass != null, "JDBC Driver class must be found.");
+    ensureValidConnection(jdbcDriverClass);
```
chtyim: Similar to above, need to cleanup and deregister as well.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -124,6 +131,48 @@ private String getJDBCPluginId() {
...
+    } catch (Exception e) {
+      throw new IllegalArgumentException("could not ensure JDBCDriver is available", e);
+    }
+    try (Connection connection = createConnection()
+    ) {
+      assert(connection.isValid(0));
+    } catch (SQLException e) {
+      throw new IllegalArgumentException("SQL Exception thrown when trying to connect to driver", e);
+    } finally {
+      DBUtils.cleanup(jdbcDriverClass);
```
chtyim: Should move to caller.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -124,6 +131,48 @@ private String getJDBCPluginId() {
...
+      DriverManager.getDriver(dbSourceConfig.connectionString);
+    } catch (SQLException e) {
+      // Driver not found. We will try to register it with the DriverManager.
+      LOG.debug("Plugin Type: {} and Plugin Name: {}; Driver Class: {} not found. Registering JDBC driver via shim {} ",
+                dbSourceConfig.jdbcPluginType, dbSourceConfig.jdbcPluginName, jdbcDriverClass.getName(),
+                JDBCDriverShim.class.getName());
+      if (driverShim == null) {
+        driverShim = new JDBCDriverShim(jdbcDriverClass.newInstance());
+      }
+      DBUtils.deregisterAllDrivers(jdbcDriverClass);
```
chtyim: Not sure if it is correct to call in here (similarly for the DBSink class). Can you verify with @bdmogal ?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -168,25 +170,53 @@ private void setResultSetMetadata() throws Exception {
...
+    try {
+      ensureJDBCDriverIsAvailable(jdbcDriverClass);
+    } catch (Exception e) {
+      throw new IllegalArgumentException("JDBCDriver is not available", e);
+    }
+    try (Connection connection = createConnection();
+         ResultSet rs = connection.getMetaData().getTables(null, null, dbSinkConfig.tableName, null)
+    ) {
+      return rs.next();
+    } catch (java.sql.SQLException e) {
```
bdmogal: shouldn't need to be fully qualified.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
bdmogal: I don't think we should call cleanup here. ``DBUtils.cleanup`` should only be called from the ``destroy`` method

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
bdmogal: cleanup should only be called in the destroy method.

chtyim: Since now the `configure` method also instantiate SQL connection, it needs to be called in a `finally` block of the `configure` method as well, as destroy won't be called at configure time.

bdmogal: Oh true, yes, I missed that this was called from ``configure``. Yup, agree that its better to move it to a finally block in ``configure``. Easier to read that way as well.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -124,6 +131,48 @@ private String getJDBCPluginId() {
...
+    try {
+      ensureJDBCDriverIsAvailable(jdbcDriverClass);
+    } catch (Exception e) {
+      throw new IllegalArgumentException("could not ensure JDBCDriver is available", e);
+    }
+    try (Connection connection = createConnection()
+    ) {
+      assert(connection.isValid(0));
+    } catch (SQLException e) {
+      throw new IllegalArgumentException("SQL Exception thrown when trying to connect to driver", e);
```
bdmogal: Can this method instead just throw ``SQLException``? I feel its better to propagate checked exceptions as far as you can, then convert them into unchecked only if you have to (for example, because the calling method does not allow you to throw an exception)

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -124,6 +131,48 @@ private String getJDBCPluginId() {
...
   }
 
+  private void ensureValidConnection(Class<? extends Driver> jdbcDriverClass) {
+    try {
+      ensureJDBCDriverIsAvailable(jdbcDriverClass);
+    } catch (Exception e) {
+      throw new IllegalArgumentException("could not ensure JDBCDriver is available", e);
+    }
+    try (Connection connection = createConnection()
+    ) {
```
bdmogal: This line can move to the previous line

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -124,6 +131,48 @@ private String getJDBCPluginId() {
     return String.format("%s.%s.%s", "source", dbSourceConfig.jdbcPluginType, dbSourceConfig.jdbcPluginName);
   }
 
+  private void ensureValidConnection(Class<? extends Driver> jdbcDriverClass) {
+    try {
+      ensureJDBCDriverIsAvailable(jdbcDriverClass);
+    } catch (Exception e) {
```
bdmogal: Again, better to propagate the exception, so you don't have so many try..catch blocks. You can just have the call to ``ensureValidConnection`` in a try...catch.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
bdmogal: This is especially true because you are not taking any specific action in the catch, just propagating it as an unchecked exception, unlike line 157 below.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
bdmogal: I take back my comment. Like @chtyim said, cleanup should be called form the caller.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -168,25 +170,53 @@ private void setResultSetMetadata() throws Exception {
...
+      }
+      DBUtils.deregisterAllDrivers(jdbcDriverClass);
       DriverManager.registerDriver(driverShim);
     }
   }
 
+  private boolean tableExists(Class<? extends Driver> jdbcDriverClass) {
+    try {
+      ensureJDBCDriverIsAvailable(jdbcDriverClass);
+    } catch (Exception e) {
```
bdmogal: Please see my comments below for ``ensureValidConnection``, Better for this method to also throw checked exceptions instead of these ``try...catch`` blocks.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -270,6 +269,73 @@ public void testDBSource() throws Exception {
...
+      .put(Properties.DB.IMPORT_QUERY, importQuery)
+      .put(Properties.DB.COUNT_QUERY, countQuery)
+      .put(Properties.DB.JDBC_PLUGIN_NAME, "hypersql")
+      .build()
+    );
+
+    ETLBatchConfig etlConfigBadConn = new ETLBatchConfig("* * * * *", sourceBadConn, sink,
+                                                         Lists.<ETLStage>newArrayList());
+    AdapterConfig adapterConfigBadConn = new AdapterConfig("", TEMPLATE_ID.getId(),
+                                                           GSON.toJsonTree(etlConfigBadConn));
```
bdmogal: Is ``toJsonTree`` needed here or is ``toJson`` sufficient?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -270,6 +269,73 @@ public void testDBSource() throws Exception {
...
+
+    managerBadConn.start();
+    managerBadConn.waitForOneRunToFinish(5, TimeUnit.MINUTES);
+    managerBadConn.stop();
+  }
+
+  @Test(expected = Exception.class)
+  @Category(SlowTests.class)
+  public void testBadNameSink() throws Exception {
+
```
bdmogal: please remove these two empty lines

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -270,6 +269,73 @@ public void testDBSource() throws Exception {
...
+                                                        hsqlDBServer.getConnectionUrl(),
+                                                        Properties.DB.TABLE_NAME, "bad_sink_name",
+                                                        Properties.DB.COLUMNS, cols,
+                                                        Properties.DB.JDBC_PLUGIN_NAME, "hypersql"
+                                        ));
+    List<ETLStage> transforms = Lists.newArrayList();
+    ETLBatchConfig etlConfigBadSinkName = new ETLBatchConfig("* * * * *", source, sinkBadName, transforms);
+    Id.Adapter adapterIdBadSinkName = Id.Adapter.from(NAMESPACE, "dbSinkTestBadName");
+    AdapterConfig adapterConfigBadSinkName = new AdapterConfig("", TEMPLATE_ID.getId(),
+                                                               GSON.toJsonTree(etlConfigBadSinkName));
```
bdmogal: can you just use ``toJson()`` here as well?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -270,6 +269,73 @@ public void testDBSource() throws Exception {
     Assert.assertEquals(clobData, Bytes.toString(row2.get("CLOB_COL"), 0, clobData.length()));
   }
 
+  @Test(expected = Exception.class)
+  @Category(SlowTests.class)
+  public void testBadConnSource() throws Exception {
```
bdmogal: This test is missing some verification. I think just expecting ``Exception`` to be thrown may not be sufficient here, because, any sublass of ``Exception`` could be thrown at any LOC in this test?

Can we instead perhaps make sure that the job did not start? Or that there we sent some data from the source but there was no data in the ``Table`` after the test?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -270,6 +269,73 @@ public void testDBSource() throws Exception {
...
+    AdapterManager managerBadConn = createAdapter(adapterIdBadConn, adapterConfigBadConn);
+
+    managerBadConn.start();
+    managerBadConn.waitForOneRunToFinish(5, TimeUnit.MINUTES);
+    managerBadConn.stop();
+  }
+
+  @Test(expected = Exception.class)
+  @Category(SlowTests.class)
+  public void testBadNameSink() throws Exception {
```
bdmogal: same comment as https://github.com/caskdata/cdap/pull/3038/files#r33291094

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
bdmogal: So the purpose of this is to clean up default drivers that get added when you create a new instance of the ``jdbcDriverClass``. I think you could move it inside the above ``if`` statement, where you create the new instance. 

It would also be good to confirm this. You can run the ``BatchETLDBAdapterTest`` in debug mode, place a breakpoint in the ``deregisterDriver`` method in ``DriverManager.java`` and inspect how the ``registeredDrivers`` list gets modified when this code is executed. You should ensure that that list does not get filled up with new instances of ``jdbcDriverClass``, without getting cleaned up. 

I can help you with this tomorrow too, if you are unsure about

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
ataylor4: toJsonTree is necessary because AdapterConfig requires a Json element,
and toJson returns a string.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -91,6 +92,13 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
                                                                                 getJDBCPluginId(),
                                                                                 PluginProperties.builder().build());
     Preconditions.checkArgument(jdbcDriverClass != null, "JDBC Driver class must be found.");
+    try {
+      Preconditions.checkArgument(tableExists(jdbcDriverClass), "Invalid table name %s", dbSinkConfig.tableName);
+    } catch (Exception e) {
+      throw new IllegalArgumentException(e);
+    } finally {
+      DBUtils.cleanup(jdbcDriverClass);
```
poornachandra: Unrelated to this PR - can you move `DBUtils.cleanup(driverClass);` in `destroy()` method also into `finally` block?

poornachandra: Also, it would be a good idea to call `destroy()` method from here, so that the driver gets de-registered too.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -91,6 +92,13 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
                                                                                 getJDBCPluginId(),
                                                                                 PluginProperties.builder().build());
     Preconditions.checkArgument(jdbcDriverClass != null, "JDBC Driver class must be found.");
+    try {
+      Preconditions.checkArgument(tableExists(jdbcDriverClass), "Invalid table name %s", dbSinkConfig.tableName);
```
poornachandra: Having a method name as `tableExists` with a non-table name as parameter is confusing. Can we instead rename this method `validateDBConnection`? We can move the precondition into the method then. `validateDBConnection` will then validate if you can connect to the DB and whether the table exists.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -295,7 +368,7 @@ public void testDBSink() throws Exception {
     AdapterConfig adapterConfig = new AdapterConfig("", TEMPLATE_ID.getId(), GSON.toJsonTree(etlConfig));
     AdapterManager manager = createAdapter(adapterId, adapterConfig);
 
-    createInputData();
+//    createInputData();
```
poornachandra: Good idea to remove commented code if it is no longer required.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
```diff
@@ -168,25 +168,44 @@ private void setResultSetMetadata() throws Exception {
...
       LOG.debug("Plugin Type: {} and Plugin Name: {}; Driver Class: {} not found. Registering JDBC driver via shim {} ",
-                dbSinkConfig.jdbcPluginType, dbSinkConfig.jdbcPluginName, driverClass.getName(),
+                dbSinkConfig.jdbcPluginType, dbSinkConfig.jdbcPluginName, jdbcDriverClass.getName(),
                 JDBCDriverShim.class.getName());
-      driverShim = new JDBCDriverShim(driverClass.newInstance());
-      DBUtils.deregisterAllDrivers(driverClass);
+      if (driverShim == null) {
+        driverShim = new JDBCDriverShim(jdbcDriverClass.newInstance());
+      }
+      DBUtils.deregisterAllDrivers(jdbcDriverClass);
```
poornachandra: Unrelated to this PR - can you add the following comment to the above statement?
When JDBC Driver class is loaded, the driver class automatically registers itself with DriverManager. However, this driver is not usable due to different classloader used in plugins. Hence de-register this driver class, and any other driver classes registered by this classloader.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
```diff
@@ -81,13 +86,20 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
                                                                                 jdbcPluginId,
                                                                                 PluginProperties.builder().build());
     Preconditions.checkArgument(jdbcDriverClass != null, "JDBC Driver class must be found.");
+    try {
+      ensureValidConnection(jdbcDriverClass);
+    } catch (Exception e) {
+      throw new IllegalArgumentException(e);
+    } finally {
+      DBUtils.cleanup(jdbcDriverClass);
```
poornachandra: Would be a good idea to call `destroy()` from there too.

poornachandra:
@ataylor4 I see that we have a lot of duplicate code between `DBSource` and `DBSink` (again, unrelated to this PR). It would be a good idea to re-factor the duplicate code into a helper class. Can you talk to @bdmogal about this?

ataylor4:
We decided to do that in a separate PR to minimize reviewing the code
a second time.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/DBSource.java
ataylor4: https://github.com/caskdata/cdap/pull/3038#discussion_r33396842 Based on this discussion, I chose to call the destroy method instead and make the driverClass a field

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/DBSink.java
ataylor4: https://github.com/caskdata/cdap/pull/3038#discussion_r33396842 Based on this discussion, I chose to make the driverClass a field so that I could call destroy

gokulavasan:
The PR needs more investigation related to Classloader leak/running out of PermGen space. Will reopen once we have addressed those issues.

PR #3040, opened 2015-06-24
[CDAP-2767] Fix error messages in namespace delete

bdmogal:
Fixed error messages in namespace delete operations.

Also fixed some cdap clients to reuse HttpResponse.toString instead of redefining the logic in private methods.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
@@ -162,7 +162,10 @@ public class NamespaceHttpHandler extends AbstractAppFabricHttpHandler {
   @Path("/unrecoverable/namespaces/{namespace-id}")
   public void delete(HttpRequest request, HttpResponder responder, @PathParam("namespace-id") String namespace) {
     if (!cConf.getBoolean(Constants.Dangerous.UNRECOVERABLE_RESET, Constants.Dangerous.DEFAULT_UNRECOVERABLE_RESET)) {
-      responder.sendStatus(HttpResponseStatus.FORBIDDEN);
+      responder.sendString(HttpResponseStatus.FORBIDDEN,
+                           String.format("Namespace '%s' cannot be deleted because '%s' is not enabled. " +
+                                           "Please enable it and restart CDAP Master.",
+                                         namespace, Constants.Dangerous.UNRECOVERABLE_RESET));
       return;
     }
     Id.Namespace namespaceId = Id.Namespace.from(namespace);
@@ -184,7 +187,10 @@ public class NamespaceHttpHandler extends AbstractAppFabricHttpHandler {
   public void deleteDatasets(HttpRequest request, HttpResponder responder,
                              @PathParam("namespace-id") String namespace) {
     if (!cConf.getBoolean(Constants.Dangerous.UNRECOVERABLE_RESET, Constants.Dangerous.DEFAULT_UNRECOVERABLE_RESET)) {
-      responder.sendStatus(HttpResponseStatus.FORBIDDEN);
+      responder.sendString(HttpResponseStatus.FORBIDDEN,
+                           String.format("All datasets in namespace %s cannot be deleted because '%s' is not enabled." +
+                                           " Please enable it and restart CDAP Master.",
+                                         namespace, Constants.Dangerous.UNRECOVERABLE_RESET));
       return;
     }
     Id.Namespace namespaceId = Id.Namespace.from(namespace);
@@ -194,7 +200,8 @@ public class NamespaceHttpHandler extends AbstractAppFabricHttpHandler {
     } catch (NotFoundException e) {
       responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Namespace %s not found.", namespace));
     } catch (NamespaceCannotBeDeletedException e) {
-      responder.sendString(HttpResponseStatus.CONFLICT, e.getMessage());
+      responder.sendString(HttpResponseStatus.CONFLICT, String.format("Datasets in namespace %s cannot be deleted. " +
+                                                                       "Reason: %s", namespace, e.getReason()));
     } catch (Exception e) {
       LOG.error("Internal error while deleting namespace.", e);
       responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
@@ -258,20 +258,19 @@ public final class DefaultNamespaceAdmin implements NamespaceAdmin {
 
     if (checkProgramsRunning(namespaceId)) {
       throw new NamespaceCannotBeDeletedException(namespaceId,
-                                                  "Some programs are currently running in namespace " + namespaceId);
+                                                  String.format("Some programs are currently running in namespace " +
+                                                                  "'%s', please stop them before deleting datasets " +
+                                                                  "in the namespace.",
+                                                                namespaceId));
     }
 
-    LOG.info("Deleting data in namespace '{}'.", namespaceId);
     try {
       dsFramework.deleteAllInstances(namespaceId);
-    } catch (DatasetManagementException e) {
-      LOG.warn("Error while deleting data in namespace {}", namespaceId, e);
-      throw new NamespaceCannotBeDeletedException(namespaceId, e);
-    } catch (IOException e) {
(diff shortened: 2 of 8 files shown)


cdap-common/src/main/java/co/cask/cdap/common/namespace/AbstractNamespaceClient.java
```diff
@@ -88,11 +88,10 @@ public void create(NamespaceMeta namespaceMeta)
 
     Id.Namespace namespace = Id.Namespace.from(namespaceMeta.getName());
     URL url = resolve(String.format("namespaces/%s", namespace.getId()));
-    HttpResponse response = execute(HttpRequest.put(url).withBody(new Gson().toJson(namespaceMeta)).build());
+    HttpResponse response = execute(HttpRequest.put(url).withBody(GSON.toJson(namespaceMeta)).build());
     String responseBody = response.getResponseBodyAsString();
     if (response.getResponseCode() == HttpURLConnection.HTTP_OK) {
-      if (responseBody != null && responseBody.equals(String.format("Namespace '%s' already exists.",
-                                                                    namespaceMeta.getName()))) {
+      if (responseBody.equals(String.format("Namespace '%s' already exists.", namespaceMeta.getName()))) {
```
jwang47: I think responseBody is still nullable.

jwang47: Nevermind. https://github.com/caskdata/common/blob/develop/common-http/src/main/java/co/cask/common/http/HttpResponse.java#L60-L62

Still, `HttpResponse.responseBody` is nullable, but I don't know what you get if you do `new String(null, Charsets.UTF_8);`. Seems that `HttpResponse.getResponseBodyAsString()` is bad. Should return null if responseBody is null probably.

bdmogal: yeah, currently its not, but it probably should be. new PR?

jwang47: Let's have the code here assume it's nullable, and then fix common-http in a later release.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
```diff
@@ -162,7 +162,8 @@ public void create(HttpRequest request, HttpResponder responder, @PathParam("nam
   @Path("/unrecoverable/namespaces/{namespace-id}")
   public void delete(HttpRequest request, HttpResponder responder, @PathParam("namespace-id") String namespace) {
     if (!cConf.getBoolean(Constants.Dangerous.UNRECOVERABLE_RESET, Constants.Dangerous.DEFAULT_UNRECOVERABLE_RESET)) {
-      responder.sendStatus(HttpResponseStatus.FORBIDDEN);
+      responder.sendString(HttpResponseStatus.FORBIDDEN,
+                           "Namespace deletion is not enabled. Please enable it and restart CDAP master.");
```
awholegunch: CDAP Master

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -258,16 +258,16 @@ public synchronized void deleteDatasets(Id.Namespace namespaceId)
 
     if (checkProgramsRunning(namespaceId)) {
       throw new NamespaceCannotBeDeletedException(namespaceId,
-                                                  "Some programs are currently running in namespace " + namespaceId);
+                                                  String.format("Some programs are currently running in namespace " +
+                                                                  "'%s', please stop them before deleting datasets " +
```
awholegunch: Shouldn't this be "deleting data"?

cdap-common/src/main/java/co/cask/cdap/common/namespace/AbstractNamespaceClient.java
```diff
@@ -100,7 +99,7 @@ public void create(NamespaceMeta namespaceMeta)
     if (response.getResponseCode() == HttpURLConnection.HTTP_BAD_REQUEST) {
       throw new BadRequestException("Bad request: " + responseBody);
     }
-    throw new IOException("Cannot get create namespace. Reason: " + "getDetails(response)");
+    throw new IOException("Cannot get create namespace. Reason: " + response);
```
awholegunch: "Cannot create namespace. Reason: ", though if you can add the namespace name, even better:
```
String.format("Cannot create namespace '%s'. Reason: ", namespaceMeta.getName())
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
bdmogal: The API is to delete datasets -  the 'datasets' entity in CDAP. I think its fair here? Perhaps if we have something that is even more user facing than this (cli command/UI), we can make it data there?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
awholegunch: I'm wondering of this is good enough. We don't have a property "namespace deletion"; instead, we have "enable.unrecoverable.reset" . Perhaps this should be:
```
"Namespaces cannot be deleted because 'enable.unrecoverable.reset' is not enabled. Please enable it and restart CDAP Master."
```

awholegunch: Of course, this is fragile; is there a static for "enable.unrecoverable.reset"?

jwang47: Yeah, there's `Constants.Dangerous.UNRECOVERABLE_RESET`

bdmogal: I'm not sure about that. I thought about it, but to me, it is a dangerous property to expose as part of an HTTP response. If users really wish to delete, they should look at documentation?

jwang47: But if we tell them to enable, then we should tell them what the property is.

jwang47:
LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -258,16 +258,16 @@ public synchronized void deleteDatasets(Id.Namespace namespaceId)
...
+                                                                  "in the namespace.",
+                                                                namespaceId));
     }
 
     LOG.info("Deleting data in namespace '{}'.", namespaceId);
```
awholegunch: OK, but why is this "Deleting data in namespace '{}'." and not deleting datasets?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
```diff
@@ -258,16 +258,16 @@ public synchronized void deleteDatasets(Id.Namespace namespaceId)
...
     try {
       dsFramework.deleteAllInstances(namespaceId);
-    } catch (DatasetManagementException e) {
-      LOG.warn("Error while deleting data in namespace {}", namespaceId, e);
-      throw new NamespaceCannotBeDeletedException(namespaceId, e);
-    } catch (IOException e) {
+    } catch (DatasetManagementException | IOException e) {
       LOG.warn("Error while deleting data in namespace {}", namespaceId, e);
```
awholegunch: OK, but why is this not "Error while deleting datasets in namespace {}"?

awholegunch:
Two responses vis-a-vis your comment...

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
bdmogal: Fixed

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/namespace/DefaultNamespaceAdmin.java
bdmogal: Fixed

bdmogal:
http://builds.cask.co/browse/CDAP-DUT2103-3

awholegunch:
LGTM2

bdmogal:
Build passed. Merging.

PR #3042, opened 2015-06-24
add adapter data modelling

elfenheart:
Cluster: http://adapter1833-1000.dev.continuuity.net:9999/


cdap-ui/app/features/adapters/controllers/list-ctrl.js
```diff
@@ -32,18 +37,24 @@ angular.module(PKG.name + '.feature.adapters')
...
-      dataSrc.poll({
-        _cdapNsPath: '/adapters/' + app.name + '/status'
-      }, function(res) {
-        app.status = res.status;
-      });
+      var statusParams = angular.extend({
+        app: app.name
+      }, params);
+
+      myAdapterApi.pollStatus(statusParams)
```
ajainarayanan: Could we check if the status is STOPPED and stop the poll? Could avoid polling during STOPPED duration (as the status is not going to change unless the user starts it manually)

cdap-ui/app/features/adapters/controllers/list-ctrl.js
```diff
@@ -53,7 +64,7 @@ angular.module(PKG.name + '.feature.adapters')
             $state.go($state.current, $state.params, {reload: true});
           });
         }, function(err){
-          console.info("Adapter Delete Failed", err);
+          console.info('Adapter Delete Failed', err);
```
ajainarayanan: Should probably alert the user with the error?

cdap-ui/app/services/adapter/myadapterapi.js
```diff
@@ -1,11 +1,12 @@
-angular.module(PKG.name + '.feature.adapters')
-  .factory('AdapterApiFactory', function($resource, myHelpers, myCdapUrl) {
+angular.module(PKG.name + '.services')
+  .factory('myAdapterApi', function($resource, myHelpers, myCdapUrl) {
     var url = myCdapUrl.constructUrl,
         templatePath = '/templates',
-        adapterPath = '/namespaces/:namespace/adapters/:adapter'
+        adapterPath = '/namespaces/:namespace/adapters/:adapter',
```
ajainarayanan: we have ```save``` as ```/adapters/:adapterId``` shouldn't our get too be ```/adapter/:adapterId```

When I saw ```myAdapterApi.get(params)``` I thought we were getting the details of an adapter. Should we move this out to a separate resource file? Because we are having save, get, delete, list and only save uses a different url (however we can GET details of the adapter using the url that we use for save too).

ajainarayanan: However I just noticed it we are passing ```adapterId `` in the place of ```/:app``` resource param.

cdap-ui/app/features/adapters/controllers/list-ctrl.js
elfenheart: What happens in the case where there are more than 1 person? The other can start the adapter too right?

cdap-ui/app/services/adapter/myadapterapi.js
elfenheart: resolved. Got rid of Basepath and using adapterPath instead

ajainarayanan:
:+1: LGTM

PR #3045, opened 2015-06-25
Adding TPFSAvro as Source

rohanahata:
https://issues.cask.co/browse/CDAP-2733
http://builds.cask.co/browse/CDAP-DUT2071
#2999 
Have an LGTM on this PR... Created this branch to squash commits. 
http://builds.cask.co/browse/CDAP-DUT2106


bdmogal:
Rebase and reword commit to include Jira.  Other than that, this PR looks similar to #2999 at a quick glance. LGTM.

elfenheart:
Need to add JSON file for this ETLBatch source

gsps1:
@elfenheart , i noticed we have json files for types and not specific to source/sink. Eg: i only see one json file for kvTable and NOT two (source and sink). for TPFS we already have a json file, correct me if i am missing something

elfenheart:
@gsps1 As long as the requirement for the customizable fields for both source and sink are the same, then yes, it can be just one file.

ajainarayanan:
Changes to config json LGTM :+1:

PR #3046, opened 2015-06-25
Spark Refactor using Controller As

elfenheart:
Refactor Spark programs to use Controller As syntax.

Cannot test Spark in Cluster. Need to pull locally.


ajainarayanan:
:+1: LGTM. Any specific reason why we are hiding everything in the status page until the spark program is run atleast once? The other programs shows a default status view with/without a run.

ajainarayanan:
Also wondering if we need to be consistent with the changes made in this PR (https://github.com/caskdata/cdap/pull/3047) Pulled it locally and it seemed to work fine. Could we figure out if we need those changes or is that PR invalid (as in the runs dropdown works just fine when we navigate between states?)

elfenheart:
Updated PR to be consistent with #3047

ajainarayanan:
:+1: LGTM

PR #3048, opened 2015-06-25
Feature/cdap 2866 Ensure files are read in even after failure

jayfjin:
JIRA- https://issues.cask.co/browse/CDAP-2739
BUILD - http://builds.cask.co/browse/CDAP-DUT2108-1

An arraylist of times that still need to be read are now stored in the table. If the file date falls within one of the ranges specified by this arraylist, then the file is accepted. If not, then the file is rejected.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
@@ -31,6 +31,7 @@ import co.cask.cdap.template.etl.api.PipelineConfigurer;
 import co.cask.cdap.template.etl.api.batch.BatchSource;
 import co.cask.cdap.template.etl.api.batch.BatchSourceContext;
 import co.cask.cdap.template.etl.common.BatchFileFilter;
+import com.google.common.collect.Lists;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import org.apache.hadoop.conf.Configuration;
@@ -43,8 +44,10 @@ import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Type;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
@@ -88,13 +91,13 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
   private static final Gson GSON = new Gson();
   private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
   private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
-
+  private static final Type ARRAYLIST_DATE_TYPE  = new TypeToken<ArrayList<Date>>() { }.getType();
   private static final Type MAP_STRING_STRING_TYPE = new TypeToken<Map<String, String>>() { }.getType();
 
   private final FileBatchConfig config;
   private KeyValueTable table;
   private Date prevMinute;
-  private String previousTimeRead;
+  private String datesToRead;
 
   public FileBatchSource(FileBatchConfig config) {
     this.config = config;
@@ -133,13 +136,17 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
 
     if (config.timeTable != null) {
       table = context.getDataset(config.timeTable);
-      String lastTimeRead = Bytes.toString(table.read(LAST_TIME_READ));
-      if (lastTimeRead == null) {
-        lastTimeRead = "0";
+      String datesToRead = Bytes.toString(table.read(LAST_TIME_READ));
+      if (datesToRead == null) {
+        List<Date> firstRun = Lists.newArrayList(new Date(0));
+        datesToRead = GSON.toJson(firstRun, ARRAYLIST_DATE_TYPE);
       }
-      previousTimeRead = lastTimeRead;
-      table.write(LAST_TIME_READ, DATE_FORMAT.format(prevMinute));
-      conf.set(LAST_TIME_READ, lastTimeRead);
+      List<Date> attempted = Lists.newArrayList(prevMinute);
+      String updatedDatesToRead = GSON.toJson(attempted, ARRAYLIST_DATE_TYPE);
+      if (!updatedDatesToRead.equals(datesToRead)) {
+        table.write(LAST_TIME_READ, updatedDatesToRead);
+      }
+      conf.set(LAST_TIME_READ, datesToRead);
     }
 
     conf.set(CUTOFF_READ_TIME, DATE_FORMAT.format(prevMinute));
@@ -165,7 +172,11 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
   @Override
   public void onRunFinish(boolean succeeded, BatchSourceContext context) {
     if (!succeeded && table != null && USE_TIMEFILTER.equals(config.fileRegex)) {
-      table.write(LAST_TIME_READ, previousTimeRead);
+      List<Date> existing = GSON.fromJson(Bytes.toString(table.read(LAST_TIME_READ)), ARRAYLIST_DATE_TYPE);
+      List<Date> failed = GSON.fromJson(datesToRead, ARRAYLIST_DATE_TYPE);
+      failed.add(prevMinute);
+      failed.addAll(existing);
+      table.write(LAST_TIME_READ, GSON.toJson(failed, ARRAYLIST_DATE_TYPE));
     }
   }
 
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
@@ -17,6 +17,8 @@
 package co.cask.cdap.template.etl.common;
 
 import co.cask.cdap.template.etl.batch.source.FileBatchSource;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.conf.Configured;
 import org.apache.hadoop.fs.Path;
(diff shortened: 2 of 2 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -40,12 +44,21 @@
...
+  private static final Gson GSON = new Gson();
+  private static final Type ARRAYLIST_DATE_TYPE  = new TypeToken<ArrayList<Date>>() { }.getType();
   private boolean useTimeFilter;
 
   private Pattern regex;
   private String pathName;
   private String lastRead;
   private Date prevMinute;
+  /*
+   * dateRangesToRead is an odd length ArrayList of Dates. The non terminal elements are tuples of Dates that
```
anwar6953: Perhaps, also describe what the last element signifies? Or did I miss that somewhere?

jayfjin: On line 57-59

anwar6953: got it

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -108,11 +113,32 @@ public void setConf(Configuration conf) {
...
 
+  /**
+   * Determines if a file should be read in
+   *
+   * @param fileDate when the file was created
+   * @return true if the file is to be read, false otherwise
+   */
+  private boolean toRead(Date fileDate) {
+    for (int i = 0; i < dateRangesToRead.size() / 2; i++) {
+      if (fileDate.compareTo(dateRangesToRead.get(2 * i)) > 0 &&
```
anwar6953: Seems like the time ranges are exclusive of the begin value, and inclusive of the end value?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -85,7 +90,7 @@ public boolean accept(Path path) {
       LOG.warn("Couldn't parse file: " + path.getName());
       fileDate = prevMinute;
```
anwar6953: It seems like if `fileDate` is equal to `prevMinute`, then `toRead(fileDate)` will return true.
So, if a `ParseException` does happen, it will definitely read the file.
Am I understanding that correctly?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
jayfjin: Yes, or otherwise events can be read in twice

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
jayfjin: Yes, do you think not reading in the file would be better behavior?

anwar6953: I'm unsure, but if there's a file with a "malformed" name, this'll mean that we read it every run, right?
Instead, would it be better to not read it or also propagate the exception?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
anwar6953: In our code base, we more commonly have the start range being inclusive and the end range exclusive. Either way works, but I think having a common convention is better.
Even some Java APIs are like that, where the beginIndex is inclusive but the endIndex is exclusive:
```java
String.substring(int beginIndex, int endIndex)
String.subSequence(int beginIndex, int endIndex)
List.subList(int fromIndex, int toIndex)
```

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -40,12 +44,21 @@
   //length of 'YYYY-MM-dd-HH-mm"
   private static final int DATE_LENGTH = 16;
   private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
+  private static final Gson GSON = new Gson();
+  private static final Type ARRAYLIST_DATE_TYPE  = new TypeToken<ArrayList<Date>>() { }.getType();
   private boolean useTimeFilter;
```
bdmogal: nit: remove unnecessary empty line

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -133,13 +134,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
 
     if (config.timeTable != null) {
       table = context.getDataset(config.timeTable);
-      String lastTimeRead = Bytes.toString(table.read(LAST_TIME_READ));
-      if (lastTimeRead == null) {
-        lastTimeRead = "0";
+      String datesToRead = Bytes.toString(table.read(LAST_TIME_READ));
+      if (datesToRead == null) {
+        ArrayList<Date> firstRun = new ArrayList();
```
bdmogal: Use ``Lists.newArrayList()``

bdmogal: Also, better to keep the type as ``List``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -133,13 +134,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+        firstRun.add(new Date(0));
+        datesToRead = GSON.toJson(firstRun, ARRAYLIST_DATE_TYPE);
+      }
+      ArrayList<Date> attempted = new ArrayList();
```
bdmogal: ``List<Date> attempted = Lists.newArrayList()``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -133,13 +134,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
 
     if (config.timeTable != null) {
       table = context.getDataset(config.timeTable);
-      String lastTimeRead = Bytes.toString(table.read(LAST_TIME_READ));
-      if (lastTimeRead == null) {
-        lastTimeRead = "0";
+      String datesToRead = Bytes.toString(table.read(LAST_TIME_READ));
+      if (datesToRead == null) {
+        ArrayList<Date> firstRun = new ArrayList();
+        firstRun.add(new Date(0));
```
bdmogal: Also, you can merge this with the last statement.

``List<Date> firstRun = Lists.newArrayList(new Date(0));``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -133,13 +134,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+        datesToRead = GSON.toJson(firstRun, ARRAYLIST_DATE_TYPE);
+      }
+      ArrayList<Date> attempted = new ArrayList();
+      attempted.add(prevMinute);
```
bdmogal: merge with previous line.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -133,13 +134,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+      String updatedDatesToRead = GSON.toJson(attempted, ARRAYLIST_DATE_TYPE);
+      if (!updatedDatesToRead.equals(datesToRead)) {
+        table.write(LAST_TIME_READ, GSON.toJson(attempted, ARRAYLIST_DATE_TYPE));
```
bdmogal: don't need ``toJson`` again? can just reuse ``updatedDatesToRead``?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -165,7 +172,11 @@ public void transform(KeyValue<LongWritable, Object> input, Emitter<StructuredRe
   @Override
   public void onRunFinish(boolean succeeded, BatchSourceContext context) {
     if (!succeeded && table != null && USE_TIMEFILTER.equals(config.fileRegex)) {
-      table.write(LAST_TIME_READ, previousTimeRead);
+      ArrayList<Date> existing = GSON.fromJson(Bytes.toString(table.read(LAST_TIME_READ)), ArrayList.class);
```
bdmogal: Make these type ``List`` as well?

bdmogal: Also use ``ARRAYLIST_DATE_TYPE`` as the type

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -108,11 +113,32 @@ public void setConf(Configuration conf) {
...
     try {
       prevMinute = sdf.parse(conf.get(FileBatchSource.CUTOFF_READ_TIME));
     } catch (ParseException pe) {
       prevMinute = new Date(System.currentTimeMillis());
     }
   }
-}
 
+  /**
+   * Determines if a file should be read in
```
bdmogal: Although this is a private method, I think we should describe the logic in detail in the javadoc

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -108,11 +113,32 @@ public void setConf(Configuration conf) {
...
+   *
+   * @param fileDate when the file was created
+   * @return true if the file is to be read, false otherwise
+   */
+  private boolean toRead(Date fileDate) {
```
bdmogal: nit: Just by convention, a method ``toX`` to me means that you're converting a certain entity to another type (e.g. ``toString``, ``Bytes.toLong``, etc. ). Should we call this ``isWithinRange`` or something on those lines?

bdmogal:
LGTM, presume that this scenario has been tested.

PR #3049, opened 2015-06-25
CDAP-2776: Configure an App during App deployment+creation

gokulavasan:
* Currently performs a GSON deserialization of the configuration which is provided as part of the Header when deploying an Application (this is the single step process for both ease and backward compatibility).
Separate endpoint will be added later to create from an Artifact which will accept the configuration in the body instead of the Header.

* Add +ve and -ve test cases for the inMemoryConfigurator and deployment pipeline.

Build : http://builds.cask.co/browse/CDAP-DUT2110


cdap-api/src/main/java/co/cask/cdap/api/app/ApplicationContext.java
```diff
@@ -30,7 +32,8 @@
   /**
    * Get the configuration object.
    *
-   * @return configuration object
+   * @return application configuration provided during application creation, else null
```
albertshau: I think it would be better to return an empty Config object rather than returning null. Otherwise every app needs to perform null checks before using the config.

cdap-app-fabric/src/main/java/co/cask/cdap/app/DefaultApplicationContext.java
```diff
@@ -26,8 +30,25 @@
...
+  private static final Gson GSON = new Gson();
+  private final T appConfig;
+
+  public DefaultApplicationContext() {
+    this.appConfig = null;
+  }
+
+  //TODO : Handle JsonSyntax Exception better and throw a more meaningful error. This will be done when we
+  // use PluginInstantiator methods instead of GSON deserialization
+  public DefaultApplicationContext(TypeToken<?> configToken, @Nullable String configString) {
```
albertshau: think it would be better to take the Type rather than the TypeToken. Then gson is an internal detail and I can just pass in MyConfig.class directly instead of wrapping it in a type token.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
```diff
@@ -113,14 +117,16 @@ public InMemoryConfigurator(Location archive) {
...
     return new DefaultConfigResponse(0, CharStreams.newReaderSupplier(specJson));
   }
 
-  private static String getSpecJson(Application app, final String bundleVersion) {
+  private static String getSpecJson(Application app, final String bundleVersion, final String configString) {
     // Now, we call configure, which returns application specification.
     DefaultAppConfigurer configurer = new DefaultAppConfigurer(app);
-    app.configure(configurer, new DefaultApplicationContext());
+    TypeToken typeToken = TypeToken.of(app.getClass());
+    TypeToken<?> resultToken = typeToken.resolveType(Application.class.getTypeParameters()[0]);
```
albertshau: does this work if the app is not parameterized?

cdap-app-fabric/src/main/java/co/cask/cdap/app/DefaultApplicationContext.java
```diff
@@ -26,8 +30,25 @@
...
 public class DefaultApplicationContext<T extends Config> implements ApplicationContext<T> {
 
+  private static final Gson GSON = new Gson();
+  private final T appConfig;
+
+  public DefaultApplicationContext() {
+    this.appConfig = null;
+  }
+
+  //TODO : Handle JsonSyntax Exception better and throw a more meaningful error. This will be done when we
```
albertshau: is there a jira open for this? if so lets put it in the comment, otherwise lets open one and then put it in the comment.

cdap-app-fabric/src/test/java/co/cask/cdap/ConfigTestApp.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+package co.cask.cdap;
+
+import co.cask.cdap.api.Config;
+import co.cask.cdap.api.app.AbstractApplication;
+import co.cask.cdap.api.data.stream.Stream;
+import co.cask.cdap.api.dataset.table.Table;
+import co.cask.cdap.api.worker.AbstractWorker;
+import com.clearspring.analytics.util.Preconditions;
```
albertshau: use google's

albertshau:
some small comments, otherwise lgtm

cdap-app-fabric/src/test/java/co/cask/cdap/ConfigTestApp.java
anwar6953: while you're at it, perhaps change the import in following two files as well.
`AdapterDefinition.java`
`AdapterService.java`

cdap-api/src/main/java/co/cask/cdap/api/app/ApplicationContext.java
gokulavasan: But we have to return an object of type T. So will returning empty Config object work?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
gokulavasan: it will work as long as user doesn't provide any configuration string in the header. if he does, then an GSON deserialization will fail since it will try to create a Config object with the config string provided.

gokulavasan:
@albertshau Addressed comments. Please take a look again. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/app/ApplicationContext.java
albertshau: we should be able to return an empty T I think.  With a future PR that will fail app creation if required fields are not present, similar to what happens with PluginConfig today.

gokulavasan: Changed logic to invoke default Constructor. Hence the user need not check for null.

gokulavasan:
@albertshau Updated PR. Please take a look. Thanks.

cdap-app-fabric/src/main/java/co/cask/cdap/app/DefaultApplicationContext.java
```diff
@@ -26,8 +32,30 @@
...
+  private static final Gson GSON = new Gson();
+  private final T appConfig;
+
+  public DefaultApplicationContext() {
+    this.appConfig = null;
+  }
+
+  //We know that the type of the configToken instance will be T.
+  @SuppressWarnings("unchecked")
+  public DefaultApplicationContext(Class<? extends Application> app, @Nullable String configString)
```
albertshau: hm sorry we've visited this multiple times, but I think it would be cleaner to just pass in the config and do the deserialization outside of this class.

albertshau:
one comment, then lgtm

gokulavasan:
@albertshau Addressed comment. Please take a look. Thanks!

albertshau:
one minor comment, then good to merge

gokulavasan:
@albertshau Addressed comment. Thanks for the review. Will merge after build passes.

PR #3047, opened 2015-06-25
Fix/ui programs backendbutton

ajainarayanan:
- Back button in programs view didn't change the run dropdown - FIXED
- Mapreduce runs detailed view doesn't show "No Runs!" when there is no runs of the mapreduce - FIXED
runsdropdownbackbutton

PR #3050, opened 2015-06-25
Fix/ui programs backbutton develop

ajainarayanan:
Exactly the same PR as (https://github.com/caskdata/cdap/pull/3047) opening it against develop. Can be cherry-picked if we are releasing a 3.0.2 release later.


elfenheart:
:+1: LGTM

PR #3051, opened 2015-06-25
[CDAP-2932] [CDAP-2660] [CDAP-2833] Fix/ui polling intervals

elfenheart:
- Change polling interval for Flows Metrics (for flow diagrams)
- Change polling interval for Workflow Schedule status
- Add transition disabled button when Workflow Schedule is on status transition

screen shot 2015-06-25 at 1 33 29 pm


ajainarayanan:
:+1: LGTM

PR #3053, opened 2015-06-26
Feature/ui workflow babelify

ajainarayanan:
- Adds babel to gulp build process
- Adds sourcemaps to debug es6 code in browsers
- Modifies workflows feature to use basic es6 features.

es6workflows
Cluster: nodeproxy1736-1000.dev.continuuity.net:9999


cdap-ui/app/features/workflows/controllers/tabs/schedule-ctrl.js
```diff
@@ -1,87 +1,92 @@
...
+      scope: this.$scope
+    }, {},
+    function success() {},
+    function error(err) {
+      this.myAlert({
+        title: 'Cannot Suspend Schedule',
+        content: err
       });
-    };
+    }.bind(this));
```
elfenheart: do we still need the .bind(this)?

cdap-ui/app/features/workflows/controllers/tabs/schedule-ctrl.js
```diff
@@ -1,87 +1,92 @@
...
+      scheduleId: obj.schedule.name,
+      scope: this.$scope
+    }, {},
+    function success() {},
+    function error(err) {
+      this.myAlert({
+        title: 'Cannot Resume Schedule',
+        content: err
+      });
+    }.bind(this));
```
elfenheart: same comment as line 72

cdap-ui/app/features/workflows/service/graph-helpers.js
```diff
@@ -118,9 +119,33 @@ angular.module(PKG.name + '.services')
       }
     }
 
+    function addStartAndEnd(nodes) {
```
elfenheart: do we need the *function* keyword?

cdap-ui/app/features/workflows/controllers/tabs/schedule-ctrl.js
ajainarayanan: Only fat arrow functions ```()=>``` don't need the executing scope. Apart from that if we are using ```function () {..}``` syntax we need to bind it with a scope if it needs to be executed in a specific scope.

cdap-ui/app/features/workflows/controllers/tabs/schedule-ctrl.js
ajainarayanan: Only fat arrow functions ```()=>``` don't need the executing scope. Apart from that if we are using ```function () {..}``` syntax we need to bind it with a scope if it needs to be executed in a specific scope.

cdap-ui/app/features/workflows/service/graph-helpers.js
ajainarayanan: function still exists in es6. We just have an extra syntax (fat arrow functions) that allows us to express anonymous functions.

ajainarayanan:
TODO: We need to modify our gulp build process for development. Right now each change takes around 7s to build app.js file and reload the browser. Its extremely in-efficient for us during development. Will fix it in the next PR.

elfenheart:
@ajainarayanan I think part of modifying the Gulp should be deprecating angular-strap modal (or if we can even remove angular-strap completely).

We have to modify Cask Confirmable for that though...

elfenheart:
:+1: LGTM

PR #3057, opened 2015-06-26
CDAP-2846 support for snapshot versions to the artifact store

albertshau:
Snapshot versions are mutable unlike other artifacts.
Also fixing up some bugs where the absence of a plugin or artifact
was not correctly throwing a not found exception.
Also changing ArtifactStore to return unmodifiable collections,
and to return location of the artifact when getting plugins.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
@@ -333,6 +333,23 @@ public class ArtifactStore {
   public void write(final Id.Artifact artifactId, final ArtifactMeta artifactMeta, final InputStream artifactContents)
     throws WriteConflictException, ArtifactAlreadyExistsException, IOException {
 
+    // if we're not a snapshot version, check that the artifact doesn't exist already.
+    final ArtifactCell artifactCell = new ArtifactCell(artifactId);
+    if (!artifactId.getVersion().isSnapshot()) {
+      byte[] existingMeta = metaTable.executeUnchecked(
+        new TransactionExecutor.Function<DatasetContext<Table>, byte[]>() {
+          @Override
+          public byte[] apply(DatasetContext<Table> context) throws Exception {
+            Table table = context.get();
+            return table.get(artifactCell.rowkey, artifactCell.column);
+          }
+        });
+
+      if (existingMeta != null) {
+        throw new ArtifactAlreadyExistsException(artifactId);
+      }
+    }
+
     Location fileDirectory =
       locationFactory.get(artifactId.getNamespace(), ARTIFACTS_PATH).append(artifactId.getName());
     Locations.mkdirsIfNotExists(fileDirectory);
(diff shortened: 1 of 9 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2119

cdap-proto/src/main/java/co/cask/cdap/proto/artifact/ArtifactVersion.java
```diff
@@ -0,0 +1,196 @@
...
+ * Represents version of an artifact. It will try to parse the version string of format
+ *
+ * <pre>
+ * {@code
+ * parse version string in format of [major].[minor].[fix](-|.)[suffix]
+ * }
+ * </pre>
+ */
+@Beta
+public final class ArtifactVersion implements Comparable<ArtifactVersion> {
```
gokulavasan: This class looks similar to the PluginVersion class. Will we remove the latter later?

albertshau: yep

cdap-app-fabric/src/main/java/co/cask/cdap/internal/filesystem/LocationCodec.java
```diff
@@ -49,6 +51,11 @@ public JsonElement serialize(Location src, Type typeOfSrc, JsonSerializationCont
   public Location deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) {
     JsonObject jsonObj = json.getAsJsonObject();
     String uri = jsonObj.get("uri").getAsString();
-    return lf.create(uri);
+    try {
```
gokulavasan: Any reason to switch to create(URI) instead of create(String) ?

albertshau: yup the string version is interpreted as a path, so the whole thing doesn't work (this class apparently wasn't used anywhere before this PR). It would deserialize 'file:/my/path' to a location of '/my/path/file:/my/path'.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -287,26 +257,126 @@ public ArtifactInfo getArtifact(Id.Artifact artifactId) throws ArtifactNotExists
...
+
+        @Override
+        public Boolean apply(DatasetContext<Table> context) throws Exception {
+          Table table = context.get();
+
+          ArtifactCell artifactCell = new ArtifactCell(artifactId);
+          byte[] existingMetaBytes = table.get(artifactCell.rowkey, artifactCell.column);
+          boolean isSnapshot = artifactId.getVersion().isSnapshot();
+          if (existingMetaBytes != null && !isSnapshot) {
+            // non-snapshot artifacts are immutable. If there is existing metadata, stop here.
```
gokulavasan: So we will always copy the JAR to location irrespective of whether we finally choose to update the metadata or not?

gokulavasan:
One question. Rest LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: hm thats a good point, will check before copying (though we'll still have to check after copying)

albertshau:
@gokulavasan addressed comment, please take another look

gokulavasan:
LGTM :+1:

PR #3058, opened 2015-06-26
Fix/ui remove unusedcode

ajainarayanan:
- Removes un-used directives
- Renames files to be consistent.


elfenheart:
Need to modify Flowlet Input and Output graphs with C3 so we can deprecate Epoch.

ajainarayanan:
I think the comment (https://github.com/caskdata/cdap/pull/3060#issuecomment-116847774) should belong here. Anyhow removed references to modules that we no longer use.

elfenheart:
Are we still using Angular Strap *button, dropdown, and alert* modules?

ajainarayanan:
@elfenheart Yes ```$dropdown and $alert``` services are from angular-strap. Have removed bs-button (used in some admin template which we are not using)

elfenheart:
:+1: LGTM

PR #3060, opened 2015-06-27
Modifies Gulpfile to speed up build process for development

ajainarayanan:
- Separates out dev build vs prod build (ui js build process)
Mainly addresses ```gulp watch``` when developing UI. ```gulp build``` should still work exactly the same.

### DIFF
--- a/cdap-ui/Gulpfile.js
+++ b/cdap-ui/Gulpfile.js
@@ -162,24 +162,84 @@ gulp.task('js:$modal', function() {
   .pipe(gulp.dest('./bower_components/angular-bootstrap'));
 });
 
+function getEs6Features(isNegate) {
+  /*
+    When we move a feature to use es6 all we need to do is add it here for the build/dev process.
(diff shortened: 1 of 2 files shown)


elfenheart:
There are Bower components that we are not using anymore, should we remove that as well?

Angular Strap: tabs (my-tabs is using this, but are we using my-tabs anywhere else), dropdown, alert, button

We can remove these from the gulp and also in main.js

cdap-ui/Gulpfile.js
```diff
@@ -162,24 +162,84 @@ gulp.task('js:$modal', function() {
...
     v: pkg.version
   });
 
+  return gulp.src(
+      [
+        './app/main.js',
+        '!./app/lib/c3.js',
+        './app/features/*/module.js',
+        './app/**/*.js',
+        '!./app/**/*-test.js'
```
elfenheart: since we don't have proper test yet, should we comment this out?

elfenheart:
:+1: LGTM

PR #3061, opened 2015-06-29
CDAP-2761: instrumentation of Cube and metrics code

abaranau:
Tested in local, distributed modes.

Couple notes:

* You'll find in watchdog a lot of passing thru of MetricsContext or MetricsCollector. As per discussion with Trex we decided to use that approach first. The alternative would be using approach similar to logging: pass context thru inheritable threadlocal. Current metrics stuff differs from logging (at least) by changing context, including in same thread. In future, we will likely do same for logging (as discussed like 2 years ago:)) - in any case we'll likely use unified approach. We'll see how - as we have more cases.

* Some metrics processing metrics are written directly to store, some are going thru normal aggregation (Kafka). For now, only "operating" metrics, e.g. metrics processing latency and # of processed metrics are emitted directly: we want to see these right away, without any delay. Others are collected in re-usable components and it is natural easiest way to emit those in a standard way. That means in idle system couple metrics (less than 10) are always aggregated and processed. Which is fine. But we may improve in future.

* Added metrics names may seem confusing or need to go into code to understand and interpret. We may improve naming in future. Note that those are "debug" level metrics - you only need those during heavy optimization and debugging.. So for now - it is ok.

* Cube dataset instrumentation (separate from metrics code - txnl) to be tested separatelly.


abaranau:
http://builds.cask.co/browse/CDAP-DUT2123-1 GREEN

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
```diff
@@ -76,15 +79,22 @@ public FactTable get(int resolution, int ignoredRollTime) {
...
       }
     };
+
     this.cube = Suppliers.memoize(new Supplier<Cube>() {
       @Override
       public Cube get() {
-        // 1 sec, 1 min, 1 hour and "all time totals"
-        return new DefaultCube(resolutions, factTableSupplier, createAggregations());
+        DefaultCube cube = new DefaultCube(resolutions, factTableSupplier, createAggregations());
+        cube.setMetricsCollector(metricsContext);
```
sreevatsanraman: We should add capabilities to turn off the instrumentation.

sreevatsanraman:
The changes look good, however we should have capabilities to turn on/off instrumentation. By default should be off.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/cube/DefaultCube.java
```diff
@@ -82,12 +85,14 @@ public void add(CubeFact fact) {
...
   public void add(Collection<? extends CubeFact> facts) {
     List<Fact> toWrite = Lists.newArrayList();
+    int dimValuesCount = 0;
     for (CubeFact fact : facts) {
       for (Aggregation agg : aggregations.values()) {
         if (agg.accept(fact)) {
           List<DimensionValue> dimensionValues = Lists.newArrayList();
           for (String dimensionName : agg.getDimensionNames()) {
             dimensionValues.add(new DimensionValue(dimensionName, fact.getDimensionValues().get(dimensionName)));
+            dimValuesCount++;
```
gsps1: can do, dimValuesCount += dimensionValues.size() after this for loop, i am not sure if this metric will be helpful though.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/cube/DefaultCube.java
```diff
@@ -173,7 +193,14 @@ SELECT count('read.ops')                                           << measure na
     FactTable table = resolutionToFactTable.get(query.getResolution());
     FactScanner scanner = table.scan(scan);
     Table<Map<String, String>, String, Map<Long, Long>> resultMap = getTimeSeries(query, scanner);
-    return convertToQueryResult(query, resultMap);
+
+    incrementMetric("cube.query.request.success.count", 1);
+    incrementMetric("cube.query.result.map.size", resultMap.size());
```
gsps1: table.size ?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/cube/DefaultCube.java
```diff
@@ -154,11 +169,16 @@ SELECT count('read.ops')                                           << measure na
...
       agg = findAggregation(query);
       if (agg == null) {
+        incrementMetric("cube.query.request.failure.count", 1);
         throw new IllegalArgumentException("There's no data aggregated for specified dimensions " +
                                              "to satisfy the query: " + query.toString());
       }
     }
 
+    // tell how many queries end up querying specific pre-aggregated views and resolutions
+    incrementMetric("cube.query.agg." + agg + ".count", 1);
```
gsps1: how does this metric name look like ? , i don't see toString in DefaultAggregation

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/timeseries/FactTable.java
```diff
@@ -127,6 +135,10 @@ public void add(List<Fact> facts) {
     // todo: replace with single call, to be able to optimize rpcs in underlying table
     timeSeriesTable.put(convertedGaugesTable);
     timeSeriesTable.increment(convertedIncrementsTable);
+    if (metrics != null) {
+      metrics.increment("factTable.put.count", convertedGaugesTable.size());
+      metrics.increment("factTable.increment.count", convertedIncrementsTable.size());
```
gsps1: would it be useful to add the resolution of the factTable in the metric name ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
```diff
@@ -120,6 +130,9 @@ public Cube get() {
     aggs.put("flow.queue", new DefaultAggregation(
       ImmutableList.of(Constants.Metrics.Tag.NAMESPACE, Constants.Metrics.Tag.APP,
                        Constants.Metrics.Tag.FLOW, Constants.Metrics.Tag.CONSUMER,
+                       Constants.Metrics.Tag.PRODUCER, Constants.Metrics.Tag.FLOWLET_QUEUE),
+      ImmutableList.of(Constants.Metrics.Tag.NAMESPACE, Constants.Metrics.Tag.APP,
+                       Constants.Metrics.Tag.FLOW, Constants.Metrics.Tag.CONSUMER,
```
gsps1: unrelated change ?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
```diff
@@ -44,21 +46,22 @@
...
   public MetricsMessageCallback(DatumReader<MetricValues> recordReader,
                                 Schema recordSchema,
-                                MetricStore metricStore) {
+                                MetricStore metricStore,
+                                @Nullable
+                                MetricsContext metricsContext) {
     this.recordReader = recordReader;
     this.recordSchema = recordSchema;
     this.metricStore = metricStore;
+    this.metricsContext = metricsContext == null ? Collections.<String, String>emptyMap() : metricsContext.getTags();
```
gsps1: what does an empty map mean, they don't have any context? , then it becomes inefficient to search/query them right ?

abaranau:
> The changes look good, however we should have capabilities to turn on/off instrumentation. By default should be off.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/cube/DefaultCube.java
abaranau: yeah, not sure :)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/cube/DefaultCube.java
abaranau: oh shoot. Nice catch. It should be agg name :)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/cube/DefaultCube.java
abaranau: result.size

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/MetricsMessageCallback.java
abaranau: yes, there could be empt context. It is accepting null for conveninece

abaranau: it may be not super efficient, but fine, if that is what desired

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
abaranau: why?

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
abaranau: well, while testing and debugging - found that we do write too many data than needed - i.e. that filter was absent. It needs to be there. Fine within this PR:)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/timeseries/FactTable.java
abaranau: yes it will be useful!

abaranau:
@sreevatsanraman @gsps1 please take another look

abaranau:
http://builds.cask.co/browse/CDAP-DUT2123-2

sreevatsanraman:
> Why do we ever want to turn it off? Esp. by default. Why?
Discussed with Sasha, since we are going to be emitting fewer metrics (~10) compared to the rest of the system/app metrics, this is no longer a concern.

abaranau:
@gsps1 please take a look

gsps1:
LGTM

abaranau:
http://builds.cask.co/browse/CDAP-DUT2123-5 GREEN

PR #3063, opened 2015-06-29
[CDAP-2890] include templates profile when building parcel to include…

dereklwood:
… ETL jars

Fixes one aspect of https://issues.cask.co/browse/CDAP-2890.  Updating ``BUILD.rst`` command for parcel build to include the templates profile, and setting MAVEN_OPTS (required on my machine).

I will update the bamboo builds after merging.


wolf31o2:
:+1: LGTM

PR #3067, opened 2015-06-30
[CDAP-2893] Log parser transform

jayfjin:
JIRA - https://issues.cask.co/browse/CDAP-2893
BUILD - http://builds.cask.co/browse/CDAP-DUT2134-2

Custom Transform that parses various log formats (currently only CLF and S3) and extracts information such as URI, IP, browser, device, and time. Part of the Caskalytics project.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
@@ -90,7 +90,6 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
   private static final String FILESYSTEM_DESCRIPTION = "Distributed file system to read in from.";
   private static final Gson GSON = new Gson();
   private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
-  private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
   private static final Type ARRAYLIST_DATE_TYPE  = new TypeToken<ArrayList<Date>>() { }.getType();
   private static final Type MAP_STRING_STRING_TYPE = new TypeToken<Map<String, String>>() { }.getType();
 
@@ -112,6 +111,9 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
 
   @Override
   public void prepareRun(BatchSourceContext context) throws Exception {
+    //SimpleDateFormat needs to be local because it is not threadsafe
+    SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
+
     //calculate date one minute ago, rounded down to the nearest minute
     prevMinute = new Date(context.getLogicalStartTime() - TimeUnit.MINUTES.toMillis(1));
     Calendar cal = Calendar.getInstance();
@@ -149,7 +151,7 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
       conf.set(LAST_TIME_READ, datesToRead);
     }
 
-    conf.set(CUTOFF_READ_TIME, DATE_FORMAT.format(prevMinute));
+    conf.set(CUTOFF_READ_TIME, dateFormat.format(prevMinute));
     if (config.inputFormatClass != null) {
       ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
       Class<? extends FileInputFormat> classType = (Class<? extends FileInputFormat>)
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
@@ -54,9 +54,9 @@ public class BatchFileFilter extends Configured implements PathFilter {
   private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
   //length of 'YYYY-MM-dd-HH-mm"
   private static final int DATE_LENGTH = 16;
-  private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
   private static final Gson GSON = new Gson();
   private static final Type ARRAYLIST_DATE_TYPE  = new TypeToken<ArrayList<Date>>() { }.getType();
+  private final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
   private boolean useTimeFilter;
   private Pattern regex;
   private String pathName;
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
@@ -1,0 +1,228 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.template.etl.transform;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.templates.plugins.PluginConfig;
+import co.cask.cdap.template.etl.api.Emitter;
+import co.cask.cdap.template.etl.api.Transform;
+import co.cask.cdap.template.etl.api.TransformContext;
+import net.sf.uadetector.ReadableUserAgent;
+import net.sf.uadetector.UserAgentStringParser;
+import net.sf.uadetector.service.UADetectorServiceFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.ByteBuffer;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/**
+ * Parses log formats to extract access information
+ */
+@Plugin(type = "transform")
+@Name("LogParser")
+@Description("Parses logs from any input source for relevant information such as URI, IP, Browser, Device, and " +
+  "Timestamp.")
+public class LogParserTransform extends Transform<StructuredRecord, StructuredRecord> {
+  private static final Schema LOG_SCHEMA = Schema.recordOf(
+    "event",
+    Schema.Field.of("uri", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("ip", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("browser", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("device", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("ts", Schema.of(Schema.Type.LONG))
+  );
+  private static final String LOG_FORMAT_DESCRIPTION = "Log format to parse. Currently only supports S3 and " +
+    "CLF Log format.";
+  private static final String INPUT_NAME_DESCRIPTION = "Name of the field in the input schema which encodes the " +
+    "log information. The given field must be of type String or Bytes";
+  private static final Logger LOG = LoggerFactory.getLogger(LogParserTransform.class);
+  //Regex used to parse a CLF log, each field is commented above
+  private static final Pattern CLF_LOG_PATTERN = Pattern.compile(
+    //   IP                    id    user      date          request     code     size    referrer    user agent
+    "^([\\d.]+|[:][:][\\d]) (\\S+) (\\S+) \\[([^\\]]+)\\] \"([^\"]+)\" (\\d{3}) ([-\"\\d]+) \"([^\"]+)\" \"([^\"]+)\"");
+  //Regex used to parse a S3 log, each field is commented above
+  private static final Pattern S3_LOG_PATTERN = Pattern.compile(
+    // bucket owner name   time           ip                    req   reqID operation  key    request
+    "^(\\S+) (\\S+) \\[(\\p{Print}+)\\] ([\\d.]+|[:][:][\\d]) (\\S+) (\\S+) (\\S+) (\\S+) \"([^\"]+)\" " +
+      //  HTTP stat error code  bytes sent     obj size  time    turn time  referrer         user agent
+      "(\\d{3}) (\\p{Print}+) ([-\"\\d]+) ([-\"\\d]+) ([\\d]+) ([-\"\\d]+) \"(\\p{Print}+)\" \"([^\"]+)\" " +
+      //  version id
+      "(\\p{Print}+)");
+  //Regex used to parse the request field for the URI
+  private static final Pattern REQUEST_PAGE_PATTERN = Pattern.compile("(\\S+)\\s(\\S+).*");
+  //Indices of which group request, time, ip, and user agent are in the S3 regex
+  private static final int[] S3_INDICES = {9, 3, 4, 17};
+  //Indices of which group request, time, ip, and user agent are in the CLF regex
+  private static final int[] CLF_INDICES = {5, 4, 1, 9};
+  //Number of groups matched in the S3 regex
+  private static final int S3_REGEX_LENGTH = 18;
+  //Number of groups matched in the CLF regex
+  private static final int CLF_REGEX_LENGTH = 9;
+  private static final String S3_LOG = "S3";
+  private static final String CLF_LOG = "CLF";
+  private final LogParserConfig config;
+  private final SimpleDateFormat sdf = new SimpleDateFormat("dd/MMM/yyyy:HH:mm:ss Z");
+
+  public LogParserTransform(LogParserConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void initialize(TransformContext context) throws Exception {
+    if (!S3_LOG.equals(config.logFormat) && !CLF_LOG.equals(config.logFormat)) {
+      LOG.error("Log format not currently supported.");
+      throw new IllegalStateException("Unsupported log format");
+    }
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
+    String log = getLog(input);
+    if (log == null) {
+      LOG.error("Couldn't read schema, log message was null");
+      return;
+    }
+
+    StructuredRecord output;
+    if (S3_LOG.equals(config.logFormat)) {
+      Matcher logMatcher = S3_LOG_PATTERN.matcher(log);
+      if (!logMatcher.matches() || logMatcher.groupCount() < S3_REGEX_LENGTH) {
+        LOG.debug("Couldn't parse log because log did not match the S3 format, log: {}", log);
+        return;
+      }
+      output = parseRequest(logMatcher, S3_INDICES);
+    } else {
+      Matcher logMatcher = CLF_LOG_PATTERN.matcher(log);
+      if (!logMatcher.matches() || logMatcher.groupCount() < CLF_REGEX_LENGTH) {
+        LOG.debug("Couldn't parse log because the log did not match the CLF format. log: {}", log);
+        return;
+      }
+      output = parseRequest(logMatcher, CLF_INDICES);
+    }
+
+    if (output != null) {
+      emitter.emit(output);
+    }
+  }
+
+  /**
+   * Gets the log message from the input
+   * @param input the StructuredRecord to extract the log from
+   * @return the log message, or null on failure
+   */
+  @Nullable
+  private String getLog(StructuredRecord input) {
+    Schema.Field inputField = input.getSchema().getField(config.inputName);
+    if (inputField == null) {
+      LOG.debug("Invalid inputName, no known inputField matches given input of " + config.inputName);
+      return null;
+    }
+
+    Schema inputSchema = inputField.getSchema();
+    if (inputSchema.isNullableSimple()) {
+      inputSchema = inputSchema.getNonNullable();
+    }
+    Schema.Type inputType = inputSchema.getType();
+
+    if (!Schema.Type.STRING.equals(inputType) && !Schema.Type.BYTES.equals(inputType)) {
+      LOG.error("Unsupported inputType in schema, only Schema.Type.BYTES and Schema.Type.STRING are supported. " +
+                  "InputType: {}", inputType.toString());
+      return null;
+    }
+
+    if (Schema.Type.STRING.equals(inputType)) {
+      return input.get(config.inputName);
+    } else {
+      Object data = input.get(config.inputName);
+      if (data instanceof ByteBuffer) {
+        return Bytes.toString((ByteBuffer) input.get(config.inputName));
+      } else if (data instanceof byte[]) {
+        return Bytes.toString((byte[]) input.get(config.inputName));
+      } else {
+        LOG.debug("Not a byte type, type is {}", data.getClass().toString());
+        return null;
+      }
+    }
+  }
+
+  /**
+   * Parses a request for the URI, IP, Browser, Device, and Time
+   * @param logMatcher the regex matcher to use
+   * @param indices array of indices that define what position in the regex the fields are,
+   *                 in the order of Request, Time, IP, and User Agent
+   */
+  @Nullable
+  private StructuredRecord parseRequest(Matcher logMatcher, int[] indices) {
+    String request = logMatcher.group(indices[0]);
+    Matcher requestMatcher = REQUEST_PAGE_PATTERN.matcher(request);
+    if (!requestMatcher.matches() || requestMatcher.groupCount() < 2) {
+      LOG.debug("Couldn't parse uri because request does not match request pattern, request: {}", request);
+      return null;
+    }
+
+    String uri = requestMatcher.group(2);
+    long ts = System.currentTimeMillis();
+    try {
+      ts = sdf.parse(logMatcher.group(indices[1])).getTime();
+    } catch (ParseException e) {
+      LOG.debug("Couldn't parse time from the input record, using current timestamp instead. Exception: {}",
+                e.getMessage());
+    }
+
+    String ip = logMatcher.group(indices[2]);
+    UserAgentStringParser parser = UADetectorServiceFactory.getResourceModuleParser();
+    ReadableUserAgent userAgent = parser.parse(logMatcher.group(indices[3]));
+    String browser = userAgent.getFamily().getName();
+    String device = userAgent.getDeviceCategory().getCategory().getName();
+
+    return StructuredRecord.builder(LOG_SCHEMA)
+      .set("uri", uri)
+      .set("ip", ip)
+      .set("browser", browser)
+      .set("device", device)
+      .set("ts", ts)
+      .build();
+  }
+
+  /**
+   * Config class for LogParserTransform
+   */
+  public static class LogParserConfig extends PluginConfig {
+    @Name("logFormat")
+    @Description(LOG_FORMAT_DESCRIPTION)
+    private String logFormat;
+
+    @Name("inputName")
+    @Description(INPUT_NAME_DESCRIPTION)
+    private String inputName;
+
+    public LogParserConfig(String logFormat, String inputName) {
+      this.logFormat = logFormat;
+      this.inputName = inputName;
+    }
+  }
+}
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.template.etl.transform;
+
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.template.etl.api.Transform;
+import co.cask.cdap.template.etl.common.MockEmitter;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+
+public class LogParserTransformTest {
+  private static final Schema STRING_SCHEMA = Schema.recordOf(
+    "event",
+    Schema.Field.of("body", Schema.of(Schema.Type.STRING))
+  );
+  private static final Schema BYTE_SCHEMA = Schema.recordOf(
+    "event",
+    Schema.Field.of("body", Schema.of(Schema.Type.BYTES))
+  );
+  private static final LogParserTransform.LogParserConfig S3_CONFIG =
+    new LogParserTransform.LogParserConfig("S3", "body");
+  private static final Transform<StructuredRecord, StructuredRecord> S3_TRANSFORM = new LogParserTransform(S3_CONFIG);
+  private static final LogParserTransform.LogParserConfig CLF_CONFIG =
+    new LogParserTransform.LogParserConfig("CLF", "body");
+  private static final Transform<StructuredRecord, StructuredRecord> CLF_TRANSFORM =
+    new LogParserTransform(CLF_CONFIG);
+
+  @Test
+  public void transformS3Log() throws Exception {
+    StructuredRecord botRecord = StructuredRecord.builder(STRING_SCHEMA)
+      .set("body", "13a9f69e4a00effd6b4b891dcbcabef632ef9a9da7c localhost " +
+        "[22/Jan/2015:11:03:21 +0000] 122.122.111.11 - 6006CA0AE4 REST.GET.OBJECT " +
+        "ubuntu/this/is/some/folder " +
+        "\"GET /my/uri.gif releases/dists/precise/releases/i18n/Translation-en HTTP/1.1\" " +
+        "403 AccessDenied 231 - 10 - \"-\" \"Debian APT-HTTP/1.3 (0.8.16~exp12ubuntu10.  17)\" -")
+      .build();
+
+    StructuredRecord browserRecord = StructuredRecord.builder(STRING_SCHEMA)
+      .set("body", "13a9f69e4a00effd6b4b891dcef632ef9afe38cc8b0 localhost " +
+             "[31/Jan/2015:21:57:57 +0000] 133.133.133.133 - 0E94306589 REST.GET.OBJECT " +
+             "downloads/this/is/another/folder/with/a/file/file.zip " +
+             "\"GET /my/uri.jpg HTTP/1.1\" 304 - - 195750039 198 - " +
+             "\"-\" \"Mozilla/5.0 Gecko/20100115 Firefox/3.6\" -")
+      .build();
+
+    MockEmitter<StructuredRecord> emitter = new MockEmitter<>();
+    S3_TRANSFORM.transform(botRecord, emitter);
+    StructuredRecord output = emitter.getEmitted().get(0);
+    Assert.assertEquals("/my/uri.gif", output.get("uri"));
+    Assert.assertEquals("122.122.111.11", output.get("ip"));
+    Assert.assertEquals("unknown", output.get("browser"));
+    Assert.assertEquals("", output.get("device"));
+    Assert.assertEquals(1421924601000L, output.get("ts"));
+
+    S3_TRANSFORM.transform(browserRecord, emitter);
+    output = emitter.getEmitted().get(1);
+    Assert.assertEquals("/my/uri.jpg", output.get("uri"));
+    Assert.assertEquals("133.133.133.133", output.get("ip"));
+    Assert.assertEquals("Firefox", output.get("browser"));
+    Assert.assertEquals("Personal computer", output.get("device"));
+    Assert.assertEquals(1422741477000L, output.get("ts"));
+  }
+
+  @Test
+  public void transformCLFLog() throws Exception {
+    StructuredRecord record = StructuredRecord.builder(BYTE_SCHEMA)
+      .set("body", ByteBuffer.wrap((Bytes.toBytes("127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] " +
+                                                    "\"GET /apache_pb.gif HTTP/1.0\" 200 2326 " +
+                                                    "\"http://www.example.com/start.html\" \"Mozilla/5.0 " +
+                                                    "(Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 " +
+                                                    "(KHTML, like Gecko) Chrome/43.0.2357.124 Safari/537.36\""))))
+      .build();
+
+    MockEmitter<StructuredRecord> emitter = new MockEmitter<>();
+    CLF_TRANSFORM.transform(record, emitter);
+    StructuredRecord output = emitter.getEmitted().get(0);
+    Assert.assertEquals("/apache_pb.gif", output.get("uri"));
+    Assert.assertEquals("127.0.0.1", output.get("ip"));
+    Assert.assertEquals("Chrome", output.get("browser"));
+    Assert.assertEquals("Personal computer", output.get("device"));
+    Assert.assertEquals(971211336000L, output.get("ts"));
+  }
+}
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
@@ -190,6 +190,11 @@
       <artifactId>parquet-avro</artifactId>
       <version>1.6.0</version>
     </dependency>
+    <dependency>
+      <groupId>net.sf.uadetector</groupId>
+      <artifactId>uadetector-resources</artifactId>
+      <version>2014.04</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/cdap-ui/templates/common/LogParser.json
+++ b/cdap-ui/templates/common/LogParser.json
@@ -1,0 +1,21 @@
+{
+  "id": "LogParser",
+  "groups" : {
+    "position": [ "group1" ],
+    "group1": {
+      "display" : "Log Parser Transform",
+      "position" : [ "logFormat", "inputName" ],
+      "fields" : {
+        "logFormat": {
+          "widget": "textbox",
+          "label": "Log Format"
+        },
+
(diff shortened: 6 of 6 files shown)


bdmogal:
Before reviewing the code, let's first make sure that the library you're using is Apache Licensed. If not, we should probably find another.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -39,7 +39,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
   //length of 'YYYY-MM-dd-HH-mm"
   private static final int DATE_LENGTH = 16;
-  private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
```
anwar6953: why not final?

jayfjin: SimpleDateFormat is not thread safe

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+import java.text.SimpleDateFormat;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Parses log formats to extract access information
+ */
+@Plugin(type = "transform")
+@Name("LogParser")
+@Description("Parses logs for relevant information.")
```
bdmogal: There should be a lot more relevant information in the description. This field is exposed to users. Describe the kinds of input supported, etc

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+    "^([\\d.]+|[:][:][\\d]) (\\S+) (\\S+) \\[([^\\]]+)\\] \"([^\"]+)\" (\\d{3}) ([-\"\\d]+) \"([^\"]+)\" \"([^\"]+)\"");
+  private static final Pattern S3_LOG_PATTERN = Pattern.compile(
+    // bucket owner name   time           ip                    req   reqID operation  key     URI
+    "^(\\S+) (\\S+) \\[(\\p{Print}+)\\] ([\\d.]+|[:][:][\\d]) (\\S+) (\\S+) (\\S+) (\\S+) \"([^\"]+)\" " +
+      //  HTTP stat error code  bytes sent     obj size  time    turn time  referrer         user agent
+      "(\\d{3}) (\\p{Print}+) ([-\"\\d]+) ([-\"\\d]+) ([\\d]+) ([-\"\\d]+) \"(\\p{Print}+)\" \"([^\"]+)\" " +
+      //  version id
+      "(\\p{Print}+)");
+  private static final Pattern REQUEST_PAGE_PATTERN = Pattern.compile("(\\S+)\\s(\\S+).*");
+  private static final int[] S3_INDICES = {9, 3, 4, 17};
```
bdmogal: describe these static constants. especially ``REQUEST_PAGE_PATTERN``, ``S3_INDICES``, ``CLF_INDICES``, ``S3_REGEX_LENGTH``, ``CLF_REGEX_LENGTH``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+    this.config = config;
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
+    String log;
+    if ("string".equals(config.inputType)) {
+      log = input.get(config.inputName);
+    } else if ("bytes".equals(config.inputType)) {
+      log = Bytes.toString((ByteBuffer) input.get(config.inputName));
```
bdmogal: why cast to ``ByteBuffer``?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+  private final LogParserConfig config;
+
+  public LogParserTransform(LogParserConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
+    String log;
+    if ("string".equals(config.inputType)) {
```
bdmogal: Can you have records with different ``inputType`` in the same run? If so, you should move this parsing to the ``initialize`` method.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
jayfjin: That is the type that is outputted by a stream.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
jayfjin: No, each run must be the same inputType.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
bdmogal: Hmmm, so this transform can only be applied to a stream source? Or something that sends a ``ByteBuffer``. If so, that should be documented in the description.

However, I think it would be better if you could

```
Object data = input.get(config.inputName);
if (data instanceof ByteBuffer) {
  // cast to ByteBuffer
} else if (data instanceof byte[]) {
  // cast to byte[]
}
```

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+      log = input.get(config.inputName);
+    } else if ("bytes".equals(config.inputType)) {
+      log = Bytes.toString((ByteBuffer) input.get(config.inputName));
+    } else {
+      LOG.error("Unsupported inputType");
+      return;
+    }
+
+    if (log == null) {
+        LOG.error("Couldn't read schema");
```
bdmogal: incorrect spacing for this and the next line. Also, the error message is not sufficient.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
+    String log;
+    if ("string".equals(config.inputType)) {
+      log = input.get(config.inputName);
+    } else if ("bytes".equals(config.inputType)) {
+      log = Bytes.toString((ByteBuffer) input.get(config.inputName));
+    } else {
+      LOG.error("Unsupported inputType");
```
bdmogal: Also specify what is supported in the error message

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+      return;
+    }
+
+    if (log == null) {
+        LOG.error("Couldn't read schema");
+        return;
+    }
+
+    if ("S3".equals(config.logFormat)) {
```
bdmogal: You probably already have a constant for S3?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+      Matcher logMatcher = S3_LOG_PATTERN.matcher(log);
+      if (!logMatcher.matches() || logMatcher.groupCount() < S3_REGEX_LENGTH) {
+        LOG.error("Couldn't parse log " + log);
```
bdmogal: these are input errors. shouldn't be error level. maybe debug

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+        return;
+      }
+      parseRequest(emitter, logMatcher, S3_INDICES);
+    } else if ("CLF".equals(config.logFormat)) {
+      Matcher logMatcher = CLF_LOG_PATTERN.matcher(log);
+      if (!logMatcher.matches() || logMatcher.groupCount() < CLF_REGEX_LENGTH) {
+        LOG.error("Couldn't parse log " + log);
```
bdmogal: debug

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+        return;
+      }
+      parseRequest(emitter, logMatcher, CLF_INDICES);
+    } else {
```
bdmogal: If you parse the log format in ``initialize`` and throw an ``Exception`` if it is not one of the supported formats, you won't need this ``else``.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
bdmogal: yup, this logic needs to move to ``initialize`` then. Shouldn't do this parsing for every record.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
bdmogal: Also, more descriptive message please, because parsing could fail in the ``parseRequest`` method too, right? :)

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  /**
+   * Parses a request for the URI, IP, Browser, Device, and Time
+   * @param emitter the emitter to emit parsed information to
+   * @param logMatcher the regex matcher to use
+   * @param indices array of indices that define what position in the regex the fields are,
+   *                 in the order of Request, Time, IP, and User Agent
+   */
+  private void parseRequest(Emitter<StructuredRecord> emitter, Matcher logMatcher, int[] indices) {
```
bdmogal: I think it would be better for this method to return ``StructuredRecord``. ``emitter.emit()`` seems better suited to the ``transform`` method. Otherwise, this method's contract is not just *parse*, but also *emit*

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+    String request = logMatcher.group(indices[0]);
+    long ts = System.currentTimeMillis();
+    Matcher requestMatcher = REQUEST_PAGE_PATTERN.matcher(request);
+    if (!requestMatcher.matches() || requestMatcher.groupCount() < 2) {
+      LOG.error("Couldn't parse uri " + request);
```
bdmogal: debug. also, state why you couldn't parse.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+      return;
+    }
+
+    String uri = requestMatcher.group(2);
+    try {
+      //Needs to be local because SimpleDateFormat is not thread safe
+      SimpleDateFormat sdf = new SimpleDateFormat("dd/MMM/yyyy:HH:mm:ss Z");
+      ts = sdf.parse(logMatcher.group(indices[1])).getTime();
+    } catch (ParseException e) {
+      LOG.info("Couldn't parse time, using ts of System.currentTimeMillis()");
```
bdmogal: debug. 'Couldn't parse time from the input record, using current timestamp instead.'?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    String ip = logMatcher.group(indices[2]);
+    UserAgentStringParser parser = UADetectorServiceFactory.getResourceModuleParser();
+    ReadableUserAgent userAgent = parser.parse(logMatcher.group(indices[3]));
+    String browser = userAgent.getFamily().getName();
+    String device = userAgent.getDeviceCategory().getCategory().getName();
+
+    StructuredRecord output = StructuredRecord.builder(LOG_SCHEMA)
+      .set("uri", uri)
```
bdmogal: upto you, but should these fields be constants?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+  private static final Schema SCHEMA = Schema.recordOf(
+    "event",
+    Schema.Field.of("body", Schema.of(Schema.Type.STRING))
+  );
+
+  @Test
+  public void transformS3Log() throws Exception {
+    StructuredRecord record = StructuredRecord.builder(SCHEMA)
+      .set("body", "13a9f69e4a00effd6b4b891dcbcabef632ef9afe38cc8b0a9da7caf6fec8c152 repository.cask.co " +
+        "[22/Jan/2015:11:03:21 +0000] 70.42.248.20 - 60068D4CA00DCAE4 REST.GET.OBJECT " +
```
bdmogal: are these known ip addresses? If so, please mask them with arbitrary numbers

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+public class LogParserTransformTest {
+  private static final Schema SCHEMA = Schema.recordOf(
+    "event",
+    Schema.Field.of("body", Schema.of(Schema.Type.STRING))
+  );
+
+  @Test
+  public void transformS3Log() throws Exception {
+    StructuredRecord record = StructuredRecord.builder(SCHEMA)
+      .set("body", "13a9f69e4a00effd6b4b891dcbcabef632ef9afe38cc8b0a9da7caf6fec8c152 repository.cask.co " +
```
bdmogal: mask hostname. never have actual hostnames or ip addresses in code, logs, PRs or jiras.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+        "[22/Jan/2015:11:03:21 +0000] 70.42.248.20 - 60068D4CA00DCAE4 REST.GET.OBJECT " +
+        "ubuntu/precise/amd64/releases/dists/precise/releases/i18n/Translation-en " +
+        "\"GET /ubuntu/precise/amd64/      releases/dists/precise/releases/i18n/Translation-en HTTP/1.1\" " +
+        "403 AccessDenied 231 - 10 - \"-\" \"Debian APT-HTTP/1.3 (0.8.16~exp12ubuntu10.  17)\" -")
+      .build();
+
+    StructuredRecord record1 = StructuredRecord.builder(SCHEMA)
+      .set("body", "13a9f69e4a00effd6b4b891dcbcabef632ef9afe38cc8b0a9da7caf6fec8c152 repository.cask.co " +
```
bdmogal: mask hostname

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+             "[31/Jan/2015:21:57:57 +0000] 159.226.251.229 - 0E9AF143067C1589 REST.GET.OBJECT " +
```
bdmogal: mask ip address

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+  public void transformS3Log() throws Exception {
+    StructuredRecord record = StructuredRecord.builder(SCHEMA)
+      .set("body", "13a9f69e4a00effd6b4b891dcbcabef632ef9afe38cc8b0a9da7caf6fec8c152 repository.cask.co " +
+        "[22/Jan/2015:11:03:21 +0000] 70.42.248.20 - 60068D4CA00DCAE4 REST.GET.OBJECT " +
+        "ubuntu/precise/amd64/releases/dists/precise/releases/i18n/Translation-en " +
+        "\"GET /ubuntu/precise/amd64/      releases/dists/precise/releases/i18n/Translation-en HTTP/1.1\" " +
+        "403 AccessDenied 231 - 10 - \"-\" \"Debian APT-HTTP/1.3 (0.8.16~exp12ubuntu10.  17)\" -")
+      .build();
+
+    StructuredRecord record1 = StructuredRecord.builder(SCHEMA)
```
bdmogal: nit: what's the difference between these two input records? It would be nicer if the difference is called out clearly in the variable name.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+      .set("body", "13a9f69e4a00effd6b4b891dcbcabef632ef9afe38cc8b0a9da7caf6fec8c152 repository.cask.co " +
+             "[31/Jan/2015:21:57:57 +0000] 159.226.251.229 - 0E9AF143067C1589 REST.GET.OBJECT " +
+             "downloads/co/cask/cdap/cdap-sdk/2.5.2/cdap-sdk-2.5.2.zip " +
+             "\"GET /downloads/co/cask/cdap/cdap-sdk/2.5.2/cdap-sdk-2.5.2.zip HTTP/1.1\" 304 - - 195750039 198 - " +
+             "\"-\" \"Mozilla/5.0 Gecko/20100115 Firefox/3.6\" -")
+      .build();
+    LogParserTransform.LogParserConfig config = new LogParserTransform.LogParserConfig("S3", "string", "body");
```
bdmogal: These should probably all be constants.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+    Transform<StructuredRecord, StructuredRecord> transform = new LogParserTransform(config);
+    transform.initialize(null);
```
bdmogal: why initialize with ``null``?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    MockEmitter<StructuredRecord> emitter = new MockEmitter<>();
+    transform.transform(record, emitter);
+    StructuredRecord output = emitter.getEmitted().get(0);
+    Assert.assertEquals("/ubuntu/precise/amd64/", output.get("uri"));
+    Assert.assertEquals("70.42.248.20", output.get("ip"));
```
bdmogal: mask ip address

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+    Assert.assertEquals("unknown", output.get("browser"));
+    Assert.assertEquals("", output.get("device"));
+    Assert.assertEquals(1421924601000L, output.get("ts"));
+
+    transform.transform(record1, emitter);
+    output = emitter.getEmitted().get(1);
+    Assert.assertEquals("/downloads/co/cask/cdap/cdap-sdk/2.5.2/cdap-sdk-2.5.2.zip", output.get("uri"));
+    Assert.assertEquals("159.226.251.229", output.get("ip"));
```
bdmogal: mask ip address

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
bdmogal: in fact, it may be better to find a 'test' CLF record from somewhere, as opposed to using an actual one.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+    Assert.assertEquals("Firefox", output.get("browser"));
+    Assert.assertEquals("Personal computer", output.get("device"));
+    Assert.assertEquals(1422741477000L, output.get("ts"));
+  }
+
+  @Test
+  public void transformCLFLog() throws Exception {
+    StructuredRecord record = StructuredRecord.builder(SCHEMA)
+      .set("body", ByteBuffer.wrap(Bytes.toBytes("::1 - - [29/Jun/2015:14:01:34 -0700] \"GET / HTTP/1.1\" 304 - " +
```
bdmogal: better if you can find a 'test' CLF log

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/transform/LogParserTransformTest.java
bdmogal: you could just use ``localhost`` or ``yourhost.yourdomain.com``

bdmogal:
Done with review

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
jayfjin: The log format is already parsed into a pattern in the static final constant. Isn't it not possible to do anything further since that requires the actual log?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+
+  /**
+   * Parses a request for the URI, IP, Browser, Device, and Time
+   * @param logMatcher the regex matcher to use
+   * @param indices array of indices that define what position in the regex the fields are,
+   *                 in the order of Request, Time, IP, and User Agent
+   */
+  private StructuredRecord parseRequest(Matcher logMatcher, int[] indices) {
+    String request = logMatcher.group(indices[0]);
+    long ts = System.currentTimeMillis();
```
anwar6953: I think it makes sense to have this at least 6 lines down, closer to the code block where it's used.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+  private static final String CLF_LOG = "CLF";
+  private final LogParserConfig config;
+
+  public LogParserTransform(LogParserConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
+    String log;
```
bdmogal: move these declarations closer to where they are used

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+    StructuredRecord output;
+    Schema.Field inputField = input.getSchema().getField(config.inputName);
+    if (inputField == null) {
+      LOG.error("Invalid inputName, no known field matches given input.");
```
bdmogal: better to add ``inputField`` to the message. Also, this is not ``error`` level. Make it ``debug``.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+      return;
+    }
+    Schema.Type inputType = inputField.getSchema().getType();
```
bdmogal: what if users specify the ``Type`` as nullable? In that case, it'll be a union of ``Type.NULL`` and ``Type.STRING``/``Type.BYTES``. There is a way to get the non-nullable type from a nullable. Should we worry about this?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+        log = Bytes.toString((ByteBuffer) input.get(config.inputName));
+      } else if (data instanceof byte[]) {
+        log = Bytes.toString((byte[]) input.get(config.inputName));
+      } else {
+        LOG.debug("Not a byte type");
+        return;
+      }
+    } else {
+      LOG.error("Unsupported inputType, only bytes and string are supported.");
+      return;
```
bdmogal: Just a code-style comment: Its better to have these conditions that ``return`` be called out in the beginning of the method, and as much of them together as possible. So would it be better to do the following right after line 100?
```
if (!Schema.Type.STRING.equals(inputType) && !Schema.Type.BYTES.equals(inputType)) {
  return;
```

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+    } else if (CLF_LOG.equals(config.logFormat)) {
+      Matcher logMatcher = CLF_LOG_PATTERN.matcher(log);
+      if (!logMatcher.matches() || logMatcher.groupCount() < CLF_REGEX_LENGTH) {
+        LOG.debug("Couldn't parse log because the log did not match the CLF format. log:" + log);
+        return;
+      }
+      output = parseRequest(logMatcher, CLF_INDICES);
+    } else {
+      LOG.error("Log format not currently supported.");
+      return;
```
bdmogal: again, an early ``if`` return bloc would be nice?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+  private static final String S3_LOG = "S3";
+  private static final String CLF_LOG = "CLF";
+  private final LogParserConfig config;
+
+  public LogParserTransform(LogParserConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
```
bdmogal: This method is fairly long. Could you break it up easily?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+
+  /**
+   * Parses a request for the URI, IP, Browser, Device, and Time
+   * @param logMatcher the regex matcher to use
+   * @param indices array of indices that define what position in the regex the fields are,
+   *                 in the order of Request, Time, IP, and User Agent
+   */
+  private StructuredRecord parseRequest(Matcher logMatcher, int[] indices) {
```
bdmogal: Since this can return ``null``, annotate with ``@Nullable``

bdmogal:
Mostly style comments remaining. Has this been tested on a cluster?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+    Matcher requestMatcher = REQUEST_PAGE_PATTERN.matcher(request);
+    if (!requestMatcher.matches() || requestMatcher.groupCount() < 2) {
+      LOG.debug("Couldn't parse uri because request does not match request pattern, request:" + request);
+      return null;
+    }
+
+    String uri = requestMatcher.group(2);
+    try {
+      //Needs to be local because SimpleDateFormat is not thread safe
+      SimpleDateFormat sdf = new SimpleDateFormat("dd/MMM/yyyy:HH:mm:ss Z");
```
chtyim: Make this a field. It's guaranteed no concurrent calls to the transform method over the same object instance will be made.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+      ts = sdf.parse(logMatcher.group(indices[1])).getTime();
+    } catch (ParseException e) {
+      LOG.debug("Couldn't parse time from the input record, using current timestamp instead.");
```
chtyim: Log the exception message as well (`e.getMessage()`), otherwise hard to tell what's wrong.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    String ip = logMatcher.group(indices[2]);
+    UserAgentStringParser parser = UADetectorServiceFactory.getResourceModuleParser();
+    ReadableUserAgent userAgent = parser.parse(logMatcher.group(indices[3]));
+    String browser = userAgent.getFamily().getName();
+    String device = userAgent.getDeviceCategory().getCategory().getName();
+
+    StructuredRecord output = StructuredRecord.builder(LOG_SCHEMA)
```
chtyim: Just `return StructuredRecord.builder(..)....build();`, no need to have an extra field.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Gets the log message from the input
+   * @param input the StructuredRecord to extract the log from
+   * @return the log message, or null on failure
+   */
+  @Nullable
+  private String getLogFromInput(StructuredRecord input) {
+    Schema.Field inputField = input.getSchema().getField(config.inputName);
+    if (inputField == null) {
+      LOG.debug("Invalid inputName, no known inputField matches given input");
```
anwar6953: Would it be worth it to also log the 'given input'?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    Schema inputSchema = inputField.getSchema();
+    if (inputSchema.isNullableSimple()) {
+      inputSchema = inputSchema.getNonNullable();
+    }
+    Schema.Type inputType = inputSchema.getType();
+
+    if (!Schema.Type.STRING.equals(inputType) && !Schema.Type.BYTES.equals(inputType)) {
+      LOG.error("Unsupported inputType, only bytes and string are supported.");
```
anwar6953: Would it be worth it to also log the inputType?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,220 @@
...
+    } else if (Schema.Type.STRING.equals(inputType)) {
+      return input.get(config.inputName);
+    } else if (Schema.Type.BYTES.equals(inputType)) {
+      Object data = input.get(config.inputName);
+      if (data instanceof ByteBuffer) {
+        return Bytes.toString((ByteBuffer) input.get(config.inputName));
+      } else if (data instanceof byte[]) {
+        return Bytes.toString((byte[]) input.get(config.inputName));
+      } else {
+        LOG.debug("Not a byte type");
```
anwar6953: Perhaps make the actual class of the object a part of the log message.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,220 @@
...
+    if (log == null) {
+      LOG.error("Couldn't read schema, log message was null");
+      return;
+    }
+
+    StructuredRecord output = null;
+    if (!S3_LOG.equals(config.logFormat) && !CLF_LOG.equals(config.logFormat)) {
+      LOG.error("Log format not currently supported.");
+      return;
+    } else if (S3_LOG.equals(config.logFormat)) {
```
bdmogal: no need for ``else if`` since the ``if`` condition above ``return``s

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+  }
+
+  /**
+   * Gets the log message from the input
+   * @param input the StructuredRecord to extract the log from
+   * @return the log message, or null on failure
+   */
+  @Nullable
+  private String getLogFromInput(StructuredRecord input) {
```
bdmogal: nit: ``getLogFromStructuredRecord``? Or just ``getLog(StructuredRecord input)``. ``FromInput`` makes it a little vague?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,220 @@
...
+    Schema.Field inputField = input.getSchema().getField(config.inputName);
+    if (inputField == null) {
+      LOG.debug("Invalid inputName, no known inputField matches given input");
+      return null;
+    }
+
+    Schema inputSchema = inputField.getSchema();
+    if (inputSchema.isNullableSimple()) {
```
bdmogal: what happens if the schema is not simple?

jayfjin: error will be logged because its not string or bytes

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
bdmogal: Also, good to mention here that  its "Unsupported type **in schema**"  (Also perhaps ``Schema.Type.BYTES`` (not just bytes) and ``Schema.Type.STRING`` (not just string))

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,220 @@
...
+      inputSchema = inputSchema.getNonNullable();
+    }
+    Schema.Type inputType = inputSchema.getType();
+
+    if (!Schema.Type.STRING.equals(inputType) && !Schema.Type.BYTES.equals(inputType)) {
+      LOG.error("Unsupported inputType, only bytes and string are supported.");
+      return null;
+    } else if (Schema.Type.STRING.equals(inputType)) {
```
bdmogal: no need for ``else``, since the ``if`` returns

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,220 @@
...
+      return input.get(config.inputName);
+    } else if (Schema.Type.BYTES.equals(inputType)) {
```
bdmogal: same here. no need for ``else``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
anwar6953: I would say to put the initial `if` condition:
`if (!Schema.Type.STRING.equals(inputType) && !Schema.Type.BYTES.equals(inputType)) {`
as the final `else` condition.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -112,6 +111,9 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
 
   @Override
   public void prepareRun(BatchSourceContext context) throws Exception {
+    //SimpleDateFormat needs to be local because it is not threadsafe
+    SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd-HH-mm");
```
bdmogal: I thought @chtyim had a comment to make this a ``final`` member? Or was it in another class. Maybe make the change here as well.

bdmogal: Actually, I take that back. is it only used in ``prepareRun``? Maybe leave as is then

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,222 @@
...
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
+    String log = getLog(input);
+    if (log == null) {
+      LOG.error("Couldn't read schema, log message was null");
+      return;
+    }
+
+    StructuredRecord output = null;
```
bdmogal: nit: move this 4 lines down. Also shouldn't need assignment to ``null``?

bdmogal:
Just 1 minor comment from me. LGTM otherwise. Presume its tested in all modes.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,222 @@
...
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
+    String log = getLog(input);
+    if (log == null) {
+      LOG.error("Couldn't read schema, log message was null");
+      return;
+    }
+
+    if (!S3_LOG.equals(config.logFormat) && !CLF_LOG.equals(config.logFormat)) {
```
chtyim: Shouldn't you do this check in the `initialize` phase instead of doing it again and again for every record?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,222 @@
...
+      inputSchema = inputSchema.getNonNullable();
+    }
+    Schema.Type inputType = inputSchema.getType();
+
+    if (!Schema.Type.STRING.equals(inputType) && !Schema.Type.BYTES.equals(inputType)) {
+      LOG.error("Unsupported inputType in schema, only Schema.Type.BYTES and Schema.Type.STRING are supported. " +
+                  "InputType: " + inputType.toString());
+      return null;
+    }
+    if (Schema.Type.STRING.equals(inputType)) {
```
chtyim: Add a new line before if.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/transform/LogParserTransform.java
```diff
@@ -0,0 +1,222 @@
...
+      LOG.debug("Couldn't parse uri because request does not match request pattern, request:" + request);
+      return null;
+    }
+
+    String uri = requestMatcher.group(2);
+    long ts = System.currentTimeMillis();
+    try {
+      ts = sdf.parse(logMatcher.group(indices[1])).getTime();
+    } catch (ParseException e) {
+      LOG.debug("Couldn't parse time from the input record, using current timestamp instead. Exception: " +
```
chtyim: When using Logger, don't use `+`, use the `{}` place holder instead.

```java
LOG.debug("..... Exception: {}", e.getMessage());
```

PR #3068, opened 2015-06-30
Fix/ui create adapter jserror

ajainarayanan:
Fixes jshint errors in cdap.

Cluster: http://nodeproxy1736-1000.dev.continuuity.net:9999/ns/default

jslint

### DIFF
--- a/cdap-ui/app/features/adapters/services/adapter-create-model.js
+++ b/cdap-ui/app/features/adapters/services/adapter-create-model.js
@@ -89,7 +89,6 @@ angular.module(PKG.name + '.feature.adapters')
         return defer.promise;
       }
       EventPipe.emit('showLoadingIcon', 'Saving adapter as draft');
-      var adapterDrafts = {};
       this.adapterDrafts[this.metadata.name] = {
         config: {
           metadata: this.metadata,
@@ -142,7 +141,7 @@ angular.module(PKG.name + '.feature.adapters')
         });
       }
       return errObj;
-    }
+    };
 
     Model.prototype.checkForValidRequiredField = function checkForValidRequiredField(plugin) {
       var i;
@@ -183,7 +182,7 @@ angular.module(PKG.name + '.feature.adapters')
       angular.forEach(source.properties, pruneProperties.bind(this, source));
       angular.forEach(sink.properties, pruneProperties.bind(this, sink));
 
-      for (i=0; i<trans.length; i++) {
+      for (let i=0; i<trans.length; i++) {
         angular.forEach(trans[i].properties, pruneProperties.bind(this, trans[i]));
         if (!trans[i].placeHolder) {
           delete trans[i]._backendProperties;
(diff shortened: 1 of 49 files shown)


cdap-ui/app/features/adapters/services/adapter-create-model.js
```diff
@@ -183,7 +182,7 @@ angular.module(PKG.name + '.feature.adapters')
       angular.forEach(source.properties, pruneProperties.bind(this, source));
       angular.forEach(sink.properties, pruneProperties.bind(this, sink));
 
-      for (i=0; i<trans.length; i++) {
+      for (let i=0; i<trans.length; i++) {
```
elfenheart: Isn't **let** an ES6 property?

cdap-ui/app/features/dashboard/services/chart-helpers.js
```diff
@@ -94,15 +96,15 @@ angular.module(PKG.name + '.feature.dashboard')
 
         // columns will be in the format: [ [metric1Name, v1, v2, v3, v4], [metric2Name, v1, v2, v3, v4], ... xCoords ]
         columns = [];
-        for (i = 0; i < newVal.length; i++) {
-          metricMap = newVal[i];
+        newVal.forEach(function (value) {
+          metricMap = value;
```
elfenheart: Do we need to assign value to metricMap? it seems like we can just directly call Object.keys on value. 

Minor performance enhancement :)

cdap-ui/app/features/workers/controllers/detail-ctrl.js
```diff
@@ -1,4 +1,4 @@
 angular.module(PKG.name + '.feature.worker')
-  .controller('WorkersDetailController', function($scope) {
+  .controller('WorkersDetailController', function() {
```
elfenheart: Just delete this file?

cdap-ui/app/services/my-promise.js
```diff
@@ -61,13 +61,17 @@ angular.module(PKG.name + '.services')
...
           }
           fulfill(result);
-        } catch (e) {
-          (console.error)? console.error(ex): console.log(ex);
-          reject(e);
+        } catch (ex) {
+          if (console.error) {
+            console.error(ex);
+          } else {
+             console.log(ex);
```
elfenheart: seems to be different spacing?

cdap-ui/app/services/my-promise.js
```diff
@@ -106,7 +110,11 @@ angular.module(PKG.name + '.services')
               try {
                 return resolve(onFulfilled(result));
               } catch (ex) {
-                (console.error)? console.error(ex): console.log(ex);
+                if (console.error) {
+                  console.error(ex);
+                } else {
+                   console.log(ex);
```
elfenheart: same comment about spacing

astewes:
LGTM

cdap-ui/app/features/adapters/services/adapter-create-model.js
ajainarayanan: This should have been fine on a production build since we run babel through the entire code when we do gulp build. But for development this would have been issue and since I didn't create an adapter I didn't notice this. Good catch! Fixed.

elfenheart:
:+1: LGTM

PR #3069, opened 2015-06-30
Adding Configuration string to the Application Specification

gokulavasan:
* Add the configuration string, which was used to create the application, to the Application Specification.

Build : http://builds.cask.co/browse/CDAP-DUT2136


cdap-app-fabric/src/main/java/co/cask/cdap/app/ApplicationSpecification.java
```diff
@@ -50,6 +50,12 @@
   String getVersion();
 
   /**
+   * @return Configuraiton string used during the creation of the Application.
```
albertshau: typo, Configuration

albertshau:
typo, then lgtm

gokulavasan:
Addressed comment. Build passed. Merging.

PR #3071, opened 2015-07-01
fix app.template.* to match cdap-default.xml

dereklwood:
updating cdap-site.xml appendix in release branch docs:
   - [x] fixing default value for ``app.template.dir``
   - [x] adding ``app.template.plugin.dir``


wolf31o2:
:+1: LGTM

PR #3075, opened 2015-07-01
CDAP-2747 - Use IndexedTable for PartitionedFileSet.

anwar6953:
Index partitions on transaction id, to allow consumption of partitions as they are created/committed, along with a simple consumer which holds its state in memory.

https://issues.cask.co/browse/CDAP-2747
http://builds.cask.co/browse/CDAP-DUT2124-21

This is based off the design here:
https://wiki.cask.co/display/CE/Periodically+consuming+partitions+of+a+PartitionedFileSetDataset

The design (API and implementation) for the client side to this functionality (likely a periodically-running workflow) has yet to be finalized.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.dataset.lib;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+import com.google.common.primitives.Longs;
+
+import java.nio.ByteBuffer;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Contains the state necessary to keep track of which partitions are processed and which partitions would need to be
+ * processed as they are created.
+ */
+public class PartitionConsumerState {
+  // useful on initial query of partitions
+  public static final PartitionConsumerState FROM_BEGINNING =
+    new PartitionConsumerState(0, Collections.<Long>emptyList());
+
+  // Read pointer of the transaction from the previous query of partitions. This is used to scan for new partitions
+  // created since then.
+  private final long startVersion;
+  // The list of in progress transactions from the previous query of partitions that are smaller than the startVersion.
+  // We do not need to include the in-progress transaction Ids that are larger than the startVersion because those will
+  // be picked up in the next scan anyways, since we will start the scan from the startVersion.
+  // Keeping track of these in-progress transactions is necessary because these might be creations of partitions that
+  // fall before the startVersion.
+  private final List<Long> versionsToCheck;
+
+  public PartitionConsumerState(long startVersion, List<Long> versionsToCheck) {
+    Preconditions.checkNotNull(versionsToCheck);
+    this.startVersion = startVersion;
+    this.versionsToCheck = ImmutableList.copyOf(versionsToCheck);
+  }
+
+  public long getStartVersion() {
+    return startVersion;
+  }
+
+  public List<Long> getVersionsToCheck() {
+    return versionsToCheck;
+  }
+
+  public static PartitionConsumerState fromBytes(byte[] bytes) {
+    Preconditions.checkArgument((bytes.length - 1) % Longs.BYTES == 0,
+                                "bytes does not have length divisible by Longs.BYTES");
+    ByteBuffer bb = ByteBuffer.wrap(bytes);
+    byte serializationFormatVersion = bb.get();
+    Preconditions.checkArgument(serializationFormatVersion == 0,
+                                "Unsupported serialization format: {}", serializationFormatVersion);
+    long startVersion = bb.getLong();
+    List<Long> versionsToCheck = Lists.newArrayList();
+    while (bb.hasRemaining()) {
+      versionsToCheck.add(bb.getLong());
+    }
+    return new PartitionConsumerState(startVersion, versionsToCheck);
+  }
+
+  public byte[] toBytes() {
+    int numLongs = 1 + versionsToCheck.size();
+    // first byte for serialization format version
+    ByteBuffer bb = ByteBuffer.allocate(1 + Longs.BYTES * numLongs);
+    // currently, serialization format is 0
+    bb.put((byte) 0);
+    bb.putLong(startVersion);
+    for (long l : versionsToCheck) {
+      bb.putLong(l);
+    }
+    return bb.array();
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    PartitionConsumerState that = (PartitionConsumerState) o;
+
+    if (startVersion != that.startVersion) {
+      return false;
+    }
+    return versionsToCheck.equals(that.versionsToCheck);
+  }
+
+  @Override
+  public int hashCode() {
+    int result = (int) (startVersion ^ (startVersion >>> 32));
+    result = 31 * result + versionsToCheck.hashCode();
+    return result;
+  }
+}
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionConsumer.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionConsumer.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data2.dataset2.lib.partitioned;
+
+import co.cask.cdap.api.dataset.lib.Partition;
+import co.cask.cdap.api.dataset.lib.PartitionConsumerResult;
+import co.cask.cdap.api.dataset.lib.PartitionConsumerState;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
+
+import java.util.Iterator;
+
+/**
+ * A simple consumer for {@link Partition}s of a {@link PartitionedFileSet}
+ */
+public class PartitionConsumer {
+  private final PartitionedFileSet partitionedFileSet;
+  private PartitionConsumerState partitionConsumerState;
+
+  /**
+   * Creates an instance of PartitionConsumer which consumes, starting from the given PartitionConsumerState
+   *
+   * @param partitionedFileSet the PartitionedFileSet to consume from
+   * @param partitionConsumerState the PartitionConsumerState to begin consuming from
+   */
+  public PartitionConsumer(PartitionedFileSet partitionedFileSet,
+                           PartitionConsumerState partitionConsumerState) {
+    this.partitionedFileSet = partitionedFileSet;
+    this.partitionConsumerState = partitionConsumerState;
+  }
+
+  /**
+   * Creates an instance of a PartitionConsumer which begins consuming from the beginning.
+   *
+   * @param partitionedFileSet the PartitionedFileSet to consume from
+   */
+  public PartitionConsumer(PartitionedFileSet partitionedFileSet) {
+    this(partitionedFileSet, PartitionConsumerState.FROM_BEGINNING);
+  }
+
+  /**
+   * @return an iterator to {@link Partition}s of the underlying {@link PartitionedFileSet} created since the last
+   * call to this method. This excludes partitions created in in-progress transactions including the one in which the
+   * call to this method is made.
+   */
+  public Iterator<Partition> consumePartitions() {
+    PartitionConsumerResult partitionConsumerResult = partitionedFileSet.consumePartitions(partitionConsumerState);
+    partitionConsumerState = partitionConsumerResult.getPartitionConsumerState();
+    return partitionConsumerResult.getPartitionIterator();
+  }
+}
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
@@ -25,7 +25,10 @@ import co.cask.cdap.api.dataset.lib.AbstractDataset;
 import co.cask.cdap.api.dataset.lib.FileSet;
 import co.cask.cdap.api.dataset.lib.FileSetArguments;
 import co.cask.cdap.api.dataset.lib.FileSetProperties;
+import co.cask.cdap.api.dataset.lib.IndexedTable;
 import co.cask.cdap.api.dataset.lib.Partition;
+import co.cask.cdap.api.dataset.lib.PartitionConsumerResult;
+import co.cask.cdap.api.dataset.lib.PartitionConsumerState;
 import co.cask.cdap.api.dataset.lib.PartitionDetail;
 import co.cask.cdap.api.dataset.lib.PartitionFilter;
 import co.cask.cdap.api.dataset.lib.PartitionKey;
@@ -38,12 +41,12 @@ import co.cask.cdap.api.dataset.lib.Partitioning.FieldType;
 import co.cask.cdap.api.dataset.table.Put;
 import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Scanner;
-import co.cask.cdap.api.dataset.table.Table;
 import co.cask.cdap.explore.client.ExploreFacade;
 import co.cask.cdap.proto.Id;
 import co.cask.tephra.Transaction;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
@@ -56,8 +59,10 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Set;
 import javax.annotation.Nullable;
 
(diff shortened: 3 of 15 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Contains the state necessary to keep track of which partitions are processed and which partitions would need to be
+ * processed as they are created.
+ */
+public class PartitionConsumerState {
+  // useful on initial query of partitions
+  public static final PartitionConsumerState FROM_BEGINNING = new PartitionConsumerState(null, new long[0]);
+
+  // Write pointer of the transaction from the previous query of partitions. This is used to scan for new partitions
```
anew: this should be the upper bound of the read pointer instead.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
```diff
@@ -0,0 +1,49 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Contains the state necessary to keep track of which partitions are processed and which partitions would need to be
+ * processed as they are created.
+ */
+public class PartitionConsumerState {
+  // useful on initial query of partitions
+  public static final PartitionConsumerState FROM_BEGINNING = new PartitionConsumerState(null, new long[0]);
```
anew: this is useful. We probably also need a way to initialize one from the current time (or a current transaction), and one that starts at a certain point in time.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionConsumer.java
```diff
@@ -0,0 +1,63 @@
...
+import co.cask.cdap.api.dataset.lib.PartitionConsumerResult;
+import co.cask.cdap.api.dataset.lib.PartitionConsumerState;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
+
+import java.util.Iterator;
+
+/**
+ * A simple consumer for {@link Partition}s of a {@link PartitionedFileSet}
+ */
+public class PartitionConsumer {
```
anew: this would have to be extended to be able to persist its state

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
```diff
@@ -0,0 +1,49 @@
...
+    this.startVersion = startVersion;
+    this.versionsToCheck = versionsToCheck;
+  }
+
+  @Nullable
+  public Long getStartVersion() {
+    return startVersion;
+  }
+
+  public long[] getVersionsToCheck() {
```
anew: would it make sense to make this a set?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -170,6 +193,99 @@ protected void addPartition(PartitionKey key, String path, boolean explorable, M
   }
 
   @Override
+  public PartitionConsumerResult consumePartitions(PartitionConsumerState partitionConsumerState) {
+    long[] previousInProgress = partitionConsumerState.getVersionsToCheck();
+    long[] inProgress = tx.getInProgress();
+    Long[] noLongerInProgress = setDiff(previousInProgress, inProgress);
+
+    Iterator<Partition> noLongerInProgressPartitions = Iterators.emptyIterator();
+    if (noLongerInProgress.length > 0) {
```
anew: it may be better to do point lookups for each of the "no longer in progress". The index scan is relatively expensive because it scans everything in between, and fetches the primary row for everything in between.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -170,6 +193,99 @@ protected void addPartition(PartitionKey key, String path, boolean explorable, M
...
+                                                    Bytes.stopKeyForPrefix(Bytes.toBytes(lastElement)));
+      noLongerInProgressPartitions = new PartitionIterator(scanner, noLongerInProgress);
+    }
+
+    Long startVersion = partitionConsumerState.getStartVersion();
+    long readPointer = tx.getReadPointer();
+    // no read your own writes
+    Scanner scanner = partitionsTable.scanByIndex(PartitionedFileSetDefinition.INDEX_COLUMN,
+                                                  startVersion == null ? null : Bytes.toBytes(startVersion),
+                                                  Bytes.toBytes(readPointer + 1));
```
anew: can readpointer == writePointer? You want to use the minimum of both as stop of the scan.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -170,6 +193,99 @@ protected void addPartition(PartitionKey key, String path, boolean explorable, M
...
+    Iterator<Partition> partitionIterator = Iterators.concat(noLongerInProgressPartitions,
+                                                             new PartitionIterator(scanner, null));
+    return new PartitionConsumerResult(new PartitionConsumerState(tx.getWritePointer(), inProgress), partitionIterator);
```
anew: this should be readPointer, not writePointer

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -170,6 +193,99 @@ protected void addPartition(PartitionKey key, String path, boolean explorable, M
...
+  }
+
+  private Long[] setDiff(long[] oldLongs, long[] newLongs) {
```
anew: this method becomes unnecessary if you just keep them as sets

anew:
@anwar6953 I did one pass and overall it looks good. I have some comments, though. Also, have not paid too much attention to test cases yet.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: My understanding is that readPointer is always smaller than writePointer, because readPointer signifies the largest transactionId that has committed.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: I think that this should be writePointer because there are transaction Ids between readPointer and writePointer that are also represented in the inProgress list.
Because of this, if we perform the next scan from the current tx's readPointer, its possible to process some partitions twice.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
anwar6953: relevant:
https://github.com/caskdata/cdap/pull/3075#discussion_r33749421

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anew: I  think then we need to be smart and ignore the in-progress ones that are greater than the read pointer

anwar6953: Ok, that would work. I'm saving tx.getReadPointer + 1, then, because we can already process the readPointer in this scan.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: Good point. Updated to do point lookups for each of the transaction Ids.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
anwar6953: done.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/**
+ * Contains the state necessary to keep track of which partitions are processed and which partitions would need to be
+ * processed as they are created.
+ */
+public class PartitionConsumerState {
+  // useful on initial query of partitions
+  public static final PartitionConsumerState FROM_BEGINNING =
```
anwar6953: https://github.com/caskdata/cdap/pull/3075#discussion_r33746967

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionConsumer.java
anwar6953: I think that would be good to design + implement in the next iteration, as the consumer of this is built (mapreduce of workflow).

anwar6953: implemented `toBytes` and `fromBytes` method on this class for now, also.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/partitioned/TimePartitionedFileSetTest.java
```diff
@@ -102,73 +101,78 @@ public void apply() throws Exception {
   }
 
   @Test
-  public void testAddGetPartitions() throws IOException, ParseException, DatasetManagementException {
```
anwar6953: Note that the changes in this file are simply indentation (due to wrapping within transaction).

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
```diff
@@ -0,0 +1,103 @@
...
+    long startVersion = bb.getLong();
+    Set<Long> versionsToCheck = Sets.newHashSet();
+    while (bb.hasRemaining()) {
+      versionsToCheck.add(bb.getLong());
+    }
+    return new PartitionConsumerState(startVersion, versionsToCheck);
+  }
+
+  public byte[] toBytes() {
+    int numLongs = 1 + versionsToCheck.size();
```
anew: Can we add an additional byte at the beginning for the version of the serialization format? That can be 0 for now, but if we ever change this format, then we can bump it to 1, and have smart logic to deserialize the current or previous versions.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
```diff
@@ -0,0 +1,103 @@
...
+
+  // Read pointer of the transaction from the previous query of partitions. This is used to scan for new partitions
+  // created since then.
+  private final long startVersion;
+  // The list of in progress transactions from the previous query of partitions that are smaller than the read pointer.
+  // We do not need to include the in-progress transaction Ids that are larger than the read pointer because those will
+  // be picked up in the next scan anyways, since we will start the scan from the read pointer.
+  // Keeping track of these in-progress transactions is necessary because these might be creations of partitions that
+  // fall before the startVersion.
+  private final Set<Long> versionsToCheck;
```
anew: Is it a list or a set? Is it ordered? Seems that it may make sense to order it.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
```diff
@@ -0,0 +1,103 @@
...
+
+  public PartitionConsumerState(long startVersion, Set<Long> versionsToCheck) {
+    this.startVersion = startVersion;
```
anew: Add a precondition to make sure veresionsToCheck is never null?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionConsumer.java
```diff
@@ -0,0 +1,63 @@
...
+   *
+   * @param partitionedFileSet the PartitionedFileSet to consume from
+   */
+  public PartitionConsumer(PartitionedFileSet partitionedFileSet) {
+    this(partitionedFileSet, PartitionConsumerState.FROM_BEGINNING);
+  }
+
+  /**
+   * @return an iterator to {@link Partition}s of the underlying {@link PartitionedFileSet} created since the last
+   * call to this method.
```
anew: Maybe mention that this excludes all in progress transactions, including the current one.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -173,6 +195,84 @@ protected void addPartition(PartitionKey key, String path, boolean explorable, M
...
+    for (Long txId : noLongerInProgress) {
+      Scanner scanner = partitionsTable.readByIndex(WRITE_PTR_COL, Bytes.toBytes(txId));
+      partitionIterators.add(new PartitionIterator(scanner));
+    }
+
+    long readPointer = tx.getReadPointer();
+    // no read your own writes (partitions)
+    Scanner scanner = partitionsTable.scanByIndex(WRITE_PTR_COL,
+                                                  Bytes.toBytes(partitionConsumerState.getStartVersion()),
+                                                  Bytes.toBytes(readPointer + 1));
```
anew: I still believe that tephra makes no guarantees that the read pointer is less than the write version of a transaction. I think you need to take the minimum of (write version, read pointer + 1) as the upper bound (and also in the following loop).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -173,6 +195,84 @@ protected void addPartition(PartitionKey key, String path, boolean explorable, M
...
+      inProgressBeforeReadPointer.add(txId);
+    }
+    return new PartitionConsumerResult(new PartitionConsumerState(readPointer + 1, inProgressBeforeReadPointer),
+                                       Iterators.concat(partitionIterators.iterator()));
+  }
+
+  private Set<Long> setDiff(Set<Long> oldLongsSet, long[] newLongs) {
+    Set<Long> newLongsSet = Sets.newHashSet();
+    for (long newLong : newLongs) {
+      newLongsSet.add(newLong);
```
anew: why do you need newLongSet? Can't you just call oldLongSet.remove(newLong) here?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -173,6 +195,84 @@ protected void addPartition(PartitionKey key, String path, boolean explorable, M
...
+      if (txId >= readPointer) {
+        break;
+      }
+      inProgressBeforeReadPointer.add(txId);
+    }
+    return new PartitionConsumerResult(new PartitionConsumerState(readPointer + 1, inProgressBeforeReadPointer),
+                                       Iterators.concat(partitionIterators.iterator()));
+  }
+
+  private Set<Long> setDiff(Set<Long> oldLongsSet, long[] newLongs) {
```
anew: not entirely sure whether this does what it should... 
You want all versions that were in oldLongsSet but are not in newLongs. But do you want to alter oldLongSet?
I think you better make a copy of oldLongSet, then remove every element of newLongs from it, and return it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -173,6 +195,84 @@ protected void addPartition(PartitionKey key, String path, boolean explorable, M
   }
 
   @Override
+  public PartitionConsumerResult consumePartitions(PartitionConsumerState partitionConsumerState) {
+    Set<Long> previousInProgress = partitionConsumerState.getVersionsToCheck();
+    Set<Long> noLongerInProgress = setDiff(previousInProgress, tx.getInProgress());
```
anew: since partitionConsumerState does no defensive copying, and your setDiff() modifies previousInProgress, this effectively alters the old state.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: done

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDefinition.java
```diff
@@ -62,20 +64,27 @@
...
     Preconditions.checkArgument(filesetDef != null, "FileSet definition is required");
     this.filesetDef = filesetDef;
-    this.tableDef = tableDef;
+    this.indexedTableDef = indexedTableDef;
   }
 
   @Override
   public DatasetSpecification configure(String instanceName, DatasetProperties properties) {
+    // define the columns for indexing on the partitionsTable
+    String indexedColumns = Bytes.toString(PartitionedFileSetDataset.WRITE_PTR_COL) + ','
```
anew: this can be a static final?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: ok, I thought the readPointer was the largest tx that had committed at the time the current tx started?

anew:
Done with my second pass. Looks pretty good, except for a few comments. Nice job on the test cases -Andreas

anwar6953:
@anew Thanks; addressed comments.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
anwar6953: added

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anew: Thanks for making the change. I know that is the current implementation in tephra, but it is nowhere documented. That means it can change with the next version of tephra, but we would not know, because the logic here would simply become non-deterministic with regards to the current transaction, but we would never see an error.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionConsumerState.java
```diff
@@ -0,0 +1,111 @@
...
+  // We do not need to include the in-progress transaction Ids that are larger than the startVersion because those will
+  // be picked up in the next scan anyways, since we will start the scan from the startVersion.
+  // Keeping track of these in-progress transactions is necessary because these might be creations of partitions that
+  // fall before the startVersion.
+  private final List<Long> versionsToCheck;
+
+  public PartitionConsumerState(long startVersion, List<Long> versionsToCheck) {
+    Preconditions.checkNotNull(versionsToCheck);
+    this.startVersion = startVersion;
+    this.versionsToCheck = versionsToCheck;
```
anew: should do defensive copying here. ImmutableList.copyOf(versionsToCheck)

anew:
Nice work. LGTM except one comment (to copy the versionsToCheck in the PartitionConsumerState).

PR #3077, opened 2015-07-01
Feature/ui required fields

astewes:
* adds uniform styling for required field asterisks
* adds asterisks to required fields where an asterisk was previously missing

required


elfenheart:
Can you add this on Create Stream modal as well?

ajainarayanan:
@astewes Should we have the name and the icons in two separate ```span```s.
For instance,

```
<label class="col-xs-12 col-sm-2 control-label" 
           for="namespaceDisplayName"> 
    <span>Name </span>
    <span class="fa fa-asterisk"></span>
</label>
```

elfenheart:
:+1: LGTM

PR #3080, opened 2015-07-02
Fix handling of absolute paths in FileSet

anew:
- absolute base paths now are treated as absolute in the file system
- they may not be inside the cdap top level directory
- all existing tests and examples changed to use relative paths
- added new test cases for absolute paths


anew:
build passes: http://builds.cask.co/browse/CDAP-DUT2142-2

chtyim:
LGTM

PR #3083, opened 2015-07-02
CDAP-2913 EMR install script

wolf31o2:
This script does everything necessary to install CDAP on an Amazon EMR cluster.

- Setup YUM repository and install CDAP packages
- Install CDAP dependencies
- Create `kafka.log.dir` directory
- Configure CDAP
- Create `hdfs.namespace` HDFS directory
- Start CDAP services

### DIFF
--- a/cdap-distributions/src/emr/install.sh
+++ b/cdap-distributions/src/emr/install.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+#
+# Copyright © 2015 Cask Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+#
+# Install CDAP under hadoop user for EMR
+#
+
+# Fetch repo file for YUM
+curl http://repository.cask.co/centos/6/x86_64/cdap/3.0/cask.repo > /etc/yum.repos.d/cdap-3.0.repo
+rpm --import http://repository.cask.co/centos/6/x86_64/cdap/3.0/pubkey.gpg
+# Enable EPEL packages (req'd for nodejs)
+sed -i -e '0,/enabled=0/s//enabled=1/' /etc/yum.repos.d/epel.repo
+yum makecache
+# Install dependencies
+yum install --downloadonly chkconfig -y
+rpm --install --force /var/cache/yum/x86_64/latest/amzn-main/packages/chkconfig-*
+yum install gyp http-parser-devel redhat-rpm-config icu -y
+yum install nodejs nodejs-devel npm -y
+yum install -y cdap-{gateway,kafka,master,ui}
+# Update permissions and symlink logs
+mkdir -p /mnt/var/log/cdap /mnt/cdap/kafka-logs
+chown -R hadoop:hadoop /var/cdap/run /var/run/cdap /var/tmp/cdap /mnt/var/log/cdap /mnt/cdap/kafka-logs
+rm -rf /var/log/cdap
+ln -sf /mnt/var/log/cdap /var/log/cdap
+# Configure CDAP
+__ipaddr=`ifconfig eth0 | grep addr: | cut -d: -f2 | head -n 1 | awk '{print $1}'`
+sed \
+  -e 's/yarn/hadoop/' \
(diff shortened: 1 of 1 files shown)


cdap-distributions/src/emr/install.sh
```diff
@@ -0,0 +1,63 @@
...
+  -e 's/LOCAL-WATCHDOG-IP//' \
+  -e "s/ROUTER-HOST-IP/${__ipaddr}/" \
+  -e 's/router.server.port/router.bind.port/' \
+  -e 's/10000/11015/' \
+  /etc/cdap/conf/cdap-site.xml.example > /etc/cdap/conf/cdap-site.xml
+# Create HDFS directory
+su - hadoop -c "hadoop fs -mkdir -p /cdap"
+# Start services
+for i in \
+  /opt/cdap/gateway/bin/svc-router \
```
dereklwood: any reason not to use ``/etc/init.d/cdap-*``?  i thought it should still work if run as hadoop user?

cdap-distributions/src/emr/install.sh
```diff
@@ -0,0 +1,63 @@
...
+  /opt/cdap/kafka/bin/svc-kafka-server \
+  /opt/cdap/master/bin/svc-master \
+  /opt/cdap/ui/bin/svc-ui
+  do
+  su - hadoop -c "$i start"
+  done
```
dereklwood: minor: indentation

dereklwood:
comment/question, not blockers.  also would be nice to add some validation checks at the top.  :+1: lgtm

cdap-distributions/src/emr/install.sh
wolf31o2: It does. I just went with the service scripts directly, since we will not need to support Kerberos.

cdap-distributions/src/emr/install.sh
```diff
@@ -0,0 +1,63 @@
...
+rpm --import http://repository.cask.co/centos/6/x86_64/cdap/3.0/pubkey.gpg
+# Enable EPEL packages (req'd for nodejs)
+sed -i -e '0,/enabled=0/s//enabled=1/' /etc/yum.repos.d/epel.repo
+yum makecache
+# Install dependencies
+yum install --downloadonly chkconfig -y
+rpm --install --force /var/cache/yum/x86_64/latest/amzn-main/packages/chkconfig-*
+yum install gyp http-parser-devel redhat-rpm-config icu -y
+yum install nodejs nodejs-devel npm -y
+yum install -y cdap-{gateway,kafka,master,ui}
```
duddun: At the risk of sounding picky: I would have the '-y' right after 'yum install' for every command or at the end (for every command).

cdap-distributions/src/emr/install.sh
```diff
@@ -0,0 +1,64 @@
...
+# the License.
+
+#
+# Install CDAP under hadoop user for EMR
+#
+
+die() { echo "ERROR: ${*}"; exit 1; };
+
+# Fetch repo file for YUM
+curl http://repository.cask.co/centos/6/x86_64/cdap/3.0/cask.repo > /etc/yum.repos.d/cdap-3.0.repo
```
sreevatsanraman: Check for 200 after the call and fail if non 200

cdap-distributions/src/emr/install.sh
```diff
@@ -0,0 +1,64 @@
...
+rpm --import http://repository.cask.co/centos/6/x86_64/cdap/3.0/pubkey.gpg
```
sreevatsanraman: Check for status and bail on error

cdap-distributions/src/emr/install.sh
```diff
@@ -0,0 +1,64 @@
...
+  -e 's/LOCAL-WATCHDOG-IP//' \
+  -e "s/ROUTER-HOST-IP/${__ipaddr}/" \
+  -e 's/router.server.port/router.bind.port/' \
+  -e 's/10000/11015/' \
+  /etc/cdap/conf/cdap-site.xml.example > /etc/cdap/conf/cdap-site.xml
+# Create HDFS directory
+su - hadoop -c "hadoop fs -mkdir -p /cdap" || die "Cannot create /cdap in HDFS"
+# Start services
+for i in \
+  /opt/cdap/gateway/bin/svc-router \
```
sreevatsanraman: authserver is missing

wolf31o2: This is on purpose. The auth server requires configuration to be useful. We cannot possibly know this configuration. Instead, we stick with the Amazon choice of not having auth.

cdap-distributions/src/emr/install.sh
wolf31o2: This should never fail, unless DNS is down. These files are hosted in S3 and EMR hosts have direct attachment to S3. I've added error checking, but it's extremely unlikely to ever trigger, which is a good thing.

sreevatsanraman:
LGTM

PR #3084, opened 2015-07-02
[CDAP-1202] Scheduler APIs (Only AbstractSchedulerService) now takes …

bdmogal:
…care of MDS operations on schedules like add update and delete.

Also refactored the ``Scheduler`` interface to achieve this, mostly around passing in the entire ``ScheduleSpecification`` as opposed to just the ``Schedule`` object, because the MDS needs it.
Also added a way to add a schedule without updating the MDS. This is needed during the deployment pipeline, where ``ApplicationRegistrationStage`` already adds ``ScheduleSpecification`` to MDS (via ``ApplicationSpecification``.

Jira: [CDAP-1202](https://issues.cask.co/browse/CDAP-1202)
Build: http://builds.cask.co/browse/CDAP-DUT2128

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
@@ -19,6 +19,7 @@ package co.cask.cdap.internal.app.runtime.schedule;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
 import co.cask.cdap.api.schedule.Schedule;
 import co.cask.cdap.api.schedule.ScheduleSpecification;
+import co.cask.cdap.api.workflow.ScheduleProgramInfo;
 import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.store.Store;
 import co.cask.cdap.common.ApplicationNotFoundException;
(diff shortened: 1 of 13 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
```diff
@@ -704,6 +707,60 @@ public Void apply(AppMds mds) throws Exception {
...
+      public Void apply(AppMds mds) throws Exception {
+        for (ApplicationSpecification appSpec : getAllApplications(namespaceId)) {
+          ApplicationSpecification newAppSpec =
+            removeSchedulesFromApp(appSpec, new Predicate<ScheduleProgramInfo>() {
+              @Override
+              public boolean apply(@Nullable ScheduleProgramInfo scheduleProgramInfo) {
+                return true;
+              }
+            });
+//          replaceAppSpecInProgramJar(program, newAppSpec);
```
bdmogal: need to fix this. perhaps get all programs and call this on each?

poornachandra: Filed JIRA https://issues.cask.co/browse/CDAP-2957 to use app spec from MDS instead of the app spec in program jar. Once the JIRA is fixed, we'll no longer have to update program jar for changes to app spec.

In this case - if delete schedules for a namespace can only be called while deleting a namespace then there is no point in updating the program jars. I think it should be okay to add a comment for now, and not update the program jars.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
```diff
@@ -34,23 +35,34 @@
...
-  void schedule(Id.Program program, SchedulableProgramType programType, Schedule schedule,
+  void schedule(Id.Program program, ScheduleSpecification scheduleSpec, boolean updateMds)
+    throws SchedulerException;
+
+  /**
+   * Schedule a program to be run in a defined schedule.
+   *
+   * @param program Program that needs to be run.
+   * @param scheduleSpec specification of the schedule.
+   * @param properties system properties to be passed to the schedule.
```
poornachandra: `scheduleSpec` contains `properties`, and we pass in additional `properties` to `schedule` method. It is a bit confusing as to whether they are the same properties or not.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
```diff
@@ -290,6 +262,65 @@ public ScheduleState scheduleState(Id.Program program, SchedulableProgramType pr
...
+        } catch (NotFoundException e) {
+          // Should not happen - we just created it. Could have been deleted just in between
+          LOG.info("Schedule could not be suspended - it did not exist: {}", schedule.getName());
+        }
+      }
+    }
+  }
+
+  private void deleteSchedules(Id.Program program, SchedulableProgramType programType,
+                               boolean deleteFromStore) throws SchedulerException {
```
poornachandra: Are there any cases where we would want to delete a schedule, but not remove it from store?

poornachandra: Looks like the only place where we would want to do it is from `deleteAllSchedules

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/Scheduler.java
```diff
@@ -34,23 +35,34 @@
...
-   * @param schedule Schedule with which the program runs.
-   * @param properties system properties to be passed to the schedule
+   * @param scheduleSpec specification of the schedule.
+   * @param updateMds indicates whether to update the application specification in the MDS with this schedule. This is
+   *                  needed because during application deployment, the app spec in the MDS already contains a schedule
+   *                  spec, so we do not want to add it again/override it.
    * @throws SchedulerException on unforeseen error.
    */
-  void schedule(Id.Program program, SchedulableProgramType programType, Schedule schedule,
+  void schedule(Id.Program program, ScheduleSpecification scheduleSpec, boolean updateMds)
```
poornachandra: Having a flag called `updateMds

poornachandra:
The other thing that is confusing is that both SchedulerService and Schedulers implement the same interface. Due to this we need to have methods that now throw `UnsupportedOperationException`. Can you add this refactoring to the JIRA too?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
```diff
@@ -704,6 +707,60 @@ public Void apply(AppMds mds) throws Exception {
...
+          mds.apps.updateAppSpec(namespaceId.getId(), appSpec.getName(), newAppSpec);
+        }
+        return null;
+      }
+    });
+  }
+
+  private ApplicationSpecification removeSchedulesFromApp(ApplicationSpecification appSpec,
+                                                          final Predicate<ScheduleProgramInfo> predicate) {
```
poornachandra: Just to make things clear, can you call `predicate` as `removalPredicate` or something similar?

poornachandra:
I don't see any test for updating schedules, can you add one?

poornachandra:
@bdmogal can this PR wait until the re-factoring of Scheduler APIs is complete? I feel that merging this code without re-factoring will make code more confusing.

bdmogal:
Yes, it can wait. The only implication will be that CDAP-1202 will stay open. I agree that the APIs are confusing and need to be refactored a fair bit.

bdmogal:
@poornachandra created https://issues.cask.co/browse/CDAP-2974 to refactor Scheduler APIs. Closing this PR. Will re-open once CDAP-2974 is fixed.

PR #3085, opened 2015-07-02
Always create MapReduce job jar on local fs.

chtyim:
- Job.submit will copy it to HDFS
- Need this in order to work on both local, HDFS and MapR FS

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
@@ -77,6 +77,7 @@ import org.apache.hadoop.mapreduce.Reducer;
 import org.apache.hadoop.security.Credentials;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
+import org.apache.twill.filesystem.LocalLocationFactory;
 import org.apache.twill.filesystem.Location;
 import org.apache.twill.filesystem.LocationFactory;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -699,12 +702,18 @@ private Location buildJobJar(BasicMapReduceContext context) throws IOException {
...
-    Location jobJar =
-      locationFactory.create(String.format("%s.%s.%s.%s.%s.jar",
-                                           ProgramType.MAPREDUCE.name().toLowerCase(),
-                                           programId.getNamespaceId(), programId.getApplicationId(),
-                                           programId.getId(), context.getRunId().getId()));
-
+    // Build the jobJar on local temp directory. On Job.submit(), the MapReduce framework will copy the file to HDFS
+    // and distribute it to task containers.
+    // It has to be local, otherwise it won't work on MapR.
+    File tempDir = new File(cConf.get(Constants.CFG_LOCAL_DATA_DIR),
```
jwang47: Where or when is this cleaned up?

chtyim: tempDir is the "data/tmp" directory, which the directory itself won't get delete (in standalone, the content will get removed on start).

In distributed mode, it's the container directory, hence also don't need to worry about the cleanup. The JobJar file created inside the tempDir already get cleanup properly (existing code, cleanup on job finish as well as any exception happens in between).

gsps1:
LGTM :+1:

chtyim:
Verified in all different mode.

PR #3087, opened 2015-07-02
external filesets

anew:
Adds a new property "data.external" for file sets (and (time-)partitioned file sets)
- this property may be used by other dataset types in the future
- this means that the dataset does not own its data (the actual files). It only refers to existing files
- also, the file set becomes read-only because of that - we assume some other process writes the files
- for partitioned file sets this means that partitions point to existing files, but dropping a partition does not delete the file
- also, dropping the dataset does not delete the base directory

This also fixes a bug in the existing (T)PFS: dropPartition() was not deleting the underlying file.


jwang47:
Also update the javadoc for `DatasetInstanceHandler.drop()`: https://github.com/caskdata/cdap/blob/develop/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java#L153

jwang47:
What happens when a user changes data.external from `false` to `true` and vice versa? Is this possible?

anew:
Good point. This should never be changed. I need to figure out how to prevent that.

anew:
OK. Turns out that there is no way for a dataset to reject an update of its properties. DatasetAdmin.upgrade() was meant for that, but it is only called after the properties have been updated, to apply the changes. Opened a Jira [CDAP-2916].

anew:
About the DatasetInstanceHandler. I changed "associated with it" to "owned by it".

anew:
Addressed your comments, @jwang47. Is it good to merge?

anew:
Fixed another existing bug, where pre-2.8 time partitions were not properly deleted. 

Build is passing now: http://builds.cask.co/browse/CDAP-DUT2150-5

anew:
@jwang47 I addressed everything I believe. Can you take another look and tell me if it's good to merge?

jwang47:
LGTM

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -220,12 +223,25 @@ protected void addPartitionToExplore(PartitionKey key, String path) {
   @Override
   public void dropPartition(PartitionKey key) {
     byte[] rowKey = generateRowKey(key, partitioning);
+    Partition partition = getPartition(key);
+    if (partition == null) {
+      // silently ignore non-existing partitions
```
anwar6953: Why do we need to ignore deletion of non-existent partition keys now?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/file/FileSetDataset.java
```diff
@@ -149,21 +149,28 @@ private Location createLocation(String relativePath) {
 
   @Override
   public Location getBaseLocation() {
+    // TODO: if the file set is external, we could return a ReadOnlyLocation that prevents writing
```
anwar6953: Probably should a JIRA should at least be made for these, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anew: We have always silently ignored it (by deleting a non-existent row). I only added comment to make that clear.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/file/FileSetDataset.java
anew: added https://issues.cask.co/browse/CDAP-2934

PR #3089, opened 2015-07-02
[CDAP-2740] [CDAP-2830] Fix/ui node server crash

elfenheart:
Cluster: 
http://securecluster1870-1000.dev.continuuity.net:9999/
http://nodejsfix1872-1000.dev.continuuity.net:9999/


ajainarayanan:
:+1: LGTM

PR #3093, opened 2015-07-06
CDAP-2846 artifact plugins extend another artifact

albertshau:
Doing some refactoring to reflect the fact that plugins extend
a set of parent artifacts and are not stand alone objects. Changes
the table schema to support finding the plugins that extend a
given parent artifact.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
@@ -18,22 +18,37 @@ package co.cask.cdap.internal.app.runtime.artifact;
 
 import co.cask.cdap.api.templates.plugins.PluginClass;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 /**
- * Metadata about an artifact, such as what plugins are contained in the artifact.
+ * Metadata about an artifact, such as what plugins are contained in the artifact, and what other artifacts can use
+ * the plugins in this artifact.
  */
 public class ArtifactMeta {
   private final List<PluginClass> plugins;
+  // can't call this 'extends' since that's a reserved keyword
+  private final Set<ArtifactRange> usableBy;
 
   public ArtifactMeta(List<PluginClass> plugins) {
+    this(plugins, ImmutableSet.<ArtifactRange>of());
+  }
+
+  public ArtifactMeta(List<PluginClass> plugins, Set<ArtifactRange> usableBy) {
     this.plugins = ImmutableList.copyOf(plugins);
+    this.usableBy = ImmutableSet.copyOf(usableBy);
   }
 
   public List<PluginClass> getPlugins() {
     return plugins;
+  }
+
+  public Set<ArtifactRange> getUsableBy() {
+    return usableBy;
   }
 
   @Override
@@ -47,11 +62,19 @@ public class ArtifactMeta {
 
     ArtifactMeta that = (ArtifactMeta) o;
 
-    return Objects.equals(plugins, that.plugins);
+    return Objects.equals(plugins, that.plugins) && Objects.equals(usableBy, that.usableBy);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(plugins);
+    return Objects.hash(plugins, usableBy);
+  }
+
+  @Override
+  public String toString() {
+    return "ArtifactMeta{" +
+      "plugins=" + plugins +
+      ", usableBy=" + usableBy +
+      '}';
   }
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
@@ -27,7 +27,10 @@ import java.util.Set;
 
 /**
  * Metadata about an artifact, such as what plugins are contained in the artifact, and what other artifacts can use
- * the plugins in this artifact.
+ * the plugins in this artifact. For example, we could have an etl-batch-lib artifact that contains
+ * 20 different plugins that are meant to be used by the application contained in the etl-batch artifact.
+ * In this case, the artifact meta for etl-batch-lib would contain details about each of those 20 plugins, as well
+ * as information about which versions of the etl-batch artifact can use the plugins it contains.
  */
 public class ArtifactMeta {
   private final List<PluginClass> plugins;
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
@@ -332,7 +332,7 @@ public class ArtifactStore {
   private boolean matches(PluginData pluginData, ArtifactVersion version) {
     ArtifactVersion lower = new ArtifactVersion(pluginData.parentVersionLower);
     ArtifactVersion upper = new ArtifactVersion(pluginData.parentVersionUpper);
-    return lower.compareTo(version) <= 0 && upper.compareTo(version) > 0;
+    return version.compareTo(lower) >= 0 && version.compareTo(upper) < 0;
   }
 
   /**
@@ -510,9 +510,11 @@ public class ArtifactStore {
     }
   }
 
+  // this method examines all plugins in the given row and checks if they extend the given parent artifact.
+  // if so, information about the plugin artifact and the plugin details are added to the given map.
   private void addPluginsToMap(Id.Artifact parentArtifactId, Map<ArtifactInfo, List<PluginClass>> map,
                                Row row) throws IOException {
-    // column is the artifact name and version, value is the serialized PluginClass
+    // column is the artifact namespace, name, and version. value is the serialized PluginData
     for (Map.Entry<byte[], byte[]> column : row.getColumns().entrySet()) {
       ArtifactColumn artifactColumn = ArtifactColumn.parse(column.getKey());
       PluginData pluginData = gson.fromJson(Bytes.toString(column.getValue()), PluginData.class);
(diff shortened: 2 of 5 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2154

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
```diff
@@ -18,24 +18,39 @@
...
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 /**
- * Metadata about an artifact, such as what plugins are contained in the artifact.
+ * Metadata about an artifact, such as what plugins are contained in the artifact, and what other artifacts can use
+ * the plugins in this artifact.
```
gokulavasan: It would really help if you add an example here.

gokulavasan: Also can the usableBy info about an Artifact change after the artifact has been deployed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -64,26 +66,58 @@
...
+ * rowkey                            columns
+ * a:system:ETLBatch                 etlbatch:3.1.0 -> {AppData}
+ * r:system:etlbatch                 3.1.0 -> {ArtifactData}
+ *
+ * After that, a system artifact etlbatch-lib-3.1.0 is added, which extends etlbatch and contains
+ * stream sink and table sink plugins. The meta table will look like:
+ *
+ * rowkey                            columns
+ * a:system:ETLBatch                 etlbatch:3.1.0 -> {AppData}
+ * p:system:etlbatch:sink:stream     system:etlbatch-lib:3.1.0 -> {PluginData}
```
gokulavasan: Wondering if {PluginData} refers to the actual object (serialized) that is stored or just a high level terminology. If possible, can we use the actual class that represents the PluginData?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -262,60 +296,43 @@ public ArtifactData apply(DatasetContext<Table> context) throws Exception {
...
-        }
-      });
-    if (plugin == null) {
-      throw new PluginNotExistsException(artifactId, type, name);
-    }
-    return plugin;
+  private boolean matches(PluginData pluginData, ArtifactVersion version) {
+    ArtifactVersion lower = new ArtifactVersion(pluginData.parentVersionLower);
+    ArtifactVersion upper = new ArtifactVersion(pluginData.parentVersionUpper);
+    return lower.compareTo(version) <= 0 && upper.compareTo(version) > 0;
```
gokulavasan: Would it be more readable if we switch the logic to do version.compareTo(lower) >=0 && version.compareTo(upper) < 0 ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -398,33 +416,35 @@ public Boolean apply(DatasetContext<Table> context) throws Exception {
...
+        Table table = context.get();
+        Scanner scanner = table.scan(scanArtifacts(namespace));
+        Row row;
+        while ((row = scanner.next()) != null) {
+          table.delete(row.getRow());
         }
-      });
-    }
+        Scan pluginsScan = new Scan(
+          Bytes.toBytes(String.format("%s:%s:", PLUGIN_PREFIX, namespace.getId())),
```
gokulavasan: Why both %s:%s: and %s:%s; ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -398,33 +416,35 @@ public Boolean apply(DatasetContext<Table> context) throws Exception {
...
+          Bytes.toBytes(String.format("%s:%s;", PLUGIN_PREFIX, namespace.getId()))
```
gokulavasan: Why %s:%s; ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -481,88 +510,84 @@ private void addArchivesToList(List<ArtifactDetail> archives, Row row) throws IO
     }
   }
 
-  // given a row representing plugin metadata, add all plugins in the row to the given map
-  private void addPluginsToMap(Map<ArtifactInfo, List<PluginClass>> map, Row row) throws IOException {
-    // plugin key contains namespace, plugin type, and plugin name
-    PluginKey pluginKey = PluginKey.parse(row.getRow());
-
+  private void addPluginsToMap(Id.Artifact parentArtifactId, Map<ArtifactInfo, List<PluginClass>> map,
```
gokulavasan: Some comment on what the Row is expected to contain would be helpful.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -481,88 +510,84 @@ private void addArchivesToList(List<ArtifactDetail> archives, Row row) throws IO
...
+                               Row row) throws IOException {
     // column is the artifact name and version, value is the serialized PluginClass
```
gokulavasan: Is this comment still true? I see namespace of the artifact also in the column in the javadoc for the class?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
gokulavasan: Also a comment on what this method does?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -64,26 +66,58 @@
...
+ * a:system:ETLBatch                 etlbatch:3.1.0 -> {AppData}
+ * r:system:etlbatch                 3.1.0 -> {ArtifactData}
+ *
+ * After that, a system artifact etlbatch-lib-3.1.0 is added, which extends etlbatch and contains
+ * stream sink and table sink plugins. The meta table will look like:
+ *
+ * rowkey                            columns
+ * a:system:ETLBatch                 etlbatch:3.1.0 -> {AppData}
+ * p:system:etlbatch:sink:stream     system:etlbatch-lib:3.1.0 -> {PluginData}
+ * p:system:etlbatch:sink:table      system:etlbatch-lib:3.1.0 -> {PluginData}
```
gokulavasan: Just to confirm, PluginSelector can be used to not only choose among different versions of the artifacts with same name, but also among different artifact names. So a plugin named table of type sink, can be part of artifacts with different names and versions?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -481,88 +510,84 @@ private void addArchivesToList(List<ArtifactDetail> archives, Row row) throws IO
...
     // column is the artifact name and version, value is the serialized PluginClass
     for (Map.Entry<byte[], byte[]> column : row.getColumns().entrySet()) {
       ArtifactColumn artifactColumn = ArtifactColumn.parse(column.getKey());
       PluginData pluginData = gson.fromJson(Bytes.toString(column.getValue()), PluginData.class);
 
-      Id.Artifact artifactId =
-        Id.Artifact.from(pluginKey.namespace, artifactColumn.name, artifactColumn.version);
-      ArtifactInfo artifactInfo = new ArtifactInfo(artifactId, pluginData.artifactLocation);
+      // filter out plugins that don't extend this version of the parent artifact
+      if (matches(pluginData, parentArtifactId.getVersion())) {
```
gokulavasan: Just curious, if the only thing we check if the artifact version is supported in the plugin's support range. What about artifact id (and namespace) match? If we don't need that, they should we be just taking in ArtifactVersion instead of Id.Artifact?

gokulavasan:
@albertshau Finished review. Few questions/comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: yep its the actual class (private to ArtifactStore)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: sure can change that

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: ';' is one bigger than ':'

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: good catch, no its not true. will fix.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: that is correct

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: we don't need to check namespace and name because the scan will filter out anything that doesn't match.  But there isn't a way to do something similar with a version range, at least none that I could think of.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactMeta.java
albertshau: hm thats a good question.  The current code would not let you (unless its a snapshot), but it may make sense. I think to start out with we will not allow it, but could relax that later on.

albertshau:
@gokulavasan updated the PR, could you take another look?

gokulavasan:
LGTM :+1:

PR #3094, opened 2015-07-06
(CDAP-344) Runs spark in all modes (in-memory, standalone, distributed)

chtyim:
High level idea:

- Replace the ExecutorURLClassLoader (Spark 1.2) / MutableURLClassLoader (Spark 1.3+) in Spark with our own implementation
  - Do class loading in CDAP way, which support program classloader and system classloader correctly
  - Stores the SparkContext object inside the classloader
  - Reconstruct the SparkContext object in remote node when the classloader is being created (or when Spark metrics sink is created, depending on which happens first).
  - Since we always have the proper SparkContext object to work with (with all CDAP services, like metrics, logging, tx...), we don't need to do anything for serializing/deserializing system services objects

- Enabled access to dataset through `SparkContext`
  - `SparkContext` now extends from `DatasetContext`
  - User can access `Dataset` in `beforeSubmit`, `onFinish` as well as in `run`.
  - Access of `Dataset` inside closure is not yet supported, since `Dataset` are not serializable, nor the `SparkContext`.

- Properly integrated with Spark metrics system. All metrics emitted by Spark framework are also collected.

- Integration with Workflow and WorkflowToken is not done. Need to have another PR for that.

- The `beforeSubmit()`, `run()` and `onFinish()` are still executed in different transaction to preserve the same behavior (and same as MR as well). It's a separate discussion whether we should makes all of them run in a single long transaction or not.


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillApplication.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Override
+  protected ProgramType getType() {
+    return ProgramType.SPARK;
+  }
+
+  @Override
+  protected void addRunnables(Map<String, RunnableResource> runnables) {
+    // TODO: Make it configurable
```
bdmogal: Should we create a jira for this, if it is not there already?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillRunnable.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.twill.api.TwillRunnable;
+import org.spark_project.protobuf.GeneratedMessage;
+
+/**
+ * A {@link TwillRunnable} wrapper for {@link SparkProgramRunner}.
+ */
+public class SparkTwillRunnable extends AbstractProgramTwillRunnable<SparkProgramRunner> {
+
+  // Dummy vars for dependency tracing
+  private GeneratedMessage gm;
```
bdmogal: Should we add a ``@SuppressWarnings("unused")`` as well?

anew:
I don't see test cases in this PR. Doesn't a change of this size deserve some tests?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.mapreduce.InputFormat;
+
+import java.util.Map;
+
+/**
+ * Interface to provide abstraction for accessing either scala or java version of the Apache SparkContext.
+ */
+public interface ApacheSparkContext {
```
bdmogal: Should we add javadoc for methods of this interface?

chtyim:
Currently relying on existing test cases. Unless we want to make this PR even bigger, otherwise, going to do it in separate PR.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
chtyim: That's true. Missed that.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,376 @@
...
+  public ExecutionSparkContext(SparkSpecification specification, Id.Program programId, RunId runId,
+                                ClassLoader programClassLoader, long logicalStartTime,
+                                Map<String, String> runtimeArguments,
+                                Transaction transaction, DatasetFramework datasetFramework,
+                                DiscoveryServiceClient discoveryServiceClient, MetricsContext metricsContext,
+                                LoggingContext loggingContext, Configuration hConf, StreamAdmin streamAdmin) {
+    super(specification, programId, runId, programClassLoader, logicalStartTime,
+          runtimeArguments, discoveryServiceClient, metricsContext, loggingContext);
+
+    this.datasets = new HashMap<>();
```
jayfjin: do you think Maps.newHashMap() is better?

chtyim: It's longer, so no. Also the new Java 7 syntax is meant to replace guava usage in this case.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  @Override
+  public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass) {
+    Map<String, String> datasetArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
+
+    // Verify the dataset exists. If it doesn't exist, exception will be thrown
+    Closeables.closeQuietly(createDataset(datasetName, datasetArgs));
+
+    Configuration configuration = new Configuration(hConf);
```
bdmogal: why do we need to clone the ``hConf`` here?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
bdmogal: Also, should these methods throw ``Exception``s, so we don't have to ``Throwables.propagate``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: A Spark program can call `readFromDataset` multiple times to create different RDD. The configurations setup by each invocation shouldn't affect others.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
chtyim: Where do we do propagate? These methods act as bridges to the underlying Apache SparkContext, which try to stay as close as to the actual API, which they don't throw checked exception.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
bdmogal: Cool! Good to know. Should we add a comment?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
bdmogal: I see. Fair enough. I saw that you had to ``Throwables.propagate`` in [ExecutionSparkContext](https://github.com/caskdata/cdap/pull/3094/files#diff-e575b23f20d63cbf38943488b1f63ff1R133). But I agree that these should be as close to the original API as possible.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,376 @@
...
+                              long startTime, long endTime, Class<? extends StreamEventDecoder> decoderType) {
+
+    StreamBatchReadable stream = (decoderType == null)
+      ? new StreamBatchReadable(streamName, startTime, endTime)
+      : new StreamBatchReadable(streamName, startTime, endTime, decoderType);
+
+    try {
+      Configuration configuration = configureStreamInput(new Configuration(hConf), stream, vClass);
+      T streamRDD = getApacheSparkContext().createRDD(StreamInputFormat.class,
+                                                         LongWritable.class, vClass, configuration);
```
bdmogal: mis-alignment

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
rohitsinha54: I remember we decided not to put Apache or any organization name in naming of classes or functions.

chtyim: We decided not to for public API. This is internal API.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillApplication.java
chtyim: I am going to address this right after this PR to not to make this PR even bigger.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillRunnable.java
chtyim: sure

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/metrics/SparkUserMetrics.java
```diff
@@ -16,10 +16,13 @@
...
 package co.cask.cdap.internal.app.runtime.spark.metrics;
 
+import co.cask.cdap.api.ServiceDiscoverer;
 import co.cask.cdap.api.metrics.Metrics;
 import co.cask.cdap.api.metrics.MetricsContext;
 import co.cask.cdap.api.spark.Spark;
-import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.internal.app.runtime.spark.ExecutionSparkContext;
+import co.cask.cdap.internal.app.runtime.spark.SparkContextProvider;
+import org.apache.twill.discovery.DiscoveryServiceClient;
```
rohitsinha54: There are some unnecessary imports here. Can you remove them.

rohitsinha54:
Were you able to run the SparkProgramRunnerTest ? I am trying to run it and it fails with:
```
Exception in thread "Spark - SparkComputeChars" java.lang.IllegalStateException: SparkContextProvider.getSparkContext should only be called in Spark executor process.
	at com.google.common.base.Preconditions.checkState(Preconditions.java:149)
	at co.cask.cdap.internal.app.runtime.spark.SparkContextProvider.createIfNotExists(SparkContextProvider.java:99)
	at co.cask.cdap.inte	at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:110) ~[spark-core_2.10-1.3.1.jar:1.3.1]
	at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala) ~[spark-core_2.10-1.3.1.jar:1.3.1]
	at co.cask.cdap.internal.app.runtime.spark.SparkRuntimeService.run(SparkRuntimeService.java:161) ~[classes/:na]
	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$1.run(AbstractExecutionThreadService.java:52) ~[guava-13.0.1.jar:na]
	at co.cask.cdap.internal.app.runtime.spark.SparkRuntimeService$2$1.run(SparkRuntimeService.java:232) [classes/:na]
	at java.lang.Thread.run(Thread.java:745) [na:1.7.0_79]
rnal.app.runtime.spark.SparkContextProvider.getSparkContext(SparkContextProvider.java:80)
	at co.cask.cdap.internal.app.runtime.spark.dataset.SparkDatasetInputFormat.getSplits(SparkDatasetInputFormat.java:68)
	at org.apache.spark.rdd.NewHadoopRDD.getPartitions(NewHadoopRDD.scala:95)
	at org.apache.spark.rdd.RDD$$anonfun$partitions$2.apply(RDD.scala:219)
	at org.apache.spark.rdd.RDD$$anonfun$partitions$2.apply(RDD.scala:217)
	at scala.Option.getOrElse(Option.scala:120)
	at org.apache.spark.rdd.RDD.partitions(RDD.scala:217)
	at org.apache.spark.rdd.MapPartitionsRDD.getPartitions(MapPartitionsRDD.scala:32)
	at org.apache.spark.rdd.RDD$$anonfun$partitions$2.apply(RDD.scala:219)
	at org.apache.spark.rdd.RDD$$anonfun$partitions$2.apply(RDD.scala:217)
	at scala.Option.getOrElse(Option.scala:120)
	at org.apache.spark.rdd.RDD.partitions(RDD.scala:217)
	at org.apache.spark.SparkContext.runJob(SparkContext.scala:1505)
	at org.apache.spark.rdd.PairRDDFunctions.saveAsNewAPIHadoopDataset(PairRDDFunctions.scala:1019)
	at co.cask.cdap.internal.app.runtime.spark.ScalaApacheSparkContext.saveAsDataset(ScalaApacheSparkContext.java:70)
	at co.cask.cdap.internal.app.runtime.spark.ExecutionSparkContext.writeToDataset(ExecutionSparkContext.java:129)
	at co.cask.cdap.internal.app.runtime.spark.ScalaCharCountProgram.run(ScalaCharCountProgram.scala:36)
	at co.cask.cdap.internal.app.runtime.spark.SparkProgramWrapper.<init>(SparkProgramWrapper.java:58)
	at co.cask.cdap.internal.app.runtime.spark.SparkProgramWrapper.main(SparkProgramWrapper.java:38)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.apache.spark.deploy.SparkSubmit$.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:569)
	at org.apache.spark.deploy.SparkSubmit$.doRunMain$1(SparkSubmit.scala:166)
	at org.apache.spark.deploy.SparkSubmit$.submit(SparkSubmit.scala:189)
	at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:110)
	at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)
	at co.cask.cdap.internal.app.runtime.spark.SparkRuntimeService.run(SparkRuntimeService.java:161)
	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$1.run(AbstractExecutionThreadService.java:52)
	at co.cask.cdap.internal.app.runtime.spark.SparkRuntimeService$2$1.run(SparkRuntimeService.java:232)
	at java.lang.Thread.run(Thread.java:745)
```
Does the unit test needs changes ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/metrics/SparkUserMetrics.java
chtyim: done

chtyim:
Good catch @rohitsinha54 That's due to one later wrong change I made that hasn't been running though bamboo yet.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextConfig.java
```diff
@@ -16,191 +16,134 @@
...
-    return hConf.get(HCONF_ATTR_RUN_ID);
-  }
-
-  private static void setLogicalStartTime(long startTime) {
-    hConf.setLong(HCONF_ATTR_LOGICAL_START_TIME, startTime);
+  /**
+   * Returns the runtime arguments stored in the configuration.
+   */
+  public Map<String, String> getArguments() {
+    return GSON.fromJson(hConf.get(HCONF_ATTR_ARGS), new TypeToken<Map<String, String>>() { }.getType());
```
bdmogal: Should we create a constant for the ``TypeToken``?

cdap-app-fabric/src/main/java/org/apache/spark/util/MutableURLClassLoader.java
```diff
@@ -0,0 +1,62 @@
...
+    LOG.info("MutableURLClassLoader intercepted");
+  }
+
+  public void addURL(URL url) {
+    // no-op
+  }
+
+  public URL[] getURLs() {
+    // return empty array
+    return new URL[0];
```
rohitsinha54: Can't we have an empty array as a static and return that always ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextConfig.java
chtyim: Just one usage in this class. It doesn't matter (in fact constant takes up perm-gen space).

cdap-common/src/main/java/co/cask/cdap/common/lang/ProgramResources.java
```diff
@@ -115,8 +115,8 @@ public String apply(ClassPath.ClassInfo input) {
     Set<String> resources = getBaseResources();
 
     // Base on the type, add extra resources
-    // Current only Spark type has extra visible resources
-    if (type == ProgramType.SPARK) {
+    // Current only Spark and Workflow type has extra visible resources
+    if (type == ProgramType.SPARK || type == ProgramType.WORKFLOW) {
```
rohitsinha54: Is it because Workflow can have Spark in them which will have extra visible resource ?
Nice!

cdap-app-fabric/src/main/java/org/apache/spark/util/MutableURLClassLoader.java
chtyim: Doesn't really matter in this case. This method is not frequently called and I believe JVM does empty array creation optimization (http://sgdev-blog.blogspot.com/2015/06/can-java-optimize-empty-array-allocation.html)

cdap-common/src/main/java/co/cask/cdap/common/lang/ProgramResources.java
chtyim: Yes. I don't know how it works before. When running Spark inside Workflow, we use the same classloader as the Workflow program to load the Spark program.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramWrapper.java
```diff
@@ -18,221 +18,73 @@
...
- * </li>
- * <li>
- * Sets {@link SparkContext} to concrete implementation of {@link JavaSparkContext} if user program implements {@link
- * JavaSparkProgram} or to {@link ScalaSparkContext} if user's program implements {@link ScalaSparkProgram}
- * </li>
- * <li>
- * Run user's program with extracted arguments from the argument list
- * </li>
- * </ol>
+ * This first command line argument to this class is the name of the user's Spark program class.
```
bdmogal: typo: *The* first command

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkServiceDiscoverer.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  public void writeExternal(ObjectOutput out) throws IOException {
+    // no-op
+  }
+
+  @Override
+  public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
+    // no-op
+  }
+}
```
bdmogal: missing newline at the end of file?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/dataset/CloseableBatchWritable.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.Closeable;
+
+/**
+ * A {@link BatchWritable} that is also {@link Closeable}.
+ *
+ * @param <K> The key type.
+ * @param <V> The value type.
+ */
+public interface CloseableBatchWritable<K, V> extends BatchWritable<K, V>, Closeable {
+}
```
bdmogal: missing newline?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkServiceDiscoverer.java
chtyim: I saw it in IDE and checkstyle is not complaining.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/dataset/CloseableBatchWritable.java
chtyim: I saw it in IDE and checkstyle is not complaining.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkServiceDiscoverer.java
bdmogal: Don't know why though. Seen that with other PRs too. My IDE seems to complain.

chtyim: This is the hexdump of the file, which shows it has a `0x0a` newline character at the end.

```
000008a0  20 20 20 20 2f 2f 20 6e  6f 2d 6f 70 0a 20 20 7d  |    // no-op.  }|
000008b0  0a 7d 0a                                          |.}.|
000008b3
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractContext.java
```diff
@@ -108,15 +111,19 @@ protected AbstractContext(Program program, RunId runId, Arguments arguments,
     this.pluginInstantiator = pluginInstantiator;
   }
 
-  public List<Id> getOwners() {
+  private List<Id> createOwners(Id.Program programId, @Nullable AdapterDefinition adapterSpec) {
     ImmutableList.Builder<Id> result = ImmutableList.builder();
-    result.add(program.getId());
+    result.add(programId);
```
anew: shouldn't it be either the program _or_ the adapter?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedSparkProgramRunner.java
```diff
@@ -0,0 +1,77 @@
...
+    // Extract and verify parameters
+    ApplicationSpecification appSpec = program.getApplicationSpecification();
+    Preconditions.checkNotNull(appSpec, "Missing application specification.");
+
+    ProgramType processorType = program.getType();
+    Preconditions.checkNotNull(processorType, "Missing processor type.");
+    Preconditions.checkArgument(processorType == ProgramType.SPARK, "Only SPARK process type is supported.");
+
+    SparkSpecification spec = appSpec.getSpark().get(program.getName());
+    Preconditions.checkNotNull(spec, "Missing SparkSpecification for %s", program.getName());
```
anew: Better if all the other messages also contain the program name

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedSparkProgramRunner.java
```diff
@@ -0,0 +1,77 @@
...
+
+    LOG.info("Launching Spark program: " + program.getName() + ":" + spec.getName());
```
anew: can the program name and the spec name be different?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkServiceDiscoverer.java
bdmogal: Hmmm, maybe just github then.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedSparkProgramRunner.java
```diff
@@ -0,0 +1,77 @@
...
+    TwillController controller = launcher.launch(new SparkTwillApplication(program, spec,
+                                                                           localizeFiles, eventHandler));
+
+    RunId runId = RunIds.fromString(options.getArguments().getOption(ProgramOptionConstants.RUN_ID));
+    return new SparkTwillProgramController(program.getName(), controller, runId).startListen();
```
anew: should it use the program name or the spec's name?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractContext.java
chtyim: It's the existing logic (I just change it to use arguments other than fields). I think the adapter name is added below.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillApplication.java
anew: still the new PR needs a Jira, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedSparkProgramRunner.java
chtyim: I think they are the same. Wasn't paying too much attention to the log since it's pretty much a copy from other distributed program runner. Will make a round of refactoring across all of them.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillRunnable.java
```diff
@@ -0,0 +1,46 @@
...
+import co.cask.cdap.internal.app.runtime.spark.SparkProgramRunner;
+import org.apache.twill.api.TwillRunnable;
+import org.spark_project.protobuf.GeneratedMessage;
+
+/**
+ * A {@link TwillRunnable} wrapper for {@link SparkProgramRunner}.
+ */
+public class SparkTwillRunnable extends AbstractProgramTwillRunnable<SparkProgramRunner> {
+
+  // Dummy vars for dependency tracing
```
anew: not sure I understand. Has this to do with making sure that Twill includes GeneratedMessage in the jar? In that case, why not say it in the comment... or if it means something else?

chtyim: Yes, that's what Twill needs for the dependency tracing so that the jar containing the `GeneratedMessage` class will get included.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillApplication.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Map;
+
+/**
+ * A {@link TwillApplication} for running {@link Spark} program in distributed mode.
+ */
+public class SparkTwillApplication extends AbstractProgramTwillApplication {
+
+  private final String name;
+
+  public SparkTwillApplication(Program program, SparkSpecification spec,
```
anew: javadocs for public constructor?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillApplication.java
chtyim: Created CDAP-2936

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillRunnable.java
```diff
@@ -0,0 +1,46 @@
...
+  @SuppressWarnings("unused")
+  private GeneratedMessage gm;
+
+  SparkTwillRunnable(String name, String hConfName, String cConfName) {
```
anew: What is the name of an hConf or a cConf? Is it a file name? I guess I don't know because the super constructor also does not document it

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillApplication.java
chtyim: Maybe it's better to add to the parent class, since it's a parameter required by the parent only (not for this class).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillRunnable.java
chtyim: It's the file name. It's part of the existing class hierarchy, which actually is pretty old and in fact the name is no longer needed with some small refactoring. Will open a JIRA for code refactoring around `ProgramRunner` and `AbstractContext` to streamline code and doc.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/AbstractSparkContext.java
```diff
@@ -17,286 +17,210 @@
...
-abstract class AbstractSparkContext implements SparkContext {
+public abstract class AbstractSparkContext implements SparkContext, Closeable {
 
-  private static final Logger LOG = LoggerFactory.getLogger(AbstractSparkContext.class);
-  private static final Gson GSON = new Gson();
-  private static final Pattern SPACES = Pattern.compile("\\s+");
-  private static final String[] NO_ARGS = {};
-  private static final String SPARK_METRICS_CONF_KEY = "spark.metrics.conf";
+  // A Regex pattern that split with whitespace that supports "" quotation.
+  private static final Pattern QUOTED_STRING_PATTERN = Pattern.compile("([^\"]\\S*|\".+?\")\\s*");
```
anew: is .+? the same as .*     ?
Also would you want to support escaped quotes inside the quotes?

anew: I am thinking of the case where you have a Json string as the value of an argument

chtyim: this regex doesn't support escape quotes. This regex is just for supporting the `public String[] getRuntimeArguments(String argsKey)` method (which is defined in API, hence not going to change it). If use wants to get the original arg value (meaning not to get splitted into `String[]`), he can do it with the `Map<String, String> getRuntimeArguments()` method.

cdap-unit-test/src/test/java/co/cask/cdap/spark/metrics/SparkMetricsIntegrationTestRun.java
```diff
@@ -44,7 +45,7 @@
 @Category(XSlowTests.class)
 public class SparkMetricsIntegrationTestRun extends TestFrameworkTestBase {
 
-  public static final String METRICS_KEY = "system.driver.BlockManager.memory.remainingMem_MB";
+  private static final String METRICS_KEY = "system.<driver>.BlockManager.memory.remainingMem_MB";
```
chtyim: Spark 1.2 emits as "driver.*", while Spark 1.3 emits as "<driver>.*". Not sure if we should normalize it or not. Thoughts?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillApplication.java
anew: That also makes sense. I actually looked at the parent class and missed it there, too

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/AbstractSparkContext.java
anew: Not sure what that method is for? It takes the value for a given key, and splits it? So I could have a runtime argument 
myKey=abc "x y z" jkl, and this will return ["abc", "x y z", "jkl"] ? What is a use case for such an API?

anew: The parent class' javadoc does not make sense to me either.

chtyim: I also don't know why it's in the API. Maybe we should deprecate it and update the doc saying it just for splitting with whitespace?

anew: As discussed in person, let's remove this

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.hadoop.mapreduce.InputFormat;
+import org.apache.spark.SparkContext;
+import org.apache.spark.api.java.JavaSparkContext;
+
+import java.util.Map;
+
+/**
+ * Interface to provide abstraction for accessing either scala or java version of the Apache SparkContext.
+ */
+interface ApacheSparkContext {
```
anew: Name is a little misleading, because this has a method getContext() that returns an Apache SparkContext - according to its javadoc. Can we call this a SparkFacade or if you have a better idea...

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
```diff
@@ -0,0 +1,71 @@
...
+   * @param hConf the hadoop configuration
+   * @param <R> type of the RDD
+   * @param <K> type of the key
+   * @param <V> type of the value
+   */
+  <R, K, V> void saveAsDataset(R rdd, String datasetName, Map<String, String> arguments,
+                               Class<K> keyClass, Class<V> valueClass, Configuration hConf);
+
+  /**
+   * Returns the Apache {@link SparkContext} or {@link JavaSparkContext}, depending on the implementation.
```
anew: depending on the language used?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
```diff
@@ -0,0 +1,71 @@
...
+   */
+  <T> T getContext();
+
+  /**
+   * Stops the Apache SparkContext.
```
anew: you mean the Apache {@link SparkContext} or {@link JavaSparkContext}, depending on the language used? Or does it stop this ApacheSparkContext?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,381 @@
...
+    } else {
+      StreamInputFormat.setDecoderClassName(configuration, decoderType);
+    }
+    return configuration;
+  }
+
+  /**
+   * Creates a new instance of dataset.
+   */
+  private <T extends Dataset> T createDataset(String datasetName, Map<String, String> arguments) {
```
anew: this does not create the dataset, it only instantiates it. Can you rename the method to getDatasetInstance() or instantiateDataset()?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,381 @@
...
+    this.streamAdmin = streamAdmin;
+    this.datasetInstantiator = new DatasetInstantiator(programId.getNamespace(), datasetFramework,
+                                                       programClassLoader, getOwners(), getMetricsContext());
+  }
+
+  @Override
+  public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass) {
+    Map<String, String> datasetArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
+
+    // Verify the dataset exists. If it doesn't exist, exception will be thrown
```
anew: this is confusing. "create" a dataset to verify that it exists?

anew: Do we have to do this at all? Wouldn't it be fine to allow the RDD to fail if it does not exist? Or is there a more lightweight way to verify that it exists?

anew: Also, we do not verify this in writeToDataset(). If we verify, we should do it consistently across all methods

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,381 @@
...
+    Closeables.closeQuietly(createDataset(datasetName, datasetArgs));
+
+    // Clone the configuration since it's dataset specification and shouldn't affect the global hConf
+    Configuration configuration = new Configuration(hConf);
+    SparkDatasetInputFormat.setDataset(configuration, datasetName, datasetArgs);
+    return getApacheSparkContext().createRDD(SparkDatasetInputFormat.class, kClass, vClass, configuration);
```
anew: should this method register the Spark program as a user of the dataset? Like we do in readFromStream()?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  @Override
+  public <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass) {
+    Map<String, String> datasetArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
+    getApacheSparkContext().saveAsDataset(rdd, datasetName, datasetArgs, kClass, vClass, new Configuration(hConf));
```
anew: and same here?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,381 @@
...
+
+  /**
+   * Creates a new instance of dataset.
+   */
+  private <T extends Dataset> T createDataset(String datasetName, Map<String, String> arguments) {
+    T dataset = datasetInstantiator.getDataset(datasetName, arguments);
+
+    // Provide the current long running transaction to the given dataset.
+    // Remove it from the transaction aware list from the dataset instantiator, since we won't use it in this context
+    // Need to remove it to avoid memory leak.
```
anew: this is confusing. What context do we use it in?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: The way for registering usage of dataset vs stream is not consistent. For dataset, the DS service do it on the server side, hence no need to do registration here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: Same as above.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: It is use inside the `SparkProgram.run()` method as well as in remote Spark executor (when executing `RecordReader` / `RecordWriter`).

chtyim: "This context" in the comment I refer to the context when this class (`ExecutionSparkContext`) is use.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,381 @@
...
+
+    @SuppressWarnings("unchecked")
+    final BatchReadable<K, V> delegate = (BatchReadable<K, V>) dataset;
+    return new BatchReadable<K, V>() {
+      @Override
+      public List<Split> getSplits() {
+        try {
+          return delegate.getSplits();
+        } finally {
+          commitAndClose(dataset);
```
anew: this is only a good idea if you know that getSplits() is the last method that is called for the batch-readable. Do we know that for sure?

chtyim: It is. `getSplits` get called on the client side, while `createSplitReader` executes on the executor. However, if there is a better hook for closing the dataset instance created during the `getSplits` call, that would be great.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,381 @@
...
+   */
+  private void commitAndClose(Dataset dataset) {
+    try {
+      if (dataset instanceof TransactionAware) {
+        try {
+          ((TransactionAware) dataset).commitTx();
+        } catch (Exception e) {
+          // Nothing much we can do besides propagating it.
+          // Since the propagation will end up in the Spark Framework, we do the logging here.
+          LOG.error("Failed to commit dataset changes", e);
```
anew: it would be good to log the name of the dataset here. I know that Dataset does not have a getName() method. But since all callers of this method know the dataset name, can you add a parameter for that?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -0,0 +1,381 @@
...
+          throw Throwables.propagate(e);
+        }
+      }
+    } finally {
+      Closeables.closeQuietly(dataset);
+    }
+  }
+
+  /**
+   * Setups the configuration for reading from stream through {@link StreamInputFormat}.
```
anew: Sets up

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: Maybe call it `createDatasetInstance` or `newDatasetInstance`? I prefer to have "create" or "new" in the method name to indicate that a new object instance is created when this method is called.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ForwardingSplitReader.java
```diff
@@ -0,0 +1,64 @@
...
+import co.cask.cdap.api.data.batch.Split;
+import co.cask.cdap.api.data.batch.SplitReader;
+
+/**
+ * A {@link SplitReader} that forwards all methods call to another {@link SplitReader}.
+ *
+ * @param <KEY> The key type.
+ * @param <VALUE> The value type.
+ */
+public class ForwardingSplitReader<KEY, VALUE> extends SplitReader<KEY, VALUE> {
```
anew: this class by itself makes not much sense. The idea is that callers override at least one method, right? So perhaps make this class abstract?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: Actually this call is not needed. Not sure how it gets there, probably because I was testing something else.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkClassLoader.java
```diff
@@ -0,0 +1,83 @@
...
+  public static SparkClassLoader create() {
+    ExecutionSparkContext sparkContext = SparkContextProvider.getSparkContext();
+    return new SparkClassLoader(sparkContext);
+  }
+
+  /**
+   * Creates a new SparkClassLoader with the given {@link ExecutionSparkContext}.
+   */
+  public SparkClassLoader(ExecutionSparkContext context) {
+    super(null, ImmutableList.of(context.getProgramClassLoader(), SparkClassLoader.class.getClassLoader()));
```
anew: Is it valid to give CombineClassLoader null as the parent? It is not @Nullable...

chtyim: It is. null means bootstrap classloader as the parent. It's common practice in java ClassLoader hierarchy. I can add Nullable and javadoc to the CombineClassLoader for that.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
```diff
@@ -16,57 +16,203 @@
...
 
-  public SparkContextProvider(Configuration hConf) {
-    this.hConf = hConf;
-    this.contextConfig = new SparkContextConfig(hConf);
-    this.contextBuilder = null;
+    // Try to find it from the context classloader
+    SparkClassLoader sparkClassLoader = ClassLoaders.find(Thread.currentThread().getContextClassLoader(),
+                                                          SparkClassLoader.class);
+    if (sparkClassLoader != null) {
+      return sparkClassLoader.getContext();
```
anew: would you want to set sparkContext=sparkClassLoader.getContext(); return sparkContext;

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
```diff
@@ -16,57 +16,203 @@
...
+      // The problem with not shutting down nicely is that some logs/metrics might be lost
+      Services.chainStart(logAppenderService, zkClientService,
+                          kafkaClientService, metricsCollectionService, streamCoordinatorClient);
+      Runtime.getRuntime().addShutdownHook(new Thread() {
+        @Override
+        public void run() {
+          // The logger may already been shutdown. Use System.out/err instead
+          System.out.println("Shutting SparkClassLoader services");
+          Future<List<ListenableFuture<Service.State>>> future = Services.chainStop(logAppenderService,
+                                                                                    streamCoordinatorClient,
```
anew: can any of these throw an exception when stopping? If so, would the chain fail?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
```diff
@@ -16,57 +16,203 @@
...
 
-  private synchronized AbstractSparkContextBuilder getBuilder(CConfiguration conf) {
-    if (contextBuilder == null) {
-      String mrFramework = hConf.get(MRConfig.FRAMEWORK_NAME, MRConfig.LOCAL_FRAMEWORK_NAME);
-      if ("local".equals(mrFramework)) {
-        contextBuilder = new InMemorySparkContextBuilder(conf);
-      } else {
-        throw new RuntimeException("Spark does not run in distributed mode right now");
+  private static CConfiguration createCConf() throws MalformedURLException {
+    CConfiguration cConf = CConfiguration.create();
```
anew: is this common across other program contexts? Perhaps we should add a CConfiguration.create(URL ... resources)?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
```diff
@@ -16,57 +16,203 @@
...
+  private static Configuration createHConf() throws MalformedURLException {
+    Configuration hConf = new Configuration();
+    hConf.clear();
+    hConf.addResource(new File("hConf.xml").toURI().toURL());
+    return hConf;
+  }
+
+  private static Injector createInjector(CConfiguration cConf, Configuration hConf) {
+    return Guice.createInjector(
+      new ConfigModule(cConf, hConf),
```
anew: So this uses the distributed modules. Is it certain that this will never be used in local mode?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramWrapper.java
```diff
@@ -18,221 +18,73 @@
...
-   * @return String[] if the command line arguments are sufficient else throws a {@link RuntimeException}
-   * @throws IllegalArgumentException if the required numbers of command line arguments were not present
-   */
-  private String[] validateArgs(String[] arguments) {
-    if (arguments.length < PROGRAM_WRAPPER_ARGUMENTS_SIZE) {
-      throw new IllegalArgumentException("Insufficient number of arguments. Program class name followed by its" +
-                                           " arguments (if any) should be provided");
-    }
-    return arguments;
+    sparkProgramClass.newInstance().run(context);
```
anew: it is a little strange to call run() in a constructor.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
chtyim: chainStop will stop everything, regardless of failure or not.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
chtyim: yes it happens couple times here and there.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
chtyim: yes. That's what the check in https://github.com/caskdata/cdap/pull/3094/files#diff-72bffd56ae78926e49675a54f167eab9R99 is for.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramWrapper.java
chtyim: I also thought about whether it should be like the current way or like this in the `main()` method.

```java
new SparkProgramWrapper(args).run();
```

Any preferences?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
```diff
@@ -21,181 +21,176 @@
...
-      } catch (Throwable t) {
-        Locations.deleteQuietly(programJarCopy);
-        throw t;
+      SparkContextConfig contextConfig = new SparkContextConfig(hConf);
+
+      File jobJar = generateJobJar(tempDir);
+      File metricsConf = SparkMetricsSink.generateSparkMetricsConfig(File.createTempFile("metrics",
+                                                                                         ".properties", tempDir));
+
+      List<File> localizeArchives = Lists.newArrayList();
```
anew: localizedArchives

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
```diff
@@ -21,181 +21,176 @@
...
-      // depends on.
-      Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
+      LOG.debug("Submitting to spark with arguments: {}", Arrays.toString(sparkSubmitArgs));
+      SparkSubmit.main(sparkSubmitArgs);
+      success = true;
+    } catch (Exception e) {
+      success = false;
+      // See if the context is stopped. If it is, then it's not an error.
+      if (!executionContext.isStopped()) {
+        LOG.error("Spark program execution failure: {}", executionContext, e);
```
anew: does it have to be stopped if it is not yet stopped?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
```diff
@@ -269,112 +261,174 @@ public void apply() throws Exception {
...
   }
 
-  private Runnable createCleanupTask(final Iterable<File> files, final Iterable<Location> locations) {
+  /**
+   * Generates an empty JAR file.
+   *
+   * @return The generated {@link File} in the given target directory
+   */
+  private File generateJobJar(File targetDir) throws IOException {
+    File tempFile = new File(targetDir, "dummy.jar");
```
anew: why call it "dummy" and not "empty" or "job"?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
```diff
@@ -269,112 +261,174 @@ public void apply() throws Exception {
...
+    return tempFile;
+  }
+
+  /**
+   * Serialize {@link CConfiguration} to a file.
+   *
+   * @return The {@link File} of the serialized configuration in the given target directory.
+   */
+  private File saveCConf(CConfiguration cConf, File targetDir) throws IOException {
+    File file = new File(targetDir, "cConf.xml");
```
anew: should we use constants for these file name? That way it becomes a lot easier to find out where they are used.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/dataset/DatasetRecordReader.java
```diff
@@ -17,42 +17,31 @@
...
 
 import java.io.IOException;
 
 final class DatasetRecordReader<KEY, VALUE> extends RecordReader<KEY, VALUE> {
-  private static final Logger LOG = LoggerFactory.getLogger(DatasetRecordReader.class);
+
   private final SplitReader<KEY, VALUE> splitReader;
-  private final BasicSparkContext context;
-  private final String dataSetName;
   //TODO: Needs support for metrics when implemented
```
anew: please add a Jira for this Todo

cdap-app-fabric/src/main/java/org/apache/spark/executor/ExecutorURLClassLoader.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static ClassLoader createParent(ClassLoader parent) {
+    // If SparkClassLoader is already in the ClassLoader hierarchy, simply use the given parent.
+    if (ClassLoaders.find(parent, SparkClassLoader.class) != null) {
+      return parent;
+    }
+
+    // Ignore the given parent and create a SparkClassLoader as the parent
+    // This is what needed in Spark distributed mode
+    return SparkClassLoader.create();
```
anew: smart!

cdap-common/src/main/java/co/cask/cdap/common/lang/ProgramResources.java
anew: Spark in workflow never really worked.

cdap-unit-test/src/test/java/co/cask/cdap/spark/metrics/SparkMetricsIntegrationTestRun.java
anew: don't normalize. We would have to do that for many other things, and many versions. Unless we want to map this to a fixed documented CDAP metric. In that case I would keep the original plus the CDAP one.

anew:
Done with my first pass. 
Overall looks good and much cleaner than before. 
Please see my comments.

cdap-common/src/main/java/co/cask/cdap/common/lang/ProgramResources.java
rohitsinha54: We have support for Spark in Workflow from a long time and it works fine as far as I know. We also did demos where Spark was running as a part of Workflow in Spark summit. Do you mean it lacks some feature inside workflow ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
chtyim: It does , it's in the `finally` block below.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
chtyim: yeah, probably emptyJob.jar is better.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
chtyim: sure.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/dataset/DatasetRecordReader.java
chtyim: Actually this TODO is not accurate anymore since we now support metrics.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedSparkProgramRunner.java
chtyim: Sure fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedSparkProgramRunner.java
chtyim: They are actually the same.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/SparkTwillApplication.java
chtyim: Added in parent class.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
chtyim: Sure. Renamed to SparkFacade.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ApacheSparkContext.java
chtyim: It stops the underlying (Java)SparkContext, depending on the language.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: Good idea.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ForwardingSplitReader.java
chtyim: sure

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
chtyim: sure

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
chtyim: Will do it in separate PR. Created https://issues.cask.co/browse/CDAP-2947

cdap-common/src/main/java/co/cask/cdap/common/lang/ProgramResources.java
anew: I meant class loading was never thoroughly working. I may be wrong.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractContext.java
anew: OK, I guess this will change with the upcoming application changes.

chtyim: Yes it will.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: newDatasetInstance is fine with me, too.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: got it

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: That explains my confusion.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: In general, anybody using a BatchReadable can call getSplits at any time. But if this particular BatchReadable is only used for passing it to DatasetInputFormat, with in turn is only called by Spark in the way you describe, then it is ok.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkClassLoader.java
anew: thanks

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
anew: ok

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkContextProvider.java
anew: cool

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramWrapper.java
anew: I like the other way better than the current way.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
anew: Oh, it is the close() that stops it. Got it

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramWrapper.java
anew: I like the other way better than the current way.

anew:
Only one more thing left: https://github.com/caskdata/cdap/pull/3094/files#discussion_r34087579

anew:
LGTM if build passes

anew:
Last change looks good. LGTM after build passes

chtyim:
Build passed http://builds.cask.co/browse/CDAP-DUT2121-30
Will squash and merge.

PR #3095, opened 2015-07-06
CDAP-2618: simplified cube sink

abaranau:
Note: lots of mapping flexibility from StructuredRecord to CubeFact was removed in favor of simpler configuration.


abaranau:
http://builds.cask.co/browse/CDAP-DUT2159-1

cdap-api/src/main/java/co/cask/cdap/api/templates/plugins/PluginProperties.java
```diff
@@ -80,7 +80,7 @@ public Builder add(String key, String value) {
      * Creates a new instance of {@link PluginProperties} with the properties added to this builder prior to this call.
      */
     public PluginProperties build() {
-      return new PluginProperties(Collections.unmodifiableMap(new HashMap<>(properties)));
+      return new PluginProperties(new HashMap<>(properties));
```
chtyim: Why the change? The `PluginProperties` doesn't do any defensive copying and relaying on the fact that the builder gives it an immutable map (which will get returned through the public `getProperties` method).

abaranau:
@chtyim please take another look. @gokulavasan too

abaranau:
http://builds.cask.co/browse/CDAP-DUT2159-4

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/BatchCubeSink.java
```diff
@@ -62,49 +61,74 @@
...
+      this.measurements = measurements;
     }
   }
 
-  private final BatchCubeConfig batchCubeConfig;
+  private final BatchCubeSinkConfig config;
 
-  public BatchCubeSink(BatchCubeConfig realtimeCubeConfig) {
-    this.batchCubeConfig = realtimeCubeConfig;
+  public BatchCubeSink(BatchCubeSinkConfig realtimeCubeConfig) {
```
gokulavasan: realtimeCubeConfig -> config

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/StructuredRecordToCubeFact.java
```diff
@@ -126,63 +83,98 @@ public CubeFact transform(StructuredRecord record) throws Exception {
...
+      this.measurementResolvers = Lists.newArrayList();
+      addMeasurementsFromProperty(props);
+      for (Map.Entry<String, String> property : props.entrySet()) {
+        if (property.getKey().startsWith(Properties.Cube.MEASUREMENT_PREFIX)) {
+          measurementResolvers.add(new MeasurementResolver(property.getKey(), property.getValue()));
+        }
       }
-      if (config.measurements == null) {
-        throw new IllegalArgumentException("Missing required 'measurements' configuration: " + config);
+      if (measurementResolvers.isEmpty()) {
```
gokulavasan: So we will throw these exceptions only after the adapter is started? Is it possible to do these validations during adapter configuration time. DBSource#configurePipeline is a good example.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/sink/RealtimeCubeSink.java
```diff
@@ -64,62 +63,90 @@
...
+    " used in CubeFact. Provide properties as a JSON Map. For example use 'price' field as a measurement of type" +
+    " gauge, and 'count' field as a measurement of type counter, the property should have the value: " +
+    "\"{\"cubeFact.measurement.price\":\"GAUGE\", \"cubeFact.measurement.count\":\"COUNTER\"}.";
 
   /**
-   * Config class for RealtimeCube
+   * Config class for RealtimeCubeSink
    */
-  public static class RealtimeCubeConfig extends PluginConfig {
+  public static class RealtimeCubeSinkConfig extends PluginConfig {
```
gokulavasan: Is the RealtimeCubeSinkConfig any different from BatchCubeSinkConfig? It looks like we can share most of the config, description etc. DBConfig is an example of how a common config is shared between two plugins.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/common/StructuredRecordToCubeFactTest.java
```diff
@@ -184,104 +139,28 @@ public void testTransform() throws Exception {
...
-    config.dimensions = new StructuredRecordToCubeFact.ValueMapping[7];
-    config.dimensions[0] = new StructuredRecordToCubeFact.ValueMapping();
-    config.dimensions[0].name = "dim1";
-    config.dimensions[0].sourceField = "dimField1";
-
-    config.dimensions[1] = new StructuredRecordToCubeFact.ValueMapping();
-    config.dimensions[1].name = "dim2";
-    config.dimensions[1].value = "value2";
+  private Map<String, String> createValidConfig() {
+    Map<String, String> config = Maps.newHashMap();
```
gokulavasan: new HashMap<>() instead of Guava Maps?

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/test/java/co/cask/cdap/template/etl/realtime/RealtimeCubeSinkTest.java
```diff
@@ -0,0 +1,127 @@
...
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ *
+ */
+public class RealtimeCubeSinkTest extends TestBase {
```
gokulavasan: Should we add this test to the ETLWorkerTest. Or we should create a BaseRealtimeTest class similar to BaseETLBatchTest that has the logic to setup the test?

cdap-ui/templates/common/Cube.json
```diff
@@ -1,40 +1,51 @@
...
-            "widget" : "json-editor",
-            "label" : "Cube Configuration",
-            "description" : "Provide the cube configuration"
-         },
-         "custom.properties" : {
+         "dataset.cube.properties" : {
            "widget" : "json-editor",
-           "label"  : "Custom Dataset Properties",
-           "description" : "Provide any custom dataset properties"
+           "label"  : "Other Dataset Properties"
```
gokulavasan: @ajainarayanan Please review the changes to cdap-ui *.json

gokulavasan:
@abaranau Finished review.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/BatchCubeSink.java
abaranau: done

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/StructuredRecordToCubeFact.java
abaranau: calling StructuredRecordToCubeFact ctor for validation in sinks configure

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/sink/RealtimeCubeSink.java
abaranau: true. Moved out

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/common/StructuredRecordToCubeFactTest.java
abaranau: this is etl-lib.. not api. why guava is problem?

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/test/java/co/cask/cdap/template/etl/realtime/RealtimeCubeSinkTest.java
abaranau: no. Instead, as I mentioned, ETLWorkerTest should be broken down into classes!

It is separate and this one will differ from ETLWorkerTest. It has similar one method due to the bug we discussed

abaranau: keeping as is

abaranau:
@gokulavasan please take another look

gokulavasan:
LGTM :+1:

PR #3097, opened 2015-07-06
refines active/hover state styling for slanted tabs

astewes:
* Removes pointer cursor when hovering over active tab
* Removes orange font color when hovering over active tab

active


elfenheart:
:+1: LGTM

PR #3098, opened 2015-07-06
Remove mkdir as HDFS isn't available at bootstrap time

wolf31o2:
Otherwise, EMR installs fail. This was there to prevent a big, ugly exception from showing up in the logs, but doesn't affect overall functionality.


duddun:
LGTM

PR #3100, opened 2015-07-06
Adding missing = in adapter client query creation

rohanahata:


rohanahata:
https://issues.cask.co/browse/CDAP-2932

anwar6953:
I know this is a easy bug fix, but can you also update AdapterClientTestRun.java, so that it would fail without this fix (and then pass with this change)?

cdap-client-tests/src/test/java/co/cask/cdap/client/AdapterClientTestRun.java
```diff
@@ -86,6 +86,12 @@ public void testAdapters() throws Exception {
     adapterClient.waitForStatus(adapterName, AdapterStatus.STOPPED, 30, TimeUnit.SECONDS);
     adapterClient.start(adapterName);
     adapterClient.waitForStatus(adapterName, AdapterStatus.STARTED, 30, TimeUnit.SECONDS);
+
+    // The following test makes sure that if the request is malformed by the client, it will result in an error
```
anwar6953: I don't think you need to specify this comment (describes a no longer existing bug). Just testing it is fine.

anwar6953:
LGTM.
squash commits + merge, once build passes

rohanahata:
http://builds.cask.co/browse/CDAP-DUT2182
https://issues.cask.co/browse/CDAP-2932

PR #3104, opened 2015-07-07
Disallow widget creation, if user specifies uneven number of parts fo…

anwar6953:
…r the metrics context

This is in response to the FIXME from:
https://github.com/caskdata/cdap/commit/86efa3b6d40c8d220486d889031c1fd28c1758a9

### DIFF
--- a/cdap-ui/app/directives/metric-picker/metric-picker.js
+++ b/cdap-ui/app/directives/metric-picker/metric-picker.js
@@ -60,8 +60,13 @@ angular.module(PKG.name + '.commons')
 
         if(attr.required!==undefined) {
           elem.find('input').attr('required', true);
+          // TODO: if the validators fail, propagate this information to the user
           ngModel.$validators.metricAndContext = function (m, v) {
             var t = m || v;
+            // Metrics context requires an even number of parts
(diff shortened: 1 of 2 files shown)


ajainarayanan:
@anwar6953 Technically I can still type in a context (app.ETLBatch.something.something or just app.ETLBatch.. ) and it would let me add the widget. But technically it wouldn't add any widget right? Would we even want the user to type in the context/metrics textboxes? If yes should we validate whatever they have typed is a valid context/metric?

anwar6953:
@ajainarayanan This disallows the user from creating the widget (the green submit button is disabled), if this validation returns false.

ajainarayanan:
@anwar6953 ok. Can we then add a TODO to add a check for valid context and to alert the user that widget is not created because of wrong context. 

Also if we are disabling the add button for odd #of contexts then can we remove the throwing error in the query-helper.js utility function?

ajainarayanan:
If for minor comments :+1: LGTM

anwar6953:
I added the TODO. I think we can leave the error checking in query-helper.js in place, in case that utility method is incorrectly used from elsewhere also (from flow page or overview page, etc).

PR #3105, opened 2015-07-07
Always exclude slf4j-log4j12 bridge, otherwise logback.xml has no eff…

chtyim:
…ect (since all logs get bridged to log4j)


jwang47:
LGTM

PR #3106, opened 2015-07-07
CDAP-2941 fix filter classloader's get resource(s)

albertshau:
The FilterClassLoader should try and use its bootstrap classloader
before doing any filtering. This is so that java system classes
and resources are available. For example, the classloader must be
able to get the javax.script engines as resources.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2169

chtyim:
LGTM

PR #3107, opened 2015-07-07
CDAP-2940 fix bug in artifact concurrent write test case

albertshau:
There was a bug in the test case where the final writer was
being determined incorrectly. Writers add themselves to a list
after they finish writing, and the last writer was assumed to be
the last writer in the list. This is wrong,
since writerX can write after writerY, but writerX may add itself
to the list before writerY adds itself to the list.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2170

gokulavasan:
LGTM :+1:

PR #3111, opened 2015-07-08
Create root ZK node right after connected to ZK.

anwar6953:
https://issues.cask.co/browse/CDAP-2569
http://builds.cask.co/browse/CDAP-DUT2176-1


chtyim:
LGTM

PR #3112, opened 2015-07-08
(CDAP-2931) Get location of spark-assembly.jar from environment

chtyim:
- Support direct location through $SPARK_ASSEMBLY_JAR
- Support search in $SPARK_HOME/lib
- Localize the spark jar to the spark client container in distributed mode

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkUtils.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkUtils.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.internal.app.runtime.spark;
+
+import com.google.common.base.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+
+/**
+ * A utility class to help determine Spark supports and locating Spark jar.
+ */
+public final class SparkUtils {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SparkUtils.class);
+
+  // Environment variable name for locating spark assembly jar file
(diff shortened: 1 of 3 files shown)


chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2121

Although the changes doesn't affect any unit-test at all since it only affect spark in distributed mode.

henry-cask:
Code looks good but any way to test this in unit or functional test?

chtyim:
I tested it on a cluster. Need to add regular integration test for Spark (not entirely related to this change, but be able to run spark on the cluster means this change works).

rohitsinha54:
LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkUtils.java
```diff
@@ -0,0 +1,120 @@
...
+    // It's useful for overridding what being set for SPARK_HOME
+    String jarEnv = System.getenv(SPARK_ASSEMBLY_JAR);
+    if (jarEnv != null) {
+      File file = new File(jarEnv);
+      if (file.isFile()) {
+        LOG.info("Located Spark Assembly JAR in {}", file);
+        sparkAssemblyJar = file;
+        return file;
+      }
+    }
```
wolf31o2: Should we WARN or anything if this is set but it does **not** point to the JAR?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkUtils.java
```diff
@@ -0,0 +1,120 @@
...
+                                        "Please set environment variable " + SPARK_HOME + " or " + SPARK_ASSEMBLY_JAR);
+    }
+
+    // Look for spark-assembly.jar symlink
+    Path assemblyJar = Paths.get(sparkHome, "lib", "spark-assembly.jar");
+    if (Files.isSymbolicLink(assemblyJar)) {
+      sparkAssemblyJar = assemblyJar.toFile();
+      return sparkAssemblyJar;
+    }
+
```
wolf31o2: What if it isn't a symlink? I have seen this, especially with the mysql-connector, that people will copy the JAR as mysql-connector-java.jar versus doing a symlink.

wolf31o2:
Minor comment, nothing blocking... :smile:

henry-cask:
Is there any JIRA to add that integration test? If not then could you file it to make sure it is not forgotten when this PR is merged?

Other than that +1 LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkUtils.java
chtyim: The current logic doesn't check file extension nor the file content to see if it s a JAR or not. Maybe we should check the file extension? Not sure if worth it.

wolf31o2: I meant if it was invalid.
`export SPARK_ASSEMBLY_JAR=/tmp/not-a-file.whatever`

We log that we found the JAR if it exists and is a file. I think that we should log if the variable is pointed to an invalid location. This would still continue to doing the search to find the JAR, but would alert the user that there's an incorrect setting in their environment. We're already doing the right thing and moving on, there's just no indication the user that the SPARK_ASSEMBLY_JAR setting from their environment is being ignored.

chtyim: Ar. I see what you mean now. Good point. will add a log.

chtyim:
created JIRA https://issues.cask.co/browse/CDAP-2964 for integration test.

chtyim:
Build green again: http://builds.cask.co/browse/CDAP-DUT2121-34

Merging.

PR #3113, opened 2015-07-08
Moving the Stream Event Record classes into a cdap-formats module

ataylor4:
Moved the Stream Event Record classes into their own module with minimal dependencies.

http://builds.cask.co/browse/CDAP-DUT2177-2


cdap-formats/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+    <version>3.1.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>cdap-formats</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-watchdog-api</artifactId>
```
chtyim: Should be cdap-api, not watchdog-api.

cdap-formats/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+      <groupId>io.thekraken</groupId>
+      <artifactId>grok</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-common</artifactId>
```
chtyim: Shouldn't depends on hbase. If you need avro, just depends on avro only.

pom.xml
```diff
@@ -23,6 +23,9 @@
   <groupId>co.cask.cdap</groupId>
   <artifactId>cdap</artifactId>
   <version>3.1.0-SNAPSHOT</version>
+  <modules>
+    <module>cdap-formats</module>
```
chtyim: Don't add it here. Add it near the end of this file, under the profile `<id>default</id>`.

cdap-formats/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+      <artifactId>grok</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.avro</groupId>
+      <artifactId>avro</artifactId>
+      <version>${avro.version}</version>
```
chtyim: You don't need the `version` and `exclusions` session. They comes from dependency management.

pom.xml
```diff
@@ -2067,6 +2067,7 @@
         <module>cdap-test</module>
         <module>cdap-integration-test</module>
         <module>cdap-ui</module>
+        <module>cdap-formats</module>
```
chtyim: Move it to higher position, right after `cdap-api`, since the only inter-module dependency it has is only `cdap-api`.

chtyim:
Just one comment left, otherwise LGTM.

PR #3117, opened 2015-07-08

ajainarayanan:
DagreD3 updated the version of lodash from 2.* to 3.* straight away.  (https://github.com/cpettitt/dagre-d3/commit/e746821ab29e0cab58dc8a9f12a12f475ff2f0bc)
Lodash changed its built files location from dist/lodash.js to just ./lodash.js  (https://github.com/lodash/lodash/wiki/Changelog#v300)

So when we took an update from DagreD3 we didn't update the path of lodash. That is the reason the UI was failing.

TODO: 
  Remove epoch.js and protractor completely in a separate PR


elfenheart:
:+1: LGTM

PR #3118, opened 2015-07-08
Programs can create streams/datasets

gokulavasan:
Changes in Programmatic API to allow creation of stream and datasets in Program's configure method.

Note: 
1. Validation of whether dataset/stream is added with the same name but different properties is not part of this PR. Opening this PR to get feedback on the approach and public API changes.
2. Workflow (and WorkflowAction) can't create streams/datasets. It is blocked because of this: https://issues.cask.co/browse/CDAP-2956

JIRA : https://issues.cask.co/browse/CDAP-2777

Build : http://builds.cask.co/browse/CDAP-DUT2167-4


cdap-api/src/main/java/co/cask/cdap/api/ProgramConfigurer.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package co.cask.cdap.api;
+
+import java.util.Map;
+
+/**
+ * Basic Program Configurer.
+ */
+public interface ProgramConfigurer {
```
chtyim: Is it specific to Program? Flowlet configurer also fits for extending from this interface, yet it's not a program (Flow is a program)

anew: But we also have FlowletProgramRunner :)

gokulavasan:
In this PR, the programs delegate the addition of datasets, streams to ApplicationConfigurer and thus the ApplicationSpecification contains all the streams, datasets information and the ProgramSpecification doesn't. Other option is to add these streams, datasets info to individual ProgramSpecifications.

gokulavasan:
Closing this PR. Will reopen for 3.2.

PR #3119, opened 2015-07-08
Add test case for TimePartitionedFileSet#addMetadata methods

anwar6953:
Simply add test case for unused methods, even though the implementation is simple.

http://builds.cask.co/browse/CDAP-DUT2183-2


cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/partitioned/TimePartitionedFileSetTest.java
```diff
@@ -85,15 +85,37 @@ public void apply() throws Exception {
...
         Assert.assertNotNull(partitionByTime);
         Assert.assertEquals(metadata, partitionByTime.getMetadata().asMap());
+
+        tpfs.addMetadata(time, "key3", "value4");
+        allMetadata.put("key3", "value4");
+
+        try {
+          // attempting to update an existing key throws a DatasetException
+          tpfs.addMetadata(time, "key3", "value5");
+        } catch (DataSetException expected) {
```
poornachandra: It would be a good idea to have `Assert.fail()` if the exception is not thrown.

poornachandra:
Just one comment, fix and merge.

PR #3121, opened 2015-07-08
CDAP-2955: NoSuchMethodException when trying to explore Datasets/Stream

jwang47:
- Hive13ExploreService.doFetchNextResults wasn't getting the
  correct fetchResults method

- [x] Verified that the NoSuchMethodException no longer occurs in distributed cluster on Hive 0.13


poornachandra:
This change itself looks good, but how do we make sure there are no other remaining issues across all the Hive versions we support? Do we have integration tests for Explore where we can run a standard set of tests across all Hive versions we support?

jwang47:
No integration tests currently. Need to work on that next.

poornachandra:
I see that @sreevatsanraman filed JIRA https://issues.cask.co/browse/CDAP-2960 for explore integration tests. So this PR can be merged.

PR #3116, opened 2015-07-08
Feature/ui adapter details

elfenheart:
Surfacing Adapter Properties on already created Application Template

screen shot 2015-07-08 at 11 29 28 am

PR #3124, opened 2015-07-09
Feature/ui adapter prototype integration 2

ajainarayanan:
- Modifies MyPlumbService to store list of nodes and connections.
- Has a better laid-out implementation of communicating between plumb and side-panel directives.
- Will be able to add property edits as soon as this PR (https://github.com/caskdata/cdap/pull/3116) is merged.

editproperties

#### TODO:
Needs to some heavy lifting to be done in styling.


cdap-ui/app/services/plumber/my-plumb-service.js
```diff
@@ -1,37 +1,113 @@
...
+      var plugin = this.nodes[pluginId];
+      var prom = $q.defer();
+      var propertiesApiMap = {
+        'source': myAdapterApi.fetchSourceProperties,
+        'sink': myAdapterApi.fetchSinkProperties,
+        'transform': myAdapterApi.fetchTransformProperties
+      };
+      var params = {
+        scope: scope,
+        adapterType: 'ETLBatch'
```
elfenheart: This only applies to ETLBatch?

elfenheart:
one minor comment, otherwise :+1: LGTM

PR #3125, opened 2015-07-09
Revert "Moving the Stream Event Record classes into a cdap-formats mo…

ataylor4:
Reverts caskdata/cdap#3113 
Testing to find source of cluster error


ataylor4:
The PR this reverted is not the source of the cluster error (of hbase-default.xml null)

PR #3127, opened 2015-07-09
CDAP-2665 add artifact repository

albertshau:
The ArtifactRepository is a layer on top of the ArtifactStore
and is responsible for inspecting artifacts to determine the
metadata for the artifact before writing it to the store.
This will eventually replace the PluginRepository, though there
is some more refactoring required before that is possible.
For now, extracting jar inspection logic to a common
ArtifactInspector class that both repositories use, and refactoring
plugin instantiation and classloading so that it can be used by
both repositories.

Changing the API of the ArtifactStore a little so that it
returns sorted maps so that it is easier to select which
artifact to use for a specific plugin.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2190

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -0,0 +1,179 @@
...
+   *
+   * @param artifactId the id of the artifact to inspect and store
+   * @param artifactFile the artifact to inspect and store
+   * @param parentArtifacts artifacts the given artifact extends.
+   *                        If null, the given artifact does not extend another artifact
+   * @throws IOException if there was an exception reading from the artifact store
+   * @throws ArtifactRangeNotFoundException if none of the parent artifacts could be found
+   */
+  @VisibleForTesting
+  void inspectArtifact(Id.Artifact artifactId, File artifactFile, @Nullable Set<ArtifactRange> parentArtifacts)
```
gokulavasan: As discussed, this should be public and we can rename this method to addArtifact?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -0,0 +1,179 @@
...
+    if (parentArtifacts.isEmpty()) {
+      // if this artifact doesn't extend another, use itself to create the parent classloader
+      parentClassLoader = createArtifactClassLoader(Locations.toLocation(artifactFile));
+    } else {
+      // otherwise, use any of the parent artifacts to create the parent classloader.
+      List<ArtifactDetail> parents = artifactStore.getArtifacts(parentArtifacts);
+      if (parents.isEmpty()) {
+        throw new ArtifactRangeNotFoundException(parentArtifacts);
+      }
+      parentClassLoader = createArtifactClassLoader(parents.get(0).getDescriptor().getLocation());
```
gokulavasan: Why use only the first parent artifact?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -162,13 +164,43 @@
...
+   *         is returned
+   */
+  public List<ArtifactDetail> getArtifacts(final Set<ArtifactRange> ranges) {
+    return metaTable.executeUnchecked(new TransactionExecutor.Function<DatasetContext<Table>, List<ArtifactDetail>>() {
+      @Override
+      public List<ArtifactDetail> apply(DatasetContext<Table> context) throws Exception {
+        List<ArtifactDetail> artifacts = Lists.newArrayList();
+        Table table = context.get();
+        for (ArtifactRange range : ranges) {
+          ArtifactKey artifactKey = new ArtifactKey(range.getNamespace(), range.getName());
```
gokulavasan: ArtifactRange contains namespace, artifact name and [lower, upper) versions. Do we use the lower and upper version info? Also the list returned back contains the ArtifactDetail of all artifacts that are within the lower, upper versions? Would it be useful to return Map<(Namespace+ArtifactName), List<ArtifactDetail>> ?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepositoryTest.java
```diff
@@ -0,0 +1,238 @@
...
+      @Override
+      public Map.Entry<ArtifactDescriptor, PluginClass> select(SortedMap<ArtifactDescriptor, PluginClass> plugins) {
+        return plugins.entrySet().iterator().next();
+      }
+    });
+    Assert.assertNotNull(plugin);
+    Assert.assertEquals(new ArtifactVersion("1.0"), plugin.getKey().getVersion());
+    Assert.assertEquals("TestPlugin2", plugin.getValue().getName());
+
+    // Load the Plugin class and the "EmptyClass" again from the current plugin selected
```
gokulavasan: As discussed, EmptyClass is not applicable here.

gokulavasan:
Few comments. Rest LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
albertshau: the assumption is that any of the parents will work, otherwise it is a user error in specifying the wrong parents. If I say etl-lib extends etl-batch versions [1.0.0 , 2.0.0), any etl-batch within the range should work.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: good catch, forgot to add the version check.

this is also kind of a weird api, going to change it to take a single range instead of a set.

albertshau: will also add a unit test

albertshau:
@gokulavasan addressed comments, please take another look

gokulavasan:
LGTM :+1:

PR #3128, opened 2015-07-09
Added scoping in the WorkflowToken api.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-2894

Link to the document - https://wiki.cask.co/display/CE/WorkflowToken+Enhancements

Pending thing to be added in separate PR: Utility method which accepts the WorkflowToken and program name and give the MapReduce counters.


sagarkapare:
@anew @bdmogal @chtyim Please help review. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
```diff
@@ -0,0 +1,81 @@
...
+
+package co.cask.cdap.api.workflow;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Class representing the value of the key in the {@link WorkflowToken}.
+ */
+public class Value {
```
bdmogal: Should we add ``getAsInt`` and ``getAsBoolean``?

bdmogal: oops, you already have ``getAsInt``.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -23,39 +25,80 @@
...
+
+  /**
+   * Put the specified key and value into the {@link WorkflowToken}.
    * The token may store additional information about the context in which
    * this key is being set, for example, the unique name of the workflow node.
-   * @param key   the key representing the entry
+   * @param key the key representing the entry
    * @param value the value for the key
    */
+  // TODO [CDAP-2895] put operation should throw certain exceptions
```
bdmogal: Can remove this TODO from here, no? We can fix that jira later, but I don't think it applies here, since its going to be a runtime exception?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -89,9 +171,18 @@
...
    */
   boolean containsKey(String key);
+
+  /**
+   * Return true if the {@link WorkflowToken} contains the specified key.
+   * @param key the key to be tested for the presence in the {@link WorkflowToken}
+   * @param scope the {@link WorkflowToken.Scope} for the key
+   * @return the result of the test
+   */
+  boolean containsKey(String key, Scope scope);
```
bdmogal: If we disallow ``null``, then we don't need these ``containsKey`` methods

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/WorkflowTokenCodec.java
```diff
@@ -0,0 +1,73 @@
...
+    JsonObject jsonObj = json.getAsJsonObject();
+
+    Type mapReduceCounterType = new TypeToken<Map<String, Map<String, Long>>>() { }.getType();
+    Map<String, Map<String, Long>> mapReduceCounters = context.deserialize(jsonObj.get("mapReduceCounters"),
+                                                                           mapReduceCounterType);
+
+    String nodeName = jsonObj.get("nodeName").getAsString();
+
+    Table<WorkflowToken.Scope, String, List<NodeValueEntry>> tokenValueMap = HashBasedTable.create();
+    JsonElement tableBackingMap = ((JsonObject) jsonObj.get("tokenValueMap")).get("backingMap");
```
bdmogal: can ``((JsonObject) jsonObj.get("tokenValueMap"))`` ever return ``null``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/WorkflowTokenCodec.java
```diff
@@ -0,0 +1,73 @@
...
+    if (tableBackingMap != null) {
+      for (Map.Entry<String, JsonElement> entry : ((JsonObject) tableBackingMap).entrySet()) {
+        WorkflowToken.Scope row = WorkflowToken.Scope.valueOf(entry.getKey());
+        for (Map.Entry<String, JsonElement> column : ((JsonObject) entry.getValue()).entrySet()) {
+          Type listType = new TypeToken<List<NodeValueEntry>>() {
+          }.getType();
```
bdmogal: nit: can move to previous line. Also, do you need to create it for every iteration?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -32,26 +35,31 @@
...
-  private final Map<String, List<NodeValueEntry>> tokenValueMap = Maps.newHashMap();
+  private final Table<Scope, String, List<NodeValueEntry>> tokenValueMap = HashBasedTable.create();
   private String nodeName;
 
   public BasicWorkflowToken() {
   }
 
-  public BasicWorkflowToken(BasicWorkflowToken other) {
-    for (Map.Entry<String, List<NodeValueEntry>> entry : other.tokenValueMap.entrySet()) {
+  public BasicWorkflowToken(Table<Scope, String, List<NodeValueEntry>> tokenValueMap, String nodeName,
```
bdmogal: this can be package-private, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -32,26 +35,31 @@
...
+    this.nodeName = nodeName;
 
-    if (other.mapReduceCounters != null) {
-      this.mapReduceCounters = copyHadoopCounters(other.mapReduceCounters);
+    if (mapReduceCounters != null) {
+      this.mapReduceCounters = copyHadoopCounters(mapReduceCounters);
     }
   }
 
+  public BasicWorkflowToken(BasicWorkflowToken other) {
```
bdmogal: Where is this used? Can we remove it?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,55 +99,69 @@ void mergeToken(BasicWorkflowToken other) {
...
+        nodeValueList.set(i, new NodeValueEntry(nodeName, new Value(value)));
         return;
       }
     }
 
-    nodeValueList.add(new NodeValueEntry(nodeName, value));
+    nodeValueList.add(new NodeValueEntry(nodeName, new Value(value)));
   }
 
+  @Nullable
```
bdmogal: If this method returns null, we should move the ``@Nullable`` annotation to the interface.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,55 +99,69 @@ void mergeToken(BasicWorkflowToken other) {
...
-  public List<NodeValueEntry> getAll(String key) {
-    if (containsKey(key)) {
-      return ImmutableList.copyOf(tokenValueMap.get(key));
-    }
-    return ImmutableList.of();
+  public Value get(String key) {
+    return get(key, Scope.USER);
   }
 
+  @Nullable
```
bdmogal: same here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,55 +99,69 @@ void mergeToken(BasicWorkflowToken other) {
...
+    if (nodeValueList.isEmpty()) {
+      // List of NodeValueEntry cannot be empty if the key is added in the WorkflowToken as
+      // when we add key, we also add single NodeValueEntry.
+      throw new IllegalStateException(String.format("List of NodeValueEntry for the key %s cannot be empty", key));
     }
-    return tokenValuesFromNode;
+    return nodeValueList.get(nodeValueList.size() - 1).getValue();
   }
 
   @Nullable
```
bdmogal: same here

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,55 +99,69 @@ void mergeToken(BasicWorkflowToken other) {
...
   @Override
-  public String get(String key, String nodeName) {
-    List<NodeValueEntry> nodeValueList = tokenValueMap.get(key);
+  public Value get(String key, String nodeName) {
+    return get(key, nodeName, Scope.USER);
+  }
+
+  @Nullable
```
bdmogal: same here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,55 +99,69 @@ void mergeToken(BasicWorkflowToken other) {
...
+    }
 
-      List<NodeValueEntry> nodeValueEntryList = entry.getValue();
-      for (NodeValueEntry nodeValueEntry : nodeValueEntryList) {
-        if (nodeValueEntry.getNodeName().equals(nodeName)) {
-          tokenValuesFromNode.put(entry.getKey(), nodeValueEntry.getValue());
-          break;
-        }
-      }
+    if (nodeValueList.isEmpty()) {
```
bdmogal: can we use ``Preconditions.checkState`` instead?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,55 +99,69 @@ void mergeToken(BasicWorkflowToken other) {
...
   }
 
+  @Nullable
   @Override
-  public List<NodeValueEntry> getAll(String key) {
-    if (containsKey(key)) {
-      return ImmutableList.copyOf(tokenValueMap.get(key));
-    }
-    return ImmutableList.of();
+  public Value get(String key) {
```
bdmogal: Since this class has a bunch of variants of the same method that delegate to other methods, would it be better to turn ``WorkflowToken`` into an abstract class with these default implementations, so ``BasicWorkflowToken`` has to implement fewer methods? 

I'm unsure if that makes the code more readable, or if it introduces backward-incompatibility though. What do you think?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -177,8 +221,13 @@ public String get(String key) {
   }
 
   @Override
+  public boolean containsKey(String key, Scope scope) {
+    return tokenValueMap.contains(scope, key);
+  }
+
+  @Override
   public boolean containsKey(String key) {
```
bdmogal: nit: to make it consistent, this method should move above the previous method.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -25,6 +25,7 @@
 import co.cask.cdap.WorkflowAppWithFork;
 import co.cask.cdap.WorkflowAppWithScopedParameters;
 import co.cask.cdap.api.schedule.ScheduleSpecification;
+import co.cask.cdap.api.workflow.Value;
```
bdmogal: unused?

cdap-app-fabric/src/test/java/co/cask/cdap/ConditionalWorkflowApp.java
```diff
@@ -186,29 +161,7 @@ public void beforeSubmit(MapReduceContext context) throws Exception {
...
-                                          " is started from the Workflow.");
-      }
-
-      Counters counters = ((Job) context.getHadoopJob()).getCounters();
-      for (CounterGroup group : counters) {
-        for (Counter counter : group) {
-          workflowToken.put(group.getName() + "." + counter.getName(), String.valueOf(counter.getValue()));
-        }
-      }
+      context.getWorkflowToken().put("start_time", Value.of(System.currentTimeMillis()));
```
bdmogal: did you mean ``end_time``, or am I missing something?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -32,26 +35,31 @@
  */
 public class BasicWorkflowToken implements WorkflowToken {
   private Map<String, Map<String, Long>> mapReduceCounters;
-  private final Map<String, List<NodeValueEntry>> tokenValueMap = Maps.newHashMap();
+  private final Table<Scope, String, List<NodeValueEntry>> tokenValueMap = HashBasedTable.create();
```
bdmogal: nit: Since this is not a ``Map`` anymore, maybe call it ``tokenValueTable``? Or even just ``data``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -32,26 +35,31 @@
...
   private String nodeName;
 
   public BasicWorkflowToken() {
   }
 
-  public BasicWorkflowToken(BasicWorkflowToken other) {
-    for (Map.Entry<String, List<NodeValueEntry>> entry : other.tokenValueMap.entrySet()) {
+  public BasicWorkflowToken(Table<Scope, String, List<NodeValueEntry>> tokenValueMap, String nodeName,
+                            @Nullable Map<String, Map<String, Long>> mapReduceCounters) {
+    for (Table.Cell<Scope, String, List<NodeValueEntry>> cell : tokenValueMap.cellSet()) {
```
bdmogal: hmmm, there seems to be a [Table#putAll](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Table.html#putAll(com.google.common.collect.Table) method. Can you use it instead of this loop?

bdmogal: Although, why do we need copying? Is there a problem with just assigning the parameter to the member?

cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
```diff
@@ -0,0 +1,81 @@
...
+  private final String value;
+  private static final Logger LOG = LoggerFactory.getLogger(Value.class);
```
chtyim: unused field?

cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public Value(String value) {
```
chtyim: This should be private and have a static `Value.of(String)` to be consistent.

cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
```diff
@@ -0,0 +1,81 @@
...
+    Value value1 = (Value) o;
+    return value.equals(value1.value);
+  }
+
+  @Override
+  public int hashCode() {
+    return value.hashCode();
+  }
+
+  public static Value of(int value) {
```
chtyim: Javadoc needed for public method in API module.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -23,39 +25,80 @@
...
+   * this key is being set, for example, the unique name of the workflow node.
+   * @param key the key representing entry
+   * @param value the {@link Value} for the key
+   */
+  void put(String key, Value value);
+
+  /**
+   * Get the most recent value added for the specified key for a {@link Scope#USER} scope.
+   * @param key the key to be searched
+   * @return the {@link Value} for the key
```
chtyim: Need to define when return `null` as well.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -23,39 +25,80 @@
...
+   */
+  @Nullable
+  Value get(String key);
+
+  /**
+   * Get the most recent value for the specified key for a given scope.
    * @param key the key to be searched
-   * @return the value for the key
+   * @param scope the {@link WorkflowToken.Scope} for the key
+   * @return the {@link Value} for the key from the specified scope
```
chtyim: same here. Check javadoc of all `@Nullable` method.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -71,12 +114,51 @@
...
+
+  /**
+   * Same key can be added to the WorkflowToken by multiple nodes.
+   * This method returns the key to {@link List} of {@link NodeValueEntry}
+   * added in the {@link WorkflowToken.Scope} provided.
+   * @param scope the scope for the key
+   * @return the {@link Map} of key to {@link List} of {@link NodeValueEntry} added for
+   * the given scope
+   */
+  Map<String, List<NodeValueEntry>> getAll(Scope scope);
```
chtyim: To be consistent, there should be `getAll()` as well to return a map for all user scope tokens.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
chtyim: :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/WorkflowTokenCodec.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public class WorkflowTokenCodec extends AbstractSpecificationCodec<WorkflowToken> {
+  @Override
+  public JsonElement serialize(WorkflowToken src, Type typeOfSrc, JsonSerializationContext context) {
+    return context.serialize(src);
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public WorkflowToken deserialize(JsonElement json, Type typeOfT,
```
chtyim: This method seems more complicated than it needs. Is there a simpler form? Maybe do something at the serialize end?

chtyim: E.g. if you make `BasicWorkflowToken` more Gson friendly, by using `EnumMap<Scope, Map<String, List<NodeEntryValue>>>`.

cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
sagarkapare: Added `getAsBoolean` as well.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Removed TODO

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Removed these methods and now not allowing null key or value in the WorkflowToken.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/WorkflowTokenCodec.java
sagarkapare: Removed this codec.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: This ctor is using while making `deepCopy`. Made it private though.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: This was used from codec which is removed now. Hence removed this ctor as well.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Removed `@Nullable` and `@Deprecated` method annotations from the implementation class.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: sure. fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Removed these methods.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: Removed.

cdap-app-fabric/src/test/java/co/cask/cdap/ConditionalWorkflowApp.java
sagarkapare: No. Its `start_time` only. It is just the key that is added in the beforeSubmit of the MapReduce programs and used to verify.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Its better to do copy otherwise changes in the parameter will affect this member variable.

cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
sagarkapare: Removed.

cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
sagarkapare: Added `Value.of(String)` and made the ctor private.

cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
sagarkapare: Added javadocs.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Fixed the documentation.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
sagarkapare: Added `getAll()` method.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/WorkflowTokenCodec.java
sagarkapare: Using `EnumMap` now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: I think its better not to expose the implementation in the API. Also we are not getting any advantage by using abstract class since `BasicWorkflowToken` is the only default implementation that we have now.

sagarkapare:

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -71,12 +117,60 @@
   List<NodeValueEntry> getAll(String key);
```
bdmogal: If the key is not found, does this return null? If so, mark this as ``@Nullable`` too

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -71,12 +117,60 @@
...
+   * and passed along each branch. At the join, all copies of the
+   * WorkflowToken are merged together. While merging, the order in which the values were
+   * inserted for a specific key is guaranteed within the same branch, but not across
+   * different branches.
+   * @param key the key to be searched
+   * @param scope the {@link WorkflowToken.Scope} for the key
+   * @return the list of {@link NodeValueEntry} from node name to the value that node
+   * added for the input key for a given scope
+   */
+  List<NodeValueEntry> getAll(String key, Scope scope);
```
bdmogal: same here

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -71,12 +117,60 @@
...
+
+  /**
+   * Get the {@link Map} of key to {@link Value}s that were added to the {@link WorkflowToken}
+   * by specific node for a {@link Scope#USER} scope.
    * @param nodeName the unique name of the node
-   * @return the map of key-values that were added by the specified node
+   * @return the map of key to values that were added by the specified node
    */
-  Map<String, String> getAllFromNode(String nodeName);
+  Map<String, Value> getAllFromNode(String nodeName);
```
bdmogal: what happens if ``nodeName`` is not found

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
```diff
@@ -71,12 +117,60 @@
...
+
+  /**
+   * Get the {@link Map} of key to {@link Value}s that were added to the {@link WorkflowToken}
+   * by specific node for a given scope.
+   * @param nodeName the unique name of the node
+   * @param scope the {@link WorkflowToken.Scope} for the key
+   * @return the map of key to values that were added by the specified node for a given scope
+   */
+  Map<String, Value> getAllFromNode(String nodeName, Scope scope);
```
bdmogal: same here

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
bdmogal: ignore: checked that it returns empty list

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowToken.java
bdmogal: Ignore again. empty map.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
bdmogal: hmmm.. if you changed this, it does not show up.

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
```diff
@@ -18,14 +18,14 @@
 
 /**
  * Multiple nodes in the Workflow can add the same key to the {@link WorkflowToken}.
- * This class provides a mapping from node name to the value which was set for the
+ * This class provides a mapping from node name to the {@link Value} which was set for the
  * specific key.
  */
 public final class NodeValueEntry {
```
bdmogal: Can this class be an inner class of ``WorkflowToken``?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -17,32 +17,37 @@
...
   }
 
-  public BasicWorkflowToken(BasicWorkflowToken other) {
-    for (Map.Entry<String, List<NodeValueEntry>> entry : other.tokenValueMap.entrySet()) {
-      List<NodeValueEntry> nodeValueList = Lists.newArrayList();
-      nodeValueList.addAll(entry.getValue());
-      this.tokenValueMap.put(entry.getKey(), nodeValueList);
+  private BasicWorkflowToken(BasicWorkflowToken other) {
+    for (Map.Entry<Scope, Map<String, List<NodeValueEntry>>> entry : other.tokenValueMap.entrySet()) {
+      Map<String, List<NodeValueEntry>> tokenValueMapByScope = Maps.newHashMap();
```
bdmogal: can just use ``new HashMap<>()`` now, with the Java 7 syntax

bdmogal: also nit: ``tokenValueMapForScope``, since this is for a single scope. 'by' suggests that this data structure contains multiple token values keyed by scope

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -62,29 +67,37 @@ void setCurrentNode(String nodeName) {
...
-      // Iterate over the list of NodeValueEntry corresponding to the current key.
-      // Only add those NodeValueEntry to the merged token which already do not exist.
-      for (NodeValueEntry otherNodeValueEntry : otherTokenValueMap.get(entry.getKey())) {
-        boolean otherNodeValueEntryExist = false;
-        for (NodeValueEntry thisNodeValueEntry : tokenValueMap.get(entry.getKey())) {
-          if (thisNodeValueEntry.equals(otherNodeValueEntry)) {
-            otherNodeValueEntryExist = true;
-            break;
-          }
+      Map<String, List<NodeValueEntry>> thisTokenValueMapByScope = this.tokenValueMap.get(entry.getKey());
```
bdmogal: nit: s/By/For/

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -62,29 +67,37 @@ void setCurrentNode(String nodeName) {
...
+
+      for (Map.Entry<String, List<NodeValueEntry>> otherTokenValueMapByScopeEntry : entry.getValue().entrySet()) {
```
bdmogal: same here

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -62,29 +67,37 @@ void setCurrentNode(String nodeName) {
...
+        if (!thisTokenValueMapByScope.containsKey(otherTokenValueMapByScopeEntry.getKey())) {
+          thisTokenValueMapByScope.put(otherTokenValueMapByScopeEntry.getKey(), Lists.<NodeValueEntry>newArrayList());
```
bdmogal: can use shorter Java 7 syntax here too

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,13 +105,30 @@ void mergeToken(BasicWorkflowToken other) {
...
+
+  @Override
+  public void put(String key, Value value) {
+    put(key, value, Scope.USER);
+  }
+
+  void put(String key, Value value, Scope scope) {
+    Preconditions.checkNotNull(key, "Null key cannot be added in the WorkflowToken.");
+    Preconditions.checkNotNull(value, String.format("Null value provided for the key '%s'.", key));
+    Preconditions.checkNotNull(value.toString(), String.format("Null value provided for the key '%s'.", key));
```
bdmogal: move this check to the ``Value`` class?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,13 +105,30 @@ void mergeToken(BasicWorkflowToken other) {
...
+    Preconditions.checkState(nodeName != null);
```
bdmogal: Add a message here?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -92,13 +105,30 @@ void mergeToken(BasicWorkflowToken other) {
...
+
+    Map<String, List<NodeValueEntry>> tokenValuesWithinScope = tokenValueMap.get(scope);
+    if (tokenValuesWithinScope == null) {
+      tokenValuesWithinScope = Maps.newHashMap();
```
bdmogal: use Java 7 syntax?

bdmogal: However, will the code look cleaner if you initialize ``tokenValueMap`` for both ``Scope.USER`` and ``SCOPE.SYSTEM`` in the constructor itself? That way, you won't have to do this null check here and perhaps at other places too.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
bdmogal: fair enough. we can revisit if we have more implementations (unlikely)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -153,32 +183,53 @@ public String get(String key, String nodeName) {
...
-      throw new IllegalStateException(String.format("List of NodeValueEntry for the key %s cannot be empty", key));
+  @Override
+  public Map<String, Value> getAllFromNode(String nodeName) {
+    return getAllFromNode(nodeName, Scope.USER);
+  }
+
+  @Override
+  public Map<String, Value> getAllFromNode(String nodeName, Scope scope) {
+    Map<String, Value> tokenValuesFromNode = Maps.newHashMap();
+    for (Map.Entry<String, List<NodeValueEntry>> entry : tokenValueMap.get(scope).entrySet()) {
```
bdmogal: ``tokenValueMap.get(scope)`` could be null here. However,  doing https://github.com/caskdata/cdap/pull/3128/files#r34424148 will fix it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Its better to include the `key` in the exception, hence having these checks here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: sure. done.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Good point! added the initialization in the constructor.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Using the Java 7 syntax.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: renamed.

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
sagarkapare: Yes. It can be. However not sure what will be the benefit. Maybe `Entry` in the name of the class is confusing, since this is not the key-value entry but only represents the value part. Perhaps rename the class to `NodeValue` and keep it outside of the `WorkflowToken`?

sagarkapare:
@bdmogal Addressed comments with one question. Please take a look again. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
bdmogal: Fair enough. Don't feel too strongly about it.

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT2191-5

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
chtyim: The differences between checking in here vs checkin the `Value.of(String)` is whether `null` is not permitted as a token vs `null` is not permitted as a `Value`.

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
chtyim: Make sense to call it `NodeValue`, because it's a class containing node name and value.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -153,32 +176,53 @@ public String get(String key, String nodeName) {
...
+      List<NodeValueEntry> nodeValueEntryList = entry.getValue();
+      for (NodeValueEntry nodeValueEntry : nodeValueEntryList) {
+        if (nodeValueEntry.getNodeName().equals(nodeName)) {
+          tokenValuesFromNode.put(entry.getKey(), nodeValueEntry.getValue());
+          break;
+        }
+      }
     }
-    return nodeValueList.get(nodeValueList.size() - 1).getValue();
+    return tokenValuesFromNode;
```
chtyim: So this method returns a new, mutable map. Other methods in this class returns immutable collection/map. It's better be consistent.

cdap-app-fabric/src/test/java/co/cask/cdap/ConditionalWorkflowApp.java
chtyim: Can you call it "start.time" instead? I know it's just a unit-test, however someone will eventually copy the concept and make it into example, which using "." looks much nicer and is the convention in hadoop and cdap.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Returning ImmutableMap now.

cdap-api/src/main/java/co/cask/cdap/api/workflow/NodeValueEntry.java
sagarkapare: Renamed to `NodeValue`.

sagarkapare:
@bdmogal @chtyim Can you please take a one more look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
bdmogal: Ok. I think its ok for ``Value`` to continue to accept ``null``, but for ``WorkflowToken`` to make sure its non-null for now.

bdmogal:
:+1:

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT2191-6. Merging.

PR #3130, opened 2015-07-10
[CDAP-2915] SQS realtime source

jayfjin:
JIRA - https://issues.cask.co/browse/CDAP-2915
BUILD - http://builds.cask.co/browse/CDAP-DUT2187-2

Realtime SQS source for ETL. Used in Caskalytics. Tested and works on clusters (with etlrealtimesourcestate disabled).

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
@@ -191,6 +191,21 @@
     </dependency>
     <!-- End for Kafka Source -->
     <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>amazon-sqs-java-messaging-lib</artifactId>
+      <version>1.0.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.elasticmq</groupId>
+      <artifactId>elasticmq-core_2.10</artifactId>
+      <version>0.6.3</version>
+    </dependency>
+    <dependency>
+      <groupId>org.elasticmq</groupId>
+      <artifactId>elasticmq-rest-sqs_2.10</artifactId>
+      <version>0.6.3</version>
+    </dependency>
+    <dependency>
       <groupId>com.twitter</groupId>
       <artifactId>parquet-avro</artifactId>
       <version>1.6.0</version>
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
@@ -1,0 +1,175 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.template.etl.realtime.source;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.templates.plugins.PluginConfig;
+import co.cask.cdap.template.etl.api.Emitter;
+import co.cask.cdap.template.etl.api.realtime.RealtimeContext;
+import co.cask.cdap.template.etl.api.realtime.RealtimeSource;
+import co.cask.cdap.template.etl.api.realtime.SourceState;
+import com.amazon.sqs.javamessaging.SQSConnection;
+import com.amazon.sqs.javamessaging.SQSConnectionFactory;
+import com.amazonaws.regions.Region;
+import com.amazonaws.regions.Regions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import javax.jms.Message;
+import javax.jms.MessageConsumer;
+import javax.jms.Session;
+import javax.jms.TextMessage;
+
+/**
+ * Realtime source that reads from Amazon SQS.
+ * TODO: CDAP-2978: Extend JMS source so this class can be deleted.
+ */
+@Plugin(type = "source")
+@Name("Sqs")
+@Description("Sqs Realtime Source - Emits a record with a field 'body' of string type")
+public class SqsSource extends RealtimeSource<StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(SqsSource.class);
+  private static final String REGION_DESCRIPTION = "Region where the queue is located.";
+  private static final String ACCESSKEY_DESCRIPTION = "Access Key of the AWS account to use.";
+  private static final String ACCESSID_DESCRIPTION = "Access ID of the AWS account to use.";
+  private static final String QUEUENAME_DESCRIPTION = "Name of the queue.";
+  private static final String ENDPOINT_DESCRIPTION = "Endpoint of the SQS server to connect to. Omit this field to" +
+    "connect to AWS.";
+  private static final Schema DEFAULT_SCHEMA = Schema.recordOf(
+    "event",
+    Schema.Field.of("body", Schema.of(Schema.Type.STRING))
+  );
+  private static final int MAX_MESSAGE_COUNT = 20;
+  private static final int TIMEOUT_LENGTH = 1000;
+  private final SqsConfig config;
+  private SQSConnectionFactory connectionFactory;
+  private MessageConsumer consumer;
+  private Session session;
+  private SQSConnection connection;
+
+  public SqsSource(SqsConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void initialize(RealtimeContext contex) {
+    try {
+      SQSConnectionFactory.Builder sqsBuild = SQSConnectionFactory.builder()
+        .withRegion(Region.getRegion(Regions.fromName(config.region)));
+      connectionFactory = config.endpoint == null ? sqsBuild.build() : sqsBuild.withEndpoint(config.endpoint).build();
+      connection = connectionFactory.createConnection(config.accessID, config.accessKey);
+      session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+      consumer = session.createConsumer(session.createQueue(config.queueName));
+      connection.start();
+    } catch (Exception e) {
+      if (session != null) {
+        try {
+          session.close();
+        } catch (Exception ex1) {
+          LOG.warn("Exception when closing session", ex1);
+        }
+      }
+      if (connection != null) {
+        try {
+          connection.close();
+        } catch (Exception ex2) {
+          LOG.warn("Exception when closing connection", ex2);
+        }
+      }
+      if (consumer != null) {
+        try {
+          consumer.close();
+        } catch (Exception ex3) {
+          LOG.warn("Exception when closing consumer", ex3);
+        }
+      }
+      LOG.error("Failed to connect to SQS");
+      throw new IllegalStateException("Could not connect to SQS.");
+    }
+  }
+
+  @Override
+  public SourceState poll(Emitter<StructuredRecord> writer, SourceState currentState) throws Exception {
+    Message msg;
+    int count = 0;
+    while ((msg = consumer.receive(TIMEOUT_LENGTH)) != null && count < MAX_MESSAGE_COUNT) {
+      String text = ((TextMessage) msg).getText();
+      if (text.isEmpty()) {
+        msg.acknowledge();
+        continue;
+      }
+
+      StructuredRecord record = StructuredRecord.builder(DEFAULT_SCHEMA)
+        .set("body", text)
+        .build();
+      writer.emit(record);
+      msg.acknowledge();
+      count++;
+    }
+    return currentState;
+  }
+
+  @Override
+  public void destroy() {
+    try {
+      consumer.close();
+      session.close();
+      connection.close();
+    } catch (Exception ex) {
+      throw new RuntimeException("Exception on closing SQS connection: " + ex.getMessage(), ex);
+    }
+  }
+
+  /**
+   * Config class for {@link SqsSource}
+   */
+  public static class SqsConfig extends PluginConfig {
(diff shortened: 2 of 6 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.elasticmq.rest.RestServer;
+import org.elasticmq.rest.sqs.SQSRestServerBuilder;
+import org.elasticmq.storage.inmemory.InMemoryStorage;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class SqsSourceTest {
+  private static final int ELASTIMQ_PORT = 9325;
```
sreevatsanraman: You should not hardcode the port. We should use a random port. Use
`Networks.getRandomPort()`

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
```diff
@@ -0,0 +1,175 @@
...
+      LOG.error("Failed to connect to SQS");
+      throw new IllegalStateException("Could not connect to SQS.");
+    }
+  }
+
+  @Override
+  public SourceState poll(Emitter<StructuredRecord> writer, SourceState currentState) throws Exception {
+    Message msg;
+    int count = 0;
+    while ((msg = consumer.receive(TIMEOUT_LENGTH)) != null && count < MAX_MESSSAGE_COUNT) {
```
dtzeng: Using just a receive will result in a lot of empty messages.  Do we want to use long polling instead?  http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html
Long polling will let us receive messages in batch and also reduce a lot of cost for our SQS server.  We might also be able to just use the Amazon SDK only instead of all the other libraries

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
```diff
@@ -0,0 +1,175 @@
...
+  private static final String ACCESSKEY_DESCRIPTION = "Access Key of the AWS account to use.";
+  private static final String ACCESSID_DESCRIPTION = "Access ID of the AWS account to use.";
+  private static final String QUEUENAME_DESCRIPTION = "Name of the queue.";
+  private static final String ENDPOINT_DESCRIPTION = "Endpoint of the SQS server to connect to. Leave blank to " +
+    "connect to AWS.";
+  private static final Schema DEFAULT_SCHEMA = Schema.recordOf(
+    "event",
+    Schema.Field.of("body", Schema.of(Schema.Type.STRING))
+  );
+  private static final int MAX_MESSSAGE_COUNT = 20;
```
dtzeng: "messsage" typo

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
jayfjin: probably want to keep this as generic as possible because eventually we want to merge this with JMS source. Can use Strings.isNullOrEmpty() in the meantime to avoid the empty message problem.

dtzeng: Sorry, meant empty as in null message, not empty string.  When you do it this way, the client will only select a subset of the servers and check if any messages are on those servers.  Long polling will check all servers of a queue for some time, and return all of them.  We might want to keep SQS separate, because it is unique in this way.  Thoughts?

jayfjin: Don't think the benefits of receiving a few fewer empty messages is worth keeping separate. If the queue is highly populated/contested then this isn't a problem, and if the queue isn't heavily used then the performance impact is negligible. Also didn't run into this problem when testing. Furthermore long polling must be enabled in order to work.

cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
```diff
@@ -191,6 +191,22 @@
     </dependency>
     <!-- End for Kafka Source -->
     <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>amazon-sqs-java-messaging-lib</artifactId>
+      <version>1.0.0</version>
+      <type>jar</type>
```
chtyim: Why need to specify the type?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
```diff
@@ -0,0 +1,182 @@
...
+    Schema.Field.of("body", Schema.of(Schema.Type.STRING))
+  );
+  private static final int MAX_MESSAGE_COUNT = 20;
+  private static final int TIMEOUT_LENGTH = 1000;
+  private final SqsConfig config;
+  private SQSConnectionFactory connectionFactory;
+  private MessageConsumer consumer;
+  private Session session;
+  private SQSConnection connection;
+  public SqsSource(SqsConfig config) {
```
chtyim: Add an empty separating fields from method.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
```diff
@@ -0,0 +1,182 @@
...
+      msg.acknowledge();
+      count++;
+    }
+    return currentState;
+  }
+
+  @Override
+  public void destroy() {
+    try {
+      if (consumer != null) {
```
chtyim: Can any of them be null? This method only get called if `initialize` returned without exception.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
```diff
@@ -0,0 +1,182 @@
...
+@Plugin(type = "source")
+@Name("Sqs")
+@Description("Sqs Realtime Source - Emits a record with a field 'body' of string type")
+public class SqsSource extends RealtimeSource<StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(SqsSource.class);
+  private static final String REGION_DESCRIPTION = "Region where the queue is located.";
+  private static final String ACCESSKEY_DESCRIPTION = "Access Key of the AWS account to use.";
+  private static final String ACCESSID_DESCRIPTION = "Access ID of the AWS account to use.";
+  private static final String QUEUENAME_DESCRIPTION = "Name of the queue.";
+  private static final String ENDPOINT_DESCRIPTION = "Endpoint of the SQS server to connect to. Leave blank to " +
```
bdmogal: Leave blank or omit?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
```diff
@@ -0,0 +1,80 @@
...
+public class SqsSourceTest {
+  private static final int ELASTIMQ_PORT = Networks.getRandomPort();
+  private RestServer sqsServer;
+  private Node elasticNode;
+  private static final String HOST_URL = "http://localhost:" + ELASTIMQ_PORT;
+  private static final String AWS_CRED = "testKey";
+
+  @Before
+  public void setupElasticMQ() {
+    elasticNode = org.elasticmq.NodeBuilder.withStorage(new InMemoryStorage());
```
bdmogal: shouldn't be fully qualified?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class SqsSourceTest {
+  private static final int ELASTIMQ_PORT = Networks.getRandomPort();
+  private RestServer sqsServer;
+  private Node elasticNode;
+  private static final String HOST_URL = "http://localhost:" + ELASTIMQ_PORT;
```
bdmogal: better to group static finals together

bdmogal: Also, use ``InetAddress.getLocalHost().getHostName()``?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
```diff
@@ -0,0 +1,80 @@
...
+  @Before
+  public void setupElasticMQ() {
+    elasticNode = org.elasticmq.NodeBuilder.withStorage(new InMemoryStorage());
+    sqsServer = new SQSRestServerBuilder(elasticNode.nativeClient(), ELASTIMQ_PORT,
+                                         new NodeAddress("http", "localhost", ELASTIMQ_PORT, "")).start();
+  }
+
+  @Test
+  public void testSQS() throws Exception {
+    final String queueName = "testQueue";
```
bdmogal: do these have to be final?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
bdmogal: also, nit: ``HOST_URL`` is slightly vague, can we qualify it with what host it is? ``SQS_HOST_URL``?

jayfjin: using INetAddress.getLocalHost().getHostName() throws an exception, is it really necessary for this purpose?

bdmogal: Take a look at ``BatchETLDBAdapterTest``, to see how you can encapsulate these connection params in to a Server class.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
```diff
@@ -0,0 +1,175 @@
...
+      connectionFactory = config.endpoint == null ? sqsBuild.build() : sqsBuild.withEndpoint(config.endpoint).build();
+      connection = connectionFactory.createConnection(config.accessID, config.accessKey);
+      session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+      consumer = session.createConsumer(session.createQueue(config.queueName));
+      connection.start();
+    } catch (Exception e) {
+      if (session != null) {
+        try {
+          session.close();
+        } catch (Exception ex1) {
```
bdmogal: If you're not doing anything much other than logging on exception, it perhaps makes sense to use ``Closeables.closeQuietly`` - here and the two below as well.

bdmogal: Ok nvm. these aren't ``Closeable``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
```diff
@@ -0,0 +1,110 @@
...
+  private RestServer sqsServer;
+  private SQSServer sqsConfig;
+
+  @Before
+  public void setupElasticMQ() {
+    elasticNode = NodeBuilder.withStorage(new InMemoryStorage());
+    sqsConfig = new SQSServer(Networks.getRandomPort());
+    sqsServer = new SQSRestServerBuilder(elasticNode.nativeClient(), sqsConfig.getPort(),
+                                         new NodeAddress("http", sqsConfig.getHostname(), sqsConfig.getPort(), ""))
+      .start();
```
bdmogal: Can this indentation be better?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
```diff
@@ -0,0 +1,110 @@
...
+  @After
+  public void stopElasticMQ() {
+    sqsServer.stop();
+    elasticNode.shutdown();
+  }
+
+  private class SQSServer {
+    private static final String AWS_CRED = "testKey";
+    private String sqsHostName;
+    private int port;
```
bdmogal: final

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @After
+  public void stopElasticMQ() {
+    sqsServer.stop();
+    elasticNode.shutdown();
+  }
+
+  private class SQSServer {
+    private static final String AWS_CRED = "testKey";
+    private String sqsHostName;
```
bdmogal: final

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/realtime/source/SqsSourceTest.java
```diff
@@ -0,0 +1,110 @@
...
+    private int port;
+
+    public SQSServer(int port) {
+      try {
+        sqsHostName = InetAddress.getLocalHost().getHostName();
+      } catch (Exception e) {
```
bdmogal: can throw Exception instead

bdmogal:
only minor comments. LGTM otherwise

PR #3132, opened 2015-07-10
[CDAP-2886] add upgrade for file sets with absolute base paths

anew:
removes the leading / from all file sets with absolute base paths (also embedded ones). This is needed because the absolute base path was interpreted as relative to the namespace's data dir until 3.0.1, so the fix in CDAP-2884 would make these file set point to the wrong location.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDefinition.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDefinition.java
@@ -31,6 +31,7 @@ import co.cask.cdap.api.dataset.lib.PartitionedFileSetProperties;
 import co.cask.cdap.api.dataset.lib.Partitioning;
 import co.cask.cdap.api.dataset.table.Table;
 import co.cask.cdap.explore.client.ExploreFacade;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
 import com.google.inject.Inject;
@@ -51,8 +52,10 @@ public class PartitionedFileSetDefinition extends AbstractDatasetDefinition<Part
 
   private static final Logger LOG = LoggerFactory.getLogger(PartitionedFileSetDefinition.class);
 
-  protected static final String PARTITION_TABLE_NAME = "partitions";
-  protected static final String FILESET_NAME = "files";
+  @VisibleForTesting
+  public static final String PARTITION_TABLE_NAME = "partitions";
+  @VisibleForTesting
+  public static final String FILESET_NAME = "files";
 
   protected final DatasetDefinition<? extends Table, ?> tableDef;
   protected final DatasetDefinition<? extends FileSet, ?> filesetDef;
--- a/cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/DatasetFrameworkTestUtil.java
+++ b/cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/DatasetFrameworkTestUtil.java
@@ -47,9 +47,14 @@ import java.util.Map;
 public final class DatasetFrameworkTestUtil extends ExternalResource {
   public static final Id.Namespace NAMESPACE_ID = Id.Namespace.from("myspace");
 
+  private Injector injector;
   private CConfiguration cConf;
   private TemporaryFolder tmpFolder;
   private DatasetFramework framework;
+
+  public Injector getInjector() {
+    return injector;
+  }
 
   public CConfiguration getConfiguration() {
     return cConf;
@@ -63,7 +68,7 @@ public final class DatasetFrameworkTestUtil extends ExternalResource {
     cConf = CConfiguration.create();
     cConf.set(Constants.CFG_LOCAL_DATA_DIR, localDataDir.getAbsolutePath());
 
-    final Injector injector = Guice.createInjector(
+    injector = Guice.createInjector(
       new ConfigModule(cConf),
       new LocationRuntimeModule().getInMemoryModules(),
       new SystemDatasetRuntimeModule().getInMemoryModules(),
--- a/cdap-master/pom.xml
+++ b/cdap-master/pom.xml
@@ -149,6 +149,13 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-data-fabric</artifactId>
+      <version>${project.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <profiles>
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
@@ -17,11 +17,17 @@
 package co.cask.cdap.data.tools;
 
 import co.cask.cdap.api.dataset.DatasetAdmin;
+import co.cask.cdap.api.dataset.DatasetSpecification;
+import co.cask.cdap.api.dataset.lib.FileSet;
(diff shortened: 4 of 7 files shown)


cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/DatasetFrameworkTestUtil.java
```diff
@@ -47,10 +47,15 @@
 public final class DatasetFrameworkTestUtil extends ExternalResource {
   public static final Id.Namespace NAMESPACE_ID = Id.Namespace.from("myspace");
 
+  private Injector injector;
   private CConfiguration cConf;
   private TemporaryFolder tmpFolder;
   private DatasetFramework framework;
 
+  public Injector getInjector() {
```
sreevatsanraman: I don't think we need this change. Since we don't use this in any unit test, atleast I did not see any usage of `getInjector()` in this PR

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDefinition.java
```diff
@@ -31,6 +31,7 @@
 import co.cask.cdap.api.dataset.lib.Partitioning;
 import co.cask.cdap.api.dataset.table.Table;
 import co.cask.cdap.explore.client.ExploreFacade;
+import com.google.common.annotations.VisibleForTesting;
```
sreevatsanraman: These changes are not needed.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -52,12 +67,15 @@
...
   private final DatasetFramework dsFramework;
+  private final TransactionExecutorFactory transactionExecutorFactory;
   private final Pattern userTablePrefix;
 
   @Inject
-  private DatasetUpgrader(CConfiguration cConf, Configuration hConf, LocationFactory locationFactory,
-                          NamespacedLocationFactory namespacedLocationFactory,
-                          HBaseTableUtil hBaseTableUtil, DatasetFramework dsFramework) {
+  @VisibleForTesting
+  DatasetUpgrader(CConfiguration cConf, Configuration hConf, LocationFactory locationFactory,
```
sreevatsanraman: can be private and remove VisibleForTesting. We are not using this in any tests yet.

cdap-master/pom.xml
```diff
@@ -149,6 +149,13 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
sreevatsanraman: un-related change?

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -128,4 +149,116 @@ public void create() throws IOException {
...
+            DatasetSpecification newSpec = migrateDatasetSpec(spec.getName(), spec);
+            mds.write(entry.getKey(), newSpec);
+          }
+        }
+      }
+    });
+  }
+
+  @VisibleForTesting
+  DatasetSpecification migrateDatasetSpec(DatasetSpecification spec) {
```
sreevatsanraman: This function can be removed. Directly use `DatasetSpecification migrateDatasetSpec(String datasetName, DatasetSpecification spec) `

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -128,4 +149,116 @@ public void create() throws IOException {
...
+      .properties(newProperties)
+      .build();
+  }
+
+  private boolean isFileSet(DatasetSpecification spec) {
+    return spec.getType().equals(FileSet.class.getName()) || spec.getType().equals("fileSet");
+  }
+
+  @VisibleForTesting
+  String containsFileSetWithAbsolutePath(DatasetSpecification spec) {
```
sreevatsanraman: can be private and remove `VisibleForTesting`

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -104,15 +105,17 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(UpgradeTool.class);
 
+  @VisibleForTesting
```
sreevatsanraman: Same comment as above.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/DatasetFrameworkTestUtil.java
anew: DatasetUpgraderTest needs it to set up the upgrade tool.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDefinition.java
anew: Same here, DatasetUpgraderTest needs it

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
anew: Yes we are, in DatasetUpgraderTest

cdap-master/pom.xml
anew: needed for the DatasetUpgraderTest. It needs a DatasetFrameworkTestUtil, which is from data fabric test jar

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
anew: Yes it could. However, in the test case, it calls this one.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
anew: ditto.

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
anew: ditto.

anew:
OK, I just realized that for some reason, the last commit did not include the test case... (strange, I did a commit -a). I pushed that now.

anew:
This is tested manually on a cluster with CDAP-3.0.1. 
- wrote an app that has datasets: a fileset, a pfs, a tpfs, a custom dataset that embeds a tpfs
- the app also has a service that allows uploading files to and downloading files from all these datasets
- deployed the app into default and "dummy" namespace
- created a file in each of the datasets through the service, in both namespaces
- upgraded the cluster to 3.1.0-SNAPSHOT
- ran the upgrade tool
- restarted the app in both namespaces
- retrieved all the files in all the datasets and in each namespace through the service.
- looked at the dataset specs before and after upgrade tool and validated that leading /s are removed

sreevatsanraman:
> OK, I just realized that for some reason, the last commit did not include the test case.
Makes sense. I did not see a test case related when I reviewed. Continuing review now.

sreevatsanraman:
LGTM

anwar6953:
Is there an issue if user runs this upgrade step multiple times?
Or if they run it after creating a dataset with the new semantics?

anew:
Multiple times, no. But running after creating a new one, yes, that would cause an issue. My original proposal was to indicate that using an extra property. However, in the design review Nitin decided that we don't want that extra property.

anew:
We could add a version id to the properties of the dataset when it is created. New datasets would have that version id, whereas legacy ones won't. That way the upgrader can distinguish. I kind of dislike the idea of explicitly versioning the properties as part of the dataset's code (I think that shouyld be a capability of the framework), but for now, we can take that approach.

anew:
opened https://issues.cask.co/browse/CDAP-2981 for this issue

PR #3134, opened 2015-07-10
CDAP-2548 : Upgrade usage.registry dataset type from Table to UsageDataset

rohitsinha54:
Did a complete test on cluster: It works.
1. Deployed 3.0.0 cluster through coopr autobuild
2. Deployed Purchase app
3. Sent some events to the stream (which will create usage.registry table)
4. Verified through hbase shell that dataset.instance has the table type for above table as Table.
5. Stopped all services
6. Beamed the fix version branch from the PR. 
7. Ran the upgrade tool 
8. After upgrade verified that the type in dataset.instance is changed to UsageDataset
9. Started all services
10. Ran flow, services, workflow and queries for results.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/UsageRegistry.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/UsageRegistry.java
@@ -17,9 +17,11 @@
 package co.cask.cdap.data2.registry;
 
 import co.cask.cdap.api.dataset.DatasetProperties;
+import co.cask.cdap.api.dataset.DatasetSpecification;
 import co.cask.cdap.api.dataset.table.Table;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.data2.datafabric.dataset.DatasetsUtil;
+import co.cask.cdap.data2.datafabric.dataset.instance.DatasetInstanceManager;
 import co.cask.cdap.data2.dataset2.DatasetFramework;
 import co.cask.cdap.data2.dataset2.tx.Transactional;
 import co.cask.cdap.proto.Id;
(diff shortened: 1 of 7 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/UsageRegistry.java
```diff
@@ -323,6 +325,26 @@ public Void apply(UsageDatasetIterable input) throws Exception {
...
   /**
+   * Upgrades the UsageRegistry. In its current implementation it changed the type of usage.registry table in
+   * dataset.instances table from table to UsageDataset.
+   *
+   * @param datasetInstanceManager {@link DatasetInstanceManager} for the dataset instance meta data
+   */
+  public void upgrade(DatasetInstanceManager datasetInstanceManager) {
+    DatasetSpecification oldDatasetSpecification = datasetInstanceManager.get(USAGE_INSTANCE_ID);
+
+    if (!oldDatasetSpecification.getType().equals("UsageDataset")) {
```
rohitsinha54: @chtyim : Is there a way to get this name programmatically when I don't have access to the dataset to do getName ?

anwar6953:
The underlying table name is exactly the same, even before it was a custom dataset?
Wouldn't it be `customDatasetName.tableName` now, whereas before it could simply be `tableName`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/UsageRegistry.java
chtyim: It is the simple class name of `UsageDataset` class. See `UsageDatasetModule`.

anwar6953:
Scratch my previous comment out. Apparently, it's not using an embedded dataset, so my concern that the underlying table name is concatenated with the dataset name is not a concern.

rohitsinha54:
Will test one more time on a cluster and then merge it.

anwar6953:
I think as part of your testing, you should query which datasets are used by program X, and which programs use dataset Y.

rohitsinha54:
Tested to work on cluster.

anwar6953:
LGTM

PR #3135, opened 2015-07-10
CDAP-2784 Support specifying metadata for the Output Partition of a M…

anwar6953:
…apReduce job.

Also add test cases for other PartitionedFileSetArguments.

Resolves:
https://issues.cask.co/browse/CDAP-2784
http://builds.cask.co/browse/CDAP-DUT2200-3

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionFilter.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionFilter.java
@@ -143,6 +143,25 @@ public class PartitionFilter {
     }
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    PartitionFilter that = (PartitionFilter) o;
+
+    return conditions.equals(that.conditions);
+  }
+
+  @Override
+  public int hashCode() {
+    return conditions.hashCode();
+  }
+
   /**
    * Represents a condition on a partitioning field, by means of an inclusive lower bound and an exclusive upper bound.
    * As a special case, if only one value is passed to the constructor, then this represents an equality filter.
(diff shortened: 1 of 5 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
```diff
@@ -130,6 +152,7 @@ public static PartitionFilter getInputPartitionFilter(Map<String, String> argume
     return builder.build();
   }
 
+  // TODO: should be private, or is it designed to be used externally?
```
anwar6953: was this intended to be public, @anew

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
```diff
@@ -28,13 +28,15 @@
 @Beta
 public class PartitionedFileSetArguments {
 
-  public static final String OUTPUT_PARTITION_KEY_PREFIX = "output.partition.";
+  // these fields should be private?
```
anwar6953: were these intended to be public, @anew

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
anew: I was following hadoop style, where, for example, FileInputFormat has a method addInputPath() but also a public constant INPUT_DIR. Since this is part of API, we can hardly change it now.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
anew: same here.

anwar6953:
Ok, removed my TODO comments for the public/private, then.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionFilter.java
```diff
@@ -141,6 +141,32 @@ public PartitionFilter build() {
       Preconditions.checkState(!map.isEmpty(), "Partition filter cannot be empty.");
       return new PartitionFilter(map);
     }
+
+    /**
+     * @return <tt>true</tt> if the underlying map contains no key-value mappings
```
anew: Javadoc should not explain by means of the implementation. It should always explain by means of what the user sees. In this case, it would be better to say" returns true if this filter is always true. Or better: if this filter imposes no conditions

anew:
It seems that this is only half of the job? where is the code in MapReduce that actually adds the metadata to the output partition? Or was that code already there? 

Is there a test that runs a map/reduce, which writes to a partition with additional metadata, and validates after the job suceeds that the partition has the metadata? We should have one. Perhaps just modify the existing test for MR with PFS?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionFilter.java
anwar6953: I actually needed this in a separate branch. Removing it from here and will update the javadoc in that branch.

anwar6953:
I did modify an existing test case to test that metadata will be written to the output partition of the job:
https://github.com/caskdata/cdap/pull/3135/files#diff-6573cfe6a0e6a573ec5c20803fd88190R167

anew:
LGTM

PR #3136, opened 2015-07-10
CDAP-2975: Pre-split FactTables

abaranau:
NOTE: each split corresponds to a pre-aggregated view of FactTable in Cube


abaranau:
http://builds.cask.co/browse/CDAP-DUT2199-1

abaranau:
GREEN

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/timeseries/FactCodec.java
```diff
@@ -255,6 +255,24 @@ public long getTimestamp(byte[] rowKey, byte[] column) {
...
 
+  static byte[][] getSplits(int aggGroupsCount) {
+    // Row key format:
+    // <version><encoded agg group><time base>...
+    // Version is fixed. We assume agg group name is encoded with 1, ..., <aggGroupsCount>.
+    int encodedIdSize = EntityTable.computeSize();
+    int rowKeySize = VERSION.length + encodedIdSize;
+    // NOTE: we don't need to include first split, it will be added automatically (e.g. by HBase).
+    byte[][] splits = new byte[aggGroupsCount - 1][];
+    for (int i = 2; i <= aggGroupsCount; i++) {
```
gsps1: we start from 2, as the first split is added by hbase ? if so, would the first aggregation would get matched in the first split ?

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/timeseries/FactTableTest.java
```diff
@@ -505,6 +510,58 @@ public void testMaxResolution() throws Exception {
...
+      for (int i = 0; i < splits.length; i++) {
+        if (Bytes.compareTo(row.getRow(), splits[i]) < 0) {
+          splitsWithRows.add(i);
+          added = true;
+          break;
+        }
+      }
+      if (!added) {
+        // falls into last split
+        splitsWithRows.add(splits.length);
```
gsps1: when will this happen ?

gsps1:
just have 2 questions.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/timeseries/FactCodec.java
abaranau: yes, it will be added by HBase. Record in the first agg will go into it, yes

cdap-data-fabric/src/test/java/co/cask/cdap/data2/dataset2/lib/timeseries/FactTableTest.java
abaranau: if this is a record of the last aggregation

gsps1:
LGTM

PR #3138, opened 2015-07-10
Feature/ui styling dev start new

astewes:
*  Adds "multiple" helper class to namespace dropdown
*  Removes "No namespaces to display" list item from the namespace dropdown (tooltip text will be used instead)
*  Refines styling for the header dropdown menus

Notes:
*  On smaller devices, the namespace dropdown will look and function as it has been. Responsive design for the namespace dropdown will be handled in a separate PR
*  The namespace list bug will need to be fixed in order for this design change to be fully realized. For demo purposes, in the attached GIF I show: 1) no namespaces to display; 2) list of namespaces to display

Cluster:
http://ns-dropdown1945-1000.dev.continuuity.net:9999/ns/default

header


elfenheart:
:+1: LGTM

PR #3139, opened 2015-07-10
Schema Propagation in Adapter Create

elfenheart:
schemapropagation

### DIFF
--- a/cdap-ui/app/features/adapters/templates/create/tabs/edit-transform-properties.html
+++ b/cdap-ui/app/features/adapters/templates/create/tabs/edit-transform-properties.html
@@ -1,7 +1,7 @@
 <br />
 
 <div class="row">
-  <div class="col-xs-12" ng-init="plugin=transform">
+  <div class="col-xs-12" ng-init="plugin=transform; type=AdapterCreateController.model.metadata.type">
     <div ng-include="'/assets/features/adapters/templates/create/tabs/plugin-edit-form.html'">
     </div>
   </div>
--- a/cdap-ui/app/services/plumber/my-plumb-service.js
+++ b/cdap-ui/app/services/plumber/my-plumb-service.js
@@ -81,15 +81,7 @@ angular.module(PKG.name + '.services')
         var source = sourceConn.length ? this.nodes[sourceConn[0].source] : null;
 
         sourceSchema = source.outputSchema;
-        // try {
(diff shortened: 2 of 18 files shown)


cdap-ui/app/features/adapters/templates/create/tabs/edit-transform-properties.html
```diff
@@ -1,7 +1,7 @@
 <br />
 
 <div class="row">
-  <div class="col-xs-12" ng-init="plugin=transform">
+  <div class="col-xs-12" ng-init="plugin=transform; type=AdapterCreateController.model.metadata.type">
```
ajainarayanan: Do we use this in the /plugin-edit-form.html view?

cdap-ui/app/services/plumber/my-plumb-service.js
```diff
@@ -62,25 +62,37 @@ angular.module(PKG.name + '.services')
...
     this.editPluginProperties = function (scope, pluginId, pluginType) {
-      var plugin = this.nodes[pluginId];
-      var prom = $q.defer();
       var propertiesApiMap = {
         'source': myAdapterApi.fetchSourceProperties,
         'sink': myAdapterApi.fetchSinkProperties,
         'transform': myAdapterApi.fetchTransformProperties
       };
+
+      var sourceConn = $filter('filter')(this.connections, { target: pluginId });
```
ajainarayanan: Is it possible for us to avoid ```$filter```

cdap-ui/app/features/adapters/templates/create/tabs/edit-transform-properties.html
elfenheart: hmmm after further inspection, I think we can delete the 3 files:
1. edit-sink-properties.html
2. edit-source-properties.html
3. edit-transform-properties.html

Am I missing anything?

cdap-ui/app/services/plumber/my-plumb-service.js
elfenheart: Will have to write a helper function to do that, because this is filtering on embedded property value. So we need to traverse the entire array and check for that key/value pair.  IndexOf does not have that functionality.

What do you think?

ajainarayanan:
:+1: LGTM

PR #3140, opened 2015-07-11
(CDAP-2228) (CDAP-2952) Take total control of ClassLoader for MapReduce

chtyim:
The approach is to launch every MapReduce container (AM/Mapper/Reducer) with our
own launcher (main method). The launcher class will creates a MapReduceClassLoader,
set it as the context ClassLoader and and use that to load the actual MR framework class.
By doing do, we no longer at the merce of when MR framework creates ApplicationClassLoader.


chtyim:
Tested on cluster with hadoop-2.2 and hadoop-2.4

albertshau:
lgtm

PR #3142, opened 2015-07-11
Upgrade tool fixes

sreevatsanraman:
* Upgrades all tables in system namespace
* Upgrades all user table across all namespace
* Avoids creating redundant empty tables after the upgrade
* Avoids upgrading non-cdap managed tables

JIRA: https://issues.cask.co/browse/CDAP-2962
Build: http://builds.cask.co/browse/CDAP-RBT308

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/config/DefaultConfigStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/config/DefaultConfigStore.java
@@ -82,8 +82,7 @@ public class DefaultConfigStore implements ConfigStore {
 
   public static void setupDatasets(DatasetFramework dsFramework) throws DatasetManagementException, IOException {
     dsFramework.addInstance(Table.class.getName(), Id.DatasetInstance.from(
-                              Constants.DEFAULT_NAMESPACE_ID, Joiner.on(".").join(Constants.SYSTEM_NAMESPACE,
-                                                                                  Constants.ConfigStore.CONFIG_TABLE)),
+                              Constants.SYSTEM_NAMESPACE_ID, Constants.ConfigStore.CONFIG_TABLE),
                             DatasetProperties.EMPTY);
   }
 
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
@@ -95,10 +95,7 @@ public class DatasetUpgrader extends AbstractUpgrader {
   private void upgradeUserTables() throws Exception {
     HBaseAdmin hAdmin = new HBaseAdmin(hConf);
     for (HTableDescriptor desc : hAdmin.listTables()) {
-      String tableName = desc.getNameAsString();
-      if (!isTableInSystemNamespace(tableName) &&
-           isUserTable(tableName) &&
-           isTableCreatedByCDAP(desc)) {
+      if (isCDAPUserTable(desc)) {
         upgradeUserTable(desc);
       }
     }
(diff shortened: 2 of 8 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/config/DefaultConfigStore.java
```diff
@@ -82,8 +82,7 @@ public ConfigTable get() {
 
   public static void setupDatasets(DatasetFramework dsFramework) throws DatasetManagementException, IOException {
     dsFramework.addInstance(Table.class.getName(), Id.DatasetInstance.from(
-                              Constants.DEFAULT_NAMESPACE_ID, Joiner.on(".").join(Constants.SYSTEM_NAMESPACE,
-                                                                                  Constants.ConfigStore.CONFIG_TABLE)),
+                              Constants.SYSTEM_NAMESPACE_ID, Constants.ConfigStore.CONFIG_TABLE),
```
chtyim: Is this an existing bug fix or are we changing namespace?

sreevatsanraman: This is an existing bug.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -93,39 +94,69 @@ private void upgradeSystemDatasets() throws Exception {
...
+        return HBaseTableAdmin.createCoprocessorJarInternal(cConf,
+                                                            locationFactory,
+                                                            hBaseTableUtil,
+                                                            transactional,
+                                                            supportsIncrement);
+      }
+
+      @Override
+      protected boolean upgradeTable(HTableDescriptor tableDescriptor) {
+        return false;
```
chtyim: Why return false? Meaning not to perform upgrade?

cdap-app-fabric/src/main/java/co/cask/cdap/config/DefaultConfigStore.java
bdmogal: Yes, this would work for upgrade from 2.6 -> 2.8, not from 2.8 ->3.0

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -93,39 +94,69 @@ private void upgradeSystemDatasets() throws Exception {
...
+
+  // Note: This check can be safely used for user table since we create meta.
+  // CDAP-2963 should be fixed so that we can make use of this check generically for all cdap tables
+  private boolean isTableCreatedByCDAP(HTableDescriptor desc) {
+    return (desc.getValue("cdap.version") != null);
+  }
+
+  private boolean isTableInSystemNamespace(String tableName) {
+    // dataset in system namespace starts with <prefix>_system ex: cdap_system
+    return tableName.startsWith(String.format("%s_%s", this.datasetTablePrefix, "system"));
```
bdmogal: Use ``Constants.SYSTEM_NAMESPACE``

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
```diff
@@ -93,39 +94,69 @@ private void upgradeSystemDatasets() throws Exception {
...
+  }
+
+  private boolean isUserTable(String tableName) {
```
bdmogal: This method depends on being used in conjunction with ``isTableInSystemNamespace``, right - meaning you have to call ``!isTableInSystemNamespace(tableName) && isUserTable(tableName)``? 

Perhaps change name/add doc to reflect that?

Or maybe just call this method ``isCDAPUserTable``, and do the ``isTableInSystemNamespace`` and ``isTableCreatedByCDAP`` checks inside this method, so users only have to call this method.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
sreevatsanraman: Fixed.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
sreevatsanraman: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/config/DefaultConfigStore.java
rohitsinha54: +1

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetUpgrader.java
sreevatsanraman: The contract says return true, if HTableDescriptor is modified. Since we don't returning false here.
```
/**
   * Modifies the table descriptor for upgrade.
   *
   * @return true if the table descriptor is modified.
   */
  protected abstract boolean upgradeTable(HTableDescriptor tableDescriptor);
```

chtyim:
LGTM

bdmogal:
:+1:

sreevatsanraman:
Will test upgrade - one final time with the changes included in code review and then merge.

PR #3143, opened 2015-07-11
[cdap-585] spark to support filesets

anew:
Work in progress. This only has reads so far. 
@chtyim can you please take a look, especially the first two commits as they fix bugs in the Spark runner. Thanks.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
@@ -21,6 +21,7 @@ import co.cask.cdap.api.common.Scope;
 import co.cask.cdap.api.data.DatasetInstantiationException;
 import co.cask.cdap.api.data.batch.BatchReadable;
 import co.cask.cdap.api.data.batch.BatchWritable;
+import co.cask.cdap.api.data.batch.InputFormatProvider;
 import co.cask.cdap.api.data.batch.Split;
 import co.cask.cdap.api.data.batch.SplitReader;
 import co.cask.cdap.api.data.stream.StreamBatchReadable;
@@ -50,6 +51,7 @@ import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.io.Closeables;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.LongWritable;
+import org.apache.hadoop.mapreduce.InputFormat;
 import org.apache.twill.api.RunId;
 import org.apache.twill.discovery.DiscoveryServiceClient;
 import org.apache.twill.filesystem.Location;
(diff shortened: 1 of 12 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -160,7 +195,8 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
 
   @Override
   public synchronized <T extends Dataset> T getDataset(String name, Map<String, String> arguments) {
-    Map<String, String> datasetArgs = RuntimeArguments.extractScope(Scope.DATASET, name, getRuntimeArguments());
+    Map<String, String> datasetArgs = new HashMap<>();
+    datasetArgs.putAll(RuntimeArguments.extractScope(Scope.DATASET, name, getRuntimeArguments()));
     datasetArgs.putAll(arguments);
```
anew: this would throw an exception if the user passes in ImmutableMap.of()

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -169,6 +205,9 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
     T dataset = (T) datasets.get(key);
     if (dataset == null) {
       dataset = instantiateDataset(name, datasetArgs);
+      if (dataset instanceof TransactionAware) {
+        ((TransactionAware) dataset).startTx(getTransaction());
+      }
```
anew: This was needed because if a dataset is instantiated in the middle of a spark job, it needs the transaction. Apparently we do not have existing test cases for Spark jobs with datasets.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/JavaSparkFacade.java
```diff
@@ -42,9 +42,8 @@ public JavaSparkFacade(SparkConf sparkConf) {
   @Override
   public <R, K, V> R createRDD(Class<? extends InputFormat> inputFormatClass,
                                Class<K> keyClass, Class<V> valueClass, Configuration hConf) {
-    Configuration configuration = new Configuration(hConf);
-    configuration.set(MRJobConfig.INPUT_FORMAT_CLASS_ATTR, inputFormatClass.getName());
-    return (R) sparkContext.newAPIHadoopRDD(configuration, inputFormatClass, keyClass, valueClass);
+    hConf.set(MRJobConfig.INPUT_FORMAT_CLASS_ATTR, inputFormatClass.getName());
```
anew: this - and the following - is what we discussed in person, @chtyim - to avoid double copying.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
```diff
@@ -167,11 +167,7 @@ protected void run() throws Exception {
         throw e;
       }
     } finally {
-      try {
-        Closeables.closeQuietly(executionContext);
-      } finally {
-        Thread.currentThread().setContextClassLoader(oldClassLoader);
-      }
```
anew: Removed this because the transaction gets committed in shutdown(). If the datasets are already closed at that time, they will throw an exception in commitTx(), and the whole job fails.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: Seems like the contract of the `RuntimeArguments.extractScope` is unclear. If the `getRuntimeArguments()` map passed to the method is an empty Map, then it returns the same map (which is immutable, since runtime args are immutable). But if there is something in the map, then it creates a HashMap and returns it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
chtyim: hum, so need to move the close to shutdown, but both contexts need to be closed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: the startTx has been called in `instantiateDataset`, isn't it?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -113,10 +115,43 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
...
+      String inputFormatName = ((InputFormatProvider) dataset).getInputFormatClassName();
+
+      // load the input format class
+      if (inputFormatName == null) {
+        throw new DatasetInstantiationException(
+          String.format("Dataset '%s' provided null as the input format class name", datasetName));
+      }
+      try {
+        @SuppressWarnings("unchecked")
+        Class<? extends InputFormat> inputFormatClass = (Class<? extends InputFormat>) Class.forName(inputFormatName);
```
chtyim: Need to load it through the `SparkClassLoader`, otherwise cannot load classes inside user program jar.

chtyim: You can use `SparkClassLoader.findFromContext()` to get it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: I can change that. sure. I still think it is better to to it this way here

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: hmm, without this, I received an exception...

chtyim: That's strange. Need to try out to tell what's wrong.

anew: you are right. The exception I saw must have been from something else

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
anew: right, so I added Closeables.closeQuietly(executionContext) to shutdown.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: No objection to do a defensive copy here. I also think it's good to make the contract for the extractScope clear.

anew:
Closing this one and opening a new one

PR #3144, opened 2015-07-11
[cdap-585] spark to support filesets

anew:
Allows Spark to read and write ((time-)partitioned) file sets.
- adds support for InputFormatProvider and OutputFormatProvider in ExecutionSparkContext
- fixes a bug in SparkRuntimeService that was closing the datasets before committing them
- fixes a bug in SparkExecutionContext, which was throwing exception if dataset had no runtime args
- avoid double copying of the Hadoop conf in the SparkFacade
- adds tests for Java and Scala Spark programs that read and write each of the three kinds of file sets
- minor fix to RuntimeArguments.extractScope(), to always return a new map
- minor fix in FileSetArguments, to use FileInputFormat.INPUT_DIR instead of deprecated "mapred.input.dir"
 
The only thing that is not implemented is to call the onFailure() method of the DatasetOutputCommitter if the Spark job fails. For MapReduce it is obvious that this happens exactly for one dataset, at the end of the job. Spark can write multiple times, and it is not clear what the semantics are. If a write to one dataset succeeds, we call its onSuccess(), but if subsequently, a write to another dataset fails (or the processing in between), do we also have to call onFailure()? Or only for the one that failed? [CDAP-2983]


anew:
Build passed: http://builds.cask.co/browse/CDAP-DUT2208-4

cdap-api/src/main/java/co/cask/cdap/api/common/RuntimeArguments.java
```diff
@@ -89,11 +89,14 @@ private RuntimeArguments() {
    * @param scope The type of the scope
    * @param name The name of the scope, e.g. "myTable"
    * @param arguments the runtime arguments of the enclosing scope
-   * @return a map that contains the arguments with and without prefix
+   * @return a new map that contains the arguments with and without prefix; or null if the input was null.
```
chtyim: Will it return `null`? Should be marked with `@Nullable`?

chtyim: Also, it may be easier to use if it returns empty map instead of null.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -113,18 +118,95 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
...
   @Override
   public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass) {
-    Map<String, String> datasetArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
-
     // Clone the configuration since it's dataset specification and shouldn't affect the global hConf
     Configuration configuration = new Configuration(hConf);
+
+    // first try if it is InputFormatProvider
+    Map<String, String> dsArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
+    Dataset dataset = getDataset(datasetName, dsArgs);
```
chtyim: I think this dataset instance need to be closed in this method.

chtyim: hum, maybe it can defer to when this context is closed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -160,7 +242,8 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
 
   @Override
   public synchronized <T extends Dataset> T getDataset(String name, Map<String, String> arguments) {
-    Map<String, String> datasetArgs = RuntimeArguments.extractScope(Scope.DATASET, name, getRuntimeArguments());
+    Map<String, String> datasetArgs = new HashMap<>();
```
chtyim: If the contract of the `extractScope` always return a new HashMap, then we don't need to copy it one more time here.

cdap-app-fabric/src/test/scala/co/cask/cdap/internal/app/runtime/spark/ScalaFileCountProgram.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    // read the dataset
+    val inputData: NewHadoopRDD[Long, String] = sc.readFromDataset(input, classOf[Long], classOf[String])
+
+    // create a new RDD with the same key but the value is the length of the string
+    val stringLengths = inputData.map(str => (str._2, str._2.length))
+
+    // write to dataset
+    sc.writeToDataset(stringLengths, output, classOf[String], classOf[Integer])
+    sc.getOriginalSparkContext.asInstanceOf[org.apache.spark.SparkContext].stop()
```
chtyim: No need to call stop from the spark program. The framework does it for you.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkAppUsingFileSet.java
```diff
@@ -0,0 +1,169 @@
...
+        new PairFunction<Tuple2<Long, String>, String, Integer>() {
+        @Override
+        public Tuple2<String, Integer> call(Tuple2<Long, String> pair) throws Exception {
+          return new Tuple2<>(pair._2(), pair._2().length());
+        }
+      });
+
+      // write the character count to dataset
+      context.writeToDataset(stringLengths, output, String.class, Integer.class);
+      ((JavaSparkContext) context.getOriginalSparkContext()).stop();
```
chtyim: No need to call stop().

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: But since Spark program can be iterative, meaning this method can be called many times. I think it's better to close it in this method.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -113,18 +118,95 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
...
   }
 
   @Override
   public <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass) {
+    // Clone the configuration since it's dataset specification and shouldn't affect the global hConf
+    Configuration configuration = new Configuration(hConf);
+
+    // first try if it is OutputFormatProvider
+    Map<String, String> dsArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
+    Dataset dataset = getDataset(datasetName, dsArgs);
```
chtyim: Same as above, I think it's better to close this dataset instance in this method.

cdap-api/src/main/java/co/cask/cdap/api/common/RuntimeArguments.java
anew: agreed

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: However, then it will be closed again when the context is closed. Can we be sure that any dataset is ok with being closed repeatedly?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: ok

cdap-app-fabric/src/test/scala/co/cask/cdap/internal/app/runtime/spark/ScalaFileCountProgram.scala
anew: Oh, I did not pay attention. Just copied the existing ScalaCharCount program and modified it. I am fixing both now.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkAppUsingFileSet.java
anew: Same here. Will fix both.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: So, the getDatasets() remembers all datasets it instantiates in the datasets Map. If this is iterative, and the same dataset is used again, getDataset() will return the same instance. Which would be bad if I close it here. So, if I close it, I have to remove it from the datasets Map, for which currently there is no way provided. But if I do that, then rollbackTx() won't be called in case the program fails later. So I believe I have to keep it open.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: See my comment about this above.

anew:
Thanks for the review, @chtyim. I addressed all your comments. Please see especially my comments about closing the dataset. Thanks

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
chtyim: Since dataset instances created in the `ExecutionSparkContext` are in long transaction, they won't get rollback, but instead the tx will just get invalided if the job fail.

chtyim: By close, I mean `commitAndClose()` and remove from the local cache since the dataset instance created in here doesn't return back to user.

chtyim: And to just get a new dataset instance without putting into the `datasets` instance cache, you can call `instantiateDataset` instead of `getDataset` (and that was why the `instantiateDataset` exists, which you can see both the `getBatchReadable` and `getBatchWritable` calls instead of `getDataset`).

anew:
@chtyim I made the suggested change. It is not that simple as for BatchWritable (using the delegate is easy if you only use a single method... here we need multiple). But it appears to work.

New build running here: http://builds.cask.co/browse/CDAP-DUT2208

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -113,18 +118,107 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
...
+          throw new DatasetInstantiationException(String.format(
+            "Input format class %s provided by dataset '%s' is not an input format", outputFormatName, datasetName), e);
+        }
+        try {
+          getSparkFacade().saveAsDataset(rdd, outputFormatClass, kClass, vClass, configuration);
+        } catch (Throwable t) {
+          // whatever went wrong, give the dataset a chance to handle the failure
+          if (dataset instanceof DatasetOutputCommitter) {
+            ((DatasetOutputCommitter) dataset).onFailure();
+          }
```
chtyim: Do you need to throw? Otherwise the `onSuccess()` will also get executed.

chtyim:
Just one comment left. Rest LGTM.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
anew: absolutely. It must have been late...

anew:
@chtyim I fixed your last comment, also added a test case for a custom dataset that implements onSuccess/onFailure.

Also a fix for the AppFabricTestHelper: it failed when deploying an app with a custom dataset. When deploying the dataset module as part of the pipeline, it failed because the namespace data dir does not exist. Apparently the AppFabricTestHelper implicitly creates the default namespace (or its data dir) somewhere later in the pipeline I could not figure out where, but the fix is to create the namespace before deploying the app.

Please do one more review. Thanks

chtyim:
LGTM

anew:
@chtyim one last commit... please review -thanks

chtyim:
LGTM.. again.

anew:
Merged this before the build finished. But the last commit was only changing one test case, and that succeeds on my laptop. Merged this to unblock @rohitsinha54 for creating today's demo.

PR #3149, opened 2015-07-13
Feature/partitioned file set arguments add partitions

anwar6953:
First commit (which is small) is the method added to PartitionedFileSetArguments, to aid in adding Partitions to be processed by an input dataset, and a two line change for `PartitionedFileSetDataset#getInputFormatConfiguration` to delegate to `FileSet#getInputFormatConfiguration` in the case that output paths are specified.

Second commit (most of this PR) is simply a test case to exercise the functionality of the first commit.

http://builds.cask.co/browse/CDAP-DUT2219-1

ignore: http://builds.cask.co/browse/CDAP-DUT2214-5

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
@@ -19,6 +19,7 @@ package co.cask.cdap.api.dataset.lib;
 import co.cask.cdap.api.annotation.Beta;
 import co.cask.cdap.api.dataset.lib.Partitioning.FieldType;
 
+import java.util.Iterator;
 import java.util.Map;
 import javax.annotation.Nullable;
 
@@ -121,6 +122,7 @@ public class PartitionedFileSetArguments {
    *
    * @param arguments the runtime arguments for a partitioned dataset
    * @param partitioning the declared partitioning for the dataset, needed for proper interpretation of values
+   * @return the PartitionFilter specified in the arguments or null if no filter is specified.
    */
   @Nullable
   public static PartitionFilter getInputPartitionFilter(Map<String, String> arguments, Partitioning partitioning) {
@@ -148,7 +150,7 @@ public class PartitionedFileSetArguments {
       @SuppressWarnings({ "unchecked", "unused" }) // we know it's type safe, but Java does not
       PartitionFilter.Builder unused = builder.addRangeCondition(fieldName, lowerValue, upperValue);
     }
-    return builder.build();
+    return builder.isEmpty() ? null : builder.build();
   }
 
   // helper to convert a string value into a field value in a partition key or filter
@@ -170,4 +172,15 @@ public class PartitionedFileSetArguments {
                       where, kind, stringValue, fieldName, fieldType.name()), e);
     }
   }
+
+  public static void addPartitions(Map<String, String> arguments, Iterator<Partition> partitionIterator) {
+    while (partitionIterator.hasNext()) {
+      addPartition(arguments, partitionIterator.next());
+    }
+  }
+
+  public static void addPartition(Map<String, String> arguments, Partition partition) {
+    FileSetArguments.addInputPath(arguments, partition.getRelativePath());
+  }
+
 }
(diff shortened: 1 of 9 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
```diff
@@ -170,4 +172,15 @@ public static Comparable convertFieldValue(String where, String kind, String fie
                       where, kind, stringValue, fieldName, fieldType.name()), e);
     }
   }
+
+  public static void addPartitions(Map<String, String> arguments, Iterator<Partition> partitionIterator) {
```
anew: shouldn't this be called addInputPartitions?

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
```diff
@@ -170,4 +172,15 @@ public static Comparable convertFieldValue(String where, String kind, String fie
                       where, kind, stringValue, fieldName, fieldType.name()), e);
     }
   }
+
```
anew: This is in API, so why is there no Javadoc?
Especially, you should explain what happens if both an input partition filter, and one or more input partitions are specified.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
anew: Why does this take a partition and not a partition key? Since runtime arguments come from outside, we can't assume that the caller has an instance of the dataset that he can call getPartition() on. It is more generic to accept a partition key. Didn't we discuss that, that a PartitionKey is a special case of a PartitionFilter, where all fields are specified as exact values?

anew: Or is this meant to be _in addition_ to the method that adds an input partition key?

anwar6953: This PR isn't implementing multiple PartitionFilters (which would be used for setting PartitionKeys).
Instead, this method is just a wrapper around `FileSetArguments#addInputPath`, for the use case of PartitionedFileSetDataset delegating to FileSetDataset, when only the input paths are set in the arguments (an no PartitionFilter).

anew:
Nice test case, but I am not sure I understand the new addPartition(s) API.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
anwar6953: Since it's just a wrapper method, the new method on PartitionedFileSetArguments isn't necessary.

The main change of this PR that's needed is delegating to the underlying files dataset for getInputFormatConfiguration, in case the PartitionFilter is missing.
https://github.com/caskdata/cdap/pull/3149/files#diff-3c5d2259d081358844e619de52b3741aR517

anwar6953:
I also updated TPFS getInputConfigurationFormat code to require a PartitionFilter, if start and end times are missing (to be consistent with existent behavior).
Otherwise, test case was failing:
http://builds.cask.co/browse/CDAP-DUT2214-JOB1-3/test/case/31210170

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/PartitionedFileSetArguments.java
anwar6953: added relevant javadoc.

anew:
One more comment...

anwar6953:
Hmm, I didn't see a use case for when users would know the exact input paths for the time partitioned partitions.

Though I don't have a strong opinion, so if you think that's a good idea, I will undo the changes in the last commit (and simply remove the relevant assert in the test case).

anwar6953:
Updated to also allow TPFS to specify input paths.

anew:
LGTM

PR #3153, opened 2015-07-13
(CDAP-2836) Add support for restart specific CDAP system services instances

henry-cask:
Add support for restart for CDAP system services instances. Need to be able to send restart commands to individual instances in System Services for CDAP Master.

Currently, to restart one of the CDAP master services (data-fabric, explore, metrics, etc.), you have to restart them all. It should be possible to just stop/start/restart a specific one.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/store/ServiceStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/store/ServiceStore.java
@@ -16,6 +16,7 @@
 
 package co.cask.cdap.app.store;
 
+import co.cask.cdap.proto.RestartServiceInstancesStatus;
 import com.google.common.util.concurrent.Service;
 
 import javax.annotation.Nullable;
@@ -39,4 +40,35 @@ public interface ServiceStore extends Service {
    * @param instances Instance Count.
    */
   void setServiceInstance(String serviceName, int instances);
+
+  /**
+   * Update the service instances restart request.
+   *
+   * @param serviceName Service name to be restarted.
+   * @param startTime Start time in Ms from Epoch.
+   * @param endTime End time in Ms from Epoch.
+   * @param isSuccess Whether or not the operation successful.
+   * @param instanceId The instance Id to be restarted.
+   */
+  void setRestartInstanceRequest(String serviceName, long startTime, long endTime, boolean isSuccess,
+                                  int instanceId);
+
+  /**
+   * Update the service instances restart request.
+   *
+   * @param serviceName Service name to be restarted.
+   * @param startTime Start time in Ms from Epoch.
+   * @param endTime End time in Ms from Epoch.
+   * @param isSuccess Whether or not the operation successful.
+   */
+  void setRestartAllInstancesRequest(String serviceName, long startTime, long endTime, boolean isSuccess);
+
+  /**
+   * Get the latest service instances restart as JSON String.
+   *
+   * @param serviceName Service name for the restart record.
+   * @return JSON string representation of latest restart instances for the service.
+   * @throws IllegalStateException when restart request can not be found for the service name.
+   */
+  RestartServiceInstancesStatus getLatestRestartInstancesRequest(String serviceName) throws IllegalStateException;
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
@@ -29,6 +29,8 @@ import com.google.gson.JsonSyntaxException;
 import com.google.inject.Inject;
 import org.jboss.netty.handler.codec.http.HttpRequest;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -41,6 +43,8 @@ import java.util.TreeSet;
  * Monitor Handler returns the status of different discoverable services
  */
 public class AbstractMonitorHandler extends AbstractAppFabricHttpHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractMonitorHandler.class);
+
   private final Map<String, MasterServiceManager> serviceManagementMap;
   private static final String STATUSOK = Constants.Monitor.STATUS_OK;
   private static final String STATUSNOTOK = Constants.Monitor.STATUS_NOTOK;
@@ -202,4 +206,55 @@ public class AbstractMonitorHandler extends AbstractAppFabricHttpHandler {
     }
   }
 
+  public void restartAllServiceInstances(HttpRequest request, HttpResponder responder, String serviceName) {
+    restartInstances(responder, serviceName, -1);
+  }
+
+  public void restartServiceInstance(HttpRequest request, HttpResponder responder, String serviceName,
+                                     int instanceId) {
+    restartInstances(responder, serviceName, instanceId);
+  }
+
+  private void restartInstances(HttpResponder responder, String serviceName, int instanceId) {
+    long startTimeMs = System.currentTimeMillis();
+    boolean isSuccess = true;
+    try {
+      if (!serviceManagementMap.containsKey(serviceName)) {
+        responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Invalid service name %s", serviceName));
+        return;
+      }
+
+      MasterServiceManager masterServiceManager = serviceManagementMap.get(serviceName);
+      if (instanceId < 0) {
+        masterServiceManager.restartAllInstances();
+      } else {
+        masterServiceManager.restartInstances(instanceId);
+      }
+
+      responder.sendStatus(HttpResponseStatus.OK);
+    } catch (Exception ex) {
+      LOG.warn(String.format("Exception when trying to restart instances for service %s", serviceName), ex);
+
+      isSuccess = false;
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR,
+                           String.format("Error restarting instance %d for service: %s", instanceId, serviceName));
+    } finally {
+      // Record the time
+      long endTimeMs = System.currentTimeMillis();
+      if (instanceId < 0) {
+        serviceStore.setRestartAllInstancesRequest(serviceName, startTimeMs, endTimeMs, isSuccess);
+      } else {
+        serviceStore.setRestartInstanceRequest(serviceName, startTimeMs, endTimeMs, isSuccess, instanceId);
+      }
+    }
+  }
+
+  public void getLatestRestartServiceInstanceStatus(HttpRequest request, HttpResponder responder, String serviceName) {
+    try {
+      responder.sendJson(HttpResponseStatus.OK, serviceStore.getLatestRestartInstancesRequest(serviceName));
+    } catch (IllegalStateException ex) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND,
+                           String.format("No restart instances request found or %s", serviceName));
+    }
+  }
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/DatasetServiceStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/DatasetServiceStore.java
@@ -24,14 +24,26 @@ import co.cask.cdap.data2.datafabric.dataset.DatasetsUtil;
 import co.cask.cdap.data2.dataset2.DatasetFramework;
 import co.cask.cdap.data2.dataset2.lib.kv.NoTxKeyValueTable;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.RestartServiceInstancesStatus;
+import co.cask.cdap.proto.RestartServiceInstancesStatus.RestartStatus;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.DiscreteDomains;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Ranges;
 import com.google.common.util.concurrent.AbstractIdleService;
+import com.google.gson.Gson;
 import com.google.inject.Inject;
 import com.google.inject.name.Named;
+
+import java.util.Set;
 
 /**
  * DatasetService Store implements ServiceStore using Datasets without Transaction.
  */
 public final class DatasetServiceStore extends AbstractIdleService implements ServiceStore {
+  private static final Gson GSON = new Gson();
+
   private final DatasetFramework dsFramework;
   private NoTxKeyValueTable table;
 
@@ -63,5 +75,45 @@ public final class DatasetServiceStore extends AbstractIdleService implements Se
   @Override
   protected void shutDown() throws Exception {
     table.close();
+  }
+
+  @Override
+  public synchronized void setRestartInstanceRequest(String serviceName, long startTimeMs, long endTimeMs,
+                                                     boolean isSuccess, int instanceId) {
+    Preconditions.checkNotNull(serviceName, "Service name should not be null.");
+    Preconditions.checkArgument(instanceId >= 0, "Instance id has to be greater than or equal to zero.");
+
+    RestartStatus status = isSuccess ? RestartStatus.SUCCESS : RestartStatus.FAILURE;
+    RestartServiceInstancesStatus restartStatus =
+      new RestartServiceInstancesStatus(serviceName, startTimeMs, endTimeMs, status, ImmutableSet.of(instanceId));
+    String toJson = GSON.toJson(restartStatus, RestartServiceInstancesStatus.class);
+
+    table.put(Bytes.toBytes(serviceName + "-restart"), Bytes.toBytes(toJson));
+  }
+
+  @Override
+  public synchronized void setRestartAllInstancesRequest(String serviceName, long startTimeMs, long endTimeMs,
+                                                         boolean isSuccess) {
+    Preconditions.checkNotNull(serviceName, "Service name should not be null.");
+
+    RestartStatus status = isSuccess ? RestartStatus.SUCCESS : RestartStatus.FAILURE;
+    int instanceCount = (this.getServiceInstance(serviceName) == null) ? 0 : this.getServiceInstance(serviceName);
+    Set<Integer> instancesToRestart = Ranges.closedOpen(0, instanceCount).asSet(DiscreteDomains.integers());
+
+    RestartServiceInstancesStatus restartStatus =
+      new RestartServiceInstancesStatus(serviceName, startTimeMs, endTimeMs, status, instancesToRestart);
+    String toJson = GSON.toJson(restartStatus, RestartServiceInstancesStatus.class);
+
+    table.put(Bytes.toBytes(serviceName + "-restart"), Bytes.toBytes(toJson));
+  }
+
+  @Override
+  public synchronized RestartServiceInstancesStatus getLatestRestartInstancesRequest(String serviceName)
+      throws IllegalStateException {
+    String jsonString = Bytes.toString(table.get(Bytes.toBytes(serviceName + "-restart")));
+    if (jsonString == null) {
+      throw new IllegalStateException("Unable to find latest restart request for " + serviceName);
+    }
+    return GSON.fromJson(jsonString, RestartServiceInstancesStatus.class);
   }
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MonitorHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MonitorHandler.java
@@ -91,4 +91,36 @@ public class MonitorHandler extends AbstractMonitorHandler {
   public void getServiceSpec(HttpRequest request, HttpResponder responder) throws Exception {
     super.getServiceSpec(request, responder);
   }
+
+  /**
+   * Send request to restart all instances for a CDAP system service.
+   */
+  @Path("/system/services/{service-name}/instances/restart")
+  @PUT
+  public void restartAllServiceInstances(HttpRequest request, HttpResponder responder,
+                                         @PathParam("service-name") String serviceName) {
+    super.restartAllServiceInstances(request, responder, serviceName);
+  }
+
+  /**
+   * Send request to restart single instance identified by <instance-id>
+   */
+  @Path("/system/services/{service-name}/instances/{instance-id}/restart")
+  @PUT
+  public void restartServiceInstance(HttpRequest request, HttpResponder responder,
+                                     @PathParam("service-name") String serviceName,
+                                     @PathParam("instance-id") int instanceId) {
+    super.restartServiceInstance(request, responder, serviceName, instanceId);
+  }
+
+  /**
+   * Send request to get the status of latest restart instances request for a CDAP system service.
+   */
+  @Path("/system/services/{service-name}/instances/restart")
+  @GET
+  public void getLatestRestartServiceInstanceStatus(HttpRequest request, HttpResponder responder,
+                                                    @PathParam("service-name") String serviceName) {
+    super.getLatestRestartServiceInstanceStatus(request, responder, serviceName);
+  }
+
 }
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
@@ -18,6 +18,7 @@ package co.cask.cdap.internal.app.services.http.handlers;
 
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.internal.app.services.http.AppFabricTestBase;
+import co.cask.cdap.proto.RestartServiceInstancesStatus;
 import co.cask.cdap.proto.SystemServiceMeta;
 import com.google.common.base.Charsets;
 import com.google.common.io.ByteStreams;
@@ -35,6 +36,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Monitor handler tests.
@@ -96,5 +98,29 @@ public class MonitorHandlerTest extends AppFabricTestBase {
     urlConn.disconnect();
     Assert.assertEquals(1, (int) result.get("requested"));
     Assert.assertEquals(1, (int) result.get("provisioned"));
+  }
+
+  @Test
+  public void testRestartInstances() throws Exception {
+    String path = String.format("system/services/%s/instances/restart", Constants.Service.APP_FABRIC_HTTP);
+    HttpURLConnection urlConn = openURL(path, HttpMethod.PUT);
+
+    Assert.assertEquals(HttpResponseStatus.OK.getCode(), urlConn.getResponseCode());
+
+    urlConn.disconnect();
+
+    urlConn = openURL(path, HttpMethod.GET);
+
+    Assert.assertEquals(HttpResponseStatus.OK.getCode(), urlConn.getResponseCode());
+
+    RestartServiceInstancesStatus result =
+      GSON.fromJson(new String(ByteStreams.toByteArray(urlConn.getInputStream()), Charsets.UTF_8),
+                    RestartServiceInstancesStatus.class);
+
+    urlConn.disconnect();
+
+    Assert.assertNotNull(result);
+    Assert.assertEquals(Constants.Service.APP_FABRIC_HTTP, result.getServiceName());
+    Assert.assertEquals(RestartServiceInstancesStatus.RestartStatus.SUCCESS, result.getStatus());
   }
 }
--- a/cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
@@ -179,4 +179,31 @@ public abstract class AbstractDistributedMasterServiceManager implements MasterS
       }
     }
   }
+
+  @Override
+  public void restartAllInstances() {
+    try {
+      Iterable<TwillController> twillControllers = twillRunnerService.lookup(Constants.Service.MASTER_SERVICES);
+      for (TwillController twillController : twillControllers) {
+        // Call restart instances
+        twillController.restartAllInstances(serviceName).get();
+      }
+    } catch (Throwable t) {
+      throw new RuntimeException(String.format("Could not change service instances of %s", serviceName), t);
+    }
+  }
+
+  @Override
+  public void restartInstances(int instanceId, int... moreInstanceIds) {
+    // Get Twill controller for the service
+    try {
+      Iterable<TwillController> twillControllers = twillRunnerService.lookup(Constants.Service.MASTER_SERVICES);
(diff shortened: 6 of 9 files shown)


henry-cask:
This one pending Apache Twill PR:
https://github.com/apache/incubator-twill/pull/52

cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
```diff
@@ -179,4 +179,41 @@ protected final HttpResponseStatus checkGetStatus(String url) throws Exception {
...
+    return true;
+  }
+
+  @Override
+  public boolean restartInstance(int instanceId, int... moreInstanceIds) {
+    Preconditions.checkArgument(serviceName != null);
+    // Get Twill controller for the service
+    try {
+      Iterable<TwillController> twillControllerList = twillRunnerService.lookup(Constants.Service.MASTER_SERVICES);
+      if (twillControllerList != null) {
```
chtyim: I don't think lookup will every return null.

hsaputra: Ah cool, will just iterate through it then =)

henry-cask: Ah cool, will just iterate through it then =)

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/ServiceStore.java
```diff
@@ -39,4 +39,25 @@
...
+  /**
+   * Update the service instances restart request.
+   *
+   * @param serviceName Service name to be restarted.
+   * @param startTime Start time in Ms from Epoch.
+   * @param endTime End time in Ms from Epoch.
+   * @param isSuccess Whether or not the operation successful.
+   * @param instanceId The instance Id to be restarted. If null then it restart all instances.
+   */
+  void setRestartInstancesRequest(String serviceName, long startTime, long endTime, boolean isSuccess,
```
jwang47: Seems weird to restart all if `instanceId` is null. Why not have two methods, `requestInstanceRestart` and `requestAllInstancesRestart`?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/DatasetServiceStore.java
```diff
@@ -64,4 +72,41 @@ protected void startUp() throws Exception {
   protected void shutDown() throws Exception {
     table.close();
   }
+
+  @Override
+  public synchronized void setRestartInstancesRequest(String serviceName, long startTimeMs, long endTimeMs,
+                                                      boolean isSuccess, Integer instanceId) {
```
jwang47: `@Nullable Integer instanceId`

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/DatasetServiceStore.java
```diff
@@ -64,4 +72,41 @@ protected void startUp() throws Exception {
...
+    RestartServiceInstancesStatus.RestartStatus status = RestartServiceInstancesStatus.RestartStatus.SUCCESS;
```
jwang47: Why not use `status = isSuccess ? SUCCESS : FAILURE` instead of setting then setting again conditionally?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/DatasetServiceStore.java
```diff
@@ -64,4 +72,41 @@ protected void startUp() throws Exception {
...
+    if (!isSuccess) {
+      status = RestartServiceInstancesStatus.RestartStatus.FAILURE;
+    }
+
+    ImmutableSet.Builder<Integer> instanceIdsBuilder = ImmutableSet.builder();
+    if (instanceId == null) {
+      int instanceCount = (this.getServiceInstance(serviceName) == null) ? 0 : this.getServiceInstance(serviceName);
```
jwang47: Can alternatively use `Objects.firstNonNull()`.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MonitorHandler.java
```diff
@@ -91,4 +91,36 @@ public void monitor(HttpRequest request, HttpResponder responder,
...
+   */
+  @Path("/system/services/{service-name}/instances/restart")
+  @PUT
+  public void restartServiceInstances(HttpRequest request, HttpResponder responder,
+                                      @PathParam("service-name") final String serviceName) {
+    super.restartServiceInstances(request, responder, serviceName);
+  }
+
+  /**
+   * Send request to restart all instances for a CDAP system service.
```
jwang47: Fix javadoc

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/ServiceStore.java
gsps1: +1

cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
```diff
@@ -179,4 +179,41 @@ protected final HttpResponseStatus checkGetStatus(String url) throws Exception {
       }
     }
   }
+
+  @Override
+  public boolean restartAllInstances() {
+    Preconditions.checkArgument(serviceName != null);
```
jwang47: Better to have a slightly more descriptive message here. E.g. "Name of service to restart is missing"

cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
```diff
@@ -179,4 +179,41 @@ protected final HttpResponseStatus checkGetStatus(String url) throws Exception {
...
+      }
+    } catch (Throwable t) {
+      LOG.error("Could not change service instance of {} : {}", serviceName, t.getMessage(), t);
+      return false;
+    }
+    return true;
+  }
+
+  @Override
+  public boolean restartInstance(int instanceId, int... moreInstanceIds) {
```
jwang47: Why return boolean instead of throwing exception?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -202,4 +202,72 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+        return;
+      }
+
+      // Need to get Twill controller for the service.
+      MasterServiceManager masterServiceManager = serviceManagementMap.get(serviceName);
+      if (masterServiceManager.restartAllInstances()) {
+        responder.sendStatus(HttpResponseStatus.OK);
+      } else {
+        isSuccess = false;
+        responder.sendString(HttpResponseStatus.BAD_REQUEST, "Operation did not succeed");
```
gsps1: it can also fail due to Internal_Server_Error right?

gsps1: api isn't descriptive about the failure case.

jwang47: Shouldn't this actually be Internal Server Error rather than Bad Request? Bad Request is more for something wrong with the input/request.

henry-cask: We could use Internal Server Error.

cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
henry-cask: It is consistent with rest of methods in this class.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -202,4 +202,72 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+      MasterServiceManager masterServiceManager = serviceManagementMap.get(serviceName);
+      if (masterServiceManager.restartInstance(instanceId)) {
+        responder.sendStatus(HttpResponseStatus.OK);
+      } else {
+        isSuccess = false;
+        responder.sendString(HttpResponseStatus.BAD_REQUEST, "Operation did not succeed");
+      }
+
+    } catch (Exception e) {
+      isSuccess = false;
```
gsps1: could be useful to add log here. LOG.Warn ?

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/ServiceStore.java
henry-cask: Agree with 2 separate methods

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MonitorHandler.java
henry-cask: Fix which part?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -202,4 +202,72 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+                           String.format("Error starting instance %d for service: %s", instanceId, serviceName));
+    } finally {
+      // Record the time
+      long endTimeMs = System.currentTimeMillis();
+      serviceStore.setRestartInstancesRequest(serviceName, startTimeMs, endTimeMs, isSuccess, null);
+    }
+  }
+
+  public void getLatestRestartServiceInstance(HttpRequest request, HttpResponder responder, String serviceName) {
+    Object restartRequest = serviceStore.getLatestRestartInstancesRequest(serviceName);
```
gsps1: why Object? why not RestartServiceInstancesStatus ?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -202,4 +202,72 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+    if (restartRequest == null) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND,
+                           String.format("No restart instances request for %s", serviceName));
```
gsps1: ``No restart instances request found for``

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MonitorHandler.java
jwang47: Isn't this method for restarting a single instance?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/DatasetServiceStore.java
```diff
@@ -64,4 +72,36 @@ protected void startUp() throws Exception {
...
+
+  @Override
+  public synchronized void setRestartInstancesRequest(String serviceName, long startTimeMs, long endTimeMs,
+                                                      boolean isSuccess, Integer instanceId) {
+    RestartServiceInstancesStatus.RestartStatus status = RestartServiceInstancesStatus.RestartStatus.SUCCESS;
+    if (!isSuccess) {
+      status = RestartServiceInstancesStatus.RestartStatus.FAILURE;
+    }
+
+    ImmutableSet.Builder<Integer> instanceIdsBuilder = ImmutableSet.builder();
```
gsps1: why do you need a builder ?  can just have an List<Integer> right ?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/DatasetServiceStore.java
henry-cask: agree

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -202,4 +202,72 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+                           String.format("Error starting instances for service: %s", serviceName));
+    } finally {
+      // Record the time
+      long endTimeMs = System.currentTimeMillis();
+      serviceStore.setRestartInstancesRequest(serviceName, startTimeMs, endTimeMs, isSuccess, null);
+    }
+  }
+
+  public void restartServiceInstance(HttpRequest request, HttpResponder responder, final String serviceName,
+                                     final int instanceId) {
```
gsps1: this method and ``restartServiceInstances`` are almost similar, can you remove duplication by extracting into a common method.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MonitorHandler.java
gsps1: invalid comment. should be ``restart single instance identified by <instance-id>``

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MonitorHandler.java
```diff
@@ -91,4 +91,36 @@ public void monitor(HttpRequest request, HttpResponder responder,
...
+  @Path("/system/services/{service-name}/instances/{instance-id}/restart")
+  @PUT
+  public void restartServiceInstance(HttpRequest request, HttpResponder responder,
+                                     @PathParam("service-name") final String serviceName,
+                                     @PathParam("instance-id") final int instanceId) {
+    super.restartServiceInstance(request, responder, serviceName, instanceId);
+  }
+
+  /**
+   * Send request to get the latest restart instances request for a CDAP system service.
```
gsps1: ``get the status of latest restart instances request``

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
```diff
@@ -97,4 +99,27 @@ public void testInstances() throws Exception {
...
   }
+
+  @Test
+  public void testRestartInstances() throws Exception {
+    String path = String.format("system/services/%s/instances/restart", Constants.Service.APP_FABRIC_HTTP);
+    HttpURLConnection urlConn = openURL(path, HttpMethod.PUT);
+
+    Assert.assertEquals(HttpResponseStatus.OK.getCode(), urlConn.getResponseCode());
+
+    TimeUnit.SECONDS.sleep(5L);
```
gsps1: anyway we can avoid this ?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
```diff
@@ -97,4 +99,27 @@ public void testInstances() throws Exception {
...
+
+    path = String.format("system/services/%s/instances/restart", Constants.Service.APP_FABRIC_HTTP);
```
gsps1: i don't see a change in path. remove?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
```diff
@@ -97,4 +99,27 @@ public void testInstances() throws Exception {
...
+    urlConn = openURL(path, HttpMethod.GET);
+
+    Assert.assertEquals(HttpResponseStatus.OK.getCode(), urlConn.getResponseCode());
+
+    RestartServiceInstancesStatus result =
+      GSON.fromJson(new String(ByteStreams.toByteArray(urlConn.getInputStream()), Charsets.UTF_8),
+                    RestartServiceInstancesStatus.class);
```
gsps1: we usually recommend using HttpGet, HttpPost api. that will make the decoding easier as well. as you can do something like, ``GSON.fromJson(EntityUtils.toString(response.getEntity(), Charsets.UTF_8), RestartServiceInstancesStatus.class)``

cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
```diff
@@ -179,4 +179,38 @@ protected final HttpResponseStatus checkGetStatus(String url) throws Exception {
       }
     }
   }
+
+  @Override
+  public boolean restartAllInstances() {
+    Preconditions.checkArgument(serviceName != null);
```
gsps1: you can use ``Preconditions.checkNotNull``

cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
```diff
@@ -179,4 +179,38 @@ protected final HttpResponseStatus checkGetStatus(String url) throws Exception {
...
+    try {
+      Iterable<TwillController> twillControllerList = twillRunnerService.lookup(Constants.Service.MASTER_SERVICES);
```
gsps1: rename to ``twillControllers`` ?

cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
```diff
@@ -179,4 +179,38 @@ protected final HttpResponseStatus checkGetStatus(String url) throws Exception {
...
+      for (TwillController twillController : twillControllerList) {
+        // Call restart instances
+        twillController.restartAllInstances(serviceName).get();
+      }
+
+    } catch (Throwable t) {
+      LOG.error("Could not change service instance of {} : {}", serviceName, t.getMessage(), t);
```
gsps1: ``Could not restart``

cdap-common/src/main/java/co/cask/cdap/common/twill/MasterServiceManager.java
```diff
@@ -86,5 +86,20 @@
 
   SystemServiceLiveInfo getLiveInfo();
 
-  //TODO: Add method to get the metrics name to get event rate on UI
```
gsps1: is this to-do irrelevant? why is it removed

cdap-proto/src/main/java/co/cask/cdap/proto/RestartServiceInstancesStatus.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package co.cask.cdap.proto;
+
+import java.util.Set;
+
+/**
+ * Class to describe status of restart instances of service request.
+ */
+public class RestartServiceInstancesStatus {
+  private final Set<Integer> instanceIds;
```
gsps1: why this has to be a ``Set`` ?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/MonitorHandler.java
henry-cask: Ah yes, sorry the review did not show which method being commented on.

cdap-proto/src/main/java/co/cask/cdap/proto/RestartServiceInstancesStatus.java
jwang47: Makes sense to me. Doesn't make sense to have multiple instances of the same ID in a restart status.

gsps1: i see. that makes sense.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
henry-cask: Yes we can but lets address it in bug fix mode to make sure we make time to feature freeze.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
henry-cask: One is GET and one is PUT

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
henry-cask: The other method is the test is using GSON so lets keep it that way and address it in separate PR.

cdap-proto/src/main/java/co/cask/cdap/proto/RestartServiceInstancesStatus.java
henry-cask: This is to prevent duplicate.

cdap-common/src/main/java/co/cask/cdap/common/twill/MasterServiceManager.java
henry-cask: Ah accidental remove. Will put it back

henry-cask: Ah no, this is due to some mistake I made commit to my local develop branch. Ignore this one =)

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: No log for rest of classes, will add more logging in separate PR.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/DatasetServiceStore.java
henry-cask: No longer needed with separate method.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: I rename the other as restartAllServiceInstances to separate restart all or an instance for a system.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
jwang47: +1, please create a JIRA and add a comment linking the JIRA issue though.

henry-cask: Issue filed: https://issues.cask.co/browse/CDAP-3004

henry-cask:
Updated the PR to address comments. CC @jwang47, @gsps1

jwang47:
Instead of amending the commit, please add a new commit when addressing comments as it makes it much easier to view the changes that addressed the comments. Then squash just before merging.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/ServiceStore.java
```diff
@@ -39,4 +40,35 @@
...
+   * Update the service instances restart request.
+   *
+   * @param serviceName Service name to be restarted.
+   * @param startTime Start time in Ms from Epoch.
+   * @param endTime End time in Ms from Epoch.
+   * @param isSuccess Whether or not the operation successful.
+   * @param instanceId The instance Id to be restarted.
+   */
+  void setRestartInstancesRequest(String serviceName, long startTime, long endTime, boolean isSuccess,
+                                  Integer instanceId);
```
jwang47: Should be `int` right? Since it's no longer nullable.

cdap-common/src/main/java/co/cask/cdap/common/twill/AbstractDistributedMasterServiceManager.java
jwang47: I think this should be changed, but probably in another PR.. seems bad to log exception here instead of propagating so that the caller knows what happened.

henry-cask: Yeah, we could send another PR to change all "set" methods to throw exceptions at fault. For now I prefer to keep it consistent.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/ServiceStore.java
henry-cask: Oops yeah, missing commit push

henry-cask:
Will do it next time

jwang47:
LGTM

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
gsps1: that should not change the string right

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
gsps1: oh, i meant the code inside the method.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
gsps1: we will still use GSON, instead of urlConnection we will use HttpPost and HttpGet. but its fine if you don't want to change it.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
henry-cask: Ah you mean the path? Will update the tests

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
henry-cask: Actually we dont need sleep here, we just need to verify the request is successfully submitted.

henry-cask:
I have updated the PR with some changes @jwang47, @gsps1, could you help do another pass?

jwang47:
LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: Ah got it, will converge into one helper method.

henry-cask:
@gsps1 updated the PR to address your concern

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -202,4 +202,51 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+
+      MasterServiceManager masterServiceManager = serviceManagementMap.get(serviceName);
+      isSuccess = (instanceId < 0) ? masterServiceManager.restartAllInstances() :
+        masterServiceManager.restartInstance(instanceId);
+
+      if (isSuccess) {
+        responder.sendStatus(HttpResponseStatus.OK);
+      } else {
+        // throw exception to have unified error handling.
+        throw new RuntimeException();
```
gsps1: have some message.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -202,4 +202,51 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+      }
+    } catch (Exception e) {
+      isSuccess = false;
```
gsps1: log the exception, so its useful for debug

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
jwang47: `masterServiceManager.restartAllInstances()` already logs, so I don't think it's necessary here. This does need to change though, probably in another PR.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -202,4 +202,51 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR,
+                           String.format("Error restarting instance %d for service: %s", instanceId, serviceName));
+    } finally {
+      // Record the time
+      long endTimeMs = System.currentTimeMillis();
+      serviceStore.setRestartInstanceRequest(serviceName, startTimeMs, endTimeMs, isSuccess, instanceId);
+    }
+  }
+
+  public void getLatestRestartServiceInstance(HttpRequest request, HttpResponder responder, String serviceName) {
```
gsps1: status? can you make the api consistent

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: Since the exception only used to unified error messages in catch clause

gsps1:
just minor nits. LGTM otherwise

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: will do

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: Lets just fix it then, I will change the underlying service manager to throw exception rather than returning boolean.

henry-cask:
Thanks for the review, if no more comments, pending successful build will rebase commits into one.

henry-cask:
Waiting for successful build ...

gsps1:
http://builds.cask.co/browse/CDAP-DUT2204-11

gsps1:
build green - merging

henry-cask:
Successful build: 
http://builds.cask.co/browse/CDAP-DUT2204-11

PR #3157, opened 2015-07-14
temporarily remove dependency on hbase-compat-1.0, until it actually …

dereklwood:
…be built

This will fix our internal auto-build process for the time-being, until the ``cdap-hbase-compat-1.0`` package can be built.  At that time, this change will be reverted.  ETA 1-2 days.


sreevatsanraman:
LGTM :+1:

ghelmling:
+1

PR #3158, opened 2015-07-14
fix deprecated size methods on svg elements

elfenheart:
Fix for dragging svg elements (flow-graph directive)


ajainarayanan:
:+1: LGTM

PR #3160, opened 2015-07-14
Feature/ui styling dev start spacing links

astewes:
*  Adjusts top margin of main.container for each tab state to reduce excess spacing underneath the namespace dropdown
*  System Health: Moves Services link to just within the status circle
*  System Health: Adds left/right padding to the warning/alert text

dev


elfenheart:
:+1: LGTM

PR #3161, opened 2015-07-14
CDAP-2868 Added hbase-compat-1.0 module and abstract out changed APIs

ghelmling:
This commit adds support for Apache HBase 1.0:
* enables the cdap-hbase-compat-1.0 module
* abstracts use of the methods which have changed between prior HBase versions and 1.0 in certain client classes: Scan, Get, Put, Delete, HTableDescriptor.  Here we use builder classes which override the method calls in a version-specific implementation where necessary.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/coprocessor/ConsumerConfigCache.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/coprocessor/ConsumerConfigCache.java
@@ -26,6 +26,7 @@ import co.cask.tephra.persist.TransactionSnapshot;
 import co.cask.tephra.util.TxUtils;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Supplier;
+import com.google.common.base.Throwables;
 import com.google.common.collect.Maps;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.TableNotFoundException;
@@ -38,6 +39,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.NavigableMap;
@@ -147,10 +150,8 @@ public class ConsumerConfigCache {
       // Scan the table with the transaction snapshot
       Scan scan = new Scan();
       scan.addFamily(QueueEntryRow.COLUMN_FAMILY);
-      scan.setCacheBlocks(false);
-      scan.setCaching(1000);
       Transaction tx = TxUtils.createDummyTransaction(txSnapshot);
-      scan.setAttribute(TxConstants.TX_OPERATION_ATTRIBUTE_KEY, txCodec.encode(tx));
+      setScanAttribute(scan, TxConstants.TX_OPERATION_ATTRIBUTE_KEY, txCodec.encode(tx));
       ResultScanner scanner = table.getScanner(scan);
       int configCnt = 0;
       for (Result result : scanner) {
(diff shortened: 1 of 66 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/coprocessor/ConsumerConfigCache.java
```diff
@@ -232,12 +235,28 @@ public void run() {
     refreshThread.start();
   }
 
-  public static ConsumerConfigCache getInstance(Configuration hConf, byte[] tableName,
+  /**
+   * Sets an attribute to the given {@link Scan} object. Instead of calling {@link Scan#setAttribute(String, byte[])}
+   * directly, it uses reflection to call the method. This is because the return type for the setAttribute method
+   * is different in different HBase version.
+   */
+  private void setScanAttribute(Scan scan, String name, byte[] value) {
```
abaranau: can we instead use ScanBuilder?

abaranau:
one Q. LGTM otherwise.

Should we consider extracting "abstraction for HBase API compatible with any 0.96+ version" into separate project? I can see it could be very useful for community for some apps/projects that use HBase. :)

abaranau:
As long as testing on cluster succeeds..

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/queue/hbase/coprocessor/ConsumerConfigCache.java
chtyim: Can't in here. This class is used inside Coprocessor, which doesn't have access to HBaseTableUtil. We can have the ScanBuilder being passed in through the getInstance method so that version specific coprocessor can pass in the right one. But seems like it just too much for a small operation like this.

ghelmling: This does not currently have access to HBaseTableUtil, which is I think why this bit was done differently.

ghelmling:
Still waiting on a successful build.  Latest is here: http://builds.cask.co/browse/CDAP-DUT2228-4

ghelmling:
Successfully ran PurchaseFlow and PurchaseHistoryBuilder MR job on a test cluster.

cdap-hbase-compat-1.0/src/main/java/co/cask/cdap/data2/transaction/queue/coprocessor/hbase10/HBaseQueueRegionObserver.java
```diff
@@ -137,13 +139,31 @@ public InternalScanner preCompact(ObserverContext<RegionCoprocessorEnvironment>
...
+
+  private ConsumerConfigCache getConfigCache(CoprocessorEnvironment env) {
     if (!configCache.isAlive()) {
-      configCache = ConsumerConfigCache.getInstance(conf, configTableNameBytes, cConfReader, txSnapshotSupplier);
+      configCache = createConfigCache(env);
     }
     return configCache;
   }
 
+  private ConsumerConfigCache createConfigCache(final CoprocessorEnvironment env) {
```
henry-cask: Do we need to check if env is not null?

chtyim: no. Why? It's a private method and it only gets called with the proper env.

henry-cask: Just checking, it if it guaranteed passed proper env then should be ok

henry-cask:
Reviewed the replacement changes and looks good, LGTM. Need to pass build to make sure.

ghelmling:
+1 on both additional commits to clean up resources

PR #3163, opened 2015-07-14

ajainarayanan:
Merging this change https://github.com/caskdata/cdap/pull/3117 to release/3.0

Cluster: http://release3-01996-1000.dev.continuuity.net:9999/ns/default


elfenheart:
:+1: LGTM

PR #3164, opened 2015-07-14
CDAP-3002: Tick initialDelay does not work properly

jwang47:
- Handle @Tick initialDelay by treating it as a delay instead of a
  timestamp relative to `System.nanoTime()`

https://issues.cask.co/browse/CDAP-3002
http://builds.cask.co/browse/CDAP-DUT2232

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowletProcessEntry.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowletProcessEntry.java
@@ -15,6 +15,7 @@
  */
 package co.cask.cdap.internal.app.runtime.flow;
 
+import com.google.common.math.LongMath;
 import com.google.common.primitives.Longs;
 
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowletProcessEntry.java
```diff
@@ -60,6 +71,10 @@ private FlowletProcessEntry(ProcessSpecification<T> processSpec, ProcessSpecific
     this.isTick = processSpec.isTick();
   }
 
+  public long getNextDeque() {
```
chtyim: Who call this?

jwang47: Oops. The unit test should be calling it, but I didn't do it properly.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/FlowletProcessEntry.java
```diff
@@ -42,11 +43,21 @@
...
   private long nextDeque;
   private long currentBackOff = BACKOFF_MIN;
 
   static <T> FlowletProcessEntry<T> create(ProcessSpecification<T> processSpec) {
-    return new FlowletProcessEntry<>(processSpec, null, processSpec.getInitialCallDelay());
+    long nextDeque;
+    try {
+      nextDeque = LongMath.checkedAdd(System.nanoTime(), processSpec.getInitialCallDelay());
+    } catch (ArithmeticException e) {
+      nextDeque = Long.MAX_VALUE;
```
poornachandra: Can you add a comment saying what `Long.MAX_VALUE

jwang47: No need to log, as the exception can only mean overflow.

chtyim:
LGTM

PR #3165, opened 2015-07-14
batch partition consuming helper

anwar6953:
Helper class `BatchPartitionConsumer.java` to remove some boilerplate code from the mapreduce job which wants to incrementally consume partitions.

In this PR, there's also an example app which shows how it can be used.

http://builds.cask.co/browse/CDAP-DUT2227-5


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/BatchPartitionConsumer.java
```diff
@@ -0,0 +1,79 @@
...
+ * A utility class to assist an incremental consumer of {@link Partition}s of a {@link PartitionedFileSet}. The
+ * implementing class is required to handle persistence of the consumption state via the abstract readBytes and
+ * writeBytes methods.
+ */
+@Beta
+public abstract class BatchPartitionConsumer {
+  private PartitionConsumerState finalConsumerState;
+
+  /**
+   * @return the serialized bytes of the state of the partition consuming MapReduce; return null to indicate a fresh
```
chtyim: Is it mapreduce specific?

anwar6953: Ah, it originally was implemented to be, but then tried to make it more generic (so also can be used from spark in the future).
I will update the javadoc.

anwar6953: docs updated.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/BatchPartitionConsumer.java
```diff
@@ -0,0 +1,79 @@
...
+  public PartitionedFileSet getPartitionedFileSet(DatasetContext datasetContext, String partitionedFileSetName) {
+    byte[] stateBytes = readBytes(datasetContext);
+    PartitionConsumerState partitionConsumerState =
+      stateBytes == null ? PartitionConsumerState.FROM_BEGINNING : PartitionConsumerState.fromBytes(stateBytes);
+
+    PartitionedFileSet partitionedFileSet = datasetContext.getDataset(partitionedFileSetName);
+    PartitionConsumerResult partitionConsumerResult = partitionedFileSet.consumePartitions(partitionConsumerState);
+    finalConsumerState = partitionConsumerResult.getPartitionConsumerState();
+
+    Map<String, String> arguments = Maps.newHashMap();
```
bdmogal: use Java 7 syntax instead of guava, more so coz this is in API, but even otherwise since its shorter now.

bdmogal:
other than 1 comment, LGTM

anwar6953:
Comments were addressed; pending build (few minutes remain).

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/BatchPartitionConsumer.java
```diff
@@ -0,0 +1,79 @@
...
+   * Used from the beforeSubmit method of the implementing batch job to get a PartitionedFileSet that has specified
+   * a set of {@link Partition}s of a {@link PartitionedFileSet} to be processed by the run of the batch job.
+   *
+   * @param datasetContext dataset context used to access the PartitionedFileSet.
+   * @param partitionedFileSetName the name of the PartitionedFileSet of which partitions will be processed.
+   * @return a PartitionedFileSet with arguments specified such that a call to its getInputFormatConfiguration method
+   * will return the paths of the Partitions that have yet to be consumed, according to the state deserialized from
+   * the implementing readBytes(DatasetContext) method.
+   */
+  public PartitionedFileSet getPartitionedFileSet(DatasetContext datasetContext, String partitionedFileSetName) {
```
anew: does this also work for TimePartitionedFileSet?

anwar6953: Since TimePartitionedFileSet delegates to PartitonedFileSet#getInputFormatConfiguration if neither start/end time are missing in the arguments, this will also work for TPFS.

PR #3166, opened 2015-07-14

ajainarayanan:
Porting this change https://github.com/caskdata/cdap/pull/3117/ from develop to release/2.8 branch

Cluster: http://release2-81995-1000.dev.continuuity.net:9999/ns/default


elfenheart:
:+1: LGTM

PR #3168, opened 2015-07-14
CDAP-2967 inspect artifacts for application classes

albertshau:
Enhance the ArtifactInspector to inspect the artifact for
Application Classes. For now, using the current application
requirement of having the application main class contained
in the artifact manifest.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactInspector.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactInspector.java
@@ -16,9 +16,11 @@
 
 package co.cask.cdap.internal.app.runtime.artifact;
 
+import co.cask.cdap.api.Config;
 import co.cask.cdap.api.annotation.Description;
 import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.app.Application;
 import co.cask.cdap.api.data.schema.UnsupportedTypeException;
 import co.cask.cdap.api.templates.plugins.PluginClass;
 import co.cask.cdap.api.templates.plugins.PluginConfig;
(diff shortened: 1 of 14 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2235

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactInspector.java
```diff
@@ -96,14 +113,72 @@ public ArtifactClasses inspectArtifact(Id.Artifact artifactId, File artifactFile
...
+        TypeToken<?> resultToken = typeToken.resolveType(Application.class.getTypeParameters()[0]);
+        Type configType;
+        // if the user parameterized their template, like 'xyz extends ApplicationTemplate<T>',
+        // we can deserialize the config into that object. Otherwise it'll just be a Config
+        if (resultToken.getType() instanceof Class) {
+          configType = resultToken.getType();
+        } else {
+          configType = Config.class;
+        }
+        apps.add(new ApplicationClass(mainClassName, "", schemaGenerator.generate(configType)));
```
gokulavasan: Any reason to generate and store schema of the application config class?

gokulavasan:
One question. LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactInspector.java
albertshau: this is so that people will know what to send for an app's config.  May not end up being exactly this, but wanted to at least put something here.

gokulavasan: Ok :+1:

PR #3175, opened 2015-07-15
Remove the unnecessary field

chtyim:
- We no longer need those fields since we always ship the spark-assembly.jar to the spark containers, which contains everything spark needs.


bdmogal:
:+1:

chtyim:
Verified on cluster, doesn't affect unit-test or standalone at all. Merging.

PR #3177, opened 2015-07-15
Fixes margin top for the entire view to be consistent (for namespace dropdown)

ajainarayanan:
Fixes the margin top on all view . The current style prevented user from accessing any part of UI on top of the page.


elfenheart:
:+1: LGTM

PR #3180, opened 2015-07-15
(CDAP-3012) Only localize spark assembly jar if the Workflow has Spar…

chtyim:
…k program inside.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkflowProgramRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkflowProgramRunner.java
@@ -15,7 +15,15 @@
  */
 package co.cask.cdap.internal.app.runtime.distributed;
 
+import co.cask.cdap.api.Resources;
+import co.cask.cdap.api.schedule.SchedulableProgramType;
+import co.cask.cdap.api.spark.SparkSpecification;
+import co.cask.cdap.api.workflow.ScheduleProgramInfo;
 import co.cask.cdap.api.workflow.Workflow;
+import co.cask.cdap.api.workflow.WorkflowActionNode;
+import co.cask.cdap.api.workflow.WorkflowConditionNode;
+import co.cask.cdap.api.workflow.WorkflowForkNode;
+import co.cask.cdap.api.workflow.WorkflowNode;
 import co.cask.cdap.api.workflow.WorkflowSpecification;
 import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.program.Program;
@@ -29,6 +37,7 @@ import co.cask.cdap.internal.app.runtime.spark.SparkContextConfig;
 import co.cask.cdap.internal.app.runtime.spark.SparkUtils;
 import co.cask.cdap.proto.ProgramType;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.twill.api.RunId;
@@ -38,7 +47,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
+import java.util.Queue;
+import javax.annotation.Nullable;
 
 /**
  * A {@link ProgramRunner} to start a {@link Workflow} program in distributed mode.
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkflowProgramRunner.java
```diff
@@ -66,14 +81,23 @@ protected ProgramController launch(Program program, ProgramOptions options,
     WorkflowSpecification workflowSpec = appSpec.getWorkflows().get(program.getName());
     Preconditions.checkNotNull(workflowSpec, "Missing WorkflowSpecification for %s", program.getName());
 
-    // Localize the spark-assembly jar, since workflow can execute spark.
-    File sparkAssemblyJar = SparkUtils.locateSparkAssemblyJar();
-    localizeFiles.put(sparkAssemblyJar.getName(), sparkAssemblyJar);
+    // It the workflow has Spark, localize the spark-assembly jar
+    List<String> extraClassPaths = new ArrayList<>();
```
henry-cask: Should probably use Lists.newArrayList() to be consistent

chtyim: We are switching to Java7 syntax, which is more compact and better be using language feature.

chtyim:
Verified on cluster with Workflow with/without Spark

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkflowProgramRunner.java
henry-cask: We are? Did not know about it

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkflowProgramRunner.java
```diff
@@ -84,4 +108,70 @@ private static Configuration createConfiguration(Configuration hConf) {
...
+    while (!nodes.isEmpty()) {
+      WorkflowNode node = nodes.poll();
+      switch (node.getType()) {
+        case ACTION: {
+          ScheduleProgramInfo programInfo = ((WorkflowActionNode) node).getProgram();
+          if (programInfo.getProgramType() == SchedulableProgramType.SPARK) {
+            hasSpark = true;
+            // The sparkSpec shouldn't be null, otherwise the Workflow is not valid
+            Resources driverResources = sparkSpecs.get(programInfo.getProgramName()).getDriverResources();
+            if (driverResources != null) {
```
henry-cask: If driverResources should not be null, is there a reason for the check?

bdmogal:
@chtyim the changes look good. However, the problem that we saw (other than the fact that we localize the spark assembly jar even when it is not required) is that because the spark assembly jar (~150MB) is localized, ``MapReduceRuntimeService`` throws OOM: Heap Space in ApplicationBundler/BundleJarUtil.

Will this work if the workflow has both a spark and a mapreduce program?

bdmogal:
We may need to somehow exclude the ``spark-assembly.jar`` while bundling the jar in ``MapReduceRuntimeService`` or increase the memory in such a case?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkflowProgramRunner.java
chtyim: `SparkSpecification.getDriverResources()` is `@Nullable`

chtyim: It's only the spec shouldn't be null (meaning the `sparkSpecs.get(...)` part, that's why I just chain it instead of getting it back and check for null).

chtyim:
For OOM shouldn't be affected by the assembly jar presence, since building of the MapReduce job jar is explicitly excluding spark and scala. However, since the spark assembly jar is a fat jar which contains a lot of "stuff" inside, maybe we should exclude the jar entirely as well (the latest twill has that capability). But anyway, it's a separate issue then this PR intended to fix.

chtyim:
And BTW, no matter how big the spark assembly jar is, it won't get loaded into memory entirely when building the MR job jar. So, the OOM maybe caused by something else that needs investigation.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkflowProgramRunner.java
chtyim: now you know :)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/DistributedWorkflowProgramRunner.java
henry-cask: Got it

chtyim:
Created https://issues.cask.co/browse/CDAP-3013 for the MR job jar building issue.

PR #3181, opened 2015-07-15
(CDAP-3013) Excludes Spark classes and classes from spark-assembly ja…

chtyim:
…r when building MR job jar.


chtyim:
Tested on cluster, no OOM observed. Also the job jar built for the MR doesn't contains spark jars/classes.

bdmogal:
:+1:

PR #3183, opened 2015-07-15
Change PartitionedFileSetDataset#dropPartition to recursively delete …

anwar6953:
…the partition files. Also throw dataset exception if the call to delete fails without throwing an exception.

The necessary change is passing `true` to the call to `delete`. The remainder is additional error-checking.

http://builds.cask.co/browse/CDAP-RBT309-4
https://issues.cask.co/browse/CDAP-3019

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
@@ -331,9 +331,13 @@ public class PartitionedFileSetDataset extends AbstractDataset implements Partit
     }
     if (!isExternal) {
       try {
-        partition.getLocation().delete();
+        boolean deleteSuccess = partition.getLocation().delete(true);
+        if (!deleteSuccess) {
+          throw new DataSetException(String.format("Error deleting file(s) for partition %s at path %s.",
+                                                   key, partition.getLocation().toURI().getPath()));
+        }
       } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


anwar6953:
@chtyim The behavior for HDFSLocation is different than LocalLocation when delete() is called without recursive=true and there existing files under that Location.
LocalLocation simply returns `false`, but `HDFSLocation` threw an exception.
Shouldn't this inconsistency not exist? (Would it be a good idea to make LocalLocation also throw an exception in that case)?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -330,11 +330,15 @@ public void dropPartition(PartitionKey key) {
       return;
     }
     if (!isExternal) {
+      String deleteFailureMessage = String.format("Error deleting file(s) for partition %s at path %s.",
```
anew: it is better to create the message only if there is an error.

chtyim:
Can you open a Twill issue with the stacktrace attached?

anwar6953:
Created: https://issues.apache.org/jira/browse/TWILL-140

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: ok; done.

anwar6953:
Added an additional check to check for file existence before delete (it's possible the user didn't create a file at that partition's path when creating a partition.

anew:
LGTM

PR #3185, opened 2015-07-15
CDAP-3001 fix classloading for getting user dataset admins

albertshau:
Fixing a bug introduced by the fix to CDAP-2564. The previous
change changed DatasetFramework to just use the classloader it
is given instead of trying to unpack dataset jars to create
classloaders. Since the CDAP system does not have access to
user dataset jars in its classpath, it needs to pass a
DatasetClassLoaderProvider to DatasetFramework, which is
responsible for unpacking user dataset jars to create the
classloaders needed to instantiate a dataset.

That refactoring was done for dataset instantiation, but not for
dataset admin instantiation. This fix adds a corresponding
method for instantiating a DatasetAdmin, which must be used when
the CDAP system needs to instantiate a DatasetAdmin for a user
dataset.


albertshau:
http://builds.cask.co/browse/CDAP-RBT310

chtyim:
The changes LGTM. From the description, it sounds like we should refactor the DatasetFramework to unify the code path for creating Dataset instances and creating DatasetAdmin instances.

PR #3188, opened 2015-07-15
Feature/cdap 2840 pfs upgrade

anwar6953:
PartitionedFileSet's partitions table used to be `Table`. It is now an `IndexedTable`, indexed on two fields: creation time and transaction write pointer. Both of these fields are columns on the rows that did not exist before 3.1.0.

https://issues.cask.co/browse/CDAP-2840
http://builds.cask.co/browse/CDAP-RBT311-3

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
@@ -1,0 +1,185 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data2.dataset2.lib.partitioned;
+
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.dataset.DatasetProperties;
+import co.cask.cdap.api.dataset.DatasetSpecification;
+import co.cask.cdap.api.dataset.lib.IndexedTable;
+import co.cask.cdap.data2.datafabric.dataset.DatasetsUtil;
+import co.cask.cdap.data2.dataset2.DatasetFramework;
+import co.cask.cdap.data2.dataset2.lib.table.hbase.HBaseTableAdmin;
+import co.cask.cdap.data2.util.TableId;
+import co.cask.cdap.data2.util.hbase.HBaseTableUtil;
+import co.cask.cdap.data2.util.hbase.ScanBuilder;
+import co.cask.cdap.proto.Id;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
+import com.google.inject.Inject;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.client.Delete;
+import org.apache.hadoop.hbase.client.HBaseAdmin;
+import org.apache.hadoop.hbase.client.HTable;
+import org.apache.hadoop.hbase.client.Put;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+import java.util.NavigableMap;
+
+/**
+ * Migrates the tables of a PartitionedFileSet.
+ */
+public class PartitionedFileSetTableMigrator {
+  private static final Logger LOG = LoggerFactory.getLogger(PartitionedFileSetTableMigrator.class);
+
+  // copied from IndexedTable.java. This upgrade is only needed from 3.0 -> 3.1, so this code will be removed
+  // after 3.1 release. This is done to avoid publicizing the fields on IndexedTable.
+  private static final byte[] IDX_COL = {'r'};
+  private static final byte DELIMITER_BYTE = 0;
+  private static final byte[] KEY_DELIMITER = new byte[] { DELIMITER_BYTE };
+
+  private static final long MILLION = 1000 * 1000;
+  protected final HBaseTableUtil tableUtil;
+  protected final Configuration conf;
+  private final DatasetFramework dsFramework;
+
+  @Inject
+  public PartitionedFileSetTableMigrator(HBaseTableUtil tableUtil, Configuration conf, DatasetFramework dsFramework) {
+    this.tableUtil = tableUtil;
+    this.conf = conf;
+    this.dsFramework = dsFramework;
+  }
+
+  /**
+   * Given a specification of a PartitionedFileSet, creates an IndexedTable for the new embedded partitions Dataset
+   * and migrates the data from the old partitions table to the new partitions table.
+   *
+   * @param namespaceId the namespace that contains the PartitionedFileSet
+   * @param dsSpec the specification of the PartitionedFileSet that needs migrating
+   * @throws Exception
+   */
+  public void upgrade(Id.Namespace namespaceId, DatasetSpecification dsSpec) throws Exception {
+    DatasetSpecification newPartitionsSpec = dsSpec.getSpecification(PartitionedFileSetDefinition.PARTITION_TABLE_NAME);
+    TableId oldPartitionsTableId = TableId.from(namespaceId, newPartitionsSpec.getName());
+    TableId indexTableId = TableId.from(namespaceId, newPartitionsSpec.getName() + ".i");
+    TableId dataTableId = TableId.from(namespaceId, newPartitionsSpec.getName() + ".d");
+
+    byte[] columnFamily = HBaseTableAdmin.getColumnFamily(newPartitionsSpec);
+
+    HBaseAdmin hBaseAdmin = new HBaseAdmin(conf);
+    if (!tableUtil.tableExists(hBaseAdmin, oldPartitionsTableId)) {
+      LOG.info("Old Partitions table does not exist: {}. Nothing to migrate from.", oldPartitionsTableId);
+      return;
+    }
+
+    DatasetsUtil.createIfNotExists(dsFramework, Id.DatasetInstance.from(namespaceId, newPartitionsSpec.getName()),
+                                   IndexedTable.class.getName(),
+                                   DatasetProperties.builder().addAll(newPartitionsSpec.getProperties()).build());
+
+    HTable oldTable = tableUtil.createHTable(conf, oldPartitionsTableId);
+    HTable newIndexTable = tableUtil.createHTable(conf, indexTableId);
+    HTable newDataTable = tableUtil.createHTable(conf, dataTableId);
+
+    LOG.info("Starting upgrade for table {}", Bytes.toString(oldTable.getTableName()));
+    try {
+      ScanBuilder scan = tableUtil.buildScan();
+      scan.setTimeRange(0, HConstants.LATEST_TIMESTAMP);
+      // migrate all versions. Since we write the same versions in the new table, transactional semantics should carry
+      // across (invalid transactions' data remains invalid, etc). There shouldn't be too many versions of a given row
+      // of the partitions table anyways.
+      scan.setMaxVersions();
+      try (ResultScanner resultScanner = oldTable.getScanner(scan.build())) {
+        Result result;
+        while ((result = resultScanner.next()) != null) {
+          Put put = new Put(result.getRow());
+
+          Long writeVersion = null;
+          for (Map.Entry<byte[], NavigableMap<byte[], NavigableMap<Long, byte[]>>> familyMap :
+            result.getMap().entrySet()) {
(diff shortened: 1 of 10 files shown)


anwar6953:
This needs to be updated to also work for TimePartitionedFileSets

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
```diff
@@ -100,7 +100,7 @@
   /**
    * Column key used to store the existence of a row in the secondary index.
    */
-  private static final byte[] IDX_COL = {'r'};
+  public static final byte[] IDX_COL = {'r'};
```
anew: does it have to be public? Since we are in the API, I am not sure we want to make this public.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
```diff
@@ -0,0 +1,175 @@
...
+   * @param namespaceId the namespace that contains the PartitionedFileSet
+   * @param dsSpec the specification of the PartitionedFileSet that needs migrating
+   * @throws Exception
+   */
+  public void upgrade(Id.Namespace namespaceId, DatasetSpecification dsSpec) throws Exception {
+    DatasetSpecification newPartitionsSpec = dsSpec.getSpecification(PartitionedFileSetDefinition.PARTITION_TABLE_NAME);
+    TableId oldPartitionsTableId = TableId.from(namespaceId, newPartitionsSpec.getName());
+    TableId indexTableId = TableId.from(namespaceId, newPartitionsSpec.getName() + ".i");
+    TableId dataTableId = TableId.from(namespaceId, newPartitionsSpec.getName() + ".d");
+
```
anew: remove extra newline

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
```diff
@@ -0,0 +1,175 @@
...
+    HTable oldTable = tableUtil.createHTable(conf, oldPartitionsTableId);
+    HTable newIndexTable = tableUtil.createHTable(conf, indexTableId);
+    HTable newDataTable = tableUtil.createHTable(conf, dataTableId);
+
+    LOG.info("Starting upgrade for table {}", Bytes.toString(oldTable.getTableName()));
+    try {
+      ScanBuilder scan = tableUtil.buildScan();
+      scan.setTimeRange(0, HConstants.LATEST_TIMESTAMP);
+      scan.setMaxVersions(1); // we only need to see one version of each row
```
anew: Not true. What if a partition was created by a MR job, and that job then failed? The transaction would be invalidated (and not rolled back). That is, the write remains, but its write version is now in the invalid list maintained by the tx manager. So, you need to talk to tx manager, get a transaction. Count the number of excludes in that transaction, and that is the max version that you need. Then, while reading, you need to filter out excluded versions.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
```diff
@@ -0,0 +1,175 @@
...
+          Long hBaseVersion = null;
+
+          for (Map.Entry<byte[], NavigableMap<byte[], NavigableMap<Long, byte[]>>> familyMap :
+            result.getMap().entrySet()) {
+            for (Map.Entry<byte[], NavigableMap<Long, byte[]>> columnMap : familyMap.getValue().entrySet()) {
+              for (Map.Entry<Long, byte[]> columnEntry : columnMap.getValue().entrySet()) {
+                Long timeStamp = columnEntry.getKey();
+                byte[] colVal = columnEntry.getValue();
+                put.add(familyMap.getKey(), columnMap.getKey(), timeStamp, colVal);
+                LOG.info("Migrating row. family: '{}', column: '{}', ts: '{}', colVal: '{}'",
```
anew: that would create a lot of log messages. Make it debug?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
```diff
@@ -0,0 +1,175 @@
...
+          // here, we make the assumption that dropping the six right-most digits of the transaction write pointer
+          // yields the timestamp at which it started
+          byte[] creationTimeBytes = Bytes.toBytes(hBaseVersion / MILLION);
+          put.add(columnFamily, PartitionedFileSetDataset.CREATION_TIME_COL,
+                  hBaseVersion, creationTimeBytes);
+
+          newDataTable.put(put);
+
+          // index the data table on the two columns
+          Put indexTableWritePointer = new Put(IndexedTable.createIndexKey(result.getRow(),
```
anew: call this variable indexPut instead of indexTableWritePointer? it is a put, not a write pointer...

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
```diff
@@ -0,0 +1,175 @@
...
+      ScanBuilder scan = tableUtil.buildScan();
+      scan.setTimeRange(0, HConstants.LATEST_TIMESTAMP);
+      scan.setMaxVersions(1); // we only need to see one version of each row
+      try (ResultScanner resultScanner = oldTable.getScanner(scan.build())) {
+        Result result;
+        while ((result = resultScanner.next()) != null) {
+          Put put = new Put(result.getRow());
+
+
+          Long hBaseVersion = null;
```
anew: why not call this the writeVersion?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
```diff
@@ -0,0 +1,175 @@
...
+      } finally {
+        oldTable.close();
+        newIndexTable.close();
+        newDataTable.close();
+      }
+      LOG.info("Successfully migrated data from table {} Now deleting it.", Bytes.toString(oldTable.getTableName()));
+      hBaseAdmin.deleteTable(oldTable.getTableName());
+      LOG.info("Succsefully deleted old data table {}", Bytes.toString(oldTable.getTableName()));
+    } catch (Exception e) {
+      LOG.error("Error while migrating data from table: {}", oldPartitionsTableId, e);
```
anew: In case of an error: Say I have migrated 1 out of 3 pfs's, and the 2nd one fails for some reason. The user will have to run the upgrade tool again. Will this method now fail for #1? And with it the upgrade tool?

cdap-master/src/main/java/co/cask/cdap/data/tools/AbstractQueueUpgrader.java
```diff
@@ -104,8 +104,8 @@ void upgrade() throws Exception {
               LOG.debug("Adding entry {} -> {} for upgrade",
                         Bytes.toString(entry.getKey()), Bytes.toString(entry.getValue()));
               put.add(QueueEntryRow.COLUMN_FAMILY, entry.getKey(), entry.getValue());
-              mutations.add(put);
             }
+            mutations.add(put);
```
anew: unrelated change?

cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
```diff
@@ -0,0 +1,217 @@
...
+
+  boolean alreadyUpgraded(DatasetSpecification dsSpec) {
+    DatasetSpecification partitionsSpec = dsSpec.getSpecification(PartitionedFileSetDefinition.PARTITION_TABLE_NAME);
+    // the partitions table is now an IndexedTable (was Table in < v3.1.0 CDAP)
+    return isIndexedTable(partitionsSpec);
+  }
+
+  boolean isPartitionedFileSet(DatasetSpecification dsSpec) {
+    String dsType = dsSpec.getType();
+    return (PartitionedFileSet.class.getName().equals(dsType) || "partitionedFileSet".equals(dsType));
```
anew: add support for timePartitionedFileSet?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
anwar6953: Hmm, a simpler method would to simple get all versions, and write them using the same versions as in the original table, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
anew: Never mind. I see that is dealt with in the PFSUpgrader

anew: Actually, this can happen. In the very rare case that a dataset embeds two PFS, the PFSUpgrader first rewrites the spec and stores it in MDS, then attempts to migrate both embedded datasets. If the second one fails, then the user retries the upgrade, it will not even attempt to migrate the embedded tables again.

anew: oh and you check for that before opening it. I think it is correct the way it is.

anew:
@anwar6953 please see my comments.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
anwar6953: Hmm, the only thing I still need to do is delete the old table, once it's done migrating (which isn't absolutely necessary, though).

cdap-master/src/main/java/co/cask/cdap/data/tools/AbstractQueueUpgrader.java
anwar6953: yea, AbstractQueueUpgrader was adding the Put to the mutations before the put was fully built.
Changed it to only add it to the mutation, once all changes to the put have been mode (After the for loop).

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anwar6953: It seems difficult to not make this public, since the migrator class is in the same package is `PartitionedFileSet`, in order to make the variables we need in `PartitionedFileSet` not public, but just protected.
If we instantiated the indexed table and did transactional writes to it, that would be one way to get round it.
Alternatively, we could at least document that it shouldn't be used?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
anwar6953: https://github.com/caskdata/cdap/commit/8e6cf081518516d5ed40ea81a965f54a714a073e

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
anwar6953: Nevermind, it was already deleting the table. Changed it to use tableUtil's dropTable though, since that disables+deletes.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
anwar6953: done

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/IndexedTable.java
anew: Since this will only be done one time (3.0 -> 3.1) and we will delete this code after 3.1, I almost think it is better to copy it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetTableMigrator.java
anew: yeah and that is ok, because in 99%+ of the cases there will only be one version.

anew:
Looks good except that I think the IndexedTable constants should not be made public.

anew:
Oh, and add a test case for TPFS?

anwar6953:
Added tpfs spec upgrade.
Once code review is complete, I'll test on a cluster once again (due to code refactoring).

anew:
LGTM. Please test on a cluster before merging.

anew:
can this be merged?

anwar6953:
I have yet to test this again after the refactoring. Will do that today or early tomorrow and merge following that.

anwar6953:
@andreas pushed one last commit, which resolves an issue which I discovered while testing the upgrade.
Other than that, this is good to merge.

anew:
Good catch. I could have found that in the first round of review... 
LGTM

anwar6953:
Squashed commits; will merge once build passes.

PR #3189, opened 2015-07-15
(CDAP-3016) Tune down number of HBase threads

chtyim:
- Tune down hconnection threads and regionserver handler count.
- Refactor HBaseTestBase code so that more configurations can be set before starting mini HBase cluster for testing.

### DIFF
--- a/cdap-data-fabric/src/test/java/co/cask/cdap/data/hbase/HBaseTestBase.java
+++ b/cdap-data-fabric/src/test/java/co/cask/cdap/data/hbase/HBaseTestBase.java
@@ -55,7 +55,16 @@ public abstract class HBaseTestBase {
 
   // Test startup / teardown
 
-  public abstract void startHBase() throws Exception;
+  public final void startHBase() throws Exception {
+    // Tune down the connection thread pool size
+    getConfiguration().setInt("hbase.hconnection.threads.core", 5);
+    getConfiguration().setInt("hbase.hconnection.threads.max", 10);
+    // Tunn down handler threads in regionserver
+    getConfiguration().setInt("hbase.regionserver.handler.count", 10);
+    doStartHBase();
+  }
+
+  public abstract void doStartHBase() throws Exception;
 
   public abstract void stopHBase() throws Exception;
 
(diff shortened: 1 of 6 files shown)


chtyim:
Test passed locally and verified there are much fewer threads than before. Suggest merging this before bamboo build to unblock other PR builds.

sreevatsanraman:
LGTM :+1:

cdap-data-fabric/src/test/java/co/cask/cdap/data/hbase/HBaseTestBase.java
```diff
@@ -55,7 +55,16 @@ public String getZkConnectionString() {
 
   // Test startup / teardown
 
-  public abstract void startHBase() throws Exception;
+  public final void startHBase() throws Exception {
+    // Tune down the connection thread pool size
+    getConfiguration().setInt("hbase.hconnection.threads.core", 5);
+    getConfiguration().setInt("hbase.hconnection.threads.max", 10);
```
ghelmling: These settings will only be effective when you call HConnection.getTable() to obtain table instances, which I don't think we really do anywhere.

We should set hbase.htable.threads.max to limit the number of threads in the thread pool constructed when instantiating HTable directly -- new HTable(Configuration, TableName).  I think this is the usage we have almost everywhere.

chtyim: Can add that too. From the testing I conducted, there are far more HConnection threads than HTable threads (probably because we are closing HTable connections during test, but not HConnection. Also, the HTable thread pool has one core thread, while the HConnection thread pool has 256 core threads).

ghelmling: Acutally, that's not complete true.  From the thead dumps, it looks like the HConnection pool is getting allocated somewhere.  Looking in to why that is.

ghelmling: It's through use of the RegionCoprocessorEnvironment.getTable() call.  So this config change will have a big impact in the ConsumerConfigCache usage.

The HTable config I suggested may not make much of a difference since HTable always sets the core pool size to 1.

I'll keep looking for other configs which could help.

PR #3191, opened 2015-07-15
[CDAP 3021] Accept cloudfront file naming convention

jayfjin:
JIRA - https://issues.cask.co/browse/CDAP-3021
BUILD - http://builds.cask.co/browse/CDAP-DUT2250

Cloudfront names files in a different format than S3 log files. Update BatchFileFilter to read cloudfront filenames.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
@@ -44,7 +44,7 @@ import java.util.regex.Pattern;
  * a string and passed in as dateRangesToRead, which is described below. The files are assumed to succeed so that files
  * will not be read in multiple times. Thus, a singleton element is written into the table as the list denoting that
  * the only files that haven't been read in are after that element. If the read fails, then dateRangesToRead with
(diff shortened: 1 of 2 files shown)


jayfjin:
build passed http://builds.cask.co/browse/CDAP-DUT2250-3

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/BatchFileFilter.java
```diff
@@ -52,54 +52,59 @@
 public class BatchFileFilter extends Configured implements PathFilter {
 
   private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
-  //length of 'YYYY-MM-dd-HH-mm"
-  private static final int DATE_LENGTH = 16;
+  //length of 'YYYY-MM-dd-HH"
+  private static final int DATE_LENGTH = 13;
```
anwar6953: can we just derive this from either the pattern string directly, or `sdf.toPattern().length()`, so we don't need to compute this manually

jayfjin: this would make DATE_LENGTH non-static since sdf is not static for concurrency reasons. would we prefer that? or just eliminate the field entirely and use sdf.toPattern().length() directly.

anwar6953:
lgtm

PR #3194, opened 2015-07-15
(CDAP-3016) limit max size of HTable thread pool to 10

chtyim:


ghelmling:
lgtm

chtyim:
Seems like this change is not needed for unit-test. Closing it for now. It's an easy change, so we can add it if proved to be needed.

PR #3200, opened 2015-07-15
Feature/ui adapter create styling

astewes:
*  Adds new grid background image and styling to the group side panels 
*  Bootstraps adapter modal content
*  Removes some unused code
*  Replaces adapter shape background colors with backgrounded images
*  Adds variable icon size to sidebar item headers

cluster name: adapter-create


cdap-ui/app/directives/group-side-panel/group-side-panel.less
```diff
@@ -7,29 +9,68 @@ my-side-panel {
...
     }
 
     &.right {
+      .btn.btn-info.btn-block {
+        text-align: right;
+      }
       right: 10px;
+      .item-heading {
+        span.fa:first-child {
+          font-size: 15px;
```
ajainarayanan: Should we make this consistent? Left side panel has different font size on open and closed and right side panel has different one.

cdap-ui/app/features/adapters/etlapps.less
```diff
@@ -125,25 +122,317 @@ body.theme-cdap.state-adapters-create {
...
+    .modal-header {
+      background-color: @body-bg;
+      color: @cdap-header;
+      height: 60px;
+      padding: 0;
+      .border-radius(4px 4px 0 0);
+      div.pull-left {
+        height: 60px;
+        padding-left: 15px;
+        max-width: 725px;
```
ajainarayanan: I believe this would cause issues if the browser is resized. It be might be better if we use the bootstrap grid here. On a laptop full browser size it looks perfect but should scale atleast for md screen size. Thoughts?

cdap-ui/app/features/adapters/etlapps.less
```diff
@@ -125,25 +122,317 @@ body.theme-cdap.state-adapters-create {
...
+        color: @cdap-header;
+        display: block;
+        font-size: 18px;
+        line-height: 60px;
+        padding: 0;
+        width: 100%;
+      }
+    }
+    .modal-body {
+      background-color: white;
```
ajainarayanan: When we have more properties/schema fields (output schema) and reduce the height of browser the content beneath gets hidden (in-accessible). May be setting height only on certain screen sizes might be re-thought?

cdap-ui/app/features/adapters/etlapps.less
```diff
@@ -125,25 +122,317 @@ body.theme-cdap.state-adapters-create {
...
+      > .row > [class*="col-"] {
+        padding-top: 20px;
+      }
+      .modal-content-scroll {
+        @media (min-width: @screen-lg-min) {
+          height: 700px;
+          overflow: scroll;
+        }
+      }
+      .col-lg-2 {
```
ajainarayanan: I am not an avid fan of styling specific bootstrap grid classes for specific screen sizes. We style bootstrap columns in a more generic way like this,
```
[class*="col-lg"] {
....
}
```
And this applies to all bootstrap large classes across all screen sizes.  Even if a widget under the modal uses a bootstrap grid class it would be affected by it. In this case we are modifying col-lg-2 class alone and for specific screen sizes. If a widget in the modal happens to use the grid would it be affected by this in those screen sizes. Is it possible for to make it generic?

ajainarayanan: Could same be applied to other instances of specific bootstrap classes in this same file?

cdap-ui/app/features/adapters/etlapps.less
```diff
@@ -125,25 +122,317 @@ body.theme-cdap.state-adapters-create {
...
+        }
+      }
+      &[disabled] {
+        .form-control {
+          height: 32px;
+        }
+      }
+    }
+
+    .my-widget-container {
```
ajainarayanan: Shouldn't this be in my-widget-container.less file?

cdap-ui/app/directives/group-side-panel/group-side-panel.less
```diff
@@ -60,17 +135,27 @@ my-side-panel {
         }
       }
 
-
       .item-body {
         padding: 5px;
```
ajainarayanan: Should there be a max-height on the item-body. What happens if more #of plugins (source) gets added in. It should automatically fix a height for the item-body and let it scroll when #of items grows more . Thoughts?

ajainarayanan:
:+1: LGTM

PR #3201, opened 2015-07-16
CDAP-2530 making RecordScannable<StructuredRecord> explorable

albertshau:
There was a bug where Table and ObjectMappedTable are
explorable because they are ```RecordScannable<StructureRecord>```,
but custom datasets were not explorable in the same way. So
somebody could create a custom dataset that uses an explorable
table, but would not be able to use that to make their custom
dataset explorable.

This fix will create a Hive table for custom datasets that are
```RecordScannable<StructuredRecord>```, assuming the schema dataset
property is set. If the schema property is not set, enabling
exploration will fail with a message that the schema property
is required.

### DIFF
--- a/cdap-docs/developers-manual/source/data-exploration/custom-datasets.rst
+++ b/cdap-docs/developers-manual/source/data-exploration/custom-datasets.rst
@@ -132,6 +132,48 @@ The SQL schema of the dataset would be::
 
 Refer to the Hive language manual for more details on schema and data types.
 
+StructuredRecord Type
+---------------------
+There are times when your record type cannot be expressed as a plain old Java object. For example, you may want to write
+a custom dataset whose schema may change depending on the properties it is given. In these situations, you can implement
+a record-scannable dataset that uses ``StructuredRecord`` as the record type::
+
+  class MyStructuredDataset ... implements RecordScannable<StructuredRecord> {
+    ...
+    public Type getRecordType() {
+      return StructuredRecord.class;
+    }
+
+The ``StructuredRecord`` class is essentially a map of fields to values, with a ``Schema`` describing the fields and values::
+
+  public class StructuredRecord {
+    ...
+    public Schema getSchema() { ... }
+
+    public <T> T get(String fieldName) { ... }
+
+Datasets that use ``StructuredRecord`` as the record type must also set the schema dataset property when they are created::
+
+  @Override
+  public void configure() {
+    Schema schema = Schema.recordOf("mySchema",
+      Schema.Field.of("id", Schema.of(Schema.Type.STRING)),
+      Schema.Field.of("name", Schema.of(Schema.Type.STRING)),
+      Schema.Field.of("age", Schema.of(Schema.Type.INT)),
+      Schema.Field.of("email", Schema.of(Schema.Type.STRING))
+    );
+    createDataset("users", MyStructuredDataset.class,
+                  DatasetProperties.builder()
+                    .add(DatasetProperties.SCHEMA, schema.toString())
+                    .build());
+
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by albertshau** [comment]
http://builds.cask.co/browse/CDAP-RBT313

File: cdap-docs/developers-manual/source/data-exploration/custom-datasets.rst [hunk 1]
```diff
@@ -132,6 +132,48 @@ The SQL schema of the dataset would be::
...
+      Schema.Field.of("age", Schema.of(Schema.Type.INT)),
+      Schema.Field.of("email", Schema.of(Schema.Type.STRING))
+    );
+    createDataset("users", MyStructuredDataset.class,
+                  DatasetProperties.builder()
+                    .add(DatasetProperties.SCHEMA, schema.toString())
+                    .build());
+
+Failure to set the schema property will result in errors when enabling exploration on the dataset. The dataset will still
+be created, but it will not be explorable until the schema property is set correctly and another call to enable exploration
```
awholegunch: Is it "another call to enable exploration" or is it "another call to the ``build`` method that enables exploration"? What is the other call that is needed?

awholegunch:
With regard to Docs: a minor comment (more for my edification than the docs); otherwise, that page LGTM.

cdap-docs/developers-manual/source/data-exploration/custom-datasets.rst
anew: I guess what this means is that you make one call to update the dataset properties. And then you make a call to enable explore. But doesn't the update-properties call enable explore already?

cdap-explore/src/test/java/co/cask/cdap/explore/service/HiveExploreStructuredRecordTestRun.java
```diff
@@ -0,0 +1,130 @@
...
+    table.writeEmail("email1", "this is the subject", "this is the body", "sljackson@boss.com");
+
+    Assert.assertTrue(table.commitTx());
+
+    transactionManager.canCommit(tx1, table.getTxChanges());
+    transactionManager.commit(tx1);
+
+    table.postTxCommit();
+
+    Transaction tx2 = transactionManager.startShort(100);
```
anew: what is this for?

anew:
Do you have a test case for a dataset that is RecordWritable<StructuredRecord>?
Do you have a test case for a dataset that does not have the schema property? (A negative test case)

anew:
Code actually looks pretty good. Please see my few comments, @albertshau

cdap-docs/developers-manual/source/data-exploration/custom-datasets.rst
albertshau: Yeah I meant it as a call to enable explore, but Andreas is right, updating properties will already do this.  I will remove that part.

cdap-explore/src/test/java/co/cask/cdap/explore/service/HiveExploreStructuredRecordTestRun.java
albertshau: oops leftover from copy-paste. will remove.

albertshau:
StructuredRecord is currently not supported as a type for RecordWritable. I will make the change in the explore service so that it errors if it sees one and add a test case for it and missing schema

anew:
only a couple more comments

albertshau:
@anew addressed comments, can you take another look?

anew:
please do not squash your commits until the review is done. After squashing it is impossible to see the incremental change since last review.

anew:
LGTM

PR #3202, opened 2015-07-16
CDAP-1580: Auto-balancing logs data processing from Kafka

abaranau:
NOTE: extracted generic ResourceBalancerService to be used by metrics processor and log saver


cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
```diff
@@ -49,46 +47,44 @@
...
 import java.util.Set;
 import java.util.concurrent.Executor;
-import javax.annotation.Nullable;
 
 /**
- * Metrics processor service that processes events from Kafka.
+ * A services that automatically balances resource assignments between its instances.
  */
-public final class KafkaMetricsProcessorService extends AbstractExecutionThreadService {
+public abstract class ResourceBalancerService extends AbstractExecutionThreadService {
```
chtyim: Since this class and sub-classes has nothing to "run", using `AbstractIdleService` instead (so that you don't need a completion and just block on it).

cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
```diff
@@ -49,46 +47,44 @@
...
-  @Inject
-  public KafkaMetricsProcessorService(CConfiguration conf,
-                                      final ZKClientService zkClient,
-                                      DiscoveryService discoveryService,
-                                      final DiscoveryServiceClient discoveryServiceClient,
-                                      KafkaMetricsProcessorServiceFactory metricsProcessorFactory) {
-    this.conf = conf;
-    this.metricsProcessorFactory = metricsProcessorFactory;
+  protected ResourceBalancerService(String serviceName,
+                                 int partitionCount,
```
chtyim: Alignment.

cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
```diff
@@ -49,46 +47,44 @@
...
+                                 final ZKClientService zkClient,
+                                 DiscoveryService discoveryService,
+                                 final DiscoveryServiceClient discoveryServiceClient) {
+    this.serviceName = serviceName;
+    this.partitionCount = partitionCount;
     this.discoveryService = discoveryService;
 
+    this.zk = ZKClients.namespace(zkClient, "/" + serviceName);
```
chtyim: Seems like you can convert this to a local field.

cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
```diff
@@ -49,46 +47,44 @@
...
-  public KafkaMetricsProcessorService(CConfiguration conf,
-                                      final ZKClientService zkClient,
-                                      DiscoveryService discoveryService,
-                                      final DiscoveryServiceClient discoveryServiceClient,
-                                      KafkaMetricsProcessorServiceFactory metricsProcessorFactory) {
-    this.conf = conf;
-    this.metricsProcessorFactory = metricsProcessorFactory;
+  protected ResourceBalancerService(String serviceName,
+                                 int partitionCount,
+                                 final ZKClientService zkClient,
```
chtyim: No need to have final.

chtyim:
Does it mean we can delete the `MultiLeaderElection` class as well? That's not a good class which I love to get rid of it.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/runtime/MetricsProcessorService.java
```diff
@@ -0,0 +1,67 @@
...
+  private static final String SERVICE_NAME = "metrics.processor.kafka.consumer";
+
+  private final KafkaMetricsProcessorServiceFactory factory;
+
+  @Nullable
+  private MetricsContext metricsContext;
+
+  @Inject
+  public MetricsProcessorService(CConfiguration conf,
+                                 final ZKClientService zkClient,
```
chtyim: remove unnecessary final.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/runtime/MetricsProcessorService.java
```diff
@@ -0,0 +1,67 @@
...
+                                 final DiscoveryServiceClient discoveryServiceClient,
+                                 KafkaMetricsProcessorServiceFactory metricsProcessorFactory) {
+    super(SERVICE_NAME,
+          conf.getInt(Constants.Metrics.KAFKA_PARTITION_SIZE, Constants.Metrics.DEFAULT_KAFKA_PARTITION_SIZE),
+          zkClient, discoveryService, discoveryServiceClient);
+
+    this.factory = metricsProcessorFactory;
+  }
+
+  public void setMetricsContext(MetricsContext metricsContext) {
```
chtyim: I don't see this method getting called besides in unit-test. Is it being called in existing code, but because of the refactoring making this class looks like a new class?

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/KafkaLogSaverService.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public final class KafkaLogSaverService extends ResourceBalancerService {
+
+  private static final String SERVICE_NAME = "log.saver.kafka.consumer";
+
+  private final LogSaverFactory logSaverFactory;
+
+  @Inject
+  public KafkaLogSaverService(CConfiguration conf,
+                              final ZKClientService zkClient,
```
chtyim: Remove unnecessary final.

cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
chtyim: You might want to have a javadoc for this constructor, saying that the ZKClient passed in will get further namespaced by the service name. This is to prevent children classes do the unnecessary namespacing by themselves.

chtyim:
In general looks good, just couple comments about the abstract base service class.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/runtime/MetricsProcessorService.java
abaranau: it is. Yeah, it is not new - this is how git diff showed it..

abaranau:
> Does it mean we can delete the MultiLeaderElection class as well? That's not a good class which I love to get rid of it.

cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
abaranau: true

cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
abaranau: yup

cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
abaranau: all true

cdap-common/src/main/java/co/cask/cdap/common/resource/ResourceBalancerService.java
abaranau: yes

cdap-watchdog/src/main/java/co/cask/cdap/metrics/runtime/MetricsProcessorService.java
abaranau: yes

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/KafkaLogSaverService.java
abaranau: yes

abaranau:
http://builds.cask.co/browse/CDAP-DUT2258-3

abaranau:
@chtyim please take another look

chtyim:
LGTM

PR #3203, opened 2015-07-16
(CDAP-3027) Fix the DFSStreamHeartbeatsTest

chtyim:
- It shares variable from multiple threads without using correct memory visibility construct
- It didn’t start dataset service, which causes issue in creating the metrics table, which is needed by the test instance, potentially causing test hang due to infinite retry inside metrics table.


abaranau:
LGTM

abaranau:
:+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT314-3

PR #3204, opened 2015-07-16
[CDAP-2480] Deprecate current workflow /current endpoint in favor of …

bdmogal:
…a new API with /runs/ in the path

Jira: [CDAP-2480](https://issues.cask.co/browse/CDAP-2480)
Build: http://builds.cask.co/browse/CDAP-RBT315

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
@@ -16,6 +16,7 @@
 
 package co.cask.cdap.gateway.handlers;
 
+import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.metrics.MetricStore;
 import co.cask.cdap.api.schedule.SchedulableProgramType;
 import co.cask.cdap.api.schedule.ScheduleSpecification;
@@ -143,13 +144,22 @@ public class WorkflowHttpHandler extends ProgramLifecycleHttpHandler {
     responder.sendString(HttpResponseStatus.OK, "Program run resumed.");
   }
 
+  // TODO: CDAP-2481. Deprecated API. Remove in 3.2.
   @GET
   @Path("/apps/{app-id}/workflows/{workflow-name}/{run-id}/current")
+  public void getWorkflowStatusOld(HttpRequest request, HttpResponder responder,
+                                  @PathParam("namespace-id") String namespaceId,
+                                  @PathParam("app-id") String appId, @PathParam("workflow-name") String workflowName,
+                                  @PathParam("run-id") String runId) {
+    getWorkflowStatus(request, responder, namespaceId, appId, workflowName, runId);
+  }
+
+  @GET
+  @Path("/apps/{app-id}/workflows/{workflow-name}/runs/{run-id}/current")
   public void getWorkflowStatus(HttpRequest request, final HttpResponder responder,
                                 @PathParam("namespace-id") String namespaceId,
                                 @PathParam("app-id") String appId, @PathParam("workflow-name") String workflowName,
                                 @PathParam("run-id") String runId) {
-
     try {
       workflowClient.getWorkflowStatus(namespaceId, appId, workflowName, runId,
                                        new WorkflowClient.Callback() {
@@ -158,8 +168,11 @@ public class WorkflowHttpHandler extends ProgramLifecycleHttpHandler {
                                            if (status.getCode() == WorkflowClient.Status.Code.NOT_FOUND) {
                                              responder.sendStatus(HttpResponseStatus.NOT_FOUND);
                                            } else if (status.getCode() == WorkflowClient.Status.Code.OK) {
+                                             // This uses responder.sendByteArray because status.getResult returns a
+                                             // json string, and responder.sendJson would need deserialization and
+                                             // serialization.
                                              responder.sendByteArray(HttpResponseStatus.OK,
-                                                                     status.getResult().getBytes(),
+                                                                     Bytes.toBytes(status.getResult()),
                                                                      ImmutableMultimap.of(
                                                                        HttpHeaders.Names.CONTENT_TYPE,
                                                                        "application/json; charset=utf-8"));
--- a/cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
+++ b/cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
@@ -619,9 +619,19 @@ For services, you can retrieve:
 
 For workflows, you can retrieve:
 
-- the information about the specific run currently running::
+- the information about the currently running node(s) in the workflow:
 
-    GET <base-url>/namespaces/<namespace>/apps/<app-id>/workflows/<workflow-id>/<run-id>/current
+  .. topic::  **Note: Workflow Current Node(s) RESTful API Deprecated**
+
+    As of *CDAP v3.1.0*, the *Workflow Current Node(s) RESTful API* has been deprecated, pending removal in a later
+    version. Replace all use of this *Workflow Current Node(s) RESTful API*  ::
+
+      GET <base-url>/namespaces/<namespace>/apps/<app-id>/workflows/<workflow-id>/<run-id>/current
+
+    with the revised API shown below for the *currently running node(s) of the workflow*.
+    Note the addition of a ``/runs/`` component in the path::
+
+      GET <base-url>/namespaces/<namespace>/apps/<app-id>/workflows/<workflow-id>/runs/<run-id>/current
 
 - the schedules defined for a workflow (using the parameter ``schedules``)::
 
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
@@ -96,7 +96,8 @@ public class WorkflowHttpHandlerTest  extends AppFabricTestBase {
   }
 
   private Integer runningProgramCount(Id.Program program, String runId) throws Exception {
-    String path = String.format("apps/%s/workflows/%s/%s/current", program.getApplicationId(), program.getId(), runId);
+    String path = String.format("apps/%s/workflows/%s/runs/%s/current", program.getApplicationId(), program.getId(),
+                                runId);
     HttpResponse response = doGet(getVersionedAPIPath(path, program.getNamespaceId()));
     if (response.getStatusLine().getStatusCode() == 200) {
       String json = EntityUtils.toString(response.getEntity());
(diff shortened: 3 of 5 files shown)


sagarkapare:
Do we need this API now? In order to fetch the currently running nodes, UI is querying the RunRecord for the Workflow. The properties field of the Workflow RunRecord contains mapping from `nodeName -> runid of the node`. From node's runid, we can get the status of the the node whether its RUNNING, COMPLETED, or FAILED.

bdmogal:
Discussed this with @anew on https://issues.cask.co/browse/CDAP-2480. UI is not the only client for REST APIs. This is still a useful API to expose to other clients.

sagarkapare:
Ohh ok. sounds good! LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -158,6 +167,9 @@ public void handle(WorkflowClient.Status status) {
                                            if (status.getCode() == WorkflowClient.Status.Code.NOT_FOUND) {
                                              responder.sendStatus(HttpResponseStatus.NOT_FOUND);
                                            } else if (status.getCode() == WorkflowClient.Status.Code.OK) {
+                                             // This uses responder.sendByteArray because status.getResult returns a
+                                             // json string, and responder.sendJson will need deserialization +
```
awholegunch: Perhaps: "would need deserialization and"

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
 
 For workflows, you can retrieve:
 
-- the information about the specific run currently running::
```
awholegunch: I would leave this line in, as even though it is erroneous, as people may be used to it. We can remove it when the depreciation warning is removed.

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
...
+- the information about the specific node currently running::
```
awholegunch: I would remove this line, essentially moving it to the new line 632.

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
...
+
+  .. topic::  **Note: Workflow current node RESTful API Deprecated**
+
+    As of *CDAP v3.1.0*, *Workflow current node RESTful API* has been deprecated, pending removal in a later version.
+    Replace all use of the following *Workflow current node RESTful API* with the new, modified API below (Note the
+    extra ``/runs/`` component in the path).
 
     GET <base-url>/namespaces/<namespace>/apps/<app-id>/workflows/<workflow-id>/<run-id>/current
 
+- the information about the specific node currently running (modified API)::
```
awholegunch:
```
- information about the specific node of the workflow run currently running::
```

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
awholegunch: You need to change the trailing "::" to a single ":"

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
...
 For workflows, you can retrieve:
 
-- the information about the specific run currently running::
+- the information about the specific node currently running::
+
+  .. topic::  **Note: Workflow current node RESTful API Deprecated**
+
+    As of *CDAP v3.1.0*, *Workflow current node RESTful API* has been deprecated, pending removal in a later version.
+    Replace all use of the following *Workflow current node RESTful API* with the new, modified API below (Note the
+    extra ``/runs/`` component in the path).
```
awholegunch: Add "  ::" (two spaces with two colons) to trigger the correct formatting.

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
...
 
     GET <base-url>/namespaces/<namespace>/apps/<app-id>/workflows/<workflow-id>/<run-id>/current
```
awholegunch: Insert:
```
with the revised API shown below for the *specific node of the workflow currently running*. 
Note the addition of a ``/runs/`` component in the path.
```

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
 
 For workflows, you can retrieve:
 
-- the information about the specific run currently running::
+- the information about the specific node currently running::
+
+  .. topic::  **Note: Workflow current node RESTful API Deprecated**
+
+    As of *CDAP v3.1.0*, *Workflow current node RESTful API* has been deprecated, pending removal in a later version.
```
awholegunch:
```
As of *CDAP v3.1.0*, the *Workflow Current Node RESTful API* has been deprecated, 
pending removal in a later version.
Replace all use of this *Workflow Current Node RESTful API*::

File: cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst [hunk 8]
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
 
 For workflows, you can retrieve:
 
-- the information about the specific run currently running::
+- the information about the specific node currently running::
+
+  .. topic::  **Note: Workflow current node RESTful API Deprecated**
```
awholegunch: .. topic::  **Note: Workflow Current Node RESTful API Deprecated**

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
 
 For workflows, you can retrieve:
 
-- the information about the specific run currently running::
+- the information about the specific run currently running:
+
+  .. topic::  **Note: Workflow Current Node RESTful API Deprecated**
+
+    As of *CDAP v3.1.0*, the *Workflow current node RESTful API* has been deprecated, pending removal in a later version.
```
awholegunch: Workflow Current Node RESTful

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
...
+    Replace all use of this *Workflow current node RESTful API*  ::
```
awholegunch: Workflow Current Node RESTful

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
...
 
     GET <base-url>/namespaces/<namespace>/apps/<app-id>/workflows/<workflow-id>/<run-id>/current
```
awholegunch: Add two spaces before GET so that the line is formatted correctly.

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
bdmogal: I think since we have a single bullet now, *run* should change to *node*. What do you think?

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
...
 
+    with the revised API shown below for the *specific node of the workflow currently running*.
+    Note the addition of a ``/runs/`` component in the path.  ::
```
awholegunch: component in the path.  ::
to
component in the path::

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
```diff
@@ -619,10 +619,20 @@ For services, you can retrieve:
...
+
+    GET <base-url>/namespaces/<namespace>/apps/<app-id>/workflows/<workflow-id>/runs/<run-id>/current
```
awholegunch: Add two spaces before the GET.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -158,6 +167,9 @@ public void handle(WorkflowClient.Status status) {
                                            if (status.getCode() == WorkflowClient.Status.Code.NOT_FOUND) {
                                              responder.sendStatus(HttpResponseStatus.NOT_FOUND);
                                            } else if (status.getCode() == WorkflowClient.Status.Code.OK) {
+                                             // This uses responder.sendByteArray because status.getResult returns a
+                                             // json string, and responder.sendJson would need deserialization and
```
anew: you mean it would send it as a String constant, enclosed in "..."?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
```diff
@@ -158,6 +167,9 @@ public void handle(WorkflowClient.Status status) {
...
+                                             // serialization.
                                              responder.sendByteArray(HttpResponseStatus.OK,
                                                                      status.getResult().getBytes(),
```
anew: If getResult() returns a String, then we should not use getBytes() without a Charset. Or use Bytes.toBytes(String)

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
bdmogal: yeah, I saw this unusual usage of ``sendByteArray`` with the ``application/json`` header, so tried to change it to use ``sendJson``. But the usage seems to be to get the response content as a string, and parse it as a json with type ``List<WorkflowActionNode>``. There are a couple of tests that do this. 

``status.getResult`` is a json string, and not an object perhaps because it is populated with the response from the ``/status`` API from the workflow http service. I added a comment so people should not see this as an anomaly and spend time trying to change it to ``responder.sendJson`` in future.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowHttpHandler.java
bdmogal: Sure, will change it.

bdmogal: Fixed.

cdap-docs/reference-manual/source/http-restful-api/lifecycle.rst
anew: My question is: is it always a single current node? Because we have forks in the workflow, there can be more than one node currently executing.

bdmogal: No in fact the output is a ``List<WorkflowActionNode>``, as json. So this should be ``node(s)``

anew:
LGTM

PR #3205, opened 2015-07-16
CDAP-3003: Choose aggregation to query for flow metrics correctly

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-RBT317-1

abaranau:
close to open against release

PR #3206, opened 2015-07-16
CDAP-3003: Choose aggregation to query for flow metrics correctly

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-RBT317-1

abaranau:
ГРЕЕН

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
```diff
@@ -249,10 +267,27 @@ public void add(Collection<? extends MetricValues> metricValues) throws Exceptio
...
   }
 
+  @Nullable
+  private String getAggregation(MetricDataQuery q) {
+    // We mostly rely on auto-selection of aggregation during query (in which case null is returned from
+    // this method). In some specific cases we need to help resolve the aggregation though.
+    Set<String> tagNames = ImmutableSet.<String>builder()
+      .addAll(q.getSliceByTags().keySet()).addAll(q.getGroupByTags()).build();
+    if (tagNames.contains(Constants.Metrics.Tag.FLOW)) {
+      if (tagNames.contains(Constants.Metrics.Tag.PRODUCER) || tagNames.contains(Constants.Metrics.Tag.CONSUMER)) {
```
gsps1: just add a note here, why when we add user metrics for flow, we don't aggregate that in this aggregation group. since the size of this BY_FLOWLET_QUQUE is smaller compared to BY_FLOW, we need to choose appropriately, smth like that.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
```diff
@@ -249,10 +267,27 @@ public void add(Collection<? extends MetricValues> metricValues) throws Exceptio
...
 
   private CubeQuery buildCubeQuery(MetricDataQuery q) {
-    return new CubeQuery(null, q.getStartTs(), q.getEndTs(), q.getResolution(), q.getLimit(), q.getMetrics(),
+    String aggregation = getAggregation(q);
+    return new CubeQuery(aggregation, q.getStartTs(), q.getEndTs(), q.getResolution(), q.getLimit(), q.getMetrics(),
                          q.getSliceByTags(), q.getGroupByTags(), q.getInterpolator());
   }
 
+  @Nullable
+  private String getAggregation(MetricDataQuery q) {
```
gsps1: use variable name `query`

gsps1:
just two minor comments. LGTM otherwise :+1:

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/DefaultMetricStore.java
abaranau: > since the size of this BY_FLOWLET_QUQUE is smaller compared to BY_FLOW

there's no guarantee on how it is chosen, but yes - I'll add a comment

gsps1: there is, right now, we pick the shortest aggregation group that has all the dimensions.

PR #3207, opened 2015-07-16
Introduce a ClassRule, TestConfiguration, for setting CDAP configs in unit-test

chtyim:
It can be used by unit-test classes to alter CDAP configurations easily in unit-test.

- Deprecate the ConfigurableTestBase, since it’s harder to use
- Convert all existing unit-tests to use the TestConfiguration
- Disable explore for tests that doesn’t need explore capability.
  - It cuts ~10-15 seconds from unit-tests that doesn't need explore service


chtyim:
Against wrong branch...

PR #3208, opened 2015-07-16
Introduce a ClassRule, TestConfiguration, that unit-test can use to set CDAP configs

chtyim:
It can be used by unit-test classes to alter CDAP configurations easily in unit-test.

- Deprecate the ConfigurableTestBase, since it’s harder to use
- Convert all existing unit-tests to use the TestConfiguration
- Disable explore for tests that doesn’t need explore capability.
  - It cuts ~10-15 seconds from unit-tests that doesn't need explore service

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BaseETLBatchTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BaseETLBatchTest.java
@@ -41,6 +41,7 @@ import co.cask.cdap.template.etl.transform.StructuredRecordToGenericRecordTransf
 import co.cask.cdap.template.test.sink.MetaKVTableSink;
 import co.cask.cdap.template.test.source.MetaKVTableSource;
 import co.cask.cdap.test.TestBase;
+import co.cask.cdap.test.TestConfiguration;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.gson.Gson;
@@ -54,6 +55,7 @@ import org.apache.hadoop.fs.Path;
 import org.apache.twill.filesystem.Location;
 import org.hsqldb.jdbc.JDBCDriver;
 import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import parquet.avro.AvroParquetOutputFormat;
 import parquet.avro.AvroParquetReader;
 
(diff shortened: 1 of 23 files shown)


cdap-unit-test/src/main/java/co/cask/cdap/test/TestConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+   *
+   * @param configs a Map of configurations
+   */
+  public TestConfiguration(Map<String, String> configs) {
+    this.configs = new HashMap<>(configs);
+  }
+
+  @Override
+  protected void before() throws Throwable {
+    for (Map.Entry<String, String> entry : configs.entrySet()) {
```
abaranau: add a note why do you do that. The same you explained me a minute ago :)

abaranau:
one ask, otherwise LGTM

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestAppWithCube.java
```diff
@@ -51,6 +53,10 @@
  *
  */
 public class TestAppWithCube extends TestBase {
+
+  @ClassRule
+  public static final TestConfiguration CONFIG = new TestConfiguration("explore.enabled", false);
```
anew: we should use constants when they exist. I believe in this case a constant exists. (Constants.Explore.EXPLORE_ENABLED)

chtyim: Fixed. For those tests in cdap-example, I am keeping them as string, since it's not supposed to use CDAP internal class in regular app.

cdap-unit-test/src/main/java/co/cask/cdap/test/TestConfiguration.java
chtyim: Added doc.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BaseETLBatchTest.java
```diff
@@ -64,6 +66,10 @@
  * Base test class that sets up plugins and the batch template.
  */
 public class BaseETLBatchTest extends TestBase {
+
+  @ClassRule
+  public static final TestConfiguration CONFIG = new TestConfiguration("explore.enabled", false);
```
anew: Use the constant?

cdap-app-templates/cdap-etl/cdap-etl-realtime/src/test/java/co/cask/cdap/template/etl/realtime/ETLWorkerTest.java
```diff
@@ -88,6 +89,9 @@
   @ClassRule
   public static final TemporaryFolder TMP_FOLDER = new TemporaryFolder();
 
+  @ClassRule
+  public static final TestConfiguration CONFIG = new TestConfiguration("explore.enabled", false);
```
anew: Use the constant?

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestAppWithCube.java
anew: oh I see. Perhaps that constant should move to the api... but ok for now

anew:
LGTM

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT318-3
Cut ~10mins from build time.
Will squash and merge.

PR #3209, opened 2015-07-16
update hbase-compat modules, add 1.0, 1.0-cdh, remove 0.94

dereklwood:
parcel buildscript needs updated to include the new hbase-compat modules, and drop the 0.94 module.


wolf31o2:
:+1: LGTM

PR #3212, opened 2015-07-16
Documentation for incrementally consuming partitions of a partitioned file set.

anwar6953:
Documentation for incrementally consuming partitions of a partitioned file set.
Additionally, minor updates to existing PartitionedFileSet documentation.

### DIFF
--- a/cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
@@ -349,16 +349,20 @@ Then, in the ``beforeSubmit()`` method of the MapReduce, specify the partitioned
 
   @Override
     public void beforeSubmit(MapReduceContext context) throws Exception {
-      PartitionedFileSet inputRecords = batchPartitionConsumer.getPartitionedFileSet(context, "inputRecords");
+      PartitionedFileSet inputRecords =
+        batchPartitionConsumer.getPartitionedFileSetForConsuming(context, "inputRecords");
       context.setInput("inputRecords", inputRecords);
 
-To save the state of partition processing, call the consumer's ``onFinish()`` method. This ensures that the
+This will read back the previously persisted state, determine the new partitions to read based upon this
+state, and compute a new state to store in memory until a call to ``persist()``. The dataset it returns
+is instantiated with the set of new partitions to read as input.
+To save the state of partition processing, call the consumer's ``persist()`` method. This ensures that the
 next time the MapReduce job runs, it processes only the newly committed partitions::
 
   @Override
   public void onFinish(boolean succeeded, MapReduceContext context) throws Exception {
     if (succeeded) {
-      batchPartitionConsumer.onFinish(context);
+      batchPartitionConsumer.persist(context);
     }
     super.onFinish(succeeded, context);
   }
--- a/cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
@@ -320,12 +320,13 @@ For example, give these arguments when starting the MapReduce through a RESTful 
 Incrementally Processing Partitions of a PartitionedFileSets in MapReduce
 =========================================================================
 
-Another way to process a partitioned file set from a MapReduce is via the beta feature of incrementally
-consuming its partitions. Using this, a MapReduce job can run and process the files of all partitions
-committed to a particular partitioned file set since the last run of the MapReduce job. Currently, the
-MapReduce job utilizing this feature is responsible for storing its own state of processing. The following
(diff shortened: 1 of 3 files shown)


cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
```diff
@@ -317,6 +317,54 @@ For example, give these arguments when starting the MapReduce through a RESTful
     "dataset.totals.output.partition.key.league" : "nfl"
   }
 
+Incrementally Processing Partitions of a PartitionedFileSets in MapReduce
+=========================================================================
+
+Another way to process a partitioned file set from a MapReduce is via the beta feature of incrementally
+consuming its partitions. Using this, a MapReduce job can run and process the files of all partitions
+committed to a particular partitioned file set since the last run of the MapReduce job. Currently, the
```
anew: I think this is phrased with a little bit of assumptions. Perhaps rephrase: 

One way to process a partitioned file set is with a MapReduce program that runs repeatedly and, in every run, reads the partitions that have been added since its previous run. This requires that the MapReduce program persists between runs what partitions have already been consumed.  An easy way to do that is the ``BatchPartitionConsumer``, an experimental feature introduced in CDAP 3.1.0. The MapReduce program is responsible for persisting the state of this consumer. In the following example, the state is persisted to a row in a KeyValue Table, but other types of Datasets can also be used::

cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
```diff
@@ -317,6 +317,54 @@ For example, give these arguments when starting the MapReduce through a RESTful
...
+        return ((KeyValueTable) datasetContext.getDataset("consumingState")).read(STATE_KEY);
+      }
+
+      @Override
+      protected void writeBytes(DatasetContext datasetContext, byte[] stateBytes) {
+        ((KeyValueTable) datasetContext.getDataset("consumingState")).write(STATE_KEY, stateBytes);
+      }
+    };
+
+Then, in the ``beforeSubmit()`` method of the MapReduce, specify the partitioned file set to be used as input::
```
anew: I think it does not become clear that the "getPartitionedFileSet()" does two things (at least):
- read back the previous state
- determine the new partitions to read 
- modify (but not persist) the state 
- instantiate the dataset with the discovered partitions as input. 
I think you can write more prose around that. Also consider renaming the method (which looks like a simple getter) right now

cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
```diff
@@ -317,6 +317,54 @@ For example, give these arguments when starting the MapReduce through a RESTful
...
+
+  @Override
+    public void beforeSubmit(MapReduceContext context) throws Exception {
+      PartitionedFileSet inputRecords = batchPartitionConsumer.getPartitionedFileSet(context, "inputRecords");
+      context.setInput("inputRecords", inputRecords);
+
+To save the state of partition processing call the consumer's ``onFinish()`` method. This ensures that the
```
anew: processing,

cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
```diff
@@ -317,6 +317,54 @@ For example, give these arguments when starting the MapReduce through a RESTful
...
+
+  @Override
+  public void onFinish(boolean succeeded, MapReduceContext context) throws Exception {
+    if (succeeded) {
+      batchPartitionConsumer.onFinish(context);
+    }
+    super.onFinish(succeeded, context);
+  }
+
+A limitation of the current implementation of BatchPartitionConsumer is that there can not be concurrent
```
anew: A limitation of the BatchPartitionConsumer

anew: Actually, if you save the state in beforeSubmit, then the next invocation would do the right thing. So perhaps the method should not be called onFinish().

sreevatsanraman:
This needs to be against release/3.1 branch. Can you create a new PR?

cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
anwar6953: We can call the method `persist()`.

cdap-docs/developers-manual/source/building-blocks/datasets/fileset.rst
anwar6953: Does `getPartitionedFileSetForConsuming()` seem better?
I can put the description in the javadocs of the method as well, to make it clear what it does.

anwar6953:
Reopening against release/3.1:
https://github.com/caskdata/cdap/pull/3228

PR #3214, opened 2015-07-17
CDAP-2763 adding rest apis for adding and getting artifacts

albertshau:
Adding restful endpoints to add artifacts and get metadata
about the artifacts. Also adding a few missing methods to
ArtifactRepository to expose some more get functions from the
ArtifactStore.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
@@ -1,0 +1,422 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.gateway.handlers;
+
+import co.cask.cdap.api.artifact.ArtifactDescriptor;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.templates.plugins.PluginClass;
+import co.cask.cdap.common.conf.CConfiguration;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.http.AbstractBodyConsumer;
+import co.cask.cdap.internal.app.namespace.NamespaceAdmin;
+import co.cask.cdap.internal.app.runtime.artifact.ArtifactAlreadyExistsException;
+import co.cask.cdap.internal.app.runtime.artifact.ArtifactDetail;
+import co.cask.cdap.internal.app.runtime.artifact.ArtifactNotExistsException;
+import co.cask.cdap.internal.app.runtime.artifact.ArtifactRange;
+import co.cask.cdap.internal.app.runtime.artifact.ArtifactRangeNotFoundException;
+import co.cask.cdap.internal.app.runtime.artifact.ArtifactRepository;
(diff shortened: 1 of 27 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2263

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactDetail.java
```diff
@@ -16,6 +16,8 @@
 
 package co.cask.cdap.internal.app.runtime.artifact;
 
+import co.cask.cdap.api.artifact.ArtifactDescriptor;
```
gokulavasan: Just the import? Unused may be?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
```diff
@@ -0,0 +1,422 @@
...
+      return null;
+    }
+  }
+
+  // check that the namespace exists, writing a response if it doesn't and returning null.
+  // also checks if the request is only supposed to include system artifacts, and returning
+  // the system namespace if so.
+  private Id.Namespace validateAndGetNamespace(String namespaceId, boolean isSystem, HttpResponder responder) {
+    Id.Namespace namespace = Id.Namespace.from(namespaceId);
+    if (!namespaceAdmin.hasNamespace(namespace)) {
```
gokulavasan: Will this if statement be false for system namespace?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -53,16 +57,67 @@
...
+    List<ArtifactSummary> summaries = Lists.newArrayList();
+    if (includeSystem) {
+      convertAndAdd(summaries, artifactStore.getArtifacts(Constants.SYSTEM_NAMESPACE_ID));
+    }
+    return Collections.unmodifiableList(convertAndAdd(summaries, artifactStore.getArtifacts(namespace)));
+  }
+
+  /**
+   * Get all artifacts in the given namespace of the given name. Will never return null.
+   * If no artifacts exist, an exception is thrown. Namespace existence is not checked.
```
gokulavasan: Why the difference this method's possible exceptions and the one above?

cdap-proto/src/main/java/co/cask/cdap/proto/artifact/ArtifactInfo.java
```diff
@@ -0,0 +1,88 @@
...
+import co.cask.cdap.api.annotation.Beta;
+import co.cask.cdap.api.artifact.ArtifactClasses;
+
+import java.util.Objects;
+
+/**
+ * Represents an artifact returned by /artifacts/{artifact-name}/versions/{artifact-version}.
+ */
+@Beta
+public class ArtifactInfo {
```
gokulavasan: ArtifactInfo can extend ArtifactSummary?

gokulavasan:
Couple of questions. Rest LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
albertshau: it will be false until we actually create the namespace, which is not in this PR.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
albertshau: this one throws not found because the caller asks for a specific artifact by name, whereas the other just lists all artifacts.  corresponds to GET /artifacts vs GET /artifacts/etlbatch, with the first call never returning a 404 (because the 'artifacts' collection always exists) and the second possibly returning a 404.

cdap-proto/src/main/java/co/cask/cdap/proto/artifact/ArtifactInfo.java
albertshau: good point, will do

albertshau:
@gokulavasan addressed comments, can you please take another look?

gokulavasan:
LGTM :+1:

PR #3215, opened 2015-07-17
CDAP-2986: spark takes dataset arguments when reading writing

rohitsinha54:
- Add new API in SparkContext to take take arguments while reading and writing dataset.
- Unit test for the above

- Fix : Earlier testSparkWithTimePartitionedFileSet was calling testSparkWithPartitionedFileSet() rather than testSparkWithTimePartitionedFileSet()

Build: http://builds.cask.co/browse/CDAP-RBT319-1

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/spark/SparkContext.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/spark/SparkContext.java
@@ -29,6 +29,7 @@ import co.cask.cdap.api.workflow.Workflow;
 import co.cask.cdap.api.workflow.WorkflowToken;
 
 import java.io.Serializable;
+import java.util.Map;
 import javax.annotation.Nullable;
 
 /**
@@ -72,6 +73,19 @@ public interface SparkContext extends RuntimeContext, DatasetContext {
   <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass);
 
   /**
+   * Create a Spark RDD that uses {@link Dataset} instantiated using the provided arguments as an input source
+   *
+   * @param datasetName the name of the {@link Dataset} to be read as an RDD
+   * @param kClass      the key class
+   * @param vClass      the value class
+   * @param datasetArgs arguments for the dataset
+   * @param <T>         type of RDD
+   * @return the RDD created from Dataset
+   * @throws UnsupportedOperationException if the SparkContext is not yet initialized
+   */
+  <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass, Map<String, String> datasetArgs);
+
+  /**
    * Writes a Spark RDD to {@link Dataset}
    *
    * @param rdd         the rdd to be stored
@@ -82,6 +96,19 @@ public interface SparkContext extends RuntimeContext, DatasetContext {
    * @throws UnsupportedOperationException if the SparkContext is not yet initialized
    */
   <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass);
+
+  /**
+   * Writes a Spark RDD to {@link Dataset} instantiated using the provided arguments
+   *
+   * @param rdd         the rdd to be stored
+   * @param datasetName the name of the {@link Dataset} where the RDD should be stored
+   * @param kClass      the key class
+   * @param vClass      the value class
+   * @param datasetArgs arguments for the dataset
+   * @param <T>         type of RDD
+   * @throws UnsupportedOperationException if the SparkContext is not yet initialized
+   */
+  <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass, Map<String, String> datasetArgs);
 
   /**
    * Create a Spark RDD that uses complete {@link Stream} as input source
--- a/cdap-api/src/main/java/co/cask/cdap/api/spark/SparkContext.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/spark/SparkContext.java
@@ -73,7 +73,7 @@ public interface SparkContext extends RuntimeContext, DatasetContext {
   <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass);
 
   /**
-   * Create a Spark RDD that uses {@link Dataset} instantiated using the provided arguments as an input source
+   * Create a Spark RDD that uses {@link Dataset} instantiated using the provided arguments as an input source.
    *
    * @param datasetName the name of the {@link Dataset} to be read as an RDD
    * @param kClass      the key class
@@ -98,7 +98,7 @@ public interface SparkContext extends RuntimeContext, DatasetContext {
   <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass);
 
   /**
-   * Writes a Spark RDD to {@link Dataset} instantiated using the provided arguments
+   * Writes a Spark RDD to {@link Dataset} instantiated using the provided arguments.
    *
    * @param rdd         the rdd to be stored
    * @param datasetName the name of the {@link Dataset} where the RDD should be stored
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ClientSparkContext.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ClientSparkContext.java
@@ -67,18 +67,8 @@ public final class ClientSparkContext extends AbstractSparkContext {
   }
 
   @Override
-  public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass) {
-    return readFromDataset(datasetName, kClass, vClass, Collections.<String, String>emptyMap());
-  }
-
-  @Override
   public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass, Map<String, String> datasetArgs) {
     throw new UnsupportedOperationException("Only supported in SparkProgram.run() execution context");
-  }
-
-  @Override
-  public <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass) {
-    writeToDataset(rdd, datasetName, kClass, vClass, Collections.<String, String>emptyMap());
   }
 
   @Override
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ClientSparkContext.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ClientSparkContext.java
@@ -32,6 +32,7 @@ import org.apache.twill.api.RunId;
 import org.apache.twill.discovery.DiscoveryServiceClient;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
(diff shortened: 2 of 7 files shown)


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkAppUsingFileSet.java
```diff
@@ -105,20 +107,48 @@ public void run(SparkContext context) {
...
-          }
-        });
-
-      // write the character count to dataset
-      context.writeToDataset(stringLengths, output, String.class, Integer.class);
+      String inputPartitionKey = context.getRuntimeArguments().get("inputKey");
+      String outputPartitinKey = context.getRuntimeArguments().get("outputKey");
+
+      // read and write datasets with dataset arguments
+      if (inputPartitionKey != null && outputPartitinKey != null) {
```
rohitsinha54: Is it a good idea to separate this out to another app class ?

cdap-api/src/main/java/co/cask/cdap/api/spark/SparkContext.java
```diff
@@ -72,6 +73,19 @@
   <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass);
 
   /**
+   * Create a Spark RDD that uses {@link Dataset} instantiated using the provided arguments as an input source
```
chtyim: Need a full stop `.` for the first sentence in javadoc.

cdap-api/src/main/java/co/cask/cdap/api/spark/SparkContext.java
```diff
@@ -84,6 +98,19 @@
   <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass);
 
   /**
+   * Writes a Spark RDD to {@link Dataset} instantiated using the provided arguments
```
chtyim: same here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ClientSparkContext.java
```diff
@@ -67,11 +68,22 @@ public ClientSparkContext(Program program, RunId runId, long logicalStartTime, M
...
+  }
+
+  @Override
+  public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass, Map<String, String> datasetArgs) {
     throw new UnsupportedOperationException("Only supported in SparkProgram.run() execution context");
   }
 
   @Override
   public <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass) {
+    writeToDataset(rdd, datasetName, kClass, vClass, Collections.<String, String>emptyMap());
```
chtyim: Move this implementation to `AbstractSparkContext` so that you don't need to do the same delegation in here and in `ExecutionSparkContext`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ClientSparkContext.java
```diff
@@ -67,11 +68,22 @@ public ClientSparkContext(Program program, RunId runId, long logicalStartTime, M
 
   @Override
   public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass) {
+    return readFromDataset(datasetName, kClass, vClass, Collections.<String, String>emptyMap());
```
chtyim: Move this to `AbstractSparkContext`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -122,12 +123,19 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
...
+
+  @Override
+  public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass, Map<String, String> userDsArgs) {
     // Clone the configuration since it's dataset specification and shouldn't affect the global hConf
     Configuration configuration = new Configuration(hConf);
 
     // first try if it is InputFormatProvider
-    Map<String, String> dsArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
-    Dataset dataset = instantiateDataset(datasetName, dsArgs);
+    Map<String, String> runtimeDsArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName,
```
chtyim: no need to rename to `runtimeDsArgs`. Just call it `dsArgs`.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkAppUsingFileSet.java
```diff
@@ -105,20 +107,48 @@ public void run(SparkContext context) {
...
-          @Override
-          public Tuple2<String, Integer> call(Tuple2<Long, String> pair) throws Exception {
-            return new Tuple2<>(pair._2(), pair._2().length());
-          }
-        });
-
-      // write the character count to dataset
-      context.writeToDataset(stringLengths, output, String.class, Integer.class);
+      String inputPartitionKey = context.getRuntimeArguments().get("inputKey");
+      String outputPartitinKey = context.getRuntimeArguments().get("outputKey");
```
chtyim: outputPartitionKey

chtyim: Also, from the code below, it seems like it is not representing a key, but rather outputPartitionTime?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkAppUsingFileSet.java
```diff
@@ -105,20 +107,48 @@ public void run(SparkContext context) {
...
+          new PairFunction<Tuple2<Long, String>, String, Integer>() {
+            @Override
+            public Tuple2<String, Integer> call(Tuple2<Long, String> pair) throws Exception {
+              return new Tuple2<>(pair._2(), pair._2().length());
+            }
+          });
+
+        // write the character count to dataset with user custom dataset args
+        Map<String, String> outputArgs = new HashMap<>();
+        TimePartitionedFileSetArguments.setOutputPartitionTime(outputArgs, Long.parseLong(outputPartitinKey));
```
chtyim: So this is for not requiring the user to provide a runtime argument with the name needed by TimePartitionedFileSet? It seems a bit redundant to me (the user provides `outputKey` instead of `output.partition.time`, which to me the latter is clearer).

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkAppUsingFileSet.java
```diff
@@ -105,20 +107,48 @@ public void run(SparkContext context) {
...
+        context.writeToDataset(customPartitionStringLengths, output, String.class, Integer.class, outputArgs);
+      } else {
+        // read the dataset
+        JavaPairRDD<Long, String> inputData = context.readFromDataset(input, Long.class, String.class);
```
chtyim: Why repeat the code. You can have the `if` block constructs the dataset arguments map and always call the read/write method with the input/output dataset arguments.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
```diff
@@ -306,6 +297,61 @@ public void apply() throws Exception {
   }
 
   @Test
+  public void testSparkWithTimePartitionedFileSetAndDSArgs() throws Exception {
+    testSparkWithTimePartitionedFileSetAndDSArgs(SparkAppUsingFileSet.class, SparkAppUsingFileSet.JavaCharCount.class);
```
chtyim: Instead of creating two new tests, can you modify the app instead so that it do two rounds of read/write, such that one round with dataset arguments and one without?

Adding new test methods for this case is not necessary and it gives a higher cost (hence longer run time) for running unit-tests.

cdap-app-fabric/src/test/scala/co/cask/cdap/internal/app/runtime/spark/ScalaFileCountProgram.scala
```diff
@@ -29,13 +32,36 @@ class ScalaFileCountProgram extends ScalaSparkProgram {
...
+
+      // read the dataset with custom user dataset arguments
+      val customInputData: NewHadoopRDD[Long, String] = sc.readFromDataset(input, classOf[Long], classOf[String], inputArgs)
+
+      // create a new RDD with the same key but the value is the length of the string
+      val customPartitionStringLengths = customInputData.map(str => (str._2, str._2.length))
+
+
+      // write the character count to dataset with user custom dataset args
+      val outputArgs: util.HashMap[String, String] = new util.HashMap[String, String]
```
chtyim: Similar comments to the Java version applies here as well.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
rohitsinha54: I had it that way earlier but then changed it as this one looked cleaner to me since it separates outt the test cases. But yes you are right the trade off longer run time is way too high.

rohitsinha54:
New Build: http://builds.cask.co/browse/CDAP-RBT319-4

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -121,12 +122,14 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
...
   @Override
-  public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass) {
+  public <T> T readFromDataset(String datasetName, Class<?> kClass, Class<?> vClass, Map<String, String> userDsArgs) {
     // Clone the configuration since it's dataset specification and shouldn't affect the global hConf
     Configuration configuration = new Configuration(hConf);
 
     // first try if it is InputFormatProvider
-    Map<String, String> dsArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
+    Map<String, String> dsArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName,
+                                                                      getRuntimeArguments());
```
chtyim: Why the misaligned newline change?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -168,12 +171,15 @@ public ExecutionSparkContext(SparkSpecification specification, Id.Program progra
...
-  public <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass) {
+  public <T> void writeToDataset(T rdd, String datasetName, Class<?> kClass, Class<?> vClass,
+                                 Map<String, String> userDsArgs) {
     // Clone the configuration since it's dataset specification and shouldn't affect the global hConf
     Configuration configuration = new Configuration(hConf);
 
     // first try if it is OutputFormatProvider
-    Map<String, String> dsArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName, getRuntimeArguments());
+    Map<String, String> dsArgs = RuntimeArguments.extractScope(Scope.DATASET, datasetName,
+                                                                      getRuntimeArguments());
```
chtyim: Same here.

chtyim:
Just two minor comments about unnecessary formatting changes. Rest LGTM.

PR #3216, opened 2015-07-17
(CDAP-3028) Improve the system service restart endpoints to handle invalid instance id and serbive not available/ready.

henry-cask:
Add different HTTP status for restart system services endpoint to indicate service disabled/not ready and illegal instance id to be restarted.


henry-cask:
Wrong branch

PR #3217, opened 2015-07-17
(CDAP-3028) Improve the system service restart endpoints to handle invalid instance id and serbive not available/ready

henry-cask:
Add different HTTP status for service disabled/not ready and illegal instance id to be restarted.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
@@ -239,14 +239,14 @@ public class AbstractMonitorHandler extends AbstractAppFabricHttpHandler {
       }
       responder.sendStatus(HttpResponseStatus.OK);
     } catch (IllegalStateException ise) {
-      LOG.warn(String.format("IllegalStateException when trying to restart instances for service %s", serviceName));
+      LOG.debug(String.format("IllegalStateException when trying to restart instances for service %s", serviceName));
 
       isSuccess = false;
       responder.sendString(HttpResponseStatus.FORBIDDEN,
                            String.format("Fail to restart instance for % because the service may not be enabled or " +
                                            "not ready yet", serviceName));
     } catch (IllegalArgumentException iex) {
-      LOG.warn(String.format("IllegalArgumentException when trying to restart instances for service %s", serviceName),
+      LOG.debug(String.format("IllegalArgumentException when trying to restart instances for service %s", serviceName),
                iex);
 
       isSuccess = false;
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
@@ -207,31 +207,52 @@ public class AbstractMonitorHandler extends AbstractAppFabricHttpHandler {
   }
 
   public void restartAllServiceInstances(HttpRequest request, HttpResponder responder, String serviceName) {
-    restartInstances(responder, serviceName, -1);
+    restartInstances(responder, serviceName, -1, true);
   }
 
   public void restartServiceInstance(HttpRequest request, HttpResponder responder, String serviceName,
                                      int instanceId) {
-    restartInstances(responder, serviceName, instanceId);
+    restartInstances(responder, serviceName, instanceId, false);
   }
 
-  private void restartInstances(HttpResponder responder, String serviceName, int instanceId) {
+  private void restartInstances(HttpResponder responder, String serviceName, int instanceId, boolean restartAll) {
     long startTimeMs = System.currentTimeMillis();
     boolean isSuccess = true;
+    if (!serviceManagementMap.containsKey(serviceName)) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Invalid service name %s", serviceName));
+      return;
+    }
+    MasterServiceManager masterServiceManager = serviceManagementMap.get(serviceName);
+
     try {
-      if (!serviceManagementMap.containsKey(serviceName)) {
-        responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Invalid service name %s", serviceName));
-        return;
+      if (!masterServiceManager.isServiceEnabled()) {
+        throw new IllegalStateException();
       }
 
-      MasterServiceManager masterServiceManager = serviceManagementMap.get(serviceName);
-      if (instanceId < 0) {
+      if (restartAll) {
         masterServiceManager.restartAllInstances();
       } else {
+        if (instanceId < 0 || instanceId >= masterServiceManager.getInstances()) {
+          throw new IllegalArgumentException();
+        }
         masterServiceManager.restartInstances(instanceId);
       }
-
       responder.sendStatus(HttpResponseStatus.OK);
+    } catch (IllegalStateException ise) {
+      LOG.warn(String.format("IllegalStateException when trying to restart instances for service %s", serviceName));
+
+      isSuccess = false;
+      responder.sendString(HttpResponseStatus.FORBIDDEN,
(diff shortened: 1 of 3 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -207,31 +207,52 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
-    restartInstances(responder, serviceName, instanceId);
+    restartInstances(responder, serviceName, instanceId, false);
   }
 
-  private void restartInstances(HttpResponder responder, String serviceName, int instanceId) {
+  private void restartInstances(HttpResponder responder, String serviceName, int instanceId, boolean restartAll) {
     long startTimeMs = System.currentTimeMillis();
     boolean isSuccess = true;
+    if (!serviceManagementMap.containsKey(serviceName)) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Invalid service name %s", serviceName));
```
bdmogal: I think the new way of doing this is to throw a ``NotFoundException``. The ``netty-http`` framework can translate that into a 404

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -207,31 +207,52 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+        throw new IllegalStateException();
       }
 
-      MasterServiceManager masterServiceManager = serviceManagementMap.get(serviceName);
-      if (instanceId < 0) {
+      if (restartAll) {
         masterServiceManager.restartAllInstances();
       } else {
+        if (instanceId < 0 || instanceId >= masterServiceManager.getInstances()) {
+          throw new IllegalArgumentException();
```
bdmogal: Instead, you can throw a ``BadRequestException``, and it will translate into 404. Check ``HttpExceptionHandler`` to see how these exceptions translate into responses.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -207,31 +207,52 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
+      return;
+    }
+    MasterServiceManager masterServiceManager = serviceManagementMap.get(serviceName);
+
     try {
-      if (!serviceManagementMap.containsKey(serviceName)) {
-        responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Invalid service name %s", serviceName));
-        return;
+      if (!masterServiceManager.isServiceEnabled()) {
+        throw new IllegalStateException();
```
bdmogal: Should this translate into 403 or 409? If it should be 409, then you can just throw a ``ConflictException`` from here. If 403, then you may want to add a ``ForbiddenException`` and a clause for it in ``HttpExceptionHandler`` (its missing right now).

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -207,31 +207,52 @@ private int getSystemServiceInstanceCount(String serviceName) throws Exception {
...
       responder.sendStatus(HttpResponseStatus.OK);
+    } catch (IllegalStateException ise) {
+      LOG.warn(String.format("IllegalStateException when trying to restart instances for service %s", serviceName));
+
+      isSuccess = false;
+      responder.sendString(HttpResponseStatus.FORBIDDEN,
+                           String.format("Fail to restart instance for % because the service may not be enabled or " +
+                                           "not ready yet", serviceName));
+    } catch (IllegalArgumentException iex) {
+      LOG.warn(String.format("IllegalArgumentException when trying to restart instances for service %s", serviceName),
```
bdmogal: lets make this debug, since its a user error

bdmogal: although you may not need this if you use exceptions. So ignore.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
hsaputra: I am following pattern in this handler class. I think we should send new PR to revamp the whole file with the new pattern.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
hsaputra: I a following same pattern in this handler class so lets change to 409 if needed for the whole cases

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
hsaputra: Previous review ask to add war log so lets keep it for now.

hsaputra:
@bdmogal, thanks for the review but lets move changing to new exception pattern in new PR since this class still setting to the status to the responder.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: I am following pattern in this handler class. I think we should send new PR to revamp the whole file with the new pattern.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: I am following same pattern in this handler class so lets change to 409 if needed for the whole cases

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: Previous review ask to add warn log so lets keep it for now.

henry-cask:
@bdmogal, thanks for the review but lets move changing to new exception pattern in new PR since this class still setting to the status to the responder.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: Actually, you are right lets change it as debug since there was no logging before and not useful for admin.

henry-cask:
Build:
https://builds.cask.co/browse/CDAP-RBT316-5

henry-cask:
Filed JIRA to address updating the AbstractMonitorHandler to use latest mechanism (via Exception throws) to handle non OK HTTP status code:
https://issues.cask.co/browse/CDAP-3034

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: I actually throw exception here to have unified error handle in the catch clause for invalid service.

henry-cask:
@bdmogal I made the recommended change to lower logs.
But would still address new exception handler with new JIRA filed above.

bdmogal:
I think this can be refactored a lot to make it simpler/more readable, using the new way of handling non-OK responses (throwing exception). Given that you have a jira for that, LGTM.

henry-cask:
Build pass: http://builds.cask.co/browse/CDAP-RBT316-6

PR #3218, opened 2015-07-17
fixing explore classpath issues 

gsps1:
1) editing yarn_application_classpath to include container working dir jars before job.jar for map-reduce job containers launched by explore queries. and changing appropriate parameters to use YARN_APP_CLASSPATH as classpath before job.jar
2) excluding hive-exec-*.jar from the container working directory

JIRA : https://issues.cask.co/browse/CDAP-2991

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
@@ -55,11 +55,13 @@ import com.google.inject.util.Modules;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.mapreduce.MRConfig;
+import org.apache.twill.api.ClassAcceptor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URL;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


gsps1:
build running : http://builds.cask.co/browse/CDAP-RBT321-1

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
```diff
@@ -332,15 +346,65 @@ public boolean accept(String className, URL classUrl, URL classPathUrl) {
...
+      throw Throwables.propagate(e);
+    }
+
+    try {
+      conf.writeXml(fos);
+    } catch (IOException e) {
+      LOG.error("Could not write modified configuration to temporary yarn-site.xml at {}", newYarnConfFile, e);
+      throw Throwables.propagate(e);
+    } finally {
+      Closeables.closeQuietly(fos);
```
chtyim: Shouldn't do close quietly. It there is exception during close, meaning the file content may not be right. Also, I dont't think we need to distinguish file not found vs io exception for the logging purpose since we are logging the exception already. BTW, use the new ARM syntax:

```java
try {
  try (OutputStream os = new FileOutputStream(newYarnConfFile)) {
    conf.writeXml(os);
  }
} catch (IOException e) {
  LOG.error(...);
}
```

cdap-explore/src/test/java/co/cask/cdap/explore/service/ExploreServiceUtilsTest.java
```diff
@@ -54,25 +56,41 @@ public void hijackHiveConfFileTest() throws Exception {
...
+    Assert.assertEquals("false", conf.get(Job.MAPREDUCE_JOB_USER_CLASSPATH_FIRST));
+    Assert.assertEquals("false", conf.get(Job.MAPREDUCE_JOB_CLASSLOADER));
     Assert.assertEquals("bar", conf.get("foo"));
 
-    // Ensure conf files that are not hive-site.xml are unchanged
+    // check yarn-site changes
     confFile = tmpFolder.newFile("yarn-site.xml");
-    Assert.assertEquals(confFile, ExploreServiceUtils.hijackHiveConfFile(confFile));
+    conf = new YarnConfiguration();
+    fos = new FileOutputStream(confFile);
```
chtyim: Use the new ARM syntax.

cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
```diff
@@ -256,21 +258,39 @@ private static void setupClasspath() throws IOException {
...
+        * We need to remove Kryo dependency in the Explore container. Spark introduced version 2.21 version of Kryo,
+        * which would be normally shipped to the Explore container. Yet, Hive requires Kryo 2.22,
+        * and gets it from the Hive jars - hive-exec.jar to be precise.
+        * hive-exec is the job.jar in hive, since hive has guava-11.0.2 classes
+        * we want to exclude the hive-exec.jar from classpath.
+        * */
+      @Override
+      public boolean accept(String className, URL classUrl, URL classPathUrl) {
+        if (bootstrapClassPaths.contains(classPathUrl.getFile()) ||
+          className.startsWith("com.esotericsoftware.kryo") || classPathUrl.getFile().contains("hive-exec")) {
```
chtyim: In distributed mode, actually we can exclude all transitive dependencies that brought in by Spark. We don't have to change the code node, but can you open a JIRA for that?

chtyim:
LGTM. Please open a JIRA about removing Spark dependencies.

cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
gsps1: added jira here : https://issues.cask.co/browse/CDAP-3035

PR #3219, opened 2015-07-17
Fix documentation build

awholegunch:
This attempts to fix the documentation build by improving the determination of the parent branch.

However, the real problem with the documentation build is that there is no 3.1-compatible branch in the CDAP-Apps repository.(This has been addressed directly in the CDAP-Apps repo.)

It moves an echo statement from a method into the place where the method is called, cleaning up the method.

This also fixes https://issues.cask.co/browse/CDAP-3072, a documentation Javascript error.


sreevatsanraman:
LGTM

PR #3224, opened 2015-07-17
make zookeeper jar as provided, as its included in classpath from hba…

gsps1:
…se classpath


gsps1:
build running : http://builds.cask.co/browse/CDAP-RBT322-1

ghelmling:
Hmm, better test that this still works for standalone as well.

gsps1:
tested on standalone as well.

ghelmling:
lgtm, assuming test build passes and testing on a cluster works as well.

PR #3227, opened 2015-07-21
CDAP-3058 NamespacedLocationFactory was not appending correctly

albertshau:
The method to get a subdirectory of a namespace location was
returning the namespace location instead of the subdirectory.
This is really bad if somebody decides to delete that
'subdirectory', then ends up deleting the entire namespace.


bdmogal:
ugh! LGTM :+1:  

Thanks for fixing.

albertshau:
http://builds.cask.co/browse/CDAP-RBT323

poornachandra:
It would be a good idea to add a unit test for this.

albertshau:
ok will add tests

anew:
Thanks for mentioning that, Poorna! 
A bug fix should always be accompanied by a unit test for that bug :)

albertshau:
@poornachandra added a unit test, can you take a look?

poornachandra:
Thanks Albert! LGTM

PR #3230, opened 2015-07-21
Spark improvements and fixes

chtyim:
- Spark should be available from the cluster, with $SPARK_HOME pointing to it
- For standalone, we package with spark-core and spark-sql (we don’t support spark-streaming yet)
- Also fix an annoying error message in standalone by doing proper cleanup of System properties that are set by spark-submit.


chtyim:
Verified on standalone and cluster that it can run spark apps (with spark SQL as well) correctly.

abaranau:
LGTM

chtyim:
Test passed: https://builds.cask.co/browse/CDAP-RBT331-4

PR #3232, opened 2015-07-21
[CDAP-3023] fix flowlet instance default in diagram to 1

elfenheart:
screen shot 2015-07-21 at 2 24 33 pm


ajainarayanan:
:+1: LGTM

PR #3233, opened 2015-07-21
CDAP-3040: fix metrics server reporting notok

abaranau:
the problem was in using same discoverable name for kafka consumer service and status service


abaranau:
http://builds.cask.co/browse/CDAP-RBT328-1

gsps1:
LGTM

PR #3234, opened 2015-07-21
Add new HTTP status for system service instances restart to separate between 503 Service unavailable and 403 Service is disabled.

henry-cask:
Following basic HTTP status rule of thumb: 
5XX code is our fault, and 4XX code is caller fault.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
@@ -226,12 +226,12 @@ public class AbstractMonitorHandler extends AbstractAppFabricHttpHandler {
 
     try {
       if (!masterServiceManager.isServiceEnabled()) {
-        LOG.debug(String.format("Fail restart instances because service %s is not enabled.", serviceName));
+        String message = String.format("Failed to restart instance for % because the service is not enabled.",
+                                       serviceName);
+        LOG.debug(message);
 
         isSuccess = false;
-        responder.sendString(HttpResponseStatus.FORBIDDEN,
-                             String.format("Fail to restart instance for % because the service is not enabled.",
-                                           serviceName));
+        responder.sendString(HttpResponseStatus.FORBIDDEN, message);
         return;
       }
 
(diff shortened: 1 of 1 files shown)


henry-cask:
Build: http://builds.cask.co/browse/CDAP-RBT325-3

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -242,9 +248,9 @@ private void restartInstances(HttpResponder responder, String serviceName, int i
       LOG.debug(String.format("IllegalStateException when trying to restart instances for service %s", serviceName));
```
bdmogal: Better to have twill throw a better, checked exception, so you don't have to remember to catch an ``IllegalStateException`` everywhere. Can you file a jira?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -226,7 +226,13 @@ private void restartInstances(HttpResponder responder, String serviceName, int i
 
     try {
       if (!masterServiceManager.isServiceEnabled()) {
-        throw new IllegalStateException();
+        LOG.debug(String.format("Fail restart instances because service %s is not enabled.", serviceName));
+
+        isSuccess = false;
+        responder.sendString(HttpResponseStatus.FORBIDDEN,
+                             String.format("Fail to restart instance for % because the service is not enabled.",
+                                           serviceName));
```
awholegunch: Why are these messages not the same? I'd go with this instead for each:
```
"Failed to restart instance for % because the service is not enabled.", serviceName
```

bdmogal:
One comment to open a Jira and TODO. :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
awholegunch: This says "instances", but the next message says "instance"..should these be consistent?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
```diff
@@ -242,9 +248,9 @@ private void restartInstances(HttpResponder responder, String serviceName, int i
       LOG.debug(String.format("IllegalStateException when trying to restart instances for service %s", serviceName));
 
       isSuccess = false;
-      responder.sendString(HttpResponseStatus.FORBIDDEN,
-                           String.format("Fail to restart instance for % because the service may not be enabled or " +
-                                           "not ready yet", serviceName));
+      responder.sendString(HttpResponseStatus.SERVICE_UNAVAILABLE,
+                           String.format("Fail to restart instance for % because the service may not be ready yet",
```
awholegunch: "Failed to restart instance for..."

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: Filed:
https://issues.apache.org/jira/browse/TWILL-143

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
awholegunch: Plus, "instances" in one and "instance" in the other...

henry-cask: Will change the text to be the same

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AbstractMonitorHandler.java
henry-cask: Will update the message

henry-cask:
Pending build: http://builds.cask.co/browse/CDAP-RBT325-6

PR #3236, opened 2015-07-21
[CDAP-2870] Feature/ui workflow diagram cursor fixes

elfenheart:
- Fix cursor on nodes that have not run yet (not-allowed)
- add status color to Killed status (red border)
- Making WorkFlow resources to be consistent with other program types (adding namespace to params). This was causing issue when you switch namespace, the program was still trying to fetch information from the previous namespace for some reason, thus throwing 404.

Cluster: singlenode2012-1000


ajainarayanan:
:+1: LGTM

PR #3239, opened 2015-07-21
reset flowlet metrics data on scope destroy

elfenheart:
Use case: When there are 2 namespaces with identical programs, run a flow and it will generate metrics. When you switch to a different namespace and go to the identical flow, the metrics from the other namespace will show up. It disappears on refresh.

Solution: Adding a reset function on FlowDiagramData service which is called on $scope $destroy event of the status-ctrl.


ajainarayanan:
:+1: LGTM

PR #3240, opened 2015-07-22
service endpoints include only public methods

anwar6953:
The first commit is the bugfix and the second commit is miscellaneous other cleanup within the test class.

https://issues.cask.co/browse/CDAP-3044
http://builds.cask.co/browse/CDAP-RBT327-1


anwar6953:
Reopened against release/3.1:
https://github.com/caskdata/cdap/pull/3241

PR #3241, opened 2015-07-22
Bugfix/service endpoints include only public methods

anwar6953:
The first commit is the bugfix and the second commit is miscellaneous other cleanup within the test class.

https://issues.cask.co/browse/CDAP-3044
http://builds.cask.co/browse/CDAP-RBT327-4

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
@@ -24,6 +24,9 @@ import co.cask.cdap.DummyAppWithTrackingTable;
 import co.cask.cdap.SleepingWorkflowApp;
 import co.cask.cdap.WordCountApp;
 import co.cask.cdap.api.schedule.ScheduleSpecification;
+import co.cask.cdap.api.service.ServiceSpecification;
+import co.cask.cdap.api.service.http.HttpServiceHandlerSpecification;
+import co.cask.cdap.api.service.http.ServiceHttpEndpoint;
 import co.cask.cdap.api.workflow.WorkflowActionSpecification;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.queue.QueueName;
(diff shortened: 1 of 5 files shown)


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -529,6 +533,32 @@ public void testWorkerSpecification() throws Exception {
...
+                                                                 new ServiceHttpEndpoint("GET", "/multi"),
+                                                                 new ServiceHttpEndpoint("GET", "/multi/ping"));
+
+    GsonBuilder gsonBuidler = new GsonBuilder();
+    gsonBuidler.registerTypeAdapter(ServiceSpecification.class, new ServiceSpecificationCodec());
+    gsonBuidler.registerTypeAdapter(HttpServiceHandlerSpecification.class, new HttpServiceSpecificationCodec());
+    Gson gson = gsonBuidler.create();
+    ServiceSpecification specification = readResponse(response, ServiceSpecification.class, gson);
+
+    Set<ServiceHttpEndpoint> returnedEndpoints = Sets.newHashSet();
```
bdmogal: use Java 7 syntax?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -851,9 +877,8 @@ private HttpResponse sendLiveInfoRequest(String namespace, String appId, String
     return doGet(versionedUrl);
   }
 
-  private int changeFLowletStreamConnection(String namespace, String appId, String flow, String flowlet,
-                                            String oldStreamId, String newStreamId)
-    throws Exception {
+  private int changeFlowletStreamConnection(String namespace, String appId, String flow, String flowlet,
```
bdmogal: should we have this method accept ``Id.Flow`` too?

bdmogal:
Nice cleanup! just a couple of minor comments, otherwise :+1: !

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
anwar6953: this method wasn't being used, so I just left it as is

anwar6953:
@bdmogal Addressed comments.

bdmogal:
:+1:

PR #3243, opened 2015-07-22
CDAP-2999 an improper if condition resulted in in-progress mappers to…

anwar6953:
CDAP-2999 an improper if condition resulted in in-progress mappers to show a duration of 17 days

https://issues.cask.co/browse/CDAP-2999


ajainarayanan:
:+1: LGTM

PR #3244, opened 2015-07-22
Issue 3015: Add Workflow Token documentation

awholegunch:
Documentation for Workflow Tokens

The RESTful API examples need updated samples of the returned tokens.

- [Developers' Manual: Workflow](http://docs-staging.cask.co/cdap/3.1.0-SNAPSHOT-feature-r3.1_3015/en/developers-manual/building-blocks/workflows.html)
- [Ref Manual: Workflow RESTful API](http://docs-staging.cask.co/cdap/3.1.0-SNAPSHOT-feature-r3.1_3015/en/reference-manual/http-restful-api/workflow.html)

### DIFF
--- a/cdap-docs/developers-manual/build.sh
+++ b/cdap-docs/developers-manual/build.sh
@@ -86,6 +86,9 @@ function download_includes() {
 #   download_readme_file_and_test ${includes_dir} ${ingest_url} 277ded1924cb8d9b52a007f262820002 cdap-stream-clients/javascript
   download_readme_file_and_test ${includes_dir} ${ingest_url} b6dedd629c708dbc68bc918b768edda5 cdap-stream-clients/python
   download_readme_file_and_test ${includes_dir} ${ingest_url} 5fc88ec3a658062775403f5be30afbe9 cdap-stream-clients/ruby
+
+  echo_red_bold "Check included example files for changes"
+  test_an_include 55738256b6c668914e0dde5c0ec44bd5 ../../cdap-examples/WikipediaPipeline/src/main/scala/co/cask/cdap/examples/wikipedia/ScalaSparkLDA.scala
 }
 
 function test_includes () {
--- a/cdap-docs/developers-manual/build.sh
+++ b/cdap-docs/developers-manual/build.sh
@@ -88,7 +88,8 @@ function download_includes() {
   download_readme_file_and_test ${includes_dir} ${ingest_url} 5fc88ec3a658062775403f5be30afbe9 cdap-stream-clients/ruby
 
   echo_red_bold "Check included example files for changes"
-  test_an_include 55738256b6c668914e0dde5c0ec44bd5 ../../cdap-examples/WikipediaPipeline/src/main/scala/co/cask/cdap/examples/wikipedia/ScalaSparkLDA.scala
+#   test_an_include 3b877d261bf74f7521a14369d186a21b ../../cdap-examples/WikipediaPipeline/src/main/scala/co/cask/cdap/examples/wikipedia/ScalaSparkLDA.scala
+#   test_an_include 3b877d261bf74f7521a14369d186a21b ../../cdap-examples/WikipediaPipeline/src/main/scala/co/cask/cdap/examples/wikipedia/TopNMapReduce.java
 }
 
 function test_includes () {
--- a/cdap-docs/developers-manual/source/building-blocks/index.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/index.rst
@@ -44,8 +44,8 @@ write data through the data abstraction layer in CDAP.
 - :doc:`MapReduce Programs <mapreduce-programs>`
 - :doc:`Workflows <workflows>`
 - :doc:`Schedules <schedules>`
-- :doc:`Workers <workers>`
 - :doc:`Spark Programs <spark-programs>`
+- :doc:`Workers <workers>`
 - :doc:`Services <services>`
 
 A :doc:`Namespace <namespaces>` is a physical grouping of application and data in CDAP.
--- a/cdap-docs/developers-manual/source/building-blocks/index.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/index.rst
@@ -19,11 +19,11 @@ Building Blocks
     Datasets <datasets/index>
     Flows and Flowlets <flows-flowlets/index>
     MapReduce Programs<mapreduce-programs>
-    Workflows <workflows>
     Schedules <schedules>
+    Services <services>
     Spark Programs <spark-programs>
     Workers <workers>
-    Services <services>
+    Workflows <workflows>
     Namespaces <namespaces>
     Transaction System <transaction-system>
 
@@ -42,11 +42,11 @@ write data through the data abstraction layer in CDAP.
 
 - :doc:`Flows (and flowlets) <flows-flowlets/index>`
 - :doc:`MapReduce Programs <mapreduce-programs>`
-- :doc:`Workflows <workflows>`
 - :doc:`Schedules <schedules>`
+- :doc:`Services <services>`
 - :doc:`Spark Programs <spark-programs>`
 - :doc:`Workers <workers>`
-- :doc:`Services <services>`
+- :doc:`Workflows <workflows>`
 
 A :doc:`Namespace <namespaces>` is a physical grouping of application and data in CDAP.
 Conceptually, namespaces can be thought of as a partitioning of a CDAP instance.
--- a/cdap-docs/developers-manual/source/building-blocks/workflows.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/workflows.rst
@@ -287,8 +287,8 @@ In this code sample, we show how to update the WorkflowToken in a MapReduce prog
 **You will get an exception** if you try to update the workflow token in:
 
 - map or reduce methods;
-- Executors in Spark programs;
-- ``run`` method in A custom actions.
+- Executors in Spark programs; or
+- ``run`` method in custom actions.
 
 You can always read the workflow token in any of the above situations. The :ref:`Wikipedia
 Pipeline example <examples-wikipedia-data-pipeline>` demonstrates some of these techniques.
--- a/cdap-docs/developers-manual/source/building-blocks/workflows.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/workflows.rst
@@ -189,12 +189,8 @@ passed, available to each of the programs in the workflow. This allows programs 
 - alter the job configuration based on a key in the token; for example, set a different
   mapper/reducer class or a different input/output dataset for a Spark or MapReduce program.
   
-The API is intended to allow appropriate action to be taken in response to the token:
-logging, modifying the conditional execution, or terminating the execution of the workflow.
-
-Execution in the workflow can be made conditional, based on the information contained in
-the token. Execution can be terminated if a node in the workflow produces unexpected
-results.
+The API is intended to allow appropriate action to be taken in response to the token, including
+logging and modifying the conditional execution of the workflow based on the token.
 
 Once a run is completed, you can query the tokens from past workflow runs for analyses that
 determine which node was executed more frequently and when. You can retrieve the token values
@@ -203,7 +199,7 @@ that were added by a specific node in the workflow to debug the flow of executio
 Scope
 -----
 Two scopes |---| *System* and *User* |---| are provided for workflow keys. CDAP adds keys
-(such as MapReduce counters) under the *System* scope. User programs have their keys
+(such as MapReduce counters) under the *System* scope. Keys added by user programs are
 stored under the *User* scope.
 
 Putting and Getting Token Values
@@ -212,7 +208,7 @@ When a value is put into a token, it is stored under a specific key. Both keys a
 corresponding values must be non-null. The token stores additional information about the 
 context in which the key is being set, such as the unique name of the workflow node. 
 
-To put a value to a token, first obtain access to the token from the workflow context, and
+To put a value into a token, first obtain access to the token from the workflow context, and
 then set a value for a specific key, as shown in this example from a MapReduce mapper::
 
   public static class MyVerifier extends Mapper<LongWritable, Text, Text, NullWritable> {
@@ -250,9 +246,7 @@ values in the accumulators to be accessed through workflow tokens and vice-versa
 
 Persisting the WorkflowToken
 ----------------------------
-
-The RunRecord for the workflow contains the WorkflowToken as a property. This token
-is persisted after each action completes in a workflow. 
+The ``WorkflowToken`` is persisted after each action in the workflow has completed.
 
 Examples
 --------
@@ -265,9 +259,9 @@ In this code sample, we show how to update the WorkflowToken in a MapReduce prog
     WorkflowToken workflowToken = context.getWorkflowToken();
     if (workflowToken != null) {
       // Put the action type in the WorkflowToken
-      workflowToken.put("action_type", "MAPREDUCE");
+      workflowToken.put("action.type", "MAPREDUCE");
       // Put the start time for the action
-      workflowToken.put("startTime", String.valueOf(System.currentTimeMillis()));
+      workflowToken.put("start.time", String.valueOf(System.currentTimeMillis()));
     }
     ...
   }
@@ -278,7 +272,7 @@ In this code sample, we show how to update the WorkflowToken in a MapReduce prog
     WorkflowToken workflowToken = context.getWorkflowToken();
     if (workflowToken != null) {
       // Put the end time for the action
-      workflowToken.put("endTime", String.valueOf(System.currentTimeMillis()));
+      workflowToken.put("end.time", String.valueOf(System.currentTimeMillis()));
     }
     ...
   }
@@ -502,7 +496,7 @@ and from within a workflow with a predicate, fork and joins::
     WorkflowToken token = getContext().getToken();
     
     // Set the type of action of the current node:
-    token.put("action_type", "CUSTOM_ACTION");
+    token.put("action.type", "CUSTOM_ACTION");
  
     // Assume that we have the following Workflow: 
     //                                              |--> PurchaseByCustomer -->|
@@ -532,10 +526,10 @@ and from within a workflow with a predicate, fork and joins::
     // structure, in the "StatusReporter", you may want to know how many actions were
     // executed as a part of a run. If the number of nodes executed were below a certain
     // threshold, send an alert. Assuming that every node in the Workflow adds the key
-    // "action_type" with the value as action type for that node in the WorkflowToken,
+    // "action.type" with the value as action type for that node in the WorkflowToken,
     // you can determine the breakdown by action type in a particular Workflow run:
     
-    List<NodeValueEntry> nodeValues = token.getAll("action_type");
+    List<NodeValueEntry> nodeValues = token.getAll("action.type");
     int totalNodeExecuted = nodeValues.size();
     int mapReduceNodes = 0;
     int sparkNodes = 0;
@@ -565,7 +559,7 @@ and from within a workflow with a predicate, fork and joins::
  
     // To get the start time of the MapReduce program with unique name "PurchaseHistoryBuilder":
     
-    String startTime = token.get("startTime", "PurchaseHistoryBuilder");
+    String startTime = token.get("start.time", "PurchaseHistoryBuilder");
  
     // To get the most recent value of counter with group name
     // 'org.apache.hadoop.mapreduce.TaskCounter' and counter name 'MAP_INPUT_RECORDS':
--- a/cdap-docs/developers-manual/source/building-blocks/workflows.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/workflows.rst
@@ -160,7 +160,8 @@ then set a value for a specific key, as shown in this example from a MapReduce m
     }
   }
 
-The `Java API <>`__
+The `WorkflowToken Java API 
+<../../reference-manual/javadocs/co/cask/cdap/api/workflow/WorkflowToken.html)>`__
 includes methods for getting values for different keys, scopes, and nodes. The same
 key can be added to the workflow by different nodes, and there are methods to return a map of those
 key-value pairs. Convenience methods allow the putting and getting of non-string values
--- a/cdap-docs/developers-manual/source/building-blocks/workflows.rst
+++ b/cdap-docs/developers-manual/source/building-blocks/workflows.rst
@@ -92,6 +92,50 @@ By default, a workflow runs sequentially. Multiple instances of a workflow can b
 concurrently. To enable concurrent runs for a workflow, set its runtime argument
 ``concurrent.runs.enabled`` to ``true``.
 
+
+.. _workflow_token:
+
+Workflow Tokens
+===============
+
+In addition to passing the control flow from one node to the next, a **workflow token** is
+passed, available to each of the programs in the workflow. This allows programs to:
+
+- pass custom data (such as metrics, a status, or an error code) from one program in the 
+  workflow to the next subsequent programs; 
+- query and set the data in the token;
+- fetch the data from the token which was set by a specific node; and
+- determine the name of the node which most recently set the token value for a specific key;
+  e.g., the node who last set the ERROR flag in the token.
+  
+The last example is intended to allow appropriate action to be taken in response, such as
+logging, modifying the conditional execution, or terminating the execution of workflow.
+
+Execution in the workflow can be made conditional, based on the information contained in
+the token. Execution can be terminated if a node in the workflow produces unexpected
+results.
+
+Once a run is completed, you can query the tokens from past workflow runs for analyses that
+determine which node was executed more frequently and when. You can retrieve the token values
+that were added by a specific node in the workflow to debug the flow of execution.
+
+Scope
+
+Two scopes |---| *System* and *User* |---| are provided for workflow keys. CDAP adds keys
+(such as MapReduce counters) under the *System* scope. User programs add their keys under
+the *User* scope.
+
+
+Putting and Getting Token Values
+
+When a value is put into a token, it is stored under a specific key. Both keys and their
+corresponding values must be non-null. The token stores additional information about the 
+context in which the key is being set, such as the unique name of the workflow node. 
+
+
+
+
+
 .. _workflow_parallel:
 
 Parallelizing Workflow Execution
@@ -221,12 +265,11 @@ where ``MyPredicate`` is a public class which implements the ``Predicate`` inter
          if (input == null) {
             return false;
          }
-         Map<String, Map<String, Long>> mapReduceCounters = input.getToken().getMapReduceCounters();
-         if (mapReduceCounters == null) {
+         Value tokenValue = input.getToken().get("BuildProductProfile");
+         if (tokenValue == null) {
             return false;
          }
-         Map<String, Long> customCounters = mapReduceCounters.get("MyCustomCounters");
-         if (customCounters.get("BuildProductProfile") > 0) {
+         if (tokenValue.getAsInt() > 0) {
            return true;
          }
          return false;
@@ -274,11 +317,9 @@ otherwise, the other path will be taken. The diagram for this code would be:
    :width: 8in
    :align: center
 
-For this 3.0 release, CDAP only supports predicate conditions based on counters from
-previous MapReduce programs. Because of this, condition nodes are currently only useful
-when they follow MapReduce nodes; they can then make use of counters emitted by the
-MapReduce program to take the decision. In a later version, we will add more information
-to the ``WorkflowContext`` which will allow for more complex decisions.
+In addition to using counters of MapReduce programs, you can use :ref:`workflow tokens
+<workflow_token>` and base the logic on data stored in a token.
+
 
 Example of Using a Workflow
 ===========================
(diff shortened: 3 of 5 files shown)


cdap-docs/developers-manual/source/building-blocks/index.rst
```diff
@@ -44,8 +44,8 @@ write data through the data abstraction layer in CDAP.
 - :doc:`MapReduce Programs <mapreduce-programs>`
 - :doc:`Workflows <workflows>`
 - :doc:`Schedules <schedules>`
-- :doc:`Workers <workers>`
```
bdmogal: how do we decide the order in this index? Should it be alphabetical?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+.. _workflow_token:
+
+Workflow Tokens
+===============
+
+In addition to passing the control flow from one node to the next, a **workflow token** is
+passed, available to each of the programs in the workflow. This allows programs to:
+
+- pass custom data (such as metrics, a status, or an error code) from one program in the 
+  workflow to the next subsequent programs;
```
bdmogal: *next subsequent* seems like a Tautology? perhaps use either one of them but not both?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+===============
+
+In addition to passing the control flow from one node to the next, a **workflow token** is
+passed, available to each of the programs in the workflow. This allows programs to:
+
+- pass custom data (such as metrics, a status, or an error code) from one program in the 
+  workflow to the next subsequent programs; 
+- query and set the data in the token;
+- fetch the data from the token which was set by a specific node; and
+- determine the name of the node which most recently set the token value for a specific key;
```
bdmogal: hmmm, I'm not sure we expose this functionality. We expose the most recently set value for a key. However, there's no API to retrieve the node that set this value. There is an indirect way, ofcourse, but no direct API.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+  e.g., the node who last set the ERROR flag in the token.
+  
+The last example is intended to allow appropriate action to be taken in response, such as
```
bdmogal: Not just the last example, but all of the above use-cases are intended to allow this. You could also possibly:
- Alter job configuration based on a key in the token - e.g. set a different mapper/reducer class or a different input/output dataset for a spark/map-reduce program.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+results.
+
+Once a run is completed, you can query the tokens from past workflow runs for analyses that
+determine which node was executed more frequently and when. You can retrieve the token values
+that were added by a specific node in the workflow to debug the flow of execution.
+
+Scope
+-----
+Two scopes |---| *System* and *User* |---| are provided for workflow keys. CDAP adds keys
+(such as MapReduce counters) under the *System* scope. User programs add their keys under
```
bdmogal: "User programs add their keys *are stored* under..."

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+
+To put a value to a token, first obtain access to the token from the workflow context, and
+then set a value for a specific key, as shown in this example from a MapReduce mapper::
+
+  public static class MyVerifier extends Mapper<LongWritable, Text, Text, NullWritable> {
+    public void map(LongWritable key, Text value, Context context)
+      throws IOException, InterruptedException {
+      String profile;
+      WorkflowToken workflowToken = context.getWorkflowToken();
+      if (workflowToken != null) {
```
bdmogal: Should we point out somewhere that this null check is only required because this Mapper could be used outside of a workflow. When run from within a workflow, it is guaranteed to be non-null.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+<../../reference-manual/javadocs/co/cask/cdap/api/workflow/WorkflowToken.html)>`__
+includes methods for getting values for different keys, scopes, and nodes. The same
+key can be added to the workflow by different nodes, and there are methods to return a map of those
+key-value pairs. Convenience methods allow the putting and getting of non-string values
+through the use of the class Value.
+
+Persisting the WorkflowToken
+----------------------------
+
+The RunRecord for the workflow contains the WorkflowToken as a property. This token
```
bdmogal: nit: upper case Workflow?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+is persisted after each action completes in a Workflow. 
+
+Examples
+--------
+
+In this code sample, we show how to update the WorkflowToken in a MapReduce program::
```
bdmogal: It would be good to point out here that:
1. A token can only be updated in:
  -  ``beforeSubmit`` and ``onFinish`` methods of a MapReduce program
  - the Driver of a Spark program. 
  - ``initialize``, ``destroy`` methods of a custom action
  - Predicates of condition nodes.
2. You will get an exception if you try to update the workflow token in:
  - ``map``, ``reduce`` methods
  - Executors in Spark
  - ``run`` method in custom actions
Of course, you can read the workflow token in all the above.

The example in #3307 demonstrates some of the above.
2. There is however

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -274,11 +385,104 @@ otherwise, the other path will be taken. The diagram for this code would be:
...
-previous MapReduce programs. Because of this, condition nodes are currently only useful
-when they follow MapReduce nodes; they can then make use of counters emitted by the
-MapReduce program to take the decision. In a later version, we will add more information
-to the ``WorkflowContext`` which will allow for more complex decisions.
+Workflow Token with Forks and Joins
+-----------------------------------
+For workflows that involve forks and joins, a single instance of the workflow token is
+shared by all branches of the fork. Updates to the singleton are made thread-safe through
+synchronized updates, guaranteeing that value you obtain from reading the token is the
+last value written.
```
bdmogal: The last value written at *runtime*. This is a time-based guarantee.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -274,11 +385,104 @@ otherwise, the other path will be taken. The diagram for this code would be:
...
+This code sample shows how to obtain values from the token from within a custom action,
+and from within a workflow with a predicate, fork and joins::
+
+  @Override
+  public void run() {
+    ...
+    WorkflowToken token = getContext().getToken();
+    
+    // Set the type of action of the current node:
+    token.put("action_type", "CUSTOM_ACTION");
```
bdmogal: Hmmm, we should not allow updates to workflow token in ``run()``, since we do not allow it in ``map``, ``reduce``. Perhaps override the ``initialize(WorkflowContext context)`` method in this example, and update it there.

bdmogal: @sagarkapare thoughts ^^^?

cdap-docs/reference-manual/source/http-restful-api/workflow.rst
```diff
@@ -0,0 +1,142 @@
...
+   :stub-columns: 1
+
+   * - HTTP Request
+     - ``GET <base-url>/namespaces/default/apps/Purchase/workflows/PurchaseHistoryWorkflow/runs/57c...75e2/token``
+   * - HTTP Response
+     - ``{"key1":[{"nodeName": "node1", "value":"1"}]}``
+   * - Description
+     - Retrieves the token for a specific run of *PurchaseHistoryWorkflow* 
+
+   * -
```
bdmogal: Is this intentional?

bdmogal:
@awholegunch mostly looks good. The one thing I see missing is assigning unique names for programs. Please find an example for it in #3307 
[Passing a unique name to a program](https://github.com/caskdata/cdap/pull/3307/files#diff-b4709a94fe98f6ba902350502c57fae1R48)
[Setting the received unique name](https://github.com/caskdata/cdap/pull/3307/files#diff-b4709a94fe98f6ba902350502c57fae1R53)

bdmogal:
@awholegunch another thing we need to document is adding [Spark Accumulators](https://spark.apache.org/docs/latest/programming-guide.html#accumulators-a-nameaccumlinka) to workflow token. Also done in lines 67-88 in [here|https://github.com/caskdata/cdap/blob/feature/workflow-example/cdap-examples/WikipediaPipeline/src/main/scala/co/cask/cdap/examples/wikipedia/ScalaSparkLDA.scala]

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+- query and set the data in the token;
+- fetch the data from the token which was set by a specific node; and
+- determine the name of the node which most recently set the token value for a specific key;
+  e.g., the node who last set the ERROR flag in the token.
+  
+The last example is intended to allow appropriate action to be taken in response, such as
+logging, modifying the conditional execution, or terminating the execution of workflow.
+
+Execution in the workflow can be made conditional, based on the information contained in
+the token. Execution can be terminated if a node in the workflow produces unexpected
```
sagarkapare: I think its better to remove the line `Execution can be terminated if a node in the workflow produces unexpected..`,  as we don't have that capability yet.

cdap-docs/developers-manual/source/building-blocks/index.rst
awholegunch: That makes sense, now that we have so many. I have organized the Application "Building Block" portion to be alphabetical.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
sagarkapare: Maybe "Keys added by the user programs are stored under.."?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: fixed: "to subsequent programs; "

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+(such as MapReduce counters) under the *System* scope. User programs add their keys under
+the *User* scope.
+
+Putting and Getting Token Values
+--------------------------------
+When a value is put into a token, it is stored under a specific key. Both keys and their
+corresponding values must be non-null. The token stores additional information about the 
+context in which the key is being set, such as the unique name of the workflow node. 
+
+To put a value to a token, first obtain access to the token from the workflow context, and
```
sagarkapare: To put a value into a token,

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: I removed the upper-case workflow at the end of the sentence.

sagarkapare: Should we hide these details from the user? and just say that, `WorkflowToken` is persisted after each action in the Workflow completes.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+
+In this code sample, we show how to update the WorkflowToken in a MapReduce program::
+
+  @Override
+  public void beforeSubmit(MapReduceContext context) throws Exception {
+    ...
+    WorkflowToken workflowToken = context.getWorkflowToken();
+    if (workflowToken != null) {
+      // Put the action type in the WorkflowToken
+      workflowToken.put("action_type", "MAPREDUCE");
```
sagarkapare: Sorry, my bad, I should have updated document to reflect this. 
Its better to put the keys by `.` separated, instead of `_`. so the key will be `action.type`.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+      // Put the start time for the action
+      workflowToken.put("startTime", String.valueOf(System.currentTimeMillis()));
```
sagarkapare: here key will be "start.time"

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,114 @@ implement the ``run()`` method::
...
+    ...
+  }
+ 
+  @Override
+  public void onFinish(boolean succeeded, MapReduceContext context) throws Exception {
+    ...
+    WorkflowToken workflowToken = context.getWorkflowToken();
+    if (workflowToken != null) {
+      // Put the end time for the action
+      workflowToken.put("endTime", String.valueOf(System.currentTimeMillis()));
```
sagarkapare: `end.time`

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -274,11 +385,104 @@ otherwise, the other path will be taken. The diagram for this code would be:
...
+                                                     WorkflowToken.Scope.SYSTEM);
+  
+    // Use case 3: Since Workflow can have multiple complex conditions and forks in its
+    // structure, in the "StatusReporter", you may want to know how many actions were
+    // executed as a part of a run. If the number of nodes executed were below a certain
+    // threshold, send an alert. Assuming that every node in the Workflow adds the key
+    // "action_type" with the value as action type for that node in the WorkflowToken,
+    // you can determine the breakdown by action type in a particular Workflow run:
+    
+    List<NodeValueEntry> nodeValues = token.getAll("action_type");
```
sagarkapare: `action.type`. Basically all occurrences of "action_type" -> "action.type", "startTime" to "start.time",  and "endTime" to "end.time".

cdap-docs/developers-manual/source/building-blocks/workflows.rst
sagarkapare: I think its better to allow the updates from the `run()` method, if needed user can still add values to the token in the map and reduce methods by defining custom counters.

cdap-docs/reference-manual/source/http-restful-api/workflow.rst
awholegunch: Yes, I am attempting to introduce some space between the examples.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Added.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Added.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Added after example.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Added.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Adjusted.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Removed.

cdap-docs/reference-manual/source/http-restful-api/workflow.rst
awholegunch: I've made this better by adding or-bars.

awholegunch:
@bdmogal Please take another look.
@sagarkapare Please respond to @bdmogal 's comment/request.

sagarkapare:
@awholegunch already responded to that. Let me know if anything else I am missing.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Revised as per Sagar's comment.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Changed to:
```
The ``WorkflowToken`` is persisted after each action in the workflow has completed.
```

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: I've edited it accordingly.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,206 @@ implement the ``run()`` method::
...
+- ``beforeSubmit`` and ``onFinish`` methods of a MapReduce program;
+- Driver of a Spark program;
+- ``initialize`` and ``destroy`` methods of a custom action; and 
+- predicates of condition nodes.
+
+**You will get an exception** if you try to update the workflow token in:
+
+- map or reduce methods;
+- Executors in Spark programs;
+- ``run`` method in A custom actions.
```
bdmogal: typo: lower-case 'a'

awholegunch: fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -274,14 +477,108 @@ otherwise, the other path will be taken. The diagram for this code would be:
...
+      if (entry.getValue().equals("CONDITION")) {
+        conditions++;
+      }
+    }
+ 
+    // Use case 4: Retrieve values from the Workflow token.
+    
+    // To get the name of the last node which set the "ERROR" flag in the WorkflowToken:
+    
+    List<NodeValueEntry> errorNodeValueList = token.getAll("ERROR");
```
bdmogal: s/NodeValueEntry/NodeValue/g

That class got renamed

awholegunch: Fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,206 @@ implement the ``run()`` method::
...
+It's important to assign unique names to each component of the workflow, especially when you
+use  multiple instances of the same program in the same workflow.
+
+These unique names can be set when the Workflow is first configured, passed to the instance of the
+program, and then be used when the program performs its own configuration. An example
+of this is the :ref:`Wikipedia Pipeline<wikipedia-data-pipeline>` example, and its use of the
+*StreamToDataset* MapReduce program multiple times::
+
+  public class StreamToDataset extends AbstractMapReduce {
+    private static final Logger LOG = LoggerFactory.getLogger(StreamToDataset.class);
```
bdmogal: should we remove this line? Seems unnecessary.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,206 @@ implement the ``run()`` method::
...
 shown above.
 
+.. _workflow-unique-names:
+
+Assigning Unique Names
+----------------------
+It's important to assign unique names to each component of the workflow, especially when you
+use  multiple instances of the same program in the same workflow.
+
+These unique names can be set when the Workflow is first configured, passed to the instance of the
```
bdmogal: should we break this sentence?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,206 @@ implement the ``run()`` method::
...
+
+.. _workflow_token:
+
+Workflow Tokens
+===============
+
+In addition to passing the control flow from one node to the next, a **workflow token** is
+passed, available to each of the programs in the workflow. This allows programs to:
+
+- pass custom data (such as metrics, a status, or an error code) from one program in the
```
bdmogal: is there a chance that this could be confused with CDAP metrics? I ask because currently, there is no integration between workflow token and metrics

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,206 @@ implement the ``run()`` method::
...
+includes methods for getting values for different keys, scopes, and nodes. The same
+key can be added to the workflow by different nodes, and there are methods to return a map of those
+key-value pairs. Convenience methods allow the putting and getting of non-string values
+through the use of the class Value.
+
+Spark Accumulators and Workflow Tokens
+--------------------------------------
+`Spark Accumulators <https://spark.apache.org/docs/latest/programming-guide.html#accumulators-a-nameaccumlinka>`__ 
+can be accessed through the SparkContext, and used with workflow tokens. This allows the 
+values in the accumulators to be accessed through workflow tokens and vice-versa.
```
bdmogal: This is a little misleading. The vice-versa is not true. Also, for 3.1, users will have to manually add Accumulators to the workflow token. Example [here](https://github.com/caskdata/cdap/pull/3307/files#diff-aad4b0a2c7724edbf7d46f46590cd421R81)

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,206 @@ implement the ``run()`` method::
...
+context in which the key is being set, such as the unique name of the workflow node. 
+
+To put a value into a token, first obtain access to the token from the workflow context, and
+then set a value for a specific key, as shown in this example from a MapReduce mapper::
+
+  public static class MyVerifier extends Mapper<LongWritable, Text, Text, NullWritable> {
+    public void map(LongWritable key, Text value, Context context)
+      throws IOException, InterruptedException {
+      String profile;
+      WorkflowToken workflowToken = context.getWorkflowToken();
```
bdmogal: This is wrong. Mostly carried over from the wiki. The ``Context`` object here is a Hadoop class and has no idea about CDAP/workflow tokens. For accessing the workflow token in ``map()``/``reduce()`` methods, users will have to have their ``Mapper``/``Reducer`` classes implement ``ProgramLifecycle<MapReduceContext>``. After doing so, they can access workflow token in the ``initialize()``/``destroy`` methods. To access it in the ``map()``/``reduce()`` methods, they will have to cache a reference to the workflow token object as a class member in the ``initialize()`` method.

Example [here](https://github.com/caskdata/cdap/pull/3307/files#diff-aad4b0a2c7724edbf7d46f46590cd421R81)

The example doesn't quite cache it, because it doesn't need the token in the ``reduce`` method, but can be easily modified to store the workflow token as a class member. Infact, maybe we should add that to the example.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
bdmogal: I think this change may not have gotten pushed. Still shows as ``NodeValueEntry``.

awholegunch: Fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: I had it in, as this was copied from the example, but I'll remove it for this.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Fixed.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Could a user write their own code to read or write to metrics, and pass the values or names along? If so, I would leave it in, because that is how I would interpret it.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Added example,and adjusted accordingly.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: The example points the same Scala code as previous. Did you mean something else?

bdmogal: Oh! I meant [this](https://github.com/caskdata/cdap/pull/3307/files#diff-b2a58b12e4947fbeffac64abfef19557R109)

awholegunch: That makes more sense...did you want to change the example to include the caching of the token in a class member?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
bdmogal: Hmmm, so you can certainly emit CDAP metrics like usual. However, from application code, you cannot read CDAP metrics currently. So you cannot add them to the Workflow Token. 

As for the reverse, you can certainly read a Workflow Token key and emit it as a metric, but I can't think of a use case for it. 


In the general sense, metrics is fine. Just wondering if the distinction with CDAP metrics should be made clearer?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Updated example.

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Made clear by replacing "metric" with "a counter".

awholegunch:
@bdmogal Please take another look. Addressed all comments.

cdap-docs/developers-manual/build.sh
```diff
@@ -86,6 +86,10 @@ function download_includes() {
 #   download_readme_file_and_test ${includes_dir} ${ingest_url} 277ded1924cb8d9b52a007f262820002 cdap-stream-clients/javascript
   download_readme_file_and_test ${includes_dir} ${ingest_url} b6dedd629c708dbc68bc918b768edda5 cdap-stream-clients/python
   download_readme_file_and_test ${includes_dir} ${ingest_url} 5fc88ec3a658062775403f5be30afbe9 cdap-stream-clients/ruby
+
+  echo_red_bold "Check included example files for changes"
+#   test_an_include 3b877d261bf74f7521a14369d186a21b ../../cdap-examples/WikipediaPipeline/src/main/scala/co/cask/cdap/examples/wikipedia/ScalaSparkLDA.scala
```
bdmogal: why is this commented out?

cdap-docs/developers-manual/source/building-blocks/workflows.rst
```diff
@@ -86,12 +94,210 @@ implement the ``run()`` method::
...
+workflow token in either the ``initialize`` or ``destroy`` methods. To access it in the
+``map`` or ``reduce`` methods, you would need to cache a reference to the workflow token
+object as a class member in the ``initialize()`` method. This is because the context
+object passed to those methods is a Hadoop class that is unaware of CDAP and its workflow
+tokens.
+
+Here is an example, taken from the
+:ref:`Wikipedia Pipeline<wikipedia-data-pipeline>` example's ``TopNMapReduce.java``:
+
+.. literalinclude:: /../../../cdap-examples/WikipediaPipeline/src/main/java/co/cask/cdap/examples/wikipedia/TopNMapReduce.java
```
bdmogal: So the inline text (which was to be updated) just got replaced with this?

bdmogal:
Just 1 question and 1 comment, otherwise LGTM! Thanks @awholegunch !

cdap-docs/developers-manual/build.sh
awholegunch: I need to get the hashes of these files in before we can turn on the guards. I will do that in a later PR. Tracked with this issue: https://issues.cask.co/browse/CDAP-3133

cdap-docs/developers-manual/source/building-blocks/workflows.rst
awholegunch: Yes. Is that fine?

bdmogal: yup. all good.

bdmogal:
LGTM!

PR #3245, opened 2015-07-22
CDAP-3066 fix FileBatchSource

jayfjin:
JIRA - https://issues.cask.co/browse/CDAP-3066
BUILD - http://builds.cask.co/browse/CDAP-DUT2289-1

Local variable aliased class variable blocking functionality.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
@@ -139,7 +139,7 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
 
     if (config.timeTable != null) {
       table = context.getDataset(config.timeTable);
-      String datesToRead = Bytes.toString(table.read(LAST_TIME_READ));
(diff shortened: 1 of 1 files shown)


bdmogal:
The code fix is pretty innocuous. However, what was the failure? Is it possible to add a unit test?

chtyim:
From the descriptions, seems like it's a bad choice of variable name?

jayfjin:
Upon failure, the MapReduce should try to re-read in data. This does not happen. datesToRead is an accurate description of the variable, it denotes the dates that still need to be read in.

anwar6953:
I think @chtyim 's point is that it gets its value from a field denoted by `LAST_TIME_READ`, which seems like a single value, whereas datesToRead seems like it has multiple values.

jayfjin:
that's because datesToRead can have multiple values.

anwar6953:
Also, `LAST_TIME_READ` denotes that it was already read/processed, but `datesToRead` denotes that it has yet to be processed.

jayfjin:
datesToRead has yet to be processed upon read in from the table...

anwar6953:
The confusion was that `LAST_TIME_READ` seems to contradict `datesToRead`, but it's fine because `LAST_TIME_READ` was the exclusive end-range of the previous run's range to read.

anwar6953:
LGTM

PR #3246, opened 2015-07-22
Make QueueTest and HBaseTableTest runs faster

chtyim:
About 6-7 mins faster in bamboo


abaranau:
LGTM

chtyim:
And build passed in http://builds.cask.co/browse/CDAP-DUT2288-5

PR #3247, opened 2015-07-22
[CDAP-3037] handle type conflict for duplicate dependencies of a dataset type

anew:
If a dataset embeds Table, and also embeds another dataset from a different module that itself embeds a Table, DatasetDefinitionLoader will try to register the Table's module twice (once directly for the embedded Table, once indirectly when adding the dependencies of the other dataset). The second time it will fail with a conflict because the type is already defined. This happens not only for table, but for any dataset that is embedded both directly and transitively through a different module (note this does not happen for IndexedTable, because it is in the same module as Table).

Fix is to ignore the TypeConflictException that happens in this case.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/DatasetDefinitionLoader.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/DatasetDefinitionLoader.java
@@ -24,6 +24,7 @@ import co.cask.cdap.common.lang.ProgramClassLoader;
 import co.cask.cdap.common.lang.jar.BundleJarUtil;
 import co.cask.cdap.common.utils.DirUtils;
 import co.cask.cdap.data2.dataset2.InMemoryDatasetDefinitionRegistry;
+import co.cask.cdap.data2.dataset2.TypeConflictException;
 import co.cask.cdap.data2.dataset2.module.lib.DatasetModules;
 import co.cask.cdap.proto.DatasetModuleMeta;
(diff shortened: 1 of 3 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/DatasetDefinitionLoader.java
```diff
@@ -86,7 +87,15 @@ public DatasetDefinitionLoader(LocationFactory locationFactory) {
         }
         Class<?> moduleClass = ClassLoaders.loadClass(moduleMeta.getClassName(), classLoader, this);
         DatasetModule module = DatasetModules.getDatasetModule(moduleClass);
-        module.register(registry);
+        try {
+          // TODO avoid duplicate registering of the same module (a type may have multiple transitive
```
albertshau: is there a jira open for this todo?

albertshau:
minor comment, then lgtm

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/type/DatasetDefinitionLoader.java
anew: It is more a reminder that we should do this when we redo the dataset framework. In the current framework it is not doable.

anew: OK, I changed the comment to explain why this is done this way, and why it can't be done better.

albertshau:
:+1:

PR #3249, opened 2015-07-22
Add support to access tpfs through service in LogAnalysisApp

rohitsinha54:
- Spark program now writes the TPFS with partition key as the logicalStartTime by passing it as arguments while calling writeToDataset rather than requiring the user user to provide the partition time as a dataset argument. https://github.com/caskdata/cdap/pull/3215
- Exposes to handlers: One to get a list of all partitions and another to get contents of a partition as a string.
- Update the unit test for the same.

Build: http://builds.cask.co/browse/CDAP-RBT333-1

### DIFF
--- a/cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
+++ b/cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
@@ -22,6 +22,7 @@ import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.data.stream.Stream;
 import co.cask.cdap.api.dataset.lib.FileSetProperties;
 import co.cask.cdap.api.dataset.lib.KeyValueTable;
+import co.cask.cdap.api.dataset.lib.TimePartitionDetail;
 import co.cask.cdap.api.dataset.lib.TimePartitionedFileSet;
 import co.cask.cdap.api.service.Service;
 import co.cask.cdap.api.service.http.AbstractHttpServiceHandler;
@@ -30,12 +31,21 @@ import co.cask.cdap.api.service.http.HttpServiceResponder;
 import co.cask.cdap.api.spark.AbstractSpark;
 import co.cask.cdap.api.workflow.AbstractWorkflow;
 import co.cask.cdap.api.workflow.Workflow;
+import co.cask.cdap.common.io.Locations;
 import com.google.common.base.Charsets;
+import com.google.common.collect.Sets;
+import com.google.common.io.CharStreams;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import org.apache.hadoop.mapreduce.lib.output.TextOutputFormat;
+import org.apache.twill.filesystem.Location;
 
+import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.util.Date;
+import java.util.Set;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
@@ -49,9 +59,10 @@ public class LogAnalysisApp extends AbstractApplication {
   public static final String LOG_STREAM = "logStream";
   public static final String HIT_COUNTER_SERVICE = "HitCounterService";
   public static final String RESPONSE_COUNTER_SERVICE = "ResponseCounterService";
+  public static final String REQUEST_COUNTER_SERVICE = "RequestCounterService";
   public static final String RESPONSE_COUNT_STORE = "responseCount";
   public static final String HIT_COUNT_STORE = "hitCount";
-  public static final String IP_COUNT_STORE = "ipCount";
+  public static final String REQ_COUNT_STORE = "reqCount";
 
   @Override
   public void configure() {
@@ -69,11 +80,12 @@ public class LogAnalysisApp extends AbstractApplication {
     // Services to query for result
     addService(HIT_COUNTER_SERVICE, new HitCounterServiceHandler());
     addService(RESPONSE_COUNTER_SERVICE, new ResponseCounterHandler());
+    addService(REQUEST_COUNTER_SERVICE, new RequestCounterHanlder());
 
     // Datasets to store output after processing
     createDataset(RESPONSE_COUNT_STORE, KeyValueTable.class);
     createDataset(HIT_COUNT_STORE, KeyValueTable.class);
-    createDataset(IP_COUNT_STORE, TimePartitionedFileSet.class, FileSetProperties.builder()
+    createDataset(REQ_COUNT_STORE, TimePartitionedFileSet.class, FileSetProperties.builder()
       .setOutputFormat(TextOutputFormat.class)
       .setOutputProperty(TextOutputFormat.SEPERATOR, ":").build());
   }
@@ -163,6 +175,82 @@ public class LogAnalysisApp extends AbstractApplication {
       } else {
         responder.sendString(String.valueOf(Bytes.toLong(read)));
       }
+    }
+  }
+
+  /**
+   * A Service which serves the number of requests made by unique ip address from a {@link TimePartitionedFileSet}
+   */
+  public static final class RequestCounterHanlder extends AbstractHttpServiceHandler {
+
+    private static final Gson GSON = new Gson();
+    static final String REQUEST_COUNTER_PARTITIONS_HANDLER = "reqcount";
+    static final String REQUEST_FILE_CONTENT_HANDLER = "reqfile";
+    static final String REQUEST_FILE_PATH_HANDLER_KEY = "time";
+
+    @UseDataSet(REQ_COUNT_STORE)
+    private TimePartitionedFileSet reqCountStore;
+
+    /**
+     * Handler which lists all the different time partitions available in the {@link LogAnalysisApp#REQ_COUNT_STORE}
+     * {@link TimePartitionedFileSet}
+     */
+    @Path(REQUEST_COUNTER_PARTITIONS_HANDLER)
+    @GET
+    public void getRequestFilesetPartitions(HttpServiceRequest request, HttpServiceResponder responder) {
+
+      // get all the existing paritions
+      Set<TimePartitionDetail> partitionsByTime = reqCountStore.getPartitionsByTime(0, Long.MAX_VALUE);
+
+      if (partitionsByTime.isEmpty()) {
+        responder.sendString(HttpURLConnection.HTTP_NO_CONTENT, "No request count information found", Charsets.UTF_8);
+      } else {
+        Set<String> formatedTimes = Sets.newHashSet();
+        for (TimePartitionDetail timePartitionDetail : partitionsByTime) {
+          String partitionTime = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT).format(
+            new Date(timePartitionDetail.getTime()));
+          formatedTimes.add(partitionTime);
+        }
+        responder.sendJson(HttpURLConnection.HTTP_OK, formatedTimes);
+      }
+    }
+
+    /**
+     * Handler which reads all the parts files from a given partition in {@link LogAnalysisApp#REQ_COUNT_STORE}
+     * {@link TimePartitionedFileSet} and send it as a string.
+     * Note: We make an assumption here that the contents for partitions in the tpfs for this example is not very huge.
+     * This method of serving contents is not ideal for large contents.
+     */
+    @Path(REQUEST_FILE_CONTENT_HANDLER)
+    @POST
+    public void getRequestFilesetContents(HttpServiceRequest request, HttpServiceResponder responder) {
+
+      String partition = GSON.fromJson(Charsets.UTF_8.decode(request.getContent()).toString(),
+                                 JsonObject.class).get(REQUEST_FILE_PATH_HANDLER_KEY).getAsString();
+      long partitionKey = 0;
+      try {
+        partitionKey = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT).parse(partition).getTime();
+      } catch (ParseException e) {
+        responder.sendError(HttpURLConnection.HTTP_BAD_REQUEST, "Failed to parse the given string to a timestamp");
+      }
+
+      final Location location = reqCountStore.getPartitionByTime(partitionKey).getLocation();
+      if (location == null) {
+        responder.sendError(HttpURLConnection.HTTP_NO_CONTENT, "No files for the given date time string");
+      }
+
+      StringBuilder contents = new StringBuilder();
+      try {
+        for (Location file : location.list()) {
+          if (file.getName().startsWith("part")) {
+            contents.append(CharStreams.toString(CharStreams.newReaderSupplier(Locations.newInputSupplier(file),
+                                                                               Charsets.UTF_8)));
+          }
+        }
+      } catch (IOException e) {
+        responder.sendError(HttpURLConnection.HTTP_INTERNAL_ERROR, "Failed to access files.");
+      }
+      responder.sendString(HttpURLConnection.HTTP_OK, contents.toString(), Charsets.UTF_8);
     }
   }
 }
--- a/cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
+++ b/cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
@@ -190,7 +190,7 @@ public class LogAnalysisApp extends AbstractApplication {
     static final String REQUEST_COUNTER_PARTITIONS_PATH = "reqcount";
     static final String REQUEST_FILE_CONTENT_PATH = "reqfile";
     static final String REQUEST_FILE_PATH_HANDLER_KEY = "time";
-    public static final DateFormat SHORT_DATE_FORMAT = DateFormat.getDateTimeInstance(DateFormat.SHORT,
+    private static final DateFormat SHORT_DATE_FORMAT = DateFormat.getDateTimeInstance(DateFormat.SHORT,
                                                                                       DateFormat.SHORT);
 
     @UseDataSet(REQ_COUNT_STORE)
@@ -245,21 +245,18 @@ public class LogAnalysisApp extends AbstractApplication {
       try {
         for (Location file : location.list()) {
           if (file.getName().startsWith("part")) {
-            String line;
             try (BufferedReader reader = new BufferedReader(new InputStreamReader(file.getInputStream(),
                                                                                   Charsets.UTF_8))) {
+              String line;
               while ((line = reader.readLine()) != null) {
                 int idx = line.indexOf(":");
                 requestCountsMap.put(line.substring(0, idx), Integer.parseInt(line.substring(idx + 1)));
               }
-            } catch (IOException e) {
-              responder.sendError(HttpURLConnection.HTTP_INTERNAL_ERROR, "Failed to read file.");
-              return;
             }
           }
         }
       } catch (IOException e) {
-        responder.sendError(HttpURLConnection.HTTP_INTERNAL_ERROR, "Failed to access files.");
+        responder.sendError(HttpURLConnection.HTTP_INTERNAL_ERROR, e.getMessage());
         return;
       }
       responder.sendJson(HttpURLConnection.HTTP_OK, requestCountsMap);
--- a/cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
+++ b/cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
@@ -190,6 +190,8 @@ public class LogAnalysisApp extends AbstractApplication {
     static final String REQUEST_COUNTER_PARTITIONS_PATH = "reqcount";
     static final String REQUEST_FILE_CONTENT_PATH = "reqfile";
     static final String REQUEST_FILE_PATH_HANDLER_KEY = "time";
+    public static final DateFormat SHORT_DATE_FORMAT = DateFormat.getDateTimeInstance(DateFormat.SHORT,
+                                                                                      DateFormat.SHORT);
 
     @UseDataSet(REQ_COUNT_STORE)
     private TimePartitionedFileSet reqCountStore;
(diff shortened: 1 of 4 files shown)


anew:
also validated on cluster: http://unew2015-1000.dev.continuuity.net:9999/ns/dummy/apps/SportResults/programs/mapreduce/ScoreCounter/runs/d895da03-3167-11e5-8ea9-42010af019f0
Under logs, we see that it uses EmptyInputFormat. and 
{noformat}
[andreas@unew2015-1000 ~]$ hadoop fs -ls -R /cdap/namespaces/dummy/data/totals/files/nba
-rw-r--r--   1 yarn supergroup          0 2015-07-23 18:23 /cdap/namespaces/dummy/data/totals/files/nba/_SUCCESS
-rw-r--r--   1 yarn supergroup          0 2015-07-23 18:23 /cdap/namespaces/dummy/data/totals/files/nba/part-r-00000
{noformat}
shows how the MR successfully wrote an output of size 0 (no records written).

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
       }
     }
   }
+
+  /**
+   * A Service which serves the number of requests made by unique ip address from a {@link TimePartitionedFileSet}
+   */
+  public static final class RequestCounterHanlder extends AbstractHttpServiceHandler {
```
anew: should this be called a Handler, not a Hanlder?

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+
+    private static final Gson GSON = new Gson();
+    static final String REQUEST_COUNTER_PARTITIONS_HANDLER = "reqcount";
+    static final String REQUEST_FILE_CONTENT_HANDLER = "reqfile";
```
anew: these are not handlers, they are paths

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+      Set<TimePartitionDetail> partitionsByTime = reqCountStore.getPartitionsByTime(0, Long.MAX_VALUE);
+
+      if (partitionsByTime.isEmpty()) {
+        responder.sendString(HttpURLConnection.HTTP_NO_CONTENT, "No request count information found", Charsets.UTF_8);
+      } else {
+        Set<String> formatedTimes = Sets.newHashSet();
+        for (TimePartitionDetail timePartitionDetail : partitionsByTime) {
+          String partitionTime = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT).format(
+            new Date(timePartitionDetail.getTime()));
+          formatedTimes.add(partitionTime);
```
anew: formatted

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+     */
+    @Path(REQUEST_COUNTER_PARTITIONS_HANDLER)
+    @GET
+    public void getRequestFilesetPartitions(HttpServiceRequest request, HttpServiceResponder responder) {
+
+      // get all the existing paritions
+      Set<TimePartitionDetail> partitionsByTime = reqCountStore.getPartitionsByTime(0, Long.MAX_VALUE);
+
+      if (partitionsByTime.isEmpty()) {
+        responder.sendString(HttpURLConnection.HTTP_NO_CONTENT, "No request count information found", Charsets.UTF_8);
```
anew: I think it is better to return an empty list with status 200.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+      } else {
+        Set<String> formatedTimes = Sets.newHashSet();
```
anew: Should this somehow be sorted by date? I think with HashSet they will be in random order...

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+      } catch (ParseException e) {
+        responder.sendError(HttpURLConnection.HTTP_BAD_REQUEST, "Failed to parse the given string to a timestamp");
+      }
+
+      final Location location = reqCountStore.getPartitionByTime(partitionKey).getLocation();
+      if (location == null) {
+        responder.sendError(HttpURLConnection.HTTP_NO_CONTENT, "No files for the given date time string");
+      }
+
+      StringBuilder contents = new StringBuilder();
```
anew: why build this string in memory? Can't you write directly to the responder's channel?

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+      long partitionKey = 0;
+      try {
+        partitionKey = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT).parse(partition).getTime();
+      } catch (ParseException e) {
+        responder.sendError(HttpURLConnection.HTTP_BAD_REQUEST, "Failed to parse the given string to a timestamp");
+      }
+
+      final Location location = reqCountStore.getPartitionByTime(partitionKey).getLocation();
+      if (location == null) {
+        responder.sendError(HttpURLConnection.HTTP_NO_CONTENT, "No files for the given date time string");
```
anew: this should be a 404.

cdap-examples/LogAnalysis/src/test/java/co/cask/cdap/examples/loganalysis/LogAnalysisAppTest.java
```diff
@@ -57,7 +50,7 @@
   private static final String TOTAL_HITS_VALUE = "2";
   private static final String TOTAL_RESPONSE_VALUE = "2";
   private static final String RESPONSE_CODE = "200";
-  private static final String TPFS_RESULT = "127.0.0.1:2";
+  private static final String TPFS_RESULT = "127.0.1.1:1\n127.0.0.1:2\n";
```
anew: how did the expected results change? I don't see a change input data or the program... How did the test pass before?

anew:
@rohitsinha54 I have done a pass through this PR. Please see my comments. -Andreas

cdap-examples/LogAnalysis/src/test/java/co/cask/cdap/examples/loganalysis/LogAnalysisAppTest.java
rohitsinha54: Earlier, we were only reading one part file and verifying the result not all the part files.
Also, I have a question on this one:
Say, first line comes from part0 file which is written by reducer0 and second line comes from part1 which is written by reducer1. Can we assume that they will always be in this order ?

anew: No you can't, there is no guarantee. It is unlikely, but the next version of Hadoop might order them differently. So the clean way of doing this would be to return a Map in Json format, then the test case can parse it as a Map and to a map.equals with en expected result.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+    @POST
+    public void getRequestFilesetContents(HttpServiceRequest request, HttpServiceResponder responder) {
+
+      String partition = GSON.fromJson(Charsets.UTF_8.decode(request.getContent()).toString(),
+                                 JsonObject.class).get(REQUEST_FILE_PATH_HANDLER_KEY).getAsString();
+      long partitionKey = 0;
+      try {
+        partitionKey = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT).parse(partition).getTime();
+      } catch (ParseException e) {
+        responder.sendError(HttpURLConnection.HTTP_BAD_REQUEST, "Failed to parse the given string to a timestamp");
```
chtyim: Without returning, it will keep going.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
chtyim: should return after sending error.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+      StringBuilder contents = new StringBuilder();
+      try {
+        for (Location file : location.list()) {
+          if (file.getName().startsWith("part")) {
+            contents.append(CharStreams.toString(CharStreams.newReaderSupplier(Locations.newInputSupplier(file),
+                                                                               Charsets.UTF_8)));
+          }
+        }
+      } catch (IOException e) {
+        responder.sendError(HttpURLConnection.HTTP_INTERNAL_ERROR, "Failed to access files.");
```
chtyim: Same here, should return.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+      final Location location = reqCountStore.getPartitionByTime(partitionKey).getLocation();
+      if (location == null) {
+        responder.sendError(HttpURLConnection.HTTP_NO_CONTENT, "No files for the given date time string");
+      }
+
+      StringBuilder contents = new StringBuilder();
+      try {
+        for (Location file : location.list()) {
+          if (file.getName().startsWith("part")) {
+            contents.append(CharStreams.toString(CharStreams.newReaderSupplier(Locations.newInputSupplier(file),
```
chtyim: Locations is an internal class. You shouldn't use it. Did you added dependency on cdap-common? You shouldn't do that for example projects.

cdap-examples/LogAnalysis/src/main/scala/co/cask/cdap/examples/loganalysis/ResponseCounterProgram.scala
```diff
@@ -51,7 +51,9 @@ class ResponseCounterProgram extends ScalaSparkProgram {
     val ipCounts = logsData.map(x => (ApacheAccessLog.
       parseFromLogLine(x._2.toString).getIpAddress, 1L)).reduceByKey(_ + _)
 
-    context.writeToDataset(ipCounts, LogAnalysisApp.IP_COUNT_STORE,
-      classOf[String], classOf[java.lang.Long])
+    val outputArgs: util.HashMap[String, String] = new util.HashMap[String, String]
```
chtyim: No need to have type. Can simply be `var outputArgs = new util.HashMap[String, String]`

cdap-examples/LogAnalysis/src/main/scala/co/cask/cdap/examples/loganalysis/ResponseCounterProgram.scala
```diff
@@ -51,7 +51,9 @@ class ResponseCounterProgram extends ScalaSparkProgram {
...
+    TimePartitionedFileSetArguments.setOutputPartitionTime(outputArgs, endTime)
+    context.writeToDataset(ipCounts, LogAnalysisApp.REQ_COUNT_STORE,
+      classOf[String], classOf[java.lang.Long], outputArgs)
```
chtyim: No need to have `java.lang`.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +177,80 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+                                                                               Charsets.UTF_8)));
+          }
+        }
+      } catch (IOException e) {
+        responder.sendError(HttpURLConnection.HTTP_INTERNAL_ERROR, "Failed to access files.");
+      }
+      responder.sendString(HttpURLConnection.HTTP_OK, contents.toString(), Charsets.UTF_8);
```
anwar6953: can't you just use `responder.sendString(contents.toString())`?

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +180,83 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+    @Path(REQUEST_COUNTER_PARTITIONS_PATH)
+    @GET
+    public void getRequestFilesetPartitions(HttpServiceRequest request, HttpServiceResponder responder) {
+
+      // get all the existing paritions
+      Set<TimePartitionDetail> partitionsByTime = reqCountStore.getPartitionsByTime(0, Long.MAX_VALUE);
+
+      SortedSet<String> formattedTimes = new TreeSet<>();
+      for (TimePartitionDetail timePartitionDetail : partitionsByTime) {
+        String partitionTime = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT).format(
```
chtyim: It's pretty expensive to create the DataFormat inside a loop, while it is not changing. Since CDAP guarantees there would be no concurrent calls made to the same handler instance, you can make the DateFormat a static constant.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +180,83 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+     */
+    @Path(REQUEST_FILE_CONTENT_PATH)
+    @POST
+    public void getRequestFilesetContents(HttpServiceRequest request, HttpServiceResponder responder) {
+
+      String partition = GSON.fromJson(Charsets.UTF_8.decode(request.getContent()).toString(),
+                                       JsonObject.class).get(REQUEST_FILE_PATH_HANDLER_KEY).getAsString();
+      long partitionKey = 0;
+      try {
+        partitionKey = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT).parse(partition).getTime();
```
chtyim: Same as above.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +180,83 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+
+    /**
+     * Handler which reads all the parts files from a given partition in {@link LogAnalysisApp#REQ_COUNT_STORE}
+     * {@link TimePartitionedFileSet} and send it as a string.
+     * Note: We make an assumption here that the contents for partitions in the tpfs for this example is not very huge.
+     * This method of serving contents is not ideal for large contents.
+     */
+    @Path(REQUEST_FILE_CONTENT_PATH)
+    @POST
+    public void getRequestFilesetContents(HttpServiceRequest request, HttpServiceResponder responder) {
```
chtyim: Why the get call is a POST?

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +180,83 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+      }
+
+      Map<String, Integer> requestCountsMap = Maps.newHashMap();
+      try {
+        for (Location file : location.list()) {
+          if (file.getName().startsWith("part")) {
+            String line;
+            BufferedReader reader = new BufferedReader(new InputStreamReader(file.getInputStream(), Charsets.UTF_8));
+            while ((line = reader.readLine()) != null) {
+              String[] reqeustCount = line.split(":");
```
chtyim: Calling `String.split` in a nested loop is really bad (it compiles a new regex on every call). Better call `indexOf` and use substring.

```java
int idx = line.indexOf(":");
map.put(line.substring(0, idx), line.substring(idx + 1));
```

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +180,83 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+        responder.sendError(HttpURLConnection.HTTP_NOT_FOUND, "No files for the given date time string");
+        return;
+      }
+
+      Map<String, Integer> requestCountsMap = Maps.newHashMap();
+      try {
+        for (Location file : location.list()) {
+          if (file.getName().startsWith("part")) {
+            String line;
+            BufferedReader reader = new BufferedReader(new InputStreamReader(file.getInputStream(), Charsets.UTF_8));
```
chtyim: You have to close the reader after use. Use the `try(Resource)` syntax.

rohitsinha54:
Build: http://builds.cask.co/browse/CDAP-RBT333-4

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +180,89 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+  /**
+   * A Service which serves the number of requests made by unique ip address from a {@link TimePartitionedFileSet}
+   */
+  public static final class RequestCounterHandler extends AbstractHttpServiceHandler {
+
+    private static final Gson GSON = new Gson();
+    static final String REQUEST_COUNTER_PARTITIONS_PATH = "reqcount";
+    static final String REQUEST_FILE_CONTENT_PATH = "reqfile";
+    static final String REQUEST_FILE_PATH_HANDLER_KEY = "time";
+    public static final DateFormat SHORT_DATE_FORMAT = DateFormat.getDateTimeInstance(DateFormat.SHORT,
```
chtyim: Why it is public?

rohitsinha54: aahh! Intellij refactoring :/ 
Will change one sec.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +180,89 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+      if (location == null) {
+        responder.sendError(HttpURLConnection.HTTP_NOT_FOUND, "No files for the given date time string");
+        return;
+      }
+
+      Map<String, Integer> requestCountsMap = Maps.newHashMap();
+      try {
+        for (Location file : location.list()) {
+          if (file.getName().startsWith("part")) {
+            String line;
```
chtyim: Move it closer to where it is used (move it inside try block).

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
```diff
@@ -165,4 +180,89 @@ public void centers(HttpServiceRequest request, HttpServiceResponder responder,
...
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(file.getInputStream(),
+                                                                                  Charsets.UTF_8))) {
+              while ((line = reader.readLine()) != null) {
+                int idx = line.indexOf(":");
+                requestCountsMap.put(line.substring(0, idx), Integer.parseInt(line.substring(idx + 1)));
+              }
+            } catch (IOException e) {
```
chtyim: No need to have catch here, right? You already catching it at the outer level.

chtyim:
Just couple minor comments left.

cdap-examples/LogAnalysis/src/main/java/co/cask/cdap/examples/loganalysis/LogAnalysisApp.java
rohitsinha54: I think its cleaner as I want to differentiate between error in reading file and error in accessing the partition dir. Don't you agree ?

chtyim: But you response with the same status code. Also, the root cause should be determined from the exception message. If you response with the e.getMessage, then you only need one catch.

chtyim:
LGTM

PR #3251, opened 2015-07-22
Don't try to upgrade usage.registy on fresh cluster

rohitsinha54:
- usage.registry table is created when something runs on a cluster and a usage record needs to be written. On a fresh cluster the usage.registry will not exists and will not be found. No upgrade is necessary in this case.


ghelmling:
lgtm

PR #3252, opened 2015-07-22
[CDAP-3053] schema javadoc should mention that toString() returns JSON

anew:


cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
```diff
@@ -473,6 +473,9 @@ public Schema getUnionSchema(int idx) {
     return (unionSchemas == null || idx < 0 || unionSchemas.size() <= idx) ? null : unionSchemas.get(idx);
   }
 
+  /**
+   * @return a JSON representation of this schema, which can be parsed with {@link #parseJson}.
```
awholegunch: No trailing period on @param or @return tags.

"the JSON representation of this schema,..."

awholegunch:
one comment, otherwise LGTM.

PR #3254, opened 2015-07-22
Feature/ui adapter create edit schema

elfenheart:
- Only the first empty row is clickable to edit
- 'Enter' key now works again
- Created adapters properties styling update

schemaedit


ajainarayanan:
:+1: LGTM

PR #3258, opened 2015-07-23
CDAP-3042 fix bug where buffered operations weren't seen in multiget

albertshau:
Fixes a bug in BufferingTable where multiget calls for HBase were
only returning persisted data and could not see changes made in the
same transaction. In order to fix this, BufferingTable now has a
getPersisted(List<Get> gets) method for subclasses to override.
BufferingTable will then apply buffered changes on top of the
persisted changes.

Also adding a service endpoint to the UserProfiles example app to
demonstrate usage of multi-get on a Table.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
@@ -395,11 +395,6 @@ public abstract class BufferingTable extends AbstractTable implements MeteredDat
     reportRead(1);
     // checking if the row was deleted inside this tx
     NavigableMap<byte[], Update> buffCols = buff.get(row);
-    boolean rowDeleted = buffCols == null && buff.containsKey(row);
-    // ANDREAS: can this ever happen?
-    if (rowDeleted) {
-      return new Result(row, Collections.<byte[], byte[]>emptyMap());
-    }
 
     // NOTE: since we cannot tell the exact column set, we always have to go to persisted store.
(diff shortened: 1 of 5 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-RBT336

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/BufferingTable.java
```diff
@@ -406,6 +426,52 @@ public Row get(byte[] row, byte[] startColumn, byte[] stopColumn, int limit) {
...
+
+        // navigable copy of the persisted data. Implementation may return immutable or unmodifiable maps,
+        // so we make a copy here.
+        NavigableMap<byte[], byte[]> rowColumns = Maps.newTreeMap(Bytes.BYTES_COMPARATOR);
+        rowColumns.putAll(persistedRow);
+
+        byte[] row = get.getRow();
+        NavigableMap<byte[], Update> buffCols = buff.get(row);
+        // if the row was deleted in the buffer, add an empty row to the results
+        if (buffCols == null && buff.containsKey(row)) {
```
ghelmling: I see a similar check in the get(Get) method, but I don't think this is actually needed, as we never seem to put a null value for the entire row.  Instead the delete(byte[] row) method first retrieves all columns, then puts a null value for each column.

ghelmling:
One comment, otherwise looks good.

albertshau:
@ghelmling addressed comment, can you take another look?

ghelmling:
lgtm

PR #3260, opened 2015-07-23
[CDAP 2945] PFS should not fail if the input partition filter matches no partitions

anew:
This changes partitioned file sets to use an EmptyInputFormat if the input filter does not match anything. (the actual input format, which most likely extends FileInputFormat, would throw an exception if no input paths are configured). This does not change the behavior if no input filter was given - that will still throw an exception, unless the input was specified directly using FileSetArguments, bypassing the partition filter (not a common use case, but supported). 

Similar for TimePartitionedFileSet: If an input time range was specified, and it matches no partitions, use EmptyInputFormat. If no time range was given, fall back to PartitionedFileSet's method of input selection (using partition filters).

Also changes the behavior if not output path is given for a file set. It used to write to the base location, which is wrong. I will now fail (just like FileOutputFormat would) if no output path is given.

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/batch/MapReduceWithPartitionedTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/batch/MapReduceWithPartitionedTest.java
@@ -251,6 +251,23 @@ public class MapReduceWithPartitionedTest {
           Assert.assertNull(row.get("y"));
         }
       });
+
+    // now run a map/reduce that reads no partitions
+    TimePartitionedFileSetArguments.setInputStartTime(inputArgs, time - TimeUnit.MINUTES.toMillis(10));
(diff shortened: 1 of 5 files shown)


anew:
Build passes: http://builds.cask.co/browse/CDAP-RBT338-2

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/batch/MapReduceWithPartitionedTest.java
```diff
@@ -379,6 +396,28 @@ public void apply() {
           Assert.assertNull(row.get("y"));
         }
       });
+
+    // a partition filter that matches the no key
```
anwar6953: I think this comment reads more easily without "the"

anwar6953:
Besides the comments about comments, remainder looked good to me.

PR #3262, opened 2015-07-23
Feature/ui styling sink output modal

astewes:
*  Adds variable for modal table border color
*  Adds nested row to Sink modals to address layout issues
*  Consolidates some helper classes and renames them to be more intuitive

Cluster name: adapter-create

sink


ajainarayanan:
:+1: LGTM

PR #3263, opened 2015-07-23
Remove unnecessary sleep for 2 seconds in various places in MR.

chtyim:
- The sleep was there for waiting metrics be published. However, on
  stopping the metrics collection service, metrics are being force pushed,
  hence the sleep is unnecessary

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapperWrapper.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapperWrapper.java
@@ -110,9 +110,6 @@ public class MapperWrapper extends Mapper {
         ClassLoaders.setContextClassLoader(oldClassLoader);
       }
 
-      // sleep to allow metrics to be written
-      TimeUnit.SECONDS.sleep(2L);
-
       // transaction is not finished, but we want all operations to be dispatched (some could be buffered in
       // memory by tx agent
       try {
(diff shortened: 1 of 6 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapperWrapper.java
```diff
@@ -135,8 +132,11 @@ public void run(Context context) throws IOException, InterruptedException {
       }
 
     } finally {
-      basicMapReduceContext.close();
-      basicMapReduceContext.getMetricsCollectionService().stop();
+      try {
+        basicMapReduceContext.close();
+      } finally {
+        basicMapReduceContext.getMetricsCollectionService().stop();
```
anew: so why does the context not stop the metrics collection service?

anew: I guess it is a bigger change... ok

anew:
LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapperWrapper.java
chtyim: It cannot. There are multiple places that call BasicMapReduceContext.close(), and in some context (e.g. MapReduceRuntimeService in local mode), it shares the same instance as the system metrics collection service. The current code base is pretty messy around creation and closing of the MR context object. It's part of the refactoring tracked by https://issues.cask.co/browse/CDAP-2937

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT340-1

PR #3264, opened 2015-07-23
reverts back to old namespace dropdown and positioning

astewes:
*  Reverts back to previous namespace dropdown positioning and styling

old


cdap-ui/app/styles/themes/cdap/header.less
```diff
@@ -467,4 +432,4 @@ body {
       }
     }
   }
-}
+}
```
anwar6953: please include the newline at the end

ajainarayanan:
:+1: LGTM

PR #3267, opened 2015-07-23
adds new table markup to prevent endpoint path overlap issues

astewes:
*  Changes layout of the endpoint table to prevent endpoint path from overlapping with the Make Request button

path


ajainarayanan:
:+1: LGTM

PR #3268, opened 2015-07-23
Fixes adapter delete to be part of admin (without metadata view)

ajainarayanan:
Fixes adapter delete to be part of admin (part 2 of unifying app and adapters under one view)


astewes:
LGTM

PR #3269, opened 2015-07-24
explore classpath fix 

gsps1:
when hive runs a local mapred task, it cannot find 'org.apache.hadoop.hive.serde2.SerDe' due to missing hive-exec jar in this classpath, adding it to hiveAuxJars and using it to set the LocalMapreduceClasspathSetter classpath

https://issues.cask.co/browse/CDAP-3033
build : http://builds.cask.co/browse/CDAP-RBT343-1/log


gsps1:
closing. accidentally opened against develop.

PR #3270, opened 2015-07-24
explore classpath fix 

gsps1:
when hive runs a local mapred task, it cannot find 'org.apache.hadoop.hive.serde2.SerDe' due to missing hive-exec jar in this classpath, adding it to hiveAuxJars and using it to set the LocalMapreduceClasspathSetter classpath

https://issues.cask.co/browse/CDAP-3033
build : http://builds.cask.co/browse/CDAP-RBT343-1/log

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
@@ -38,6 +38,7 @@ import co.cask.http.HttpHandler;
 import com.google.common.base.Joiner;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.google.common.io.Files;
 import com.google.inject.Exposed;
 import com.google.inject.Inject;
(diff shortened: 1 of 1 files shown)


cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
```diff
@@ -298,12 +299,15 @@ public boolean accept(String className, URL classUrl, URL classPathUrl) {
...
-    List<String> orderedDependenciesStr = builder.build();
+    List<String> orderedDependenciesStr = Lists.newArrayList(builder.build());
 
     System.setProperty(HiveConf.ConfVars.HIVEAUXJARS.toString(), Joiner.on(',').join(orderedDependenciesStr));
     LOG.debug("Setting {} to {}", HiveConf.ConfVars.HIVEAUXJARS.toString(),
               System.getProperty(HiveConf.ConfVars.HIVEAUXJARS.toString()));
 
+    // add hive-exec.jar to the end of auxjars, we need this, otherwise when hive runs a MapRedLocalTask it cannot find
+    // "org.apache.hadoop.hive.serde2.SerDe" class in its classpath.
+    orderedDependenciesStr.add(new HiveConf().getJar());
```
poornachandra: Since `hive-exec.jar` is only required for local mapreduce task, it would be better to not add it to `orderedDependenciesStr`, but add it only while giving the dependencies to `LocalMapreduceClasspathSetter`

cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
```diff
@@ -298,12 +299,15 @@ public boolean accept(String className, URL classUrl, URL classPathUrl) {
     for (File dep : orderedDependencies) {
       builder.add("file://" + dep.getAbsolutePath());
     }
-    List<String> orderedDependenciesStr = builder.build();
+    List<String> orderedDependenciesStr = Lists.newArrayList(builder.build());
```
poornachandra: This should remain immutable. Look at my comment below.

cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
```diff
@@ -298,12 +299,15 @@ public boolean accept(String className, URL classUrl, URL classPathUrl) {
...
 
     System.setProperty(HiveConf.ConfVars.HIVEAUXJARS.toString(), Joiner.on(',').join(orderedDependenciesStr));
     LOG.debug("Setting {} to {}", HiveConf.ConfVars.HIVEAUXJARS.toString(),
               System.getProperty(HiveConf.ConfVars.HIVEAUXJARS.toString()));
 
+    // add hive-exec.jar to the end of auxjars, we need this, otherwise when hive runs a MapRedLocalTask it cannot find
```
poornachandra: We are not adding `hive-exec.jar` to Hive aux jars, but only to the mapreduce classpath, right? The comment should reflect that I think.

gsps1:
@poornachandra i have addressed the comments, please take a look again. Thanks

poornachandra:
LGTM

PR #3271, opened 2015-07-24
CDAP-1873: reduce stack traces from non critical errors on master start

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-RBT345-1

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/DistributedStreamService.java
```diff
@@ -436,7 +436,8 @@ public void onSuccess(ResourceRequirement result) {
 
       @Override
       public void onFailure(Throwable t) {
-        LOG.warn("Failed to update stream resource requirement.", t);
+        LOG.warn("Failed to update stream resource requirement: " + t.getMessage());
```
chtyim: Use the `{}` syntax. `LOG.warn("Failed to .... {}", t.getMessage());`

PR #3272, opened 2015-07-24
CDAP-3031 surface exception causes when datasets cant be created

albertshau:
Changing the dataset create endpoint to include the cause of any
IOException seen while creating the dataset. If there was some
exception instantiating the dataset admin, a generic message will
be included in the response, but the cause will not be shown since
that is a CDAP error and not an error in a dataset admin.


chtyim:
LGTM

albertshau:
http://builds.cask.co/browse/CDAP-RBT348

bdmogal:
This may be unrelated, but should we also switch to the new way of handling HTTP responses in this handler (throwing exceptions), perhaps separate jira + PR?

albertshau:
@bdmogal what's the new way? do the exceptions automatically get translated into the right error code? If so I can remove a couple of those catches.

PR #3273, opened 2015-07-24
[CDAP-1983] Added capability for batch upload to stream in TestBase

bdmogal:
Build: http://builds.cask.co/browse/CDAP-RBT346

### DIFF
--- a/cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
+++ b/cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
@@ -26,9 +26,14 @@ import co.cask.cdap.internal.MockResponder;
 import co.cask.cdap.internal.io.SchemaTypeAdapter;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.test.StreamManager;
+import co.cask.http.BodyConsumer;
 import com.google.common.base.Charsets;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.io.ByteProcessor;
+import com.google.common.io.ByteStreams;
+import com.google.common.io.Files;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
@@ -43,6 +48,7 @@ import org.jboss.netty.handler.codec.http.HttpRequest;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
 import org.jboss.netty.handler.codec.http.HttpVersion;
 
+import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Type;
 import java.nio.ByteBuffer;
@@ -144,6 +150,34 @@ public class DefaultStreamManager implements StreamManager {
     if (responder.getStatus() != HttpResponseStatus.OK) {
       throw new IOException("Failed to write to stream. Status = " + responder.getStatus());
     }
+  }
+
+  @Override
+  public void send(File file, String contentType) throws Exception {
+    String path = String.format("/v3/namespaces/%s/streams/%s/batch", streamId.getNamespaceId(), streamId.getId());
+    HttpRequest request = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.POST, path);
+    request.setHeader(HttpHeaders.Names.CONTENT_TYPE, contentType);
+    final MockResponder responder = new MockResponder();
+    final BodyConsumer bodyConsumer = streamHandler.batch(request, responder, streamId.getNamespaceId(),
+                                                          streamId.getId());
+    Preconditions.checkNotNull(bodyConsumer, "BodyConsumer from stream batch load call should not be null");
+
+    ByteStreams.readBytes(Files.newInputStreamSupplier(file), new ByteProcessor<BodyConsumer>() {
+      @Override
+      public boolean processBytes(byte[] buf, int off, int len) throws IOException {
+        bodyConsumer.chunk(ChannelBuffers.wrappedBuffer(buf, off, len), responder);
+        return true;
+      }
+
+      @Override
+      public BodyConsumer getResult() {
+        bodyConsumer.finished(responder);
+        return bodyConsumer;
+      }
+    });
+
+    Preconditions.checkState(HttpResponseStatus.OK.equals(responder.getStatus()),
+                             "Failed to load events to stream %s in batch", streamId);
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
```diff
@@ -147,6 +157,38 @@ public void send(Map<String, String> headers, ByteBuffer buffer) throws IOExcept
   }
 
   @Override
+  public void send(File file, String contentType) throws Exception {
+    String path = String.format("/v3/namespaces/%s/streams/%s/batch", streamId.getNamespaceId(), streamId.getId());
```
anwar6953: Why not use an implementation similar to StreamClient#sendFile?

bdmogal: StreamClient#sendFile makes HTTP calls, whereas DefaultStreamManager does not start up services, instead makes method calls on handler objects. This is similar to AppFabricClient.deployApplication.

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
```diff
@@ -147,6 +157,36 @@ public void send(Map<String, String> headers, ByteBuffer buffer) throws IOExcept
...
+    HttpRequest request = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.POST, path);
+    byte[] fileContent = Files.readAllBytes(Paths.get(file.toURI()));
```
chtyim: This is bad. Why read the whole file? You are streaming the data to BodyConsumer, there is no point to put the content in the request object.

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
```diff
@@ -147,6 +157,36 @@ public void send(Map<String, String> headers, ByteBuffer buffer) throws IOExcept
...
+    request.setHeader(HttpHeaders.Names.CONTENT_TYPE, contentType);
+    MockResponder responder = new MockResponder();
+    BodyConsumer bodyConsumer = streamHandler.batch(request, responder, streamId.getNamespaceId(), streamId.getId());
+    Preconditions.checkNotNull(bodyConsumer, "BodyConsumer from stream batch load call should not be null");
+
+    Location location = new LocalLocationFactory().create(file.toURI());
+    try (BufferFileInputStream is = new BufferFileInputStream(location.getInputStream(), 1024 * 1024)) {
+      byte[] chunk = is.read();
+      while (chunk.length > 0) {
+        responder = new MockResponder();
```
chtyim: Why create a new responder?

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
```diff
@@ -147,6 +157,36 @@ public void send(Map<String, String> headers, ByteBuffer buffer) throws IOExcept
...
+    request.setContent(content);
+    request.setHeader(HttpHeaders.Names.CONTENT_LENGTH, content.readableBytes());
+    request.setHeader(HttpHeaders.Names.CONTENT_TYPE, contentType);
+    MockResponder responder = new MockResponder();
+    BodyConsumer bodyConsumer = streamHandler.batch(request, responder, streamId.getNamespaceId(), streamId.getId());
+    Preconditions.checkNotNull(bodyConsumer, "BodyConsumer from stream batch load call should not be null");
+
+    Location location = new LocalLocationFactory().create(file.toURI());
+    try (BufferFileInputStream is = new BufferFileInputStream(location.getInputStream(), 1024 * 1024)) {
+      byte[] chunk = is.read();
```
chtyim: This is bad. It may create a new byte[] on every call. Why need to be so complicated? You can simply use a `FileInputStream` with a `byte[]` or use `ByteStreams.readBytes` from guava.

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
```diff
@@ -341,6 +343,20 @@ public void testDeployWorkflowApp() throws Exception {
     Assert.assertEquals(false, Boolean.parseBoolean(workflowToken.getTokenData().get("running").get(0).getValue()));
   }
 
+  @Test
+  public void testBatchStreamUpload() throws Exception {
+    StreamManager batchStream = getStreamManager("batchStream");
+    batchStream.createStream();
+    URL streamDataUrl = Thread.currentThread().getContextClassLoader().getResource("test-stream-data.txt");
```
chtyim: Better not to put a file in resource. You can use Temporary folder and generate a file with the desired content. This makes the file content and the assertion code together.

pom.xml
```diff
@@ -1245,6 +1245,7 @@
                   <exclude>cdap-app-templates/**</exclude>
                   <exclude>**/*.json</exclude>
                   <exclude>archetypes/**/resources/**/*.txt</exclude>
+                  <exclude>**/*.txt</exclude>
```
chtyim: Shouldn't need this if you generate the file for upload test.

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
bdmogal: my bad. I had tried two approaches, this was left over from the approach that didn't work. removed.

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
bdmogal: changed to use ``ByteStreams.readBytes``.

bdmogal:
@chtyim addressed comments. Can you please check again? Especially https://github.com/caskdata/cdap/pull/3273#discussion_r35394468

cdap-unit-test/src/main/java/co/cask/cdap/test/internal/DefaultStreamManager.java
```diff
@@ -147,6 +153,34 @@ public void send(Map<String, String> headers, ByteBuffer buffer) throws IOExcept
...
+    final MockResponder responder = new MockResponder();
+    final BodyConsumer bodyConsumer = streamHandler.batch(request, responder, streamId.getNamespaceId(),
+                                                          streamId.getId());
+    Preconditions.checkNotNull(bodyConsumer, "BodyConsumer from stream batch load call should not be null");
+
+    ByteStreams.readBytes(Files.newInputStreamSupplier(file), new ByteProcessor<BodyConsumer>() {
+      @Override
+      public boolean processBytes(byte[] buf, int off, int len) throws IOException {
+        bodyConsumer.chunk(ChannelBuffers.wrappedBuffer(buf, off, len), responder);
+        return (len - off) > -1;
```
chtyim: It's always true. Just `return true`.

chtyim:
Just one comment left, rest LGTM.

bdmogal:

PR #3276, opened 2015-07-24
CDAP-3039 Fix to make Worker program stop

gokulavasan:
When the run method of the worker completes. This bug is relevant only for inmemory mode and thus no change is required for distributed mode logic

JIRA : https://issues.cask.co/browse/CDAP-3039

Build : http://builds.cask.co/browse/CDAP-RBT347-1

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
@@ -23,6 +23,7 @@ import co.cask.cdap.app.runtime.ProgramOptions;
 import co.cask.cdap.app.runtime.ProgramRunner;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.internal.app.runtime.AbstractListener;
 import co.cask.cdap.internal.app.runtime.AbstractProgramController;
 import co.cask.cdap.internal.app.runtime.BasicArguments;
 import co.cask.cdap.internal.app.runtime.ProgramOptionConstants;
@@ -39,11 +40,13 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.inject.Inject;
 import org.apache.twill.api.RunId;
+import org.apache.twill.common.Threads;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -130,17 +133,36 @@ public abstract class AbstractInMemoryProgramRunner implements ProgramRunner {
     private final ProgramOptions options;
     private final Lock lock = new ReentrantLock();
     private final ProgramRunnerFactory.Type type;
+    private final AtomicLong liveComponents;
 
     public InMemoryProgramController(Table<String, Integer, ProgramController> components,
-                              RunId runId, Program program, ProgramSpecification spec,
-                              ProgramOptions options, ProgramRunnerFactory.Type type) {
+                                     RunId runId, Program program, ProgramSpecification spec,
+                                     ProgramOptions options, ProgramRunnerFactory.Type type) {
       super(program.getName(), runId);
       this.program = program;
       this.components = components;
       this.spec = spec;
       this.options = options;
       this.type = type;
+      this.liveComponents = new AtomicLong(components.size());
       started();
+      monitorComponents();
+    }
+
+    // Add listener to monitor completion/killed status of individual components, so that the program can be marked
+    // as completed once all the components have completed/killed.
+    private void monitorComponents() {
+      for (ProgramController controller : components.values()) {
+        controller.addListener(new AbstractListener() {
+          @Override
+          public void completed() {
+            if (liveComponents.decrementAndGet()  == 0) {
+              LOG.info("Program stopped: {}", spec.getName());
+              complete();
+            }
+          }
+        }, Threads.SAME_THREAD_EXECUTOR);
+      }
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
```diff
@@ -554,6 +554,19 @@ public void testAppWithWorker() throws Exception {
...
+  @Test
+  public void testWorkerStop() throws Exception {
+    // Test to make sure the worker program's status goes to stopped after the run method finishes
+    ApplicationManager manager = deployApplication(NoOpWorkerApp.class);
+    WorkerManager workerManager = manager.getWorkerManager("DummyWorker");
+    try {
+      workerManager.start();
+      workerManager.waitForStatus(false, 5, 1);
+    } finally {
+      manager.stopAll();
```
anwar6953: we don't need this manager.stopAll, right? nothing should be running at that point in time?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
```diff
@@ -130,17 +133,49 @@ private ProgramOptions createComponentOptions(String name, int instanceId, int i
...
+    }
+
+    // Add listener to monitor completion/killed status of individual components, so that the program can be marked
+    // as completed once all the components have completed/killed.
+    private void monitorComponents() {
+      for (ProgramController controller : components.values()) {
+        controller.addListener(new AbstractListener() {
+          @Override
+          public void completed() {
+            super.completed();
```
chtyim: Why need to call super?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
```diff
@@ -130,17 +133,49 @@ private ProgramOptions createComponentOptions(String name, int instanceId, int i
...
+            long value = numWorkerThreads.decrementAndGet();
+            checkThreadCompleted(value);
+          }
+
+          @Override
+          public void killed() {
+            super.killed();
+            long value = numWorkerThreads.decrementAndGet();
+            checkThreadCompleted(value);
```
chtyim: Why need to handle killed? Isn't kill is from external stop through this ProgramController?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
```diff
@@ -130,17 +133,49 @@ private ProgramOptions createComponentOptions(String name, int instanceId, int i
...
+
+    // Add listener to monitor completion/killed status of individual components, so that the program can be marked
+    // as completed once all the components have completed/killed.
+    private void monitorComponents() {
+      for (ProgramController controller : components.values()) {
+        controller.addListener(new AbstractListener() {
+          @Override
+          public void completed() {
+            super.completed();
+            long value = numWorkerThreads.decrementAndGet();
```
chtyim: This class is not Worker specific. Better call it `liveComponents`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
```diff
@@ -130,17 +133,49 @@ private ProgramOptions createComponentOptions(String name, int instanceId, int i
...
+            checkThreadCompleted(value);
```
chtyim: No need to have separate method. Simply be:

```java
public void complete() {
  if (liveComponents.decrementAndGet() == 0) {
    complete();
  }
}
```

gokulavasan:
@chtyim @anwar6953  Addressed comments. Please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
```diff
@@ -130,17 +133,36 @@ private ProgramOptions createComponentOptions(String name, int instanceId, int i
...
+
+    // Add listener to monitor completion/killed status of individual components, so that the program can be marked
+    // as completed once all the components have completed/killed.
+    private void monitorComponents() {
+      for (ProgramController controller : components.values()) {
+        controller.addListener(new AbstractListener() {
+          @Override
+          public void completed() {
+            if (liveComponents.decrementAndGet()  == 0) {
+              LOG.info("Program stopped: {}", spec.getName());
```
chtyim: I thought we already logging it at higher level. Or if not, we should log it at higher level ProgramController/listener.

gokulavasan: We log it in the doStop() method. But it is invoked only we get a stop request for the Worker program? Hence it won't be invoked in the case when all worker threads complete.

chtyim: But the logging here is only specific for programs that has components. What I mean is, it should be log at higher level, e.g. in the AbstractProgramRuntimeSercice:150, which it adds a listener to the program controller.

gokulavasan: @chtyim Ok. Makes sense. Should I make that change in this PR? Looks like other programs log at the ProgramController level (eg, FlowProgramController, FlowletProgramController). Should we move fix them all together?

chtyim: For this PR, make it at the listener added by the AbstractProgramRuntimeService. I don't think you need to make any change the flow controllers.

gokulavasan: @chtyim Added the LOG message in AbstractProgramController since we have the program started message there in started() method. Please take a look and let me know if that is not the right place.

gokulavasan:

chtyim:
LGTM

gokulavasan:

PR #3278, opened 2015-07-24
Stores the position of the nodes once they have been dragged

ajainarayanan:
Adds the ability to store nodes their location when dragged by user. So a refresh on the same page or a draft opened again will have nodes in the same place as it is left.


astewes:
LGTM

PR #3279, opened 2015-07-24
CDAP-2454 Remove twillrunid from proto RunRecord so not send back to client

henry-cask:
Since twillrunid is internal to CDAP it should not be exposed to RunRecord proto which is wired to client:
-) Removed twillRunId field from proto RunRecord
-) Added RunRecordMeta class that extends RunRecord with twillRunId field to work with app meta store.
-) Updated handler and test classes to reflect the changes

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -45,6 +45,7 @@ import co.cask.cdap.internal.app.runtime.flow.FlowUtils;
 import co.cask.cdap.internal.app.runtime.schedule.Scheduler;
 import co.cask.cdap.internal.app.services.ProgramLifecycleService;
 import co.cask.cdap.internal.app.services.PropertiesResolver;
+import co.cask.cdap.internal.app.store.RunRecordMeta;
 import co.cask.cdap.proto.Containers;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.Instances;
@@ -57,11 +58,12 @@ import co.cask.cdap.proto.ProgramStatus;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.RunRecord;
 import co.cask.cdap.proto.ServiceInstances;
-import co.cask.cdap.proto.codec.ScheduleSpecificationCodec;
 import co.cask.http.HttpResponder;
 import com.google.common.base.Charsets;
+import com.google.common.base.Function;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
@@ -164,6 +166,14 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
    */
   private static final Gson GSON = ApplicationSpecificationAdapter.addTypeAdapters(new GsonBuilder()).create();
 
+  private static final Function<RunRecordMeta, RunRecord> CONVERT_TO_RUN_RECORD =
+    new Function<RunRecordMeta, RunRecord>() {
+      @Override
+      public RunRecord apply(RunRecordMeta input) {
+        return new RunRecord(input);
+      }
+    };
+
   /**
    * Store manages non-runtime lifecycle.
    */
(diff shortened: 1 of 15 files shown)


henry-cask:
Build: http://builds.cask.co/browse/CDAP-RBT342-6

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -474,8 +476,11 @@ public void programRunRecord(HttpRequest request, HttpResponder responder,
     }
 
     try {
-      RunRecord runRecord = store.getRun(Id.Program.from(namespaceId, appId, type, programId), runid);
-      if (runRecord != null) {
+      RunRecordMeta runRecordMeta = store.getRun(Id.Program.from(namespaceId, appId, type, programId), runid);
+      if (runRecordMeta != null) {
+        RunRecord runRecord = new RunRecord(runRecordMeta.getPid(), runRecordMeta.getStartTs(),
```
chtyim: Since `RunRecordMeta` extends from `RunRecord`, it's better to have a copy constructor in `RunRecord`. With then you can simple do:

```java
responder.sendJson(HttpResponseStatus.OK, new RunRecord(runRecordMeta));
```

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1424,7 +1429,20 @@ private void getRuns(HttpResponder responder, Id.Program programId, String statu
       try {
         ProgramRunStatus runStatus = (status == null) ? ProgramRunStatus.ALL :
           ProgramRunStatus.valueOf(status.toUpperCase());
-        responder.sendJson(HttpResponseStatus.OK, store.getRuns(programId, runStatus, start, end, limit));
+        List<RunRecord> records =
+          Lists.transform(store.getRuns(programId, runStatus, start, end, limit),
+                          new Function<RunRecordMeta, RunRecord>() {
+                            @Nullable
+                            @Override
+                            public RunRecord apply(@Nullable RunRecordMeta input) {
```
chtyim: It won't be null, right?

chtyim: Also, you can have this function as a static constant. Then you can use it in both `programRunRecord` and `getRuns` methods.

chtyim:
In general looks good. Just some comment on how to make the code more concise.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
henry-cask: Should not be null, but the apply contract have nullable annotation so I added the check.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
henry-cask: sounds good

henry-cask: Actually you could not, because RunRecord is in proto module and RunRecordMeta is in common so I dont think it will work.

chtyim: ? I don't understand. `RunRecord(RunRecord copy)` won't work?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
chtyim: The Function interface has nullable for both return type and the parameter. However, for a specific implementation (like the one in here), if it know for sure it won't be null, you can remove the annotation and no need to do the check.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
henry-cask: You men RunRecord(RunRecordMeta copy) right?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
henry-cask: Ok, will do

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
chtyim: No. I mean `RunRecord(RunRecord)`.. not meta. A copy constructor should copy from itself. Since `RunRecordMeta` is a subclass of `RunRecord`, it should be ok to call that constructor.

henry-cask: Ah ok, got it

henry-cask:
Update the PR based on review

henry-cask:
Build: http://builds.cask.co/browse/CDAP-RBT342-10

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -164,6 +166,14 @@ public void setStatus(String status) {
    */
   private static final Gson GSON = ApplicationSpecificationAdapter.addTypeAdapters(new GsonBuilder()).create();
 
+  private static final Function<RunRecordMeta, RunRecord> CONVERTTORUNRECORD =
```
chtyim: Shouldn't it be `CONVERT_TO_RUN_RECORD`?

chtyim:
Just one comment about the variable name. No need to rerun build.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
henry-cask: I will update the variable name to make it easy on the eyes =)

henry-cask:
Build passes, will squash the commit and merge ...

PR #3281, opened 2015-07-24
Fix/ui adapter save reset

elfenheart:
same as #3275. Opening against ui-adaptercreate-bugfixes branch


ajainarayanan:
:+1: LGTM

PR #3282, opened 2015-07-24
Feature/es sink plugin Allows elasticsearch as a batch sink for an ETL Adapter

ataylor4:
http://builds.cask.co/browse/CDAP-DUT2319-1


chtyim:
LGTM

PR #3283, opened 2015-07-24
Correct syntax for twill.jvm.gc.opts example. 

awholegunch:
Removed all trailing periods from descriptions to make them consistent.


chtyim:
LGTM

PR #3284, opened 2015-07-24
Bugfix/custom dataset embeddeding pfs

anwar6953:
https://issues.cask.co/browse/CDAP-3030
http://builds.cask.co/browse/CDAP-RBT349-6

The issue described in the JIRA is that the dependencies declared by a module using the module partitionedFileSet (pfs) or the module timePartitionedfileSet (tpfs) did not (in pre-3.1 CDAP) include 'core', which the pfs and tpfs now depend on (3.1 and following).

This stage in the upgrade tool modifies any modules that depend on pfs or tpfs to also declare a dependency on 'core'.

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
@@ -200,13 +200,15 @@ public class PFSUpgrader {
   DatasetModuleMeta migrateDatasetModuleMeta(DatasetModuleMeta moduleMeta) {
     List<String> newUsesModules = new ArrayList<>();
     // ordering of the usesModules is important. They are loaded in order of the elements' indices
-    if (!moduleMeta.getUsesModules().contains("orderedTable-hbase")) {
-      newUsesModules.add("orderedTable-hbase");
+    // add the following two modules as the first two dependencies, and then simply exclude them from the existing
+    // declared dependencies to avoid duplicate declared dependencies.
+    newUsesModules.add("orderedTable-hbase");
+    newUsesModules.add("core");
+    for (String usesModule : moduleMeta.getUsesModules()) {
+      if (!"orderedTable-hbase".equals(usesModule) && !"core".equals(usesModule)) {
+        newUsesModules.add(usesModule);
+      }
     }
-    if (!moduleMeta.getUsesModules().contains("core")) {
-      newUsesModules.add("core");
-    }
-    newUsesModules.addAll(moduleMeta.getUsesModules());
     DatasetModuleMeta migratedModuleMeta = new DatasetModuleMeta(moduleMeta.getName(),
                                                                  moduleMeta.getClassName(),
                                                                  moduleMeta.getJarLocation(),
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
@@ -155,8 +155,8 @@ public class PFSUpgrader {
     });
   }
 
-  // need to add 'core' as a module dependency for any modules that currently depend on partitionedFileSet or
-  // timePartitionedFileSet as a resolution to: https://issues.cask.co/browse/CDAP-3030
+  // need to add 'orderedTable-hbase' and 'core' as module dependencies for any modules that currently depend on
+  // partitionedFileSet or timePartitionedFileSet as a resolution to: https://issues.cask.co/browse/CDAP-3030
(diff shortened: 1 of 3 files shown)


anew:
LGTM if tested on cluster

anwar6953:
Added another commit because the order of `usesModules` matters in terms of the order in which the dependencies are loaded.

cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
```diff
@@ -149,8 +153,63 @@ public void apply() throws Exception {
...
+      }
+    }
+    return false;
+  }
+
+  @VisibleForTesting
+  DatasetModuleMeta migrateDatasetModuleMeta(DatasetModuleMeta moduleMeta) {
+    List<String> newUsesModules = new ArrayList<>(moduleMeta.getUsesModules());
+    // core needs to be to the left of pfs or tpfs, because this list is ordered by dependency
+    newUsesModules.add(0, "core");
```
albertshau: more efficient in array list to add this first, then add the rest.

cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
```diff
@@ -149,8 +153,63 @@ public void apply() throws Exception {
...
+          DatasetModuleMeta migratedModuleMeta = migrateDatasetModuleMeta(moduleMeta);
+          dsTypeMDS.write(entry.getKey(), migratedModuleMeta);
+        }
+      }
+    });
+  }
+
+  @VisibleForTesting
+  boolean needsConverting(DatasetModuleMeta moduleMeta) {
+    for (String usesModule : moduleMeta.getUsesModules()) {
```
albertshau: does this catch datasets that depend on a module that depends on pfs or tpfs?

anwar6953: yea, they're all listed here. I don't think the dependency declaration is recursive (all are declared explicitly).

cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
```diff
@@ -149,8 +153,64 @@ public void apply() throws Exception {
...
+  }
 
+  @VisibleForTesting
+  boolean needsConverting(DatasetModuleMeta moduleMeta) {
+    // if 'core' is already a declared dependency, no need to upgrade the module meta
+    List<String> usesModules = moduleMeta.getUsesModules();
+    if (usesModules.contains("core")) {
+      return false;
+    }
+    return usesModules.contains("partitionedFileSet") || usesModules.contains("timePartitionedFileSet");
```
albertshau: technically somebody could have used "partitionedFileSet" or ```PartitionedFileSet.class.getName()```

albertshau: same thing for tpfs

cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
anwar6953: this is the reason we need to add 'core' to modules that depend on pfs/tpfs, even though the pfs/tpfs modules depend on 'core' in their module meta.

cdap-master/src/main/java/co/cask/cdap/data/tools/PFSUpgrader.java
anew: It is looking for a module name, not a dataset type name. The module name is unique

anwar6953: yea, I looked at the dataset.type table, and concluded the same thing as Andreas.

anwar6953:
Also added a dependency on 'orderedTable-hbase'.

anwar6953:
Good point.

anwar6953:
Pushed latest commit with updates.
@anew @anew
Please take another look; thanks.

anew:
LGTM. Please test before merging. This turned out to be trickier than expected...

anwar6953:
@anew @albertshau 
Pushed another commit which resolves an issue which happens when trying to migrate the tables of an embedded pfs. Explanation is in the comment of the commit.

anew:
LGTM after testing :)

anwar6953:
Tested both top-level pfs as well as an embedded tpfs, retrieval of partitions created before the upgrade was succesfull in both cases now.

anew:
merge it!

PR #3287, opened 2015-07-24
Adding ZK dependency as it is required by KafkaSource; 

gokulavasan:
Since the ZK dependency in the cdap was turned into provided scope

Tested in standalone and cluster.

Build : http://builds.cask.co/browse/CDAP-RBT352-1


sreevatsanraman:
LGTM.
Merge the adapters works on standalone and the cluster.

gokulavasan:
Build is :green_apple: 
Merging.

PR #3289, opened 2015-07-24
[CDAP-2981] undo fileset upgrade; add a version to the spec and handle backwards compatibility

anew:
... for absolute base paths


cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -110,7 +109,7 @@
 
     private final String description;
 
-    private Action(String description) {
```
anwar6953: why can't this remain private?

anew: enum constructors are always private. The JLS recommends not to declare them as private because that is redundant. Our checkstyle also warns about that.

anwar6953:
LGTM

anew:
waiting for build before merging: http://builds.cask.co/browse/CDAP-RBT351

anew:
citing @chtyim: Build failed with the known LogAnalysisAppTest only. Since that test is fixed in release, will just merge this one.

PR #3294, opened 2015-07-24
[CDAP-3086] Fixes browser resize to re-draw DAG in adapter create view

ajainarayanan:
Fixes redrawing DAG when window resized. 

resize


elfenheart:
:+1: LGTM

PR #3297, opened 2015-07-24
CDAP-3064 add artifact client

albertshau:
Adding an ArtifactClient so that it can be used for tests and
by the CLI. Also adding some more error handling along the way,
such as artifact name and version validation.


cdap-proto/src/main/java/co/cask/cdap/proto/artifact/ArtifactRange.java
```diff
@@ -96,7 +96,11 @@ public int hashCode() {
 
   @Override
   public String toString() {
-    return new StringBuilder()
+    StringBuilder stringBuilder = new StringBuilder();
+    if (namespace.getId().equals(Id.Namespace.SYSTEM)) {
```
gokulavasan: Why is that we append the namespace only if it is 'system'

cdap-client/src/main/java/co/cask/cdap/client/ArtifactClient.java
```diff
@@ -0,0 +1,346 @@
...
+   * @throws IOException if a network error occurred
+   * @throws UnauthorizedException if the request is not authorized successfully in the gateway server
+   */
+  public void add(Id.Namespace namespace, String artifactName, @Nullable String artifactVersion,
+                  @Nullable Set<ArtifactRange> parentArtifacts, InputSupplier<? extends InputStream> artifactContents)
+    throws ArtifactAlreadyExistsException, BadRequestException, IOException,
+    UnauthorizedException, ArtifactRangeNotFoundException {
+
+    URL url = config.resolveNamespacedURLV3(namespace, String.format("artifacts/%s", artifactName));
+    if (artifactVersion != null) {
```
gokulavasan: What is this if statement for?

gokulavasan:
Couple of questions. Rest LGTM :+1:

cdap-proto/src/main/java/co/cask/cdap/proto/artifact/ArtifactRange.java
albertshau: i think it was because the handler only deals with ranges within its own namespace and with system namespace.  But that logic shouldn't be a part of this class, will add namespace no matter what.

cdap-client/src/main/java/co/cask/cdap/client/ArtifactClient.java
albertshau: ogosh thats not intended. will remove

albertshau:
http://builds.cask.co/browse/CDAP-DUT2454

gokulavasan:
LGTM :+1:

PR #3298, opened 2015-07-24
CDAP-2848 fixing a bug discovered while trying to reproduce the issue reported in CDAP-2848

gokulavasan:
Could not reproduce the issue reported in CDAP-2848 but fixing the new bug discovered. The test was trying to compare JSON string instead of the maps directly.

Build : http://builds.cask.co/browse/CDAP-RBT353-1


cdap-client-tests/src/main/java/co/cask/cdap/client/app/AppReturnsArgs.java
```diff
@@ -49,7 +50,8 @@ public void configure() {
     @Path(ENDPOINT)
     @GET
     public void handler(HttpServiceRequest request, HttpServiceResponder responder) {
-      responder.sendJson(200, getContext().getRuntimeArguments());
+      TreeMap<String, String> values = new TreeMap<>(getContext().getRuntimeArguments());
```
gsps1: why treemap?

gokulavasan: Just a way to sort the map entries by key. If I just make this change and not modify the test, the test will fail more often.

chtyim: I think it's because the unit-test itself has wrong assert. You don't need to use TreeMap here, but rather just modify the unit-test assertion to `Assert.assertEquals(propMap, responseMap);`, where the responseMap can simply decoded as `Map<String, String>`.

gsps1: what is the reason for failure ? because right now in the test case the expected value is of type hashmap.

gokulavasan: @chtyim @gsps1 I will remove the TreeMap. I had changed the assert to compare maps instead of JSON strings.

gokulavasan:
@chtyim @gsps1 Please review the PR. Thanks!

cdap-client-tests/src/main/java/co/cask/cdap/client/app/AppReturnsArgs.java
gokulavasan: @gsps1 The reason for the failure was the comparison was being earlier done on Json strings instead of the Maps. If the order of the keys happened to be the same then the tests used to pass. Put once in a while the order ended up being different and the test failed. So in order to increase the changes of failure, I had added the logic to sort the keys. But removed it after @chtyim comment.

gokulavasan:
@gsps1 This also contains the fix for the actual issue filed in the JIRA. And this also fixes the failure caused here : https://builds.cask.co/browse/CDAP-RBT-JOB1-697/test/case/5558946

cdap-client-tests/src/test/java/co/cask/cdap/client/PreferencesClientTestRun.java
```diff
@@ -130,6 +134,17 @@ public void testProgramAPI() throws Exception {
     }
   }
 
+  private HttpResponse getServiceResponse(URL serviceURL) throws IOException, InterruptedException {
+    int iterations = 0;
+    HttpResponse response;
+    do {
+      TimeUnit.MILLISECONDS.sleep(50);
+      iterations++;
+      response = HttpRequests.execute(HttpRequest.builder(HttpMethod.GET, serviceURL).build());
```
gsps1: better to return if (response == 200) here and do the sleep after that statement. and the while loop will just be 
``while (iterations < 100);

gsps1:
one minor comment. lgtm otherwise

gokulavasan:
@gsps1 Please take a look. Addressed your comment.

gsps1:
LGTM :+1:

gokulavasan:

PR #3299, opened 2015-07-25
improvement to integration testbase : checking if cdap services are u…

gsps1:
…p before running integration tests, adding accessToken if the cluster is secure and also asserting if unrecoverable reset option is set in the cluster

issue : https://issues.cask.co/browse/CDAP-3033
build : http://builds.cask.co/browse/CDAP-RBT354-1

### DIFF
--- a/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
+++ b/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
@@ -22,10 +22,12 @@ import co.cask.cdap.client.ApplicationClient;
 import co.cask.cdap.client.DatasetClient;
 import co.cask.cdap.client.MetaClient;
 import co.cask.cdap.client.MetricsClient;
+import co.cask.cdap.client.MonitorClient;
 import co.cask.cdap.client.NamespaceClient;
 import co.cask.cdap.client.ProgramClient;
 import co.cask.cdap.client.StreamClient;
 import co.cask.cdap.client.config.ClientConfig;
+import co.cask.cdap.client.config.ConnectionConfig;
 import co.cask.cdap.client.util.RESTClient;
 import co.cask.cdap.common.NotFoundException;
 import co.cask.cdap.common.ProgramNotFoundException;
@@ -34,6 +36,7 @@ import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.data2.datafabric.DefaultDatasetNamespace;
 import co.cask.cdap.proto.ApplicationRecord;
+import co.cask.cdap.proto.ConfigEntry;
 import co.cask.cdap.proto.DatasetSpecificationSummary;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.NamespaceMeta;
@@ -41,8 +44,11 @@ import co.cask.cdap.proto.ProgramRecord;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.StreamDetail;
 import co.cask.cdap.security.authentication.client.AccessToken;
+import co.cask.cdap.security.authentication.client.AuthenticationClient;
+import co.cask.cdap.security.authentication.client.basic.BasicAuthenticationClient;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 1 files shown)


cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
 
+  private void checkSystemServices() throws ServiceUnavailableException {
+    MonitorClient monitorClient = getMonitorClient();
+
+    try {
+      checkSystemServicesOk(monitorClient);
+    } catch (UnauthorizedException e) {
+      LOG.info("Caught UnAuthorizedException, Getting Credentials");
+      // security is enabled, we need to get access token before checking system services
+      URL url = null;
```
anwar6953: unused?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
+      try {
+        Properties properties = new Properties();
+        properties.setProperty("security.auth.client.username", System.getProperty("userName"));
+        properties.setProperty("security.auth.client.password", System.getProperty("password"));
+        AuthenticationClient authClient = new BasicAuthenticationClient();
+        authClient.configure(properties);
+        ConnectionConfig connectionConfig = getClientConfig().getConnectionConfig();
+        authClient.setConnectionInfo(connectionConfig.getHostname(), connectionConfig.getPort(), false);
+        AccessToken accessToken = getAccessToken(authClient);
+        System.setProperty("accessToken", accessToken.getValue());
```
poornachandra: It would be a good idea to Gson encode access token here.

poornachandra: It would be better to save the access token and return it in a method called `getAcccessToken()`. Also have a method called `protected AccessToken fetchAccessToken()`, that user can over-ride if required. We can have an implementation that takes in username and password and returns the access token.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
+
+    try {
+      checkSystemServicesOk(monitorClient);
+    } catch (UnauthorizedException e) {
+      LOG.info("Caught UnAuthorizedException, Getting Credentials");
+      // security is enabled, we need to get access token before checking system services
+      URL url = null;
+      try {
+        Properties properties = new Properties();
+        properties.setProperty("security.auth.client.username", System.getProperty("userName"));
```
poornachandra: `userName` can be a single word => `username`

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
+        properties.setProperty("security.auth.client.password", System.getProperty("password"));
```
poornachandra: It might be safer to read username/password from environment variable, since system properties are visible on command line.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
+      } catch (IOException e) {
+        // no-op
+      }
+    } while (numTries <= authClientRetry);
+    throw new ServiceUnavailableException("Unable to connect to AuthServer to obtain AccessToken");
+  }
+
+  private void checkSystemServicesOk(MonitorClient monitorClient) throws UnauthorizedException {
+    // we will wait for upto 10minutes for CDAP services to be up.
+    long maxRetries = TimeUnit.MINUTES.toSeconds(10);
```
poornachandra: We should consolidate all retry time into one and make it a constant. Maybe call it `MAX_REST_WAIT_TIME`, or something.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
+    }
+    LOG.info("CDAP Services are up and running!");
+  }
+
+  private AccessToken getAccessToken(AuthenticationClient authClient) {
+    long authClientRetry = TimeUnit.MINUTES.toSeconds(3);
+    int numTries = 0;
+    do {
+      try {
+        numTries++;
```
poornachandra: Call `numTries` as `numSecs`

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
+        TimeUnit.SECONDS.sleep(1);
```
poornachandra: It would be good to sleep after the call fails.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
+      }
+    } while (numTries <= authClientRetry);
+    throw new ServiceUnavailableException("Unable to connect to AuthServer to obtain AccessToken");
+  }
+
+  private void checkSystemServicesOk(MonitorClient monitorClient) throws UnauthorizedException {
+    // we will wait for upto 10minutes for CDAP services to be up.
+    long maxRetries = TimeUnit.MINUTES.toSeconds(10);
+    int numTries = 0;
+    do {
```
poornachandra: This loop can be refactored to eliminate code duplication.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,110 @@
...
     assertIsClear();
   }
 
+  private void checkSystemServices() throws ServiceUnavailableException {
+    MonitorClient monitorClient = getMonitorClient();
+
+    try {
+      checkSystemServicesOk(monitorClient);
+    } catch (UnauthorizedException e) {
+      LOG.info("Caught UnAuthorizedException, Getting Credentials");
```
poornachandra: `UnAuthorizedException` => `UnauthorizedException`

poornachandra:
@gsps1 Few comments, please fix.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
poornachandra: On second thoughts - let's keep the system properties. We can always set environment variable, and use that to set the system property. System properties are more flexible.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
gsps1: yes. removed.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
gsps1: done, using callable now.

gsps1:
@poornachandra thanks for the review, I have addressed your comments, please take a look again.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -101,10 +199,9 @@ protected String getInstanceURI() {
 
   /**
    * CDAP access token for making requests to secure CDAP instances.
-   * Can be obtained via the CDAP authentication server.
    */
-  protected String getAccessToken() {
-    return System.getProperty("accessToken", "");
+  private AccessToken getAccessToken() {
```
poornachandra: This should be public, so that derived classes can use the access token.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,121 @@
...
+  }
+
+  /**
+   * Uses BasicAuthenticationClient to fetch {@link AccessToken} - this implementation can be overridden if desired.
+   * @return {@link AccessToken}
+   * @throws IOException
+   */
+  protected AccessToken fetchAccessToken() throws IOException {
+    Properties properties = new Properties();
+    properties.setProperty("security.auth.client.username", System.getProperty("username"));
```
poornachandra: It would be a good idea to call the properties `cdap.username` and `cdap.password`, so that it does not clash with any other username/password being used.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,121 @@
...
+      public Boolean call() throws Exception {
+        return authClient.getAccessToken() != null;
+      }
+    }, "Unable to connect to Authentication service to obtain access token");
+
+    return authClient.getAccessToken();
+  }
+
+
+  private void checkServicesWithRetry(Callable<Boolean> callable,
```
poornachandra: Isn't there some method that already does this kind of check and wait?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -53,29 +59,121 @@
...
+    authClient.configure(properties);
+    ConnectionConfig connectionConfig = getClientConfig().getConnectionConfig();
+    authClient.setConnectionInfo(connectionConfig.getHostname(), connectionConfig.getPort(), false);
+
+    checkServicesWithRetry(new Callable<Boolean>() {
+      @Override
+      public Boolean call() throws Exception {
+        return authClient.getAccessToken() != null;
+      }
+    }, "Unable to connect to Authentication service to obtain access token");
```
poornachandra: It would be a good idea to print the connection info here, it will help in debugging.

poornachandra:
@gsps1 Few more comments.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
gsps1: i am not sure if we want to allow that though, in most cases the derived classes would do ``getConnectionConfig`` and that will set accessToken.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
gsps1: https://github.com/caskdata/cdap/blob/develop/cdap-common/src/main/java/co/cask/cdap/common/utils/Tasks.java#L46
-- this does repeated checks based on callable, but they throw any exception, while here we want to throw only UnauthorizedException and skip on other exceptions.

gsps1: should we move this method over there ?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
gsps1: on second thoughts, it can be useful to have this as public, since ClientConfig builder is public as well.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
poornachandra: Since this exception handling logic is very specific to integration test, let's keep it here for now.

poornachandra:
LGTM

gsps1:
squashed commits - build green - merging.

PR #3300, opened 2015-07-25
Feature/ui styling adapters group sidepanel

astewes:
*  Updates styling of group side panels, plumb shapes and metadata box per the mocks
*  Adds conditional tooltip (but some of the logic still needs to be completed)

cluster name: adapter-create

To do:
*  [x] Add metadata description to white box
*  [x] Complete tooltip-enable logic for item headers (#3308)

adapter


elfenheart:
:+1: LGTM

PR #3303, opened 2015-07-26
Fix/ui adapter minor bugfixes

ajainarayanan:
- Fixes empty spaces to be reset on all properties
- Removes '(' from input schema
- Fixes modal close behavior. (no outside click close and or ESC button close)
- Reduce the size of 'Loading' text in properties modal.
- Fixes Adapter DAG's line-width to be smaller
- Fixes settings modal to have instances/cron schedule


elfenheart:
:+1: LGTM

PR #3304, opened 2015-07-26
Fixes url, alignment & messaging in adapters list view to be drafts

ajainarayanan:
- Fixes the url for adapter drafts,
- Fixes title and alignment
- Fixes the message to be shown when no drafts have been created


elfenheart:
:+1: LGTM

PR #3305, opened 2015-07-27
Added parentNodIds to the WorkflowNode.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-3000

Following changes are done as a part of the PR:
1. Added set of parentNodeIds to the each WorkflowNode, which points to the previous nodes in the Workflow. 
2. While doing the `get` operation on the WorkflowToken, we create the filter list of node Ids with the help of parentNodeIds set at each node. Only values from this filter list are returned.
3. While forking in the Workflow, we make copies of the `BasicWorkflowToken`. Instead of making deep copy of it, we allow sharing of the `tokenValueMap` instance inside it across different branches, so that same map gets updated by different nodes. 
4. Added unit test cases.


sagarkapare:
@chtyim @bdmogal @anew Please help review. Thanks.

sagarkapare:
Closing this PR. will open new against the release branch.

PR #3309, opened 2015-07-27
(CDAP-2899) Make MR job runner in local mode use a per run directory …

chtyim:
…so that it gets cleanup properly after job is done.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
@@ -177,32 +177,30 @@ final class MapReduceRuntimeService extends AbstractExecutionThreadService {
 
   @Override
   protected void startUp() throws Exception {
-    final Job job = createJob();
-    Configuration mapredConf = job.getConfiguration();
-
-    classLoader = new MapReduceClassLoader(context.getProgram().getClassLoader());
-    mapredConf.setClassLoader(new WeakReferenceDelegatorClassLoader(classLoader));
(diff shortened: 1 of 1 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -177,32 +177,30 @@ protected String getServiceName() {
...
-    Configuration mapredConf = job.getConfiguration();
+    // Creates a temporary directory locally for storing all generated files.
+    File tempDir = createTempDirectory();
+    this.cleanupTask = createCleanupTask(tempDir);
 
-    classLoader = new MapReduceClassLoader(context.getProgram().getClassLoader());
-    mapredConf.setClassLoader(new WeakReferenceDelegatorClassLoader(classLoader));
-    ClassLoaders.setContextClassLoader(mapredConf.getClassLoader());
+    try {
+      Job job = createJob(new File(tempDir, "mapreduce"));
```
bdmogal: ``programType`` is already part of the ``tempDir``, do we need a subdirectory called "mapreduce" too?

chtyim: Yes. This is to separate out the directory used by CDAP vs directory used by the MapReduce framework.

bdmogal: :+1:

bdmogal:
LGTM! Thanks @chtyim !

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT356

PR #3310, opened 2015-07-27
[CDAP-3092] Mapper reads in multiple files

jayfjin:
JIRA - https://issues.cask.co/browse/CDAP-3092
BUILD - http://builds.cask.co/browse/CDAP-DUT2333-1

Allows each mapper in FileBatchSource to read in multiple files. Previous behavior would generate at least one mapper per file. This solves the many small files problem, in which a large number of mappers would cause slow runtime and eventually OOM issues.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
@@ -38,6 +38,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.mapreduce.Job;
+import org.apache.hadoop.mapreduce.lib.input.CombineTextInputFormat;
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -86,12 +87,15 @@ public class FileBatchSource extends BatchSource<LongWritable, Object, Structure
   private static final String TABLE_DESCRIPTION = "Name of the Table that keeps track of the last time files " +
     "were read in.";
   private static final String INPUT_FORMAT_CLASS_DESCRIPTION = "Name of the input format class, which must be a " +
-    "subclass of FileInputFormat. Defaults to TextInputFormat";
+    "subclass of FileInputFormat. Defaults to CombineTextInputFormat";
   private static final String FILESYSTEM_DESCRIPTION = "Distributed file system to read in from.";
+  private static final String MAX_SPLIT_SIZE_DESCRIPTION = "Max split size for each mapper in the MapReduce Job. " +
+    "Defaults to 128MB.";
   private static final Gson GSON = new Gson();
   private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
   private static final Type ARRAYLIST_DATE_TYPE  = new TypeToken<ArrayList<Date>>() { }.getType();
   private static final Type MAP_STRING_STRING_TYPE = new TypeToken<Map<String, String>>() { }.getType();
+  private static final int DEFAULT_SPLIT_SIZE = 134217728;
 
   private final FileBatchConfig config;
   private KeyValueTable table;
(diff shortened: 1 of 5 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/SqsSource.java
```diff
@@ -44,8 +44,8 @@
  * TODO: CDAP-2978: Extend JMS source so this class can be deleted.
  */
 @Plugin(type = "source")
-@Name("Sqs")
-@Description("Sqs Realtime Source - Emits a record with a field 'body' of string type")
+@Name("AmazonSqs")
```
gokulavasan: Sqs or SQS?

dtzeng: i think all caps would be better style

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -216,14 +229,21 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
     @Description(INPUT_FORMAT_CLASS_DESCRIPTION)
     private String inputFormatClass;
 
+    @Name("maxSplitSize")
+    @Nullable
+    @Description(MAX_SPLIT_SIZE_DESCRIPTION)
+    private String maxSplitSize;
+
```
bdmogal: extra blank line

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -158,9 +162,18 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
         classLoader.loadClass(config.inputFormatClass);
       job.setInputFormatClass(classType);
+    } else {
+      job.setInputFormatClass(CombineTextInputFormat.class);
     }
     FileInputFormat.setInputPathFilter(job, BatchFileFilter.class);
     FileInputFormat.addInputPath(job, new Path(config.path));
+    long maxSplit;
+    try {
+      maxSplit = Long.valueOf(config.maxSplitSize);
```
bdmogal: ``maxSplit`` is a primitive ``long``. So you don't need ``valueOf``, which returns a ``new Long()``. Use ``Long.parseLong`` instead.

bdmogal: Also,  can you call it maxSplitSize, since that is what it is, its not the split itself.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/source/FileBatchSource.java
```diff
@@ -86,12 +87,15 @@
...
-    "subclass of FileInputFormat. Defaults to TextInputFormat";
+    "subclass of FileInputFormat. Defaults to CombineTextInputFormat";
   private static final String FILESYSTEM_DESCRIPTION = "Distributed file system to read in from.";
+  private static final String MAX_SPLIT_SIZE_DESCRIPTION = "Max split size for each mapper in the MapReduce Job. " +
+    "Defaults to 100KB.";
   private static final Gson GSON = new Gson();
   private static final Logger LOG = LoggerFactory.getLogger(FileBatchSource.class);
   private static final Type ARRAYLIST_DATE_TYPE  = new TypeToken<ArrayList<Date>>() { }.getType();
   private static final Type MAP_STRING_STRING_TYPE = new TypeToken<Map<String, String>>() { }.getType();
+  private static final int DEFAULT_SPLIT_SIZE = 102400;
```
bdmogal: According to [this|https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/input/FileInputFormat.java#L216], the default for the param ``mapreduce.input.fileinputformat.split.maxsize`` is ``Long.MAX_VALUE``. I think for a realistic usecase, the default you have here is very low. We should probably have it atleast as the default size of 1 block on HDFS - 64 (now 128 in most cases) MB.

bdmogal:
LGTM, pending build. Please squash commits before merging.

PR #3311, opened 2015-07-27
Added parentNodIds to the WorkflowNode.

sagarkapare:
Opening it against release/3.1 branch.

JIRA: https://issues.cask.co/browse/CDAP-3000

Following changes are done as a part of the PR:
1. Added set of parentNodeIds to the each WorkflowNode, which points to the previous nodes in the Workflow. 
2. While doing the get operation on the WorkflowToken, we create the filter list of node Ids with the help of parentNodeIds set at each node. Only values from this filter list are returned.
3. While forking in the Workflow, we make copies of the BasicWorkflowToken. Instead of making deep copy of it, we allow sharing of the tokenValueMap instance inside it across different branches, so that same map gets updated by different nodes. 
4. Added unit test cases.


sagarkapare:
@chtyim @bdmogal @anew Please help review. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowSpecification.java
```diff
@@ -45,6 +46,43 @@ public WorkflowSpecification(String className, String name, String description,
...
+  }
+
+  /**
+   * Visit all the nodes in the {@link Workflow} and generate the map of node id to
+   * the {@link WorkflowNode}. Returned map does not contain any mapping for the FORK node,
+   * however contains the ACTION and CONDITION nodes within the FORK node itself.
+   */
+  private Map<String, WorkflowNode> generateNodeIdMap() {
+    Map<String, WorkflowNode> nodeIdMap = new HashMap<>();
+    Queue<WorkflowNode> nodes = new LinkedList<>(this.nodes);
```
chtyim: Don't use `this` since this method is called from constructor. Have the List of nodes passed to this method.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowSpecification.java
```diff
@@ -45,6 +46,43 @@ public WorkflowSpecification(String className, String name, String description,
...
+    while (!nodes.isEmpty()) {
+      WorkflowNode node = nodes.poll();
+      switch (node.getType()) {
+        case ACTION: {
+          nodeIdMap.put(node.getNodeId(), node);
+          break;
+        }
+        case FORK: {
+          WorkflowForkNode forkNode = (WorkflowForkNode) node;
```
chtyim: Why not add the fork node to the map as well? I think it's better for the list and map contain the same set of nodes.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -38,8 +42,10 @@
   private static final long serialVersionUID = -1173500180640174909L;
 
   private Map<String, Map<String, Long>> mapReduceCounters;
-  private final Map<Scope, Map<String, List<NodeValue>>> tokenValueMap = new EnumMap<>(Scope.class);
+  private Map<Scope, Map<String, List<NodeValue>>> tokenValueMap =
+    Collections.synchronizedMap(new EnumMap<Scope, Map<String, List<NodeValue>>>(Scope.class));
```
chtyim: Is it enough to only have the top level map synchronized? The HashMap inside as well as the list can be updated concurrently as well, right?

chtyim: And in fact, from the current logic, the top level map doesn't need to be synchronized since you put the values in the constructor and never add new scopes afterward.

chtyim: And why `final` is removed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -38,8 +42,10 @@
...
   private String nodeName;
+  private List<String> nodeFilterList;
```
chtyim: Why not final and compute it in constructor? And shouldn't it be a `Set`?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -122,22 +122,24 @@ void put(String key, Value value, Scope scope) {
     Preconditions.checkNotNull(value.toString(), String.format("Null value provided for the key '%s'.", key));
     Preconditions.checkState(nodeName != null, "nodeName cannot be null.");
 
-    List<NodeValue> nodeValueList = tokenValueMap.get(scope).get(key);
-    if (nodeValueList == null) {
-      nodeValueList = Lists.newArrayList();
-      tokenValueMap.get(scope).put(key, nodeValueList);
-    }
+    synchronized (tokenValueMap) {
```
chtyim: It's a bad practice to synchronize on object that you don't own the class.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -122,22 +122,24 @@ void put(String key, Value value, Scope scope) {
...
+      for (int i = 0; i < nodeValueList.size(); i++) {
+        if (nodeValueList.get(i).getNodeName().equals(nodeName)) {
+          nodeValueList.set(i, new NodeValue(nodeName, value));
+          return;
+        }
       }
-    }
 
-    nodeValueList.add(new NodeValue(nodeName, value));
+      nodeValueList.add(new NodeValue(nodeName, value));
```
chtyim: The above for loop will do replacement, but then the same value will get appended to the list as well? Meaning you'll have a duplicated value?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowSpecification.java
sagarkapare: Fixed.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowSpecification.java
sagarkapare: Added fork node to the map.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Updated the code - 
1. While creating the entry in the enum map, now adding the instance of the `ConcurrentHashMap` as a value.
2. Creating the `Collections.synchronizedList` for the NodeValue list.
3. Created private static member in the BasicWorkflowToken, which is now used to synchronized where we iterate over the list.
4. final is in place now. I removed it because in the copy constructor, I was assigning the `tokenValueMap` to the `other.tokenValueMap` in order to use the same instance. However we can still keep the outer map different and have the same instance of the inner maps.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Yes. Updated to the `set`. However its not final because the same instance of the `BasicWorkflowToken` is passed by driver to the different nodes before the execution (except fork). WorkflowDriver keep calling the `setCurrentNodeInfo` to update the `nodeName` and the `nodeFilterSet`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Updated the code to use the private static member variable of the `BasicWorkflowToken` for synchronization. Let me know if this is not what you meant.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: We `return` from the loop if the value replacement is done, right?

sagarkapare:
@chtyim Addressed comments, plus some questions. Please take a look again. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowSpecification.java
```diff
@@ -45,6 +46,42 @@ public WorkflowSpecification(String className, String name, String description,
...
+   * the {@link WorkflowNode}.
+   */
+  private Map<String, WorkflowNode> generateNodeIdMap(List<WorkflowNode> nodesInWorkflow) {
+    Map<String, WorkflowNode> nodeIdMap = new HashMap<>();
+    Queue<WorkflowNode> nodes = new LinkedList<>(nodesInWorkflow);
+    while (!nodes.isEmpty()) {
+      WorkflowNode node = nodes.poll();
+      nodeIdMap.put(node.getNodeId(), node);
+      switch (node.getType()) {
+        case ACTION: {
```
gsps1: you can skip the `{` for cases.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -74,28 +78,30 @@ void setCurrentNode(String nodeName) {
...
-            if (thisNodeValue.equals(otherNodeValue)) {
-              otherNodeValueExist = true;
-              break;
+          // Iterate over the list of NodeValue corresponding to the current key.
+          // Only add those NodeValue to the merged token which already do not exist.
+
+          for (NodeValue otherNodeValue : otherTokenValueMapForScopeEntry.getValue()) {
+            boolean otherNodeValueExist = false;
+            for (NodeValue thisNodeValue :
+              thisTokenValueMapForScope.get(otherTokenValueMapForScopeEntry.getKey())) {
```
gsps1: can simplify by creating a set of ``thisTokenValueMapForScope.get(otherTokenValueMapForScopeEntry.getKey())`` and just add if its not present in the set.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -122,22 +128,24 @@ void put(String key, Value value, Scope scope) {
     Preconditions.checkNotNull(value.toString(), String.format("Null value provided for the key '%s'.", key));
     Preconditions.checkState(nodeName != null, "nodeName cannot be null.");
 
-    List<NodeValue> nodeValueList = tokenValueMap.get(scope).get(key);
-    if (nodeValueList == null) {
-      nodeValueList = Lists.newArrayList();
-      tokenValueMap.get(scope).put(key, nodeValueList);
-    }
+    synchronized (sharedLock) {
+      List<NodeValue> nodeValueList = tokenValueMap.get(scope).get(key);
```
gsps1: generally we don't include types in variable names right? nodeValues would be a better name.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowSpecification.java
sagarkapare: Skipped.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Using `set` now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: I think in this case its better to have List in the name for the readability.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -36,36 +43,34 @@
...
-      Map<String, List<NodeValue>> tokenValueMapForScope = new HashMap<>();
-      for (Map.Entry<String, List<NodeValue>> valueEntry : entry.getValue().entrySet()) {
-        tokenValueMapForScope.put(valueEntry.getKey(), Lists.newArrayList(valueEntry.getValue()));
-      }
-
-      this.tokenValueMap.put(entry.getKey(), tokenValueMapForScope);
+    // Same instance of the tokenValueMap is shared across all the
+    // nodes in the Workflow, so that values can be put concurrently.
+    for (Scope scope : Scope.values()) {
+      tokenValueMap.put(scope, other.tokenValueMap.get(scope));
```
chtyim: Better not to copy but just do the assignment directly. `this.tokenValueMap = other.tokenValueMap;`

sagarkapare:
Making some changes to the PR. Please hold off on the review. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionNode.java
```diff
@@ -28,13 +29,21 @@
   private final WorkflowActionSpecification actionSpecification;
 
   public WorkflowActionNode(String nodeId, ScheduleProgramInfo program) {
```
anew: Why does this API have no Javadocs?

anew: Has @awholegunch reviewed the Javadocs in this package? I think they need an overhaul

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowNode.java
```diff
@@ -16,16 +16,24 @@
...
 /**
  * Represents node in the {@link Workflow}.
  */
 public abstract class WorkflowNode {
   protected final String nodeId;
   protected final WorkflowNodeType nodeType;
+  protected final Set<String> parentNodeIds;
 
-  public WorkflowNode(String nodeId, WorkflowNodeType nodeType) {
+  public WorkflowNode(String nodeId, WorkflowNodeType nodeType, @Nullable Set<String> parentNodeIds) {
```
anew: when would this be null? Why not require that it is the empty set if there are no parents? (That would only be the case for the first node in a workflow, right?)

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowForkNode.java
```diff
@@ -19,13 +19,13 @@
...
 /**
- * Represents the FORK node in the {@link Workflow}
+ * Represents the FORK node in the {@link Workflow}.
  */
 public class WorkflowForkNode extends WorkflowNode {
   private final List<List<WorkflowNode>> branches;
 
   public WorkflowForkNode(String nodeId, List<List<WorkflowNode>> branches) {
-    super(nodeId, WorkflowNodeType.FORK);
+    super(nodeId, WorkflowNodeType.FORK, null);
```
anew: A fork never has parent nodes?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -36,36 +43,34 @@
...
   private BasicWorkflowToken(BasicWorkflowToken other) {
-    for (Map.Entry<Scope, Map<String, List<NodeValue>>> entry : other.tokenValueMap.entrySet()) {
-      Map<String, List<NodeValue>> tokenValueMapForScope = new HashMap<>();
-      for (Map.Entry<String, List<NodeValue>> valueEntry : entry.getValue().entrySet()) {
-        tokenValueMapForScope.put(valueEntry.getKey(), Lists.newArrayList(valueEntry.getValue()));
-      }
-
-      this.tokenValueMap.put(entry.getKey(), tokenValueMapForScope);
+    // Same instance of the tokenValueMap is shared across all the
+    // nodes in the Workflow, so that values can be put concurrently.
```
anew: this comment is inconsistent with the code. Comment say "same instance of tokenValueMap is shared:, but the code copies the all values. What exactly do you mean?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionNode.java
sagarkapare: We have javadoc for the class. I added the javadoc to the constructors in the `WorkflowActionNode` as well. Is that what you meant? If so will add it to other node's constructors as well.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowNode.java
sagarkapare: When we configure the Workflow, we do not add the parent ids and they are null. Parent id's are added during generating the specification for the Workflow.

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowForkNode.java
sagarkapare: `fork` is artificial node in the Workflow, which does not do any processing. So the parent id is not added to the fork. Parent ids of the first node in the fork branches contains the previous node of the fork.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Revised the code. So as per the current logic, we always make a deep copy of the WorkflowToken before starting the execution of the ACTION/CONDITION. Put operations happen on the copied token. The copied token is merged back to the original token after the execution of the node completes.

sagarkapare:
@anew @chtyim Updated the PR. Please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -40,63 +48,63 @@
...
-  void mergeToken(BasicWorkflowToken other) {
+  void mergeToken(final String nodeName, BasicWorkflowToken other) {
     for (Map.Entry<Scope, Map<String, List<NodeValue>>> entry : other.tokenValueMap.entrySet()) {
-      Map<String, List<NodeValue>> thisTokenValueMapForScope = this.tokenValueMap.get(entry.getKey());
 
-      for (Map.Entry<String, List<NodeValue>> otherTokenValueMapForScopeEntry : entry.getValue().entrySet()) {
-        if (!thisTokenValueMapForScope.containsKey(otherTokenValueMapForScopeEntry.getKey())) {
-          thisTokenValueMapForScope.put(otherTokenValueMapForScopeEntry.getKey(), Lists.<NodeValue>newArrayList());
-        }
+      ConcurrentHashMap<String, List<NodeValue>> thisTokenValueMapForScope =
```
anew: Why do you cast here? If this class makes explicit use of concurrent features, then the type should be declared as ConcurrentMap, not Map

anew:
The program logic appears correct to me. I do believe, though, that the code could be streamlined a bit. Since we don't have time for that before release, I am ok with merging this and refactoring it after release. Please also wait for @chtyim 's final review, as he has had more context on this from the start.

sagarkapare:
Thank you @anew for looking into it :) . yes will wait for LGTM from @chtyim !

sreevatsanraman:
@anew @chtyim - if this is not the quality we expect. Let's merge this in 3.2 time frame that doing it now and fixing later.

chtyim:
@sagarkapare To clarify, this bug only affects the RunRecord being stored in the MDS, but won't affect the correctness of the workflow execution, right?

cdap-api/src/main/java/co/cask/cdap/api/workflow/WorkflowActionNode.java
```diff
@@ -27,14 +28,44 @@
   private final ScheduleProgramInfo program;
   private final WorkflowActionSpecification actionSpecification;
 
+  /**
+   * Constructs the ACTION node in the {@link Workflow} for MapReduce or Spark program.
+   * @param nodeId the id with which the node is to be created
```
awholegunch: There should be an empty line (with a "*") before the first "@param" tag in all Javadocs.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+          parentNodeIds.add(conditionNode.getNodeId());
+        }
+        break;
+      default:
+        throw new IllegalStateException("Node type is invalid.");
+    }
+  }
 
-    return new WorkflowConditionNode(conditionNodeId, conditionNode.getPredicateClassName(), ifbranch, elsebranch);
+  // This method determines if the node is immediate child of the CONDITION node.
```
awholegunch: the immediate child

sagarkapare:
@chtyim Yes. it affects how the RunRecord being stored in the MDS inside a fork, but won't affect the correctness of the Workflow execution.

So if the fork is the end node in the Workflow persisted token will be in the inconsistent state. 

One workaround can be applied is, we can call `store.updateWorkflowToken` at the end of fork execution, so that token gets correctly persisted at the end of the fork and we can revisit the approach to fix this in 3.2

Let me know if thats ok. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+  // For e.g. consider the following workflow where c1 is the condition node.
+  // If branch of the condition node executes a fork with actions a1, a3, and a4
```
awholegunch: The 'If' branch....The 'Else' branch...

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+  // in parallel. Else branch of the condition execute a5. Node a6 gets executed after
+  // the If/Else branch of the condition finishes the execution.
```
awholegunch: finishes execution.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+  //
+  //               T |--a1---a2--|
+  //            |----|-----a3----|----|
+  //      c1----|    |-----a4----|    |---a6
+  //            |----------a5---------|
+  //               F
+  //
+  // For nodes a1, a3, 14, and a5 are considered to be immediate children of the c1 and method
```
awholegunch: Nodes a1, a3, a4, and a5 are considered to be immediate children of c1 and this method

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+  // returns true for them, so that parent set of a1, a3, a4, and a5 will contain only c1.
```
awholegunch: the parent set

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+  // For node a6, even its previous node is c1, since its not the immediate child of c1, method
```
awholegunch: For node a6, though its previous node is c1, since it's not an immediate child of c1, the method

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+
+    Set<String> parentNodeIds = new HashSet<>();
+    populateParentNodeIds(previousNode, parentNodeIds, currentNode);
+    return new WorkflowConditionNode(conditionNodeId, conditionNode.getPredicateClassName(), ifbranch, elsebranch,
+                                     parentNodeIds);
+  }
+
+  // This method populates the set parentNodeIds with the parent node ids of the current node.
+  // If the previous node is of type ACTION, then that node is considered as the parent of the current node.
+  // If the previous node if of type FORK, then the last nodes on all the branches of FORK are considered as
```
awholegunch: If the previous node is of type FORK,

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+  // the parent of the current node.
+  // If the previous node is of type CONDITION and if the node is immediate child of the CONDITION, then the
```
awholegunch: and if the node is an immediate child of the CONDITION

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/DefaultWorkflowConfigurer.java
```diff
@@ -124,26 +133,139 @@ private WorkflowNode createNodeWithId(WorkflowNode node) {
...
+  // CONDITION node is considered to be parent of the current node, otherwise last nodes on the if branch and
```
awholegunch: to be the parent of the current node; otherwise, the last nodes on the 'if' branch and
'else' branches are considered to be the parent of the current node.

sagarkapare:
This PR is blocked by refactoring the way Workflow starts the programs. Will close this for now and re-open later.

PR #3313, opened 2015-07-27
Fixes arrow head, thickness and color in my-plumb directive

ajainarayanan:
Fixes the arrow styles (thickness, color and arrow head)

### DIFF
--- a/cdap-ui/app/directives/plumb/my-plumb-factory.js
+++ b/cdap-ui/app/directives/plumb/my-plumb-factory.js
@@ -5,10 +5,10 @@ angular.module(PKG.name + '.commons')
       ConnectionsDetachable: true
     };
     var connectorStyle = {
-      strokeStyle: 'black',
-      fillStyle: '#666',
+      strokeStyle: "#666e82",
+      fillStyle: '#666e82',
       radius: 7,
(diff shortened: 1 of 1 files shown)


astewes:
LGTM

cdap-ui/app/directives/plumb/my-plumb-factory.js
```diff
@@ -16,11 +16,11 @@ angular.module(PKG.name + '.commons')
       maxConnections: -1,
       paintStyle: {
         strokeStyle: 'white',
-        fillStyle: '#666',
+        fillStyle: '#666e82',
```
elfenheart: Is it possible to use variable already defined in variable.less?

ajainarayanan: I don't believe we can use the variable defined in a less file in javascript file. We don't ship less files right.. Any suggestions to do it in a straight forward way apart from querying through css selectors ?

elfenheart: lol, my bad...didn't notice it's in JS :grin:

elfenheart:
:+1: LGTM

PR #3317, opened 2015-07-27
Do not quote OPTS

wolf31o2:
Quoting OPTS only allows a single Java property to be set. This is
fine with the default configuration of CDAP, but fails quite
spectacularly if the user specifies their own variables in the
cdap-env.sh file:

export OPTS="${OPTS} -Dhdp.version=2.2.4.2-2"

Our scripts interpret OPTS as a single parameter, and pass it to
the "java" binary as such, causing Java to not recognise this now
combined parameter. Removing the quotes allows them to be interpreted
correctly by Bash, and passed correctly to Java.


poornachandra:
LGTM

PR #3318, opened 2015-07-27
updates adapter icons and style selectors

astewes:
*  Updates font files with latest icomoon custom fonts
*  Changes style selectors to account for both font awesome and icomoon fonts

adapter name: adapter-create


ajainarayanan:
:+1: LGTM

PR #3320, opened 2015-07-27
Fix Security Configuration example

awholegunch:
Change the example XML to be correct XML, and highlight the replacements required.

Fix for https://issues.cask.co/browse/CDAP-2838


jwang47:
LGTM

PR #3322, opened 2015-07-27
CDAP-2498: Test Source has three modes - stream, table, default. 

gokulavasan:
Logic should go to default mode if the type is not set to stream or table

Build : http://builds.cask.co/browse/CDAP-RBT376-1

JIRA : https://issues.cask.co/browse/CDAP-2498


sreevatsanraman:
LGTM :+1:

gokulavasan:
Build is :green_apple: Merging.

PR #3324, opened 2015-07-27
(CDAP-3096) Make MapReduce and Workflow in distributed mode aware

chtyim:
of the mapreduce.application.framework.path setting.

This PR has three commits and is organized like this:

1. Just a simple class rename, not really related to CDAP-3096, but to make class name following the same convention
2. The actual logic for the fix to make use of mapreduce framework.
3. It's a hack for working around TWILL-144. It is set to be removed in future release one TWILL-144 is resolved (tracked by CDAP-3119).

Advise to review commit 1 and 2 only. For the third commit, it's been tested on various clusters with or without usage of mapreduce framework settings.


anew:
Not everything here is obvious, but after we discussed, it makes sense. 
Provided that this is tested on the various distros we support, LGTM

chtyim:
Test passed: http://builds.cask.co/browse/CDAP-RBT374-2

PR #3326, opened 2015-07-27
changes layout of adapter metadata box

astewes:
*  Replaces floats from metadata box
*  Sets min-width on metadata box
*  Adds placeholder description for metadata description
*  Adds myEllipsis to adapter name

metadata


elfenheart:
:+1: LGTM

PR #3328, opened 2015-07-28
[CDAP-3087] fix the updating of namespace when getList is called

elfenheart:
cluster: secure2245-1000


ajainarayanan:
:+1: LGTM

PR #3329, opened 2015-07-28
fixes padding issue affecting Settings dropdown

astewes:
*  Adds white background to entire dropdown toggle
*  Fixes padding issue affecting focus state of Settings dropdown
*  Adds caret to Settings dropdown on smaller devices
*  Caret now transitions/rotates -90deg when dropdowns are expanded

caret

### DIFF
--- a/cdap-ui/app/directives/navbar/navbar.html
+++ b/cdap-ui/app/directives/navbar/navbar.html
@@ -58,6 +58,7 @@
     <li class="dropdown">
       <a href="" class="dropdown-toggle">
         <span class="fa fa-cog"></span>
+        <span class="caret"></span>
       </a>
     </li>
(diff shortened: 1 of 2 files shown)


ajainarayanan:
@astewes Still needs fix for smaller screen size. If you reduce the browser size the settings menu item is not aligned with the namespace dropdown.

ajainarayanan:
Talked to @astewes . Had some comments about the settings menu on smaller screen. 
- Should show arrow on namespace dropdown on toggle
- Should use the same arrow on 'settings' menu on smaller screens.
- The settings menu should not shift to left & right when we open up namespace dropdown.

elfenheart:
:+1: LGTM

PR #3330, opened 2015-07-28
Fixes arrow placement attached to the end point

ajainarayanan:
- Fixes arrows to be attached to the tip of the endpoint in adapter DAG
- Hides the close button on a plugin once it is created


elfenheart:
:+1: LGTM

PR #3331, opened 2015-07-28
[CDAP-2922] adds metrics to datasets obtained through DynamicDatasetContext

anew:
This affects: workers, mapreduce, spark (driver side). 
- adds a metrics context to the DynamicDatasetContext
- changes to program contexts to pass in the metrics context
- additions to some test cases to validate that the metrics are emitted with correct context
- new test case WorkerProgramRunnerTest - we did not have a test case for workers before

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
@@ -31,6 +31,10 @@ import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
 import co.cask.cdap.api.dataset.lib.PartitionedFileSetArguments;
 import co.cask.cdap.api.dataset.lib.TimePartitionedFileSet;
 import co.cask.cdap.api.dataset.lib.TimePartitionedFileSetArguments;
+import co.cask.cdap.api.dataset.lib.cube.AggregationFunction;
+import co.cask.cdap.api.metrics.MetricDataQuery;
+import co.cask.cdap.api.metrics.MetricStore;
+import co.cask.cdap.api.metrics.MetricTimeSeries;
 import co.cask.cdap.app.program.Program;
 import co.cask.cdap.app.runtime.ProgramController;
 import co.cask.cdap.app.runtime.ProgramRunner;
@@ -77,6 +81,8 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
@@ -97,6 +103,7 @@ public class SparkProgramRunnerTest {
   private static TransactionManager txService;
   private static DatasetFramework dsFramework;
   private static DatasetInstantiator datasetInstantiator;
+  private static MetricStore metricStore;
 
   final String testString1 = "persisted data";
   final String testString2 = "distributed systems";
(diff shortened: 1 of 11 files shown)


anew:
Build: http://builds.cask.co/browse/CDAP-RBT359

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
```diff
@@ -155,6 +163,24 @@ public void testSparkWithObjectStore() throws Exception {
     prepareInputData();
     runProgram(app, SparkAppUsingObjectStore.CharCountSpecification.class);
     checkOutputData();
+
+    // validate that the table emitted metrics
+    Collection<MetricTimeSeries> metrics =
+      metricStore.query(new MetricDataQuery(
```
abaranau: it's time to have some builders.. :) We have them for Cube, but not for MetricsStore (which was used mostly internally so far). Created https://issues.cask.co/browse/CDAP-3112

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetContext.java
```diff
@@ -72,16 +76,28 @@
...
    * @param name the name of the dataset
+   *
    * @return runtime arguments for the given dataset
    */
   protected Map<String, String> getRuntimeArguments(String name) {
     return RuntimeArguments.extractScope(Scope.DATASET, name, runtimeArguments);
   }
 
+  /**
+   * @param namespace the {@link Id.Namespace} in which the transaction context is created
```
abaranau: transaction context is created in namespace? is this smth new?

abaranau: also, the tx context is passed in, and the javadoc says it is created inside..

cdap-notifications/src/main/java/co/cask/cdap/notifications/service/BasicNotificationContext.java
```diff
@@ -51,7 +51,7 @@ public boolean execute(TxRunnable runnable, TxRetryPolicy policy) {
         final TransactionContext context = new TransactionContext(transactionSystemClient);
         try {
           context.start();
-          runnable.run(new DynamicDatasetContext(namespaceId, context, dsFramework,
+          runnable.run(new DynamicDatasetContext(namespaceId, context, null, dsFramework,
```
abaranau: why is it null here?

abaranau:
couple Qs

cdap-notifications/src/main/java/co/cask/cdap/notifications/service/BasicNotificationContext.java
anew: Because there is no metrics context here. At least I am not sure what the context should be.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetContext.java
anew: I simply copied this from the other constructor, because this one was missing the javadocs. Not sure what is meant by this. Perhaps it should say, "in which the dataset context is created"

cdap-notifications/src/main/java/co/cask/cdap/notifications/service/BasicNotificationContext.java
anew: This is a system service that currently has no metric context. I think it goes beyond the scope of this PR to change that. Opened https://issues.cask.co/browse/CDAP-3114

anew:
@abaranau I answered your questions. Please take another look.

cdap-notifications/src/main/java/co/cask/cdap/notifications/service/BasicNotificationContext.java
abaranau: Got it - please add a todo with link to this jira

abaranau:
one nit and LGTM after that

PR #3332, opened 2015-07-28
CDAP-3070: Failure to configure one Plugin JAR, should not affect the configuration of other JARs

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-RBT360-1

JIRA : https://issues.cask.co/browse/CDAP-3070


sreevatsanraman:
LGTM

PR #3336, opened 2015-07-28
[CDAP-3098] add tooltip to input schema

elfenheart:
add tooltip if input schema name is too long
modify modals columns for source, sink, transforms

tooltipsinput


ajainarayanan:
Didn't we talk about resizing the layouts so that all three, input schema, properties & output schema occupy the same grid size? We can show the tooltips but I feel this fix is incomplete if we don't correct the layout. Thoughts?

elfenheart:
@ajainarayanan I agree with you. Let's put this on-hold

ajainarayanan:
:+1: LGTM

PR #3337, opened 2015-07-28
Feature/ui cherry pick namespace

elfenheart:
Cherry pick #3264 to revert namespace inside adapters


ajainarayanan:
Shouldn't the right fix be us rebasing adapter-create-prototype branch on top of release/3.1? Any specific reason why had to cherry-pick those commits if we know that adapter-create-prototype is eventually going to be merged to release/3.1?

PR #3339, opened 2015-07-28
CDAP-2895 Throw exception when the put operation on the WorkflowToken…

sagarkapare:
… is performed from Mappre/Reducer classes.

JIRA: https://issues.cask.co/browse/CDAP-2895

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
@@ -25,6 +25,7 @@ import co.cask.cdap.PauseResumeWorklowApp;
 import co.cask.cdap.WorkflowAppWithErrorRuns;
 import co.cask.cdap.WorkflowAppWithFork;
 import co.cask.cdap.WorkflowAppWithScopedParameters;
+import co.cask.cdap.WorkflowTokenTestPutApp;
 import co.cask.cdap.api.schedule.ScheduleSpecification;
 import co.cask.cdap.api.workflow.WorkflowActionNode;
 import co.cask.cdap.api.workflow.WorkflowActionSpecification;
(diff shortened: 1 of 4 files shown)


sagarkapare:
@bdmogal @chtyim @anew Please help review. Thanks!

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -1197,4 +1198,84 @@ private String appendScopeAndKeyToUrl(String workflowTokenUrl, @Nullable Workflo
...
+    // Start program with "put.in.mapper.initialize" argument.
+    // It will perform put operation on the WorkflowToken in the Initialize method of the Mapper class.
+    // This should fail.
+    String outputPath = new File(tmpFolder.newFolder(), "output").getAbsolutePath();
+    startProgram(workflowId, ImmutableMap.of("inputPath", createInputForRecordVerification("firstInput"),
+                                             "outputPath", outputPath, "put.in.mapper.initialize", "true"));
+    waitState(workflowId, ProgramRunStatus.RUNNING.name());
+    waitState(workflowId, "STOPPED");
+
+    List<RunRecord> programRuns = getProgramRuns(workflowId, ProgramRunStatus.FAILED.name());
```
anew: you wait for state STOPPED but then validate that it is actually FAILED. Does that ever work?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -1197,4 +1198,84 @@ private String appendScopeAndKeyToUrl(String workflowTokenUrl, @Nullable Workflo
...
+    // This should succeed.
+    outputPath = new File(tmpFolder.newFolder(), "output").getAbsolutePath();
+    startProgram(workflowId, ImmutableMap.of("inputPath", createInputForRecordVerification("fifthInput"),
+                                             "outputPath", outputPath));
+    waitState(workflowId, ProgramRunStatus.RUNNING.name());
+    waitState(workflowId, "STOPPED");
+
+    programRuns = getProgramRuns(workflowId, ProgramRunStatus.COMPLETED.name());
+    Assert.assertEquals(1, programRuns.size());
+  }
```
anew: Do you have test for putting from the beforeSubmit() or the onFinish()? Those should be allowed, I believe.

anew:
Hi @sagarkapare mostly looks good, but I had to comments. Thanks -Andreas

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: `waitState` queries the `/status` api, which checks for the program in the in-memory runtimeInfo object. If the program is not found, it returns `STOPPED`, otherwise it returns `RUNNING`. When the program completes/fails, it is removed from the runTimeInfo and in the `RunRecord` we mark the appropriate status as COMPLETED/FAILED etc.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
sagarkapare: Yes. We allow putting from beforeSubmit() and onFinish(). Added test case for the put operation from onFinish method in the new commit. We are already performing put operation in the beforeSubmit method - https://github.com/caskdata/cdap/pull/3339/files#diff-6d5c552cd9d1a2ee73692a5f4aa9c8f0R91

sagarkapare:
Thank you @anew for the review comments. Please take a look again.

sagarkapare:
Release build running here - http://builds.cask.co/browse/CDAP-RBT365-1

sagarkapare:
Build Passed. @anew Please take a look again. Thanks!

anew:
LGTM

PR #3340, opened 2015-07-28
Update the system services restart endpoints to be more RESTful

henry-cask:
Follow up on endpoint for system service restart instances comments after demo.

Adding concept of "restart" resource to the /system/services/{service-name} endpoint.
The restart path represent a request to restart instances on the system services.

So to restart all instances of a system service need to send request to:
PUT /system/services/{service-name}/restart

and to restart particular instance id via:
PUT /system/services/{service-name}/restart/{instance-id}

Modify the GET endpoint for latest restart status by adding concept of latest-restart resource:
GET /system/services/{service-name}/latest-restart

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
@@ -108,7 +108,7 @@ public class MonitorHandlerTest extends AppFabricTestBase {
 
     Assert.assertEquals(HttpResponseStatus.OK.getCode(), response.getStatusLine().getStatusCode());
 
-    path = String.format("%s/system/services/%s/lastrestartstatus", Constants.Gateway.API_VERSION_3,
+    path = String.format("%s/system/services/%s/latest-restart", Constants.Gateway.API_VERSION_3,
                          Constants.Service.APP_FABRIC_HTTP);
     response = doGet(path);
 
@@ -131,7 +131,7 @@ public class MonitorHandlerTest extends AppFabricTestBase {
 
     Assert.assertEquals(HttpResponseStatus.BAD_REQUEST.getCode(), response.getStatusLine().getStatusCode());
 
-    path = String.format("%s/system/services/%s/lastrestartstatus", Constants.Gateway.API_VERSION_3,
+    path = String.format("%s/system/services/%s/latest-restart", Constants.Gateway.API_VERSION_3,
                          Constants.Service.APP_FABRIC_HTTP);
     response = doGet(path);
 
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
@@ -102,22 +102,21 @@ public class MonitorHandlerTest extends AppFabricTestBase {
 
   @Test
   public void testRestartInstances() throws Exception {
-    String path = String.format("system/services/%s/instances/restart", Constants.Service.APP_FABRIC_HTTP);
-    HttpURLConnection urlConn = openURL(path, HttpMethod.PUT);
+    String path = String.format("%s/system/services/%s/restart", Constants.Gateway.API_VERSION_3,
+                                Constants.Service.APP_FABRIC_HTTP);
+    HttpResponse response = doPut(path);
 
-    Assert.assertEquals(HttpResponseStatus.OK.getCode(), urlConn.getResponseCode());
+    Assert.assertEquals(HttpResponseStatus.OK.getCode(), response.getStatusLine().getStatusCode());
 
-    urlConn.disconnect();
+    path = String.format("%s/system/services/%s/lastrestartstatus", Constants.Gateway.API_VERSION_3,
+                         Constants.Service.APP_FABRIC_HTTP);
+    response = doGet(path);
 
-    urlConn = openURL(path, HttpMethod.GET);
-
-    Assert.assertEquals(HttpResponseStatus.OK.getCode(), urlConn.getResponseCode());
+    Assert.assertEquals(HttpResponseStatus.OK.getCode(), response.getStatusLine().getStatusCode());
 
     RestartServiceInstancesStatus result =
-      GSON.fromJson(new String(ByteStreams.toByteArray(urlConn.getInputStream()), Charsets.UTF_8),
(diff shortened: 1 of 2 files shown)


henry-cask:
Successful build: http://builds.cask.co/browse/CDAP-RBT363-1

anew:
Can we do POST .../services/<name>/restart for all instances and POST .../services/<name>/instances/<n>/restart for a particular one?

henry-cask:
Thanks for the review @anew , I have updated the PR to reflect the comments.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
```diff
@@ -102,23 +102,22 @@ public void testInstances() throws Exception {
 
   @Test
   public void testRestartInstances() throws Exception {
-    String path = String.format("system/services/%s/instances/restart", Constants.Service.APP_FABRIC_HTTP);
-    HttpURLConnection urlConn = openURL(path, HttpMethod.PUT);
-
-    Assert.assertEquals(HttpResponseStatus.OK.getCode(), urlConn.getResponseCode());
+    String path = String.format("%s/system/services/%s/restart", Constants.Gateway.API_VERSION_3,
```
anew: is there a test case for restarting a single instance, and then getting the status of that restart?

anew:
@henry-cask please see my comments

henry-cask:
@anew yes, it is covered by the testInvalidIdRestartInstances.

henry-cask:
@anew To be complete, I added new test testSingleIdRestartInstances to test successful restart single instance.

henry-cask:
New build: http://builds.cask.co/browse/CDAP-RBT363-4

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
```diff
@@ -102,22 +102,44 @@ public void testInstances() throws Exception {
...
+
+    Assert.assertEquals(HttpResponseStatus.OK.getCode(), response.getStatusLine().getStatusCode());
+
+    path = String.format("%s/system/services/%s/latest-restart", Constants.Gateway.API_VERSION_3,
+                         Constants.Service.APP_FABRIC_HTTP);
+    response = doGet(path);
+
+    Assert.assertEquals(HttpResponseStatus.OK.getCode(), response.getStatusLine().getStatusCode());
+
+    RestartServiceInstancesStatus result =
```
anew: RestartServiceInstancesStatus is an ugly class name. Is it marked as @Beta? It should because as you mentioned, we have plans to replace this with a more systematic API.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
```diff
@@ -102,22 +102,44 @@ public void testInstances() throws Exception {
...
+      GSON.fromJson(new String(ByteStreams.toByteArray(response.getEntity().getContent()), Charsets.UTF_8),
+                    RestartServiceInstancesStatus.class);
 
     Assert.assertNotNull(result);
     Assert.assertEquals(Constants.Service.APP_FABRIC_HTTP, result.getServiceName());
```
anew: Does this object contain the id the of the instances that were restarted? If so, you should validate that. Also are there other fields of interest? The time of restart? The duration of the the restart? Where do you test that these fields are properly set?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
```diff
@@ -102,22 +102,44 @@ public void testInstances() throws Exception {
...
 
     RestartServiceInstancesStatus result =
-      GSON.fromJson(new String(ByteStreams.toByteArray(urlConn.getInputStream()), Charsets.UTF_8),
+      GSON.fromJson(new String(ByteStreams.toByteArray(response.getEntity().getContent()), Charsets.UTF_8),
                     RestartServiceInstancesStatus.class);
 
-    urlConn.disconnect();
+    Assert.assertNotNull(result);
+    Assert.assertEquals(Constants.Service.APP_FABRIC_HTTP, result.getServiceName());
+    Assert.assertEquals(RestartServiceInstancesStatus.RestartStatus.SUCCESS, result.getStatus());
```
anew: same as below applies here.

anew: Is there any way you can test that the service was actually restarted?

henry-cask: It is covered in Twill test

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
henry-cask: Let's address it in another PR, I need to get this in to update the doc

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
henry-cask: Let's address it in another PR because this one just to cover API change and need to be in before doc PR could be sent

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
anew: I don't think Twill tests the CDAP services.

anew:
@henry-cask OK, please merge this and address the testing in a new PR.

henry-cask:
Thanks @anew. 
Will send new PR immediately

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/MonitorHandlerTest.java
henry-cask: Ah ok, no it does not.
However since it is internally just delegate to Twill to restart the instances, currently we do not check it. We also do not have tests to test whether add and reduce new instances for service actually restarted.
Looks like good candidate to add those.

henry-cask:
Build passes, merging ...

PR #3341, opened 2015-07-28
[CDAP-2258] some refactoring to return better HTTP codes

anew:
Build passed here

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -28,6 +28,7 @@ import co.cask.cdap.app.program.Programs;
 import co.cask.cdap.app.runtime.ProgramController;
 import co.cask.cdap.app.runtime.ProgramRuntimeService;
 import co.cask.cdap.app.store.Store;
+import co.cask.cdap.common.ApplicationNotFoundException;
 import co.cask.cdap.common.BadRequestException;
 import co.cask.cdap.common.NotFoundException;
 import co.cask.cdap.common.ProgramNotFoundException;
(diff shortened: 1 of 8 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -368,11 +369,18 @@ public void performAction(HttpRequest request, HttpResponder responder,
     }
 
     if (!isValidAction(action)) {
-      responder.sendStatus(HttpResponseStatus.METHOD_NOT_ALLOWED);
+      responder.sendStatus(HttpResponseStatus.NOT_FOUND);
```
chtyim: I think now we throw exception for error response. E.g. `throw new NotFoundException(...)` in this case. There is a error handler that will turn known exception into corresponding response code and respond.

anew: I was following the existing code. The error handler is - as far as I understand - a fallback for exception thrown by methods further in the call stack. That does not prevent us from explicitly sending a response, does it?

chtyim: No it does not. Looks like we have some mixed style of error handling in this class:

https://github.com/caskdata/cdap/blob/feature/cdap-2258-better-errors/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java#L231

vs

https://github.com/caskdata/cdap/blob/feature/cdap-2258-better-errors/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java#L257

anew:
@chtyim I have added a test case.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anew: I have tried to keep the refactoring here small. The entire class (and some others - ProgramLifecycleService, DefaultStore) have a wild mix of different error handling styles. This needs more refactoring but that would go beyond a bugfix and I don't want to do this so close to a release.See https://issues.cask.co/browse/CDAP-3107

anew:
@chtyim any other concerns?

chtyim:
LGTM

PR #3342, opened 2015-07-28
CDAP-3104 enabling explore on a Table without schema should no-op

albertshau:
Fixes a bug where a request to enable exploration on a Table
without a schema would error. This is required because a call to
enable exploration is made on every single dataset when it is
created or when its properties change. All this dataset specific
logic in explore should be moved out, but that is outside the
scope of this fix and has its own open issue.

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
@@ -162,7 +162,7 @@ public class ExploreTableManager {
     // there are two ways to refer to each dataset type...
     if (ObjectMappedTableModule.FULL_NAME.equals(datasetType) ||
       ObjectMappedTableModule.SHORT_NAME.equals(datasetType)) {
-      return createFromSchemaProperty(spec, datasetID, serdeProperties);
+      return createFromSchemaProperty(spec, datasetID, serdeProperties, true);
     }
 
     try (SystemDatasetInstantiator datasetInstantiator = datasetInstantiatorFactory.create()) {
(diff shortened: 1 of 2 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-RBT368

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
```diff
@@ -220,15 +221,19 @@ public QueryHandle enableDataset(Id.DatasetInstance datasetID, DatasetSpecificat
...
+                                               Map<String, String> serdeProperties, boolean shouldErrorOnMissingSchema)
     throws ExploreException, SQLException, UnsupportedTypeException {
 
     String schemaStr = spec.getProperty(DatasetProperties.SCHEMA);
     // if there is no schema property, we cannot create the table and this is an error
     if (schemaStr == null) {
-      throw new IllegalArgumentException(String.format(
-        "Unable to enable exploration on dataset %s because the %s property is not set.",
-        datasetID.getId(), DatasetProperties.SCHEMA));
+      if (shouldErrorOnMissingSchema) {
```
bdmogal: nit: Should we use ``Preconditions`` here?

bdmogal:
1 small nit, else LGTM

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
albertshau: preconditions should be used to check conditions that should always be true.  this is more like bad user input. It should probably actually be throwing BadRequestException, but leaving that exception refactoring to another PR.

bdmogal: fair enough. :+1:

PR #3344, opened 2015-07-28
CDAP-2768 filter out system.queue from list of datasets

albertshau:
system.queue is a system dataset, but it exists in each namespace
once a flow has been started in the namespace. It should not be
visible via the get datasets rest call. This is a short term
fix, with the longer term fix to have support for system datasets
in namespaces other than the system namespace.


sreevatsanraman:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-RBT367

PR #3345, opened 2015-07-28
update tephra version to 0.6.0, remove snapshot dependency

gsps1:


chtyim:
Do we still need this? https://github.com/caskdata/cdap/blob/feature/tephra-version-update/pom.xml#L66

gsps1:
i was able to build it with forced update after removing that snapshot, so i think we can remove that.

chtyim:
LGTM

PR #3349, opened 2015-07-28
Misc log fixes

poornachandra:
This contains the following fixes -
1. Simplify log file range calculation, also fixes a bug in rage calculation.
2. Handle a case where no logs were returned when logs in Kafka were expired for the latest run.

It would be easier to review commit by commit.

JIRA - https://issues.cask.co/browse/CDAP-2935


poornachandra:
Build - http://builds.cask.co/browse/CDAP-RBT357-5

sreevatsanraman:
One minor comment. LGTM otherwise.

poornachandra:
Added the test case. Will merge after new build. 

http://builds.cask.co/browse/CDAP-RBT357-6

PR #3350, opened 2015-07-28
changes selector for plugin modal styling

astewes:
*  Modifies CSS selector for plugin modals to prevent unwanted styling of the down modal

down


ajainarayanan:
:+1: LGTM

PR #3353, opened 2015-07-28
CDAP-2765 Remove CLI security warning

wolf31o2:
The security warning removal is in the second commit. The first commit is white-space only.


jwang47:
LGTM

PR #3354, opened 2015-07-28
CDAP-2765 Remove CLI security warnings

wolf31o2:
This is the same as #3353 but against `release/3.1`

### DIFF
--- a/cdap-common/src/main/resources/cdap-default.xml
+++ b/cdap-common/src/main/resources/cdap-default.xml
@@ -1080,14 +1080,8 @@
     </property>
 
     <property>
-      <name>security.auth.server.address</name>
-      <value></value>
-      <description>Deprecated. Use security.auth.server.bind.address instead.</description>
-    </property>
-
-    <property>
-      <name>security.auth.server.bind.address</name>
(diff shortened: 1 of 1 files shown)


jwang47:
LGTM

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -1080,14 +1080,8 @@
...
-      <name>security.auth.server.address</name>
-      <value></value>
-      <description>Deprecated. Use security.auth.server.bind.address instead.</description>
-    </property>
-
-    <property>
-      <name>security.auth.server.bind.address</name>
-      <value>0.0.0.0</value>
+        <name>security.auth.server.bind.address</name>
+        <value>0.0.0.0</value>
```
awholegunch: In the docs, we currently give 127.0.0.1 as the default; which should it be?
See
http://docs-staging.cask.co/cdap/3.1.0-SNAPSHOT/en/admin-manual/installation/cdap-site.html

jwang47: Default is 0.0.0.0.

PR #3355, opened 2015-07-28
CDAP-2565: Use gauge for tx histogram metrics

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-RBT371-1

chtyim:
LGTM

ghelmling:
+1, better than summing the individual values

PR #3356, opened 2015-07-28
Updated zookeeper dependency and edited kafka properties to be schema…

ataylor4:
… and format

http://builds.cask.co/browse/CDAP-RBT372-1

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
@@ -166,6 +166,11 @@
     <!-- End for JMS Source -->
     <!-- Start for Kafka Source -->
     <dependency>
+      <groupId>org.apache.zookeeper</groupId>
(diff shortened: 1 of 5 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/pom.xml
```diff
@@ -166,6 +166,11 @@
     <!-- End for JMS Source -->
     <!-- Start for Kafka Source -->
     <dependency>
+      <groupId>org.apache.zookeeper</groupId>
+      <artifactId>zookeeper</artifactId>
+      <scope>compile</scope>
```
gokulavasan: Except for the scope, the other change should be in release branch. During conflict if any, make sure the compile scope option is included

gokulavasan:
Backend changes LGTM :+1:

ataylor4:
same as https://github.com/caskdata/cdap/pull/3363; closed so that it could be merged into release

PR #3357, opened 2015-07-28
[CDAP-2996] Moves pause/stop buttons for workflows near runid dropdown

ajainarayanan:
- Moves pause/stop buttons near the runid
- Removes the text from the buttons
- Removes 1000 interval time from pollRunDetail api of workflows.

TODO:
  We need a way for the aggregator in the client to get multiple request and send only one message to node proxy for any polling. Currently if the same page has 3 different sections polling for the same resource we make 3 different polls and this brings down the UI. We should fix this in the immediate next release.
worfkflowpausestop


elfenheart:
change color of buttons to match mock:

screen shot 2015-07-28 at 5 15 56 pm
screen shot 2015-07-28 at 5 16 01 pm

elfenheart:
:+1: LGTM on @ajainarayanan changes

PR #3362, opened 2015-07-29
make docs target paths project generic

duddun:
This change allows us to remove 'cdap' from the bamboo global variable for staging and docs paths and make the paths generic (using $PROJECT which are already deriving from pom file in this script).

Note: This change will also be backported to release 3.1, 3.0 and 2.8

This CANNOT be merged until we disable all three docs build plans temporarily, backport and merge PRs for all four branches, and modify the two corresponding bamboo global variables.


wolf31o2:
:+1: LGTM

PR #3363, opened 2015-07-29
Updated zookeeper dependency and edited kafka properties to be schema…

ataylor4:
...and format on release branch

https://github.com/caskdata/cdap/pull/3356 without front-end changes and branched from release

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/KafkaSource.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/KafkaSource.java
@@ -65,8 +65,8 @@ public class KafkaSource extends RealtimeSource<StructuredRecord> {
   public static final String KAFKA_ZOOKEEPER = "kafka.zookeeper";
   public static final String KAFKA_BROKERS = "kafka.brokers";
   public static final String KAFKA_DEFAULT_OFFSET = "kafka.default.offset";
-  public static final String KAFKA_SCHEMA = "kafka.schema";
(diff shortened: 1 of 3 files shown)


gokulavasan:
LGTM :+1:

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/KafkaSource.java
```diff
@@ -65,8 +65,8 @@
   public static final String KAFKA_ZOOKEEPER = "kafka.zookeeper";
   public static final String KAFKA_BROKERS = "kafka.brokers";
   public static final String KAFKA_DEFAULT_OFFSET = "kafka.default.offset";
-  public static final String KAFKA_SCHEMA = "kafka.schema";
-  public static final String KAFKA_FORMAT = "kafka.format";
+  public static final String KAFKA_SCHEMA = "schema";
```
hsaputra: If the constant value changed, then it is preferable to also change the variable name to match it

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/realtime/source/KafkaSource.java
```diff
@@ -65,8 +65,8 @@
...
+  public static final String KAFKA_FORMAT = "format";
```
hsaputra: If the constant value changed, then it is preferable to also change the variable name to match it

hsaputra:
Small nit on variable name, other than that LGTM

PR #3366, opened 2015-07-29
[CDAP-3077] Added another config parameter in Table Sink to allow cas…

bdmogal:
…e-insensitive find for row key field in the incoming record's schema.

Also:
- Refactored TableSink, RealtimeTableSink and TableSource to share their PluginConfig class.
- Added a delete() method to AdapterManager for cleanup in tests. Still not implemented in RemoteAdapterManager.

Build: https://builds.cask.co/browse/CDAP-RBT366
Jira: [CDAP-3077](https://issues.cask.co/browse/CDAP-3077)

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
@@ -179,7 +179,7 @@ public class BatchETLDBAdapterTest extends BaseETLBatchTest {
   @Test
   @Category(SlowTests.class)
   @SuppressWarnings("ConstantConditions")
-  public void testDBSource() throws Exception {
+  public void testDBSourceWithCaseInsensitiveRowKey() throws Exception {
     String importQuery = "SELECT ID, NAME, SCORE, GRADUATED, TINY, SMALL, BIG, FLOAT_COL, REAL_COL, NUMERIC_COL, " +
       "DECIMAL_COL, BIT_COL, DATE_COL, TIME_COL, TIMESTAMP_COL, BINARY_COL, BLOB_COL, CLOB_COL FROM my_table " +
       "WHERE ID < 3";
@@ -196,7 +196,9 @@ public class BatchETLDBAdapterTest extends BaseETLBatchTest {
     ETLStage sink = new ETLStage("Table", ImmutableMap.of(
       "name", "outputTable",
       Properties.Table.PROPERTY_SCHEMA, schema.toString(),
-      Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "ID"));
+      Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "id",
+      // Test case-insensitive row key matching
+      Properties.Table.CASE_SENSITIVE_ROW_FIELD, "false"));
 
     ETLBatchConfig etlConfig = new ETLBatchConfig("* * * * *", source, sink, Lists.<ETLStage>newArrayList());
     AdapterConfig adapterConfig = new AdapterConfig("", TEMPLATE_ID.getId(), GSON.toJsonTree(etlConfig));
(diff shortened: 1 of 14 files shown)


bdmogal:
@chtyim This is per the discussion on [CDAP-3077](https://issues.cask.co/browse/CDAP-3077). The only thing not implemented here is the optimization you suggested in your last comment on that PR. The problem is that we need to look for the ``schema.row.field`` in the incoming ``StructuredRecord`` from the database. Did you mean in the comment that we should cache it after getting the first record, and ignore for subsequent records?

bdmogal:
Also, this PR is pending testing with Netezza adapter. As I do not have access to it, @dave-finnegan is testing it for me. Will not merge until that test is certified.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/template/etl/batch/BatchETLDBAdapterTest.java
```diff
@@ -196,7 +196,9 @@ public void testDBSource() throws Exception {
     ETLStage sink = new ETLStage("Table", ImmutableMap.of(
       "name", "outputTable",
       Properties.Table.PROPERTY_SCHEMA, schema.toString(),
-      Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "ID"));
+      Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "id",
+      // Test case-insensitive row key matching
+      Properties.Table.CASE_INSENSITIVE_ROW_FIELD, "true"));
```
albertshau: how come this is just for the row field?

albertshau: this approach seems weird, so I have to name my columns accordingly, but not the row field? kind of defeats the overall goal right?

albertshau: ok nevermind, its because the table doesn't take a schema, just the row field

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/RecordPutTransformer.java
```diff
@@ -102,36 +117,54 @@ private void setField(Put put, Schema.Field field, Object val) {
...
+        @Override
+        public boolean apply(Schema.Field input) {
+          return input.getName().equalsIgnoreCase(rowField);
+        }
+      });
+      if (!Iterables.isEmpty(filtered)) {
+        Preconditions.checkArgument(Iterables.size(filtered) == 1, "Cannot have multiple fields in the schema that " +
+                                      "match %s in a case-insensitive manner when %s is enabled. Found %s.", rowField,
+                                    Properties.Table.CASE_INSENSITIVE_ROW_FIELD, Iterables.toString(filtered));
+        field = filtered.iterator().next();
```
albertshau: you could at this point set rowField to whatever you found.  Even if for some weird reason you get another record with a different casing, the search logic will still work to find it again.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/TableConfig.java
```diff
@@ -0,0 +1,71 @@
...
+    "created with this schema, which will allow the table to be explored through Hive.")
+  @Nullable
+  private String schemaStr;
+
+  @Name(Properties.Table.PROPERTY_SCHEMA_ROW_FIELD)
+  @Description("The name of the record field that should be used as the row key when writing to the table.")
+  private String rowField;
+
+  @Name(Properties.Table.CASE_INSENSITIVE_ROW_FIELD)
+  @Description("Allows case-insensitive search for schema.row.field in the schema. Disabled by default. Set this to" +
```
albertshau: 'search' may be confusing for people, since search implies some knowledge about the implementation. It's also probably easier to think a property called case sensitive. case sensitive = true is simpler than double negative case insensitive = false.  So maybe "Whether or not schema.row.field is case sensitive, defaults to true."

albertshau:
logic looks good, one comment about an optimization and one about property naming.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/RecordPutTransformer.java
```diff
@@ -102,36 +117,54 @@ private void setField(Put put, Schema.Field field, Object val) {
...
+  private Put createPut(StructuredRecord record, Schema.Field keyField) {
+    String keyFieldName = keyField.getName();
+    Object val = record.get(keyFieldName);
     Preconditions.checkArgument(val != null, "Row key cannot be null.");
 
     Schema.Type keyType = validateAndGetType(keyField);
     switch (keyType) {
       case BOOLEAN:
-        return new Put(Bytes.toBytes((Boolean) record.get(rowField)));
+        return new Put(Bytes.toBytes((Boolean) record.get(keyFieldName)));
```
albertshau: can replace all ```record.get(keyFieldName)``` with ```val```

bdmogal:
@albertshau addressed comments. Please take a look again.

albertshau:
lgtm

elfenheart:
JSON file LGTM

bdmogal:
Build passed - http://builds.cask.co/browse/CDAP-RBT366-6.

Rebased + Squashed. Merging.

PR #3370, opened 2015-07-29
[cdap 3132] fix race condition when test gets run record before (async) start/stop has completed

anew:
title says it all.

Build: http://builds.cask.co/browse/CDAP-RBT380-1


gsps1:
LGTM

PR #3371, opened 2015-07-29
moves Settings icon to top of panel

astewes:
*  Moves Settings icon to the top of the right group side panel

settings


elfenheart:
:+1: LGTM

PR #3372, opened 2015-07-29
CDAP-3106: Improve locating hive-exec.jar

jwang47:
https://issues.cask.co/browse/CDAP-3106
http://builds.cask.co/browse/CDAP-DUT2381

Integration test ExploreTest passed against a CDH 5.4.2 cluster.

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
@@ -36,6 +36,7 @@ import co.cask.cdap.gateway.handlers.CommonHandlers;
 import co.cask.cdap.hive.datasets.DatasetStorageHandler;
 import co.cask.http.HttpHandler;
 import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


cdap-explore/src/main/java/co/cask/cdap/explore/guice/ExploreRuntimeModule.java
```diff
@@ -309,7 +312,10 @@ public boolean accept(String className, URL classUrl, URL classPathUrl) {
     // we need to add this, otherwise when hive runs a MapRedLocalTask it cannot find
     // "org.apache.hadoop.hive.serde2.SerDe" class in its classpath.
     List<String> orderedDependenciesWithHiveJar = Lists.newArrayList(orderedDependenciesStr);
-    orderedDependenciesWithHiveJar.add(new HiveConf().getJar());
+    String hiveExecJar = new JobConf(Task.class).getJar();
+    Preconditions.checkNotNull(hiveExecJar, "Couldn't locate hive-exec.jar to be included in HADOOP_CLASSPATH " +
```
gsps1: log the hiveExecJar as a debug message, so we know what is the jar we are adding when we want to debug.

gsps1:
did you figure out why hive-jdbc-1.1.0-cdh5.4.4-standalone.jar was added instead?

jwang47:
Yes, previously we were getting the jar containing the `HiveConf` class, which is in both hive-exec and hive-common, so there was a possibility that we didn't get the hive-exec jar. Now we're only getting the jar containing a class only in hive-exec, so we shouldn't have the problem anymore.

jwang47:
Oops. Wanted to open against release/3.1. Will wait for build to pass then reopen against release/3.1 and merge.

gsps1:
did you figure out why hive-common is coming before hive-exec, i believe we control the classpath and make sure hive-exec comes at the beginning. if it's not, we have to see why that's happening.

jwang47:
I'm not sure why hive-common is coming before hive-exec (or if it even is), but we shouldn't depend on the order if possible. I can definitely take a look at that if it becomes an issue, but I was able to get the explore integration test to pass with this fix, so I don't think it's an issue here.

gsps1:
LGTM, please open the pr against release

jwang47:
Reopened against release/3.1 https://github.com/caskdata/cdap/pull/3380

PR #3375, opened 2015-07-29
CDAP-2925: Document how to create dataset with properties better

jwang47:


sreevatsanraman:
open against release/3.1

PR #3385, opened 2015-07-29
CDAP 2892 : Deploy schedule triggers in paused state and get rid of lazyStart() scheduler hack

rohitsinha54:
This PR does the following: 
- CDAP 2892 : When a new app is deployed with workflow its schedules are in disabled state unless resumed by the user manually. This is true for both time and stream size based schedule.
- Gets rid of lazyStart() hack which was required to deploy schedules with a stopped scheduler to ensure that they are in paused state.

Approach:
The way achieve paused triggers in Quartz scheduler is through groups. [More Details] (https://issues.cask.co/browse/CDAP-2892?focusedCommentId=17594&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17594)
- During initialization of scheduler we create add an empty job within a group. This is needed as I was not able to find any direct way to create a group.
- This group is then paused which guarantees that any new triggers added to this group will also be added with paused state. We add all our new triggers to this paused group.
- When resume is called on a schedule we move the trigger to default group and resume it. This ensure that when cdap restarts again and scheduler gets initialized it will not pause the triggers which are resumed. 
- Any operation which needs TriggerKey use getGroupedTriggerKey() to get the appropriate TriggerKey as the key will be different depending on which group it belong to.

StreamSizeScheduler: The scheduler of this scheduler does not set activeState to  true anymore. They get activated when resume is called on them.

Build: http://builds.cask.co/browse/CDAP-RBT383-2

***
Tasks pending for these changes 
- [x] Unit tests: The way we pause trigger according to groups changed to handle cdap restarts and my earlier unit tests are not valid anymore. Working on updating them now.
- [x] Testing on cluster with CDAP restart. 
- [ ] Schedules need to be resumed in cdap-integration tests. 

***
Testing: 
Following steps were followed to test the schedules across cdap restart to check that resumed schedules does not get paused when cdap  starts again. It was done on standalone and singlenode cluster and works well on both.
1. Deployed Purchase app with two schedules one being run every  3 minutes.
2. Made sure  that  the status of the schedule is suspended and the workflow does not start even if the schedule time of these paused schedule is reached.
3. Resume the 3 minute schedule and make sure the other (every day one) is still paused. 
4. Made sure that the workflow runs every 3 minutes. 
5. Stop cdap services. (I did it when the workflow was not running)
6. Restart cdap
7. Checked that the 3 min schedule is in scheduled state and every day one is paused. 
8. Redeploy the app again and this time it has one more schedule which (run every 2 minutes) - Apparently this is the only way too add schedules for now.
9. Made sure that after redeploy every 3 min is still scheduled and keeps running the workflow and every 2 min is suspended state
10. Resume every 2 min schedule and made sure it starts workflow.
11. Suspend the every 3 min schedule and made sure it does not start workflow.


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
```diff
@@ -498,12 +498,11 @@ private void startWorkflowAdapter(Id.Namespace namespace, AdapterDefinition adap
...
                          ProgramOptionConstants.ADAPTER_NAME, adapterSpec.getName(),
-                         ProgramOptionConstants.ADAPTER_SPEC, GSON.toJson(adapterSpec),
-                         // hack for scheduler weirdness in unit tests, remove once CDAP-2281 is done
-                         Constants.Scheduler.IGNORE_LAZY_START, String.valueOf(true)
-                       ));
+                         ProgramOptionConstants.ADAPTER_SPEC, GSON.toJson(adapterSpec)));
     //TODO: Scheduler API should also manage the MDS.
     store.addSchedule(workflowId, scheduleSpec);
+    scheduler.resumeSchedule(workflowId, scheduleSpec.getProgram().getProgramType(),
+                             scheduleSpec.getSchedule().getName());
```
anew: so, adapter schedules are always active right away?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -240,8 +267,7 @@ public void suspendSchedule(Id.Program program, SchedulableProgramType programTy
     throws NotFoundException, SchedulerException {
     checkInitialized();
     try {
-      scheduler.pauseTrigger(new TriggerKey(AbstractSchedulerService.scheduleIdFor(program, programType,
-                                                                                   scheduleName)));
+      scheduler.pauseTrigger(getGroupedTriggerKey(program, programType, scheduleName));
```
anew: what happens if someone attempts to pause a schedule that is still in the new paused triggers group? Will he get an error: schedule not found? Or will he get: schedule is already paused?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -395,4 +436,31 @@ public Job newJob(TriggerFiredBundle bundle, org.quartz.Scheduler scheduler)
...
+   * @return Trigger key created from program, programType and scheduleName and TimeScheuler#PAUSED_NEW_TRIGGERS_GROUP
+   * if it exists in this group else returns the {@link TriggerKey} prepared with null which gets it with
+   * {@link Key#DEFAULT_GROUP}
+   * @throws org.quartz.SchedulerException
+   */
+  private TriggerKey getGroupedTriggerKey(Id.Program program, SchedulableProgramType programType, String scheduleName)
+    throws org.quartz.SchedulerException {
+
+    TriggerKey neverResumedTriggerKey = new TriggerKey(AbstractSchedulerService.scheduleIdFor(program, programType,
+                                                                                              scheduleName),
```
anew: this: new TriggerKey(AbstractSchedulerService.scheduleIdFor(program, programType, scheduleName), and this: new TriggerKey(AbstractSchedulerService.scheduleIdFor(program, programType, scheduleName)); appear to be repeated quite often. There also seems to be a bit of duplication between this method and line 285ff. Not sure if you could refactor that.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -395,4 +436,31 @@ public Job newJob(TriggerFiredBundle bundle, org.quartz.Scheduler scheduler)
...
+                                                       PAUSED_NEW_TRIGGERS_GROUP);
+    if (scheduler.checkExists(neverResumedTriggerKey)) {
+      return neverResumedTriggerKey;
+    }
+    return new TriggerKey(AbstractSchedulerService.scheduleIdFor(program, programType, scheduleName));
```
anew: Can it happen that two concurrent request come in for the same trigger. The first calls this method, decides to resume, that is move it out of the group, and reschedule it. That while the second request calls this method. Is any race condition possible?

anew:
@rohitsinha54, overall this looks good. Please see my few comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/AdapterService.java
rohitsinha54: Yes. AdapterScheduler will not need to be resumed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
rohitsinha54: Yes I have thought about this scenario while implementing. Here is my finding: 
- Trigger Paused state is mapped to our Suspended state.
- Now if suspend action is performed on a suspended trigger our handler in ProgramLifeCycleHttpHandler will send a conflict with schedule already suspended message": https://github.com/caskdata/cdap/blob/release/3.1/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java#L434-434

anew: ok

rohitsinha54:
New Build : http://builds.cask.co/browse/CDAP-RBT383-6

rohitsinha54:
This change is to risky to go in release 3.1 so we are moving it to 3.2
Opened a new PR against develop here: https://github.com/caskdata/cdap/pull/3451

PR #3391, opened 2015-07-30
[CDAP-3077] Added another config parameter in Table Sink to allow cas…

bdmogal:
…e-insensitive find for row key field in the incoming record's schema.

Also:
- Refactored TableSink, RealtimeTableSink and TableSource to share their PluginConfig class.
- Added a delete() method to AdapterManager for cleanup in tests. Still not implemented in RemoteAdapterManager.

re-use table config in TableSource

This was wrongly merged to develop. Creating against release/3.1. Already reviewed in #3366


albertshau:
:+1:

PR #3393, opened 2015-07-30
Explicitly set list of Eclipse plugins to install and set HDP 2.1

wolf31o2:
This fixes the VM build for release.

- Using a known-good version of HDP to get Flume Agent
- Set `sdk_version` default to `3.1.0`
- Set explicit list of Eclipse plugins to install


sreevatsanraman:
lgtm

PR #3394, opened 2015-07-30
CDAP-765: access to custom and system metrics in unit-test framework

abaranau:


abaranau:
http://builds.cask.co/browse/CDAP-DUT2394-1

abaranau:
GREEN

gsps1:
added a commit to for minor fixes. build running.

gsps1:
http://builds.cask.co/browse/CDAP-DUT2394-2/test - build passed

cdap-client/src/main/java/co/cask/cdap/client/MetricsClient.java
```diff
@@ -266,6 +266,16 @@ public void waitFor(String name, long count,
         doWaitFor(name, count, timeout, timeoutUnit);
       }
 
+      @Override
+      public long getCounter(String name) {
```
anwar6953: Why is the implementation the same for `getCounter` and `getGauge` in `MetricsClient`, but the implementations are different in `RuntimeStats`?

cdap-unit-test/src/main/java/co/cask/cdap/test/RuntimeStats.java
```diff
@@ -182,6 +192,15 @@ public String toString() {
 
   private static long getTotalCounter(Map<String, String> context, String metricName) {
     MetricDataQuery query = getTotalCounterQuery(context, metricName);
+    return getSingleValueFromTotals(query);
+  }
+
+  private static long getLatestGauge(Map<String, String> context, String metricName) {
```
anwar6953: actually, maybe the question is why do we need this `getLatestGauge` method with the same implementation as `getTotalCounter` above it?

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
```diff
@@ -740,6 +740,35 @@ public void testAppWithServices() throws Exception {
...
   @Test
+  public void testWaitOnUserMetrics() throws Exception {
+
+    ApplicationManager applicationManager = deployApplication(AppWithUserMetrics.class);
+    applicationManager.getFlowManager("fooFlow").start();
+
+    // Send some inputs to streams
+    StreamManager streamManager = getStreamManager("fooStream");
+    for (int i = 0; i < 100; i++) {
+      streamManager.send(ImmutableMap.of("index", String.valueOf(i)), "testing message " + i);
```
anwar6953: Do we need to send headers?
Doesn't seem like its used.

anwar6953:
Also, it seems that the tags being queried are very limited (not customizable, like the metric name is).
Do you think that should also be more customizable? Perhaps at least in a separate PR or JIRA?

gsps1:
@anwar6953 i don't think we need customization at tag level, as metrics are emitted from one of the program and we can get context for those programs, we are using those tags to get RuntimeMetrics.

anwar6953:
The issue with RuntimeStats is that there are methods only for certain program types (flow, flowlet, service, service handler). There are many tags that are omitted (dataset, workflow, spark, mapreduce, etc).

gsps1:
then we have to add implementation for those in their respective managers.

anwar6953:
Why not simply expose a method like the `getMetrics` of `RuntimeStats` that allows user to pass in tags, and then query by metric name?

gsps1:
Deprecated 
 1) RuntimeMetrics and 2) RuntimeStats classes. 

Using MetricsManager class, which has method to getMetrics<context, metric>  and getMetrics with retry. 

Updated TestFrameworkTestRun with dependent changes, using MetricsManager instead of deprecated RuntimeMetrics now.

Things remaining:
1) Change examples and tests using RuntimeMetrics to use MetricsManager
2) Change MetricsClient API to use MetricsManager instead of RuntimeMetrics and dependent changes.

will create JIRA to track the above two after the review and do it in a separate PR.

cdap-api/src/main/java/co/cask/cdap/api/metrics/RuntimeMetrics.java
```diff
@@ -24,8 +24,8 @@
 /**
  * Provides access to metrics about running processors.
  *
- * TODO: Support custom metrics https://issues.cask.co/browse/CDAP-765
  */
+@Deprecated
```
anwar6953: indicate what to use instead

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
gsps1: its used to emit, even/odd metrics in the flowlet.

cdap-unit-test/src/main/java/co/cask/cdap/test/MetricsManager.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  public Collection<MetricTimeSeries> query(MetricDataQuery query) throws Exception {
+    return metricStore.query(query);
+  }
+
+
+  /**
+   * waitForFlowlet's events processed metric, specified by count, and the timeout and timOutUnit is provided.
+   * @param namespace
```
anwar6953: Its not useful to have params in javadoc, if they're not going to have description, right?

cdap-api/src/main/java/co/cask/cdap/api/metrics/RuntimeMetrics.java
```diff
@@ -23,9 +23,9 @@
 
 /**
  * Provides access to metrics about running processors.
- *
- * TODO: Support custom metrics https://issues.cask.co/browse/CDAP-765
+ * deprecated, Usages in MetricsClient and RuntimeStats will be replaced alternative implementation.
```
anwar6953: I think it should be similar to deprecation description from other class:
`This is deprecated, should use {@link co.cask.cdap.test.MetricsManager} instead`

Is that description correct?

cdap-unit-test/src/main/java/co/cask/cdap/test/MetricsManager.java
```diff
@@ -0,0 +1,135 @@
...
+                              Constants.Metrics.Tag.FLOW, flowName,
+                              Constants.Metrics.Tag.FLOWLET, flowletName
+    ), Constants.Metrics.Name.Flow.FLOWLET_PROCESSED, count, timeout, timeUnit);
+
+  }
+
+
+  /**
+   * waitFor a metric value count for the metric identified by metricName and context.
+   * @param context
```
anwar6953: same here

cdap-unit-test/src/main/java/co/cask/cdap/test/MetricsManager.java
anwar6953: app, flowName, flowletName are self-explanatory, but timeout and count are not (units?)

gsps1:
removed deprecation for RuntimeMetrics, just deprecated RuntimeStats. using RuntimeMetrics in MetricsManager.

anwar6953:
4 nit/trivial comments.
LGTM besides those.

gsps1:
thanks @anwar6953 for the review. i have addressed your comments. please review again.

gsps1:
build running : http://builds.cask.co/browse/CDAP-DUT2394-3

gsps1:
build GREEN, merging

PR #3395, opened 2015-07-30
Update docs for JDK version

sreevatsanraman:


awholegunch:
LGTM. Created an issue (https://issues.cask.co/browse/CDAP-3150) so that in the future this all could happen in one location.

PR #3396, opened 2015-07-30
Add @Beta annotation and make final to RestartServiceInstancesStatus class.

henry-cask:
Tagged the RestartServiceInstancesStatus class as Beta and add final modifier to make sure
it is not inherited.

This class may change in the near future to support more hollistic history of system
services statuses.

### DIFF
--- a/cdap-proto/src/main/java/co/cask/cdap/proto/RestartServiceInstancesStatus.java
+++ b/cdap-proto/src/main/java/co/cask/cdap/proto/RestartServiceInstancesStatus.java
@@ -15,12 +15,15 @@
  */
 package co.cask.cdap.proto;
 
+import co.cask.cdap.api.annotation.Beta;
+
 import java.util.Set;
 
 /**
  * Class to describe status of restart instances of service request.
  */
-public class RestartServiceInstancesStatus {
+@Beta
+public final class RestartServiceInstancesStatus {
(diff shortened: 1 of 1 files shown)


chtyim:
Proto is internal class, right? Do we need @Beta?

henry-cask:
We have couple in proto like AdapterDetail and AdapterConfig.

henry-cask:
Proto is internal to people developing CDAP so assuming people fork CDAP and start developing on top of CDAP, the @Beta annotation could help indicate this could change in near future.

chtyim:
I don't object it, but at the same time, it means a lot of internal APIs need to have @Beta :)

henry-cask:
Yeah, I personally think we should never add @Beta for release product =P
The original idea is that this class may evolve pretty soon, so @Beta is added to give heads up.

I would leave this open until tonight. 
If @anew is ok we DO NOT add @Beta annotation for this class, I will just close this PR as will not fix.

albertshau:
proto is accessible in unit test so it's kind of public in a sense

anew:
proto is not internal because it represents our REST API. Agreed that the serialized/jsonized proto objects won't have the @Beta, so I am ok with not having it there. But we should have some other way of indicating for a REST API that it is experimental.

henry-cask:
Closing it for now.

PR #3397, opened 2015-07-30
CDAP-3148 Fix mapreduce.application.classpath for HDP 2.2.x clusters for explore

poornachandra:
JIRA - https://issues.cask.co/browse/CDAP-3148

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
@@ -99,6 +99,7 @@ public class ExploreServiceUtils {
 
   private static final Pattern HIVE_SITE_FILE_PATTERN = Pattern.compile("^.*/hive-site\\.xml$");
   private static final Pattern YARN_SITE_FILE_PATTERN = Pattern.compile("^.*/yarn-site\\.xml$");
+  private static final Pattern MAPRED_SITE_FILE_PATTERN = Pattern.compile("^.*/mapred-site\\.xml$");
 
   /**
    * Get all the files contained in a class path.
@@ -349,14 +350,16 @@ public class ExploreServiceUtils {
       return hijackHiveConfFile(confFile);
     } else if (YARN_SITE_FILE_PATTERN.matcher(confFile.getAbsolutePath()).matches()) {
       return hijackYarnConfFile(confFile);
+    } else if (MAPRED_SITE_FILE_PATTERN.matcher(confFile.getAbsolutePath()).matches()) {
+      return hijackMapredConfFile(confFile);
     } else {
       return confFile;
     }
   }
 
   /**
-   * Check that the file is a yarn-site.xml file, and return a temp copy of it to which are added
-   * necessary options. If it is not a yarn-site.xml file, return it as is.
+   * Change yarn-site.xml file, and return a temp copy of it to which are added
+   * necessary options.
    */
   private static File hijackYarnConfFile(File confFile) {
     Configuration conf = new Configuration(false);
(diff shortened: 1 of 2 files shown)


poornachandra:
Build - http://builds.cask.co/browse/CDAP-RBT386-1

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
```diff
@@ -389,6 +392,39 @@ private static File hijackYarnConfFile(File confFile) {
...
+    String mrAppClassPath = conf.get(MRJobConfig.MAPREDUCE_APPLICATION_CLASSPATH,
+                                       MRJobConfig.DEFAULT_MAPREDUCE_APPLICATION_CLASSPATH);
+
+    // Add the pwd/* at the beginning of classpath. Without this change, old jars from mr framework classpath
+    // get into classpath.
+    mrAppClassPath = "$PWD/*," + mrAppClassPath;
+
+    conf.set(MRJobConfig.MAPREDUCE_APPLICATION_CLASSPATH, mrAppClassPath);
+
+    File newMapredConfFile = new File(Files.createTempDir(), "mapred-site.xml");
```
chtyim: Will this result in a lot of temp directories in the container when a lot of queries has been executed through the explore service?

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
```diff
@@ -389,6 +392,39 @@ private static File hijackYarnConfFile(File confFile) {
   }
 
   /**
+   * Check that the file is a mapred-site.xml file, and return a temp copy of it to which are added
+   * necessary options. If it is not a mapred-site.xml file, return it as is.
```
gsps1: this comment is not valid, as we are not doing those checks here. i know its from other method, can you change it?

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
poornachandra: This creation of temp files happens in master when launching explore container. So, only one copy of each file should be created for each master run.

poornachandra:
@gsps1 Fixed review comments, please check again.

poornachandra:
Build - http://builds.cask.co/browse/CDAP-RBT386-3

gsps1:
LGTM, merge after build passes.

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
chtyim: In that case, use the temp directory inside the data directory so that when master starts again, it will gets cleanup. See https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java#L120

poornachandra:
Build http://builds.cask.co/browse/CDAP-RBT386-4 is green.

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreServiceUtils.java
chtyim: This comment get ignored and merged??

poornachandra: I missed your final comment yesterday. Here is the PR for it - https://github.com/caskdata/cdap/pull/3419

PR #3398, opened 2015-07-30
CDAP-2735 remove unused tableName property from DBSource

albertshau:


albertshau:
http://builds.cask.co/browse/CDAP-RBT387

bdmogal:
:+1:

PR #3401, opened 2015-07-30
Update cdap-spark-scala-archetype for 3.1

rohitsinha54:
This is just a copy and paste from our SparkKMeansExample. 
- Bumped up the spark core version to cdap dependency version (which is currently 1.3.1)
- Tested creation of project from archetype
- Tested project build and unit tests
- Tested being able to open in IDE


cdap-archetypes/cdap-spark-scala-archetype/src/main/resources/archetype-resources/src/main/scala/Point.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  public double getY() {
+    return y;
+  }
+
+  public double getZ() {
+    return z;
+  }
+}
```
awholegunch: add a newline.

sreevatsanraman:
LGTM. Fix the one minor issue from JJ and merge.

PR #3403, opened 2015-07-30
CDAP-2376: Make sure Mapper/Reducer Logs are flushed 

gokulavasan:
For very short running MapReduce jobs, the logs were not collected. The issue was that the KakfaLogAppender was not getting closed and thus the messages were not getting flushed. PR makes sure that LogAppenderInititalizer is closed in MapperWrapper, ReducerWrapper and in DatasetInputFormat, DatasetOutputFormat (by closing MapReduceContextProvider). Also had to package logback.xml when building the MapReduce Job JAR so that messages are formatted in the expected format.

JIRA : https://issues.cask.co/browse/CDAP-2376

Build : http://builds.cask.co/browse/CDAP-RBT388-1

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextProvider.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextProvider.java
@@ -90,6 +90,12 @@ public final class MapReduceContextProvider {
     return context;
   }
 
+  public synchronized void stop() {
+    if (contextBuilder != null) {
+      contextBuilder.finish();
+    }
+  }
+
   private synchronized AbstractMapReduceContextBuilder getBuilder(CConfiguration conf) {
     if (contextBuilder != null) {
       return contextBuilder;
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextProvider.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextProvider.java
@@ -90,6 +90,8 @@ public final class MapReduceContextProvider {
     return context;
   }
 
+  // TODO: CDAP-3160 : Refactor to remove the need for the stop method below. Provider/Builder classes should not have
+  // methods like stop(), finish() or close().
   public synchronized void stop() {
     if (contextBuilder != null) {
       contextBuilder.finish();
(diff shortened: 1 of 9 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextProvider.java
```diff
@@ -90,6 +90,12 @@ public synchronized BasicMapReduceContext get() {
     return context;
   }
 
+  public synchronized void stop() {
```
chtyim: It sounds like a bad idea to have provider or builder to method like stop() or finish() or close(). Can you open a JIRA for refactoring that?

gokulavasan:
@chtyim Addressed comment. Please review.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
```diff
@@ -773,8 +774,19 @@ public boolean accept(String className, URL classUrl, URL classPathUrl) {
       LOG.warn("Not including HBaseTableUtil classes in submitted Job Jar since they are not available");
     }
 
+    // Add the logback.xml as a resource while creating the MapReduce Job JAR
+    Set<URI> logbackURI = Sets.newHashSet();
+    try {
+      URL logback = getClass().getResource("logback.xml");
```
chtyim: Where is the logback.xml file located?

gokulavasan: This will be made available by the AbstractDistributedProgramRunner (getLogbackURI method) launches the MapReduceTwillApp ?

chtyim: Have you verify? `Class.getResource()` will prepend the current class package name as the prefix. Hence it will try to look for a resource in "co/cask/cdap/internal/app/runtime/batch/logback.xml", which I don't think there is such file in the classloader path.

gokulavasan:
@chtyim Please review. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceRuntimeService.java
gokulavasan: Thanks for the catch Terence. Made the path absolute.

chtyim:
LGTM

PR #3404, opened 2015-07-30
Throw exception if WorkflowToken.put operation happens from the Spark executor.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-2895


sagarkapare:
Release build is running here - http://builds.cask.co/browse/CDAP-RBT390-1

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -40,6 +43,7 @@
   private Map<String, Map<String, Long>> mapReduceCounters;
   private final Map<Scope, Map<String, List<NodeValue>>> tokenValueMap = new EnumMap<>(Scope.class);
   private String nodeName;
+  private transient boolean putOperationAllowed = true;
```
chtyim: Just call it `putAllowed`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -117,6 +121,11 @@ public void put(String key, Value value) {
   }
 
   void put(String key, Value value, Scope scope) {
+    // Check if put operation is being performed inside the Spark executor
+    if (!putOperationAllowed) {
+      throw new UnsupportedOperationException("Put operation is not allowed from Spark executor.");
```
chtyim: Shouldn't mention Spark in this class, since it is not spark specific. E.g. in Mapper/Reducer context, put is also not allowed.

cdap-app-fabric/src/test/java/co/cask/cdap/WorkflowApp.java
```diff
@@ -120,20 +124,44 @@ public void run(SparkContext context) {
...
+
       // If there are problems accessing workflow token here, Spark will throw a NotSerializableException in the test
       final WorkflowToken workflowToken = context.getWorkflowToken();
       if (workflowToken != null) {
-        workflowToken.put("otherKey", "otherValue");
+        workflowToken.put("multiplier", "2");
       }
-      distData.map(new Function<Integer, Integer>() {
+
+      final boolean addToToken = context.getRuntimeArguments().containsKey("addToTokenInFunction");
```
chtyim: Better call it `closurePutToken`.

cdap-app-fabric/src/test/java/co/cask/cdap/WorkflowApp.java
```diff
@@ -120,20 +124,44 @@ public void run(SparkContext context) {
...
+            return multiplier * val;
+          }
+          return 0;
         }
       });
+
+      boolean exceptionThrown = false;
+      try {
+        List<Integer> result = mapData.collect();
+        Assert.assertTrue(result.equals(Arrays.asList(2, 4, 6, 8, 10)));
```
chtyim: Better not to run assert in the program. Assert should be done in the test case.

cdap-app-fabric/src/test/java/co/cask/cdap/WorkflowApp.java
```diff
@@ -120,20 +124,44 @@ public void run(SparkContext context) {
...
+      } catch (Throwable t) {
+        exceptionThrown = true;
+      }
+
+      if (addToToken) {
+        Assert.assertTrue(exceptionThrown);
```
chtyim: If the closure throws exception, the spark program execution should be in failed state. It's better to verify that from the test method, otherwise it's not really verifying.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Renamed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: It is specific to Spark in this case. For MapReduce we do it through `NonUpdatableToken` added as a part of the PR - #3339 

MapperWrapper and ReducerWrapper gets the BasicMapReduceContext from the MapReduceContextProvider. There we supply the NonUpdatableToken.

cdap-app-fabric/src/test/java/co/cask/cdap/WorkflowApp.java
sagarkapare: Renamed and testing it inside the WorkflowHttpHandlerTest.

cdap-app-fabric/src/test/java/co/cask/cdap/WorkflowApp.java
sagarkapare: Updated the test cases to reflect this.

sagarkapare:
@chtyim Addressed comments please take a look again. Thanks!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
chtyim: Why it cannot be this class with putAllowed = false for MR case?

sagarkapare: Fixed. Removed the NonUpdatableToken class and using putAllowed flag for the MR case.

sagarkapare:
@chtyim Addressed comments. Please take a look. Thanks!

chtyim:
LGTM

sagarkapare:
Thank you @chtyim for the review. Release build is running here - http://builds.cask.co/browse/CDAP-RBT390-5 Will merge once it passes.

sagarkapare:
Build passed. Merging.

PR #3405, opened 2015-07-30
[CDAP-2895] Add a max permissible size for Workflow Token

bdmogal:
Build: http://builds.cask.co/browse/CDAP-RBT389
Jira: [CDAP-2895](https://issues.cask.co/browse/CDAP-2895)

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
@@ -16,11 +16,14 @@
 
 package co.cask.cdap.api.workflow;
 
+import co.cask.cdap.api.annotation.Beta;
+
 import java.io.Serializable;
 
 /**
  * Class representing the value of the key in the {@link WorkflowToken}.
  */
+@Beta
 public class Value implements Serializable {
 
   private static final long serialVersionUID = -3420759818008526875L;
@@ -32,24 +35,31 @@ public class Value implements Serializable {
   }
 
   /**
-   * @return the boolean value
+   * @return the result of calling <code>Boolean.parseBoolean(toString())</code>
    */
   public boolean getAsBoolean() {
     return Boolean.parseBoolean(value);
   }
 
   /**
-   * @return the int value
+   * @return the result of calling <code>Integer.parseInt(toString())</code>
    */
   public int getAsInt() {
     return Integer.parseInt(value);
   }
 
   /**
-   * @return the long value
+   * @return the result of calling <code>Long.parseLong(toString())</code>
    */
   public long getAsLong() {
     return Long.parseLong(value);
+  }
+
+  /**
+   * @return the result of calling <code>Double.parseDouble(toString())</code>
+   */
+  public double getAsDouble() {
+    return Double.parseDouble(value);
   }
 
   /**
@@ -81,6 +91,7 @@ public class Value implements Serializable {
 
   /**
    * Return the {@link Value} representation of the input <code>boolean</code> parameter.
+   *
    * @param value the <code>boolean</code> value to be represented as {@link Value}
    * @return a {@link Value} representation of the argument
    */
@@ -90,6 +101,7 @@ public class Value implements Serializable {
 
   /**
    * Return the {@link Value} representation of the input <code>int</code> parameter.
+   *
    * @param value the <code>int</code> value to be represented as {@link Value}
    * @return a {@link Value} representation of the argument
    */
@@ -99,6 +111,7 @@ public class Value implements Serializable {
 
   /**
    * Return the {@link Value} representation of the input <code>long</code> parameter.
+   *
    * @param value the <code>long</code> value to be represented as {@link Value}
    * @return a {@link Value} representation of the argument
    */
@@ -107,7 +120,18 @@ public class Value implements Serializable {
   }
 
   /**
+   * Return the {@link Value} representation of the input <code>double</code> parameter.
+   *
+   * @param value the <code>double</code> value to be represented as {@link Value}
+   * @return a {@link Value} representation of the argument
+   */
+  public static Value of(double value) {
+    return new Value(String.valueOf(value));
+  }
+
+  /**
    * Return the {@link Value} representation of the input <code>String</code> parameter.
+   *
    * @param value the <code>String</code> value to be represented as {@link Value}
    * @return a {@link Value} representation of the argument
    */
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
@@ -19,6 +19,7 @@ package co.cask.cdap.internal.app.runtime.workflow;
 import co.cask.cdap.api.workflow.NodeValue;
 import co.cask.cdap.api.workflow.Value;
 import co.cask.cdap.api.workflow.WorkflowToken;
+import co.cask.cdap.common.conf.Constants;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -37,14 +38,23 @@ public class BasicWorkflowToken implements WorkflowToken, Serializable {
 
   private static final long serialVersionUID = -1173500180640174909L;
 
-  private Map<String, Map<String, Long>> mapReduceCounters;
   private final Map<Scope, Map<String, List<NodeValue>>> tokenValueMap = new EnumMap<>(Scope.class);
+  private final int maxSizeBytes;
+  private Map<String, Map<String, Long>> mapReduceCounters;
   private String nodeName;
+  private int bytesLeft;
 
-  public BasicWorkflowToken() {
+  /**
+   * Creates a {@link BasicWorkflowToken} with the specified maximum size.
+   *
+   * @param maxSizeMb the specified maximum size in MB for the {@link BasicWorkflowToken} to create.
+   */
+  public BasicWorkflowToken(int maxSizeMb) {
     for (Scope scope : Scope.values()) {
-      tokenValueMap.put(scope, new HashMap<String, List<NodeValue>>());
+      this.tokenValueMap.put(scope, new HashMap<String, List<NodeValue>>());
     }
+    this.maxSizeBytes = maxSizeMb * 1024 * 1024;
+    this.bytesLeft = maxSizeBytes;
   }
 
   private BasicWorkflowToken(BasicWorkflowToken other) {
@@ -62,6 +72,8 @@ public class BasicWorkflowToken implements WorkflowToken, Serializable {
     if (other.mapReduceCounters != null) {
       this.mapReduceCounters = copyHadoopCounters(other.mapReduceCounters);
     }
+    this.maxSizeBytes = other.maxSizeBytes;
+    this.bytesLeft = other.bytesLeft;
   }
 
   void setCurrentNode(String nodeName) {
@@ -78,8 +90,9 @@ public class BasicWorkflowToken implements WorkflowToken, Serializable {
       Map<String, List<NodeValue>> thisTokenValueMapForScope = this.tokenValueMap.get(entry.getKey());
 
       for (Map.Entry<String, List<NodeValue>> otherTokenValueMapForScopeEntry : entry.getValue().entrySet()) {
-        if (!thisTokenValueMapForScope.containsKey(otherTokenValueMapForScopeEntry.getKey())) {
-          thisTokenValueMapForScope.put(otherTokenValueMapForScopeEntry.getKey(), Lists.<NodeValue>newArrayList());
+        String otherKey = otherTokenValueMapForScopeEntry.getKey();
+        if (!thisTokenValueMapForScope.containsKey(otherKey)) {
+          thisTokenValueMapForScope.put(otherKey, Lists.<NodeValue>newArrayList());
         }
 
         // Iterate over the list of NodeValue corresponding to the current key.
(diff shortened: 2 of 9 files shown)


cdap-api/src/main/java/co/cask/cdap/api/workflow/Value.java
```diff
@@ -53,6 +56,13 @@ public long getAsLong() {
   }
 
   /**
+   * @return the double value
+   */
+  public double getAsDouble() {
```
chtyim: So this method will throw `NumberFormatException` if the value is not a double. It's better document it. Or you can simply say, it returns the result of calling `Double.parseDouble(toString())`. Document similarly for other conversion methods.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -40,11 +42,20 @@
   private Map<String, Map<String, Long>> mapReduceCounters;
   private final Map<Scope, Map<String, List<NodeValue>>> tokenValueMap = new EnumMap<>(Scope.class);
   private String nodeName;
+  private final int maxSizeBytes;
```
chtyim: Better group all `final` fields first, followed by non-final fields.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -40,11 +42,20 @@
...
+   *
+   * @param maxSizeMb the specified maximum size in MB for the {@link BasicWorkflowToken} to create.
+   */
+  public BasicWorkflowToken(int maxSizeMb) {
     for (Scope scope : Scope.values()) {
-      tokenValueMap.put(scope, new HashMap<String, List<NodeValue>>());
+      this.tokenValueMap.put(scope, new HashMap<String, List<NodeValue>>());
     }
+    this.maxSizeBytes = maxSizeMb * 1024 * 1024;
+    this.currentSizeBytes = 0;
```
chtyim: not needed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -87,15 +101,16 @@ void mergeToken(BasicWorkflowToken other) {
...
             if (thisNodeValue.equals(otherNodeValue)) {
               otherNodeValueExist = true;
               break;
             }
           }
           if (!otherNodeValueExist) {
-            thisTokenValueMapForScope.get(otherTokenValueMapForScopeEntry.getKey()).add(otherNodeValue);
+            ensureEnoughMemoryAvailable(otherKey, otherNodeValue.getValue());
+            thisTokenValueMapForScope.get(otherKey).add(otherNodeValue);
+            updateSize(otherKey, otherNodeValue.getValue());
```
chtyim: Why not encapsulate this three into a method like `addNodeValue(Map<String, List<NodeValue>> map, String key, NodeValue nodeValue)` that would do check and update size and adding to the list?
```

File: cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java [hunk 5]
```diff
@@ -247,4 +265,21 @@ public WorkflowToken deepCopy() {
     }
     return builder.build();
   }
+
+  private void ensureEnoughMemoryAvailable(String key, Value value) {
+    int totalSizeAfterAdding  = Bytes.toBytes(key).length + Bytes.toBytes(value.toString()).length + currentSizeBytes;
```
chtyim: Technically it can overflow. It's better be:

```java
int bytesLeft; /// Initialize to max size in constructor
....
int left = bytesLeft;
left -= key.length;
if (left <= 0) {
  // throw exception
}
left -= value.length;
if (left <= 0) {
  // throw exception
}
// Update the size as well
bytesLeft = left;
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -247,4 +265,21 @@ public WorkflowToken deepCopy() {
...
+                                                      "permitted size by setting the parameter '%s' in " +
+                                                      "cdap-site.xml to add more values.",
+                                                    maxSizeBytes / (1024 * 1024),
+                                                    currentSizeBytes / (1024 * 1024),
+                                                    Constants.AppFabric.WORKFLOW_TOKEN_MAX_SIZE_MB));
+    }
+  }
+
+  private void updateSize(String key, Value value) {
+    currentSizeBytes += Bytes.toBytes(key).length + Bytes.toBytes(value.toString()).length;
```
chtyim: This is redundant conversion of Bytes. Should merge this method with the previous one.

bdmogal: I had kept these two methods separate, so I check that enough memory is available as one of the first lines in ``put()`` (line 144), while only update ``currentSizeBytes`` (and now ``bytesLeft``) after the put operation was successful (line 162). However, removed the Bytes conversion and just using ``String.length()``. Let me know if that is ok.

bdmogal:
@chtyim addressed comments. Please have another look.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -247,4 +269,34 @@ public WorkflowToken deepCopy() {
...
+                                                      "adding key '%s' with value '%s'. Current size is '%sMB'. " +
+                                                      "Please increase the maximum permitted size by setting the " +
+                                                      "parameter '%s' in cdap-site.xml to add more values.",
+                                                    maxSizeBytes / (1024 * 1024), key, value,
+                                                    currentSizeBytes / (1024 * 1024),
+                                                    Constants.AppFabric.WORKFLOW_TOKEN_MAX_SIZE_MB));
+    }
+  }
+
+  private void updateSize(String key, Value value) {
```
chtyim: Why need to separate as two? Can it simply just do compute the potential new size and do the update+put? it's just 1 line each for the update and put.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -247,4 +269,34 @@ public WorkflowToken deepCopy() {
...
+                                                      "adding key '%s'. Current size is '%sMB'. Please increase the " +
+                                                      "maximum permitted size by setting the parameter '%s' in " +
+                                                      "cdap-site.xml to add more values.",
+                                                    maxSizeBytes / (1024 * 1024), key,
+                                                    currentSizeBytes / (1024 * 1024),
+                                                    Constants.AppFabric.WORKFLOW_TOKEN_MAX_SIZE_MB));
+    }
+    left -= value.toString().length();
+    if (left <= 0) {
+      throw new IllegalStateException(String.format("Exceeded maximum permitted size of workflow token '%sMB' while " +
```
chtyim: You can further simplify the logic into one throw:

```java
/**
 * Appends the given key/value to the given map.
 *
 * @throws IllegalArgumentException if the max size will be exceeded after adding the given key/value.
 */
void append(String key, Value value, Map<String, List<NodeValue>> nodeValues) {
  int left = bytesLeft - key.length;
  left = (left > 0) ? left - value.length : left;
  if (left <= 0) {
    throw ...
  }
  nodeValues.get(key).add(new NodeValue(nodeName, value));
  bytesLeft = left;
}
```

And I don't see why this method needs further breakdown (it's less than 10 lines).

chtyim: You can share this method between merge and put easily, without repeating the contract of append (which is check size, add, update size). You may modify the signature a bit to take a `List<NodeValue>` instead of a `Map`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
```diff
@@ -641,7 +641,8 @@ public WorkflowToken getWorkflowToken(Id.Workflow workflowId, String workflowRun
     // For pre-3.1 run records, there won't be a workflow token. Return an empty token for such run records.
     if (workflowToken == null) {
       LOG.debug("Could not find workflowToken for workflow: {}, runId: {}", workflowId, workflowRunId);
```
nitinmotgi: "No workflow token available for workflow {}, run id {}"

bdmogal:
@chtyim updated to address your last comments and updated tests to test more meaningful scenarios. 

I've also handled a condition that was missed previously, where we don't append to the map, but replace one of the ``NodeValue`` objects. This happens when the same node updates a key that it had set previously.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -37,14 +38,24 @@
 
   private static final long serialVersionUID = -1173500180640174909L;
 
-  private Map<String, Map<String, Long>> mapReduceCounters;
   private final Map<Scope, Map<String, List<NodeValue>>> tokenValueMap = new EnumMap<>(Scope.class);
+  private final int maxSizeBytes;
+  private Map<String, Map<String, Long>> mapReduceCounters;
   private String nodeName;
+  private int currentSizeBytes;
```
chtyim: You don't need this. If you have max and left, then current is max - left.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -247,4 +263,57 @@ public WorkflowToken deepCopy() {
...
+   * Replaces the value of a key in the workflow token. Used when the same node adds a newer value for an existing key.
+   *
+   * @param key the key whose value is to be replaced
+   * @param existingNodeValue the existing {@link NodeValue} to replace
+   * @param newNodeValue the new {@link NodeValue} to replace the existing {@link NodeValue} with
+   * @param nodeValues the list of {@link NodeValue} for the specified key
+   * @param index the index at which the value is to be replaced in the node value list
+   * @throws IllegalStateException there is not enough memory available for replacing the existing value
+   */
+  private void replace(String key, NodeValue existingNodeValue, NodeValue newNodeValue, List<NodeValue> nodeValues,
```
chtyim: No need to pass in existingNodeValue since you already got the index.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
sagarkapare: Since we store the token in the MDS, would it be better to store the `currentSizeBytes` and compute `bytesLeft` as max - current?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
chtyim: And you can actually merge this and the previous method into one.

```java
int oldValueLen = (index < 0) ? 0 : nodeValues.get(index).getValue().toString().length();
int valueLen = nodeValue.getValue().toString().length();

int left = bytesLeft - valueLen + oldValueLen;
left = (left < 0 || index >= 0) ? left : left - key.length();
if (left < 0) { 
  ...
}
if (index >= 0) {
  nodeValues.set(index, nodeValue);
} else {
  nodeValues.add(nodeValue);
}
bytesLeft = left;
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
chtyim: Why do we store the size in MDS? Seems like this field should be transient. Actually only the tokenValueMap should be saved, isn't it (or may the mapReduceCounters as well).

bdmogal: ``maxSizeBytes`` would also need to be serialized. Otherwise there is no way to recompute either of these values (if we make them transient, recomputing would be necessary)?

chtyim: When we read it back from MDS, it's for read only purpose, right?

bdmogal: @chtyim we haven't done refactoring that we discussed from ``WorkflowDriver`` -> ``ProgramRunner`` yet. So the workflow token is still passed serialized in ``ProgramOptions``.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
```diff
@@ -247,4 +261,60 @@ public WorkflowToken deepCopy() {
...
+
+  /**
+   * Appends the given key/value to the given list.
+   *
+   * @param key the key to append
+   * @param nodeValue the {@link NodeValue} to append
+   * @param nodeValues the existing, non-null list of {@link NodeValue} for the specified key
+   * @throws IllegalArgumentException if the max size will be exceeded after adding the given key/value.
+   */
+  private void append(String key, NodeValue nodeValue, List<NodeValue> nodeValues) {
```
chtyim: This method is no longer needed, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
chtyim: Oh right. But then it has nothing to do with MDS :)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/BasicWorkflowToken.java
bdmogal: huh! don't know how it got left behind. removed.

chtyim:
LGTM

bdmogal:
Build passed: http://builds.cask.co/browse/CDAP-RBT389-6. Squashed commits. Merging.

PR #3406, opened 2015-07-30
Update cdap-java-spark-archetype

rohitsinha54:
This is just a copy and paste from our SparkKMeansExample. 
- Bumped up the spark core version to cdap dependency version (which is currently 1.3.1)
- Tested creation of project from archetype
- Tested project build and unit tests
- Tested being able to open in IDE

### DIFF
--- a/cdap-archetypes/cdap-spark-java-archetype/src/main/resources/archetype-resources/src/main/java/SparkPageRankApp.java
+++ b/cdap-archetypes/cdap-spark-java-archetype/src/main/resources/archetype-resources/src/main/java/SparkPageRankApp.java
@@ -138,7 +138,7 @@ public class SparkPageRankApp extends AbstractApplication {
 
     @Path(RANKS_SERVICE_PATH)
     @POST
-    public void getRank(HttpServiceRequest request, HttpServiceResponder responder) {
+    public void retrieveRank(HttpServiceRequest request, HttpServiceResponder responder) {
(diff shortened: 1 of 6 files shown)


cdap-archetypes/cdap-spark-java-archetype/src/main/resources/archetype-resources/src/main/java/SparkPageRankApp.java
```diff
@@ -93,15 +127,27 @@ public SparkSpecification configure() {
...
+    public static final String RANKS_SERVICE_PATH = "rank";
+
     @UseDataSet("ranks")
     private ObjectStore<Integer> ranks;
 
-    @Path("rank")
-    @GET
-    public void getRank(HttpServiceRequest request, HttpServiceResponder responder, @QueryParam("url") String url) {
+    @Path(RANKS_SERVICE_PATH)
+    @POST
```
sreevatsanraman: why is getRank a POST call?

rohitsinha54: We need to use POST call as the query string being sent has / and : which can't be sent through path params and the UI does not support query parameters. 
I renamed the method and removed get from it for some clarity.

sreevatsanraman:
LGTM :+1:

rohitsinha54:
Doing one final testing to create from archetype before merge.

PR #3409, opened 2015-07-30
(CDAP-3145) Fix a bug in ResourceCoordinator.

chtyim:
- Instead of using the old set of discoverables in performAssignment, always use the live ServiceDiscovered to get the latest sets of discoverables
- This fixes the case where the coordinator needs to fetch the old assignment while the set is empty, followed by an update in the set of discoverables to become non-empty.


chtyim:
Targeted wrong branch. Closing.

PR #3410, opened 2015-07-30
(CDAP-3145) Fix a bug in ResourceCoordinator.

chtyim:
- Instead of using the old set of discoverables in performAssignment, always use the live ServiceDiscovered to get the latest sets of discoverables
- This fixes the case where the coordinator needs to fetch the old assignment while the set is empty, followed by an update in the set of discoverables to become non-empty.


chtyim:
Verified on cluster. Step to verify is to kill the metrics processor container multiple times to see if the correct assignment can be restored after the leader for resource coordinator is up.

chtyim:
Build: http://builds.cask.co/browse/CDAP-RBT391-1

gsps1:
where/how does the coordinator fetches old assignment when the set is empty ?

abaranau:
looks good from first look, will ask couple Qs in a min at office ;)

abaranau:
we talked. LGTM

chtyim:
Build passed. Merging.

PR #3415, opened 2015-07-30
CDAP-2341 Shorten the cleanup thread for invalid records to make sure store quickly updated.

henry-cask:
Some of programs like Spark, MR, and workflow may die quickly due to memory error so we want
to make sure the runs status updated quickly.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
@@ -77,7 +77,7 @@ public class ProgramLifecycleService extends AbstractIdleService {
     LOG.info("Starting ProgramLifecycleService");
 
     scheduledExecutorService.scheduleWithFixedDelay(new RunRecordsCorrectorRunnable(this),
-                                                    2L, 600L, TimeUnit.SECONDS);
+                                                    2L, 6L, TimeUnit.SECONDS);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


sagarkapare:
LGTM

henry-cask:
Will merge once build pass: http://builds.cask.co/browse/CDAP-RBT392-2

chtyim:
6 seconds? Seems pretty often.

henry-cask:
It was 10 minutes before any number between it should fine too. Any suggestion?

henry-cask:
Will update it to every 5 mins instead if no objection

chtyim:
Can you make it configurable from cdap-site.xml and default it to 300 seconds in cdap-default.xml?

anew:
how about 60 seconds default?

henry-cask:
Make it configurable with default 300 s.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -64,20 +66,26 @@
...
+    this.configuration = configuration;
   }
 
   @Override
   protected void startUp() throws Exception {
     LOG.info("Starting ProgramLifecycleService");
 
+    long interval = configuration.getLong(Constants.AppFabric.PROGRAM_RUNID_CORRECTOR_INTERVAL_SECONDS);
+    if (interval <= 0) {
+      interval = 300L;
```
anwar6953: This will only happen if user defines an invalid value in their configuration, right?
If so, should there be a warn here?

henry-cask: I dont think we need to add any log in here since we handle it with correct value.

anwar6953: ok

henry-cask:
Let's compromise and set to 180 seconds by default.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
henry-cask: I could add debug log to make it less noisy but still give use some trackability.

chtyim:
LGTM

henry-cask:
Build passes: http://builds.cask.co/browse/CDAP-RBT392-4

henry-cask:
Merging ...

PR #3416, opened 2015-07-30
Fixes json error validation to skip properties that are added in imported json & not from backend

ajainarayanan:
- When importing json in UI the user might construct the json by hand and they may add a property that is not in the backend. In that case the UI should not fail and should skip in sending that property to the backend.
- Minor js error to fix when we check if a property is required when it is not from the backend (in the malformed json being imported)


elfenheart:
:+1: LGTM

PR #3418, opened 2015-07-30
Doc build fixes

awholegunch:
- Changes the title of a column in a table in the Installation Guide to read "Supported Versions"
- Turns on the guards for two included files in the Developers' Manual
- Corrects one of the uses of those files to read correctly
- Adds error checking to the download of remote repo files so that failure to find a file doesn't result in mysterious doc build errors


cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -147,7 +147,7 @@ Hadoop/HBase Environment
 For a distributed enterprise, you must install these Hadoop components:
 
 +---------------+-------------------+-----------------------------------------------------+
-| Component     | Source            | Required Version                                    |
+| Component     | Source            | Supported Versions                                  |
 +===============+===================+=====================================================+
 | **HDFS**      | Apache Hadoop     | 2.0.2-alpha through 2.5.0                           |
```
wolf31o2: HDP 2.2 is Hadoop 2.6, which we support...

sreevatsanraman:
LGTM

PR #3420, opened 2015-07-30
Changing reference from MySQL to RDBMS in error messages

nitinmotgi:
As DBUtils handles all JDBC compatible RDBMS and it's not just limited to MySQL.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/DBUtils.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/DBUtils.java
@@ -39,7 +39,7 @@ public final class DBUtils {
    * @param driverClass the JDBC driver class
    */
   public static void cleanup(Class<? extends Driver> driverClass) {
-    shutDownMySQLAbandonedConnectionCleanupThread(driverClass.getClassLoader());
+    shutDownRDBMSAbandonedConnectionCleanupThread(driverClass.getClassLoader());
   }
 
   /**
@@ -56,7 +56,7 @@ public final class DBUtils {
       driverField.setAccessible(true);
       Driver d = (Driver) driverField.get(driverInfo);
       if (d == null) {
-        LOG.debug("Found null driver object in drivers list. Ignoring.");
+        LOG.debug("Found null driver object in drivers list. Skipping.");
         continue;
       }
       LOG.debug("Removing non-null driver object from drivers list.");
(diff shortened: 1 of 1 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/DBUtils.java
```diff
@@ -88,10 +88,10 @@ private static void shutDownMySQLAbandonedConnectionCleanupThread(ClassLoader cl
       Class<?> mysqlCleanupThreadClass = classLoader.loadClass("com.mysql.jdbc.AbandonedConnectionCleanupThread");
       Method shutdownMethod = mysqlCleanupThreadClass.getMethod("shutdown");
       shutdownMethod.invoke(null);
-      LOG.info("Successfully shutdown MySQL connection cleanup thread.");
+      LOG.info("Successfully shutdown RDBMS connection cleanup thread.");
```
gokulavasan: Looks like the class we are trying to load is a MySQL class?

bdmogal: yeah, I don't think we should do this. Other DBs do not have this thread to cleanup. Unfortunately there is no way to check if it is mysql that we're running with, without adding an explicit dependency on it. We should probably just suppress the stack trace and move it to loglevel ``trace``?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/DBUtils.java
```diff
@@ -88,10 +88,10 @@ private static void shutDownMySQLAbandonedConnectionCleanupThread(ClassLoader cl
...
     } catch (Throwable e) {
       // cleanup failed, ignoring silently
-      LOG.warn("Failed to shutdown MySQL connection cleanup thread. Ignoring.");
+      LOG.warn("Failed to shutdown RDBMS connection cleanup thread. Ignoring.");
```
gokulavasan: The old LOG message should be improved since this might be confusing if someone is using RDBMS other than MySQL.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/DBUtils.java
chtyim: We know if it is running MySQL right since we are doing loadClass("mySqlClassHere"), right?

bdmogal:
Closing this in favor of https://github.com/caskdata/cdap/pull/3421.

PR #3423, opened 2015-07-30
Dataset TTL correction

awholegunch:
Fix for https://issues.cask.co/browse/CDAP-3084

Corrects usage to show that TTL is (in these cases) specified in seconds, not milliseconds.

Related issue: https://issues.cask.co/browse/CDAP-2878

Passes local checkstyle.


anew:
I can't tell whether these are all the paces where TTL is mentioned in the docs. But these look good. 
+1

awholegunch:
Here is all the places in the docs with TTL. Some involve streams:

http://docs-staging.cask.co/cdap/3.1.0-SNAPSHOT/en/search.html?q=TTL&check_keywords=yes&area=default

PR #3432, opened 2015-07-30
CDAP-2735 shouldn't have removed table name since the sink uses it

albertshau:
Also removing descriptions since the descriptions from the backend
should be used.


gokulavasan:
LGTM :+1:

gokulavasan:
LGTM :+1:

PR #3436, opened 2015-07-30
Replace non-standard LIMIT with a WHERE clause in the query to get Re…

bdmogal:
…sultSetMetadata

Tested on Oracle DB and HsqlDB that this is a valid query.

1st step towards getting Oracle write support.


gokulavasan:
May be add a comment to say why we are executing such a query?

gokulavasan:
Comment then LGTM :+1:

gokulavasan:
Please open the PR against release

anwar6953:
yea, why is it like that?

bdmogal:
@anwar6953 the point of this query is to make a call to the table to retrieve its schema in the form of a ``ResultSetMetadata`` object. Actual queries happen via ``DBOutputFormat`` when the MapReduce program runs.

PR #3443, opened 2015-07-31
make docs target paths project generic

duddun:
This is a backport of PR 3362: (develop)

This change allows us to remove 'cdap' from the bamboo global variable for staging and docs paths and make the paths generic (using $PROJECT which are already deriving from pom file in this script).

Note: This change is being backported to releases 3.1 (PR 3431), 3.0 (PR 3442) and 2.8


wolf31o2:
:+1: LGTM

PR #3447, opened 2015-07-31
remove call to tempDir.mkDirs() because DirUtils.createTempDir on the…

anwar6953:
… previous line already creates it.

https://builds.cask.co/browse/CDAP-RBT393-3


poornachandra:
LGTM

PR #3451, opened 2015-07-31
CDAP 2892 : Deploy schedule triggers in paused state and get rid of lazyStart() scheduler hack

rohitsinha54:
This PR does the following: 
- CDAP 2892 : When a new app is deployed with workflow its schedules are in disabled state unless resumed by the user manually. This is true for both time and stream size based schedule.
- Gets rid of lazyStart() hack which was required to deploy schedules with a stopped scheduler to ensure that they are in paused state.

Approach:
The way achieve paused triggers in Quartz scheduler is through groups. [More Details] (https://issues.cask.co/browse/CDAP-2892?focusedCommentId=17594&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17594)
- During initialization of scheduler we create add an empty job within a group. This is needed as I was not able to find any direct way to create a group.
- This group is then paused which guarantees that any new triggers added to this group will also be added with paused state. We add all our new triggers to this paused group.
- When resume is called on a schedule we move the trigger to default group and resume it. This ensure that when cdap restarts again and scheduler gets initialized it will not pause the triggers which are resumed. 
- Any operation which needs TriggerKey use getGroupedTriggerKey() to get the appropriate TriggerKey as the key will be different depending on which group it belong to.

StreamSizeScheduler: The scheduler of this scheduler does not set activeState to  true anymore. They get activated when resume is called on them.

Build: http://builds.cask.co/browse/CDAP-DUT2386-4

***
Tasks pending for these changes 
- [x] Unit tests: The way we pause trigger according to groups changed to handle cdap restarts and my earlier unit tests are not valid anymore. Working on updating them now.
- [x] Testing on cluster with CDAP restart. 
- [x] Schedules need to be resumed in cdap-integration tests. 
- [ ] Test for HA

***
Testing: 
Following steps were followed to test the schedules across cdap restart to check that resumed schedules does not get paused when cdap  starts again. It was done on standalone and singlenode cluster and works well on both.
1. Deployed Purchase app with two schedules one being run every  3 minutes.
2. Made sure  that  the status of the schedule is suspended and the workflow does not start even if the schedule time of these paused schedule is reached.
3. Resume the 3 minute schedule and make sure the other (every day one) is still paused. 
4. Made sure that the workflow runs every 3 minutes. 
5. Stop cdap services. (I did it when the workflow was not running)
6. Restart cdap
7. Checked that the 3 min schedule is in scheduled state and every day one is paused. 
8. Redeploy the app again and this time it has one more schedule which (run every 2 minutes) - Apparently this is the only way too add schedules for now.
9. Made sure that after redeploy every 3 min is still scheduled and keeps running the workflow and every 2 min is suspended state
10. Resume every 2 min schedule and made sure it starts workflow.
11. Suspend the every 3 min schedule and made sure it does not start workflow.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/AbstractSchedulerService.java
@@ -23,8 +23,6 @@ import co.cask.cdap.api.schedule.ScheduleSpecification;
 import co.cask.cdap.app.store.Store;
 import co.cask.cdap.common.ApplicationNotFoundException;
 import co.cask.cdap.common.NotFoundException;
-import co.cask.cdap.common.conf.CConfiguration;
-import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.internal.schedule.StreamSizeSchedule;
 import co.cask.cdap.internal.schedule.TimeSchedule;
 import co.cask.cdap.proto.Id;
@@ -51,31 +49,21 @@ public abstract class AbstractSchedulerService extends AbstractIdleService imple
   private static final Logger LOG = LoggerFactory.getLogger(AbstractSchedulerService.class);
   private final TimeScheduler timeScheduler;
   private final StreamSizeScheduler streamSizeScheduler;
-  private final CConfiguration cConf;
   private final Store store;
 
-  public AbstractSchedulerService(TimeScheduler timeScheduler, StreamSizeScheduler streamSizeScheduler,
-                                  CConfiguration cConf, Store store) {
+  public AbstractSchedulerService(TimeScheduler timeScheduler, StreamSizeScheduler streamSizeScheduler, Store store) {
     this.timeScheduler = timeScheduler;
     this.streamSizeScheduler = streamSizeScheduler;
-    this.cConf = cConf;
     this.store = store;
-  }
-
-  private boolean isLazyStart() {
-    return cConf.getBoolean(Constants.Scheduler.SCHEDULERS_LAZY_START, false);
   }
 
   /**
    * Start the scheduler services, by initializing them and starting them
-   * right away if lazy start is not active.
    */
   protected final void startSchedulers() throws SchedulerException {
     try {
       timeScheduler.init();
-      if (!isLazyStart()) {
-        timeScheduler.lazyStart();
-      }
+      timeScheduler.start();
       LOG.info("Started time scheduler");
     } catch (Throwable t) {
       Throwables.propagateIfInstanceOf(t, SchedulerException.class);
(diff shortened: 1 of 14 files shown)


rohitsinha54:
Will re-open after doing concurrency fixes.

rohitsinha54:
Addressing @anew comment from the previous PR about concurrency issues.
Earlier we used to store all our triggerKeys in DEFAULT group. Now we are using multiple groups. Initially, the keys are stored in NEW group (group which is always paused) and when a triggerKey is resumed for the first time we move it from NEW to DEFAULT group. Now this leads to a lot of concurrency issues since our TimerScheduler is not synchronized.
For example, a thread can call resume() which will get the key from NEW group and move it to DEFAULT group but in between this another thread  can call getKey method and get the key name with NEW group and before it actually operate with this key the key can be moved by earlier thread leading to failure in update.
I discussed an initial approach with @anew to make getGroupedTriggerKey() and resume() synchronized which will guarantee that when move is happening no one else can call getGroupedTriggerKey() unless the move is completed and decided that this will work. 
But after some more thought over the weekend I believe that this approach will not work in all cases. For example, say thread A executes updates which executes getGroupedTriggerKey() under sync lock and gets a key which is under NEW group before it actually updates the trigger associated with this key thread B executes resume() under sync (it will be able to as update is not sync method) and gets the key name with NEW group and moves it. Now thread A continues and try to update the trigger through the trigger key which is moved to DEFAULT group now and this will fail with trigger not found even though the trigger is there but just in another group.

So, it looks like we will have to make update(), suspend etc method sync too besides just resume() which puts a lot overhead.

If this overhead is too restricting then we can take another approach in which we keep using just one group (the default  group) and call pause on the default group. But we will call pause on this group every time CDAP starts to make sure that the new trigger which are added are paused too but calling pause will pause all triggers (even the ones which were resumed before CDAP was restarted). So before calling pause on this group we will have to iterate through all the key to check which are in resumed state and then resume them back after calling pause on the DEFAULT group. 

@anew: Thoughts ?

rohitsinha54:
Discussed in person, we will make all the methods synchronized.

anew:
I don't think it is acceptable to pause all schedules when one schedule is added. That would mean that once in a while a schedule does not run just because some other schedule is added or updated. 

One question: Is it possible to add a new schedule in the "paused" group, then move it to the default group without resuming it? In that case the "paused " group only gets involved when new schedules are created.

Another option: Create the schedule with a trigger that never fires, or that would only fire long into the future. Then pause it. Then updateSchedule() with the actual schedule to be created. That way the "paused" group is not needed at all.

rohitsinha54:
@anew :
Yes pausing all triggers while adding one is not a good way to go through it. Agreed. 

The thought of moving from paused to default while creating did occurred to me but it does not work as the trigger state is not a property associated with the trigger. Whenever we create a trigger before being added the scheduler checks if it belongs to paused group and then pauses it.

I looked into the the suggestion of deploying a trigger key which never fires and then updating it. But sadly this will not work. We we updateSchedule we create a new trigger key from the existing which leads us back to the earlier problem in which a new trigger key can only be paused if its added to a paused group.

anew:
I think it is ok for now to synchronize all methods because the Quartz scheduler will synchronize anyway, and therefore we do not give up much concurrency.

rohitsinha54:
Build: http://builds.cask.co/browse/CDAP-DUT2676-2

anew:
@rohitsinha54 please see my comments (I made them directly on the last two commits).

rohitsinha54:
@anew: Fix comments. Please have a look. Will resolve conflicts at the end of the review.

anew:
Ok, since that was the only concern I had left, this is now good to merge. Please resolve conflicts and merge after build succeeds.

rohitsinha54:
Build running: http://builds.cask.co/browse/CDAP-DUT2676-5

rohitsinha54:
Fixing integration tests before merge.

rohitsinha54:
Last build failed due to a new test case. New one running here: http://builds.cask.co/browse/CDAP-DUT2426-3

PR #3457, opened 2015-07-31
exclude all hive jars from hive's mapreduce containers

gsps1:
https://issues.cask.co/browse/CDAP-2918


gsps1:
build : http://builds.cask.co/browse/CDAP-RBT400-1

PR #3465, opened 2015-07-31
[CDAP-3184] Only cache the bare-minimum requirements from the ResultS…

bdmogal:
…etMetadata object (column type mappings) and store them in a separate object,

instead of caching the ResultSetMetadata object itself throughout the map task.

This reduces dependencies on the JDBC ResultSetMetadata after the initialize() stage, making DBSink more robust towards conflicting JDBC driver
behavior.

Jira: [CDAP-3184](https://issues.cask.co/browse/CDAP-3184)
Build: http://builds.cask.co/browse/CDAP-RBT402


bdmogal:
Opened against wrong branch. closing.

PR #3468, opened 2015-07-31
add type map<string, string> to schema and stream properties + add ts…

elfenheart:
… & headers implicit schema to stream source


ajainarayanan:
:+1: LGTM

PR #3469, opened 2015-07-31
[CDAP-3182] TableSource should not have case.sensitive.row.key property

bdmogal:
Separated out configs for TableSource and TableSink (both Batch and Realtime)

Tested on standalone

Jira: [CDAP-3182](https://issues.cask.co/browse/CDAP-3182)
Build: http://builds.cask.co/browse/CDAP-RBT403


gokulavasan:
LGTM :+1:

bdmogal:
http://builds.cask.co/browse/CDAP-RBT403-2. Merging.

PR #3472, opened 2015-07-31
Fixes DAG traversal to construct config based on connections in random order

ajainarayanan:
- Add nodes in any random order.
- Form connections with/without source and sink.
- Re-order/change connections in anyway.

The above conditions should all be supported in UI and final config formed should be correct. There should be no restriction in the order in which we add nodes nor the order in which we make connections. The final connection should be the way we represent in UI.


elfenheart:
:+1: LGTM

PR #3478, opened 2015-07-31
Fix/ui adapters import2

ajainarayanan:
- Fixes admin apps list and adapter drafts list to not show more than one alert at a time.
- Fixes importing/cloning of etlrealtime apps.


elfenheart:
:+1: LGTM

PR #3479, opened 2015-07-31
CDAP-3187 metrics should be unique even if name is equal

albertshau:
Adding the stage number to the metrics context so that two filters
in the same pipeline will emit different metrics.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/ETLTemplate.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/ETLTemplate.java
@@ -74,6 +74,7 @@ public abstract class ETLTemplate<T extends ETLConfig> extends ApplicationTempla
     ETLStage sinkConfig = etlConfig.getSink();
     List<ETLStage> transformConfigs = etlConfig.getTransforms();
     String sourcePluginId = PluginID.from(Constants.Source.PLUGINTYPE, sourceConfig.getName(), 1).getID();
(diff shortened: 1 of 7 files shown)


gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-RBT405

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/ETLTemplate.java
```diff
@@ -73,10 +73,9 @@ public void configureAdapter(String adapterName, T etlConfig, AdapterConfigurer
     ETLStage sourceConfig = etlConfig.getSource();
     ETLStage sinkConfig = etlConfig.getSink();
     List<ETLStage> transformConfigs = etlConfig.getTransforms();
-    String sourcePluginId = String.format("%s%s%s", Constants.Source.PLUGINTYPE, Constants.ID_SEPARATOR,
-                                          sourceConfig.getName());
-    String sinkPluginId = String.format("%s%s%s", Constants.Sink.PLUGINTYPE, Constants.ID_SEPARATOR,
-                                        sinkConfig.getName());
+    String sourcePluginId = PluginID.from(Constants.Source.PLUGINTYPE, sourceConfig.getName(), 1).getID();
+    String sinkPluginId =
+      PluginID.from(Constants.Sink.PLUGINTYPE,  sinkConfig.getName(), 2 + transformConfigs.size()).getID();
```
gsps1: can you add a comment why we are adding 2 ?

PR #3482, opened 2015-07-31
Fixes 'Add App' dropdown in firefox to be aligned properly + removes …

ajainarayanan:
…custom padding on buttons in the home page


elfenheart:
:+1: LGTM

PR #3485, opened 2015-07-31
Fix the url for StreamClient#asyncSendEvent

anwar6953:
For some reason, the url that StreamClient's async send event method hits was modified in this recent commit:
https://github.com/caskdata/cdap/commit/318bf1154511f4ba14d46bb090dad39e43312a42#diff-f61d08dd39129279edde2e03ec7f9dcfR444

There is a test case for this method, but it is difficult to assert that the async endpoint was hit, because even for stream events of size 1-2MB, it's nearly effectively synchronous.

http://builds.cask.co/browse/CDAP-RBT408-2


jwang47:
LGTM

PR #3486, opened 2015-07-31
CDAP-3232 fix stream writer to persist headers

albertshau:
prefix headers with the stream name so that they are interpreted
as stream headers instead of request headers.


albertshau:
http://builds.cask.co/browse/CDAP-RBT409

gokulavasan:
LGTM :+1:

PR #3492, opened 2015-08-01
[CDAP-3188] [CDAP-3231] [CDAP-3236] [CDAP-3245] Fix/ui explore download queries

elfenheart:
- Fix downloading large query
- Fix missing description on created apps
- Fix nullable schema propagation
- Fix workflow schedules view when there is no schedule

### DIFF
--- a/cdap-ui/server/express.js
+++ b/cdap-ui/server/express.js
@@ -35,6 +35,7 @@ function makeApp (authAddress, cdapConfig) {
   catch(e) { log.error('Favicon missing! Please run `gulp build`'); }
   app.use(compression());
   app.use(bodyParser.json());
+  app.use(bodyParser.urlencoded({ extended: false }));
   app.use(cookieParser());
 
   // serve the config file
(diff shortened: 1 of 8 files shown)


cdap-ui/server/express.js
```diff
@@ -60,6 +61,42 @@ function makeApp (authAddress, cdapConfig) {
...
+  app.post('/downloadQuery', function(req, res) {
+    var fs = require('fs');
+    var StringDecoder = require('string_decoder').StringDecoder;
+    var url = req.body.backendUrl;
+    var query = req.body.queryHandle;
+
+    var path = DIST_PATH + '/assets/public';
+
+    fs.mkdir(path, function (err) {
+      if (err && err.code === 'EEXIST') {
```
anwar6953: doesn't seem like we need an error handling callback at all?
(since it's just a no-op)

elfenheart: Unfortunately we do. If I don't specify the error callback, the node server crash

anwar6953: ok

ajainarayanan:
:+1: LGTM

PR #3493, opened 2015-08-01
Fix/ui adapter imports 3

ajainarayanan:
- Cleans up importing and editing of drafts 
- Fixes drawing DAGs in already created adapters.
- Suppresses 404 error messages when adapter config is not found.


elfenheart:
:+1: LGTM

PR #3496, opened 2015-08-01
Fix/ui fix docsurl

ajainarayanan:
- Fixes info button url to be dynamic.
- Reduces the time we show any alert in UI.


elfenheart:
:+1: LGTM

PR #3497, opened 2015-08-01
Fix/ui minor improvements

elfenheart:
- Give more space to plugin description on property edit modal
- Fix Focus when opening modal (doesn't scroll to the bottom)
- Stream default format to text. If no schema is defined, propagate ts, headers, body

somefixes


ajainarayanan:
:+1: LGTM

PR #3500, opened 2015-08-01
StructuredRecordToPutTransformer, StructuredRecordToAvro Transformer

gokulavasan:
Takes in an output schema and filters out fields that are not present in it.

This allows the sink to drop fields if the schema string that it gets has fields removed compared to the previous stage. Before this change, we were converting the structured record we get directly to Put or AvroRecord without taking into consideration the output schema that was provided (output schema was used only for explore). 

Build : http://builds.cask.co/browse/CDAP-RBT413-1

JIRA :

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/TableSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/TableSink.java
@@ -20,6 +20,7 @@ import co.cask.cdap.api.annotation.Description;
 import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.dataset.table.Put;
 import co.cask.cdap.api.dataset.table.Table;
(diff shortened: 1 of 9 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/TableSink.java
```diff
@@ -61,7 +62,14 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
   @Override
   public void initialize(BatchSinkContext context) throws Exception {
     super.initialize(context);
-    recordPutTransformer = new RecordPutTransformer(tableSinkConfig.getRowField(),
+    Schema outputSchema = null;
```
sreevatsanraman: What happens when you don't specify any output schema? Will everything be filtered out? Can we add a test case for that?

gokulavasan: If outputSchema is null, all the fields from the StructuredRecord are 'put' into Put. This was the behavior earlier. The old tests already test for this. The test that I have added also tests again for the case of outputSchema being null.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/common/RecordPutTransformerTest.java
```diff
@@ -121,11 +121,31 @@ public void testCaseInsensitiveRowKeyValid() {
...
+                                         Schema.Field.of("age", Schema.of(Schema.Type.INT)));
+    StructuredRecord record = StructuredRecord.builder(inputSchema)
+      .set("id", 123L).set("name", "ABC").set("age", 10).build();
+    RecordPutTransformer transformer = new RecordPutTransformer("id", outputSchema, true);
+    Put put = transformer.toPut(record);
+    Assert.assertEquals(1, put.getValues().size());
+
+    transformer = new RecordPutTransformer("id", null, true);
+    put = transformer.toPut(record);
+    Assert.assertEquals(2, put.getValues().size());
```
gsps1: can you add one more output schema without any fields?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/TableSink.java
sreevatsanraman: ok

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/template/etl/common/RecordPutTransformerTest.java
sreevatsanraman: let's add this test to develop/

sreevatsanraman:
LGTM :+1:

PR #3504, opened 2015-08-03
CDAP-2521: CLI autocomplete for stream format

jwang47:
https://issues.cask.co/browse/CDAP-2521
http://builds.cask.co/browse/CDAP-DUT2442


gokulavasan:
LGTM :+1:

PR #3505, opened 2015-08-03
[CDAP-3251] Remove classloader approach to create instance of Apache Twill ZKBrokerService class

henry-cask:
Currently we use hacky classloader way to instantiate Apache Twill ZKBrokerService helper class in
Kafka08SimpleApiConsumer class because the constructor was package private.

With updated dependency on Apache Twill in CDAP we now could just simply call the constructor.


chtyim:
LGTM

PR #3506, opened 2015-08-03
CDAP-3147: CLI stop workflow

jwang47:
https://issues.cask.co/browse/CDAP-3147
http://builds.cask.co/browse/CDAP-DUT2443

### DIFF
--- a/cdap-cli-tests/src/test/java/co/cask/cdap/cli/CLIMainTest.java
+++ b/cdap-cli-tests/src/test/java/co/cask/cdap/cli/CLIMainTest.java
@@ -523,6 +523,9 @@ public class CLIMainTest {
     testCommandOutputContains(
       cli, String.format("get workflow token %s %s at node %s scope user key %s", workflow, runId,
                          FakeWorkflow.FakeAction.ANOTHER_FAKE_NAME, FakeWorkflow.FakeAction.TOKEN_KEY), fakeNodeValue);
+
+    // stop workflow
+    testCommandOutputContains(cli, "stop workflow " + workflow, "400: Program not running");
   }
 
   private static File createAppJarFile(Class<?> cls) throws IOException {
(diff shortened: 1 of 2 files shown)


cdap-cli-tests/src/test/java/co/cask/cdap/cli/CLIMainTest.java
```diff
@@ -523,6 +523,9 @@ public void testWorkflowToken() throws Exception {
     testCommandOutputContains(
       cli, String.format("get workflow token %s %s at node %s scope user key %s", workflow, runId,
                          FakeWorkflow.FakeAction.ANOTHER_FAKE_NAME, FakeWorkflow.FakeAction.TOKEN_KEY), fakeNodeValue);
+
+    // stop workflow
+    testCommandOutputContains(cli, "stop workflow " + workflow, "400: Program not running");
```
bdmogal: Do we also have a test that stops a workflow successfully?

jwang47: Not right now.. I'm not sure how to construct that test properly.

jwang47: However, I don't think that needs to be tested here, as it's more of an integration test for workflows. What's being tested here is whether the CLI is hitting the correct HTTP endpoint.

bdmogal: Do you think it makes sense to do a stop + restart just before line 493 in this test? Although I'm not sure if that will make the test flaky since there is no way to tell how long the fake workflow actually runs?

bdmogal:
One question about a test scenario, otherwise LGTM

cdap-cli-tests/src/test/java/co/cask/cdap/cli/CLIMainTest.java
jwang47: It seems that would make the test flaky. Not sure how to fix it right now.

PR #3508, opened 2015-08-03
CDAP-1936: Minimal instrumentation of logs processing

abaranau:
* # of processed messages
* lag of processing, in ms

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/logging/save/KafkaMessageCallback.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/logging/save/KafkaMessageCallback.java
@@ -17,6 +17,8 @@
 package co.cask.cdap.logging.save;
 
 import ch.qos.logback.classic.spi.ILoggingEvent;
+import co.cask.cdap.api.metrics.MetricsContext;
+import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.logging.LoggingContext;
 import co.cask.cdap.logging.appender.kafka.LoggingEventSerializer;
 import co.cask.cdap.logging.context.LoggingContextHelper;
@@ -41,12 +43,15 @@ public class KafkaMessageCallback implements KafkaConsumer.MessageCallback {
   private final Set<KafkaLogProcessor> kafkaLogProcessors;
   private final LoggingEventSerializer serializer;
   private final CountDownLatch stopLatch;
+  private final MetricsContext metricsContext;
 
   public KafkaMessageCallback(CountDownLatch stopLatch,
-                              Set<KafkaLogProcessor> kafkaLogProcessors) throws Exception {
+                              Set<KafkaLogProcessor> kafkaLogProcessors,
+                              MetricsContext metricsContext) throws Exception {
     this.kafkaLogProcessors = kafkaLogProcessors;
     this.serializer = new LoggingEventSerializer();
     this.stopLatch = stopLatch;
+    this.metricsContext = metricsContext;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


abaranau:
http://builds.cask.co/browse/CDAP-DUT2445-1

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/KafkaMessageCallback.java
```diff
@@ -66,12 +71,17 @@ public void onReceived(Iterator<FetchedMessage> messages) {
...
     int count = 0;
 
+    long oldestProcessed = 0;
     while (messages.hasNext()) {
       FetchedMessage message = messages.next();
       try {
         GenericRecord genericRecord = serializer.toGenericRecord(message.getPayload());
         ILoggingEvent event = serializer.fromGenericRecord(genericRecord);
 
+        if (event.getTimeStamp() < oldestProcessed) {
```
chtyim: oldestProcessed is initialized with "0", meaning this will never be true?

abaranau: yes - it is timestamp. We also check for count - if we don't process any messages, we don't report

abaranau: fixed

abaranau:
@chtyim look one more time please

abaranau:
GREEN

chtyim:
LGTM

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/KafkaMessageCallback.java
gsps1: we should probably do this after we call ``processor.process(logEvent)`` -- If we get 100 fetched messages and the last one has a old timestamp and if there were exception while processing that event, we will still gauge for that timestamp - which is not accurate as that timestamp event was not processed successfully. might not happen, but better to do there i guess.

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/LogSaver.java
```diff
@@ -67,6 +73,9 @@ public LogSaver(KafkaClientService kafkaClient,
     this.kafkaCancelMap = new HashMap<>();
     this.kafkaCancelCallbackLatchMap = new HashMap<>();
     this.messageProcessors = messageProcessors;
+
+    this.metricsContext = metricsCollectionService.getContext(
+      ImmutableMap.of(Constants.Metrics.Tag.COMPONENT, Constants.Service.LOGSAVER));
```
gsps1: should we include the instance-id of the log.saver ? so we can see the delay and processing metrics for each instance of log.saver?

gsps1:
@abaranau couple of comments.

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/LogSaver.java
abaranau: we do not aggregate for instances for system services right now.. so it'll be bigger change. CDAP-3265

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/KafkaMessageCallback.java
abaranau: with the current code we show what is the lag based on the fetched log messages from kafka. I believe it is fine.

PR #3511, opened 2015-08-04
CDAP-1618: Do schedule updates in proper way

rohitsinha54:
- Uses rescheduleJob() from Scheduler to update schedules
- Earlier approach of deleting the schedule and creating new one was not safe for concurrency. For example, thread A calls updateSchedule() and deletes the schedule() and before new one is added thread B calls checkStatus() or some other function which will fail to find the trigger even though the key is valid and exists.
- The rescheduleJob() function perform the reschedule synchronously which guarantees that while the schedule is being updated all other threads have to wait and hence will not lead to any concurrent access issues.
- Added a unit tests in SchedulerService to test updateSchedule.

Issue: https://issues.cask.co/browse/CDAP-1618
Build: http://builds.cask.co/browse/CDAP-DUT2449-1

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
@@ -262,7 +262,6 @@ final class TimeScheduler implements Scheduler {
   }
 
   @Override
-
   public void updateSchedule(Id.Program program, SchedulableProgramType programType, Schedule schedule,
                              Map<String, String> properties) throws NotFoundException, SchedulerException {
     checkInitialized();
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
@@ -267,8 +267,7 @@ final class TimeScheduler implements Scheduler {
     checkInitialized();
     try {
       Trigger trigger = getTrigger(program, programType, schedule.getName());
-      @SuppressWarnings("unchecked")
-      TriggerBuilder<Trigger> triggerBuilder = (TriggerBuilder<Trigger>) trigger.getTriggerBuilder();
+      TriggerBuilder triggerBuilder = trigger.getTriggerBuilder();
       String cronEntry =  ((TimeSchedule) schedule).getCronEntry();
 
       // create the new trigger with the new schedule schedule all other fields will remain unmodified
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
@@ -166,11 +166,7 @@ final class TimeScheduler implements Scheduler {
         .withSchedule(CronScheduleBuilder
                         .cronSchedule(getQuartzCronExpression(cronEntry))
                         .withMisfireHandlingInstructionDoNothing());
-      if (properties != null) {
-        for (Map.Entry<String, String> entry : properties.entrySet()) {
-          trigger.usingJobData(entry.getKey(), entry.getValue());
-        }
-      }
+      addProperties(trigger, properties);
       try {
         scheduler.scheduleJob(trigger.build());
       } catch (org.quartz.SchedulerException e) {
@@ -266,11 +262,25 @@ final class TimeScheduler implements Scheduler {
   }
 
   @Override
+
   public void updateSchedule(Id.Program program, SchedulableProgramType programType, Schedule schedule,
                              Map<String, String> properties) throws NotFoundException, SchedulerException {
-    // TODO modify the update flow [CDAP-1618]
-    deleteSchedule(program, programType, schedule.getName());
-    schedule(program, programType, schedule, properties);
+    checkInitialized();
+    try {
+      Trigger trigger = getTrigger(program, programType, schedule.getName());
+      @SuppressWarnings("unchecked")
+      TriggerBuilder<Trigger> triggerBuilder = (TriggerBuilder<Trigger>) trigger.getTriggerBuilder();
+      String cronEntry =  ((TimeSchedule) schedule).getCronEntry();
+
+      // create the new trigger with the new schedule schedule all other fields will remain unmodified
+      triggerBuilder.withSchedule(CronScheduleBuilder
+                                    .cronSchedule(getQuartzCronExpression(cronEntry))
+                                    .withMisfireHandlingInstructionDoNothing());
+      addProperties(triggerBuilder, properties);
+      scheduler.rescheduleJob(trigger.getKey(), triggerBuilder.build());
+    } catch (org.quartz.SchedulerException e) {
+      throw new SchedulerException(e);
+    }
   }
 
   @Override
@@ -278,11 +288,7 @@ final class TimeScheduler implements Scheduler {
     throws NotFoundException, SchedulerException {
     checkInitialized();
     try {
-      Trigger trigger = scheduler.getTrigger(
-        new TriggerKey(AbstractSchedulerService.scheduleIdFor(program, programType, scheduleName)));
-      if (trigger == null) {
-        throw new ScheduleNotFoundException(Id.Schedule.from(program.getApplication(), scheduleName));
-      }
+      Trigger trigger = getTrigger(program, programType, scheduleName);
 
       scheduler.unscheduleJob(trigger.getKey());
 
@@ -394,5 +400,29 @@ final class TimeScheduler implements Scheduler {
         }
       }
     };
+  }
+
+  /**
+   * Gets a {@link Trigger} associated with this program name, type and schedule name
+   */
+  private Trigger getTrigger(Id.Program program, SchedulableProgramType programType,
+                             String scheduleName) throws org.quartz.SchedulerException, ScheduleNotFoundException {
+    Trigger trigger = scheduler.getTrigger(
+      new TriggerKey(AbstractSchedulerService.scheduleIdFor(program, programType, scheduleName)));
+    if (trigger == null) {
+      throw new ScheduleNotFoundException(Id.Schedule.from(program.getApplication(), scheduleName));
+    }
+    return  trigger;
+  }
+
+  /**
+   * Adds properties to a {@link TriggerBuilder} to used by the {@link Trigger}
+   */
+  private void addProperties(TriggerBuilder trigger, Map<String, String> properties) {
+    if (properties != null) {
+      for (Map.Entry<String, String> entry : properties.entrySet()) {
+        trigger.usingJobData(entry.getKey(), entry.getValue());
+      }
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -266,11 +262,25 @@ public void updateSchedule(Id.Program program, SchedulableProgramType programTyp
   }
 
   @Override
+
```
poornachandra: Extra newline.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
```diff
@@ -166,6 +169,36 @@ public void testSimpleSchedulerLifecycle() throws Exception {
...
+    List<String> scheduleIds = schedulerService.getScheduleIds(program, programType);
+    Assert.assertEquals(1, scheduleIds.size());
+    checkState(Scheduler.ScheduleState.SUSPENDED, scheduleIds);
+
+    schedulerService.updateSchedule(program, programType, updatedTimeSchedule1);
+
+    schedulerService.resumeSchedule(program, programType, "Schedule1");
+    checkState(Scheduler.ScheduleState.SCHEDULED, scheduleIds);
+
+    List<ScheduledRuntime> scheduledRuntimes = schedulerService.nextScheduledRuntime(program, programType);
```
poornachandra: It would be a good idea to check the `nextScheduledRuntime` once before to assert the first schedule.

poornachandra:
@rohitsinha54 Just two comments, please address them.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/TimeScheduler.java
```diff
@@ -268,9 +264,22 @@ public void updateSchedule(Id.Program program, SchedulableProgramType programTyp
   @Override
   public void updateSchedule(Id.Program program, SchedulableProgramType programType, Schedule schedule,
                              Map<String, String> properties) throws NotFoundException, SchedulerException {
-    // TODO modify the update flow [CDAP-1618]
-    deleteSchedule(program, programType, schedule.getName());
-    schedule(program, programType, schedule, properties);
+    checkInitialized();
+    try {
+      Trigger trigger = getTrigger(program, programType, schedule.getName());
+      @SuppressWarnings("unchecked")
```
anew: this is not wrong. But this is done differently in schedule(). It would be nice follow the same style within the same file

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
```diff
@@ -165,6 +168,41 @@ public void testSimpleSchedulerLifecycle() throws Exception {
...
+
+    schedulerService.resumeSchedule(program, programType, "Schedule1");
+    checkState(Scheduler.ScheduleState.SCHEDULED, scheduleIds);
+
+    List<ScheduledRuntime> updatedScheduledRuntimes = schedulerService.nextScheduledRuntime(program, programType);
+
+    // the next schedule runtime should be in next 10 minutes from resume
+    Assert.assertTrue(updatedScheduledRuntimes.get(0).getTime() > System.currentTimeMillis() &&
+                        updatedScheduledRuntimes.get(0).getTime() < System.currentTimeMillis() +
+                          TimeUnit.MINUTES.toMillis(12));
```
anew: this does not match the comment (that it must be within 10 minutes).

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
```diff
@@ -165,6 +168,41 @@ public void testSimpleSchedulerLifecycle() throws Exception {
...
+    store.addApplication(appId, applicationSpecification, locationFactory.create("app"));
+    List<String> scheduleIds = schedulerService.getScheduleIds(program, programType);
+    Assert.assertEquals(1, scheduleIds.size());
+    checkState(Scheduler.ScheduleState.SUSPENDED, scheduleIds);
+
+    List<ScheduledRuntime> oldScheduledRuntimes = schedulerService.nextScheduledRuntime(program, programType);
+
+    schedulerService.updateSchedule(program, programType, updatedTimeSchedule1);
+
+    schedulerService.resumeSchedule(program, programType, "Schedule1");
```
anew: why does it have to be resumed?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
```diff
@@ -63,6 +64,8 @@
     Schedules.createDataSchedule("Schedule3", "Every 1M", Schedules.Source.STREAM, STREAM_ID.getId(), 1);
   private static final Schedule dataSchedule2 =
     Schedules.createDataSchedule("Schedule4", "Every 10M", Schedules.Source.STREAM, STREAM_ID.getId(), 10);
+  private static final Schedule  updatedTimeSchedule1 =
```
anew: static final variables must be all upper case

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
```diff
@@ -165,6 +168,41 @@ public void testSimpleSchedulerLifecycle() throws Exception {
...
   }
 
+  @Test
+  public void testScheduleUpdate() throws Exception {
+    AppFabricTestHelper.deployApplication(namespace, AppWithWorkflow.class);
+    ApplicationSpecification applicationSpecification = store.getApplication(appId);
+
+    schedulerService.schedule(program, programType, ImmutableList.of(timeSchedule1));
+    applicationSpecification = createNewSpecification(applicationSpecification, program, programType, timeSchedule1);
+    store.addApplication(appId, applicationSpecification, locationFactory.create("app"));
```
anew: it is not clear to me why this test needs to manipulate the store.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
```diff
@@ -165,6 +168,41 @@ public void testSimpleSchedulerLifecycle() throws Exception {
...
+    List<String> scheduleIds = schedulerService.getScheduleIds(program, programType);
+    Assert.assertEquals(1, scheduleIds.size());
+    checkState(Scheduler.ScheduleState.SUSPENDED, scheduleIds);
```
anew: this assumes that after schedule(), the schedule is initially suspended?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
```diff
@@ -165,6 +168,41 @@ public void testSimpleSchedulerLifecycle() throws Exception {
...
+
+    List<ScheduledRuntime> updatedScheduledRuntimes = schedulerService.nextScheduledRuntime(program, programType);
+
+    // the next schedule runtime should be in next 10 minutes from resume
+    Assert.assertTrue(updatedScheduledRuntimes.get(0).getTime() > System.currentTimeMillis() &&
+                        updatedScheduledRuntimes.get(0).getTime() < System.currentTimeMillis() +
+                          TimeUnit.MINUTES.toMillis(12));
+
+    // the updated next schedule runtime must be less than the old one
+    Assert.assertTrue(updatedScheduledRuntimes.get(0).getTime() < oldScheduledRuntimes.get(0).getTime());
```
anew: why would it have to be smaller? If it is 1:55, then the next runtime will be at 2:00, regardless of whether the schedule is hourly or every 10 minutes.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
bdmogal: @anew this is because the scheduler APIs currently do not do any store updates. They expect the callers to handle store operations. We have https://issues.cask.co/browse/CDAP-1202 open for this. I had a PR - https://github.com/caskdata/cdap/pull/3084 to fix this, but we decided to not do it in 3.1, because it would need a bigger refactoring of the scheduler classes. 

Basically, the problem is that currently, ``AbstractSchedulerService``, and ``TimeScheduler`` and ``StreamSizeScheduler`` implement the same interface, but really have different responsibilities. So while doing this, we ended up expanding the scheduler apis with more operations that had to throw UnsupportedOperationException in some implementations.

We should pick up that PR again, but I haven't gotten to it since 3.1

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
anew: ok, after seeing the getCron() method, I think this is ok.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
rohitsinha54: Yes, in unit test we have lazy hack in which the triggers are paused when they are initially deployed. 
This PR gets rid of the lazy hack which we have in unit tests: PR https://github.com/caskdata/cdap/pull/3451

rohitsinha54:
While working on addressing comments and adding new unit test I found this bug: https://issues.cask.co/browse/CDAP-3380
Working on it and will include its fix in this PR.

rohitsinha54:
@anew : Addressed comments. 
Also during working on unit tests encountered a bug: https://issues.cask.co/browse/CDAP-3380
and the fix is included in this PR as its related to update of schedule. 

This commit is the new changes: https://github.com/caskdata/cdap/commit/de8fa51303a2bbe42c72dcc3dca997742a4fdde3

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
```diff
@@ -165,6 +172,70 @@ public void testSimpleSchedulerLifecycle() throws Exception {
...
 
+  @Test
+  public void testScheduleUpdate() throws Exception {
+    // test time schedule update
+    testScheduleUpdate(TIME_SCHEDULE_1, UPDATED_TIME_SCHEDULE_1, TimeUnit.MINUTES.toMillis(10));
+    // test stream size schedule update
+    testScheduleUpdate(DATA_SCHEDULE_2, UPDATED_DATA_SCHEDULE_2, null);
+  }
+
+  private void testScheduleUpdate(Schedule oldSchedule, Schedule newSchedule, @Nullable Long durationInMillis)
```
anwar6953: why is the last param nullable? It seems that its passed to `verifyUpdatedNextRuntime` below on line 225, on which it is not nullable.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
anew: ok. Let's make sure this gets done and not deferred until 2017 :)

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/schedule/SchedulerServiceTest.java
anew: ok makes sense

anew:
LGTM

sreevatsanraman:
Latest build passed and see an LGTM. Merging.

PR #3513, opened 2015-08-04
[CDAP-3253] Feature/ui widget stream dataset

elfenheart:
streamselector

### DIFF
--- a/cdap-ui/Gulpfile.js
+++ b/cdap-ui/Gulpfile.js
@@ -18,8 +18,7 @@ gulp.task('css:lib', ['fonts'], function() {
       './bower_components/font-awesome/css/font-awesome.min.css',
       './bower_components/ng-sortable/dist/ng-sortable.min.css',
       './bower_components/c3/c3.min.css',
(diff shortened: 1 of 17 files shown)


cdap-ui/Gulpfile.js
```diff
@@ -18,7 +18,8 @@ gulp.task('css:lib', ['fonts'], function() {
       './bower_components/font-awesome/css/font-awesome.min.css',
       './bower_components/ng-sortable/dist/ng-sortable.min.css',
       './bower_components/c3/c3.min.css',
-      './bower_components/angular-gridster/dist/angular-gridster.min.css'
+      './bower_components/angular-gridster/dist/angular-gridster.min.css',
+      './bower_components/angular-ui-select/dist/select.min.css'
```
ajainarayanan: We don't need ui.select anymore right?

ajainarayanan:
@elfenheart Do we need references of ui.select directive? We don't use it anymore right? Is it possible to remove them?

elfenheart:
@ajainarayanan I'll remove them

ajainarayanan:
:+1: LGTM

PR #3516, opened 2015-08-04
[CDAP-2592] update dependencies

elfenheart:
Distributed Singlenode: singlenode2012-1000
Secure: securecluster2452-1000


elfenheart:
putting a hold on this.. There's still some issue with authentication

elfenheart:
ready for review

cdap-ui/app/features/admin/controllers/namespace/create-ctrl.js
```diff
@@ -33,6 +32,11 @@ angular.module(PKG.name + '.feature.admin')
             myNamespace.getList(true).then(function() {
               EventPipe.emit('namespace.update');
             });
+
+            $timeout(function () {
+              $modalInstance.close();
```
ajainarayanan: Is ```$modalInstance.dismisAll()``` not working? is ```$timeout``` the only solution? Seems like a hack. Does the docs suggest anything?

elfenheart: Tried that, and it's still throwing the $$nextSibling is null error

ajainarayanan:
:+1: LGTM

PR #3519, opened 2015-08-04
[CDAP-2723] remove delimiters from view for projection transform

elfenheart:
screen shot 2015-08-04 at 3 43 31 pm


ajainarayanan:
:+1: LGTM

PR #3520, opened 2015-08-04
CDAP-3261 fixes for adding artifacts in distributed mode

albertshau:
Removing location serialization and deserialization through a gson
type adapter since it triggers an infinite loop in gson for some
reason. Also changing add artifact api a little bit to take an
input supplier for an input stream instead of just the input
stream to ensure that the stream is closed.


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -213,7 +215,12 @@ public void addArtifact(Id.Artifact artifactId, File artifactFile, @Nullable Set
     try {
       ArtifactClasses artifactClasses = artifactInspector.inspectArtifact(artifactId, artifactFile, parentClassLoader);
       ArtifactMeta meta = new ArtifactMeta(artifactClasses, parentArtifacts);
-      artifactStore.write(artifactId, meta, new FileInputStream(artifactFile));
+      artifactStore.write(artifactId, meta, new InputSupplier<InputStream>() {
```
chtyim: `Files.newInputStreamSupplier`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -125,7 +128,8 @@
   private static final Id.DatasetInstance META_ID =
     Id.DatasetInstance.from(Constants.SYSTEM_NAMESPACE, "artifact.meta");
 
-  private final NamespacedLocationFactory locationFactory;
+  private final LocationFactory locationFactory;
+  private final NamespacedLocationFactory namespacedLocationFactory;
```
chtyim: The naming for the `NamespacedLocationFactory` is really bad.. It is not a `LocationFactory` at all. Anyway not related to this PR, but that needs to be fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -581,19 +594,30 @@ private void addPluginsToMap(Id.Artifact parentArtifactId, SortedMap<ArtifactDes
...
     }
   }
 
   private ArtifactDescriptor getDescriptor(Id.Artifact artifactId, Location location) {
     return new ArtifactDescriptor(artifactId.getName(), artifactId.getVersion(),
-                                  Constants.SYSTEM_NAMESPACE_ID.equals(artifactId.getNamespace()), location);
+      Constants.SYSTEM_NAMESPACE_ID.equals(artifactId.getNamespace()), location);
+  }
+
+  private Location toLocation(String locationURI) {
```
chtyim: Why represent `URI` using string type but not `URI` type?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStoreTest.java
```diff
@@ -663,4 +665,14 @@ private void assertEqual(Id.Artifact expectedId, ArtifactMeta expectedMeta,
     Assert.assertEquals(expectedContents, CharStreams.toString(
       new InputStreamReader(actual.getDescriptor().getLocation().getInputStream(), Charsets.UTF_8)));
   }
+
+  private void writeArtifact(Id.Artifact artifactId, ArtifactMeta meta, final String contents)
+    throws ArtifactAlreadyExistsException, IOException, WriteConflictException {
+    artifactStore.write(artifactId, meta, new InputSupplier<InputStream>() {
```
chtyim: `ByteStreams.newInputStreamSupplier()`

chtyim:
Just couple comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
albertshau: o nice didn't realize gson handled uris.  simplifies this a decent amount

albertshau:
http://builds.cask.co/browse/CDAP-DUT2455

albertshau:
@chtyim addressed comments, could you take another look?

chtyim:
LGTM

PR #3521, opened 2015-08-04
Feature/expose history on all program managers

anwar6953:
First commit moves the implementation from WorkflowManager to AbstractProgramManager to enable that functionality for all program types.
Second commit simply cleans up redundant castings.

http://builds.cask.co/browse/CDAP-DUT2456-3

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/DataSetInputFormat.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/DataSetInputFormat.java
@@ -75,7 +75,7 @@ public final class DataSetInputFormat<KEY, VALUE> extends InputFormat<KEY, VALUE
     BasicMapReduceContext mrContext = contextProvider.get();
     mrContext.getMetricsCollectionService().startAndWait();
     String dataSetName = getInputName(conf);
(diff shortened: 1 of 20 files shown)


bdmogal:
The code LGTM. However, should we add some unit tests?

anwar6953:
The test case was a good idea; it revealed that AppFabricClient#getHistory was not generic to all program types.
Add test case to test ServiceManager#getHistory and fix AppFabricClient#getHistory to be more generic.

anwar6953:
"So this method was workflow specific?"
Yes.

"don't need the ``status=completed``? This is for completed runs right?"
Only the url had status=completed. The actual call to the handler didn't specify a particular status. This is consistent in both the default as well as remote implementations, so I simply removed that part from the url to also make that consistent.

bdmogal:
> The actual call to the handler didn't specify a particular status. 

Should it, though?

Apart from that, LGTM

anwar6953:
I don't think so. This enables users to retrieve all run records, which is more flexible than simple returning the completed ones.

bdmogal:
But do the underlying usages expect only completed runs? We don't allow users to specify a filter either, right?

anwar6953:
Updated to allow user to specify a filter (ProgramRunStatus).

bdmogal:
LGTM

PR #3522, opened 2015-08-04
consolidates cdap logo into one image

astewes:
*  Removes icomoon fist from global header
*  Adds first to "CDAP" image
*  Removes unnecessary HTML markup and corresponding styles

cdap


ajainarayanan:
:+1: LGTM

PR #3523, opened 2015-08-05
(CDAP-20) Continue effort in removing Guava usage in cdap-api

chtyim:
There are two commits in this PR. They are grouped as follows:

1. More relatively trivial removal of guava usage in cdap-api
  - Also make MapReduceSpecification a class instead of interface
  -   We don’t expect anyone implementing that interface since it’s an class provided through the MapReduceContext to the user.
2. Deprecate immutableSortedMapOf methods in Bytes
  - They don’t really belong to bytes operation
  - They return NavigableMap instead of SortedMap
  - Also relax constraints on MetricsTable to accept SortedMap instead of NavigableMap
    - This constraint is only for InMemory table implementation, which actually doesn’t need NavigableMap as well


chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2451-5

albertshau:
:+1:

PR #3526, opened 2015-08-05
CDAP-2764 use artifact repo in app deployment pipeline

albertshau:
Some refactoring to make the application deployment pipeline add
the jar to the artifact repo, then call the deployment pipeline
with the id for the newly added artifact. Unfortunately, the
old code path still exists because app templates rely on it, but
should be removed soon. ArtifactID is now also part of the app
spec, which is a requirement for plugin integration with apps.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2448

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -258,6 +272,16 @@ private BodyConsumer deployApplication(final HttpResponder responder,
       return null;
     }
 
+    // error handling needs to be refactored here, should be able just to throw the exception, but the caller
```
gokulavasan: may be prefix it with TODO ?

cdap-client-tests/src/test/java/co/cask/cdap/client/PreferencesClientTestRun.java
```diff
@@ -238,7 +239,11 @@ public void testPreferences() throws Exception {
       Assert.assertEquals(propMap, client.getApplicationPreferences(FAKE_APP_ID, false));
       Assert.assertEquals(propMap, client.getProgramPreferences(flow, false));
     } finally {
-      appClient.delete(FAKE_APP_ID);
+      try {
+        appClient.delete(FAKE_APP_ID);
+      } catch (ApplicationNotFoundException e) {
+        // ok if this happens, means its already deleted.
```
anwar6953: why would it already be deleted, if its redeployed line 237?

albertshau: its in a finally, so there could be an exception in the test that happens before its redeployed. This test will fail, but we'd want the failure exception to be from that earlier failure and not from this delete.

anwar6953: I see. ok.

cdap-app-fabric/src/main/java/co/cask/cdap/app/DefaultAppConfigurer.java
```diff
@@ -85,6 +87,12 @@ public DefaultAppConfigurer(Application app, String configuration) {
     this.configuration = configuration;
   }
 
+  public DefaultAppConfigurer(Id.Artifact artifactId, Application app, String configuration) {
+    this(app);
```
gokulavasan: this(app, configuration);

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -371,6 +412,30 @@ private void getAppDetails(HttpResponder responder, String namespace, String nam
     }
   }
 
+  // parse a filename into an Id.Artifact, assuming the filename is of format {name}-{version}.jar
```
gokulavasan: Will this be the only way to get the version of the jar? Will we also have a fall back option of trying to get it from the manifest?

albertshau: yeah i will need to add that logic, but we should try and get from manifest if the header is not present. Opened CDAP-3310

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/LocalArtifactLoaderStage.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.twill.filesystem.Location;
+
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import javax.annotation.Nullable;
+
+/**
+ * LocalArchiveLoaderStage gets a {@link Location} and emits a {@link ApplicationDeployable}.
```
gokulavasan: LocalArtifactLoaderStage

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -186,12 +186,40 @@ public ArtifactDetail getArtifact(Id.Artifact artifactId) throws IOException, Ar
...
    * @param artifactId the id of the artifact to inspect and store
    * @param artifactFile the artifact to inspect and store
+   * @return detail about the newly added artifact
+   * @throws IOException if there was an exception reading from the artifact store
+   * @throws WriteConflictException if there was a write conflict writing to the ArtifactStore
+   * @throws ArtifactAlreadyExistsException if the artifact already exists
+   * @throws InvalidArtifactException if the artifact is invalid. For example, if it is not a zip file,
+   *                                  or the application class given is not an Application.
+   */
+  public ArtifactDetail addArtifact(Id.Artifact artifactId, File artifactFile)
```
gokulavasan: can't we just call addArtifact(artifactId, artifactFile, null) ?

gokulavasan:
Few comments. LGTM otherwise

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
albertshau: functionally yes, but this method doesn't throw an exception about parent artifacts not found whereas that one does.

gokulavasan: Oh ok.

gokulavasan:
LGTM :+1:

PR #3528, opened 2015-08-05
modifies .btn-styles mixin to fix button state styling

astewes:
*  Modifies .btn-styles mixin to fix issue where \<button> elements were getting "stuck" in active/focus state
*  Adds correct default styling for .btn-default class

### DIFF
--- a/cdap-ui/app/features/admin/admin.less
+++ b/cdap-ui/app/features/admin/admin.less
@@ -200,6 +200,7 @@ body.theme-cdap {
       // This is required to steal the focus from an already clicked button on previous state.
       .btn-default.active, .btn.btn-default:focus {
         background-color: white;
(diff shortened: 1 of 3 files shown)


ajainarayanan:
@astewes Do we still need the system service detail view in admin to have styles to fix metadata/logs tab buttons?

astewes:
@ajainarayanan i tried this out and yes, looks like the styles will need to be kept in admin. The default for the .active class is to darken the button color, but with Metadata/Logs, we want to override the default with a white background and use a box shadow to differentiate the active button.

elfenheart:
@astewes @ajainarayanan added a small fix to this PR, fixing typo on workflow icon classname.

:+1: LGTM on @astewes changes

ajainarayanan:
:+1: LGTM

PR #3529, opened 2015-08-06
ability to add multiple events when sending events to streams

elfenheart:
- Ability to copy paste multiple lines (events) when sending events to a stream
- User can press shift + enter to go to a newline

multipleevents


ajainarayanan:
:+1: LGTM

PR #3532, opened 2015-08-06
Wait till the workflow stop before calling the current endpoint.

sagarkapare:
Fix the flaky test case.
We give command to stop the Workflow and then call the current endpoint, which should return 404 when workflow is not running. However if the Workflow is not completely stopped, then the current endpoint may return 500. 

Added wait in the test case till the Workflow stops completely and then call current endpoint.

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
@@ -522,6 +522,9 @@ public class WorkflowHttpHandlerTest  extends AppFabricTestBase {
     // Stop the program while in fork
     stopProgram(programId, 200);
 
+    // Wait till the program stop
+    waitState(programId, "STOPPED");
+
     // Current endpoint would return 404
     response = getWorkflowCurrentStatus(programId, runId);
     Assert.assertEquals(404, response.getStatusLine().getStatusCode());


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -522,6 +522,9 @@ public void testWorkflowForkApp() throws Exception {
     // Stop the program while in fork
     stopProgram(programId, 200);
 
+    // Wait till the program stop
+    waitState(programId, "STOPPED");
```
bdmogal: is there a way to use ``ProgramRunStatus`` in ``waitState``? Not for this PR though. If its possible, can you add a note in https://issues.cask.co/browse/CDAP-2984

gokulavasan:
What does the current endpoint do? Can you point me to the docs? And why does it return 500?

bdmogal:
The reasoning and the fix for it LGTM.

sagarkapare:
@gokulavasan The link to the doc is - http://docs.cdap.io/cdap/current/en/reference-manual/http-restful-api/lifecycle.html. Please scroll down to `Retrieving Specific Run Information` section. When the workflow is stopped, there is a possibility that service handling the current endpoint is discoverable but by the time when actually query to the service id made its terminated, which is when the exception would be thrown and 500 code would be returned. The fix makes sure that we wait till the workflow is completely STOPPED and then query the endpoint so that expected code 404 is returned.

sagarkapare:
Build passed here - https://builds.cask.co/browse/CDAP-DRC2322-1. @gokulavasan if not any more questions we can merge this.

gokulavasan:
LGTM :+1:

PR #3538, opened 2015-08-06
[CDAP-3305] add confirmation when exiting adapter create with unsaved changes

elfenheart:
confirmations


cdap-ui/app/features/adapters/controllers/create-new-ctrl.js
```diff
@@ -78,7 +78,33 @@ angular.module(PKG.name + '.feature.adapters')
         }.bind(this));
     };
 
+
+    var confirmOnPageExit = function (e) {
+
+      if (!MyPlumbService.isConfigTouched) { return; }
+      // If we haven't been passed the event get the window.event
+      e = e || window.event;
```
ajainarayanan: May be we should use ```$window```? If at all when we add tests it will be easier to test. I think we have been using ```$window``` in other places I hope.

ajainarayanan:
Minor comment. Otherwise :+1: LGTM

ajainarayanan:
:+1: LGTM

PR #3540, opened 2015-08-07
Fix/ui nodeproxy ssl enabled

ajainarayanan:
- Fixes node proxy to read correct key/certificate files when ssl is enabled.
- Fixes node proxy to send it as config so that UI will change its behavior while making queries through websockets and directly to backend.

### DIFF
--- a/cdap-ui/server.js
+++ b/cdap-ui/server.js
@@ -6,18 +6,19 @@ var express = require('./server/express.js'),
     parser = require('./server/config/parser.js'),
     sockjs = require('sockjs'),
     http = require('http'),
+    fs = require('fs'),
     log4js = require('log4js'),
     https = require('https');
-  
+
 var cdapConfig, securityConfig;
 
 /**
- * Configuring the logger. In order to use the logger anywhere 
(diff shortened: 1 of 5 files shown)


cdap-ui/server.js
```diff
@@ -55,10 +56,14 @@ parser.extractConfig('cdap')
...
-        key: fs.readFileSync(securityConfig['dashboard.ssl.cert']),
-        cert: fs.readFileSync(securityConfig['dashboard.ssl.key'])
-      }, app);
+      try {
+        server = https.createServer({
+          key: fs.readFileSync(securityConfig['dashboard.ssl.key']),
+          cert: fs.readFileSync(securityConfig['dashboard.ssl.cert'])
+        }, app);
+      } catch(e) {
+        log.debug('SSL key/cert files read failed. Pleas fix the key/ssl certificate files and restart node server -  ', e);
```
elfenheart: small typo on 'Please'

elfenheart:
small comment, otherwise LGTM :+1:

PR #3541, opened 2015-08-07
CDAP-2780 add a way to scan a directory for system artifacts

albertshau:
Adding a method to ArtifactRepository to scan a directory for
system artifacts to add. This will be done when AppFabricServer
starts up, and can also be called on command through a RESTful
call in case system artifacts need to be added without restarting
AppFabric.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
@@ -74,7 +74,7 @@ import javax.ws.rs.QueryParam;
  * {@link co.cask.http.HttpHandler} for managing adapter lifecycle.
  */
 @Singleton
-@Path(Constants.Gateway.API_VERSION_3 + "/namespaces/{namespace-id}")
+@Path(Constants.Gateway.API_VERSION_3)
 public class ArtifactHttpHandler extends AbstractHttpHandler {
   private static final Logger LOG = LoggerFactory.getLogger(ArtifactHttpHandler.class);
   private static final String VERSION_HEADER = "Artifact-Version";
@@ -97,8 +97,25 @@ public class ArtifactHttpHandler extends AbstractHttpHandler {
       cConf.get(Constants.AppFabric.TEMP_DIR)).getAbsoluteFile();
   }
 
+  @POST
+  @Path("/namespaces/system/artifacts")
+  public void refreshSystemArtifacts(HttpRequest request, HttpResponder responder,
+                                     @PathParam("namespace-id") String namespaceId) {
+    try {
+      artifactRepository.addSystemArtifacts();
+      responder.sendStatus(HttpResponseStatus.OK);
+    } catch (IOException e) {
+      LOG.error("Error while refreshing system artifacts.", e);
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR,
+        "There was an IO error while refreshing system artifacts, please try again.");
+    } catch (WriteConflictException e) {
+      LOG.error("Error while refreshing system artifacts.", e);
+      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
+    }
+  }
+
   @GET
-  @Path("/artifacts")
+  @Path("/namespaces/{namespace-id}/artifacts")
   public void getArtifacts(HttpRequest request, HttpResponder responder,
                            @PathParam("namespace-id") String namespaceId,
                            @QueryParam("includeSystem") @DefaultValue("true") boolean includeSystem)
@@ -115,7 +132,7 @@ public class ArtifactHttpHandler extends AbstractHttpHandler {
   }
 
   @GET
-  @Path("/artifacts/{artifact-name}")
+  @Path("/namespaces/{namespace-id}/artifacts/{artifact-name}")
   public void getArtifactVersions(HttpRequest request, HttpResponder responder,
                                   @PathParam("namespace-id") String namespaceId,
                                   @PathParam("artifact-name") String artifactName,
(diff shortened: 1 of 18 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2473

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
```diff
@@ -97,8 +97,25 @@ public ArtifactHttpHandler(CConfiguration cConf,
       cConf.get(Constants.AppFabric.TEMP_DIR)).getAbsoluteFile();
   }
 
+  @POST
+  @Path("/namespaces/system/artifacts")
+  public void refreshSystemArtifacts(HttpRequest request, HttpResponder responder,
+                                     @PathParam("namespace-id") String namespaceId) {
```
gokulavasan: Is the ``namespaceId`` being used?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginClassDeserializer.java
```diff
@@ -0,0 +1,69 @@
...
+                                 JsonDeserializationContext context) throws JsonParseException {
+    if (!json.isJsonObject()) {
+      throw new JsonParseException("PluginClass should be a JSON Object");
+    }
+
+    JsonObject jsonObj = json.getAsJsonObject();
+
+    String type = jsonObj.has("type") ? jsonObj.get("type").getAsString() : Plugin.DEFAULT_TYPE;
+    String name = getRequired(jsonObj, "name").getAsString();
+    String description = getRequired(jsonObj, "description").getAsString();
```
gokulavasan: Why is description a required property?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
albertshau: oops shouldn't have it here, will remove.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginClassDeserializer.java
albertshau: good question, was just moving this class but doesn't seem like it should be.  Will change it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -184,6 +192,7 @@ public ArtifactDetail getArtifact(Id.Artifact artifactId) throws IOException, Ar
 
   /**
    * Inspects and builds plugin and application information for the given artifact.
+   * TODO (CDAP-3319) check parents don't have parents
```
gokulavasan: I guess this can be removed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -242,6 +253,109 @@ public ArtifactDetail addArtifact(Id.Artifact artifactId, File artifactFile,
...
+          // otherwise, don't parse it
+          systemArtifacts.add(SystemArtifactConfig.builder(artifactId, jarFile).build());
+        }
+      } catch (InvalidArtifactException e) {
+        LOG.warn(String.format("Could not add system artifact '%s' because it is invalid.", artifactFileName), e);
+      }
+    }
+
+
+    // Need to be sure to add parent artifacts before artifacts that extend them.
```
gokulavasan: How are making sure that we add parent artifacts before adding artifacts that extend them? Does Collections.sort do the trick?

gokulavasan: Ok nvmind, saw the SystemArtifactConfig implementation below.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/SystemArtifactConfig.java
```diff
@@ -0,0 +1,324 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.Reader;
+import java.lang.reflect.Type;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Represents json config file that can be placed in the system artifacts directory. Comparable so that we
```
gokulavasan: It looks like we also use this class for artifacts that don't have a json config file (such as ETLBatch, ETLRealtime app artifacs?).

albertshau: thats true, I will update the javadoc so that it is more accurate

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/SystemArtifactConfig.java
```diff
@@ -0,0 +1,324 @@
...
+      JsonObject obj = json.getAsJsonObject();
+
+      // deserialize fields
+      Set<ArtifactRange> parents = context.deserialize(obj.get("parents"), PARENTS_TYPE);
+      parents = parents == null ? Collections.<ArtifactRange>emptySet() : parents;
+      Set<PluginClass> plugins = context.deserialize(obj.get("plugins"), PLUGINS_TYPE);
+      plugins = plugins == null ? Collections.<PluginClass>emptySet() : plugins;
+
+      try {
+        validateParentList(parents);
```
gokulavasan: The validation of parentList and pluginList needs to be performed for all artifacts isn't it? Is it specific to SystemArtifacts?

albertshau: thats a good point, guess I need to move it out to some common class.

gokulavasan:
Few comments. Rest LGTM :+1:

gokulavasan:
LGTM :+1:

PR #3543, opened 2015-08-07
CDAP-3319 make sure artifacts cant have grandparents

albertshau:
This is already a limitation of the plugin system, but it was not
being enforced by the artifact framework. We want to keep things
simple so that an artifact can extend another artifact, but only
to one level. This makes it simpler for our framework, but also
prevents users from creating complicated dependency trees that
are hard for them to manage as well.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2476

gokulavasan:
@albertshau Can you explain how this is limited by the plugin system? PR changes LGTM :+1:

albertshau:
@gokulavasan Our PluginClassLoader assumes there is one app jar and one plugin jar and builds the classloader from those 2 jars. There isn't support for a whole chain of jars, and I don't think we'd want to introduce unless we have compelling use cases around it

PR #3545, opened 2015-08-07
Improved flowlet error messages.

anwar6953:
The code was using `UserErrors.INVALID_STREAM_NAME`, where it should've been using
`UserErrors.INVALID_STREAM_NULL`.
Also, some places, it was using `UserErrors.INVALID_FLOWLET_NAME`, but didn't specify the string of the requested flowlet name, so the error message was getting logged/outputted with a `%s` in it:
`java.lang.IllegalArgumentException: Your FlowSpecification attempts to connect from an invalid Flowlet; "%s" does not exist`

http://builds.cask.co/browse/CDAP-DUT2478-1


gokulavasan:
LGTM :+1:

PR #3546, opened 2015-08-07
[CDAP-2514] add upload csv to streams

elfenheart:
uploadcsv


bdmogal:
We should probably not restrict it to csv, just allow to upload any file? However I see that you're setting content-type to text/csv. Maybe allow users to upload any file with a format specifier (select box that allows you to choose csv or plain, and depending on that you set content-type to text/csv or text/plain)? We also support avro/binary.

Only allowing csv upload seems a bit restrictive.

ajainarayanan:
@elfenheart Two minor things,
1. Do we need uploadCsv function in fileUploader factory? All that is different from this to uploadFile is the content-type. Can't we make it dynamic so that we don't have redundant code?
2. Should we disable uploadCSV and submit buttons in the modal while upload takes places?

elfenheart:
addressed comments.

Will do a design with Sean to incorporate @bdmogal suggestion. Coming in later PR.

ajainarayanan:
:+1: LGTM

PR #3547, opened 2015-08-07
CDAP-3224 add api for creating an app from an artifact

albertshau:
Modifies the current PUT /apps/<appid> api so that requests with
a json content type are interpreted as a request to create an app
based on an artifact that already exists.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
@@ -265,6 +265,10 @@ public class AppLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
     responder.sendString(status.getCode(), status.getMessage());
   }
 
+  // normally we wouldn't want to use a body consumer but would just want to read the request body directly
+  // since it wont be big. But the deploy app API has one path with different behavior based on content type
+  // the other behavior requires a BodyConsumer and only have one method per path is allowed,
+  // so we have to use a BodyConsumer
   private BodyConsumer deployAppFromArtifact(final Id.Namespace namespace, final String appId) throws IOException {
 
     return new AbstractBodyConsumer(File.createTempFile(appId, ".json", tmpDir)) {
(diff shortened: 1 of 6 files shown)


awholegunch:
Should there be some documentation changes to reflect this, and indicate how backwards-compatibilty is handled?

albertshau:
there will be doc changes, but wanted to do them separately

albertshau:
http://builds.cask.co/browse/CDAP-DUT2479

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -244,6 +265,54 @@ public void deleteAllApps(HttpRequest request, HttpResponder responder,
     responder.sendString(status.getCode(), status.getMessage());
   }
 
+  private BodyConsumer deployAppFromArtifact(final Id.Namespace namespace, final String appId) throws IOException {
+
+    return new AbstractBodyConsumer(File.createTempFile(appId, ".json", tmpDir)) {
```
gokulavasan: Do we need BodyConsumer for json file that contains artifact info and configuration?

albertshau: normally no, but we decided to do this weird thing where the path is the same as the current api which does need a body consumer, so the two need to have the same method signature.

cdap-proto/src/main/java/co/cask/cdap/proto/artifact/CreateAppRequest.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Request body when creating an app
+ *
+ * @param <T> the type of application config
+ */
+public class CreateAppRequest<T> {
+  private final ArtifactSummary artifact;
+  private final T config;
+
+  public CreateAppRequest(ArtifactSummary artifact, T config) {
```
gokulavasan: Would it be useful to have a constructor that takes in only artifact summary?

gokulavasan:
Couple of comments

gokulavasan:
LGTM :+1:

PR #3548, opened 2015-08-08
Show 0 instead of a dash, when there are 0 stream events

anwar6953:
This is now consistent with the datasets page, which shows 0b (as opposed to a dash), when there isn't a non-zero metric value for that metric:
https://github.com/caskdata/cdap/blob/feature-ui/stream-page-show-0/cdap-ui/app/features/datasets/templates/tabs/status.html#L17

Before:
image

Now:
image


ajainarayanan:
:+1: LGTM

PR #3549, opened 2015-08-09
Fix race condition in testAppWithWorker that can fail occasionally.

chtyim:


cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
```diff
@@ -603,14 +603,30 @@ public void testAppWithWorker() throws Exception {
     ApplicationManager applicationManager = deployApplication(testSpace, AppWithWorker.class);
     LOG.info("Deployed.");
```
bdmogal: should we remove this log?

bdmogal:
LGTM

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
chtyim: Not the focus of this PR. Not going to change it now.

chtyim:
Build passed http://builds.cask.co/browse/CDAP-DUT2482-3

PR #3550, opened 2015-08-09
Feature/ui predefined templates

ajainarayanan:
- Adds express end points to query/get pre-defined templates.
- Modifies group-side-panel to be more generic (less verbose)
- Adds $resource endpoints to query/get predefined templates
- Modifies canvas-ctrl to load predefined templates into canvas in single click.

##### TODO:
  - After cloning/importing a template if the user decide to scrap off everything and decide to use a template what should be the behavior? Scratch everything and draw the template (which seems to be logical). If that is the case then the behavior is already there. If not we need to handle that separately.

 
predefinedtemplates


elfenheart:
naming the folder adaptertemplates might be a little confusing. Since we are going away with the name adapters, should we call this folder something else?

cdap-ui/app/services/adapter/my-adapter-templatep-api.js
```diff
@@ -0,0 +1,21 @@
+angular.module(PKG.name + '.services')
+  .factory('myAdapterTemplatesApi', function($resource) {
+    return $resource(
+      '',
+      {
+        apptype: '@apptype',
+        appname: '@appname'
+      },
+      {
+        query: {
```
elfenheart: instead of query, should we call this list instead?

elfenheart:
minor comment, otherwise :+1: LGTM

cdap-ui/app/services/adapter/my-adapter-templatep-api.js
ajainarayanan: They both ideally mean the same. Don't know which one is the correct usage. If query sounds not intuitive will change it though. @elfenheart thoughts?

elfenheart:
:+1: LGTM

PR #3551, opened 2015-08-10
Fix/ui adapter export prune

ajainarayanan:
- Removed _backendProperties from config when exporting adapters (exported json)
- Minor function renames to make it a little bit obvious (since we handle both left and right sidebars in the same controller)


elfenheart:
:+1: LGTM

PR #3553, opened 2015-08-10
CDAP-3343 make sure plugin inspector has spark access

albertshau:
Make changes so that when the framework is inspecting an artifact
for plugins it has access to spark packages in case the artifact
contains spark programs. The classloader is only used when
inspecting artifacts and is closed right after inspection finishes.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2485

gokulavasan:
LGTM :+1:

PR #3557, opened 2015-08-10
Copy deploy.sh from develop

wolf31o2:
This includes SSH host key checking fix.

### DIFF
--- a/cdap-docs/deploy.sh
+++ b/cdap-docs/deploy.sh
@@ -120,8 +120,9 @@ DOCS_SERVERS="${DOCS_SERVER1} ${DOCS_SERVER2}"
 REMOTE_STG_DIR="${REMOTE_STG_BASE}/${PROJECT}/${REMOTE_DIR}"		# e.g. /var/www/html/staging/cdap/develop
 REMOTE_DOCS_DIR="${REMOTE_DOCS_BASE}/${PROJECT}/${REMOTE_DIR}"		# e.g. /var/www/docs/cdap/release
 
+SSH_OPTS='-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null'
 RSYNC_OPTS='-aPh'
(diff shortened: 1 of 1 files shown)


cdap-docs/deploy.sh
```diff
@@ -186,6 +183,12 @@ function deploy () {
   local _version=${6}
   local _branch=${7}
   decho "deploying to ${_host}"
+  # if there is no branch name to append, append version to directory name, otherwise:
+  #   cdap dir gets blown away in remote dir delete/create
+  #   otherwise all files end up one level too high after rsync
```
awholegunch: otherwise twice. "all files end up one level too high after rsync"

cdap-docs/deploy.sh
```diff
@@ -200,16 +203,20 @@ decho "DEPLOY_TO_DOCS=${DEPLOY_TO_DOCS}"
...
-  for i in ${DOCS_SERVERS}; do
-    deploy ${USER} ${i} ${REMOTE_DOCS_DIR} ${ZIP_FILE} ${FILE_PATH} ${VERSION} ${BRANCH}
-  done
+  if [ "${BRANCH}" == '' ]; then
+    decho "Deploying artifacts to Docs servers"
+    for i in ${DOCS_SERVERS}; do
+      deploy ${USER} ${i} ${REMOTE_DOCS_DIR} ${ZIP_FILE} ${FILE_PATH} ${VERSION}
+    done
+  else
+    decho "Do not deploy artifacts from feature branches to Docs servers"
```
awholegunch: This message looks odd to me; we do deploy artifacts from the feature branches (including release feature branches) to Docs servers. Why is this not deploying if Branch != ""?

awholegunch:
A couple of minor comments, otherwise LGTM.

wolf31o2:
This was a copy from the current develop version to release/2.8... I neither wrote nor modified this code. I literally checked out the latest version of the file and made a commit the correct branch.

PR #3558, opened 2015-08-10
make IntegrationTestBase#checkSystemServices private --> protected

anwar6953:
Use case is that the system services need to be checked for `OK` status from subclass (cdap-integration-test repo, for instance), and this functionality is already implemented in here.


gsps1:
lgtm

PR #3565, opened 2015-08-11
Cleaning up dependencies. Specifically

chtyim:
- Replace the unnecessary usage of fastutil in etl-lib
- Do proper exclusion and dependency management in etl-lib
- Fix duplicated dependency in cdap-watchdog

Built standalone using command in BUILD.rst. No out of memory issue.


gokulavasan:
LGTM :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2492

Also verified in standalone and distributed mode.

PR #3566, opened 2015-08-11
CDAP-3356 remove deprecated 'id' field from RESTful APIs

albertshau:
More specifically, removing from the apps and programs RESTful
APIs. Id was deprecated because it always had the same value as
name. Also removing unused classes for streams and datasets.


cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
```diff
@@ -274,7 +273,7 @@ public void setTTL(Id.Stream stream, long ttlInSeconds)
   /**
    * Lists all streams.
    *
-   * @return list of {@link StreamRecord}s
+   * @return list of {@link StreamDetail StreamDetails}
```
bdmogal: is this valid syntax?

bdmogal:
1 question, otherwise LGTM

cdap-client/src/main/java/co/cask/cdap/client/StreamClient.java
chtyim: it is.

awholegunch: One is the link, the second is a label. As you can see, lets one use consistent pluralization.

awholegunch:
LGTM. Please copy your release note comment to the new "Release Notes" field added to the Jiras database. That way we will know it has been written up.

PR #3567, opened 2015-08-11
CDAP-3138: CLI shouldn't stack trace on unable to connect at startup

jwang47:
https://issues.cask.co/browse/CDAP-3138
http://builds.cask.co/browse/CDAP-DUT2496

With this change, connection refused messages are nicer.

Without specifying -u:

```
CDAP instance at 'http://myhost:10000/default' could not be reached: Connection refused
Specify the CDAP instance URI with the -u command line argument.
```

With specifying -u:

```
CDAP instance at 'http://myhost:10000/default' could not be reached: Connection refused
```

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
@@ -108,7 +108,8 @@ public class CLIMain {
   private final LaunchOptions options;
   private final FilePathResolver filePathResolver;
 
-  public CLIMain(final LaunchOptions options, final CLIConfig cliConfig) throws URISyntaxException, IOException {
+  public CLIMain(final LaunchOptions options,
+                 final CLIConfig cliConfig) throws URISyntaxException, IOException {
     this.options = options;
     this.cliConfig = cliConfig;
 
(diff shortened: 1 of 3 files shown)


cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -171,20 +172,26 @@ public void onConnectionChanged(CLIConnectionConfig config) {
...
         } else {
           cliConfig.getOutput().println(e.getMessage());
         }
+        if (!command.hasOption(URI_OPTION.getOpt())) {
+          cliConfig.getOutput().printf("Specify the CDAP instance URI with the -u command line argument.\n");
+        }
+        return false;
       }
     }
+    return true;
```
chtyim: Invert the if logic and move this up.

```java
if (!options.isAutoconnect()) {
  return true;
}
try {
  ...
  return true;
} catch (Exception e) {
  ...
  return false;
}
```

cdap-client/src/main/java/co/cask/cdap/client/config/ClientConfig.java
```diff
@@ -128,14 +128,14 @@ public URL resolveNamespacedURLV3(Id.Namespace namespace, String path) throws Ma
 
   public HttpRequestConfig getDefaultRequestConfig() {
     if (connectionConfig == null) {
-      throw new DisconnectedException();
+      throw new DisconnectedException(null);
```
chtyim: Why not having an overloaded constructor for `DisconnectedException` that takes no argument so that caller don't need to pass in null explicitly?

cdap-client/src/main/java/co/cask/cdap/client/exception/DisconnectedException.java
```diff
@@ -15,8 +15,20 @@
...
  */
 public class DisconnectedException extends RuntimeException {
+
+  private final ConnectionConfig connectionConfig;
+
+  public DisconnectedException(ConnectionConfig connectionConfig) {
+    this.connectionConfig = connectionConfig;
+  }
+
+  public ConnectionConfig getConnectionConfig() {
```
chtyim: From the usage, seems like this is `@Nullable`.

cdap-client/src/main/java/co/cask/cdap/client/exception/DisconnectedException.java
```diff
@@ -15,8 +15,28 @@
...
+import co.cask.cdap.client.config.ConnectionConfig;
+
+import javax.annotation.Nullable;
+
 /**
  * Thrown when the client is not connected, but a request was attempted.
  */
 public class DisconnectedException extends RuntimeException {
+
+  @Nullable
```
bdmogal: is this necessary? I think just having it on the getter is sufficient, since there's no way you can access the field directly.

bdmogal:
One minor comment, otherwise :+1:

cdap-client/src/main/java/co/cask/cdap/client/exception/DisconnectedException.java
jwang47: Why not have it? I think it makes it clear for someone reading this code.

PR #3570, opened 2015-08-11
CDAP-3270: CLI set stream format documentation

jwang47:
https://issues.cask.co/browse/CDAP-3270
http://builds.cask.co/browse/CDAP-DUT2497


awholegunch:
LGTM.

PR #3574, opened 2015-08-12
Use the correct API to check for file existence.

chtyim:


chtyim:
Verified on cluster.

poornachandra:
LGTM

chtyim:
Merging as it doesn't affect unit-test at all

PR #3575, opened 2015-08-12
[CDAP-3333] Predefined Plugins Phase 1

elfenheart:
Roadmap:

1. Ability to edit existing plugins (#3589)
2. Fix widget-dataset-selector (#3589)
3. When user adds the predefined template:
3.1 DAG should show templateName instead of plugin name
3.2 The prepopulated fields should be disabled #3606
4. Implement design of the templates creation screen

predefinedplugins


cdap-ui/app/features/admin/controllers/namespace/templates-list-ctrl.js
```diff
@@ -0,0 +1,29 @@
...
+    mySettings.get('pluginTemplates')
+      .then(function (res) {
+        vm.list = res[$stateParams.nsadmin];
+      });
+
+    vm.delete = function (template) {
+      delete template;
+      mySettings.get('pluginTemplates')
+        .then(function (res) {
+          delete res[$stateParams.nsadmin][template.templateName];
```
ajainarayanan: What if we start out with empty plugin templates? 
```res``` will be undefined and undefined[...][...] will be a js error right?

ajainarayanan:
@elfenheart 
- I guess styling is still yet to be finalized I guess?
- We talked about extracting the properties template (middle section) alone from the plugin-edit-properties view and then re-use the template and controller in multiple places (in this case admin). Are we doing in this PR or some other one?

elfenheart:
@ajainarayanan I am planning to do that at a later stage. I want to make the modifications needed to take the feature to completion to see the difference between the two and see if it still make sense to re use templates/directives.

On styling, still waiting for a mock up from Nitin/Sean, so we can leave that for last (item 4)

ajainarayanan:
:+1: LGTM

PR #3576, opened 2015-08-13
CDAP-3283 index application class information and expose it

albertshau:
adding rows to the ArtifactStore metatable for application classes
found in artifacts to allow efficient lookup of app classes in
a namespace and for a given class name. This functionality is
exposed through the ArtifactRepository and through the http
handler.

Filling in the REST endpoint for getting application classes in
a namespace and by classname. Also adding artifact name and version
filters to the get apps REST endpoint.

Finally, also added methods to ArtifactClient corresponding to the
new REST APIs, and a new method to ApplicationClient for filtering
applications by artifact.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
@@ -36,6 +36,8 @@ import co.cask.cdap.internal.app.runtime.artifact.PluginNotExistsException;
 import co.cask.cdap.internal.app.runtime.artifact.WriteConflictException;
 import co.cask.cdap.internal.io.SchemaTypeAdapter;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.artifact.ApplicationClassInfo;
+import co.cask.cdap.proto.artifact.ApplicationClassSummary;
 import co.cask.cdap.proto.artifact.ArtifactInfo;
 import co.cask.cdap.proto.artifact.ArtifactRange;
 import co.cask.cdap.proto.artifact.ArtifactSummary;
@@ -90,6 +92,8 @@ public class ArtifactHttpHandler extends AbstractHttpHandler {
   private static final String VERSION_HEADER = "Artifact-Version";
   private static final String EXTENDS_HEADER = "Artifact-Extends";
   private static final String PLUGINS_HEADER = "Artifact-Plugins";
+  private static final Type APPCLASS_SUMMARIES_TYPE = new TypeToken<List<ApplicationClassSummary>>() { }.getType();
+  private static final Type APPCLASS_INFOS_TYPE = new TypeToken<List<ApplicationClassInfo>>() { }.getType();
   private static final Gson GSON = new GsonBuilder()
     .registerTypeAdapter(Schema.class, new SchemaTypeAdapter())
     .registerTypeAdapter(PluginClass.class, new PluginClassDeserializer())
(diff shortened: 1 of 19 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2501

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
```diff
@@ -294,8 +298,14 @@ public void getApplicationClasses(HttpRequest request, HttpResponder responder,
 
     Id.Namespace namespace = validateAndGetNamespace(namespaceId);
 
-    // TODO: implement
-    responder.sendJson(HttpResponseStatus.OK, Lists.newArrayList());
+    try {
+      responder.sendJson(HttpResponseStatus.OK, artifactRepository.getApplicationClasses(namespace, includeSystem),
+        APPCLASS_SUMMARIES_TYPE, GSON);
```
anwar6953: nit: indent to align with first param in previous line?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -133,6 +136,47 @@ public ArtifactDetail getArtifact(Id.Artifact artifactId) throws IOException, Ar
...
+  /**
+   * Get all application classes in the given namespace of the given class name.
+   * Will never return null. If no artifacts exist, an empty list is returned. Namespace existence is not checked.
+   *
+   * @param namespace the namespace to get application classes from
+   * @param className the application class to get
+   * @return an unmodifiable list of application classes that belong to the given namespace
+   * @throws IOException if there as an exception reading from the meta store
+   */
+  public List<ApplicationClassInfo> getApplicationClasses(Id.Namespace namespace,
```
gokulavasan: Why are we returning ApplicationClassInfo here but ApplicationClassSummary in the method above? Also by is there no includeSystem boolean in this method?

cdap-client-tests/src/test/java/co/cask/cdap/client/ArtifactClientTestRun.java
```diff
@@ -222,6 +227,26 @@ public InputStream getInput() throws IOException {
       new ArtifactInfo(pluginId.getName(), pluginId.getVersion().getVersion(), false, pluginClasses);
     Assert.assertEquals(pluginArtifactInfo, artifactClient.getArtifactInfo(pluginId));
 
+    // test get all app classes in namespace
+    // no way to delete artifacts yet... when run in a suite will see other app classes from app deployments,
```
gokulavasan: This isn't true any more is it?

cdap-client/src/main/java/co/cask/cdap/client/ArtifactClient.java
```diff
@@ -198,6 +202,73 @@ public ArtifactInfo getArtifactInfo(Id.Artifact artifactId, boolean isSystem)
...
+   * @param namespace the namespace to list application classes from
+   * @param includeSystem whether to include application classes from system artifacts
+   * @return summaries of all application classes in the given namespace, including classes from system artifacts
+   * @throws IOException if a network error occurred
+   * @throws UnauthorizedException if the request is not authorized successfully in the gateway server
+   */
+  public List<ApplicationClassSummary> getApplicationClasses(Id.Namespace namespace, boolean includeSystem)
+    throws IOException, UnauthorizedException {
+
+    String path = String.format("classes/apps?includeSystem=%s", includeSystem);
```
gokulavasan: Where is this endpoint implemented?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
albertshau: there is no include system because the flag only applies to the top level get call (get /classes/apps).  When you ask for a specific class (get /classes/apps/co.cask....) there is a isSystem flag that determines whether or not its a system class.

albertshau: it's info here versus summary above to prevent flooding the get all call with details

cdap-client-tests/src/test/java/co/cask/cdap/client/ArtifactClientTestRun.java
albertshau: not true anymore, will update the test

cdap-client/src/main/java/co/cask/cdap/client/ArtifactClient.java
albertshau: ArtifactHttpHandler. was there before but was a no-op

gokulavasan:
Reviewed. Address this comment, https://github.com/caskdata/cdap/pull/3576#discussion_r37937257.
Rest LGTM :+1:

PR #3581, opened 2015-08-13
Document cdap-site properties dynamically

awholegunch:
Change the documentation of the cdap-site properties to be based on cdap-default.xml.

Adds to the build the calling of a python script that writes an rst table that is included in the documentation.

Staged at:
- [**New Appendix**](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT-feature-3.1_cdap_site_xml/en/admin-manual/installation/cdap-site.html#appendix-cdap-site-xml)
- [Current Appendix](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT/en/admin-manual/installation/cdap-site.html#appendix-cdap-site-xml)

**Note:** Even though the branch starts with "3.1", that's a mistake, as this was intended for 3.2...

Fix for issue: https://issues.cask.co/browse/CDAP-3257


cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -20,41 +20,40 @@
...
         <value>cdap</value>
-        <description>Specifies the root namespace</description>
+        <description>Namespace for this CDAP instance</description>
     </property>
 
     <property>
         <name>zookeeper.quorum</name>
         <value>127.0.0.1:2181/${root.namespace}</value>
-        <description>Specifies the zookeeper quorum string</description>
+        <description>Zookeeper address host:port</description>
```
poornachandra: I think we should leave this as zookeeper quorum string. Any reason why we are changing the description.

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -20,41 +20,40 @@
...
+            Specifies the maximum read buffer size in bytes used by the Thrift server;
+            value should be set to greater than the maximum frame sent on the RPC channel
         </description>
     </property>
 
     <property>
         <name>twill.zookeeper.namespace</name>
         <value>/twill</value>
-        <description>Zookeeper namespace prefix for Apache Twill</description>
+        <description>ZooKeeper namespace prefix for Apache Twill</description>
```
poornachandra: We are using both `ZooKeeper` and `Zookeeper` spellings in this doc.

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -68,13 +67,18 @@
     <property>
         <name>twill.jvm.gc.opts</name>
         <value>-verbose:gc -Xloggc:&lt;LOG_DIR&gt;/gc.log -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=10 -XX:GCLogFileSize=1M</value>
-        <description>Java GC options for all Apache Twill containers</description>
+        <description>
+            Java garbage collection options for all Apache Twill containers; &lt;LOG_DIR&gt;
+            is the location of the log directory on each machine; note that the special
```
poornachandra: Changing `is the location of the log directory on each machine` to `is the location of the log directory in the container` makes it clearer.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Thanks. I thought I had fixed all of these, but obviously not.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Fixed. No reason, but I think your choice is better.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Fixed.

awholegunch:
"On-hold" because there are two issues that will take time to address:
- properties with a "final" attribute need to be pulled and placed in a second table, as they cannot be modified;
- there are properties in this file that no longer are used or applicable; insofar as can be easily done, these will be identified and removed.

awholegunch:
Removed the "on-hold" because addressed previous concerns. An "exclusions" file has been added that is used to specify properties that are not to be documented. Documented properties that have a "final" attribute of "true" are placed in a second table after the first.

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -193,11 +197,13 @@
     <property>
         <name>stream.flume.port</name>
         <value>10004</value>
+        <description></description>
```
nitinmotgi: Missing description ?

awholegunch: Yes. I'll make a list and ask for assistance as needed.

awholegunch: These are missing descriptions:
```
stream.flume.port
stream.flume.threads
metrics.num.instances
metrics.max.instances
metrics.num.cores
metrics.memory.mb
metrics.processor.max.instances
explore.start.on.demand

**Comment by awholegunch** [comment]
Fixed missing properties.

File: cdap-common/src/main/resources/cdap-default.xml [hunk 5]
```diff
@@ -20,41 +20,40 @@
     <property>
         <name>root.namespace</name>
         <value>cdap</value>
-        <description>Specifies the root namespace</description>
+        <description>Namespace for this CDAP instance</description>
```
bdmogal: I don't think this is sufficient. Also, this could be confused with namespaces in CDAP. This particular property is used as:
1. The parent (or root) node for all nodes that CDAP creates in ZooKeeper.
2. The directory under which all CDAP data and metadata is stored on HDFS
3. The prefix for all HBase tables that CDAP creates

Also, this value *must* be alphanumeric, otherwise CDAP will throw an error while starting up.

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -20,41 +20,40 @@
...
         <value>127.0.0.1:2181/${root.namespace}</value>
-        <description>Specifies the zookeeper quorum string</description>
+        <description>ZooKeeper quorum string</description>
     </property>
 
     <property>
         <name>zookeeper.session.timeout.millis</name>
         <value>40000</value>
-        <description>Specifies the zookeeper session time out; time unit milliseconds</description>
+        <description>ZooKeeper session time-out in milliseconds</description>
```
bdmogal: is timeout one word? (no hyphen)?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -113,7 +117,7 @@
     <property>
         <name>app.connection.backlog</name>
         <value>20000</value>
-        <description>Max connection backlog of AppFabricServer</description>
+        <description>Max connection backlog of app fabric server</description>
```
bdmogal: App Fabric Server?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -324,43 +318,43 @@
     <property>
         <name>data.local.storage</name>
         <value>${local.data.dir}/ldb</value>
-        <description>Specifies the database directory</description>
+        <description>Database directory for data fabric when in standalone mode</description>
```
bdmogal: should we call out that this is the leveldb directory?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -482,15 +479,15 @@
     <property>
         <name>data.queue.table.name</name>
```
bdmogal: This property seems unused. But please re-check as well.

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -513,21 +510,19 @@
     <property>
         <name>metadata.bind.address</name>
```
bdmogal: metadata properties also seem unused, maybe we can check and remove?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -694,24 +690,25 @@
...
         <value>0.0.0.0</value>
-        <description>Specifies the address for router server to bind to</description>
+        <description>Router server address</description>
     </property>
 
     <property>
         <name>router.webapp.enabled</name>
         <value>false</value>
-        <description>Specifies if webapp listening service should be started at Router</description>
+        <description>Specifies if webapp listening service should be started at router</description>
```
bdmogal: Should also remove the "Specifies..." pattern here for consistency?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -750,43 +747,43 @@
...
         <value>45005</value>
-        <description>Port for metrics query server to listen on</description>
+        <description>Metrics query server bind port</description>
     </property>
 
     <property>
         <name>metrics.data.table.retention.resolution.1.seconds</name>
         <value>7200</value>
-        <description>Retention resolution 1 sec table in seconds</description>
+        <description>Retention resolution of the 1 second table in seconds; default retention period is 2 hours</description>
```
bdmogal: '1 second' resolution table?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -750,43 +747,43 @@
...
     </property>
 
     <property>
         <name>metrics.data.table.retention.resolution.60.seconds</name>
         <value>2592000</value>
-        <description>Retention resolution for 1 minute table (in seconds). Default retention period is 30 days</description>
+        <description>Retention resolution for 1 minute table (in seconds); default retention period is 30 days</description>
```
bdmogal: '1 minute' resolution table?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -750,43 +747,43 @@
...
     </property>
 
     <property>
         <name>metrics.data.table.retention.resolution.3600.seconds</name>
         <value>2592000</value>
-        <description>Retention resolution 1 hour table (in seconds). Default retention period is 30 days</description>
+        <description>Retention resolution 1 hour table (in seconds); default retention period is 30 days</description>
```
bdmogal: '1 hour' resolution table?

cdap-docs/admin-manual/build.sh
```diff
@@ -18,4 +18,21 @@
...
+CDAP_DEFAULT_RST="cdap-default-table.rst"
+CHECK_INCLUDES=$TRUE
+
+function download_includes() {
+  echo "Building rst file from cdap-default.xml..."
+  
+  local includes_dir=${1}
+  if [ ! -d "${includes_dir}" ]; then
+    mkdir ${includes_dir}
+    echo "Creating Includes Directory: ${includes_dir}"
```
bdmogal: nit: Created? or move this echo before ``mkdir``?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -960,19 +963,19 @@
...
         <value>kafka</value>
-        <description>ZK namespace for Kafka</description>
+        <description>Kafka ZooKeeper namespace</description>
     </property>
 
     <property>
         <name>kafka.default.replication.factor</name>
         <value>1</value>
-        <description>Kafka replication factor</description>
+        <description>Kafka replication factor [See Note 1]</description>
```
bdmogal: I don't think we should have a reference from ``cdap-default.xml`` to the rst file. People may want to look at cdap-default.xml (in their source tree) by itself, and would have difficulty in finding the note. Is it necessary for this to be a reference? Can we just pull in the text from Note 1 into the description here?

cdap-docs/admin-manual/source/installation/cdap-site.rst
```diff
@@ -1,410 +1,43 @@
...
-     - User name for running applications in YARN
-   * - ``zookeeper.quorum``
-     - ``127.0.0.1:2181/${root.namespace}``
-     - Zookeeper address host:port
-   * - ``zookeeper.session.timeout.millis``
-     - ``40000``
-     - Zookeeper session time out in milliseconds
+.. include:: ../../build/_includes/cdap-default-table.rst
 
 .. _note 1:
```
bdmogal: I think we should remove these two notes and pull their contents into the properties that they are about.

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,204 @@
...
+    root = tree.getroot()
+    items = []
+    for property in root:
+        item = Item()
+        for attribute in property:
+            item.set_attribute(attribute.tag, attribute.text)
+        if item.name not in exclusions:
+            items.append(item)
+
+    return sorted(items, key=lambda item: item.name)
```
bdmogal: should we sort things here? Properties in ``cdap-default.xml`` may not have alphabetical ordering. They could be ordered by importance (unlikely). Also, related properties may be grouped with each other, even though they may not be alphabetically ordered. For such cases, should we just preserve the ordering from ``cdap-default.xml``?

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,204 @@
...
+        f = open(filepath, 'w')
+        f.write(table)
+        f.close()
+    else:
+        print table
+    
+    
+if __name__ == '__main__':
+    main()
+
```
bdmogal: missing newline at the end of file.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Fixed.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Fixed. One word.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Checked. Removed.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Checked. Removed.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Fixed. Made all the "enable" properties identical. However, the description of this property is baroque.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: What about "the 1-second resolution table"?

cdap-common/src/main/resources/cdap-default.xml
awholegunch: ibid.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: ibid.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Done: pulled in Note 1.

cdap-common/src/main/resources/cdap-default.xml
bdmogal: I meant '1-second', '1-hour' is the resolution of the metrics tables. Just '1-second table' sounded confusing to me.

bdmogal: So essentially, add the word 'resolution' in the description :)

cdap-docs/tools/doc-cdap-default.py
awholegunch: The current table is and was sorted to make it easy for people to look a property up. The properties in the cdap-default are not alphabetical; they are thematic. I will see if I can bring in the group titles, because without them, the order appears arbitrary if it isn't sorted.

cdap-docs/tools/doc-cdap-default.py
awholegunch: fixed.

cdap-docs/admin-manual/source/installation/cdap-site.rst
awholegunch: Done.

cdap-docs/admin-manual/build.sh
awholegunch: Moved before ``mkdir``.

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Changed description to "Database directory for LevelDB, used for data fabric in standalone mode"

awholegunch:
The appendix has been revised extensively and the page re-designed. Please look at [the new page](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT-feature-3.1_cdap_site_xml/en/admin-manual/installation/cdap-site.html#appendix-cdap-site-xml) when re-built. 

In particular, consider:
- the titles used for the sections (Can they be simplified, or combined, or made more consistent with other usage in CDAP?)
- the order of the sections (What order would make sense? Alphabetical? The "General" first, and then alphabetical?)
- the order within the sections (Perhaps alphabetical?)
- are there any other properties to be removed or excluded?
- What about ``data.tx.janitor.enable``? These Tephra-related properties: are they valid?

bdmogal:
> the titles used for the sections
- I think YARN Services and Master HTTP services can be combined into 'Master'
- Queue-Related -> Queue
- Log Collection Service can be combined with Logging
- I don't know if a distinction needs to be made between Metrics and Metrics Processor for describing properties
- Workflow and Scheduler have one each. Perhaps they can be combined into a Workflow and Scheduler section.
- Artifact and App Template can be merged with AppFabric
- Maybe we should call DataFabric as Datasets, and AppFabric as Applications, since Datasets and - Applications seem to be more commonly used these days. 
- Mapreduce Info properties can be moved to the General Section as well, we should treat them similarly as the ZooKeeper properties.
- Also consider standardizing on either singular or plural in the section headings (Queue vs Queues)
- Another thing to consider is that Queue and Workflow/Scheduler are essentially part of Applications. However, it may be ok to keep them separate right now.

> the order of the sections (What order would make sense? Alphabetical? The "General" first, and then alphabetical?)
I think General first, followed by an alphabetical list is fine. Also, alphabetical within a section is good too.

awholegunch:
Before this is merged, because of the potential for breaking, a clean build and test is required. Once the last edits are made, this will be performed before merging.

awholegunch:
@bdmogal Correct me if I'm wrong, but I think you mentioned to me that we should remove the "final" section and merge those two properties into the rest, but marked appropriately?

bdmogal:
@awholegunch my memory is defeating me today, I don't remember that conversation. However, if there's only two properties, then having a 'Final?' column everywhere may be overkill? I don't feel strongly either way.

Perhaps we can keep them in a separate table as-is for now, and when we have more of those to warrant a separate column everywhere, we can merge them in?

awholegunch:
Made a number of changes, and compared the original cdap-defaults.xml to the new one. These 12 properties were removed, and two (``metrics.query.bind.address`` and ``router.ssl.server.port``) are included but excluded from the documentation:
```
appfabric.environment
data.queue.table.name
log.collection.bind.address
log.collection.bind.port
log.query.bind.address
log.query.bind.port
metadata.bind.address
metadata.bind.port
metadata.program.run.history.keepdays
security.auth.server.bind.port
stream.flume.port
stream.flume.threads

**Comment by awholegunch** [comment]
Time to take another look. New version built and staged: [New Appendix](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT-feature-3.1_cdap_site_xml/en/admin-manual/installation/cdap-site.html#appendix-cdap-site-xml)

File: cdap-docs/admin-manual/source/installation/cdap-site.rst [hunk 20]
```diff
@@ -1,410 +1,26 @@
...
-   * - ``yarn.user``
-     - ``yarn``
-     - User name for running applications in YARN
-   * - ``zookeeper.quorum``
-     - ``127.0.0.1:2181/${root.namespace}``
-     - Zookeeper address host:port
-   * - ``zookeeper.session.timeout.millis``
-     - ``40000``
-     - Zookeeper session time out in milliseconds
+Any of its values can be over-ridden by defining a modifying value in the
```
bdmogal: other than the final ones?

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  License for the specific language governing permissions and limitations under
+  the License.
+  -->
+""" % datetime.now().year
+
+
+class Item:
```
bdmogal: In this file, should we make the tab indent to be 2 spaces instead of 4 to be consistent with all our code?

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+    XML_DESCRIP_CLOSE = '        </description>\n'
+    XML_FINAL         = '        <final>true</final>\n'
+    XML_PROP_CLOSE    = '    </property>\n\n'
+    XML_CONFIG_CLOSE  = '</configuration>\n'
+
+    xml = XML_CONFIG_OPEN
+    keys = defaults.keys()
+    for key in FIRST_TWO_SECTIONS:
+        keys.remove(key)
+    keys.sort()
```
bdmogal: nit: this (or line 330) now makes hdfs.lib.dir (which references hdfs.namespace) appear before hdfs.namespace :-)

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -17,1205 +17,1535 @@
...
     </property>
 
     <property>
-        <name>zookeeper.session.timeout.millis</name>
-        <value>40000</value>
-        <description>Specifies the zookeeper session time out; time unit milliseconds</description>
+        <name>hdfs.namespace</name>
+        <value>/${root.namespace}</value>
+        <description>
+            Namespace for HDFS files written by CDAP
```
bdmogal: Root directory for HDFS files written by CDAP?

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+            name = name[0:-len(CONFIG)]
+        underline = "-" * len(name)
+        rst = "\n%s\n%s\n%s" % (name, underline, RST_TABLE_HEADER)
+        return rst
+
+    def display(self):
+        print "\nSection %s" % self.name
+       
+
+class PIParser(ET.XMLTreeBuilder):
```
bdmogal: what does PIParser stand for? Should the name be more descriptive?

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+    --help  This help statement
+"""
+
+def main():
+    """ Main program entry point.
+    """
+    if len(sys.argv) == 1:
+        print_usage()
+        return
+    elif sys.argv[1] == '--rebuild':
```
bdmogal: can we use optparse or something similar to parse arguments, instead of doing it ourselves? 

One of my concerns is there are no named single character alternatives for each of these arguments.

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+        if str(item.__class__) == '__main__.Section':
+            section_counter += 1
+            if not props_only:
+                item.display()                
+        else:
+            properties_counter += 1
+            item.display()
+    print "Sections: %d\nProperties: %d" % (section_counter, properties_counter)
+
+def print_usage():
```
bdmogal: You can get Usage in a standard form automatically if you use ``optparse``

cdap-docs/tools/doc-cdap-default.py
bdmogal: Also, what does ``rebuild`` stand for? It can just be ``build``, right? Unless we're carrying over some state from a previous run?

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+    elif sys.argv[1] == '--load-props':
+        load_xml(props_only=True)
+        return
+    elif sys.argv[1] == '--load-xml':
+        load_xml()
+        return
+    elif sys.argv[1] == '--print-rst':
+        load_create_rst('', '')
+        return
+    elif sys.argv[1] == '--help':
```
bdmogal: nit: Is this needed? Lines 429-431 will handle this, right? Ok if you want to call it out though.

bdmogal: Unless, you exit with a success code after --help and a non-success code after line 431

cdap-docs/tools/doc-cdap-default.py
bdmogal: O I see why you call it rebuild now. Its essentially re-formatting cdap-default.xml though, right?

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+        defaults[section_name] = section
+        section = []  
+    
+    print "Sections: %d Keys: %d Properties: %d" % (section_counter, len(defaults.keys()), properties_counter)
+
+    # Build XML file
+    XML_CONFIG_OPEN   = '<configuration>\n'
+    XML_SECTION_SUB   = "\n    <!-- %s -->\n\n"
+    XML_PROP_OPEN     = '    <property>\n'
+    XML_NAME_SUB      = "        <name>%s</name>\n"
```
bdmogal: same question about tab indent here too.

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+            properties_counter += 1
+            section.append(item)
+    if section: # End old section
+        defaults[section_name] = section
+        section = []  
+    
+    print "Sections: %d Keys: %d Properties: %d" % (section_counter, len(defaults.keys()), properties_counter)
+
+    # Build XML file
+    XML_CONFIG_OPEN   = '<configuration>\n'
```
bdmogal: Actually, can't we use a standard XML library here instead of manipulating strings?

cdap-docs/tools/doc-cdap-default.py
```diff
@@ -0,0 +1,434 @@
...
+                if prop.name in exclusions:
+                    print "but in the list of exclusions"
+            xml += XML_DESCRIP_CLOSE
+            if prop.final:
+                xml += XML_FINAL
+            xml += XML_PROP_CLOSE
+    xml += XML_CONFIG_CLOSE
+    
+    filepath = default_xml_filepath('_revised.xml')
+    f = open(filepath, 'wb')
```
bdmogal: for all of these ``open`` calls, what happens if the file already exists?

bdmogal:
@awholegunch did one pass. In addition:
1. Let's run a DUT build
2. Beam to cluster and ensure CDAP starts
3. Run Integration tests on the cluster. I can help you with this.

bdmogal:
The previous comment was because it is impossible to review the changes to ``cdap-default.xml`` because of the re-ordering of properties

cdap-docs/tools/doc-cdap-default.py
awholegunch: Correct; it takes it in and rebuilds it.
I can change to using optparse. Originally this app had one function with two options. Then another function was added. And then I thought of something cool it could do. And then I added...so, it's at the point to replace sys.argv...

cdap-docs/tools/doc-cdap-default.py
awholegunch: It stands for "Processing Instructions Parser", and came from this site: http://effbot.org/zone/element-pi.htm
Perhaps appropriate comment explaining it?

cdap-common/src/main/resources/cdap-default.xml
awholegunch: Fixed.

cdap-docs/tools/doc-cdap-default.py
awholegunch: Well, Python standard is 4...and given that whitespace is important in Python, I find the extra space helpful in reading the code.

cdap-docs/admin-manual/source/installation/cdap-site.rst
awholegunch: Fixed. Moved all this text about final into this file rather than being in the Python script.

cdap-docs/tools/doc-cdap-default.py
awholegunch: Add doc string with usage to class.

cdap-docs/tools/doc-cdap-default.py
awholegunch: Maybe. We have "Processing Instructions" (ie comments)...that makes it more difficult. I was able to remove all the inline ones that were causing trouble; I can see if the remaining ones (the sections) can be handled by a standard library.

awholegunch:
+1

cdap-docs/tools/doc-cdap-default.py
awholegunch: If the file exists or is open, it's just overwritten. As for error handling: I don't think it's worth the effort, because this is not a command that would be run automatically...at least, not my intention. You would run it manually and check the results. More of an editing tool. I did it to because I was getting tired trying to do it manually.

cdap-docs/tools/doc-cdap-default.py
awholegunch: You can't have it both ways. (Alphabetical *and* semantic ordering)... :8ball:

cdap-docs/tools/doc-cdap-default.py
awholegunch: See above.

cdap-docs/tools/doc-cdap-default.py
bdmogal: :+1:

awholegunch:
Updated again: [Appendix](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT-feature-3.1_cdap_site_xml/en/admin-manual/installation/cdap-site.html#appendix-cdap-site-xml)

cdap-docs/tools/doc-cdap-default.py
awholegunch: Looked into this. Because we want to generate "human-readable" XML (an oxymoron to some), and as I haven't found a library that does it to our standards (sorted in a particular order), I suggest we leave as-is. It's really just a pretty-printing feature tacked on the standard library...

bdmogal:
This LGTM - My only concern being the hand-generated XML. As long as we're only using that for generating docs, and not checking it back in, I'm ok with that. We should test this on a cluster and on standalone, so that's pending and in-progress.

bdmogal:
Build: http://builds.cask.co/browse/CDAP-DUT2505-8
Integration tests passed on standalone. 
Testing on cluster pending.

awholegunch:
I've regenerated the XML file using the Cask standard of two spaces for indentation of XML files.

bdmogal:
This is good to merge. All Integration tests passed on a CDH5.4 cluster.

awholegunch:
I'm not going to merge this until Monday morning, when people are around, just in case something goes wrong...

PR #3583, opened 2015-08-13
adds media queries to adjust height of side panels

astewes:
*  Adds media query to set dynamic height on item body of Source, Transform, and Sink
*  Adds media query to adjust height of right side panel item headings (collapsed state)

panels


ajainarayanan:
:+1: LGTM

PR #3584, opened 2015-08-13
CDAP-3272 add ability to explicitly specify artifact classes

albertshau:
Through the RESTful API, explicit plugins are given through a
header. For system artifacts, the list of plugins can be placed
in a json config file similar to how it is done for
ApplicationTemplate now.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
@@ -210,7 +210,7 @@ public class ArtifactRepository {
 
     Location artifactLocation = Locations.toLocation(artifactFile);
     try (CloseableClassLoader parentClassLoader = artifactClassLoaderFactory.createClassLoader(artifactLocation)) {
-      ArtifactClasses artifactClasses = artifactInspector.inspectArtifact(artifactId, artifactFile, parentClassLoader);
+      ArtifactClasses artifactClasses = inspectArtifact(artifactId, artifactFile, null, parentClassLoader);
       validatePluginSet(artifactClasses.getPlugins());
       ArtifactMeta meta = new ArtifactMeta(artifactClasses, ImmutableSet.<ArtifactRange>of());
       return artifactStore.write(artifactId, meta, Files.newInputStreamSupplier(artifactFile));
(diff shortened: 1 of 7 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2507

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -238,16 +238,62 @@ public ArtifactDetail addArtifact(Id.Artifact artifactId, File artifactFile,
...
       return artifactStore.write(artifactId, meta, Files.newInputStreamSupplier(artifactFile));
+    } finally {
+      parentClassLoader.close();
+    }
+  }
+
+  private ArtifactClasses inspectArtifact(Id.Artifact artifactId, File artifactFile,
+                                          @Nullable Set<PluginClass> additionalPlugins,
+                                          ClassLoader parentClassLoader) throws IOException, InvalidArtifactException {
+    ArtifactClasses artifactClasses = artifactInspector.inspectArtifact(artifactId, artifactFile, parentClassLoader);
```
gokulavasan: So we treat the config file as plugin info provided in addition to the artifact inspection? This is different from the logic in PluginRepository where we do ``configureByFile`` or ``configureByInspection``. Is this because now this info can be provided as part of the POST request to addArtifact instead of a File and thus makes it easy to do 'addition' instead of 'only' logic?

gokulavasan: Also do we check somewhere to make sure we don't have same plugin name, type in both ``additionalPlugins`` and ``artifactClasses`` ?

gokulavasan:
Couple of comments

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
albertshau: yeah was debating whether to do only or addition.  The reason decided on in addition is that I could think of a case where you've included a 3rd party plugin with a bunch of other explicitly annotated plugins, in which case you would want it to be in addition rather than only. But could not think of a situation where you would want 'only' when there are annotated plugins in the artifact.

albertshau: am not validating the additional plugins, will add that.  I think it is ok to make the plugins specified by the user overwrite what is inspected.

gokulavasan:
LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
gokulavasan: Since we only look for classes with ``Plugin`` annotation, we should be good with this approach

PR #3585, opened 2015-08-13
CDAP-482: Remove HttpServiceSpecificationCodec after end of support

jwang47:
https://issues.cask.co/browse/CDAP-482
http://builds.cask.co/browse/CDAP-DUT2508


chtyim:
LGTM

PR #3586, opened 2015-08-13
[CDAP-3333] Predefined Plugins Phase 2

elfenheart:
This PR should be merge **after** #3575 

edittemplate


elfenheart:
will reopen against phase 1 PR

PR #3591, opened 2015-08-13
CDAP-3310 pick up artifact version from manifest if not specified

albertshau:
Use the Bundle-Version from the jar manifest as the artifact
version if a version is not explicitly given. Otherwise, use
whatever is explicitly given.

Also now that http-netty version has been updated, simplifying
add artifact endpoint by letting the exception handler handle
exceptions rather than handling them in the method.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2512

gokulavasan:
LGTM :+1:

PR #3592, opened 2015-08-13
CDAP-2428: Suppress logback logging in CLI

jwang47:
https://issues.cask.co/browse/CDAP-2428
http://builds.cask.co/browse/CDAP-DUT2514

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
@@ -16,6 +16,8 @@
 
 package co.cask.cdap.cli;
 
+import ch.qos.logback.classic.Level;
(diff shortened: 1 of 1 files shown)


cdap-cli/src/main/java/co/cask/cdap/cli/CLIMain.java
```diff
@@ -237,6 +238,11 @@ public CLI getCLI() {
   }
 
   public static void main(String[] args) {
+    // disable logback logging
```
anwar6953: why not import?
why specify full class name here?

jwang47: Oh, yeah import would be better.

anwar6953:
LGTM

PR #3593, opened 2015-08-13
CDAP-2777 Squashing all commits from PR #3179

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-2777

Build : http://builds.cask.co/browse/CDAP-DUT2515

Note: This PR doesn't add any validation logic to make sure that datasets/streams with different configuration but with same names are used in two different programs. It will just use the last configuration for dataset/stream.


albertshau:
:+1:

PR #3596, opened 2015-08-14
Use the correct API to check for file existence.

sreevatsanraman:
* Cherry picking changes for logback container fix CDAP-3362 into release/3.0


chtyim:
LGTM

PR #3598, opened 2015-08-14
[CDAP-3374] Warning messages indicating that the Runrecords with status running

henry-cask:
This fix move getting the invalid run records early to avoid excessive warning logs when processing per program type.


henry-cask:
Build: http://builds.cask.co/browse/CDAP-DUT2517-1

PR #3599, opened 2015-08-14
(CDAP-3008) Refactor the way of how to submit Spark program

chtyim:
- A new interface, SparkSubmitter, is created for abstracting out submission
- Provide two implementations, Local and Distributed SparkSubmitter
  - Currently they are implemented the same way
  - Will diverge when using yarn-cluster mode

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/AbstractSparkSubmitter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/AbstractSparkSubmitter.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.internal.app.runtime.spark;
+
+import co.cask.cdap.api.spark.SparkSpecification;
+import co.cask.cdap.common.lang.ClassLoaders;
+import co.cask.cdap.common.lang.WeakReferenceDelegatorClassLoader;
+import co.cask.cdap.internal.app.runtime.distributed.LocalizeResource;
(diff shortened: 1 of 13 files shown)


chtyim:
Build passed: https://builds.cask.co/browse/CDAP-DUT2523

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/AbstractSparkSubmitter.java
```diff
@@ -0,0 +1,180 @@
...
+  @Override
+  public final <V> ExecutionFuture<V> submit(final ExecutionSparkContext sparkContext, Map<String, String> configs,
+                                             List<LocalizeResource> resources, File jobJar, final V result) {
+
+    final SparkSpecification spec = sparkContext.getSpecification();
+    final List<String> args = createSubmitArguments(spec.getMainClassName(), configs, resources, jobJar);
+
+    // Spark submit is called from this executor
+    // Use an executor to simplify logic that is needed to interrupt the running thread on stopping
+    final ExecutorService executor = Executors.newSingleThreadExecutor(new ThreadFactory() {
```
henry-cask: Could we just move this as member variable instead instantiating if in the method?

chtyim: Can't the executor thread ties to the submit call. The same submitter can have multiple spark job submitted through it (e.g. in Workflow fork).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/SparkRuntimeService.java
```diff
@@ -158,65 +161,75 @@ protected void startUp() throws Exception {
...
-      new WeakReferenceDelegatorClassLoader(sparkClassLoader));
+    // First, only submit the job if this service is in RUNNING state.
+    // This service will not be in running state if the triggeredShutdown() method
+    // is invoked between startUp() and run() methods.
+    //
+    // The second clause is to guard against case where job is submitted through submitSpark.call(),
+    // but before the atomic reference is updated, the triggeredShutdown() method is invoked and cancelled
+    // the old Future reference inside the atomic reference. If that happened, need to cancel the submitted job.
+    if (isRunning() && submissionFuture.getAndSet(submitSpark.call()).isCancelled()) {
+      submissionFuture.get().cancel(true);
```
henry-cask: Should we return immediately if this is happening?

chtyim: We cannot. We need to invalid the transaction, which happens below.

henry-cask:
Did a pass and have couple comments.

henry-cask:
Thanks for the reply. LGTM

PR #3600, opened 2015-08-14
Wait till the workflow stop before calling the current endpoint.

bdmogal:
Backporting #3532, to avoid unit test failures during release


sagarkapare:
LGTM

bdmogal:
Merging. This fix is already working fine on develop.

PR #3602, opened 2015-08-14
turn off host key check on initial ssh

dereklwood:
turns off hostkey checking on the initial ssh attempt.  This needs to be ported to release branches as well.


bdmogal:
:+1:

PR #3603, opened 2015-08-14
Fixes url construction for uploading an app in express server

ajainarayanan:
-Fixes the app upload api url in node server.


elfenheart:
:+1: LGTM

PR #3604, opened 2015-08-14
[CDAP-3374] Warning messages indicating that the Runrecords with status running

henry-cask:
Move the check for Program ID for program type and run id early to avoid excessive logging for invalid run records with wrong program type for the run records.

Previously for each program type we have 2 checks, one at the beginning to filter run records with RUNNING status and actually does not have container running for that program type, and then later check if the run record id is actually belong to that program type.


henry-cask:
Successful build: http://builds.cask.co/browse/CDAP-DUT2526-1

chtyim:
LGTM

henry-cask:
Thanks, merging ...

PR #3605, opened 2015-08-14
Correct the name of the property used to enable CDAP Explore service.

awholegunch:
A check was made in the documentation for other cases of this error, and this was the only instance found.

Fixes https://issues.cask.co/browse/CDAP-3404.


sreevatsanraman:
LGTM

PR #3606, opened 2015-08-14
[CDAP-3333] Predefined Plugins Phase 3

elfenheart:
Disabling input when user already defines the field in the template

disablefields


ajainarayanan:
1. https://issues.cask.co/browse/CDAP-3422 - Cloned or Imported adapter configs will have implicitly have custom templates but the plugin will not be recognized as custom template so properties will become editable.
2. We should have a confirmation before deleting an adapter.
3. The local cache for template config needs ```angular.copy``` talked to @elfenheart about it.
4. Minor thing but we need to confirm on a design and have the stylings in a better way (the schema table and the info icons at the very least needs some love).

ajainarayanan:
Also if we could squash some commits that would be awesome :laughing:

elfenheart:
@ajainarayanan Addressed number 2 and 3. Item 1 and 4 will come in later stage

ajainarayanan:
:+1: LGTM

PR #3607, opened 2015-08-14
[CDAP-3397] Fixes empty table headers for adapter metrics in adapter detailed view

ajainarayanan:
- Hide empty table header in adapter metrics view


elfenheart:
:+1: LGTM

PR #3608, opened 2015-08-14
Improve logic for plugin inspection

chtyim:
- Added filter to only pickup @Plugin class only without loading every classes inside a JAR.


albertshau:
:+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2530-2

PR #3611, opened 2015-08-14
CDAP-3367 add a way to delete artifacts

albertshau:
Artifact deletion can be unsafe and it is assumed the user knows
exactly what they are doing.

### DIFF
--- a/cdap-client-tests/src/test/java/co/cask/cdap/client/ArtifactClientTestRun.java
+++ b/cdap-client-tests/src/test/java/co/cask/cdap/client/ArtifactClientTestRun.java
@@ -40,7 +40,6 @@ import co.cask.cdap.proto.artifact.ArtifactSummary;
 import co.cask.cdap.proto.artifact.PluginInfo;
 import co.cask.cdap.proto.artifact.PluginSummary;
 import co.cask.cdap.test.XSlowTests;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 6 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2525

cdap-client-tests/src/test/java/co/cask/cdap/client/ArtifactClientTestRun.java
```diff
@@ -82,6 +81,10 @@ public InputStream getInput() throws IOException {
   public void setUp() throws Throwable {
     super.setUp();
     artifactClient = new ArtifactClient(clientConfig, new RESTClient(clientConfig));
+    for (ArtifactSummary artifactSummary : artifactClient.list(Id.Namespace.DEFAULT, false)) {
```
gokulavasan: Do we always expect some artifacts to be there (so that ``artifactClient.delete`` gets invoked) ?  Or should we add an explicit call in the test and check that the artifact has been deleted?

gokulavasan:
One question about testing of artifact deletion. Rest LGTM

cdap-client-tests/src/test/java/co/cask/cdap/client/ArtifactClientTestRun.java
albertshau: this is more of a catch all to clean up state since there is no delete all in a namespace.  A lot of other tests in the suite don't know about artifacts, they just deploy apps, which leaves those artifacts around even if the app is deleted.

gokulavasan:
LGTM :+1:

PR #3614, opened 2015-08-14
CDAP-3413: Remove guava dependencies in Schema

jwang47:
https://issues.cask.co/browse/CDAP-3413
http://builds.cask.co/browse/CDAP-DUT2537

Not completely necessary for 3.2 since we're going to have a copy of guava in CDAP code, but this will be needed eventually.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
@@ -17,24 +17,9 @@
 package co.cask.cdap.api.data.schema;
 
 import co.cask.cdap.api.annotation.Beta;
-import co.cask.cdap.internal.io.SQLSchemaParser;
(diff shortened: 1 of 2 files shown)


cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
```diff
@@ -123,228 +108,10 @@ public String toString() {
...
   }
 
-  /**
-   * Parse the given JSON representation, as returned by {@link #toString()} into a Schema object.
-   *
-   * @param schemaJson the json representation of the schema
-   * @return the json representation parsed into a schema object
-   * @throws IOException if there was an exception parsing the schema
-   */
-  public static Schema parseJson(String schemaJson) throws IOException {
```
bdmogal: these are all incompatible changes in API. Need to deprecate for 2 releases.

PR #3616, opened 2015-08-14
CDAP-3405 fix cube adapter json config - add transform to add a count…

gokulavasan:
… field and also update the cubeServiceApp version


sreevatsanraman:
LGTM

PR #3617, opened 2015-08-14
[CDAP-3408] Fixed Hive/Explore version detection logic for CDH 5.[2,3].

bdmogal:
Without this fix, Explore was using Hive14ExploreService for CDH 5.[2,3], which uses Hive 0.13, thus causing NoSuchMethodErrors due to incompatibilities.

Cherry picking #3608 for develop


jwang47:
LGTM

PR #3620, opened 2015-08-15
CDAP-3388: Create cdap-spi with interfaces that abstract StreamEventRecordFormat and related classes

jwang47:
https://issues.cask.co/browse/CDAP-3388
http://builds.cask.co/browse/CDAP-DUT2534


cdap-spi/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+      <artifactId>cdap-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
```
anwar6953: add newline

chtyim:
Better call the new module `cdap-api-common` because core module usually refers to the implementation. Rest LGTM.

PR #3622, opened 2015-08-15
(Cdap-20) Incremental effort to remove guava from api

chtyim:
There are three changes:

1. Replace usage of TypeToken to Type in FieldAccessor.
2. Change Reflections.visit to take Type instead of TypeToken
3. Remove usage of Guava Joiner and Splitter in cdap-api

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/common/RuntimeArguments.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/common/RuntimeArguments.java
@@ -16,8 +16,6 @@
 
 package co.cask.cdap.api.common;
 
-import com.google.common.base.Splitter;
-
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -42,7 +40,11 @@ public final class RuntimeArguments {
   public static Map<String, String> fromPosixArray(String[] args) {
     Map<String, String> kvMap = new HashMap<>();
     for (String arg : args) {
-      kvMap.putAll(Splitter.on("--").omitEmptyStrings().trimResults().withKeyValueSeparator("=").split(arg));
+      int idx = arg.indexOf('=');
+      int keyOff = arg.startsWith("--") ? "--".length() : 0;
+      String key = idx < 0 ? arg.substring(keyOff) : arg.substring(keyOff, idx);
+      String value = idx < 0 ? "" : arg.substring(idx + 1);
+      kvMap.put(key, value);
     }
     return kvMap;
   }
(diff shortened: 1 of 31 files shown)


cdap-api/src/main/java/co/cask/cdap/api/common/RuntimeArguments.java
```diff
@@ -42,7 +40,11 @@ private RuntimeArguments() {
   public static Map<String, String> fromPosixArray(String[] args) {
     Map<String, String> kvMap = new HashMap<>();
     for (String arg : args) {
-      kvMap.putAll(Splitter.on("--").omitEmptyStrings().trimResults().withKeyValueSeparator("=").split(arg));
+      int idx = arg.indexOf('=');
+      int keyOff = arg.startsWith("--") ? "--".length() : 0;
+      String key = idx < 0 ? arg.substring(keyOff) : arg.substring(keyOff, idx);
+      String value = idx < 0 ? "" : arg.substring(idx + 1);
```
gokulavasan: what if we have arg = "--validKey=" then arg.substring(idx+1) will be out of bounds

chtyim: Good catch.

cdap-common/src/main/java/co/cask/cdap/internal/io/FieldAccessorGenerator.java
```diff
@@ -126,8 +126,9 @@ private void initializeReflectionField(GeneratorAdapter mg, Field field) {
     mg.visitTryCatchBlock(beginTry, endTry, catchHandle, Type.getInternalName(Exception.class));
     mg.mark(beginTry);
 
-    // Field field = findField(classType, "fieldName")
+    // Field field = Fields.findField(classType, "fieldName")
```
gokulavasan: Fields.findField(TypeToken.of(classType), "fieldName") ?

gokulavasan:

cdap-api/src/main/java/co/cask/cdap/api/common/RuntimeArguments.java
chtyim: Actually it won't. The substring requires the begin index > 0 and <= string.length.

gokulavasan: Ah I see. Ok.

cdap-common/src/main/java/co/cask/cdap/internal/io/FieldAccessorGenerator.java
chtyim: Fixed.

gokulavasan:
LGTM :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2451-6

PR #3623, opened 2015-08-16
(CDAP-20) Remove almost all guava usages from cdap-api

chtyim:
- Almost all usages are remove except the following:
  - TypeToken
  - BiMap and related implementation
  - Multimap and related implementation

- For TypeToken usages, it's all internal now (meaning not leaking through any interfaces), hence we can just copy the TypeToken (and related classes) into cdap-api
- For BiMap usages, we can replace it with two maps or copy it from guava as well
- For Multimap, it is problematic since it is leaked as part of public API through HttpServiceRequest/Responder. We can do deprecation, however it means we cannot remove guava dependency until two versions later, which we needed for the Cassandra usage case now.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
@@ -19,24 +19,24 @@ package co.cask.cdap.api.data.schema;
 import co.cask.cdap.api.annotation.Beta;
 import co.cask.cdap.internal.io.SQLSchemaParser;
 import co.cask.cdap.internal.io.SchemaTypeAdapter;
-import com.google.common.base.Preconditions;
-import com.google.common.base.Throwables;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableBiMap;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
-import com.google.common.io.CharStreams;
 import com.google.gson.stream.JsonWriter;
 
 import java.io.IOException;
 import java.io.Reader;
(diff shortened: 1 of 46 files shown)


cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
```diff
@@ -775,15 +839,14 @@ private String buildString() {
...
-    try {
-      SCHEMA_TYPE_ADAPTER.write(writer, this);
-      writer.close();
-      return builder.toString();
+
+    StringWriter writer = new StringWriter();
+    try (JsonWriter jsonWriter = new JsonWriter(writer)) {
+      SCHEMA_TYPE_ADAPTER.write(jsonWriter, this);
     } catch (IOException e) {
       // It should never throw IOException on the StringBuilder Writer, if it does, something very wrong.
```
gokulavasan: comment update: StringWriter instead of StringBuilderWriter

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/KeyValueTable.java
```diff
@@ -226,7 +225,9 @@ public void write(KeyValue<byte[], byte[]> keyValue) throws IOException {
       private boolean closed = false;
       @Override
       protected KeyValue<byte[], byte[]> computeNext() {
-        Preconditions.checkState(!closed);
+        if (closed) {
```
gokulavasan: Why the change in logic?

chtyim: The old logic is not good. Should always use the endOfData to signal end of data instead of throwing illegal state exception.

gokulavasan:
LGTM :+1:

cdap-api/src/main/java/co/cask/cdap/api/data/schema/Schema.java
chtyim: fixed

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2451-7

PR #3625, opened 2015-08-17
CDAP-3437 need the quotes to force string comparison

duddun:
When doing if [ A == B ] string variable comparisons in bash, safest thing is to force both sides to be string.

This will be ported back to 3.1, 3.0 and 2.8 as well.


wolf31o2:
This looks fine. Reference this PR when you make the others. :smile:

PR #3626, opened 2015-08-17

jwang47:
https://issues.cask.co/browse/CDAP-3419

Output looks like this:

```
Running class co.cask.cdap.data.tools.SimpleHBaseQueueDebugger
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/opt/cdap/master/lib/ch.qos.logback.logback-classic-1.0.9.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/usr/lib/hadoop/lib/slf4j-log4j12-1.7.5.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/usr/lib/zookeeper/lib/slf4j-log4j12-1.6.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
2015-08-19 20:34:04,485 - WARN  [main:c.c.c.c.c.Configuration@570] - security.auth.server.address is deprecated. Instead, use security.auth.server.bind.address
2015-08-19 20:34:04,496 - WARN  [main:c.c.c.c.c.Configuration@570] - security.auth.server.address is deprecated. Instead, use security.auth.server.bind.address
2015-08-19 20:34:05,522 - INFO  [main:c.c.c.c.g.LocationRuntimeModule$HDFSLocationModule@88] - HDFS namespace is /cdap
2015-08-19 20:34:05,742 - WARN  [main:o.a.h.u.NativeCodeLoader@62] - Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Results for queue queue:///default/WordCount/WordCounter/counter/queue: min tx timestamp: n/a
Results for queue queue:///default/WordCount/WordCounter/splitter/wordArrayOut: min tx timestamp: n/a
Results for queue queue:///default/WordCount/WordCounter/splitter/wordOut: min tx timestamp: n/a
Results for queue queue:///default/PurchaseHistory/PurchaseFlow/reader/queue: min tx timestamp: 1440016236129
Total results for all queues: min tx timestamp: 1440016236129
```

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/SimpleHBaseQueueDebugger.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/SimpleHBaseQueueDebugger.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data.tools;
+
+/**
+ * {@link HBaseQueueDebugger} with simple defaults.
+ */
+public class SimpleHBaseQueueDebugger {
+  public static void main(String[] args) throws Exception {
+    if (args.length >= 1 && args[0].equals("help")) {
+      System.out.printf("Displays the minimum transaction time for all events in all queues.\n\n");
+      System.out.printf("Uses HBaseQueueDebugger with the following options:\n");
+      System.out.printf("-D" + HBaseQueueDebugger.PROP_SHOW_PROGRESS + "=false\n");
+      System.out.printf("-D" + HBaseQueueDebugger.PROP_SHOW_TX_TIMESTAMP_ONLY + "=true\n");
+      return;
+    }
+
+    System.setProperty(HBaseQueueDebugger.PROP_SHOW_PROGRESS, "false");
+    System.setProperty(HBaseQueueDebugger.PROP_SHOW_TX_TIMESTAMP_ONLY, "true");
+    HBaseQueueDebugger.main(args);
+  }
+}
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/HBaseQueueDebugger.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/HBaseQueueDebugger.java
@@ -17,14 +17,32 @@
 package co.cask.cdap.data.tools;
 
 import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.flow.FlowSpecification;
+import co.cask.cdap.api.flow.FlowletConnection;
+import co.cask.cdap.app.ApplicationSpecification;
+import co.cask.cdap.app.guice.AppFabricServiceRuntimeModule;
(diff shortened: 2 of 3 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/SimpleHBaseQueueDebugger.java
```diff
@@ -0,0 +1,28 @@
...
+package co.cask.cdap.data.tools;
+
+/**
+ * {@link HBaseQueueDebugger} with simple defaults.
+ */
+public class SimpleHBaseQueueDebugger {
+  public static void main(String[] args) throws Exception {
+    System.setProperty(HBaseQueueDebugger.PROP_SHOW_PROGRESS, "false");
+    System.setProperty(HBaseQueueDebugger.PROP_SHOW_TX_TIMESTAMP_ONLY, "true");
+    HBaseQueueDebugger.main(args);
```
poornachandra: Just for completeness sake, can you show SimpleHBaseQueueDebugger's help here when args[1] == "help"? Otherwise, we'll be showing HBaseQueueDebugger's help, right?

poornachandra:
I see that `minWritePointer` is initialized to `Long.MAX_VALUE`. What happens if `minWritePointer` never gets updated in `visitRow` method, either because `state == ConsumerEntryState.PROCESSED` or `stateValue == null`? Do we then print out `Long.MAX_VALUE / TxConstants.MAX_TX_PER_MS` as the timestamp?

poornachandra:
Also, why is your sample output having such large timestamps? Is it because of the previous comment?

poornachandra:
Is it easy to suppress the SLF4J messages in the output?

poornachandra:
@jwang47 Few comments, please address them.

jwang47:
Yes, the output Long.MAX timestamps due to initializing to Long.MAX. Will fix this. Also, it's possible to suppress the SLF4J output.

jwang47:
Actually, not sure if it's easy to suppress the SLF4J output here at the moment. We have multiple bindings for SLF4J in the cdap-master classpath, so we'll always have the

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/opt/cdap/master/lib/ch.qos.logback.logback-classic-1.0.9.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/usr/lib/hadoop/lib/slf4j-log4j12-1.7.5.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/usr/lib/zookeeper/lib/slf4j-log4j12-1.6.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
```

and

```
2015-08-19 20:34:04,485 - WARN  [main:c.c.c.c.c.Configuration@570] - security.auth.server.address is deprecated. Instead, use security.auth.server.bind.address
2015-08-19 20:34:04,496 - WARN  [main:c.c.c.c.c.Configuration@570] - security.auth.server.address is deprecated. Instead, use security.auth.server.bind.address
```

is due to the cdap-default.xml which we fixed in release/3.2

jwang47:
Updated description with new sample output. Now shows "n/a" instead of Long.MAX if there aren't any timestamps found.

poornachandra:
@jwang47 When n/a is displayed for timestamp, does it mean that this queue consumer did not process any events at all? In such a case it is okay to ignore the consumer while calculating the min time across all consumers?

jwang47:
Right, n/a means the queue has no processed events. In this case, we are effectively ignoring the consumer for calculating min time.

PR #3627, opened 2015-08-17
Correct the spelling of "ZooKeeper".

awholegunch:
In most cases, it should be spelt "ZooKeeper". This corrects it in both the documentation and the application template. It adds Apache ZooKeeper as trademark of the Apache Foundation.

Fixes https://issues.cask.co/browse/CDAP-3375


gokulavasan:
LGTM :+1: Just curious, in what cases should it not be spelt as "ZooKeeper"

awholegunch:
I guess the only other cases are where it should be all lower-case ("zookeeper"), which is usually code or a property name. There are a number of ``Constants.Zookeeper.*``...

awholegunch:
There are classes called ``Zookeeper``, but I have left them unchanged.

gokulavasan:
LGTM :+1: Can you squash all the commits to a single one before merging? Or do you think they all warrant a separate commit?

PR #3628, opened 2015-08-17
Upgrade tool output example and --force option.

awholegunch:
Fix for https://issues.cask.co/browse/CDAP-3403

Adds to the description of the upgrade tool an example output (created with a parsed literal block so that it will use the current CDAP version string) and explains use of the ``--force`` option.

Building at : [Installation#upgrading-an-existing-version](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT-feature-3.2_3403_Installation/en/admin-manual/installation/installation.html#upgrading-an-existing-version)


bdmogal:
:+1:

PR #3629, opened 2015-08-18
[CDAP-3142] Remove blue alerts from workflow

elfenheart:
- Remove alerts from no runs workflow (cursor already showing disabled)


ajainarayanan:
:+1: LGTM

PR #3630, opened 2015-08-18
CDAP-3065 add artifact methods to TestBase for unit tests

albertshau:
Adding methods to add an application artifact or a plugin artifact
to TestManager and its implementations. Also adding a method to
deploy an application from an artifact.

### DIFF
--- a/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
+++ b/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
@@ -49,6 +49,7 @@ import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.io.Files;
+import com.google.common.io.InputSupplier;
 import com.google.common.io.Resources;
 import com.google.common.reflect.TypeToken;
 import com.google.gson.Gson;
@@ -60,6 +61,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.JarURLConnection;
 import java.net.URL;
 import java.sql.Connection;
@@ -187,13 +189,15 @@ public class IntegrationTestManager implements TestManager {
 
   @Override
   public void addAppArtifact(Id.Artifact artifactId, Class<?> appClass) throws Exception {
-    Location appJar = AppJarHelper.createDeploymentJar(locationFactory, appClass, new Manifest());
-    File destination =
-      new File(tmpFolder, String.format("%s-%s.jar", artifactId.getName(), artifactId.getVersion().getVersion()));
-    Files.copy(Locations.newInputSupplier(appJar), destination);
-    appJar.delete();
+    final Location appJar = AppJarHelper.createDeploymentJar(locationFactory, appClass, new Manifest());
 
-    artifactClient.add(artifactId, null, Files.newInputStreamSupplier(destination));
+    artifactClient.add(artifactId, null, new InputSupplier<InputStream>() {
+      @Override
+      public InputStream getInput() throws IOException {
+        return appJar.getInputStream();
+      }
+    });
+    appJar.delete();
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2551

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
```diff
@@ -162,6 +181,68 @@ public void addTemplatePluginJson(Id.ApplicationTemplate templateId, String file
...
+  public void addArtifact(Id.Artifact artifactId, File artifactFile) throws Exception {
+    artifactClient.add(artifactId, null, Files.newInputStreamSupplier(artifactFile));
+  }
+
+  @Override
+  public void addAppArtifact(Id.Artifact artifactId, Class<?> appClass) throws Exception {
+    Location appJar = AppJarHelper.createDeploymentJar(locationFactory, appClass, new Manifest());
+    File destination =
+      new File(tmpFolder, String.format("%s-%s.jar", artifactId.getName(), artifactId.getVersion().getVersion()));
+    Files.copy(Locations.newInputSupplier(appJar), destination);
```
gokulavasan: We create the jar once and then copy it again for formatting the name of the Jar to be in the format ``%s-%s.jar``?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
```diff
@@ -162,6 +181,68 @@ public void addTemplatePluginJson(Id.ApplicationTemplate templateId, String file
...
+    appJar.delete();
+
+    artifactClient.add(artifactId, null, Files.newInputStreamSupplier(destination));
+  }
+
+  @Override
+  public void addPluginArtifact(Id.Artifact artifactId, Id.Artifact parent,
```
gokulavasan: It might be helpful to add comments for each of these methods to specify what method is used when. It is hard to figure it out from the method arguments.

cdap-unit-test/src/main/java/co/cask/cdap/test/UnitTestManager.java
```diff
@@ -156,13 +168,25 @@ public ApplicationManager deployApplication(Id.Namespace namespace, Class<? exte
...
                                                                applicationClz, appConfig, bundleEmbeddedJars);
-      return appManagerFactory.create(Id.Application.from(namespace, appSpec.getName()), deployedJar, appSpec);
+      return appManagerFactory.create(Id.Application.from(namespace, appSpec.getName()), deployedJar);
     } catch (Exception e) {
       throw Throwables.propagate(e);
     }
   }
 
   @Override
+  public ApplicationManager deployApplication(Id.Application appId,
```
gokulavasan: Just curious, is there any way we can reduce the code duplication between this class and IntegrationTestManager?

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
```diff
@@ -183,26 +185,32 @@ public void testServiceManager() throws Exception {
...
-    ApplicationManager appManager;
-    if (object != null) {
-      streamName = object.getStreamName();
-      datasetName = object.getTableName();
-      appManager = deployApplication(ConfigTestApp.class, object);
-    } else {
-      appManager = deployApplication(ConfigTestApp.class);
-    }
+  @Test
+  public void testAppFromArtifact() throws Exception {
```
gokulavasan: Will the current methods to deployApp in TestFramework still remain? If yes, what will be the version of the artifacts created from it? Is the plan to have both the current and new methods exist together?

gokulavasan:
Few comments/questions. Rest LGTM :+1:

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
albertshau: thats a good point, this was a blind copy-paste. don't need to copy at all

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
albertshau: will do

cdap-unit-test/src/test/java/co/cask/cdap/test/app/TestFrameworkTestRun.java
albertshau: the current methods will still remain to maintain compatibility and because they are valid restful api calls as well. They deploy a version that is derived from the current time, but it is transparent to the caller.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
albertshau: actually these are described by the javadocs in the interface TestManager already so won't copy them here.

gokulavasan:
LGTM :+1:

cdap-unit-test/src/main/java/co/cask/cdap/test/ConfigurableTestBase.java
```diff
@@ -525,6 +540,25 @@ protected static void addTemplatePluginJson(Id.ApplicationTemplate templateId, S
     getTestManager().addTemplatePluginJson(templateId, fileName, type, name, description, className, fields);
   }
 
+  protected static void addArtifact(Id.Artifact artifactId, File artifactFile) throws Exception {
```
bdmogal: this class is Deprecated (as of 3.1, I think). Should we add new methods to it?

gokulavasan: @bdmogal Will address this in a new PR. Merging PR now.

bdmogal: Sure. Please create a Jira, because this class is intended to be removed in 3.3

PR #3633, opened 2015-08-18
CDAP-3427: Fix cdap-master packaging

jwang47:
https://issues.cask.co/browse/CDAP-3427

Built cdap-master and verified that the expected classes are in the jar.


sreevatsanraman:
changes LGTM. Merge if it is working on a cluster.

jwang47:
Tested and working on cluster.

PR #3634, opened 2015-08-18
Set Twill log level to WARN to log master startup exceptions

poornachandra:


gokulavasan:
LGTM :+1:

poornachandra:
No need to wait for unit-tests since this does not affect them.

PR #3635, opened 2015-08-18
CDAP-3430: Move StreamEvent back to co.cask.cdap.api.flow.flowlet package

jwang47:
https://issues.cask.co/browse/CDAP-3430
http://builds.cask.co/browse/CDAP-DUT2553

Shouldn't move public API classes before deprecating them.


bdmogal:
What was the original intention behind the move? If the right place is in ``co.cask.cdap.api.stream``, then should we keep it there, and deprecate the old one in ``co.cask.cdap.api.flow.flowlet``, as opposed to moving it back?

jwang47:
I'm not sure we actually wanted to move the class to the new package yet. It was moved because I read a TODO that suggested that. Simply going to revert for now.

bdmogal:
LGTM. Please do an integration test run (standalone) before merging.

jwang47:
Got integration tests passing. Will merge after successful build.

PR #3637, opened 2015-08-18
CDAP-3282 api to update an existing application

albertshau:
An application's config can be updated, as well as the artifact
version it uses. Changing artifact name is invalid. If no config
is given, the existing config will be used. Otherwise, the
existing config is completely overwritten by the requested config.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2556

gokulavasan:
LGTM :+1:

PR #3640, opened 2015-08-19
Feature/ui styling dashboard pixel pushing

astewes:
*  Fixes bottom border issue affecting certain slanted tabs
*  Adds helper class to better target these slanted tabs

tabs


ajainarayanan:
@astewes This PR doesn't fix other slanted tabs border issue. I feel instead of fixing a specific case (border bottom) which happened based on something that we changed we should fix the original change that caused this. Do we know what caused the extra border bottom for non-active tabbed tabs?

ajainarayanan:
:+1: LGTM

PR #3641, opened 2015-08-19
Improve Documentation Search Error Response

awholegunch:
Fixes https://issues.cask.co/browse/CDAP-3438

Removes commented-out HTML and Javascript code that was not being used, and that created a duplicate message that wasn't being used.

Corrected the Javascript that produced a message that was being used when a search failed to produce any results.

Example of current message: [Search with Error Message](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT/en/search.html?q=common-&check_keywords=yes&area=default)
Example of revised message: [Search with Error Message](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT-feature-3.2_3438_Search_Error/en/search.html?q=common-&check_keywords=yes&area=default)

Currently building...

### DIFF
--- a/cdap-docs/_common/_themes/cdap/search.html
+++ b/cdap-docs/_common/_themes/cdap/search.html
@@ -49,9 +49,11 @@
     <span id="search-progress" style="padding-left: 10px"></span>
   </form>
   <p>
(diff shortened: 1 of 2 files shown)


bdmogal:
:+1:

PR #3642, opened 2015-08-19
Feature/ui styling ns dropdown scroll

astewes:
*  Removes 18px bottom margin from global header (unnecessary since the header is fixed and main.container is positioned around it)
*  Disables namespace dropdown animation, which was causing the dropdown scroll to stop working intermittently

cluster name: ns-dd-scroll

### DIFF
--- a/cdap-ui/app/directives/breadcrumbs/breadcrumbs.less
+++ b/cdap-ui/app/directives/breadcrumbs/breadcrumbs.less
@@ -84,11 +84,6 @@ my-breadcrumbs, my-breadcrumb {
       .box-shadow(0 6px 12px rgba(0, 0, 0, 0.175));
       overflow-x: hidden;
       overflow-y: scroll;
-      &.am-flip-x {
-        .transition(none);
-        -webkit-animation-duration: 0s;
-        animation-duration: 0s;
-      }
       @media (max-width: @screen-md-max) {
         position: absolute;
       }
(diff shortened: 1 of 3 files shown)


ajainarayanan:
@astewes the right place for removing the animation would be here https://github.com/caskdata/cdap/blob/feature/ui-styling-ns-dropdown-scroll/cdap-ui/app/directives/breadcrumbs/breadcrumbs-ctrl.js#L22 instead of styling around it.

cdap-ui/app/directives/breadcrumbs/breadcrumbs.less
```diff
@@ -114,6 +71,52 @@ my-breadcrumbs, my-breadcrumb {
...
+          width: 13px;
+        }
+        .active {
+          color: #c4c4c4;
+          pointer-events: none;
+        }
+      }
+    }
+
     .breadcrumb {
```
ajainarayanan: This might be out of context of this PR, but .breadcrumb style shouldn't it be under ```breadcrumb``` directive?

ajainarayanan:
One minor comment. Otherwise :+1: LGTM

astewes:
The .breadcrumbs class is assigned to the <ol> for the breadcrumbs list. But to avoid confusion, this class should be renamed to something like .breadcrumb-list, to avoid any potential confusion with the breadcrumb directive.

astewes:
HTML markup is the same in all breadcrumb instances, and the .breadcrumb class is part of ncy-breadcrumb, and therefore shouldn't be changed. Will keep the selector as is for now and perhaps revisit this in the near future.

PR #3643, opened 2015-08-19
Update the copyright header for CDAP from 2014 to 2014-2015

henry-cask:
This PR Contains large number of files but the goal is to change copyright header year.

The copyright information in the header need to be updated to reflect 2015.

From: Copyright © 2014 Cask Data, Inc.
To: Copyright © 2014-2015 Cask Data, Inc.

### DIFF
--- a/cdap-api-common/pom.xml
+++ b/cdap-api-common/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-  Copyright © 2014 Cask Data, Inc.
+  Copyright © 2014-2015 Cask Data, Inc.
 
   Licensed under the Apache License, Version 2.0 (the "License"); you may not
   use this file except in compliance with the License. You may obtain a copy of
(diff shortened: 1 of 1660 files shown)


henry-cask:
Build:
http://builds.cask.co/browse/CDAP-DUT2565-2

sagarkapare:
@henry-cask I think we only update the copyright if the file is modified? Are these the files we modified in 2015 and missed to update the copyright?

henry-cask:
Copyright only applies to whatever year posted in the header regardless been modified or not.

So for the files not modified, the Copyright end by end of year 2014

bdmogal:
Presuming this is necessary, and build passes, LGTM.

henry-cask:
Thanks, merging ...

awholegunch:
I missed this PR and issue. This is unnecessary and should not have been done. Especially since changing file hashes are used to flag real changes in the documentation, and this sort of change breaks the documentation build.

Copyright does not end at the end of a year. The year signifies when the copyright starts. In the case of corporations in the United States, the length of term is 120 years. (See http://www.copyright.gov/title17/92chap3.html#302).

Please revert this merge.

henry-cask:
Ah, then I think we need change some. I remember seeing some has 2014-2015. Ok reverting this change

PR #3644, opened 2015-08-19
Features/ui adapter schedules

ajainarayanan:
- Adds schedules view to ETLBatch adapters
- Adds schedules-view and cron-schedule-view directives for re-usability
- Fixes workflows & adapter schedule views to use the same directive (share component).


elfenheart:
:+1: LGTM

PR #3646, opened 2015-08-20
Feature/cdap-3372 workflow compare

rohanahata:
The other 2 endpoints from the WorkflowStatsSla. 
https://issues.cask.co/browse/CDAP-3370
http://builds.cask.co/browse/CDAP-DUT2570

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
@@ -501,4 +501,25 @@ public interface Store {
    */
   WorkflowStatistics getWorkflowStatistics(Id.Workflow workflowId, long startTime,
                                            long endTime, List<Double> percentiles);
+
+  /**
+   * Returns the record that represents the run of a workflow.
+   *
+   * @param workflowId The Workflow whose run needs to be queried
+   * @param runId RunId of the workflow run
+   * @return A workflow run record corresponding to the runId
+   */
+  WorkflowDataset.WorkflowRunRecord getWorkflowRun(Id.Workflow workflowId, String runId);
+
+  /**
+   * Get a list of workflow runs that are spaced apart by time interval in both directions from the run id provided.
+   *
+   * @param workflow The workflow whose statistics need to be obtained
+   * @param runId The run id of the workflow
+   * @param limit The number of the records that the user wants to compare against on either side of the run
+   * @param timeInterval The timeInterval with which the user wants to space out the runs
+   * @return Map of runId of Workflow to DetailedStatistics of the run
+   */
+  Collection<WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(Id.Workflow workflow, String runId,
+                                                                       int limit, long timeInterval);
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
@@ -16,12 +16,24 @@
 
 package co.cask.cdap.gateway.handlers;
 
+import co.cask.cdap.api.dataset.lib.cube.AggregationFunction;
+import co.cask.cdap.api.metrics.MetricDataQuery;
+import co.cask.cdap.api.metrics.MetricSearchQuery;
+import co.cask.cdap.api.metrics.MetricStore;
+import co.cask.cdap.api.metrics.MetricTimeSeries;
+import co.cask.cdap.api.metrics.TagValue;
+import co.cask.cdap.app.mapreduce.MRJobInfoFetcher;
 import co.cask.cdap.app.store.Store;
 import co.cask.cdap.common.BadRequestException;
+import co.cask.cdap.common.NotFoundException;
+import co.cask.cdap.common.app.RunIds;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.utils.TimeMathParser;
+import co.cask.cdap.internal.app.store.WorkflowDataset;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.WorkflowStatistics;
+import co.cask.cdap.proto.WorkflowStatsComparison;
 import co.cask.http.AbstractHttpHandler;
 import co.cask.http.HttpResponder;
 import com.google.inject.Inject;
@@ -31,7 +43,14 @@ import org.jboss.netty.handler.codec.http.HttpResponseStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -46,21 +65,37 @@ public class WorkflowStatsSLAHttpHandler extends AbstractHttpHandler {
 
   private static final Logger LOG = LoggerFactory.getLogger(WorkflowStatsSLAHttpHandler.class);
   private final Store store;
+  private final MRJobInfoFetcher mrJobInfoFetcher;
+  private final MetricStore metricStore;
 
   @Inject
-  WorkflowStatsSLAHttpHandler(Store store) {
+  WorkflowStatsSLAHttpHandler(Store store, MRJobInfoFetcher mrJobInfoFetcher, MetricStore metricStore) {
     this.store = store;
+    this.mrJobInfoFetcher = mrJobInfoFetcher;
+    this.metricStore = metricStore;
   }
 
+  /**
+   * Returns the statistics for a given workflow.
+   *
+   * @param request The request
+   * @param responder The responder
+   * @param namespaceId The namespace the application is in
+   * @param appId The application the workflow is in
+   * @param workflowId The workflow that needs to have it stats shown
+   * @param start The start time of the range
+   * @param end The end time of the range
+   * @param percentiles The list of percentile values on which visibility is needed
+   */
   @GET
   @Path("apps/{app-id}/workflows/{workflow-id}/statistics")
   public void workflowStats(HttpRequest request, HttpResponder responder,
                             @PathParam("namespace-id") String namespaceId,
                             @PathParam("app-id") String appId,
                             @PathParam("workflow-id") String workflowId,
-                            @QueryParam("start") String start,
-                            @QueryParam("end") String end,
-                            @QueryParam("percentile") List<Double> percentiles) throws Exception {
+                            @QueryParam("start") @DefaultValue("now-1d") String start,
+                            @QueryParam("end") @DefaultValue("now") String end,
+                            @QueryParam("percentile") @DefaultValue("90.0") List<Double> percentiles) throws Exception {
     long startTime = TimeMathParser.parseTimeInSeconds(start);
     long endTime = TimeMathParser.parseTimeInSeconds(end);
 
@@ -88,5 +123,222 @@ public class WorkflowStatsSLAHttpHandler extends AbstractHttpHandler {
       return;
     }
     responder.sendJson(HttpResponseStatus.OK, workflowStatistics);
+  }
+
+  /**
+   * The endpoint returns a list of workflow metrics based on the workflow run and a surrounding number of runs
+   * of the workflow that are spaced apart by a time interval from each other.
+   *
+   * @param request The request
+   * @param responder The responder
+   * @param namespaceId The namespace the application is in
+   * @param appId The application the workflow is in
+   * @param workflowId The workflow that needs to have it stats shown
+   * @param runId The run id of the Workflow that the user wants to see
+   * @param limit The number of the records that the user wants to compare against on either side of the run
+   * @param interval The timeInterval with which the user wants to space out the runs
+   */
+  @GET
+  @Path("apps/{app-id}/workflows/{workflow-id}/runs/{run-id}/statistics")
+  public void workflowRunDetail(HttpRequest request, HttpResponder responder,
+                                @PathParam("namespace-id") String namespaceId,
+                                @PathParam("app-id") String appId,
+                                @PathParam("workflow-id") String workflowId,
+                                @PathParam("run-id") String runId,
+                                @QueryParam("limit") int limit,
+                                @QueryParam("interval") String interval) throws Exception {
+    if (limit < 0) {
+      throw new BadRequestException("Limit has to be greater than or equal to 0. Entered value was : " + limit);
+    }
+
+    long timeInterval = TimeMathParser.resolutionInSeconds(interval);
+    Id.Workflow workflow = Id.Workflow.from(Id.Namespace.from(namespaceId), appId, workflowId);
+    Collection<WorkflowDataset.WorkflowRunRecord> workflowRunRecords =
+      store.retrieveSpacedRecords(workflow, runId, limit, timeInterval);
+
+    List<WorkflowRunMetrics> workflowRunMetricsList = new ArrayList<>();
+    Map<String, Long> startTimes = new HashMap<>();
+    for (WorkflowDataset.WorkflowRunRecord workflowRunRecord : workflowRunRecords) {
+      workflowRunMetricsList.add(getDetailedRecord(workflow, workflowRunRecord.getWorkflowRunId()));
+      startTimes.put(workflowRunRecord.getWorkflowRunId(),
+                     RunIds.getTime(RunIds.fromString(workflowRunRecord.getWorkflowRunId()), TimeUnit.SECONDS));
+    }
+
+    Collection<WorkflowStatsComparison.ProgramNodes> formattedStatisticsMap = format(workflowRunMetricsList);
+    responder.sendJson(HttpResponseStatus.OK, new WorkflowStatsComparison(startTimes, formattedStatisticsMap));
+  }
+
+  /**
+   * Compare the metrics of 2 runs of a workflow
+   *
+   * @param request The request
+   * @param responder The responder
+   * @param namespaceId The namespace the application is in
+   * @param appId The application the workflow is in
+   * @param workflowId The workflow that needs to have it stats shown
+   * @param runId The run id of the Workflow that the user wants to see
+   * @param otherRunId The other run id of the same workflow that the user wants to compare against
+   */
+  @GET
+  @Path("apps/{app-id}/workflows/{workflow-id}/runs/run-id/{run-id}/compare")
+  public void compare(HttpRequest request, HttpResponder responder,
+                      @PathParam("namespace-id") String namespaceId,
+                      @PathParam("app-id") String appId,
+                      @PathParam("workflow-id") String workflowId,
+                      @PathParam("run-id") String runId,
+                      @QueryParam("other-run-id") String otherRunId) throws Exception {
+    Id.Workflow workflow = Id.Workflow.from(Id.Namespace.from(namespaceId), appId, workflowId);
+    WorkflowRunMetrics detailedStatistics = getDetailedRecord(workflow, runId);
+    WorkflowRunMetrics otherDetailedStatistics = getDetailedRecord(workflow, otherRunId);
+    if (detailedStatistics == null) {
+      throw new NotFoundException("The run-id provided was not found : " + runId);
+    }
+    if (otherDetailedStatistics == null) {
+      throw new NotFoundException("The other run-id provided was not found : " + otherRunId);
+    }
+
+    List<WorkflowRunMetrics> workflowRunMetricsList = new ArrayList<>();
+    workflowRunMetricsList.add(detailedStatistics);
+    workflowRunMetricsList.add(otherDetailedStatistics);
+    responder.sendJson(HttpResponseStatus.OK, format(workflowRunMetricsList));
+  }
+
+  private Collection<WorkflowStatsComparison.ProgramNodes> format(List<WorkflowRunMetrics> workflowRunMetricsList) {
+    Map<String, WorkflowStatsComparison.ProgramNodes> programLevelDetails = new HashMap<>();
+    for (WorkflowRunMetrics workflowRunMetrics : workflowRunMetricsList) {
+      for (ProgramMetrics programMetrics : workflowRunMetrics.getProgramMetricsList()) {
+        String programName = programMetrics.getProgramName();
+        if (programLevelDetails.get(programName) == null) {
+          WorkflowStatsComparison.ProgramNodes programNodes = new WorkflowStatsComparison.ProgramNodes(
+            programName, programMetrics.getProgramType(),
+            new ArrayList<WorkflowStatsComparison.ProgramNodes.WorkflowProgramDetails>());
+          programLevelDetails.put(programName, programNodes);
+        }
+        programLevelDetails.get(programName).addWorkflowDetails(
+          workflowRunMetrics.getWorkflowRunId(), programMetrics.getProgramRunId(),
+          programMetrics.getProgramStartTime(), programMetrics.getMetrics());
+      }
+    }
+    return programLevelDetails.values();
+  }
+
+  /**
+   * Returns the detailed Record for the Workflow
+   *
+   * @param workflowId Workflow that needs to get its detailed record
+   * @param runId Run Id of the workflow
+   * @return Return the Workflow Run Metrics
+   */
+  @Nullable
+  private WorkflowRunMetrics getDetailedRecord(Id.Workflow workflowId, String runId) throws Exception {
+    WorkflowDataset.WorkflowRunRecord workflowRunRecord = store.getWorkflowRun(workflowId, runId);
+    if (workflowRunRecord == null) {
+      return null;
+    }
+    List<WorkflowDataset.ProgramRun> programRuns = workflowRunRecord.getProgramRuns();
+    List<ProgramMetrics> programMetricsList = new ArrayList<>();
+    for (WorkflowDataset.ProgramRun programRun : programRuns) {
+      Map<String, Long> programMap = new HashMap<>();
+      String programName = programRun.getName();
+      ProgramType programType = programRun.getProgramType();
+      Id.Program program = Id.Program.from(workflowId.getNamespaceId(), workflowId.getApplicationId(),
+                                           programType, programName);
+      String programRunId = programRun.getRunId();
+      if (programType == ProgramType.MAPREDUCE) {
+        programMap = getMapreduceDetails(program, programRunId);
+      } else if (programType == ProgramType.SPARK) {
+        programMap = getSparkDetails(program, programRunId);
+      }
+      programMap.put("timeTaken", programRun.getTimeTaken());
+      long programStartTime = RunIds.getTime(RunIds.fromString(programRunId), TimeUnit.SECONDS);
+      programMetricsList.add(new ProgramMetrics(programName, programType, programRunId, programStartTime, programMap));
+    }
+    return new WorkflowRunMetrics(runId, programMetricsList);
+  }
+
+  private Map<String, Long> getMapreduceDetails(Id.Program mapreduceProgram, String runId) throws Exception {
+    Id.Run mrRun = new Id.Run(mapreduceProgram, runId);
+    return mrJobInfoFetcher.getMRJobInfo(mrRun).getCounters();
+  }
+
+  private Map<String, Long> getSparkDetails(Id.Program sparkProgram, String runId) throws Exception {
+    Map<String, String> context = new HashMap<>();
+    context.put(Constants.Metrics.Tag.NAMESPACE, sparkProgram.getNamespaceId());
+    context.put(Constants.Metrics.Tag.APP, sparkProgram.getApplicationId());
+    context.put(Constants.Metrics.Tag.SPARK, sparkProgram.getId());
+    context.put(Constants.Metrics.Tag.RUN_ID, runId);
+
+    List<TagValue> tags = new ArrayList<>();
+    for (Map.Entry<String, String> entry : context.entrySet()) {
+      tags.add(new TagValue(entry.getKey(), entry.getValue()));
+    }
+    MetricSearchQuery metricSearchQuery = new MetricSearchQuery(0, 0, Integer.MAX_VALUE, tags);
+    Collection<String> metricNames = metricStore.findMetricNames(metricSearchQuery);
+    Map<String, Long> overallResult = new HashMap<>();
+    for (String metricName : metricNames) {
+      Collection<MetricTimeSeries> resultPerQuery = metricStore.query(
+        new MetricDataQuery(0, 0, Integer.MAX_VALUE, metricName, AggregationFunction.SUM,
+                            context, new ArrayList<String>()));
+
+      for (MetricTimeSeries metricTimeSeries : resultPerQuery) {
+        overallResult.put(metricTimeSeries.getMetricName(), metricTimeSeries.getTimeValues().get(0).getValue());
+      }
+    }
+    return overallResult;
+  }
+
+  private static class ProgramMetrics {
+    private final String programName;
+    private final ProgramType programType;
+    private final String programRunId;
+    private final long programStartTime;
+    private final Map<String, Long> metrics;
+
+    private ProgramMetrics(String programName, ProgramType programType, String programRunId,
+                          long programStartTime, Map<String, Long> metrics) {
+      this.programName = programName;
+      this.programType = programType;
+      this.programRunId = programRunId;
+      this.programStartTime = programStartTime;
+      this.metrics = metrics;
+    }
+
+    public String getProgramName() {
+      return programName;
+    }
+
+    public ProgramType getProgramType() {
+      return programType;
+    }
+
+    public Map<String, Long> getMetrics() {
+      return metrics;
+    }
+
+    public long getProgramStartTime() {
+      return programStartTime;
+    }
+
+    public String getProgramRunId() {
+      return programRunId;
+    }
+  }
+
+  private static class WorkflowRunMetrics {
+    private final String workflowRunId;
+    private final List<ProgramMetrics> programMetricsList;
+
+    private WorkflowRunMetrics(String workflowRunId, List<ProgramMetrics> programMetricsList) {
+      this.workflowRunId = workflowRunId;
+      this.programMetricsList = programMetricsList;
+    }
+
+    public String getWorkflowRunId() {
+      return workflowRunId;
+    }
+
+    public List<ProgramMetrics> getProgramMetricsList() {
+      return programMetricsList;
+    }
   }
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
@@ -331,6 +331,31 @@ public class DefaultStore implements Store {
   }
 
   @Override
+  public WorkflowDataset.WorkflowRunRecord getWorkflowRun(final Id.Workflow workflowId, final String runId) {
+    return txnlWorkflow.executeUnchecked(new TransactionExecutor.Function
+      <WorkflowStatsDataset, WorkflowDataset.WorkflowRunRecord>() {
+      @Override
+      public WorkflowDataset.WorkflowRunRecord apply(WorkflowStatsDataset dataset) throws Exception {
+        return dataset.workflowDataset.getRecord(workflowId, runId);
+      }
+    });
+  }
+
+  @Override
+  public Collection<WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(final Id.Workflow workflow,
+                                                                             final String runId,
+                                                                             final int limit,
+                                                                             final long timeInterval) {
+    return txnlWorkflow.executeUnchecked(new TransactionExecutor.Function
+      <WorkflowStatsDataset, Collection<WorkflowDataset.WorkflowRunRecord>>() {
+      @Override
+      public Collection<WorkflowDataset.WorkflowRunRecord> apply(WorkflowStatsDataset dataset) throws Exception {
+        return dataset.workflowDataset.getDetailsOfRange(workflow, runId, limit, timeInterval);
+      }
+    });
+  }
+
+  @Override
   public List<RunRecordMeta> getRuns(final Id.Program id, final ProgramRunStatus status,
                                      final long startTime, final long endTime, final int limit, final String adapter) {
     return txnl.executeUnchecked(new TransactionExecutor.Function<AppMds, List<RunRecordMeta>>() {
@@ -1216,9 +1241,9 @@ public class DefaultStore implements Store {
   }
 
   private static ApplicationSpecification replaceFlowletInAppSpec(final ApplicationSpecification appSpec,
-                                                           final Id.Program id,
-                                                           final FlowSpecification flowSpec,
-                                                           final FlowletDefinition adjustedFlowletDef) {
+                                                                  final Id.Program id,
+                                                                  final FlowSpecification flowSpec,
+                                                                  final FlowletDefinition adjustedFlowletDef) {
     // as app spec is immutable we have to do this trick
     return replaceFlowInAppSpec(appSpec, id, new FlowSpecificationWithChangedFlowlets(flowSpec, adjustedFlowletDef));
   }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
@@ -22,6 +22,7 @@ import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Scan;
 import co.cask.cdap.api.dataset.table.Scanner;
 import co.cask.cdap.api.dataset.table.Table;
+import co.cask.cdap.common.app.RunIds;
 import co.cask.cdap.data2.dataset2.lib.table.MDSKey;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.PercentileInformation;
@@ -30,8 +31,10 @@ import co.cask.cdap.proto.WorkflowStatistics;
 import com.google.common.primitives.Longs;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+import org.apache.twill.api.RunId;
 
 import java.lang.reflect.Type;
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -39,6 +42,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 /**
@@ -81,9 +85,8 @@ public class WorkflowDataset extends AbstractDataset {
    * @param timeRangeStart Start of the time range that the scan should begin from
    * @param timeRangeEnd End of the time range that the scan should end at
    * @return List of WorkflowRunRecords
-   * @throws Exception
    */
-  public List<WorkflowRunRecord> scan(Id.Workflow id, long timeRangeStart, long timeRangeEnd) throws Exception {
+  private List<WorkflowRunRecord> scan(Id.Workflow id, long timeRangeStart, long timeRangeEnd) {
     byte[] startRowKey = new MDSKey.Builder().add(id.getApplication().getNamespaceId()).add(id.getApplicationId()).
       add(id.getId()).add(timeRangeStart).build().getKey();
     byte[] endRowKey = new MDSKey.Builder().add(id.getApplication().getNamespaceId()).add(id.getApplicationId()).
@@ -204,10 +207,86 @@ public class WorkflowDataset extends AbstractDataset {
     return programToRunRecord.values();
   }
 
+  @Nullable
+  WorkflowRunRecord getRecord(Id.Workflow id, String pid) {
+    RunId runId = RunIds.fromString(pid);
+    long startTime = RunIds.getTime(runId, TimeUnit.SECONDS);
+    MDSKey mdsKey = new MDSKey.Builder().add(id.getNamespaceId())
+      .add(id.getApplicationId()).add(id.getId()).add(startTime).build();
+    byte[] startRowKey = mdsKey.getKey();
+
+    Row indexRow = table.get(startRowKey);
+    if (indexRow == null) {
+      return null;
+    }
+    Map<byte[], byte[]> columns = indexRow.getColumns();
+    String workflowRunId = Bytes.toString(columns.get(RUNID));
+    long timeTaken = Bytes.toLong(columns.get(TIME_TAKEN));
+
+    List<ProgramRun> actionRunsList = GSON.fromJson(Bytes.toString(columns.get(NODES)), PROGRAM_RUNS_TYPE);
+    return new WorkflowRunRecord(workflowRunId, timeTaken, actionRunsList);
+  }
+
+  Collection<WorkflowRunRecord> getDetailsOfRange(Id.Workflow workflow, String runId, int limit, long timeInterval) {
+    Map<String, WorkflowRunRecord> mainRunRecords = getNeighbors(workflow, RunIds.fromString(runId),
+                                                                 limit, timeInterval);
+    WorkflowRunRecord workflowRunRecord = getRecord(workflow, runId);
+    if (workflowRunRecord != null) {
+      mainRunRecords.put(workflowRunRecord.getWorkflowRunId(), workflowRunRecord);
+    }
+    return mainRunRecords.values();
+  }
+
+  /**
+   * Returns a map of WorkflowRunId to WorkflowRunRecord that are close to the WorkflowRunId provided by the user.
+   *
+   * @param id The workflow
+   * @param runId The runid of the workflow
+   * @param limit The limit on each side of the run that we want to see into
+   * @param timeInterval The time interval that we want the results to be spaced apart
+   * @return A Map of WorkflowRunId to the corresponding Workflow Run Record. A map is used so that duplicates of
+   * the WorkflowRunRecord are not obtained
+   */
+  private Map<String, WorkflowRunRecord> getNeighbors(Id.Workflow id, RunId runId, int limit, long timeInterval) {
+    long startTime = RunIds.getTime(runId, TimeUnit.SECONDS);
+    Map<String, WorkflowRunRecord> workflowRunRecords = new HashMap<>();
+    int i = -limit;
+    long prevStartTime = startTime - (limit * timeInterval);
+    // The loop iterates across the range that is startTime - (limit * timeInterval) to
+    // startTime + (limit * timeInterval) since we want to capture all runs that started in this range.
+    // Since we want to stop getting the same key, we have the prevStartTime become 1 more than the time at which
+    // the last record was found if the (interval * the count of the loop) is less than the time.
+    while (prevStartTime <= startTime + (limit * timeInterval)) {
+      MDSKey mdsKey = new MDSKey.Builder().add(id.getNamespaceId())
+        .add(id.getApplicationId()).add(id.getId()).add(prevStartTime).build();
+      byte[] startRowKey = mdsKey.getKey();
+      Scan scan = new Scan(startRowKey, null);
+      Scanner scanner = table.scan(scan);
+      Row indexRow = scanner.next();
+      if (indexRow == null) {
+        return workflowRunRecords;
+      }
+      byte[] rowKey = indexRow.getRow();
+      long time = ByteBuffer.wrap(rowKey, rowKey.length - Bytes.SIZEOF_LONG, Bytes.SIZEOF_LONG).getLong();
+      if (!((time >= (startTime - (limit * timeInterval))) && time <= (startTime + (limit * timeInterval)))) {
+        break;
+      }
+      Map<byte[], byte[]> columns = indexRow.getColumns();
+      String workflowRunId = Bytes.toString(columns.get(RUNID));
+      long timeTaken = Bytes.toLong(columns.get(TIME_TAKEN));
+      List<ProgramRun> programRunList = GSON.fromJson(Bytes.toString(columns.get(NODES)), PROGRAM_RUNS_TYPE);
+      workflowRunRecords.put(workflowRunId, new WorkflowRunRecord(workflowRunId, timeTaken, programRunList));
+      prevStartTime = startTime + (i * timeInterval) < time ?
+        time + 1 : startTime + (i * timeInterval);
+      i++;
+    }
+    return workflowRunRecords;
+  }
+
   /**
    * Class to store the name, type and list of runs of the programs across all workflow runs
    */
-  static class ProgramRunDetails {
+  private static final class ProgramRunDetails {
     private final String name;
     private final ProgramType programType;
     private final List<Long> programRunList;
@@ -236,9 +315,9 @@ public class WorkflowDataset extends AbstractDataset {
   }
 
   /**
-   * Internal class to keep track of Workflow Run Records
+   * Class to keep track of Workflow Run Records
    */
-  static class WorkflowRunRecord {
+  public static final class WorkflowRunRecord {
     private final String workflowRunId;
     private final long timeTaken;
     private final List<ProgramRun> programRuns;
@@ -263,9 +342,9 @@ public class WorkflowDataset extends AbstractDataset {
   }
 
   /**
-   * Internal Class for keeping track of programs in a workflow
+   * Class for keeping track of programs in a workflow
    */
-  static class ProgramRun {
+  public static final class ProgramRun {
     private final String runId;
     private final long timeTaken;
     private final ProgramType programType;
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
@@ -17,22 +17,35 @@
 package co.cask.cdap.internal.app.services.http.handlers;
 
 import co.cask.cdap.WorkflowApp;
+import co.cask.cdap.api.metrics.MetricStore;
+import co.cask.cdap.api.metrics.MetricType;
+import co.cask.cdap.api.metrics.MetricValues;
+import co.cask.cdap.app.metrics.MapReduceMetrics;
 import co.cask.cdap.app.store.Store;
 import co.cask.cdap.common.app.RunIds;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.internal.app.services.http.AppFabricTestBase;
 import co.cask.cdap.internal.app.store.DefaultStore;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.PercentileInformation;
 import co.cask.cdap.proto.ProgramRunStatus;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.WorkflowStatistics;
+import co.cask.cdap.proto.WorkflowStatsComparison;
+import com.google.common.collect.ImmutableMap;
 import com.google.gson.reflect.TypeToken;
+import org.apache.hadoop.mapreduce.TaskCounter;
 import org.apache.http.HttpResponse;
 import org.apache.twill.api.RunId;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
 import org.junit.Assert;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -40,10 +53,17 @@ import java.util.concurrent.TimeUnit;
  */
 public class WorkflowStatsSLAHttpHandlerTest extends AppFabricTestBase {
 
+  private static MetricStore metricStore;
+  private static Store store;
+  @BeforeClass
+  public static void beforeClass() throws Throwable {
+    AppFabricTestBase.beforeClass();
+    store = getInjector().getInstance(DefaultStore.class);
+    metricStore = getInjector().getInstance(MetricStore.class);
+  }
+
   @Test
   public void testStatistics() throws Exception {
-
-    Store store = getInjector().getInstance(DefaultStore.class);
 
     deploy(WorkflowApp.class);
     String workflowName = "FunWorkflow";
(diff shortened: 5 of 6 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -501,4 +501,25 @@ void setWorkflowProgramStart(Id.Program programId, String programRunId, String w
    */
   WorkflowStatistics getWorkflowStatistics(Id.Workflow workflowId, long startTime,
                                            long endTime, List<Double> percentiles);
+
+  /**
```
sreevatsanraman: The javadoc doesn't match the function signature. We are returning workflow run record.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+        getDetailedRecord(workflow, workflowRunRecord.getWorkflowRunId());
+      detailedStatisticsMap.put(workflowRunRecord.getWorkflowRunId(), detailedStatistics);
+    }
+    Map<String, Map<String, Map<String, Long>>> formattedStatisticsMap = convert(detailedStatisticsMap);
+
+    responder.sendJson(HttpResponseStatus.OK, formattedStatisticsMap);
+  }
+
+  @GET
+  @Path("apps/{app-id}/workflows/{workflow-id}/compare/run-id/{run-id}")
```
bdmogal: Should be ``apps/{app-id}/workflows/{workflow-id}/run-id/{run-id}/compare``

gsps1: runs/{run-id}

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -501,4 +501,25 @@ void setWorkflowProgramStart(Id.Program programId, String programRunId, String w
...
+   * Get a list of workflow runs that are spaced apart by time interval in both directions from the run id provided.
+   *
+   * @param workflow The workflow whose statistics need to be obtained
+   * @param runId The run id of the workflow
+   * @param count The count of the records that the user wants to compare against
+   * @param timeInterval The timeInterval with which the user wants to space out the runs
+   * @return Map of runId of Workflow to DetailedStatistics of the run
+   */
+  Map<String, WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(Id.Workflow workflow, String runId,
+                                                                int count, long timeInterval);
```
bdmogal: indentation.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -501,4 +501,25 @@ void setWorkflowProgramStart(Id.Program programId, String programRunId, String w
...
                                            long endTime, List<Double> percentiles);
+
+  /**
+   * Returns detailed statistics for a run of a workflow.
+   *
+   * @param workflowId The Workflow whose run needs to be queried
+   * @param runId RunId of the workflow run
+   * @return A detailed report of the statistics
+   */
+  WorkflowDataset.WorkflowRunRecord getWorkflowRun(Id.Workflow workflowId, String runId);
```
bdmogal: I think you could pass in ``Id.Run`` here, but its ok to keep this as-is as well, since you want to enforce that it only supports workflows

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+  public void compare(HttpRequest request, HttpResponder responder,
+                      @PathParam("namespace-id") String namespaceId,
+                      @PathParam("app-id") String appId,
+                      @PathParam("workflow-id") String workflowId,
+                      @PathParam("run-id") String runId,
+                      @QueryParam("other-run-id") String otherRunId) throws Exception {
+    Id.Workflow id = Id.Workflow.from(Id.Namespace.from(namespaceId), appId, workflowId);
```
bdmogal: nit: name the var ``workflowId``

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    WorkflowStatistics.DetailedStatistics detailedStatistics = getDetailedRecord(id, runId);
+    WorkflowStatistics.DetailedStatistics otherDetailedStatistics = getDetailedRecord(id, otherRunId);
+    if (detailedStatistics == null) {
+      responder.sendString(HttpResponseStatus.BAD_REQUEST, "The run-id you provided was not correct.");
```
bdmogal: Better to not use a notion of 'person' in error messages. Also, why is this not correct? Because it was not found? Then let's return ``HttpResponseStatus.NOT_FOUND``. Also include the runid in the error message

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+      return;
+    }
+    if (otherDetailedStatistics == null) {
+      responder.sendString(HttpResponseStatus.BAD_REQUEST, "The other run-id you provided was not correct.");
```
bdmogal: ditto

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+      return;
+    }
+    // TODO: Change format of return such that the results are intertwined.
```
bdmogal: if this is not applicable, remove

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    Map<String, WorkflowStatistics.DetailedStatistics> detailedStatisticsMap = new HashMap<>();
+    detailedStatisticsMap.put(runId, detailedStatistics);
+    detailedStatisticsMap.put(otherRunId, otherDetailedStatistics);
+
+    Map<String, Map<String, Map<String, Long>>> formattedStatisticsMap = convert(detailedStatisticsMap);
+    responder.sendJson(HttpResponseStatus.OK, formattedStatisticsMap);
+  }
+
+  private Map<String, Map<String, Map<String, Long>>> convert(Map<String, WorkflowStatistics.DetailedStatistics>
+                                                                        workflowDetailedStatisticsMap) {
```
bdmogal: indentation

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    // TODO: Change format of return such that the results are intertwined.
+    Map<String, WorkflowStatistics.DetailedStatistics> detailedStatisticsMap = new HashMap<>();
+    detailedStatisticsMap.put(runId, detailedStatistics);
+    detailedStatisticsMap.put(otherRunId, otherDetailedStatistics);
+
+    Map<String, Map<String, Map<String, Long>>> formattedStatisticsMap = convert(detailedStatisticsMap);
+    responder.sendJson(HttpResponseStatus.OK, formattedStatisticsMap);
+  }
+
+  private Map<String, Map<String, Map<String, Long>>> convert(Map<String, WorkflowStatistics.DetailedStatistics>
```
bdmogal: convert what? to what?

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -501,4 +501,25 @@ void setWorkflowProgramStart(Id.Program programId, String programRunId, String w
...
+  /**
+   * Get a list of workflow runs that are spaced apart by time interval in both directions from the run id provided.
+   *
+   * @param workflow The workflow whose statistics need to be obtained
+   * @param runId The run id of the workflow
+   * @param count The count of the records that the user wants to compare against
+   * @param timeInterval The timeInterval with which the user wants to space out the runs
+   * @return Map of runId of Workflow to DetailedStatistics of the run
+   */
+  Map<String, WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(Id.Workflow workflow, String runId,
```
gsps1: should be getWorkflowRuns, also why Map ? can't we just use list or collection ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
     return programToRunRecord.values();
   }
 
+  @Nullable
+  public WorkflowRunRecord getRecord(Id.Workflow id, String runId) {
+    RunId pid = RunIds.fromString(runId);
```
bdmogal: lets call it runId

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+                                @PathParam("namespace-id") String namespaceId,
+                                @PathParam("app-id") String appId,
+                                @PathParam("workflow-id") String workflowId,
+                                @PathParam("run-id") String runId,
+                                @QueryParam("count") int count,
+                                @QueryParam("interval") String interval) throws Exception {
+    long timeInterval = TimeMathParser.resolutionInSeconds(interval);
+    Id.Workflow workflow = Id.Workflow.from(Id.Namespace.from(namespaceId), appId, workflowId);
+    Collection<WorkflowDataset.WorkflowRunRecord> workflowRunRecords =
+      store.retrieveSpacedRecords(workflow, runId, count, timeInterval).values();
```
gsps1: well, here you get a collection, but the signature in ``DefaultStore`` its Map..

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
     return programToRunRecord.values();
   }
 
+  @Nullable
+  public WorkflowRunRecord getRecord(Id.Workflow id, String runId) {
```
bdmogal: javadoc

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    Map<String, WorkflowStatistics.DetailedStatistics> detailedStatisticsMap = new HashMap<>();
+    for (WorkflowDataset.WorkflowRunRecord workflowRunRecord : workflowRunRecords) {
+      WorkflowStatistics.DetailedStatistics detailedStatistics =
+        getDetailedRecord(workflow, workflowRunRecord.getWorkflowRunId());
+      detailedStatisticsMap.put(workflowRunRecord.getWorkflowRunId(), detailedStatistics);
+    }
+    Map<String, Map<String, Map<String, Long>>> formattedStatisticsMap = convert(detailedStatisticsMap);
```
gsps1: if you are formatting, call it ``format(detailedStatisticsMap)``

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
gsps1: try to remove code duplication.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+    RunId pid = RunIds.fromString(runId);
+    long startTime = RunIds.getTime(pid, TimeUnit.SECONDS);
+    MDSKey mdsKey = new MDSKey.Builder().add(id.getNamespaceId())
+      .add(id.getApplicationId()).add(id.getId()).add(startTime).build();
+    byte[] startRowKey = mdsKey.getKey();
+    Scan scan = new Scan(startRowKey, null);
+
+    WorkflowRunRecord workflowRunRecord;
+    Scanner scanner = table.scan(scan);
```
bdmogal: If you know the exact rowKey, just do a get instead of a scan.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+  @Nullable
+  public WorkflowRunRecord getRecord(Id.Workflow id, String runId) {
+    RunId pid = RunIds.fromString(runId);
+    long startTime = RunIds.getTime(pid, TimeUnit.SECONDS);
+    MDSKey mdsKey = new MDSKey.Builder().add(id.getNamespaceId())
+      .add(id.getApplicationId()).add(id.getId()).add(startTime).build();
+    byte[] startRowKey = mdsKey.getKey();
+    Scan scan = new Scan(startRowKey, null);
+
+    WorkflowRunRecord workflowRunRecord;
```
bdmogal: not necessary to declare it first. This should move to line 230

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+      return null;
+    }
+    Map<byte[], byte[]> columns = indexRow.getColumns();
+    String workflowRunId = Bytes.toString(columns.get(RUNID));
+    long timeTaken = Bytes.toLong(columns.get(TIME_TAKEN));
+
+    List<ProgramRun> actionRunsList = GSON.fromJson(Bytes.toString(columns.get(NODES)), PROGRAM_RUNS_TYPE);
+    workflowRunRecord = new WorkflowRunRecord(workflowRunId, timeTaken, actionRunsList);
+
+    if (!runId.equals(workflowRunRecord.getWorkflowRunId())) {
```
bdmogal: This if-else is not necessary

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+    Map<String, WorkflowRunRecord> mainRunRecords = getNeighbors(workflow, runId, count, timeInterval);
+    WorkflowRunRecord workflowRunRecord = getRecord(workflow, runId);
+    if (workflowRunRecord != null) {
+      mainRunRecords.put(workflowRunRecord.getWorkflowRunId(), workflowRunRecord);
+    }
+    return mainRunRecords;
+  }
+
+  private Map<String, WorkflowRunRecord> getNeighbors(Id.Workflow id, String runId, int count, long timeInterval) {
+    RunId pid = RunIds.fromString(runId);
```
bdmogal: runId

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,131 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    context.put(Constants.Metrics.Tag.NAMESPACE, sparkProgram.getNamespaceId());
+    context.put(Constants.Metrics.Tag.APP, sparkProgram.getApplicationId());
+    context.put(Constants.Metrics.Tag.SPARK, sparkProgram.getId());
+    context.put(Constants.Metrics.Tag.RUN_ID, runId);
+
+    List<TagValue> tags = new ArrayList<>();
+    for (Map.Entry<String, String> entry : context.entrySet()) {
+      tags.add(new TagValue(entry.getKey(), entry.getValue()));
+    }
+    MetricSearchQuery metricSearchQuery = new MetricSearchQuery(0, Integer.MAX_VALUE, -1, tags);
```
gsps1: ``new MetricSearchQuery(0, 0, Integer.MAX_VALUE, tags)``

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
bdmogal: hmmm, actually, why not pass in a RunId object instead of a string?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,78 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+      Scanner scanner = table.scan(scan);
+      Row indexRow = scanner.next();
+      if (indexRow == null) {
+        return workflowRunRecords;
+      }
+      mdsKey = new MDSKey(indexRow.getRow());
+      MDSKey.Splitter splitter = mdsKey.split();
+      splitter.skipString();
+      splitter.skipString();
+      splitter.skipString();
```
gsps1: hmm.. why do you do this ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+  private Map<String, WorkflowRunRecord> getNeighbors(Id.Workflow id, String runId, int count, long timeInterval) {
+    RunId pid = RunIds.fromString(runId);
+    long startTime = RunIds.getTime(pid, TimeUnit.SECONDS);
+    Map<String, WorkflowRunRecord> workflowRunRecords = new HashMap<>();
+    for (int i = (-1 * count); i <= count; i++) {
+      long prevStartTime = startTime + (i * timeInterval);
+      MDSKey mdsKey = new MDSKey.Builder().add(id.getNamespaceId())
+        .add(id.getApplicationId()).add(id.getId()).add(prevStartTime).build();
+      byte[] startRowKey = mdsKey.getKey();
+      Scan scan = new Scan(startRowKey, null);
```
bdmogal: use a get instead of a scan?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -238,7 +313,7 @@ public ProgramType getProgramType() {
   /**
    * Internal class to keep track of Workflow Run Records
    */
-  static class WorkflowRunRecord {
+  public static class WorkflowRunRecord {
```
gsps1: internal class, but public ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+      mainRunRecords.put(workflowRunRecord.getWorkflowRunId(), workflowRunRecord);
+    }
+    return mainRunRecords;
+  }
+
+  private Map<String, WorkflowRunRecord> getNeighbors(Id.Workflow id, String runId, int count, long timeInterval) {
+    RunId pid = RunIds.fromString(runId);
+    long startTime = RunIds.getTime(pid, TimeUnit.SECONDS);
+    Map<String, WorkflowRunRecord> workflowRunRecords = new HashMap<>();
+    for (int i = (-1 * count); i <= count; i++) {
```
bdmogal: Should really explain the logic of this for loop in a detailed comment

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+  Map<String, WorkflowRunRecord> getDetailsOfRange(Id.Workflow workflow, String runId, int count, long timeInterval) {
+    Map<String, WorkflowRunRecord> mainRunRecords = getNeighbors(workflow, runId, count, timeInterval);
+    WorkflowRunRecord workflowRunRecord = getRecord(workflow, runId);
+    if (workflowRunRecord != null) {
+      mainRunRecords.put(workflowRunRecord.getWorkflowRunId(), workflowRunRecord);
+    }
+    return mainRunRecords;
+  }
+
+  private Map<String, WorkflowRunRecord> getNeighbors(Id.Workflow id, String runId, int count, long timeInterval) {
```
bdmogal: even though this is a private method, because of the complexity, please add javadoc

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,133 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+  @GET
+  @Path("apps/{app-id}/workflows/{workflow-id}/runs/{run-id}/statistics")
+  public void workflowRunDetail(HttpRequest request, HttpResponder responder,
+                                @PathParam("namespace-id") String namespaceId,
+                                @PathParam("app-id") String appId,
+                                @PathParam("workflow-id") String workflowId,
+                                @PathParam("run-id") String runId,
+                                @QueryParam("count") int count,
+                                @QueryParam("interval") String interval) throws Exception {
+    if (count < 0) {
```
anwar6953: I would suggest naming this to `limit`, rather than count.
Doesn't that term seem more appropriate?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
gsps1: if you want to get the timestamp, add a comment, and also you can use the api ``splitter.forward(count)`` to skip initial bytes and then call ``splitter.getLong()``

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,78 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+    RunId pid = RunIds.fromString(runId);
+    long startTime = RunIds.getTime(pid, TimeUnit.SECONDS);
+    Map<String, WorkflowRunRecord> workflowRunRecords = new HashMap<>();
+    for (int i = (-1 * count); i <= count; i++) {
```
gsps1: this loop can we made better. use a while loop and instead of using ``prevStartTime = startTime + (i * timeInterval)``, update prevStartTime, based on the row what we find in scan, to avoid redoing some scan. you can instead do
``prevStartTime = time(L#268) + timeInterval``

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,79 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+      splitter.skipString();
+      long time = splitter.getLong();
+      if ((time >= (startTime - (count * timeInterval))) && time <= (startTime + (count * timeInterval))) {
+        Map<byte[], byte[]> columns = indexRow.getColumns();
+        String workflowRunId = Bytes.toString(columns.get(RUNID));
+        long timeTaken = Bytes.toLong(columns.get(TIME_TAKEN));
+
+        List<ProgramRun> programRunList = GSON.fromJson(Bytes.toString(columns.get(NODES)), PROGRAM_RUNS_TYPE);
+        workflowRunRecords.put(workflowRunId, new WorkflowRunRecord(workflowRunId, timeTaken, programRunList));
+      } else {
```
bdmogal: Invert this condition, and just add a break in the if. no else necessary

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -238,7 +314,7 @@ public ProgramType getProgramType() {
   /**
    * Internal class to keep track of Workflow Run Records
    */
-  static class WorkflowRunRecord {
+  public static class WorkflowRunRecord {
```
bdmogal: why?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -265,7 +341,7 @@ public String getWorkflowRunId() {
   /**
    * Internal Class for keeping track of programs in a workflow
    */
-  static class ProgramRun {
+  public static class ProgramRun {
```
bdmogal: why?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
```diff
@@ -113,5 +116,108 @@ public void testStatistics() throws Exception {
     response = doGet(request);
     Assert.assertEquals(HttpResponseStatus.BAD_REQUEST.getCode(),
                         response.getStatusLine().getStatusCode());
+    metricStore.deleteAll();
+  }
+
+  @Test
+  public void details() throws Exception {
```
bdmogal: more appropriate name. A method name should be an action, a verb. not a noun.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
```diff
@@ -113,5 +116,108 @@ public void testStatistics() throws Exception {
...
+      readResponse(response, new TypeToken<Map<String, Map<String, Map<String, Long>>>>() { }.getType());
+
+    // This should only return 3 because the 1st, 6th and 11th record were picked as the interval is 5 seconds
+    // and the workflow runs every second. For some reason the test fails when run with other test implying that
+    // state is saved somewhere. Ask someone how to delete the whole metrics thing.
+    Assert.assertEquals(2, workflowStatistics.size());
+  }
+
+  @Test
+  public void compare() throws Exception {
```
bdmogal: testCompare?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
gsps1: +1

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,133 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+  }
+
+  private Map<String, Map<String, Map<String, Long>>> convert(Map<String, WorkflowStatistics.DetailedStatistics>
+                                                                workflowDetailedStatisticsMap) {
+    Map<String, Map<String, Map<String, Long>>> formattedStatisticsMap = new HashMap<>();
+    for (Map.Entry<String, WorkflowStatistics.DetailedStatistics> entry : workflowDetailedStatisticsMap.entrySet()) {
+      Map<String, Map<String, Long>> internalMap = entry.getValue().getProgramToStatistics();
+      for (Map.Entry<String, Map<String, Long>> internalEntry : internalMap.entrySet()) {
+        Map<String, Map<String, Long>> newMap;
+        if ((newMap = formattedStatisticsMap.get(internalEntry.getKey())) == null) {
```
gsps1: refactor 
`` 
if (formattedStatisticsMap.get(key) == null) {
formattedStatisticsMap.put(key, new HashMap<>);
}
formattedStatisticsMap.get(key).put(somekey, somevalue));
``

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,133 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
     }
     responder.sendJson(HttpResponseStatus.OK, workflowStatistics);
   }
+
+  @GET
+  @Path("apps/{app-id}/workflows/{workflow-id}/runs/{run-id}/statistics")
+  public void workflowRunDetail(HttpRequest request, HttpResponder responder,
```
bdmogal: I think we decided to add javadocs for REST handler methods.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
bdmogal: Let's also convert this map into an object and add it to ``cdap-proto``

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +108,133 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    WorkflowDataset.WorkflowRunRecord workflowRunRecord = store.getWorkflowRun(workflowId, runId);
+    if (workflowRunRecord == null) {
+      return null;
+    }
+    List<WorkflowDataset.ProgramRun> programRuns = workflowRunRecord.getProgramRuns();
+    Map<String, Map<String, Long>> details = new HashMap<>();
+    for (WorkflowDataset.ProgramRun actionRun : programRuns) {
+      if (actionRun.getProgramType() == ProgramType.MAPREDUCE) {
+        details.put(actionRun.getName(),
+                    getMapreduceDetails(Id.Program.from(workflowId.getNamespaceId(), workflowId.getApplicationId(),
```
bdmogal: can extract ``Id.Program.from(workflowId.getNamespaceId(), workflowId.getApplicationId(), actionRun.getProgramType(), actionRun.getName()`` into a var so you don't have to repeat it for mapreduce and spark

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
```diff
@@ -331,6 +331,31 @@ public WorkflowStatistics apply(WorkflowStatsDataset dataset) throws Exception {
...
+      <WorkflowStatsDataset, WorkflowDataset.WorkflowRunRecord>() {
+      @Override
+      public WorkflowDataset.WorkflowRunRecord apply(WorkflowStatsDataset dataset) throws Exception {
+        return dataset.workflowDataset.getRecord(workflowId, runId);
+      }
+    });
+  }
+
+  @Override
+  public Map<String, WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(final Id.Workflow workflow,
```
bdmogal: what is a 'spaced record'?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -83,7 +86,7 @@ void write(Id.Workflow id, RunRecordMeta runRecordMeta, List<ProgramRun> program
    * @return List of WorkflowRunRecords
    * @throws Exception
```
bdmogal: when does it throw an exception?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
rohanahata: Yes we use this class in the WorkflowSLAHttpHandler

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/AppFabricTestBase.java
```diff
@@ -153,6 +156,9 @@
   private static LocationFactory locationFactory;
   private static File adapterDir;
 
+  protected static MetricStore metricStore;
```
bdmogal: why protected in AppFabricTestBase? Can you make them private in ``WorkflowSLAHttpHandlerTest``, since that's the only place they are used?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
```diff
@@ -40,11 +45,10 @@
  */
 public class WorkflowStatsSLAHttpHandlerTest extends AppFabricTestBase {
 
+
```
bdmogal: extra newline

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
```diff
@@ -113,5 +117,108 @@ public void testStatistics() throws Exception {
...
+    String request = String.format("%s/namespaces/%s/apps/%s/workflows/%s/runs/%s/statistics?count=%s&interval=%s",
+                                   Constants.Gateway.API_VERSION_3, Id.Namespace.DEFAULT,
+                                   WorkflowApp.class.getSimpleName(), workflowProgram.getId(), workflowRun1.toString(),
+                                   "20", "1s");
+
+    HttpResponse response = doGet(request);
+    Map<String, Map<String, Map<String, Long>>> workflowStatistics =
+      readResponse(response, new TypeToken<Map<String, Map<String, Map<String, Long>>>>() { }.getType());
+
+    // This should only return 3 because the 1st, 6th and 11th record were picked as the interval is 5 seconds
```
jayfjin: only return 3 but the assert is equal to 2?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
```diff
@@ -113,5 +117,108 @@ public void testStatistics() throws Exception {
...
+
+    Map<String, String> mapTypeContext = ImmutableMap.of(Constants.Metrics.Tag.NAMESPACE,
+                                                         mapreduceProgram.getNamespaceId(),
+                                                         Constants.Metrics.Tag.APP, mapreduceProgram.getApplicationId(),
+                                                         Constants.Metrics.Tag.MAPREDUCE, mapreduceProgram.getId(),
+                                                         Constants.Metrics.Tag.RUN_ID, mapreduceRunid.toString(),
+                                                         Constants.Metrics.Tag.MR_TASK_TYPE,
+                                                         MapReduceMetrics.TaskType.Mapper.getId());
+
+    metricStore.add(new MetricValues(mapTypeContext, MapReduceMetrics.METRIC_INPUT_RECORDS, 10, 38L,
```
jayfjin: magic numbers? perhaps extract as constant?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
rohanahata: So I changed the test case after this comment. If you notice, I have mentioned that I have to ask someone how to clear the previously run records. Following up on it, I have already talked to Shankar about it. Will add that assert in my next commit after doing some refactoring.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
```diff
@@ -113,5 +117,108 @@ public void testStatistics() throws Exception {
...
+    store.setWorkflowProgramStart(sparkProgram, sparkRunid.getId(), workflowProgram.getId(),
+                                  workflowRunId.getId(), sparkProgram.getId(),
+                                  RunIds.getTime(sparkRunid, TimeUnit.SECONDS), null, null);
+    store.setStop(sparkProgram, sparkRunid.getId(),
+                  TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()), ProgramRunStatus.COMPLETED);
+
+    store.setStop(workflowProgram, workflowRunId.getId(),
+                  TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()), ProgramRunStatus.COMPLETED);
+
+    TimeUnit.SECONDS.sleep(1);
```
jayfjin: what is this for?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
rohanahata: splitter.forward() does not exist

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
rohanahata: Similar reason... used in the Handler

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +111,181 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+                            context, new ArrayList<String>()));
+
+      for (MetricTimeSeries metricTimeSeries : resultPerQuery) {
+        overallResult.put(metricTimeSeries.getMetricName(), metricTimeSeries.getTimeValues().get(0).getValue());
+      }
+    }
+    return overallResult;
+  }
+
+  static class ProgramMetrics {
```
bdmogal: private?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +111,181 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    public ProgramType getProgramType() {
+      return programType;
+    }
+
+    public Map<String, Long> getMetrics() {
+      return metrics;
+    }
+  }
+
+  static class WorkflowRunMetrics {
```
bdmogal: private?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +206,73 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+    Map<String, WorkflowRunRecord> mainRunRecords = getNeighbors(workflow, RunIds.fromString(runId),
+                                                                 count, timeInterval);
+    WorkflowRunRecord workflowRunRecord = getRecord(workflow, runId);
+    if (workflowRunRecord != null) {
+      mainRunRecords.put(workflowRunRecord.getWorkflowRunId(), workflowRunRecord);
+    }
+    return mainRunRecords;
+  }
+
+  private Map<String, WorkflowRunRecord> getNeighbors(Id.Workflow id, RunId runId, int count, long timeInterval) {
```
bdmogal: please add some comments for this method

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +206,73 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+      if (indexRow == null) {
+        return workflowRunRecords;
+      }
+      mdsKey = new MDSKey(indexRow.getRow());
+      MDSKey.Splitter splitter = mdsKey.split();
+      splitter.skipString();
+      splitter.skipString();
+      splitter.skipString();
+      long time = splitter.getLong();
+      if ((time >= (startTime - (count * timeInterval))) && time <= (startTime + (count * timeInterval))) {
```
bdmogal: can you invert this condition?

```
if (!((time >= (startTime - (count * timeInterval))) && time <= (startTime + (count * timeInterval)))) {
  break;
}
```

This is to take out the else with a single break statement? Will reduce one level of indentation.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/AppFabricTestBase.java
```diff
@@ -17,11 +17,13 @@
 package co.cask.cdap.internal.app.services.http;
 
 import co.cask.cdap.api.Config;
+import co.cask.cdap.api.metrics.MetricStore;
```
bdmogal: unused imports? this, line 26 and line 40?

cdap-proto/src/main/java/co/cask/cdap/proto/WorkflowStatsComparison.java
```diff
@@ -0,0 +1,94 @@
...
+ * the License.
+ */
+package co.cask.cdap.proto;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Returns detailed statistics per run
```
bdmogal: I don't think it does :)

cdap-proto/src/main/java/co/cask/cdap/proto/WorkflowStatsComparison.java
```diff
@@ -0,0 +1,94 @@
...
+  public Map<String, Long> getStartTimes() {
+    return startTimes;
+  }
+
+  public Collection<ProgramNodes> getProgramNodesList() {
+    return programNodesList;
+  }
+
+  /**
+   * Contains information about Programs
```
bdmogal: what information?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +111,181 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+                      @PathParam("namespace-id") String namespaceId,
+                      @PathParam("app-id") String appId,
+                      @PathParam("workflow-id") String workflowId,
+                      @PathParam("run-id") String runId,
+                      @QueryParam("other-run-id") String otherRunId) throws Exception {
+    Id.Workflow workflow = Id.Workflow.from(Id.Namespace.from(namespaceId), appId, workflowId);
+    WorkflowRunMetrics detailedStatistics = getDetailedRecord(workflow, runId);
+    WorkflowRunMetrics otherDetailedStatistics = getDetailedRecord(workflow, otherRunId);
+    if (detailedStatistics == null) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND, "The run-id provided was not found : " + runId);
```
bdmogal: nit: since you are using exceptions everywhere in this class, throw a ``NotFoundException`` here too?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +111,181 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+      return;
+    }
+    if (otherDetailedStatistics == null) {
+      responder.sendString(HttpResponseStatus.NOT_FOUND, "The other run-id provided was not found : " + otherRunId);
```
bdmogal: same nit.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +111,181 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    }
+    return overallResult;
+  }
+
+  static class ProgramMetrics {
+    private final String programName;
+    private final ProgramType programType;
+    private final Map<String, Long> metrics;
+
+    public ProgramMetrics(String programName, ProgramType programType, Map<String, Long> metrics) {
```
bdmogal: why public?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +111,181 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    public Map<String, Long> getMetrics() {
+      return metrics;
+    }
+  }
+
+  static class WorkflowRunMetrics {
+    private final String workflowRunId;
+    private final List<ProgramMetrics> programMetricsList;
+
+    public WorkflowRunMetrics(String workflowRunId, List<ProgramMetrics> programMetricsList) {
```
bdmogal: why public?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
rohanahata: Since they are spaced apart by timeIntervals I called it so

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
rohanahata: This Map was the way I was using sets to ensure uniqueness. A set of workflow records did not work out because creating a new object each time with the same params still counts as different objects. Hence, a map was used. Have converted the store to reflect collection though but internally a map is used.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
gsps1: i see its private.. then you can do something like this
``
byte[] rowKey = indexRow.getRow();
// ByteBuffer.wrap(byte[], offset, length)
long time = ByteBuffer.wrap(rowKey, rowKey.length() - Bytes.SIZEOF_LONG, SIZEOF_LONG).getLong()
``
since we know the time is at the end and the length of it (long), doing it this way is robust.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowStatsSLAHttpHandlerTest.java
```diff
@@ -114,4 +159,110 @@ public void testStatistics() throws Exception {
...
+                                  workflowRunId.getId(), sparkProgram.getId(),
+                                  RunIds.getTime(sparkRunid, TimeUnit.SECONDS), null, null);
+    store.setStop(sparkProgram, sparkRunid.getId(),
+                  TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()), ProgramRunStatus.COMPLETED);
+
+    store.setStop(workflowProgram, workflowRunId.getId(),
+                  TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()), ProgramRunStatus.COMPLETED);
+
+    // Sleep so that 2 runs of the workflow dont have same start time which causes exception to be thrown
+    TimeUnit.SECONDS.sleep(1);
```
gsps1: i asked you to remove the sleep.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +111,181 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+  @Path("apps/{app-id}/workflows/{workflow-id}/runs/{run-id}/statistics")
+  public void workflowRunDetail(HttpRequest request, HttpResponder responder,
+                                @PathParam("namespace-id") String namespaceId,
+                                @PathParam("app-id") String appId,
+                                @PathParam("workflow-id") String workflowId,
+                                @PathParam("run-id") String runId,
+                                @QueryParam("limit") int limit,
+                                @QueryParam("interval") String interval) throws Exception {
+    if (limit < 0) {
+      throw new BadRequestException("Count has to be greater than or equal to 0");
```
gsps1: you have to send responder the error status and message, not throw the exception.
```responder.sendError(HttpResponseStatus.BAD_REQUEST, "Count has to be greater than or equal to 0")
   return;
```

bdmogal: @gsps1 we added support for handling specific exceptions in HttpExceptionHandler, so this will work.

gsps1: ignore the comment. i see we use ``HttpExceptionHandler`` to handle these exceptions.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
gsps1: then remove the comment about it being an internal class, otherwise move it out from this class and make it non-static.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
anwar6953: update exception message, based upon the change of count -> limit

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +111,197 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    context.put(Constants.Metrics.Tag.APP, sparkProgram.getApplicationId());
+    context.put(Constants.Metrics.Tag.SPARK, sparkProgram.getId());
+    context.put(Constants.Metrics.Tag.RUN_ID, runId);
+
+    List<TagValue> tags = new ArrayList<>();
+    for (Map.Entry<String, String> entry : context.entrySet()) {
+      tags.add(new TagValue(entry.getKey(), entry.getValue()));
+    }
+    MetricSearchQuery metricSearchQuery = new MetricSearchQuery(0, 0, Integer.MAX_VALUE, tags);
+    Collection<String> metricNames = metricStore.findMetricNames(metricSearchQuery);
```
anwar6953: @gsps1
Shouldn't the findMetricNames not require a MetricSearchQuery object (which has start time, end time, and limit - unnecessary for finding metric names)?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
```diff
@@ -331,6 +331,31 @@ public WorkflowStatistics apply(WorkflowStatsDataset dataset) throws Exception {
...
+      @Override
+      public WorkflowDataset.WorkflowRunRecord apply(WorkflowStatsDataset dataset) throws Exception {
+        return dataset.workflowDataset.getRecord(workflowId, runId);
+      }
+    });
+  }
+
+  @Override
+  public Collection<WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(final Id.Workflow workflow,
+                                                                              final String runId,
```
anwar6953: nit: indentation

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
gsps1: yeah, good point. currently its the configurable so the user can scan in multiple resolution tables for metrics.. but it will be useful to have searchAllMetrics() in MetricStore API, which will look into aggregates table for metric names by default..

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -501,4 +501,25 @@ void setWorkflowProgramStart(Id.Program programId, String programRunId, String w
...
+   * Get a list of workflow runs that are spaced apart by time interval in both directions from the run id provided.
+   *
+   * @param workflow The workflow whose statistics need to be obtained
+   * @param runId The run id of the workflow
+   * @param count The count of the records that the user wants to compare against
+   * @param timeInterval The timeInterval with which the user wants to space out the runs
+   * @return Map of runId of Workflow to DetailedStatistics of the run
+   */
+  Collection<WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(Id.Workflow workflow, String runId,
+                                                                       int count, long timeInterval);
```
gsps1: wrong indentation. please format. use (cmd + alt + i) to auto format in intelliJ

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -30,6 +31,8 @@
 import com.google.common.primitives.Longs;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+import org.apache.avro.generic.GenericData;
```
gsps1: why generic record ? unused import ?

cdap-proto/src/main/java/co/cask/cdap/proto/WorkflowStatsComparison.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+package co.cask.cdap.proto;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Returns an object that has a map from WorkflowRunId to time and a list of metrics per ProgramRun
+ * which are further divided into run that the action took in a specific workflow.
```
gsps1: this comment is very confusing.. please make is clear.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,78 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
     return programToRunRecord.values();
   }
 
+  @Nullable
+  WorkflowRunRecord getRecord(Id.Workflow id, String pid) {
```
gsps1: private ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,78 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+    }
+    Map<byte[], byte[]> columns = indexRow.getColumns();
+    String workflowRunId = Bytes.toString(columns.get(RUNID));
+    long timeTaken = Bytes.toLong(columns.get(TIME_TAKEN));
+
+    List<ProgramRun> actionRunsList = GSON.fromJson(Bytes.toString(columns.get(NODES)), PROGRAM_RUNS_TYPE);
+    return new WorkflowRunRecord(workflowRunId, timeTaken, actionRunsList);
+  }
+
+  Collection<WorkflowRunRecord> getDetailsOfRange(Id.Workflow workflow, String runId, int limit, long timeInterval) {
```
gsps1: private?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -204,6 +207,78 @@ public int compare(WorkflowDataset.WorkflowRunRecord o1, WorkflowDataset.Workflo
...
+   * @param timeInterval The time interval that we want the results to be spaced apart
+   * @return A Map of WorkflowRunId to the corresponding Workflow Run Record. A map is used so that duplicates of
+   * the WorkflowRunRecord are not obtained
+   */
+  private Map<String, WorkflowRunRecord> getNeighbors(Id.Workflow id, RunId runId, int count, long timeInterval) {
+    long startTime = RunIds.getTime(runId, TimeUnit.SECONDS);
+    Map<String, WorkflowRunRecord> workflowRunRecords = new HashMap<>();
+    int i = -count;
+    long prevStartTime = startTime - (timeInterval * count);
+    while (prevStartTime <= startTime + (count * timeInterval)) {
```
gsps1: add a comment about the program logic here for this block.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -236,9 +311,9 @@ public ProgramType getProgramType() {
   }
 
   /**
-   * Internal class to keep track of Workflow Run Records
+   * Class to keep track of Workflow Run Records
    */
-  static class WorkflowRunRecord {
+  public static class WorkflowRunRecord {
```
gsps1: make it final

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
```diff
@@ -263,9 +338,9 @@ public String getWorkflowRunId() {
   }
 
   /**
-   * Internal Class for keeping track of programs in a workflow
+   * Class for keeping track of programs in a workflow
    */
-  static class ProgramRun {
+  public static class ProgramRun {
```
gsps1: final

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
rohanahata: package-private because of use in DefaultStore

cdap-proto/src/main/java/co/cask/cdap/proto/WorkflowStatsComparison.java
```diff
@@ -0,0 +1,112 @@
...
+
+  public Collection<ProgramNodes> getProgramNodesList() {
+    return programNodesList;
+  }
+
+  /**
+   * Contains information about a Program and the statistics that correspond to it
+   * divided per run of the workflow that the program ran in.
+   */
+  public static class ProgramNodes {
```
gsps1: final

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
rohanahata: directly called from DefaultStore

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/WorkflowDataset.java
gsps1: ok

gsps1:
few minor comments, LGTM otherwise

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +125,221 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+  }
+
+  private static class ProgramMetrics {
+    private final String programName;
+    private final ProgramType programType;
+    private final String programRunId;
+    private final long programStartTime;
+    private final Map<String, Long> metrics;
+
+    ProgramMetrics(String programName, ProgramType programType, String programRunId,
```
bdmogal: private

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
```diff
@@ -89,4 +125,221 @@ public void workflowStats(HttpRequest request, HttpResponder responder,
...
+    public String getProgramRunId() {
+      return programRunId;
+    }
+  }
+
+  private static class WorkflowRunMetrics {
+    private final String workflowRunId;
+    private final List<ProgramMetrics> programMetricsList;
+
+    WorkflowRunMetrics(String workflowRunId, List<ProgramMetrics> programMetricsList) {
```
bdmogal: private

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java
gsps1: have added jira for it

PR #3648, opened 2015-08-20
Feature/ui styling adapter modals ff safari

astewes:
*  Adds custom styling for select form element
*  Adds helper class to facilitate custom styling
*  Changes some hex color values to variables
*  Contains some minor less code cleanup

Below are GIFs of the adapter modal select elements in action:

Modal header buttons using % width:
ff2

Firefox: 
firefox

Safari:
safari

Chrome:
chrome


cdap-ui/app/features/adapters/etlapps.less
```diff
@@ -193,7 +191,7 @@ body.theme-cdap.state-adapters-create {
         .modal-btns {
           margin-right: -15px;
           text-align: right;
-
+          height: 60px;
```
ajainarayanan: We might not need a height here. The individual button widths need to be managed with media queries.

cdap-ui/app/styles/themes/cdap/mixins.less
```diff
@@ -154,3 +154,13 @@
   order: @val;
 }
 
+.appearance(@value: none) {
+  appearance: @value;
+  -webkit-appearance: @value;
+  -moz-appearance: @value;
+  -o-appearance: @value;
+  ::-ms-expand {
+    display: none;  // Need this to hide default select in IE10
```
ajainarayanan: If we hide this for IE10 would we be even seeing the dropdowns in that browser?

astewes:
in IE10, the default dropdown arrow would be hidden, but the custom arrow (placed via the helper class .select-wrapper) would take care of this

ajainarayanan:
:+1: LGTM

PR #3650, opened 2015-08-20

jwang47:
https://issues.cask.co/browse/CDAP-3419
http://builds.cask.co/browse/CDAP-RBT415


cdap-master/src/test/java/co/cask/cdap/data/tools/flow/HBaseQueueDebuggerTest.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright © 2014 Cask Data, Inc.
```
bdmogal: 2015

bdmogal:
This is a cherry-pick, right? If so, one minor comment, then LGTM. Also, could you update the copyright year on develop?

PR #3651, opened 2015-08-20
Correct IRC link.

awholegunch:
Fixes the link to the IRC channel (it was being rendered as HTTP rather than IRC) and moves a period outside of an HTTP link.


bdmogal:
:+1:

PR #3655, opened 2015-08-21
Fix/ui predefined plugins structure

elfenheart:
modify saving templates plugin to be categorized into source, transforms, and sinks

### DIFF
--- a/cdap-ui/app/features/adapters/controllers/create/canvas-ctrl.js
+++ b/cdap-ui/app/features/adapters/controllers/create/canvas-ctrl.js
@@ -5,7 +5,9 @@ angular.module(PKG.name + '.feature.adapters')
         transformTemplates = [],
         sinkTemplates = [];
 
-    function objectToArray(obj, arr) {
+    function objectToArray(obj) {
+      var arr = [];
+
       angular.forEach(obj, function (val) {
         if (val.templateType === MyPlumbService.metadata.template.type) {
           val.icon = 'fa-plug';
@@ -14,15 +16,17 @@ angular.module(PKG.name + '.feature.adapters')
           arr.push(val);
         }
       });
+
+      return arr;
     }
 
     mySettings.get('pluginTemplates')
       .then(function (res) {
         var templates = res[$state.params.namespace][MyPlumbService.metadata.template.type];
 
-        objectToArray(templates.source, sourceTemplates);
-        objectToArray(templates.transform, transformTemplates);
-        objectToArray(templates.sink, sinkTemplates);
+        sourceTemplates = objectToArray(templates.source);
+        transformTemplates = objectToArray(templates.transform);
+        sinkTemplates = objectToArray(templates.sink);
       });
 
 
@@ -318,7 +322,7 @@ angular.module(PKG.name + '.feature.adapters')
           id: id,
           name: item.pluginName,
           icon: MyPlumbFactory.getIcon(item.pluginName),
-          type: item.type,
+          type: item.pluginType,
           properties: item.properties,
           outputSchema: item.outputSchema,
           templateName: item.templateName,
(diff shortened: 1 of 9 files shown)


cdap-ui/app/features/admin/controllers/namespace/templates-ctrl.js
```diff
@@ -152,7 +152,15 @@ angular.module(PKG.name + '.feature.admin')
             res[namespace] = {};
           }
 
-          if (res[namespace][properties.templateName] && !vm.isEdit) {
+          if (!res[namespace][properties.templateType]) {
```
ajainarayanan: Is it possible to use ```myHelpers.deepset``` and ```myHelpers.objectQuery```? We could do something like this,
```Javascript 

if (!myHelpers.objectQuery(res, properties.templateType, properties.type, properties.templateName)) {
  myHelpers.deepSet(res, [properties.templateType, properties.templateName, properties.type].join('.'));
}
```

cdap-ui/app/features/admin/controllers/namespace/templates-list-ctrl.js
```diff
@@ -1,27 +1,53 @@
...
     vm.isEmpty = function () {
-      return Object.keys(vm.list).length === 0;
+      return vm.list.length === 0;
     };
 
     vm.delete = function (template) {
       mySettings.get('pluginTemplates')
         .then(function (res) {
-          delete res[$stateParams.nsadmin][template.templateName];
+          delete res[$stateParams.nsadmin][template.templateType][template.type][template.templateName];
```
ajainarayanan: This might be a little subjective, but ```template.templateType``` and ```template.type``` seems confusing. When I look at it at an instant its getting hard to settle down which one means which thing. There ```ETLBatch``` and ```ETLRealtime``` and then there is ```source, sink and transforms```. One is template type and the other is plugin type. I feel we should try changing it so that it becomes appropriate for us when we come back and see this after 2-3 months :smile:

cdap-ui/app/features/admin/controllers/namespace/templates-list-ctrl.js
```diff
@@ -1,27 +1,53 @@
...
+      });
+    }
+
+    function processResult(response) {
+      if (response) {
+        vm.list = [];
+
+        var sources, transforms, sinks;
+
+        sources = myHelpers.objectQuery(response, $stateParams.nsadmin, 'ETLBatch', 'source');
```
ajainarayanan: This could wrapped something like,

```Javascript

function processResult(response) {

  if (!response) {
     return;
  } 
  vm.list = [];
  ['ETLBatch', 'ETLRealtime'].forEach(function(templatetype) {
     ['source', 'transforms', 'sink'].forEach(function(plugintype) {
       var pluginArray = objectToArray(myHelpers.objectQuery(response, $stateParams.nsadmin, templatetype, plugintype));
       vm.list = vm.list.concat(pluginArray);
     });
  });
}
function objectToArray(obj) {
  var returnArray = [];
   angular.forEach(obj, function (value) {
     returnArray.push(value);
  });
  return returnArray;
}
```

This way we know where and who updates ```vm.list``` and there is no repetition of code. Right now vm.list is manipulated in three different places and by looking at processResult I don't see why we use that function until I see ```objectToArray``` function.

cdap-ui/app/features/adapters/controllers/create/canvas-ctrl.js
```diff
@@ -5,24 +5,24 @@ angular.module(PKG.name + '.feature.adapters')
...
-            } else if (template.type === 'transform') {
-              transformTemplates.push(template);
-            } else if (template.type === 'sink') {
-              sinkTemplates.push(template);
-            }
-          }
-        });
+        var templates = res[$state.params.namespace][MyPlumbService.metadata.template.type];
 
+        objectToArray(templates.source, sourceTemplates);
```
ajainarayanan: I am guessing we could change here if we are changing it here (https://github.com/caskdata/cdap/pull/3655/files#r37775127). Instead of passing in a reference and have the utility function change the array we could get an array as return type and we could do whatever we want with it. Thoughts?

ajainarayanan:
:+1: LGTM

PR #3657, opened 2015-08-21
[CDAP-3014] Propagate the actual cause instead of wrapping a RuntimeE…

bdmogal:
…xception inside another RuntimeException in WorkflowDriver.

Also log the exception when exceptions occur.

Jira: [CDAP-3014](https://issues.cask.co/browse/CDAP-3014)
Build: http://builds.cask.co/browse/CDAP-DUT2576


sagarkapare:
LGTM :+1:

sagarkapare:
Build passed and tested on the cluster as well. Merging.

PR #3660, opened 2015-08-21
Upgrade tool fix

sreevatsanraman:
* Fixed cases where the upgrade tool doesn't update versions for all tables.


poornachandra:
LGTM

PR #3664, opened 2015-08-21
Changed timeout to avoid ES realtime test failures

ataylor4:
Elasticsearch Realtime test was failing because it timed out before the records were written, so I increased the time. This is a temporary fix and will be resolved once adapters more fully become applications and thus have access to metrics.

https://issues.cask.co/browse/CDAP-3459
http://builds.cask.co/browse/CDAP-DUT2582-2

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-realtime/src/test/java/co/cask/cdap/template/etl/realtime/RealtimeElasticsearchSinkTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-realtime/src/test/java/co/cask/cdap/template/etl/realtime/RealtimeElasticsearchSinkTest.java
@@ -17,6 +17,7 @@
 package co.cask.cdap.template.etl.realtime;
 
 import co.cask.cdap.common.utils.Networks;
+import co.cask.cdap.common.utils.Tasks;
 import co.cask.cdap.proto.AdapterConfig;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.template.etl.api.PipelineConfigurable;
@@ -63,6 +64,7 @@ import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
 import static org.elasticsearch.index.query.QueryBuilders.matchQuery;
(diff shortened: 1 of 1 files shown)


cdap-app-templates/cdap-etl/cdap-etl-realtime/src/test/java/co/cask/cdap/template/etl/realtime/RealtimeElasticsearchSinkTest.java
```diff
@@ -138,7 +138,7 @@ public void testESSink() throws Exception {
       AdapterManager adapterManager = createAdapter(adapterId, adapterConfig);
 
       adapterManager.start();
-      TimeUnit.SECONDS.sleep(5);
+      TimeUnit.SECONDS.sleep(10);
```
bdmogal: shouldn't have a sleep. I think there is an ``adapterManager.waitForOneRunToFinish()``. See other adapter tests for examples.

sreevatsanraman: instead of sleeping for random interval, can we check for the success condition sleeping every 1 seconds upto a specified retry. If the success criteria is not met after the n retries then fail the test. Sleeping for 10 seconds may fail if the build box is slower. For example see HelloWorldTest

```
 RuntimeMetrics metrics = flowManager.getFlowletMetrics("saver");

 metrics.waitForProcessed(5, 5, TimeUnit.SECONDS);
```

gokulavasan: @bdmogal I think that works well only for Batch Adapters. For realtime adapters, the run doesn't complete until we stop the adapter?

gokulavasan: @sreevatsanraman For metrics in adapter, we currently don't have a way to expose it. After @gsps1 's changes, we have this capability in Flows. But we need to add that for Workers and expose it through AdapterManager. But eventually we will remove AdapterManager. So should we make the changes?

bdmogal: @gokulavasan  Just thought of a way to do it. Spoke with @ataylor4. Fix should be coming soon.

bdmogal:
LGTM. Please rebase to squash commits and add jira in the commit message.

bdmogal:
Buil passed. Merging.

PR #3670, opened 2015-08-24
Fix testWorkerInstances test to use Tasks.waitFor instead of retry...

bdmogal:
...with explicit sleep. Currently, this test fails sometimes.

Build: https://builds.cask.co/browse/CDAP-DUT2584


chtyim:
LGTM

PR #3673, opened 2015-08-24
Use 3.1.1-2 release artifacts

wolf31o2:
Otherwise, it'll use whatever is CDAP cookbook default. Same as #3668 for 3.1


sreevatsanraman:
LGTM :+1:

PR #3677, opened 2015-08-24
(CDAP-3458) Don’t reset the position to 0 on IOException

chtyim:
- If the exception is EOFException, the StreamDataFileReader won’t get closed, hence the reopening of the underlying input stream should just resume from where it left off
- If the exception is other type of IOException, it will get propagated outside, which the expected behavior is to close and create a new StreamDataFileReader, which the initial position will be passed from the caller, hence resetting of position to 0 is not need.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamDataFileReader.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamDataFileReader.java
@@ -262,7 +262,6 @@ public final class StreamDataFileReader implements FileReader<PositionStreamEven
         }
       }
     } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


sagarkapare:
LGTM :+1:

sreevatsanraman:
Can we add a test case for this fix?

chtyim:
It's very difficult to add a unit-test to it as it only happens in distributed mode with data node replication factor > 1 and apparently only happen on Hadoop-2.6. It's probably caused by delay in data block meta data replication, which is very difficult to simulate in unit-test env.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2590-2

Also tested on cluster overnight.

PR #3679, opened 2015-08-24
CDAP-3450 move app deploy logic from http handler into service

albertshau:
Some refactoring to move logic out of the AppLifecycleHttpHandler
into AppLifecycleService. Http handlers should contain as little
logic as possible, with most logic residing in the services since
they can be re-used in different places. For example, CDAP upgrade
may need to make use of this logic when upgrading existing apps to
the new artifact framework.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2592

albertshau:
actually, need more work, putting on hold

gokulavasan:
LGTM :+1:

PR #3680, opened 2015-08-24
[CDAP-3398] Refactor Schema Editor Widget

elfenheart:
removing stream properties and consolidating that widget to the schema widget with additional config in JSON template called property-watch, which will specify the format of the schema


ajainarayanan:
@elfenheart The adpater html views are being deleted in a separate PR that I will be creating. Could add them back so that we don't have a merge conflict? I am deleting the files as part of the refactor/reorganization that we are doing for adapters.

cdap-ui/app/directives/widget-container/widget-schema-editor/widget-schema-editor.js
```diff
@@ -5,10 +5,13 @@ angular.module(PKG.name + '.commons')
       scope: {
         model: '=ngModel',
         config: '=',
+        plugin: '=',
```
ajainarayanan: Right now this is a little mis-leading. plugin here is actually plugin properties and config is schema properties. Is it possible to use the same name or something that signifies what is being passed?

ajainarayanan:
:+1: LGTM

PR #3683, opened 2015-08-24
Fixes all race condition in Worker.stop() usage. Generally, stop() me…

chtyim:
…thod happens in different thread than the init(), run() and destroy() methods. In specifics:

- Shouldn’t set running = true in the run() method and setting it to false in stop(). It is because it is possible that stop() get called from another thread which set the “running” to false, before run() is able to set it to true. If that happen, the Worker thread will be keep running forever.
- Shouldn’t use getContext() from stop() because the context field is not volatile, and also it’s possible stop() is called from another thread which even before the initialize() is being invoked.


gokulavasan:
LGTM :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2597-1

PR #3686, opened 2015-08-25
Updated code to return the root cause of the exception through the service handler.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-1892

Before the fix service always returned the same message `Error: 500: Transaction failure when committing changes. Aborted transaction.` irrespective of the root cause.

After the fix, we return the stack trace of the root cause, so that it is easier for the user to pin point the reason for the exception through UI as well as through CLI without checking the master log files.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/DelayedHttpServiceResponder.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/DelayedHttpServiceResponder.java
@@ -21,6 +21,7 @@ import co.cask.cdap.api.service.http.HttpServiceResponder;
 import co.cask.http.HttpResponder;
 import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.gson.Gson;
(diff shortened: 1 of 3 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/HttpHandlerGenerator.java
```diff
@@ -613,7 +613,9 @@ private void generateTransactionalDelegateBody(GeneratorAdapter mg, Method metho
 
       // wrappedResponder.setTransactionFailureResponse();
```
gokulavasan: wrappedResponder.setTransactionFailureResponse(txFailureException);

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/DelayedHttpServiceResponder.java
```diff
@@ -176,8 +177,11 @@ private void send(int status, ByteBuffer content, String contentType,
    * Since calling one of the send methods multiple times logs a warning, upon transaction failures this
    * method is called to allow setting the failure response without an additional warning.
    */
-  public void setTransactionFailureResponse() {
-    ByteBuffer buffer = Charsets.UTF_8.encode("Transaction failure when committing changes. Aborted transaction.");
+  public void setTransactionFailureResponse(Throwable t) {
+    Throwable rootCause = Throwables.getRootCause(t);
+    ByteBuffer buffer = Charsets.UTF_8.encode("Exception occurred while handling request: "
```
gsps1: should we say "Aborted transaction due to exception ..." In that way we also pass the information that the transaction is aborted.

gsps1:
just one minor comment, LGTM otherwise.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/DelayedHttpServiceResponder.java
sagarkapare: I think `Exception occurred while handling request: ` text is more relevant. no? Since here the user is sending/creating the request and not the transaction and from his perspective request is failed. He may not be aware of the transaction at all.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/HttpHandlerGenerator.java
sagarkapare: fixed.

sagarkapare:
@gsps1 @gokulavasan addressed/answered the comments. Please take a look again. Thanks!

gokulavasan:
LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/DelayedHttpServiceResponder.java
gsps1: ok.

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT2598-2
Merging.

PR #3690, opened 2015-08-25
Adding methods in Programs to allow use Plugins

gokulavasan:
Instead of using configurer, adding helper methods in Abstract Program classes. Modifying the unit test to make use of it.

Build : http://builds.cask.co/browse/CDAP-DUT2606-2


cdap-api/src/main/java/co/cask/cdap/api/artifact/PluginConfigurer.java
```diff
@@ -25,7 +26,7 @@
  * This interface provides methods to register plugin usage in a CDAP Program. The registered plugins will be
  * available at the runtime of the Program.
  */
-public interface PluginConfigurer {
+public interface PluginConfigurer extends DatasetConfigurer {
```
albertshau: is DatasetConfigurer ever used on its own? If not maybe its better to have just one interface?

cdap-api/src/main/java/co/cask/cdap/internal/api/AbstractPluginConfigurable.java
```diff
@@ -0,0 +1,56 @@
...
+import co.cask.cdap.api.artifact.PluginSelector;
+import co.cask.cdap.api.templates.plugins.PluginProperties;
+
+import javax.annotation.Nullable;
+
+/**
+ * Helper methods to add plugins to Programs.
+ * @param <T> Program's configurer that implements {@link PluginConfigurer}
+ */
+public abstract class AbstractPluginConfigurable<T extends PluginConfigurer>
```
albertshau: similar question here, is AbstractProgramDatasetConfigurable ever used or does everything extend AbstractPluginConfigurable now

cdap-api/src/main/java/co/cask/cdap/api/artifact/PluginConfigurer.java
gokulavasan: It is used by AbstractFlowlet, AbstractSpark classes. Until we have support for plugins in both, we need these two interfaces. Also, ApplicationConfigurer extends DatasetConfigurer but not PluginConfigurer.

cdap-api/src/main/java/co/cask/cdap/internal/api/AbstractPluginConfigurable.java
gokulavasan: Same reason as https://github.com/caskdata/cdap/pull/3690#discussion_r37902221
AbstractSpark, AbstractFlowlet extend AbstractProgramDatasetConfigurable.

cdap-api/src/main/java/co/cask/cdap/api/artifact/PluginConfigurer.java
albertshau: don't we want to be able to use plugins in the application as well?

albertshau:
one question about adding it to apps, otherwise lgtm

cdap-api/src/main/java/co/cask/cdap/api/artifact/PluginConfigurer.java
gokulavasan: Currently we send the Plugin information used in a specific program through its ProgramSpecification. So if we want to support adding Plugins in the Application, then we need to pass that info through ApplicationSpecification. Should we support that? It is easier for streams/datasets since it should be unique across the app and programs. But for plugins, we want to support local plugins and global plugins?

albertshau: I think it makes sense if the plugin is available across the entire app instead of local to a program. It is more consistent with the stream and dataset behavior.  Also, we have a use case in data quality where an aggregation function plugin is used in both a mapreduce program and a service program. In that case it's easier to register it once in the app versus in both programs.

gokulavasan:
@albertshau Addressed your comments. Please review.

cdap-api/src/main/java/co/cask/cdap/api/artifact/PluginConfigurer.java
gokulavasan: Sure, will allow Applications to add Plugins as well, which will be available to all programs. But currently not adding the ability to add plugins to Flows/Services. Need to figure out there is a use case for it.

albertshau:
one comment

gokulavasan:
@albertshau Addressed your comment. Now one can add plugins in Workers, Service (and in ServiceHandler), MapReduce and also in Application. All plugins will be visible across the entire application irrespective of where they are added, similar to streams/datasets. This also removes the problem of Location being made visible through Program specification since now all that info is stored in ApplicationSpecification.

gokulavasan:
Also verified that the change works in distributed mode.

albertshau:
:+1:

PR #3691, opened 2015-08-25
(CDAP-3458) Don’t reset the position to 0 on IOException

sreevatsanraman:
- If the exception is EOFException, the StreamDataFileReader won’t get closed, hence the reopening of the underlying input stream should just resume from where it left off
- If the exception is other type of IOException, it will get propagated outside, which the expected behavior is to close and create a new StreamDataFileReader, which the initial position will be passed from the caller, hence resetting of position to 0 is not need.
(cherry picked from commit 0d02d34d7d0222914d90abefa08aea242ae5947b)


chtyim:
LGTM

PR #3693, opened 2015-08-25

jwang47:
http://builds.cask.co/browse/CDAP-DUT2544

### DIFF
--- a/cdap-master/src/main/java/co/cask/cdap/data/tools/HBaseQueueDebugger.java
+++ b/cdap-master/src/main/java/co/cask/cdap/data/tools/HBaseQueueDebugger.java
@@ -17,14 +17,32 @@
 package co.cask.cdap.data.tools;
 
 import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.flow.FlowSpecification;
+import co.cask.cdap.api.flow.FlowletConnection;
+import co.cask.cdap.app.ApplicationSpecification;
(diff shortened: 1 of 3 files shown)


cdap-master/src/main/java/co/cask/cdap/data/tools/HBaseQueueDebugger.java
```diff
@@ -303,28 +405,123 @@ public long getTotal() {
...
+    Injector injector = Guice.createInjector(
+      new ConfigModule(CConfiguration.create(), HBaseConfiguration.create()),
+      new IOModule(),
+      new ZKClientModule(),
+      new LocationRuntimeModule().getDistributedModules(),
+      new DiscoveryRuntimeModule().getDistributedModules(),
+      new StreamAdminModules().getDistributedModules(),
+      new NotificationFeedClientModule(),
+      new TwillModule(),
+      new ExploreClientModule(),
```
chtyim: Why need to bring in so many seemingly unrelated guice modules?

jwang47: `HBaseQueueDebugger` depends on `Store` which appears to require these guice modules.

jwang47:
This is a cherry-pick without conflicts from https://github.com/caskdata/cdap/pull/3650, so merging since build passed.

PR #3698, opened 2015-08-25
Fix/ui adapter modal+gridsterupdate

ajainarayanan:
- Updates angular gridster 
   Right now we have jserror when we switch tabs. angular-gridster internally nullifies grid data when we switch tab which causes issue (https://github.com/ManifestWebDesign/angular-gridster/issues/262). Hence the upgrade.
- Fixes adapter modal right buttons when browser size is reduced.


elfenheart:
:+1: LGTM on UI changes

sreevatsanraman:
pom changes looks good to me.

PR #3700, opened 2015-08-25
[CDAP-3248] add validation schema for unique fields

elfenheart:
Adding validation for unique fields in plugin edit output schema

validateschema


ajainarayanan:
:+1: LGTM

PR #3702, opened 2015-08-25
CDAP-3497 add a way in unit tests to export packages in an app

albertshau:
This is needed if an app would normally export packages for
plugins to implement.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2614

gokulavasan:
LGTM :+1:

PR #3703, opened 2015-08-26
CDAP-3482 Sort CLASSPATH in Standalone

wolf31o2:
Standalone loads all JARs from CDAP_HOME/lib/* which relies on the
underlying file-system's ordering, which is not guaranteed. This
was causing instability due to the differences between Mac OS X
file-system (HFS+) and Linux (Ext3/4).


sreevatsanraman:
LGTM :+1:

PR #3704, opened 2015-08-26
CDAP-3501 store artifact range inclusive information

albertshau:
Fixes a bug where information about version inclusivity is dropped
in the artifact store which can cause bugs if the version range
originally used non-default inclusive flags.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2618

gokulavasan:
LGTM :+1:

PR #3705, opened 2015-08-26
CDAP-3482 Sort CLASSPATH in Standalone

sreevatsanraman:
Standalone loads all JARs from CDAP_HOME/lib/* which relies on the
underlying file-system's ordering, which is not guaranteed. This
was causing instability due to the differences between Mac OS X
file-system (HFS+) and Linux (Ext3/4).

(cherry picked from commit 3da35a5e24ad1a913aa18afe95651f070017ea40)


wolf31o2:
:+1: LGTM

PR #3706, opened 2015-08-26
CDAP-3451 adding CDAP upgrade to v.3.2 for apps

albertshau:
The upgrade will create artifacts from existing applications and
update application metadata to point to those artifacts

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
@@ -28,6 +28,7 @@ import co.cask.cdap.api.workflow.WorkflowSpecification;
 import co.cask.cdap.app.ApplicationSpecification;
 import co.cask.cdap.app.deploy.Manager;
 import co.cask.cdap.app.deploy.ManagerFactory;
+import co.cask.cdap.app.program.ManifestFields;
 import co.cask.cdap.app.program.Programs;
 import co.cask.cdap.app.runtime.ProgramRuntimeService;
 import co.cask.cdap.app.store.Store;
@@ -39,6 +40,7 @@ import co.cask.cdap.common.InvalidArtifactException;
 import co.cask.cdap.common.NotFoundException;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.io.Locations;
 import co.cask.cdap.common.namespace.NamespacedLocationFactory;
 import co.cask.cdap.config.PreferencesStore;
 import co.cask.cdap.data2.registry.UsageRegistry;
@@ -54,6 +56,7 @@ import co.cask.cdap.internal.app.runtime.artifact.WriteConflictException;
 import co.cask.cdap.internal.app.runtime.flow.FlowUtils;
 import co.cask.cdap.internal.app.runtime.schedule.Scheduler;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.NamespaceMeta;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.ProgramTypes;
 import co.cask.cdap.proto.artifact.AppRequest;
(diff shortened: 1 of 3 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2621

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -361,6 +374,101 @@ public boolean apply(Id.Program programId) {
...
+              "You will need to re-deploy the app after upgrade.",
+            appId.getNamespaceId(), appId.getId()));
+          continue;
+        }
+
+        // copy jar to local file
+        File tmpFile = File.createTempFile("tmpApp", ".jar", tmpDir);
+        Files.copy(Locations.newInputSupplier(appJarLocation), tmpFile);
+
+        // use app name as artifact name
```
gokulavasan: May be move this comment to line 427?

gokulavasan:
Minor comment. LGTM :+1:

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/ApplicationLifecycleServiceTest.java
```diff
@@ -0,0 +1,77 @@
...
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, appSpec.getName());
+    // write an app spec without an artifact id
+    store.addApplication(appId, appSpec, appJar);
+    appSpec = store.getApplication(appId);
+    Assert.assertNull(appSpec.getArtifactId());
+
+    // run upgrade
+    applicationLifecycleService.upgrade(false);
+
+    // app sepc should now have an artifact id
```
sreevatsanraman: typo.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -361,6 +374,101 @@ public boolean apply(Id.Program programId) {
...
+          if (version == null || version.isEmpty()) {
+            version = "1.0.0";
+          }
+        }
+
+        ArtifactDetail artifactDetail;
+        Id.Artifact artifactId = Id.Artifact.from(namespaceId, appId.getId(), version);
+        try {
+          artifactDetail = artifactRepository.addArtifact(artifactId, tmpFile);
+        } catch (WriteConflictException e) {
```
sreevatsanraman: This will happen if we run the upgrade tool twice. What would be the impact of that? Will this upgrade operation be idempotent?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/ApplicationLifecycleServiceTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @BeforeClass
+  public static void setup() throws Exception {
+    applicationLifecycleService = getInjector().getInstance(ApplicationLifecycleService.class);
+    store = getInjector().getInstance(DefaultStore.class);
+    locationFactory = getInjector().getInstance(LocationFactory.class);
+  }
+
+  @Test
+  public void testUpgrade() throws Exception {
```
sreevatsanraman: Can we add a test case for running upgrade tool twice? make sure the operation is idempotent.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
albertshau: this is really more of a transaction failure, will only happen if a writer repeatedly fails to write due to another writer making concurrent changes. should not happen during upgrade.

albertshau: the scenario you mention is captured in the catch below for ArtifactAlreadyExists. In that scenario, it just looks up the info from the store. It is only possible if the upgrade tool ran before, the artifact was added, but the app metadata failed to update.  It is idempotent

albertshau:
@sreevatsanraman addressed your comments can you take another look?

sreevatsanraman:
LGTM

PR #3707, opened 2015-08-26
Fixes differences between old and new plugin templates json structure +  makes pluginTemplates fetch Sync(from cache)

ajainarayanan:
- Fixes plugin templates difference in old and new JSON structures (state params changes)
- Fixes fetching plugin templates from user store in sync with clicking on left side panel. (There was a timing issue between plugin templates fetch and opening of default source tab when we load adapter create view)
  For Instance:  
     If we have a Source template of name "Database1" it wouldn't appear when we load adapter create view because we are fetching the plugin templates asynchronously with loading actual plugins. (```savedTemplates``` might not be updated before we got the list of plugins from backend)

### DIFF
--- a/cdap-ui/app/features/admin/templates/namespace/templates-list.html
+++ b/cdap-ui/app/features/admin/templates/namespace/templates-list.html
@@ -24,21 +24,15 @@
     <tbody>
       <tr ng-repeat="template in TemplatesListController.list">
         <td>
-          <a ng-if="template.pluginTemplate"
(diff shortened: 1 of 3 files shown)


cdap-ui/app/features/admin/templates/namespace/templates-list.html
```diff
@@ -32,7 +32,7 @@
           <i class="icon-{{template.templateType}}"></i>
           <span>{{ template.templateType }}</span>
         </td>
-        <td>{{ template.pluginType }}</td>
+        <td>{{ template.pluginType || template.type }}</td>
```
elfenheart: not required

elfenheart:
:+1: LGTM

PR #3708, opened 2015-08-26
need the quotes to force string comparison

duddun:
Porting (merged) CDAP-3437 to 3.1, 3.0, 2.8.
This only affects build artifacts upload.


wolf31o2:
:+1: LGTM

PR #3711, opened 2015-08-26
Some improvments/fixes while looking at the stream hive related classes

chtyim:


cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
```diff
@@ -194,7 +183,7 @@ public InputSplit createSplit(Path path, Path indexPath, long startTime, long en
 
     private final String opClassName;
 
-    private CompareOp(String opClassName) {
+    CompareOp(String opClassName) {
```
jwang47: Why default? Doesn't look like it's necessary.

albertshau:
:+1:

cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
chtyim: IntelliJ have a warning for having private, since enum constructor is always private, hence modifier is unnecessary.

jwang47:
LGTM

chtyim:
Build passed: https://builds.cask.co/browse/CDAP-DRC2470-1

PR #3712, opened 2015-08-26
CDAP-3508 moving contexts to a place where they can be shared

albertshau:
This lets other apps create default contexts


gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-DUT2627

PR #3713, opened 2015-08-26
Use correct SDK version for Docker

wolf31o2:
- Creates packer/files/cdap-sdk.json
- Copies cdap-sdk.json to /tmp
- Uses cdap-sdk.json during cdap::sdk Chef run

This will be ported to other branches (3.1 and develop) with the appropriate versions and the builds will be updated to modify this version with the version from Maven.


cdap-distributions/src/packer/files/cdap-sdk.json
```diff
@@ -0,0 +1,5 @@
+{
+  "cdap": {
+    "version": "3.0.4-1"
```
awholegunch: Will this version appear in the pom.xml?

awholegunch:
Minor comment, otherwise LGTM.

cdap-distributions/src/packer/files/cdap-sdk.json
wolf31o2: This has nothing to do with Maven.

PR #3714, opened 2015-08-26
Remove PFS Upgrader

anwar6953:
Removing upgrader for PartitionedFileSets.
This upgrade was needed when user upgrades to 3.1; no changes are planned for PFS that requires upgrade after 3.1.

http://builds.cask.co/browse/CDAP-DUT2629-2


anew:
We are still assuming that the upgrade tool is only between two consecutive versions? In that case: LGTM

anwar6953:
That was also my assumption - that we do not support upgrades across consecutive versions.
That makes the upgrade tool easier to develop, but we should verify this assumption with others.

This assumption is consistent with the documentation, so far.
http://docs.cask.co/cdap/3.1.1/en/admin-manual/installation/installation.html?highlight=upgrade

"These steps will upgrade from CDAP 3.0.3 to 3.1.0. If you are on an earlier version of CDAP, please follow the upgrade instructions for the earlier versions and upgrade first to 3.0.3 before proceeding. (Note: Some apps need to be both recompiled and re-deployed, see below.)"

anwar6953:
The user can upgrade to 3.1.1 before proceeding also (its not a requirement that they must be on 3.1.0 in order to upgrade to 3.2.x).

The instructions to upgrade to 3.1.1 would be on the docs for 3.1.1, right?

anew:
I thought we can always upgrade from 3.N.x to 3.N+1.y (from any minor-minor version of the previous release to any minor-minor version of the current release).

anwar6953:
Yea, so I changed the documentation to use 3.1.x instead of 3.1.0

awholegunch:
LGTM

PR #3716, opened 2015-08-26
[CDAP-3494] Feature/ui redesign explore

elfenheart:
cluster: newsinglenode2671-1000

newexplore


cdap-ui/app/directives/explore/explore.js
```diff
@@ -0,0 +1,155 @@
...
+              isFirstTime = false;
+
+              vm.queries = queries;
+
+              // Polling for status
+              angular.forEach(vm.queries, function(q) {
+                q.isOpen = false;
+                if (q.status !== 'FINISHED') {
+
+                  // TODO: change to use myExploreApi once figure out how to manually stop poll with $resource
```
ajainarayanan: Is this still valid? We can stop a poll using ```$resource``` right?

cdap-ui/app/directives/explore/explore.js
```diff
@@ -0,0 +1,155 @@
...
+        vm.getQueries();
+
+        vm.preview = function (query) {
+          $bootstrapModal.open({
+            templateUrl: 'explore/preview-modal.html',
+            size: 'lg',
+            resolve: {
+              query: function () { return query; }
+            },
+            controller: ['$scope', 'myExploreApi', function ($scope, myExploreApi) {
```
ajainarayanan: We resolve ```query``` but not using it in the controller?

cdap-ui/app/directives/explore/explore.js
```diff
@@ -0,0 +1,155 @@
...
+          myExploreApi.postQuery(params, { query: vm.query }, vm.getQueries);
+          vm.currentPage = 1;
+        };
+
+        vm.getQueries = function() {
+
+          myExploreApi.getQueries(params)
+            .$promise
+            .then(function (queries) {
+              if (!isFirstTime) {
```
ajainarayanan: Not sure if we require ```isFirstTime``` and the ```if``` block. Just the ```$timeout``` block should be fine right?

elfenheart: The if block will prevent all the queries being green the first time user loads the page.

ajainarayanan:
:+1: LGTM

PR #3719, opened 2015-08-27
(CDAP-20) Deprecate public API that uses Multimap.

chtyim:


cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpServiceRequest.java
```diff
@@ -43,10 +45,17 @@
...
    * @return the headers of this request; each header name can map to multiple values
+   * @deprecated Use {@link #getAllHeaders()} instead. This method will be removed in future release.
    */
+  @Deprecated
   Multimap<String, String> getHeaders();
 
   /**
+   * @return all headers of this request; each header name can map to multiple values
+   */
+  Map<String, List<String>> getAllHeaders();
```
bdmogal: nit: Should we add a method description in this javadoc, other than the description of ``@return``, or do you think that is unnecessary?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/DelayedHttpServiceResponder.java
```diff
@@ -58,109 +62,74 @@ public DelayedHttpServiceResponder(HttpResponder responder, MetricsContext metri
...
-   * Sends a string response back to the HTTP client.
-   *
-   * @param status the status of the HTTP response
-   * @param data the data to be sent back
-   * @param charset the Charset used to encode the string
-   */
   @Override
   public void sendString(int status, String data, Charset charset) {
-    send(status, charset.encode(data), "text/plain; charset=" + charset.name(), null, false);
+    doSend(status, charset.encode(data), "text/plain; charset=" + charset.name(), null, false);
```
bdmogal: Should we send empty map for headers here too, instead of ``null``?

bdmogal:
Couple of questions, LGTM otherwise

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpServiceRequest.java
chtyim: Like `Return all headers of this request.`? It's a bit repetitive and it's not like that in other methods.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/DelayedHttpServiceResponder.java
chtyim: Either is fine. It's a private method and the existing logic already handle null.

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpServiceRequest.java
bdmogal: +1

bdmogal:
LGTM!

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2451-8

PR #3721, opened 2015-08-27
modifies services layout per new program mocks

astewes:
*  Reorganizes the Services feature to accommodate new layout


ajainarayanan:
:+1: LGTM

PR #3723, opened 2015-08-27
integrate package bundling with artifact upload process

duddun:
We already have an RPM and Debian package bundling script that we run manually (as needed) on the docs server.  This replaces that manual process.


cdap-distributions/bin/build_artifact_upload.sh
```diff
@@ -71,11 +76,17 @@ function sync_build_artifacts_to_server () {
...
+  clean
+
   decho "identify packages"
   PACKAGES=$(find ${_source} -type f \( -name '*.rpm' -o -name '*.deb' \) | sort -u)
   decho "${PACKAGES}"
   decho ""
 
+  # prepare for bundling
+  mkdir -p ${PKG_PROCESSING_DIR}
+  for p in ${PACKAGES}; do cp ${p} ${PKG_PROCESSING_DIR}; done
```
wolf31o2: Why do you need this loop? You're copying multiple files to the same destination directory.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+# limitations under the License.
+#
+######################################################################################
+# Create RPM and Debian bundles and upload them directly to docs1
+# This script expects to receive 3 parameters in this order:
+#   $1 BUILD_PACKAGE    e.g. cdap
+#   $2 VERSION    e.g. 3.0.4
+#   $3 USER                     REMOTE_USER
+#   $4 remote host              REMOTE_HOST
+#   $5 target directory         REMOTE_INCOMING_DIR
```
wolf31o2: It's 5 parameters...

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+######################################################################################
+DEBUG=${DEBUG:-no}              ### set to yes for debugging
+RUN_DATE=`date '+%Y%m%d_%R'`
+SCRIPT=`basename ${BASH_SOURCE[0]}`                     ### Set Script Name variable
+BUILD_PACKAGE="${1:-cdap}"                                    ## e.g. cdap ## a.k.a. BUILD_PACKAGE
```
wolf31o2: This cannot have a default, as `$1` is required. Also, there's no need for quotes.

wolf31o2: Plus, this is in the CDAP repository, do it'll only ever be `cdap`... :smile: 

Since this is coming from local disk, there's no reason why we need to specify project and version. The project is always `cdap` and the version is whatever's on the disk.

```
VERSION=$(<cdap-distributions/target/stage-packaging/opt/cdap/distributions/VERSION)
```

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+    echo ${*}
+  else
+    RSYNC_QUIET='--quiet'
+  fi
+}
+
+# Help function
+function HELP {
+  echo -e \\n"Help documentation for ${BOLD}${SCRIPT}.${NORM}"\\n
+  echo -e "${REV}Basic usage:${NORM} ${BOLD}$SCRIPT${NORM} <remote_user> <remote_host> <remote_target_directory"\\n
```
wolf31o2: Incorrect arguments... there's 5, not 3...

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+DEB_BUNDLE_TGZ="${DISTRIBUTED_DEB_BUNDLE}-${VERSION}.tgz"
+RPM_BUNDLE_TGZ="${DISTRIBUTED_RPM_BUNDLE}-${VERSION}.tgz"
+DEB_BUNDLE_TGZ_PATH="${DEB_BUNDLE_DIR}/${DEB_BUNDLE_TGZ}"
+RPM_BUNDLE_TGZ_PATH="${RPM_BUNDLE_DIR}/${RPM_BUNDLE_TGZ}"
+S3STG_INCOMING="${S3STG_DIR}/incoming"
+S3STG_INCOMING_PROJ="${S3STG_INCOMING}/${BUILD_PACKAGE}"
+STG_DIR="${S3STG_INCOMING}/downloads/co/cask/${BUILD_PACKAGE}"
+STG_DEB_DIR="${STG_DIR}/${DISTRIBUTED_DEB_BUNDLE}/${VERSION}"
+STG_RPM_DIR="${STG_DIR}/${DISTRIBUTED_RPM_BUNDLE}/${VERSION}"
+echo ${STG_RPM_DIR}
```
wolf31o2: Why is this echo here?

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+
+#########################################################################
+create_bundles() {
+
+   ### PREP
+   # make sure bundle directories exist
+   mkdir -p ${STG_DIR} || die "Unable to create STG_DIR"
+   mkdir -p ${DEB_BUNDLE_DIR} ${RPM_BUNDLE_DIR} || die "Unable to create bundle processing directories"
+   mkdir -p ${STG_DEB_DIR} ${STG_RPM_DIR} || die "Unable to create bundle staging directories"
+   OUTGOING_DIR=${BUILD_PACKAGE}/${_version}
```
wolf31o2: Where does `${_version}` come from? Should this be `${VERSION}`?

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+   echo ${OUTGOING_DIR}
+   
+   ### FIND ARTIFACTS
+   mkdir -p ${S3STG_INCOMING_PROJ} || die "Unable to create project staging directory"
+   cd ${S3STG_INCOMING_PROJ} || die "Unable to cd to ${S3STG_INCOMING_PROJ} directory"
```
wolf31o2: Why do we create this directory?

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+   find . -name "${BUILD_PACKAGE}-*${VERSION}-*.deb" -exec cp '{}' ${DEB_BUNDLE_DIR}/. \; || die "Unable to copy Debian ${VERSION} packages to TMP directory"
+   find . -name "${BUILD_PACKAGE}-*${VERSION}-*.rpm" -exec cp '{}' ${RPM_BUNDLE_DIR}/. \; || die "Unable to copy ${VERSION} RPMs to TMP directory"
```
wolf31o2: These will copy nothing. The directory is empty.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+   
+   # create and verify debian tarball
+   cd ${DEB_BUNDLE_DIR} || die "Unable to cd to ${DEB_BUNDLE_DIR} directory"
+   tar czvf ${DEB_BUNDLE_TGZ} * || die "Unable to create Debian tar bundle"
```
wolf31o2: Current bundles have a directory in them...

```
$ tar tf ~/Downloads/cdap-distributed-rpm-bundle-3.1.1.tgz
cdap-distributed-rpm-bundle-3.1.1/
cdap-distributed-rpm-bundle-3.1.1/cdap-master-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-cli-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-hbase-compat-0.96-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-hbase-compat-1.0-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-hbase-compat-1.0-cdh-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-ui-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-security-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-kafka-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-gateway-3.1.1-1.noarch.rpm
cdap-distributed-rpm-bundle-3.1.1/cdap-hbase-compat-0.98-3.1.1-1.noarch.rpm
```

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+   # create and verify RPM tarball
+   cd ${RPM_BUNDLE_DIR} || die "Unable to cd to ${RPM_BUNDLE_DIR} directory"
+   tar czvf ${RPM_BUNDLE_TGZ} * || die "Unable to create RPM tar bundle"
+   tar tzvf ${RPM_BUNDLE_TGZ} || die "Unable to show contents of RPM tar bundle"
+   
+   # SYNC bundle(s) to remote server
+   decho "rsyncing with rsync -av -e \"ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null\" ${RSYNC_QUIET} ${DEB_BUNDLE_TGZ_PATH} ${STG_DEB_DIR}/.  2>&1"
+   decho "rsyncing with rsync -av -e \"ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null\" ${RSYNC_QUIET} ${RPM_BUNDLE_TGZ_PATH} ${STG_RPM_DIR}/.  2>&1"
+   rsync -av -e "ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" ${RSYNC_QUIET} ${DEB_BUNDLE_TGZ_PATH} ${STG_DEB_DIR}/. 2>&1 || die "could not rsync ${DEB_BUNDLE_TGZ_PATH} as ${REMOTE_USER} to ${REMOTE_HOST}: ${!}"
+   rsync -av -e "ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" ${RSYNC_QUIET} ${RPM_BUNDLE_TGZ_PATH} ${STG_RPM_DIR}/. 2>&1 || die "could not rsync ${RPM_BUNDLE_TGZ_PATH} as ${REMOTE_USER} to ${REMOTE_HOST}: ${!}"
```
wolf31o2: We should separate create from sync, logically. Put this in its own function.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,136 @@
...
+# Check number of arguments. If <5 are passed, print help and exit.
+NUMARGS=$#
+decho -e \\n"Number of arguments: ${NUMARGS}"
+if [ ${NUMARGS} -lt 5 ]; then
+  HELP
+fi
+
+decho "#######################################################################################"
+echo "Creating RPM and Debian bundles"
+create_bundles
```
wolf31o2: Now that we've gotten here, `create_bundles` (and your new `sync_bundles` function) should both be simply included into `build_artifact_upload.sh` to DRY up the code and not repeat variable definitions, functions, etc.

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: typically they have not

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: done

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: Will do this later

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+# This script expects to receive 3 parameters in this order:
+#   $1 USER                     REMOTE_USER
+#   $2 remote host              REMOTE_HOST
+#   $3 target directory         REMOTE_INCOMING_DIR
+######################################################################################
+DEBUG=${DEBUG:-no}              ### set to yes for debugging
+RUN_DATE=`date '+%Y%m%d_%R'`
+SCRIPT=`basename ${BASH_SOURCE[0]}`                     ### Set Script Name variable
+BUILD_PACKAGE='cdap'                                   ## e.g. cdap ## a.k.a. BUILD_PACKAGE
+VERSION=$(cdap-distributions/target/stage-packaging/opt/cdap/distributions/VERSION)
```
wolf31o2: You forgot the `<`

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+REMOTE_USER=${1}                                        ### remote user
+REMOTE_HOST=${2:-127.0.0.1}                             ### remote host
+REMOTE_INCOMING_DIR=${3}                                ### target directory on remote host
+BUILD_PACKAGE=${BUILD_PACKAGE:-cdap}
```
wolf31o2: Just set this to `cdap` unless you expect this to be overridden somewhere. I don't see where we'll be using this outside of a CDAP-related build, though.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+REMOTE_BASE_DIR="incoming"
```
wolf31o2: Unused.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+USER=${USER:-bamboo}
```
wolf31o2: Unused, overrides a built-in shell variable, *and* defaults to `bamboo` when there should be no mention of Cask-infrastructure-specific items.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+TMP_DIR='/tmp/bundles'
+S3STG_DIR=${S3STG_DIR:-/data/s3stg} # change it #####
+DEB_BUNDLE="${BUILD_PACKAGE}-deb-bundle"
+RPM_BUNDLE="${BUILD_PACKAGE}-rpm-bundle"
+DEB_BUNDLE_DIR="${TMP_DIR}/${DEB_BUNDLE}"
+RPM_BUNDLE_DIR="${TMP_DIR}/${RPM_BUNDLE}"
```
wolf31o2: This ends up being way more confusing when read below. Most of these constants are redundant or mask the meaning enough that I find myself forced to constantly refer back to this list to know which variable means what.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+BUILD_PACKAGE='cdap'                                   ## e.g. cdap ## a.k.a. BUILD_PACKAGE
+VERSION=$(cdap-distributions/target/stage-packaging/opt/cdap/distributions/VERSION)
+REMOTE_USER=${1}                                        ### remote user
+REMOTE_HOST=${2:-127.0.0.1}                             ### remote host
+REMOTE_INCOMING_DIR=${3}                                ### target directory on remote host
+BUILD_PACKAGE=${BUILD_PACKAGE:-cdap}
+REMOTE_BASE_DIR="incoming"
+USER=${USER:-bamboo}
+TMP_DIR='/tmp/bundles'
+S3STG_DIR=${S3STG_DIR:-/data/s3stg} # change it #####
```
wolf31o2: This is only used for S3STG_INCOMING below.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+DEB_BUNDLE_DIR="${TMP_DIR}/${DEB_BUNDLE}"
+RPM_BUNDLE_DIR="${TMP_DIR}/${RPM_BUNDLE}"
+DISTRIBUTED_DEB_BUNDLE="${BUILD_PACKAGE}-distributed-deb-bundle"
+DISTRIBUTED_RPM_BUNDLE="${BUILD_PACKAGE}-distributed-rpm-bundle"
+DEB_BUNDLE_TGZ="${DISTRIBUTED_DEB_BUNDLE}-${VERSION}.tgz"
+RPM_BUNDLE_TGZ="${DISTRIBUTED_RPM_BUNDLE}-${VERSION}.tgz"
+DEB_BUNDLE_TGZ_PATH="${DEB_BUNDLE_DIR}/${DEB_BUNDLE_TGZ}"
+RPM_BUNDLE_TGZ_PATH="${RPM_BUNDLE_DIR}/${RPM_BUNDLE_TGZ}"
+S3STG_INCOMING="${S3STG_DIR}/incoming"
+S3STG_INCOMING_PROJ="${S3STG_INCOMING}/${BUILD_PACKAGE}"
```
wolf31o2: We use this in one place only to cd to run a `find .` on an empty directory.

cdap-distributions/bin/prep_and_stage_bundles.sh
wolf31o2: Avoid ANSI code changes. They're horrible on text-only terminals that don't support bold/reverse.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+
+die ( ) { echo ; echo "ERROR: ${*}" ; echo ; exit 1; }
+
+######################################################################################
+
+clean() { test -d ${TMP_DIR} && rm -rf ${TMP_DIR}; };
+
+if [ $# -lt 2 ]; then
+  die "You need to pass project and version to ${_script}"$'\n\n'"Example: ./${_script} cdap 3.0.0"
+fi
```
wolf31o2: We check for two arguments here, and three below. Also, `_script` is undefined and this example has incorrect usage.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+
+#########################################################################
+create_bundles() {
+
+   ### PREP
+   # make sure bundle directories exist
+   mkdir -p ${STG_DIR} || die "Unable to create STG_DIR"
+   mkdir -p ${DEB_BUNDLE_DIR} ${RPM_BUNDLE_DIR} || die "Unable to create bundle processing directories"
+   mkdir -p ${STG_DEB_DIR} ${STG_RPM_DIR} || die "Unable to create bundle staging directories"
```
wolf31o2: I don't understand the creation of these directories. Also, creating `STG_DIR` is redundant with this line, since they're children of `STG_DIR` and are created with `mkdir -p` which will create any required parents.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+   OUTGOING_DIR=${BUILD_PACKAGE}/${VERSION}
+   echo ${OUTGOING_DIR}
```
wolf31o2: decho

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+   
+   ### FIND ARTIFACTS
+   cd ${S3STG_INCOMING_PROJ} || die "Unable to cd to ${S3STG_INCOMING_PROJ} directory"
+   find . -name "${BUILD_PACKAGE}-*${VERSION}-*.deb" -exec cp '{}' ${DEB_BUNDLE_DIR}/. \; || die "Unable to copy Debian ${VERSION} packages to TMP directory"
```
wolf31o2: Debian packages use underscores between name and version.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+   ### PREP
+   # make sure bundle directories exist
+   mkdir -p ${STG_DIR} || die "Unable to create STG_DIR"
+   mkdir -p ${DEB_BUNDLE_DIR} ${RPM_BUNDLE_DIR} || die "Unable to create bundle processing directories"
+   mkdir -p ${STG_DEB_DIR} ${STG_RPM_DIR} || die "Unable to create bundle staging directories"
+   OUTGOING_DIR=${BUILD_PACKAGE}/${VERSION}
+   echo ${OUTGOING_DIR}
+   
+   ### FIND ARTIFACTS
+   cd ${S3STG_INCOMING_PROJ} || die "Unable to cd to ${S3STG_INCOMING_PROJ} directory"
```
wolf31o2: The Maven build does not put packages in this location.

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+   find . -name "${BUILD_PACKAGE}-*${VERSION}-*.deb" -exec cp '{}' ${DEB_BUNDLE_DIR}/. \; || die "Unable to copy Debian ${VERSION} packages to TMP directory"
+   find . -name "${BUILD_PACKAGE}-*${VERSION}-*.rpm" -exec cp '{}' ${RPM_BUNDLE_DIR}/. \; || die "Unable to copy ${VERSION} RPMs to TMP directory"
```
wolf31o2: Why are you using `find .` from a subdirectory rather than the top-level directory? Even better, don't use `find` at all. It's a waste of time and energy when you have a known filename pattern in a known set of directory locations.

From the top of the cdap checkout:
```
cp -f */target/${BUILD_PACKAGE}-*${VERSION}-*.rpm ${RPM_BUNDLE_DIR}
```

Or even simply:
```
cp -f */target/*.rpm ${RPM_BUNDLE_DIR}
```

cdap-distributions/bin/prep_and_stage_bundles.sh
wolf31o2: OK. Even easier, then. You can just use `tar` without even staging anything...

```
tar czf ${DEB_BUNDLE_TGZ} --strip-components 2 */target/*.deb
tar czf ${RPM_BUNDLE_TGZ} --strip-components 2 */target/*.deb
```

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+   
+   # create and verify debian tarball
+   cd ${DEB_BUNDLE_DIR} || die "Unable to cd to ${DEB_BUNDLE_DIR} directory"
+   tar czvf ${DEB_BUNDLE_TGZ} * || die "Unable to create Debian tar bundle"
+   tar tzvf ${DEB_BUNDLE_TGZ} || die "Unable to show contents of Debian tar bundle"
```
wolf31o2: This should only be done during DEBUG, especially if you're using `v` in the create. Also, `v` and `t` are redundant.

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: It still find them because there is a hyphen in a different place
Will modify it to use _ -- much more precise

cdap-distributions/bin/prep_and_stage_bundles.sh
```diff
@@ -0,0 +1,138 @@
...
+   # create and verify RPM tarball
+   cd ${RPM_BUNDLE_DIR} || die "Unable to cd to ${RPM_BUNDLE_DIR} directory"
+   tar czvf ${RPM_BUNDLE_TGZ} * || die "Unable to create RPM tar bundle"
+   tar tzvf ${RPM_BUNDLE_TGZ} || die "Unable to show contents of RPM tar bundle"
+}
+
+sync_bundles() {
+   # SYNC bundle(s) to remote server
+   decho "rsyncing with rsync -av -e \"ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null\" ${RSYNC_QUIET} ${DEB_BUNDLE_TGZ_PATH} ${STG_DEB_DIR}/.  2>&1"
+   decho "rsyncing with rsync -av -e \"ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null\" ${RSYNC_QUIET} ${RPM_BUNDLE_TGZ_PATH} ${STG_RPM_DIR}/.  2>&1"
```
wolf31o2: Use single quotes on the outside and don't escape the internal quotes.

cdap-distributions/bin/prep_and_stage_bundles.sh
wolf31o2: This isn't a complex ask. There's no reason to add another script, with lots of duplicated constants. These scripts are already included in our packages (which is actually a whole different problem). If you still think this shouldn't be done, then you should file a JIRA so this work doesn't get lost.

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: For the same reason why we have had to add the logic to build_artifact_upload.sh to make sure the right artifacts get selected and the wrong ones do not. Did not want to repeat everything.
The goal of this was quick and dirty. grab the essence of what we had in two working scripts. Quickly integrate them and use that until we have a much more robust solution (coming soon).

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

cdap-distributions/bin/prep_and_stage_bundles.sh
duddun: fixed

wolf31o2:
:+1: LGTM

wolf31o2:
You'll want to also copy this to `release/3.0` and `release/3.1` branches, too.

PR #3725, opened 2015-08-27
CDAP-3488 create etl batch application

albertshau:
Creating the application version of ETLBatch. This will replace
the ETLBatch ApplicationTemplate, but we can't remove the template
until the UI is moved off templates and adapters.


albertshau:
looks like a lot, but it is mostly copies of test cases since we can't remove yet

gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-DUT2638

PR #3729, opened 2015-08-28
Use correct SDK version for Docker (3.1 edition)

wolf31o2:
This is the same as #3713 for `release/3.1` branch...


dereklwood:
:+1:

PR #3730, opened 2015-08-28
Use correct SDK version for Docker (develop)

wolf31o2:
This is the same as #3713 for release/3.1 branch...

The template is used during CI to replace the JSON file with a modified version with the current build version in it.


dereklwood:
:+1:

PR #3731, opened 2015-08-28
Fixed two MR bugs:

chtyim:
- In MapReduceContextProvider, we should always get the MapReduceClassLoader from hConf, not Confer
- Do proper resource cleanup for the artifact plugin instantiator

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
@@ -65,7 +65,10 @@ import org.apache.twill.internal.ServiceListenerAdapter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.Closeable;
 import java.security.PrivilegedExceptionAction;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


gokulavasan:
LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
```diff
@@ -249,6 +251,7 @@ public void starting() {
 
       @Override
       public void terminated(Service.State from) {
+        Closeables.closeQuietly(artifactPluginInstantiator);
```
gokulavasan: @chtyim Looks like we also close it in two more places in this class.

gokulavasan:
One comment. LGTM otherwise!

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceProgramRunner.java
chtyim: Good catch.

gokulavasan:
LGTM :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2644-3

PR #3732, opened 2015-08-28
CDAP-2965: Update docs to reflect new cdap-client API

jwang47:
https://issues.cask.co/browse/CDAP-2965
http://builds.cask.co/browse/CDAP-DUT2640

### DIFF
--- a/cdap-docs/reference-manual/source/java-client-api.rst
+++ b/cdap-docs/reference-manual/source/java-client-api.rst
@@ -92,7 +92,9 @@ Then, instantiate as follows::
   AccessToken accessToken = authenticationClient.getAccessToken();
 
   // Interact with the secure CDAP instance located at example.com, port 10000, with the provided accessToken
-  ClientConfig clientConfig = new ClientConfig("example.com", 10000, accessToken);
+  ClientConfig clientConfig = ClientConfig.builder()
+    .setConnectionConfig(new ConnectionConfig("example.com", 10000, accessToken))
+    .build();
 
 .. _application-client:
 
@@ -106,17 +108,17 @@ ApplicationClient
   ApplicationClient appClient = new ApplicationClient(clientConfig);
 
   // Fetch the list of applications
-  List<ApplicationRecord> apps = appClient.list();
+  List<ApplicationRecord> apps = appClient.list(Id.Namespace.DEFAULT);
 
   // Deploy an application
-  File appJarFile = ...;
-  appClient.deploy(appJarFile);
+  File appJarFile = new File("your-app.jar");
+  appClient.deploy(Id.Namespace.DEFAULT, appJarFile);
 
   // Delete an application
-  appClient.delete("Purchase");
+  appClient.delete(Id.Application.from(Id.Namespace.DEFAULT, "Purchase"));
 
   // List programs belonging to an application
-  appClient.listPrograms("Purchase");
+  appClient.listPrograms(Id.Application.from(Id.Namespace.DEFAULT, "Purchase"));
 
 .. _preferences-client:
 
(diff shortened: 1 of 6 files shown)


cdap-docgen/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+      <groupId>com.google.inject</groupId>
+      <artifactId>guice</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.javaparser</groupId>
+      <artifactId>javaparser</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
```
anwar6953: missing newline

cdap-docs/reference-manual/source/java-client-api.rst
```diff
@@ -335,39 +327,20 @@ QueryClient
...
-      header[index] = column.getName() + ": " + column.getType();
-    }
+  ListenableFuture<ExploreExecutionResult> resultFuture = queryClient.execute(Id.Namespace.DEFAULT, "SELECT * FROM dataset_history WHERE customer IN ('Alice','Bob')");
+  ExploreExecutionResult results = resultFuture.get();
+
+  // Fetch schema
+  List<ColumnDesc> schema = results.getResultSchema();
+  String[] header = new String[schema.size()];
+  for (int i = 0; i < header.length; i++) {
+  ColumnDesc column = schema.get(i);
```
anwar6953: indentation would have to be preserved in the same way as `nl()`?

jwang47: Indentation isn't preserved unfortunately. But yes, we can do begin/endIndent().

cdap-docs/reference-manual/source/java-client-api.rst
```diff
@@ -272,19 +265,19 @@ DatasetClient
...
-  datasetClient.truncate("someDataset");
+  datasetClient.truncate(datasetId);
 
   // Delete a dataset
-  datasetClient.delete("someDataset");
+  datasetClient.delete(datasetId);
 
-
-.. _dataset-module-client:
+.. _datasetmodule-client:
```
awholegunch: Why is this changing? ``dataset-module-client`` is correct.

cdap-docs/reference-manual/source/java-client-api.rst
```diff
@@ -296,17 +289,17 @@ DatasetModuleClient
...
-  DatasetModuleMeta datasetModuleMeta = datasetModuleClient.get("someDatasetModule");
+  DatasetModuleMeta datasetModuleMeta = datasetModuleClient.get(datasetModuleId);
 
   // Delete all dataset modules
-  datasetModuleClient.deleteAll();
-
+  datasetModuleClient.deleteAll(Id.Namespace.DEFAULT);
 
-.. _dataset-type-client:
+.. _datasettype-client:
```
awholegunch: ibid.

awholegunch: If these change, the links to them break.

awholegunch:
I don't quite understand the point of this. Using Java to generate rst seems very heavy-weight. As you see, the formatting is hard to get right. What about using Python? That's what we are doing elsewhere in the build system. Plus, it looks fragile and difficult to maintain.

Another thing to think about is that instead of generating and inserting into the main rst file, you generate a portion of it. That portion is written to a separate file (can be rst, txt, anything) and is then included.
Then you don't have to worry about changes to your file, and potentially the generation can be part of the build. But given how long it takes a JVM to spin up at times...

awholegunch:
Also, as we have a ``cdap-docs`` module; what about using ``cdap-docs-gen``?

awholegunch:
The more I read this...I can see having Java code to build and test...part of the maven build...and then run through a simple Python script that grabs the correct portions into an rst file. Much simpler, easy to integrate, even I could maintain it :)

jwang47:
True, the code that reads the .java and generates the .rst is very hacky here. I think the .java that contains the sample usage is useful, but we should definitely find a better way to generate the .rst.

awholegunch:
Here's a proposal then:
- change the module to ``cdap-docs-gen``, to keep with existing names.
- revise the ``GenerateClientUsageExample.java`` to remove all the hacks added
- perhaps rename it ``ClientUsageExample.java``
- remove all the code generating the rst in Java
- add a test case to check the ``ClientUsageExample.java``

Basically, just get the Java file ready. I will in another PR add a python script that will be called by the build system to create and link up the .rst. Make sense?

jwang47:
Sounds good.

PR #3733, opened 2015-08-28
Exclude JSON template from RAT

wolf31o2:
JSON doesn't allow comments, so there's no license header.


dereklwood:
:+1:

chtyim:
The exclusion is wrong. Develop is still broken. See 109a9971311f5056d3c59837290dc14f63f147b2

PR #3735, opened 2015-08-28
Exclude JSON template from RAT

wolf31o2:
Cherry pick #3733 to release/3.1


chtyim:
The exclusion is wrong. Develop is still broken. See 109a997

PR #3737, opened 2015-08-28
Use MockAppConfigurer instead of DefaultAppConfigurer to get the name of the App

gokulavasan:
MockConfigurer is required so that we don't have to pass in ArtifactRepository/PluginInstantiator. Moreover it is logically not possible since the artifact itself is not available and thus instantiating plugins before that won't be possible. Hence deployApplication method itself won't work for applications using plugins. Users will have to resort to using addArtifact, addPluginArtifact, addApplication route

Also, Close the ArtifactPluginInstantiator in Worker, ServiceHandler

Build : http://builds.cask.co/browse/CDAP-DUT2634


gokulavasan:
@albertshau Please review

albertshau:
:+1:

PR #3738, opened 2015-08-28
CDAP-3489 creating etl realtime app

albertshau:
This is the Application version of the current etl realtime
ApplicationTemplate. This will replace the template once the UI
has moved over to apps instead of adapters. Also included some
refactoring so that etl mapreduce and worker can share logic for
registering pipeline plugins.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2649

albertshau:
a lot of copying again... will be nice once we can remove templates

gokulavasan:
LGTM :+1:

PR #3740, opened 2015-08-28
Multiple Output Datasets

anwar6953:
This PR allows support for MR job to write to multiple output Datasets, according to the following design document: https://wiki.cask.co/display/CE/Multiple+Output+Datasets+of+MapReduce

In this PR, there are two test MapReduce jobs which utilize the new API; I am working on additional ones.

https://issues.cask.co/browse/CDAP-2756
http://builds.cask.co/browse/CDAP-DUT2564-13


cdap-api-hadoop/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-api</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-mapreduce-client-core</artifactId>
+      <version>${hadoop.version}</version>
```
chtyim: Shouldn't have these. Should be using whatever being specified in the parent pom through the dependency management.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractMapper.java
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.mapreduce;
+
+import co.cask.cdap.api.ProgramLifecycle;
+import com.google.common.io.Closeables;
```
chtyim: Try not to use guava in api.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractMapper.java
```diff
@@ -0,0 +1,66 @@
...
+  private MapReduceTaskContext<KEYIN, VALUEIN, KEYOUT, VALUEOUT> context;
+
+  @Override
+  public void initialize(MapReduceTaskContext context) throws Exception {
+    this.context = context;
+  }
+
+  @Override
+  public void destroy() {
+    Closeables.closeQuietly(context);
```
chtyim: What close the context here? In general, you only release resource that you created. Since context is passed to this class, it shouldn't be closed in here. It should be the call (which I believe is the MapperWrapper) to do the cleanup.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/MapReduceTaskContext.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * MapReduce task execution context. This is similar to MapReduceContext, but without ability to configure
+ * input/output/resources
+ *
+ * @param <KEYIN>    input key type
+ * @param <VALUEIN>  input value type
+ * @param <KEYOUT>   output key type
+ * @param <VALUEOUT> output value type
+ */
+public interface MapReduceTaskContext<KEYIN, VALUEIN, KEYOUT, VALUEOUT> extends RuntimeContext, DatasetContext,
```
chtyim: Shouldn't be a parameterized type.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/MapReduceTaskContext.java
```diff
@@ -0,0 +1,88 @@
...
+   */
+  <K, V> void write(String namedOutput, K key, V value) throws IOException, InterruptedException;
+
+  /**
+   * Write key and value to the hadoop context.
+   *
+   * @param key         the key
+   * @param value       the value
+   */
+  void write(KEYOUT key, VALUEOUT value) throws IOException, InterruptedException;
```
chtyim: Do it similar to the write method above.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/MapReduceTaskContext.java
```diff
@@ -0,0 +1,88 @@
...
+ * MapReduce task execution context. This is similar to MapReduceContext, but without ability to configure
+ * input/output/resources
+ *
+ * @param <KEYIN>    input key type
+ * @param <VALUEIN>  input value type
+ * @param <KEYOUT>   output key type
+ * @param <VALUEOUT> output value type
+ */
+public interface MapReduceTaskContext<KEYIN, VALUEIN, KEYOUT, VALUEOUT> extends RuntimeContext, DatasetContext,
+  ServiceDiscoverer, AdapterContext, PluginContext, Closeable {
```
chtyim: Shouldn't have `Closeable` in this interface.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/MapReduceTaskContext.java
```diff
@@ -0,0 +1,88 @@
...
+   * job runs.
+   *
+   * @return Time in milliseconds since epoch time (00:00:00 January 1, 1970 UTC).
+   */
+  long getLogicalStartTime();
+
+  /**
+   * Returns the Hadoop context object for the task.
+   */
+  TaskInputOutputContext<KEYIN, VALUEIN, KEYOUT, VALUEOUT> getHadoopcontext();
```
chtyim: `getHadoopContext`.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/MapReduceTaskContext.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * @return the {@link WorkflowToken} associated with the current {@link Workflow},
+   * if the {@link MapReduce} program is executed as a part of the Workflow.
```
chtyim: Document that it can returns null.

chtyim:
With the introduction of the new cdap-api-hadoop module, you need to adjust the class loading contract in the ProgramClassLoader.

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
```diff
@@ -117,6 +117,25 @@
...
+  /**
+   * Overrides the output configuration of this MapReduce job to also allow writing to the specified dataset instance.
+   * Currently, the dataset passed in must be an {@link OutputFormatProvider}.
+   * You may want to use this method instead of {@link #setOutput(String)} if your output dataset uses runtime
+   * arguments set in your own program logic.
+   *
+   * @param datasetName the name of the output dataset
+   * @param dataset the output dataset
+   */
+  void addOutput(String datasetName, Dataset dataset);
```
chtyim: Since this is a new method, we shouldn't take a `Dataset` instance, but a dataset argument instead (Map<String, String>)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
```diff
@@ -0,0 +1,158 @@
...
+  implements MapReduceTaskContext<KEYIN, VALUEIN, KEYOUT, VALUEOUT> {
+
+  private final BasicMapReduceContext basicMapReduceContext;
+  private TaskInputOutputContext hadoopContext;
+  private final MultipleOutputs multipleOutputs;
+
+  public BasicMapReduceTaskContext(BasicMapReduceContext basicMapReduceContext, TaskInputOutputContext hadoopContext) {
+    this.basicMapReduceContext = basicMapReduceContext;
+    this.hadoopContext = hadoopContext;
+    this.multipleOutputs = new MultipleOutputs<>(hadoopContext);
```
chtyim: Why there is `<>`, while the type is not a parameterized type?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
```diff
@@ -0,0 +1,293 @@
...
+import java.util.StringTokenizer;
+
+/**
+ * The MultipleOutputs class simplifies writing output data to multiple outputs.
+ * It has been adapted from org.apache.hadoop.mapreduce.lib.output.MultipleOutputs.
+ *
+ * @param <KEYOUT>   output key type
+ * @param <VALUEOUT> output value type
+ */
+public class MultipleOutputs<KEYOUT, VALUEOUT> {
```
chtyim: What the KEYOUT and VALUEOUT are for? Looks like none of the usage actually fit this. Also, isn't the MultipleOutputs allow writing different key type and value type?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputsMainOutputWrapper.java
```diff
@@ -0,0 +1,102 @@
...
+
+  public static void setRootOutputFormat(Job job,  String outputFormatClass, Map<String, String> outputConfig) {
+    job.getConfiguration().set(ROOT_OUTPUT_FORMAT, outputFormatClass);
+
+    for (Map.Entry<String, String> confEntry : outputConfig.entrySet()) {
+      job.getConfiguration().set(confEntry.getKey(), confEntry.getValue());
+    }
+  }
+
+  public OutputFormat<K, V> getRootOutputFormat(TaskAttemptContext context) {
```
chtyim: Why public? Need to pay more attention to encapsulation.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputsTest.java
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.internal.app.runtime.batch.dataset;
+
+
```
chtyim: Remove extra new line.

chtyim:
Did an initial quick scan. Please address those comments first.

anwar6953:
@chtyim Could you elaborate on this:
"With the introduction of the new cdap-api-hadoop module, you need to adjust the class loading contract in the ProgramClassLoader."

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractMapper.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /**
+   * Called once for each key/value pair in the input split. Most applications
+   * should override this, but the default is the identity function.
+   */
+  @SuppressWarnings("unchecked")
```
anew: But this only works if KEYIN-KEYOUT and same for the value. Can't we leave this as an abstract method?

chtyim:
Meaning need to change ProgramClassLoader as well. Is it possible to delay the introduction of AbstractMapper/Reducer to later PR? All you need is actually the Mapper/ReducerTaskContext being exposed in api, which doesn't have dependency on Hadoop, right?

It's better to use a separate, smaller PR for the introduction of the cdap-api-hadoop module, since it requires changes to ProgramClassLoader (and related classes) to provide a proper class loading model for CDAP programs.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractMapper.java
```diff
@@ -0,0 +1,66 @@
...
+  public void map(KEYIN key, VALUEIN value,
+                  MapReduceTaskContext context) throws IOException, InterruptedException {
+    context.write(key, value);
```
anew: We are losing some type-safeness here. Mapper.Context is parameterized with KEYOUT and VALUEOUT, so you can only write correctly typed output. MapReduceTaskContext gives that up and type errors will only be found at runtime. I don't think that is good.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractReducer.java
```diff
@@ -0,0 +1,70 @@
...
+ * Abstract Reducer class which provides functionality to a MapReduceTaskContext object, which allows writing to
+ * multiple outputs.
+ *
+ * @param <KEYIN>    input key type
+ * @param <VALUEIN>  input value type
+ * @param <KEYOUT>   output key type
+ * @param <VALUEOUT> output value type
+ */
+public abstract class AbstractReducer<KEYIN, VALUEIN, KEYOUT, VALUEOUT>
+  extends Reducer<KEYIN, VALUEIN, KEYOUT, VALUEOUT>
```
anew: I guess all comments for AbstractMapper also apply to AbstractReducer.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractMapper.java
anwar6953: Terence had suggested to remove the type parameterization of our TaskContext object:
https://github.com/caskdata/cdap/pull/3740#discussion_r38240572

This makes sense, because it supports writing keys and values of different types (since it supports multiple outputs).

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractMapper.java
anwar6953: This will work even if KEYIN != KEYOUT, since our MapReduceTaskContext object isn't type-parameterized.
We can make it an abstract method if there's another need for it, though.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractMapper.java
chtyim: We have to be careful about what the contract of the MapReduceTaskContext. We can have the context carrying KEY/VALUE out type, which only gets applied to the `write(K, V)` method, but not the `write(ds, K, V)` method (which is what you have implemented now). In that case, the initialize method of this class should be `initialize(MapReduceTaskContext<KEYOUT, VALUEOUT> context)` and the field should be adjusted accordingly.

anew: I like that.

cdap-api-hadoop/src/main/java/co/cask/cdap/mapreduce/AbstractMapper.java
anew: It will compile but fail at runtime.

anwar6953:
Ok, removed the cdap-api-hadoop for now. Bringing back the type parameterization on the MapReduceTaskContext class

anwar6953:
With MapReduceTaskContext being type-parameterized and without an AbstractMapper class, the user code is not very clean.

See:
https://github.com/caskdata/cdap/pull/3740/files#diff-ffc177771a36092be2afe2cfa0534e51R111

chtyim:
It's mainly the mapper/reducer needs to implements the ProgramLifecycle interface and memorizing the context object, right? That's ok since it's not the final change.

anwar6953:
Yea

anew:
This looks good to me, but Terence had a few more comments. Please wait for his LGTM.

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
```diff
@@ -102,21 +104,44 @@
...
+   * Overrides the output configuration of this MapReduce job to also allow writing to the specified dataset by
+   * its name.
+   *
+   * @param datasetName the name of the output dataset
+   */
+  void addOutput(String datasetName);
+
+  /**
+   * Overrides the output configuration of this MapReduce job to also allow writing to the specified dataset.
+   * Currently, the dataset specified in must be an {@link OutputFormatProvider}.
```
chtyim: Is this true? Meaning it cannot be a table dataset with arguments? Or we don't support arguments for table dataset?

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceTaskContext.java
```diff
@@ -0,0 +1,84 @@
...
+import co.cask.cdap.api.data.DatasetContext;
+import co.cask.cdap.api.templates.AdapterContext;
+import co.cask.cdap.api.workflow.Workflow;
+import co.cask.cdap.api.workflow.WorkflowToken;
+
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/**
+ * MapReduce task execution context. This is similar to MapReduceContext, but without ability to configure
```
chtyim: We don't need to say it is similar to MapReduceContext, right? They are not really related.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -202,42 +207,94 @@ public void setInput(String inputDatasetName, Dataset dataset) {
...
-   *
-   * @return the output dataset for the MapReduce job.
+   * @return the output datasets for the MapReduce job.
    */
-  public Dataset getOutputDataset() {
-    // use the dataset instance if it is set.
-    if (outputDataset != null) {
-      return outputDataset;
+  public Map<String, Dataset> getOutputDatasets() {
+    // the dataset instance in the entries may be null, so hydrate the entries if needed, prior to returning them
```
chtyim: What do mean by hydrate the entries?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
```diff
@@ -70,12 +77,11 @@
 
   private String inputDatasetName;
   private List<Split> inputDataSelection;
-  private String outputDatasetName;
   private Job job;
-  private Dataset outputDataset;
   private Dataset inputDataset;
   private Resources mapperResources;
   private Resources reducerResources;
+  private Map<String, Dataset> outputDatasets;
```
chtyim: final

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
```diff
@@ -0,0 +1,219 @@
...
+ * implement ProgramLifeCycle<MapReduceContext> in order to to get a MapReduceContext object.
+ *
+ * @param <KEYOUT>   output key type
+ * @param <VALUEOUT> output value type
+ */
+public class BasicMapReduceTaskContext<KEYOUT, VALUEOUT> implements MapReduceTaskContext<KEYOUT, VALUEOUT>,
+  MapReduceContext {
+
+  private final BasicMapReduceContext basicMapReduceContext;
+  private TaskInputOutputContext<?, ?, KEYOUT, VALUEOUT> hadoopContext;
```
chtyim: final

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * Mapreduce task runtime context which delegates to BasicMapReduceContext for non task-specific methods.
+ * It currently also extends MapReduceContext to support backwards compatability. Mapper and Reducer tasks could
+ * implement ProgramLifeCycle<MapReduceContext> in order to to get a MapReduceContext object.
+ *
+ * @param <KEYOUT>   output key type
+ * @param <VALUEOUT> output value type
+ */
+public class BasicMapReduceTaskContext<KEYOUT, VALUEOUT> implements MapReduceTaskContext<KEYOUT, VALUEOUT>,
+  MapReduceContext {
```
chtyim: Why implements MapReduceContext?

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
anwar6953: In the current framework, runtime arguments of a dataset aren't serialized into the hadoop conf and used when instantiating the dataset during task runtime.
Instead, they're used in OutputFormatProviders, to help the OutputFormatProvider figure out which configuration to return (for instance, in PFS/TPFS/FS).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
chtyim: The proper way to support calling of ProgramLifecycle interface with either MapReduceContext or MapReduceTaskContext should be done like this:

```java
class BasicMapReduceContext implements MapReduceContext {
...
}

class BasicMapReduceTaskContext implements MapReduceTaskContext {
...
}

class MapReduceLifecycleContext implements MapReduceContext, MapReduceTaskContext {
  private final MapReduceContext contextDelegate;
  private final MapReduceTaskContext taskContextDelegate;
  ...
}
```

This helps keep the actual implementation clean and easy to refactor in future.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceContext.java
anwar6953: I meant instantiate the Dataset instances for the names. Changed it to:
```
// the dataset instance in the entries may be null, so instantiate the Dataset instance if needed, prior to
    // returning the entries
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
```diff
@@ -0,0 +1,219 @@
...
+    multipleOutputs.write(namedOutput, key, value);
+  }
+
+
+  @Override
+  public void write(KEYOUT key, VALUEOUT value) throws IOException, InterruptedException {
+    hadoopContext.write(key, value);
+  }
+
+  public void close() throws IOException, InterruptedException {
```
chtyim: Why not implements Closeable?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
anwar6953: So we can pass it to `initialize(MapReduceContext)` for backwards compatability for now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
anwar6953: I thought you said not to have it.
https://github.com/caskdata/cdap/pull/3740#discussion_r38240675

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
chtyim: So will exception be thrown if the given dataset name doesn't refer to a OutputFormatProvider?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
chtyim: Right, I mean not in the interface (since it's public API, and we don't expect user to call close). But for the implementation, if that's need to get closed, it's perfectly okay to implements Closeable.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
chtyim: Do it in the way I mentioned above.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
```diff
@@ -0,0 +1,280 @@
...
+   * Cache for the taskContexts
+   */
+  private Map<String, TaskAttemptContext> taskContexts = new HashMap<>();
+
+  /**
+   * Checks the existence of a named output within a collection.
+   *
+   * @throws IllegalArgumentException if the output name is not existing/absent, as appropriate.
+   */
+  private static void checkNamedOutputName(String namedOutput, Collection namedOutputs, boolean expectToExist) {
```
chtyim: Why the `Colleciton` is without type parameter? Don't you see a warning in IntelliJ?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
```diff
@@ -0,0 +1,280 @@
...
+
+  private static final String FORMAT = ".format";
+  private static final String KEY = ".key";
+  private static final String VALUE = ".value";
+  private static final String CONF = ".conf";
+
+  /**
+   * Cache for the taskContexts
+   */
+  private Map<String, TaskAttemptContext> taskContexts = new HashMap<>();
```
chtyim: final?

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
anwar6953: Yea, there's an exception thrown.
```
if (!(dataset instanceof OutputFormatProvider)) {
      throw new IllegalArgumentException("Output dataset must be an OutputFormatProvider.");
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
anwar6953: I didn't see warning, but I added the type parameter.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
```diff
@@ -0,0 +1,280 @@
...
+      throw new IllegalArgumentException("Named output '" + namedOutput + "' already defined");
+    } else if (expectToExist && !namedOutputs.contains(namedOutput)) {
+      throw new IllegalArgumentException("Named output '" + namedOutput + "' not defined");
+    }
+  }
+
+  // Returns list of channel names.
+  static List<String> getNamedOutputsList(JobContext job) {
+    List<String> names = new ArrayList<>();
+    StringTokenizer st = new StringTokenizer(job.getConfiguration().get(MULTIPLE_OUTPUTS, ""), " ");
```
chtyim: Wouldn't it simpler by using `Splitter`?

```java
return new ArrayList<>(Splitter.on(" ").split(job.getConfiguration().get(MULTIPLE_OUTPUTS, "")));
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
```diff
@@ -0,0 +1,280 @@
...
+    conf.set(MULTIPLE_OUTPUTS, conf.get(MULTIPLE_OUTPUTS, "") + " " + namedOutput);
+    conf.set(MO_PREFIX + namedOutput + FORMAT, outputFormatClass);
+    conf.setClass(MO_PREFIX + namedOutput + KEY, keyClass, Object.class);
+    conf.setClass(MO_PREFIX + namedOutput + VALUE, valueClass, Object.class);
+    setNamedConfigurations(job, namedOutput, outputConfigs);
+  }
+
+  // instance code, to be used from Mapper/Reducer code
+
+  private TaskInputOutputContext context;
```
chtyim: Fields should be define on the top. Don't adapt the Hadoop coding style. It's horrible.

chtyim: Also, can these fields be final?

anwar6953: Yea, made them final.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
```diff
@@ -0,0 +1,280 @@
...
+
+      // add the record-writer to the cache
+      recordWriters.put(namedOutput, writer);
+    }
+    return writer;
+  }
+
+  // Create a taskAttemptContext for the named output with
+  // output format and output key/value types put in the context
+  private TaskAttemptContext getContext(String nameOutput) throws IOException {
```
chtyim: Why this method doesn't need to synchornized while the getRecordWriter needs to? Both of them seems to be looking up and modifying cache.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
```diff
@@ -0,0 +1,280 @@
...
+      context.setStatus(status);
+    }
+  }
+
+  /**
+   * Closes all the opened outputs.
+   * This should be called from cleanup method of map/reduce task.
+   */
+  @SuppressWarnings("unchecked")
+  public void close() throws IOException, InterruptedException {
```
chtyim: Implements Closeable.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
```diff
@@ -0,0 +1,280 @@
...
+    for (RecordWriter writer : recordWriters.values()) {
+      writer.close(context);
```
chtyim: If one writer.close throws exception, then the remaining won't get closed?

anwar6953: Is it suggested to LOG.ERROR if one fails? or try to close all of them and then propagate the exception at the end?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputsMainOutputWrapper.java
```diff
@@ -0,0 +1,102 @@
...
+    OutputFormat<K, V> rootOutputFormat = getRootOutputFormat(job);
+    return rootOutputFormat.<K, V>getRecordWriter(job);
+  }
+
+  @Override
+  public void checkOutputSpecs(JobContext context) throws IOException, InterruptedException {
+
+  }
+
+  public static void setRootOutputFormat(Job job,  String outputFormatClass, Map<String, String> outputConfig) {
```
chtyim: No javadoc?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputsMainOutputWrapper.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Override
+  public RecordWriter getRecordWriter(TaskAttemptContext job) throws IOException, InterruptedException {
+    OutputFormat<K, V> rootOutputFormat = getRootOutputFormat(job);
+    return rootOutputFormat.<K, V>getRecordWriter(job);
+  }
+
+  @Override
+  public void checkOutputSpecs(JobContext context) throws IOException, InterruptedException {
+
```
chtyim: Why empty? At least a comment.

cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
chtyim: better have a `@throws IllegalArgumentException ...` in the javadoc.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
chtyim: I think you still want to close all of them and throw the IOException out. You can use `addSuppressed` to help gather all IOExceptions.

```java
IOException ex = null;
for (RecordWriter writer : recordWriters.values()) {
  try {
    writer.close(context);
  } catch (IOException e) {
    if (ex == null) {
      ex = e;
    } else {
      ex.addSuppressed(e);
    }
  }
}
if (ex != null) {
  throw ex;
}
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/MultipleOutputs.java
anwar6953: This method is only used from within getRecordWriter, so it wasn't necessary.
But to protect against other usages from elsewhere (in the future), I made it synchronized now.

anwar6953:
@chtyim @anew Thanks for the review; I addressed comments. Please take another look.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceLifecycleContext.java
```diff
@@ -0,0 +1,210 @@
...
+    return delegate.getServiceURL(serviceId);
+  }
+
+  // TODO: Document usage of ProgramLifeCycle<MapReduceTaskContext> instead of ProgramLifeCycle<MapReduceContext>
+  // Until then, we can not remove the following methods.
+  private static final String UNSUPPORTED_ERROR_MESSAGE = "Operation not supported from MapReduce task-level context.";
+
+  @Override
+  public <T> T getHadoopJob() {
+    throw new UnsupportedOperationException(UNSUPPORTED_ERROR_MESSAGE);
```
chtyim: Instead of throwing exception, which can bring existing program, better log a warn and do no-op.

chtyim:
Just one comment left. Rest LGTM.

chtyim:
LGTM

anwar6953:
Squashed commits; pending build:
http://builds.cask.co/browse/CDAP-DUT2564-12

PR #3742, opened 2015-08-28
CDAP-3539 fixing upgrade to handle bad app metadata

albertshau:
Fixing an issue where the upgrade tool would choke on apps that
have the wrong location stored for the application archive. Not
sure how that can happen, it is probably a sign on multiple
cdap upgrades and one of those upgrades going wrong.

In any case, check the location exists. If not, try to find it
at the expected location. If that doesn't exist, just continue
on and log a message.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
@@ -519,11 +519,13 @@ public class AppLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
       datasets.add(new DatasetDetail(datasetSpec.getInstanceName(), datasetSpec.getTypeName()));
     }
 
-    // TODO: Remove the null check for artifact id once templates/adapters are removed
-    ArtifactSummary artifactSummary = spec.getArtifactId() == null ?
-      new ArtifactSummary(spec.getName(), null, true) : ArtifactSummary.from(spec.getArtifactId());
+    // this is only required if there are old apps lying around that failed to get upgrading during
+    // the upgrade to v3.2 for some reason. In those cases artifact id will be null until they re-deploy the app.
+    // in the meantime, we don't want this api call to null pointer exception.
+    ArtifactSummary summary = spec.getArtifactId() == null ?
+      new ArtifactSummary(spec.getName(), null, false) : ArtifactSummary.from(spec.getArtifactId());
     return new ApplicationDetail(spec.getName(), spec.getDescription(), spec.getConfiguration(),
-                                 streams, datasets, programs, artifactSummary);
+      streams, datasets, programs, summary);
   }
 
   private void ensureNamespaceExists(Id.Namespace namespace) throws NamespaceNotFoundException {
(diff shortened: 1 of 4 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
```diff
@@ -517,7 +517,12 @@ private static ApplicationDetail makeAppDetail(ApplicationSpecification spec) {
       datasets.add(new DatasetDetail(datasetSpec.getInstanceName(), datasetSpec.getTypeName()));
     }
 
+    // this is only required if there are old apps lying around that failed to get upgrading during
+    // the upgrade to v3.2 for some reason. In those cases artifact id will be null until they re-deploy the app.
+    // in the meantime, we don't want this api call to null pointer exception.
+    ArtifactSummary summary = spec.getArtifactId() == null ?
+      new ArtifactSummary(spec.getName(), null, false) : ArtifactSummary.from(spec.getArtifactId());
```
anew: It is certain that this can never happen for system artifact?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -396,11 +397,11 @@ public void upgrade(boolean continueOnFailure) throws Exception {
...
 
-        Location appJarLocation = store.getApplicationArchiveLocation(appId);
+        Location appJarLocation = findAppJarLocation(appId);
         if (appJarLocation == null) {
-          LOG.error(String.format("Unable to get location of jar for app '%s' in namespace '%s'. " +
-              "You will need to re-deploy the app after upgrade.",
-            appId.getNamespaceId(), appId.getId()));
+          // nothing we can do... skip and log a message
+          LOG.warn("Unable to find application jar for app '%s' in namespace '%s'. " +
+            "Please re-deploy the app manually after upgrade.");
```
anew: That is an important message. Will this only go to the log? Is there a way to report that this back directly to the UpgradeTool? So that the user who runs the upgrade immediately gets a report of what went wrong?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -474,6 +475,37 @@ public void upgrade(boolean continueOnFailure) throws Exception {
...
+          appJarLocation = file;
+          found = true;
+          break;
+        }
+      }
+      if (!found) {
+        return null;
+      }
+    }
+    return appJarLocation;
```
anew: It seems that instead of returning null, this method should throw an exception with a detailed message, why the location could not be found (there can be various reasons). The caller can catch the exception and log it properly.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/AppLifecycleHttpHandler.java
albertshau: this can never happen for apps created after 3.2, the deployment pipeline requires an artifact in order to create the app spec.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
albertshau: Could throw an exception, but that would halt the upgrade and it would never be able to complete. I've opened CDAP-3551to surface these types of errors in a better way.

anew:
LGTM

PR #3745, opened 2015-08-28
CDAP-3514: Add endpoint to get total count of all queries

jwang47:
https://issues.cask.co/browse/CDAP-3514
http://builds.cask.co/browse/CDAP-DUT2657

### DIFF
--- a/cdap-explore-client/src/main/java/co/cask/cdap/explore/service/Explore.java
+++ b/cdap-explore-client/src/main/java/co/cask/cdap/explore/service/Explore.java
@@ -88,7 +88,7 @@ public interface Explore {
   /**
    * Fetch a preview of the results of a Hive operation. This can be called only after the state of the operation is
    * {@link QueryStatus.OpStatus#FINISHED}. Two subsequent calls to this methods will return the same list of results.
-   * 
+   *
    * @param handle handle returned by {@link #execute(Id.Namespace, String)}.
    * @return preview list of {@link QueryResult}s.
    * @throws ExploreException on any error fetching a preview of the results.
@@ -116,6 +116,15 @@ public interface Explore {
    */
   List<QueryInfo> getQueries(Id.Namespace namespace) throws ExploreException, SQLException;
 
+  /**
+   * Get the number of active queries being executed in Hive, specifically
+   * queries that are being executed or completed queries that have not yet been closed.
+   *
+   * @param namespace namespace to count queries in.
+   * @return the number of live queries being executed.
+   * @throws ExploreException
+   */
+  int getActiveQueryCount(Id.Namespace namespace) throws ExploreException;
 
   ////// Metadata methods
 
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
@@ -742,7 +742,7 @@ public abstract class BaseHiveExploreService extends AbstractIdleService impleme
 
         operationHandle = doExecute(sessionHandle, statement);
         QueryHandle handle = saveReadWriteOperation(operationHandle, sessionHandle, sessionConf,
-                                               statement, database);
+                                                    statement, database);
         LOG.trace("Executing statement: {} with handle {}", statement, handle);
         return handle;
       } catch (Throwable e) {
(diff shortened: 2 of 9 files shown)


poornachandra:
@jwang47 This PR is to get a count of active queries, not a count of all queries, right?

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -996,6 +998,22 @@ public void close(QueryHandle handle) throws ExploreException, HandleNotFoundExc
     return result;
   }
 
+  @Override
+  public int getActiveQueryCount(Id.Namespace namespace) throws ExploreException {
+    startAndWait();
+
+    int count = 0;
+    for (Map.Entry<QueryHandle, OperationInfo> entry : activeHandleCache.asMap().entrySet()) {
+      if (entry.getValue().getNamespace().equals(getHiveDatabase(namespace.getId()))) {
```
poornachandra: `getHiveDatabase(namespace.getId())` does not change for a method call, it can move out of the loop, right?

cdap-explore-client/src/main/java/co/cask/cdap/explore/service/Explore.java
```diff
@@ -116,6 +116,14 @@
    */
   List<QueryInfo> getQueries(Id.Namespace namespace) throws ExploreException, SQLException;
 
+  /**
+   * Get the number of active queries being executed in Hive.
```
poornachandra: Active queries are queries that are being executed, or completed queries that have not been closed yet.

cdap-explore/src/test/java/co/cask/cdap/explore/service/HiveExploreServiceTestRun.java
```diff
@@ -273,6 +276,26 @@ public void testHiveIntegration() throws Exception {
   }
 
   @Test
+  public void testQueriesCount() throws Exception {
+    Id.Namespace testNamespace1 = Id.Namespace.from("testQueriesCount");
+    exploreClient.addNamespace(testNamespace1).get();
```
bdmogal: This may need to be updated now to use a combination of ``NamespaceClient`` and ``DatasetFramework`` to manage namespaces in explore tests. Look at how its done in ``BaseHiveExploreServiceTest`` in latest develop.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/InactiveOperationInfo.java
```diff
@@ -21,11 +21,14 @@
 
 import java.util.List;
 
-final class InactiveOperationInfo extends OperationInfo {
+/**
+ * OperationInfo representing an inactive operation.
+ */
+public final class InactiveOperationInfo extends OperationInfo {
```
poornachandra: I see that the test classes are not part of the same namespace as code they test. Can you move the test classes to code namespace so that these classes can remain package-private?

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
```diff
@@ -137,7 +138,8 @@
   private final SchedulerQueueResolver schedulerQueueResolver;
 
   // Handles that are running, or not yet completely fetched, they have longer timeout
-  private final Cache<QueryHandle, OperationInfo> activeHandleCache;
+  @VisibleForTesting
+  public final Cache<QueryHandle, OperationInfo> activeHandleCache;
```
poornachandra: Not a good idea to make a member accessible publicly even with an annotation. Instead can you run a query using explore service, and then verify active count?

poornachandra:
@jwang47 Done with review, please address comments.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
jwang47: Is there a way to control when a query completes? I don't want to introduce a flaky test.

poornachandra: Until the query is closed, it remains in the active handler cache. So you have control on when it gets removed from the cache. The test will not be flaky.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/InactiveOperationInfo.java
poornachandra: Actually, if you test using a query as discussed below then you will not need these changes.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/BaseHiveExploreService.java
jwang47: Cool, will fix the test then.

cdap-explore/src/main/java/co/cask/cdap/explore/service/hive/InactiveOperationInfo.java
```diff
@@ -21,11 +21,14 @@
...
 
+/**
+ * OperationInfo representing an inactive operation.
+ */
 final class InactiveOperationInfo extends OperationInfo {
   private final List<ColumnDesc> schema;
   private final QueryStatus status;
 
-  InactiveOperationInfo(OperationInfo operationInfo, List<ColumnDesc> schema, QueryStatus status) {
+  public InactiveOperationInfo(OperationInfo operationInfo, List<ColumnDesc> schema, QueryStatus status) {
```
poornachandra: This and the next few classes can revert to package-private access, right?

poornachandra:
@jwang47 One minor comment, please fix and merge.

PR #3746, opened 2015-08-28
Allow users to customize how the partitions are stored in directories

ataylor4:
http://builds.cask.co/browse/CDAP-DUT2655-8

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
@@ -33,6 +33,7 @@ import javax.annotation.Nullable;
 public class TimePartitionedFileSetArguments extends PartitionedFileSetArguments {
 
   public static final String OUTPUT_PATH_FORMAT = "output.file.path.format";
+  public static final String OUTPUT_TIME_ZONE = "output.time.zone";
   public static final String OUTPUT_PARTITION_TIME = "output.partition.time";
   public static final String INPUT_START_TIME = "input.start.time";
   public static final String INPUT_END_TIME = "input.end.time";
@@ -73,13 +74,10 @@ public class TimePartitionedFileSetArguments extends PartitionedFileSetArguments
   public static void setOutputPathFormat(Map<String, String> arguments, String pathFormat) {
     long curTime = System.currentTimeMillis();
     try {
-      SimpleDateFormat format = new SimpleDateFormat(pathFormat.split(",")[0]);
-      if (pathFormat.split(",").length > 1) {
-        format.setTimeZone(TimeZone.getTimeZone(pathFormat.split(",")[1]));
-      }
+      SimpleDateFormat format = new SimpleDateFormat(pathFormat);
       format.format(new Date(curTime));
     } catch (Exception e) {
-      Throwables.propagate(new IOException("Invalid date format: " + pathFormat + '\n' + e));
+      throw new IllegalArgumentException("Invalid date format: " + pathFormat + '\n' + e);
     }
     arguments.put(OUTPUT_PATH_FORMAT, pathFormat);
   }
@@ -92,6 +90,26 @@ public class TimePartitionedFileSetArguments extends PartitionedFileSetArguments
   @Nullable
   public static String getOutputPathFormat(Map<String, String> arguments) {
     return arguments.get(OUTPUT_PATH_FORMAT);
+  }
+
+  /**
+   * This is the time zone to format the date in. Time zone is only used in conjunction with
+   * {@link #getOutputPathFormat(Map)} to format the output path in the correct time.
+   * @param timeZone The string ID of the time zone. It is parsed by {@link TimeZone#getTimeZone(String)},
+   *                 and if the string ID is not a valid time zone, UTC is used.
+   */
+  public static void setOutputTimeZone(Map<String, String> arguments, String timeZone) {
+    arguments.put(OUTPUT_TIME_ZONE, timeZone);
+  }
+
+  /**
+   * This is the time zone used to format the date for the output partition.
+   * @return The String ID of the time zone of the date.
+   * May be null.
+   */
+  @Nullable
+  public static String getOutputTimeZone(Map<String, String> arguments) {
+    return arguments.get(OUTPUT_TIME_ZONE);
   }
 
   /**
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
@@ -17,6 +17,7 @@
 package co.cask.cdap.api.dataset.lib;
 
 import co.cask.cdap.api.annotation.Beta;
+import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
 
 import java.io.IOException;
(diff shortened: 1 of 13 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
```diff
@@ -54,6 +55,29 @@ public static Long getOutputPartitionTime(Map<String, String> arguments) {
   }
 
   /**
+   * This is the file extension for each partition when using TimePartitionedFileSet as an OutputFormatProvider.
+   * It's used to generate the end of the output file path for each partition.
+   * @param pathFormat The format for the path; for example: '%tF/%tH-%tM.%d',
```
chtyim: It doesn't tell what are all the available formats. E.g. what is `%tF`? Is it just the format string providing to `SimpleDateFormat`? If that the case, need to state it clear and have a `{@link SimpleDateFormat}`.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/TPFSSinkConfig.java
```diff
@@ -33,8 +33,13 @@
   @Nullable
   protected String basePath;
 
-  public TPFSSinkConfig(String name, @Nullable String basePath) {
+  @Description(TimePartitionedFileSetSink.PATH_FORMAT_DESC)
```
albertshau: etl plugins are also documented at cdap-docs/application-templates/source/etl/templates/sinks/batch/ (this is what the UI links to when you hit the info button for a plugin).  If we are adding

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
albertshau: doc mentions that it is what `String.format()` uses.  One thing that should be mentioned in the javadoc is that the path must be unique per partition, otherwise runtime errors will occur.

albertshau: it is also worth looking at SimpleDateFormat, it may be more user friendly and flexible to use it

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
```diff
@@ -54,6 +55,32 @@ public static Long getOutputPartitionTime(Map<String, String> arguments) {
...
+   *                   which will create a file path ending in the format of 2015-01-01/20-42.142017372000
+   *                   The pathFormat will be the format provided to
+   *                   {@link String}.format(String format, Object ... args), and args will be the appropriate
+   *                   number of instances of the time in milliseconds since epoch time (00:00:00 January 1, 1970 UTC).
+   */
+  public static void setOutputPathFormat(Map<String, String> arguments, String pathFormat) {
+    arguments.put(OUTPUT_PATH_FORMAT, pathFormat);
```
albertshau: should get that the format is a valid format and throw an exception if so.  String.format already throws IllegalFormatException

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
albertshau: what time zone gets used? is it always utc? if not we should have a way to set the timezone as well.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
```diff
@@ -54,6 +60,41 @@ public static Long getOutputPartitionTime(Map<String, String> arguments) {
...
+   *                   {@link java.text.SimpleDateFormat}. If left blank, then the partitions will be of the form
+   *                   2015-01-01/20-42.142017372000, with the time being the time UTC.
+   *                   Note that each partition must have a unique file path or a runtime exception will be thrown.
+   */
+  public static void setOutputPathFormat(Map<String, String> arguments, String pathFormat) {
+    long curTime = System.currentTimeMillis();
+    try {
+      SimpleDateFormat format = new SimpleDateFormat(pathFormat.split(",")[0]);
+      if (pathFormat.split(",").length > 1) {
+        format.setTimeZone(TimeZone.getTimeZone(pathFormat.split(",")[1]));
```
albertshau: A comma is a valid character to have in a file path, so this parsing may break. I think its better to have time zone as a separate argument. Could have one method with just the path format which defaults to utc, then another that also takes in a time zone.

albertshau:
one comment about explicitly passing timezone instead of parsing it

albertshau:
few more comments regarding behavior

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/TimePartitionedFileSetArguments.java
```diff
@@ -54,6 +62,56 @@ public static Long getOutputPartitionTime(Map<String, String> arguments) {
...
+   *                   which will create a file path ending in the format of 2015-01-01/20-42,
+   *                   with the time of the partition being the time in the timezone of Los Angeles (PST or PDT).
+   *                   The pathFormat will be the format provided to
+   *                   {@link java.text.SimpleDateFormat}. If left blank, then the partitions will be of the form
+   *                   2015-01-01/20-42.142017372000, with the time being the time UTC.
+   *                   Note that each partition must have a unique file path or a runtime exception will be thrown.
+   * @param timeZone The string ID of the time zone. It is parsed by {@link TimeZone#getTimeZone(String)},
+   *                 and if the string ID is not a valid time zone, UTC is used.
+   */
+  public static void setOutputPathFormat(Map<String, String> arguments, String pathFormat, String timeZone) {
```
albertshau: annotate timeZone as Nullable. We can also add a convenience method that doesn't take a timezone and calls this one.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/TimePartitionedFileSetSink.java
```diff
@@ -37,6 +40,15 @@
     "are written to. If it doesn't exist, it will be created.";
   protected static final String BASE_PATH_DESC = "The base path for the time partitioned fileset. Defaults to the " +
     "name of the dataset.";
+  protected static final String PATH_FORMAT_DESC = "The format for the path; for example: " +
+    "'yyyy-MM-dd/HH-mm' will create a file path ending in the format of 2015-01-01/20-42. " +
+    "The string provided will be provided to SimpleDataFormat. " +
+    "If left blank, then the partitions will be of the form 2015-01-01/20-42.142017372000. " +
+    "Note that each partition must have a unique file path or a runtime exception will be thrown.";
+  protected static final String TIME_ZONE_DESC = "The time zone to format the partition. " +
+    "This option is only used if pathFormat is set. If left blank, defaults to UTC. " +
```
albertshau: Also mention that if it's invalid, it gets interpreted as UTC

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/TimePartitionedFileSetDefinition.java
```diff
@@ -82,7 +86,16 @@ public PartitionedFileSet getDataset(DatasetContext datasetContext, DatasetSpeci
...
       if (FileSetArguments.getOutputPath(arguments) == null) {
-        String path = String.format("%tF/%tH-%tM.%d", time, time, time, time);
+        String outputPathFormat = TimePartitionedFileSetArguments.getOutputPathFormat(arguments);
+        String path;
+        if (Strings.isNullOrEmpty(outputPathFormat)) {
+          path = String.format("%tF/%tH-%tM.%d", time, time, time, time);
+        } else {
+          SimpleDateFormat format = new SimpleDateFormat(outputPathFormat);
+          String timeZoneID = TimePartitionedFileSetArguments.getOutputPathTimeZone(arguments);
+          format.setTimeZone(TimeZone.getTimeZone("" + timeZoneID));
```
albertshau: how come there is this empty string?

albertshau: o in case timeZoneID is null?  it's more clear if you explicitly handle that case. Otherwise somebody who is not familiar with the getTimeZone will get confused.

albertshau:
couple more non-functional comments. If build is green, fix those and merge

PR #3748, opened 2015-08-28
Feature/ui programs layout workflows

astewes:
*  Adds helper classes to properly display status color
*  Adds more statuses to start-stop-button directive
*  Fixes some styling issues


cdap-ui/app/directives/program-status/program-status.js
```diff
@@ -5,36 +5,50 @@ angular.module(PKG.name + '.commons')
...
+              path = '/apps/' + $state.params.appId + '/' + $scope.type + '/' + $state.params.programId + '/runs/' + $scope.runid;
+            }
+
+            pollPromise = dataSrc.poll({
+              _cdapNsPath: path,
+              interval: $scope.pollInterval || 10000
+            }, function (res) {
+              var startMs = res.start * 1000;
+              $scope.start = new Date(startMs);
+              console.info('status: ', res.status, res.runid);
```
ajainarayanan: Remove?

cdap-ui/app/features/workflows/controllers/tabs/runs/tabs/status-ctrl.js
```diff
@@ -1,6 +1,6 @@
 // FIXME: Needs to be re-thought.
 var runparams = {},
-    alertpromise,
+    // alertpromise,
```
ajainarayanan: This can be removed. This is already committed to develop.

cdap-ui/app/features/workflows/templates/detail.html
```diff
@@ -1,39 +1,199 @@
...
+  <div class="program-level">
+    <div class="container">
+      <div class="row" program-heading>
+        <div class="col-sm-4 col-md-6">
+          <h2 ng-bind="$state.params.programId"></h2>
+          <div class="type-block">
+            <span class="icon-workflow"></span>
+            <small> Type: Workflow </small>
+          </div>
+          <p> Description goes here. </p>
```
ajainarayanan: May be we should remove this? We need to decide on how we are going to show description on the detail view.

cdap-ui/app/features/workflows/templates/detail.html
```diff
@@ -1,39 +1,199 @@
...
-  </div>
-</div>
+  <div class="run-level">
+    <div class="container">
+      <div class="run-level-inner">
+
+        <div dropdown is-open="dropdown.isopen" class="flex-item">
+          <a class="btn btn-default dropdown-toggle"
+             dropdown-toggle ng-disabled="RunsController.runs.length === 0">
+            <span class="fa fa-circle" ng-class="{'text-primary': RunsController.runs.selected.status === 'COMPLETED', 'text-success': RunsController.runs.selected.status === 'STARTED' || RunsController.runs.selected.status === 'RUNNING', 'text-danger': RunsController.runs.selected.status === 'FAILED' || RunsController.runs.selected.status === 'KILLED', 'text-muted': RunsController.runs.selected.status === 'STOPPED', 'text-warning': RunsController.runs.selected.status === 'SUSPENDED'}"></span>
```
ajainarayanan: We need to figure out a way to show this a little bit more 'readable'.  May be controller should have a map of states to the css class or may be something simpler.

cdap-ui/app/features/workflows/controllers/tabs/runs/runs-detail-ctrl.js
```diff
@@ -3,7 +3,9 @@ class WorkflowsRunsDetailController {
     let filterFilter = $filter('filter');
     let match = filterFilter($scope.RunsController.runs, {runid: $state.params.runid});
     $scope.RunsController.runs.selected.runid = match[0].runid;
-
+    $scope.RunsController.runs.selected.status = match[0].status;
+    $scope.RunsController.runs.selected.start = match[0].start;
+    $scope.RunsController.runs.selected.end = match[0].end;
```
elfenheart: Can't we do $scope.RunsController.runs.selected = match[0]?

ajainarayanan: Talked to @elfenheart. We don't want to overwrite the object reference here since it is also being used in the parent controller (RunsController)

cdap-ui/app/features/workflows/templates/detail.html
nitinmotgi: Why we need to remove this and what more needs to be decided ?

cdap-ui/app/features/workflows/templates/detail.html
nitinmotgi: I agree we need to centralize this and use it everywhere. Let's not replicate this everywhere.

ajainarayanan: @nitinmotgi Talked to andrew. It is not straight forward to do this. We will sit together to work on this tomorrow. Not blocking this PR only for this change. We will change it as a separate PR (since it is going to be merged to a feature branch). This PR needs to be merged for other program views. Hence we are going forward with merging this and will change across all program views in a separate PR tomorrow.

ajainarayanan:
:+1: LGTM

We still need to bind description to the detailed view and consolidate having css classes added to template for various conditions. These can be done in separate PRs and should not be blocking from update other program views. Hence merging this PR now to address those two review comments in future PRs.

PR #3749, opened 2015-08-29
Changing pom files to create system artifacts, plugin artifacts and json files for the plugin artifacts

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-DUT2661


gokulavasan:
@albertshau Please review.

cdap-app-templates/cdap-dq/src/main/java/co/cask/cdap/dq/DataQualityApp.java
```diff
@@ -88,45 +84,8 @@
     private String datasetName;
     private Map<String, Set<String>> fieldAggregations;
 
-    public ConfigClass() {
```
albertshau: while we're at it, lets change the name of this to DQConfig or something like that.

cdap-master/pom.xml
```diff
@@ -254,6 +365,27 @@
...
+            <executions>
+              <execution>
+                <id>rename-plugin-json</id>
+                <phase>process-resources</phase>
+                <goals>
+                  <goal>run</goal>
+                </goals>
+                <configuration>
+                  <target>
+                    <move file="${stage.opt.dir}/artifacts/ETLBatch.json" tofile="${stage.opt.dir}/artifacts/cdap-etl-lib-${project.version}-batch.json" />
```
albertshau: lets rename the source to something closer to what it will finally be, otherwise its hard to tell just by looking which file is which.  ETLBatch -> cdap-etl-lib-batch.json

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/resources/DataQuality.json
```diff
@@ -0,0 +1,3 @@
+{
+  "parents": [ "cdap-dq[${project.version},${project.version}]" ]
```
albertshau: lets rename this to cdap-data-quality instead of cdap-dq

cdap-app-templates/cdap-dq/src/main/java/co/cask/cdap/dq/DataQualityApp.java
```diff
@@ -137,6 +96,26 @@ public ConfigClass(int workflowScheduleMinutes, DataQualitySource source,
...
+    Preconditions.checkArgument(configObj.workflowScheduleMinutes > 0,
+                                "Workflow Frequency in minutes (>0) should be provided");
+    Preconditions.checkArgument(configObj.source != null, "Configuration for DataQualityApp Source is missing");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(configObj.source.getName()),
+                                "Data Quality source name should not be null or empty");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(configObj.source.getId()),
+                                "Data Quality source id should not be null or empty");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(configObj.datasetName),
+                                "Output Dataset name should be not be null or empty");
+    Preconditions.checkArgument(configObj.fieldAggregations != null, "Field Aggregations needs to be specified");
```
albertshau: 'fieldAggregations' instead of Field Aggregations, since that is the name of the config field

cdap-app-templates/cdap-dq/src/main/java/co/cask/cdap/dq/DataQualityApp.java
```diff
@@ -137,6 +96,26 @@ public ConfigClass(int workflowScheduleMinutes, DataQualitySource source,
   @Override
   public void configure() {
     ConfigClass configObj = getContext().getConfig();
+    Preconditions.checkArgument(configObj.workflowScheduleMinutes > 0,
```
albertshau: can we open a jira to support more frequencies.  I think we would want to support a syntax like the one we use in stream source, '5m', '6h', '1d', etc.

albertshau:
couple comments

gokulavasan:
@albertshau Addressed comments. Please take a look again.

albertshau:
:+1:

PR #3753, opened 2015-08-31
Feature/ui programs layout mapreduce

astewes:
*  Reorganizes states
*  Removes runs.html template (markup moved to detail.html)


elfenheart:
:+1: LGTM

PR #3754, opened 2015-08-31
Feature/ui programs layout flows

astewes:
*  Reorganizes states
*  Removes runs.html template (markup moved to detail.html)


elfenheart:
History modal is not showing up for me

cdap-ui/app/features/flows/controllers/tabs/runs-ctrl.js
```diff
@@ -55,4 +56,26 @@ angular.module(PKG.name + '.feature.flows')
     this.activeTab = tab;
 
    };
+
+   this.openHistory = function() {
+     this.$bootstrapModal.open({
+       size: 'lg',
+       template: '<my-program-history data-runs="RunsController.runs" data-type="FLOWS"></my-program-history>',
```
elfenheart: I think the data-runs here is incorrect. It should be data-runs="runs"

ajainarayanan:
:+1: LGTM

PR #3755, opened 2015-08-31
removes history and data templates

astewes:
*  Removes data and history templates (in lieu of directives used in template defined in runs-ctrl)


elfenheart:
:+1: LGTM

PR #3762, opened 2015-08-31
fixes some minor styling issues in the program detail header + UI view

astewes:
*  Adds proper spacing around program description
*  Fixes some button border and color issues
*  Adds right border to the run level row
*  Adds width to runid text so that layout doesn't break when runid text exceeds a certain length
*  Top aligns status color circle and down caret in runid dropdown


elfenheart:
:+1: LGTM

PR #3763, opened 2015-08-31
CDAP-3488 use the correct package name for batch config

albertshau:


gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-DUT2673

PR #3765, opened 2015-08-31
CDAP-3432: CLI prompt doesn't show the namespace name when passed from URI

jwang47:
https://issues.cask.co/browse/CDAP-3432
http://builds.cask.co/browse/CDAP-DUT2675


gokulavasan:
LGTM :+1:

jwang47:
cdap-cli tests passed, merging.

PR #3766, opened 2015-08-31
(CDAP-3387) Store the stream format specification as part of SerDe property

chtyim:
- Detach it from StreamConfig so that different stream table view can have different format spec
- Modify the InputFormat/RecordRecord to use the FormatSpecification from the SerDe instead of reading from StreamConfig.

### DIFF
--- a/cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
+++ b/cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
@@ -18,6 +18,7 @@ package co.cask.cdap.explore.service;
 
 import co.cask.cdap.api.data.batch.RecordScannable;
 import co.cask.cdap.api.data.batch.RecordWritable;
+import co.cask.cdap.api.data.format.FormatSpecification;
 import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.data.schema.UnsupportedTypeException;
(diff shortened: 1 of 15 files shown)


cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
```diff
@@ -188,7 +196,8 @@ public QueryHandle enableDataset(Id.DatasetInstance datasetID, DatasetSpecificat
           ((RecordScannable) dataset).getRecordType() : ((RecordWritable) dataset).getRecordType();
 
         // if the type is a structured record, use the schema property to create the table
-        if (StructuredRecord.class.equals(recordType)) {
+        // Use == because that's what same class means.
+        if (StructuredRecord.class == recordType) {
```
jwang47: Why does `recordType` have to be == StructuredRecord.class instead of allowing `recordType` to be a subclass?

cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
```diff
@@ -96,7 +93,18 @@ public InputSplit createSplit(Path path, Path indexPath, long startTime, long en
                                                                StreamUtils.getGeneration(streamConfig));
 
     StreamInputSplitFinder.Builder builder = StreamInputSplitFinder.builder(streamPath.toURI());
-    return setupBuilder(conf, streamConfig, builder).build(SPLIT_FACTORY);
+
+    // Get the Hive table path for the InputSplit created. It is just to satisfy hive. The InputFormat never uses it.
```
jwang47: Why the change? Don't fully understand the comment.

jwang47: Oh, nevermind saw the StreamInputSplit constructor..

cdap-explore/src/main/java/co/cask/cdap/hive/stream/StreamInputSplit.java
```diff
@@ -45,7 +46,8 @@ public StreamInputSplit() {
   /**
    * Constructs a split.
    *
-   * @param path Path for the stream event file.
+   * @param path the hive file path. Never used by {@link HiveStreamInputFormat}; it is just to satisfy hive.
+   * @param eventPath Path for the stream event file.
```
jwang47: Why do we want eventPath and not pass that to super()?

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
chtyim: It was checking for equals, which is the same as "==" (in fact equals check with have a warning).

cdap-explore/src/main/java/co/cask/cdap/hive/stream/HiveStreamInputFormat.java
chtyim: Updated the javadoc for StreamInputSplit to include the detail of how *funny* Hive behaves.

cdap-explore/src/main/java/co/cask/cdap/explore/service/ExploreTableManager.java
chtyim: Not intended to change the logic, but just to eliminate the warning. I am not sure if the current logic works for sub-class of StructuredRecord.

cdap-explore/src/main/java/co/cask/cdap/hive/stream/StreamInputSplit.java
chtyim: See Javadoc of this class (may need to reload to see it).

jwang47:
LGTM

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2677-1

PR #3767, opened 2015-09-01
Feature/ui styling program modals

astewes:
*  Adds cdap-modal helper class for styling program view modals
*  Removes some styling that is no longer needed with content that has moved from tabs to modals


ajainarayanan:
:+1: LGTM apart from the adapter detailed view fix.

ajainarayanan:
:+1: LGTM

PR #3770, opened 2015-09-01
CDAP-3489 re-introduce app name as part of state store key

albertshau:


gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-DUT2680

anew:
Is this a compatible change? Will existing ETL pipelines work after this change? Or do we need migration?

albertshau:
this is how it was when templates except instead of app name we used adapter name.  Pre-3.2 pipelines are getting deleted as part of upgrade since adapters are being removed.  3.2.0 snapshot pipelines will be broken but don't think we're worried about that

PR #3771, opened 2015-09-01
CDAP-3559 dont delete the app archive when deleting an app

albertshau:
Now that apps are created from artifacts, this is bad behavior.
We don't have to delete the app jar since other apps may be using
it, and since futures apps may want to use it.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2681

gokulavasan:
LGTM :+1:

PR #3772, opened 2015-09-01
[CDAP-1864] Added a namespace client that services other than AppFabricServer can use to check namespace existence.

bdmogal:
Provide an AbstractNamespaceClient in Dataset Service Handlers to respond with 404 when the specified namespace does not exist.
The following implementations of AbstractNamespaceClient are now available to use in unit tests and in standalone:
1. InMemoryNamespaceClient: Lightweight client that doesn't make any HTTP calls, to be used when AppFabricServer is not started, and a DefaultNamespaceAdmin is also unavailable to interact directly with MDS.
2. LocalNamespaceClient: Delegates to DefaultNamespaceAdmin when that is available, but AppFabricServer is not started.

- Other minor refactoring in Dataset Handler tests to be able to test with multiple namespaces.
- Also cleaned up Explore tests to use a combination of AbstractNamespaceClient + DatasetFramework to manage namespaces.

Build: http://builds.cask.co/browse/CDAP-DUT2671


chtyim:
How many different implementations are there for NamespaceClient?

bdmogal:
@chtyim There are 4:
1. ``DiscoveryNamespaceClient`` - Used in standalone and distributed mode by stream and dataset service
2. ``NamespaceClient`` - Fixed ``ClientConfig`` based REST Client used in cdap-clients and cdap-cli
3. ``LocalNamespaceClient`` - Used in tests which do not use AppFabricServer REST APIs i.e. ProgramRunnerTests, etc, but which have access to the ``Store``, and have a binding to ``DefaultNamespaceAdmin``. This client delegates to the injected ``DefaultNamespaceAdmin``.
4. ``InMemoryNamespaceClient`` - Used in dataset and explore tests that do not start/interact with AppFabric. This just maintains an in-memory datastructure for namespaces.

chtyim:
The first 2 can be merged to one. The "2" can be done by simply implementing the DiscoveryServiceClient which always return the same Discoverable..

Is 4 only use in test?

bdmogal:
3 and 4 are used only in tests right now. 

The only difference between 1 and 2 is that they override the following two methods in ``AbstractNamespaceClient`` differently. - 

``HttpResponse execute(HttpRequest request)``; and
``URL resolve(String resource)``;

Their namespace related code is the same and is shared in ``AbstractNamespaceClient``

bdmogal:
I'll look into your suggestion about using a ``ClientConfig`` based ``DiscoveryServiceClient``. Currently this was modeled on the way ``ExploreClient``s are built.

chtyim:
If InMemoryNamespaceClient is only used in test, move it to src/test/java and have the modules that need that class to depends on cdap-app-fabric `<type>test-jar</type>`

bdmogal:
Spoke with @chtyim offline. ``InMemoryNamespaceClient`` and ``LocalNamespaceClient`` are injected via a Guice module, which is in production code. Hence, they cannot be moved to ``src/test``.

anwar6953:
LGTM

bdmogal:
@chtyim Merging ``DiscoveryNamespaceClient`` and ``NamespaceClient`` together might involve greater refactoring. I will explore that separately in https://issues.cask.co/browse/CDAP-3570, especially since ``NamespaceClient`` has a lot of usages across ``cdap-clients``, ``cdap-cli`` and integration tests.

bdmogal:
Build is green. Tested on standalone and cluster. Merging.

PR #3636, opened 2015-08-18
Advanced Feature: Class Loading

awholegunch:
Takes the blog post on class loading and adds it as an advanced features section. 

Fix for https://issues.cask.co/browse/CDAP-2930.

Staged at: [Java Class Loading and Distributed Data Processing Frameworks](http://docs-staging.cask.co/staging/cdap/3.2.0-SNAPSHOT-feature-3.2_classloading/en/developers-manual/advanced/class-loading.html)

PR #3774, opened 2015-09-01
Documentation on Class Loading in CDAP

awholegunch:
Fixes per comments in https://github.com/caskdata/cdap/pull/3636.


anew:
LGTM

PR #3775, opened 2015-09-01
Adds map<string,string> to schema properties to be used by schema-edi…

ajainarayanan:
…tor widget in adapter modals


elfenheart:
:+1: LGTM

PR #3778, opened 2015-09-01
Check that dataset service is up, in addition to system services' status

anwar6953:
Check that dataset service is up, in addition to /system/services/status.

http://builds.cask.co/browse/CDAP-DUT2691-1

### DIFF
--- a/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
+++ b/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
@@ -97,7 +97,13 @@ public class IntegrationTestBase {
     Callable<Boolean> monitorCallable = new Callable<Boolean>() {
       @Override
       public Boolean call() throws Exception {
-        return getMonitorClient().allSystemServicesOk();
+        // first wait for all system services to be 'OK'
+        if (!getMonitorClient().allSystemServicesOk()) {
+          return false;
+        }
+        // check that the dataset service is up
+        getNamespaceClient().list();
+        return true;
       }
     };
 


cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -97,7 +97,13 @@ protected void checkSystemServices() throws ServiceUnavailableException {
     Callable<Boolean> monitorCallable = new Callable<Boolean>() {
       @Override
       public Boolean call() throws Exception {
-        return getMonitorClient().allSystemServicesOk();
+        // first wait for all system services to be 'OK'
+        if (!getMonitorClient().allSystemServicesOk()) {
+          return false;
+        }
+        // check that the dataset service is up
+        getNamespaceClient().list();
```
bdmogal: Its very difficult to verify this fix. Do you think the following would help:

```
try {
  getNamespaceClient().list();
  LOG.info("Able to discover dataset service");
  return true;
} catch (IOException e) {
  LOG.error("Dataset Service may not be up yet");
  return false;
}
```

If there is a better way of making sure that the ``IOException`` is indeed an 'Unable to discover dataset service', that's even better, but I think it may not be super important.

anwar6953: the list() method throws an IOE, which will be retried anyways.
The http client for the namespace client will log all attempted requests and their response, which I think is enough logging. There's no need to additionally log (such as you've done in your comment).

bdmogal:
:+1:

PR #3779, opened 2015-09-01
[CDAP-3391] Improve exception/interrupt handling and logging in Abstr…

bdmogal:
…actProgramTwillRunnable and FileLogAppender

Also some additional cleanup of unused code in FileLogAppender.

Jira: [CDAP-3391](https://issues.cask.co/browse/CDAP-3391)
Build: https://builds.cask.co/browse/CDAP-DUT2684

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
@@ -250,11 +250,14 @@ public abstract class AbstractProgramTwillRunnable<T extends ProgramRunner> impl
   @Override
   public void stop() {
     try {
-      LOG.info("Stopping runnable: {}", name);
+      LOG.info("Stopping runnable: {}.", name);
       controller.stop().get();
       logAppenderInitializer.close();
+    } catch (InterruptedException e) {
+      LOG.debug("Interrupted while stopping runnable: {}.", name, e);
+      Thread.currentThread().interrupt();
     } catch (Exception e) {
-      LOG.error("Fail to stop: {}", e, e);
+      LOG.error("Failed to stop runnable: {}.", name, e);
       throw Throwables.propagate(e);
     }
   }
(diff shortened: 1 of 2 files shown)


henry-cask:
The change looks good, but why downgrade to debug level?

poornachandra:
@bdmogal I think any info logging that happens once per run can remain at info level. Any repeated messages can become debug.

bdmogal:
Reverted to info level for one-time logs. Please review again, Thanks!

cdap-watchdog/src/main/java/co/cask/cdap/logging/appender/file/FileLogAppender.java
```diff
@@ -182,6 +176,12 @@ public void stop() {
     }
 
     scheduledExecutor.shutdownNow();
+    try {
+      scheduledExecutor.awaitTermination(Integer.MAX_VALUE, TimeUnit.NANOSECONDS);
```
poornachandra: Let's wait for a reasonable time, like 5 mins for it to terminate.

poornachandra:
One more comment, please fix and merge.

bdmogal:
Addressed comments. Will merge after next successful build - http://builds.cask.co/browse/CDAP-DUT2684-3

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
```diff
@@ -313,11 +316,11 @@ public void error(Throwable cause) {
 
     try {
       state.get();
-      LOG.info("Program stopped.");
+      LOG.info("Program {} stopped.", name);
     } catch (InterruptedException e) {
-      LOG.warn("Program interrupted.", e);
+      LOG.warn("Program {} interrupted.", name, e);
```
sagarkapare: Should we set the interrupt flag here as well?

bdmogal: Added

bdmogal:
Tested on cluster. @sagarkapare please take another look.

sagarkapare:
LGTM :+1:

bdmogal:
Squashed commits. Build is green. Integration tests pass. Merging.

PR #3780, opened 2015-09-01
Update md5 of recently updated files to fix the docs build.

anwar6953:
Update md5 of recently updated files to fix the docs build.


awholegunch:
LGTM.

PR #3781, opened 2015-09-01
(CDAP-2976) Refactor ProgramController implementation

chtyim:
- Unify all state transition done and triggering of controller listener from one thread
- Use Id.Program instead of a String as the programId

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/ProgramController.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/ProgramController.java
@@ -16,6 +16,7 @@
 
 package co.cask.cdap.app.runtime;
 
+import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.ProgramRunStatus;
 import com.google.common.util.concurrent.ListenableFuture;
 import org.apache.twill.api.RunId;
@@ -103,7 +104,7 @@ public interface ProgramController {
 
     private final ProgramRunStatus runStatus;
 
-    private State(ProgramRunStatus runStatus) {
+    State(ProgramRunStatus runStatus) {
       this.runStatus = runStatus;
     }
 
@@ -116,7 +117,22 @@ public interface ProgramController {
     }
   }
 
+  /**
+   * Returns the program Id which this controller is controlling.
+   */
+  Id.Program getProgramId();
+
+  /**
+   * Returns the run Id which this controller is controlling.
+   */
   RunId getRunId();
+
+  /**
+   * Returns the component name within a given program which this controller is controlling
+   * or {@code null} if there is no component name for this controller.
+   */
+  @Nullable
+  String getComponentName();
 
   /**
    * Suspend the running {@link ProgramRunner}.
@@ -142,9 +158,9 @@ public interface ProgramController {
   /**
    * Adds a listener to watch for state changes. Adding the same listener again don't have any effect
    * and simply will get the same {@link Cancellable} back.
-   * @param listener
-   * @param executor
-   * @return
+   * @param listener listener for listening to state changes
+   * @param executor the executor used for making calls to the given listener
+   * @return a {@link Cancellable} to cancel the listening
    */
   Cancellable addListener(Listener listener, Executor executor);
 
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
@@ -138,7 +138,7 @@ public abstract class AbstractInMemoryProgramRunner implements ProgramRunner {
     public InMemoryProgramController(Table<String, Integer, ProgramController> components,
                                      RunId runId, Program program, ProgramSpecification spec,
                                      ProgramOptions options, ProgramRunnerFactory.Type type) {
-      super(program.getName(), runId);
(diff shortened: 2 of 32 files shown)


chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2689-1

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/ProgramController.java
```diff
@@ -116,9 +117,24 @@ public boolean isDone() {
...
+   */
   RunId getRunId();
 
   /**
+   * Returns the component name within a given program which this controller is controlling
+   * or {@code null} if there is no component name for this controller.
+   */
+  @Nullable
+  String getComponentName();
+
```
anew: what is a component? haven't seen that in cdap before

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
```diff
@@ -80,7 +80,8 @@ public AbstractInMemoryProgramRunner(CConfiguration cConf, ProgramRunnerFactory
   private void startComponent(Program program, String name, int instances, RunId runId, ProgramOptions options,
                                 Table<String, Integer, ProgramController> components, ProgramRunnerFactory.Type type) {
     for (int instanceId = 0; instanceId < instances; instanceId++) {
-      ProgramOptions componentOptions = createComponentOptions(name, instanceId, instances, runId, options);
+      ProgramOptions componentOptions = createComponentOptions(name + "-" + instanceId,
```
anew: isn't that redundant? So if I have a flowlet named "parser", then the name will be "parser-0" and the instance id "0" etc?

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/ProgramController.java
chtyim: This API is internal and component is for the case of controller component run of a program (e.g. Flowlet). The same terminology is being used in runtime service and program runner.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/AbstractInMemoryProgramRunner.java
chtyim: Maybe better to refactor the existing "createComponentOptions" method as well (current change is to just fit in what the current logic is, since the name is being passed as component name to the program controller).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractProgramController.java
```diff
@@ -116,10 +155,11 @@ public void run() {
     if (!state.compareAndSet(State.STARTING, State.STOPPING)
       && !state.compareAndSet(State.ALIVE, State.STOPPING)
       && !state.compareAndSet(State.SUSPENDED, State.STOPPING)) {
-      return Futures.immediateFailedFuture(new IllegalStateException("Stopping not allowed").fillInStackTrace());
+      return Futures.immediateFailedFuture(
+        new IllegalStateException("Stopping not allowed for " + name + " in " + state.get()).fillInStackTrace());
```
anew: I don't think the fillInStackTrace() is needed here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractProgramController.java
```diff
@@ -141,11 +181,11 @@ public void run() {
    */
   protected void complete() {
     if (!state.compareAndSet(State.ALIVE, State.COMPLETED)) {
-      LOG.debug("Cannot transit to COMPLETED state from {} state: {} {}", state.get(), programName, runId);
+      LOG.debug("Cannot transit to COMPLETED state from {} state: {} {}", state.get(), name);
       return;
     }
-    LOG.debug("Program stopped: {} {}", programName, runId);
-    executor(State.COMPLETED).execute(new Runnable() {
+    LOG.debug("Program controller stopped: {}", name);
```
anew: I think either all lifecycle transitions should log, or none should log. Here, it seems that only complete() logs.

chtyim: Agree. Looks like logs are added at different commits. Will take a pass on them.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/AbstractProgramController.java
```diff
@@ -260,10 +301,23 @@ public void execute(@Nonnull Runnable command) {
 
   protected abstract void doCommand(String name, Object value) throws Exception;
 
-  private Executor executor(State state) {
-    return executor(state.name());
+  /**
+   * Force this controller into error state and set the failure into the given future.
+   * This method should only gets called from the single thread executor of this class.
```
anew: gets/be

anew:
I have only minor comments left. Otherwise LGTM

chtyim:
Addressed comments. Please review again. Thanks.

chtyim:
Latest build passed: http://builds.cask.co/browse/CDAP-DUT2689

anew:
LGTM

chtyim:
Squashed commit. Merging.

PR #3782, opened 2015-09-01
adds box shadow and background-color to program buttons

astewes:
*  Adds box shadow to button bar
*  Adds green background to Services -> Make Request button when not disabled


ajainarayanan:
:+1: LGTM

PR #3784, opened 2015-09-02
CDAP-3435 Program status shows Running even after it is stopped.

henry-cask:
When doing fix for CDAP-3374 we added access to transaction within transaction to reduce logging messages which cause problems.

This PR removes the nested the call to transactions inside a transaction (call to retrieveProgramIdForRunRecord actually cause call to dataStore which has transaction) for the running run records fixer thread.
Add additional log to debug run records cleaning. Cache the program ids for run ids when filtering for valid program type.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
@@ -38,6 +38,8 @@ import co.cask.cdap.proto.ProgramRunStatus;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.RunRecord;
 import com.google.common.base.Predicate;
+import com.google.common.base.Throwables;
+import com.google.common.collect.Collections2;
 import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.AbstractIdleService;
 import com.google.inject.Inject;
@@ -48,6 +50,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


henry-cask:
Build: http://builds.cask.co/browse/CDAP-DUT2695-2

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -257,35 +259,63 @@ private void validateAndCorrectRunningRunRecords() {
...
+    LOG.debug("End getting {} run records not actually running.", notActuallyRunning.size());
+
+    final Map<String, Id.Program> runIdToProgramId = new HashMap<>();
+
+    LOG.debug("Start getting invalid run records  ...");
+    Collection<RunRecordMeta> invalidRunRecords =
+      Collections2.filter(notActuallyRunning, new Predicate<RunRecordMeta>() {
+        @Override
+        public boolean apply(@Nullable RunRecordMeta input) {
+          if (input == null) {
```
chtyim: can input be null?

henry-cask: Should not be. I will take out the null check for both filters.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -468,6 +492,8 @@ public void run() {
         RunRecordsCorrectorRunnable.LOG.debug("End correcting invalid run records.");
       } catch (Throwable t) {
         // Ignore any exception thrown since this behaves like daemon thread.
+        LOG.warn("Unable to complete correcting run records: {}",
+                 (t.getCause() != null ? t.getCause().getMessage() : t.getMessage()));
```
chtyim: Use `Throwables.getRootCause(t)`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
```diff
@@ -257,35 +259,57 @@ private void validateAndCorrectRunningRunRecords() {
...
-        return (targetProgramId != null) && !runIdToRuntimeInfo.containsKey(RunIds.fromString(runId));
+        // Check if it is not actually running.
+        return !runIdToRuntimeInfo.containsKey(RunIds.fromString(runId));
       }
     });
+    LOG.debug("End getting {} run records not actually running.", notActuallyRunning.size());
+
+    final Map<String, Id.Program> runIdToProgramId = new HashMap<>();
+
+    LOG.debug("Start getting invalid run records  ...");
```
chtyim: More like a trace to me.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
henry-cask: Sounds good

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ProgramLifecycleService.java
henry-cask: Will make the change.

henry-cask:
Last build passes: http://builds.cask.co/browse/CDAP-DUT2695-5

PR #3785, opened 2015-09-02
adds description to program detail views

astewes:
*  Adds description to program detail views
*  Adds breadcrumb box shadow


ajainarayanan:
:+1: LGTM

PR #3786, opened 2015-09-02
CDAP-3490 cdap upgrade for templates and adapters

albertshau:
We are removing templates and adapters, which means the upgrade
process should delete any existing adapters and templates.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2697

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -396,13 +354,19 @@ public void upgrade(boolean continueOnFailure) throws Exception {
           continue;
         }
 
+        // if this was etlbatch or etlrealtime template
+        if ("ETLBatch".equals(appSpec.getName()) || "ETLRealtime".equals(appSpec.getName())) {
```
gokulavasan: so the assumption here is, if the app is named ETLBatch, ETLRealtime it should be our template app?

gokulavasan:
One question. Rest LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
albertshau: yeah... there isn't really any other data to go on since templates were implemented using apps.

albertshau: though this logic could be moved into adapter service, where we would know it's a template.  Will do that.

gokulavasan:
LGTM :+1:

PR #3787, opened 2015-09-02
CDAP-3107: Initial log cleanup and refactoring of ProgramLifecycleHttpHandler for review

rohitsinha54:
- Removed redundant exception throwing for some methods and dependent methods in ProgramLifecycleHttpHandler
- Changed handlers exceptions goes through HttpExceptionHandler

Changes for handlers: 
- getMapReduceInfo
- getStatus

Opening for some feedback and will proceed with others.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
@@ -71,13 +71,9 @@ public class MRJobClient implements MRJobInfoFetcher {
 
     JobClient jobClient;
     JobStatus[] jobs;
-    try {
-      jobClient = new JobClient(hConf);
-      jobs = jobClient.getAllJobs();
-    } catch (Exception e) {
-      LOG.warn("JobClient failed to get all jobs.", e);
-      throw new IOException(e);
-    }
+
+    jobClient = new JobClient(hConf);
+    jobs = jobClient.getAllJobs();
 
     JobStatus thisJob = findJobForRunId(jobs, runId);
 
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
@@ -69,11 +69,8 @@ public class MRJobClient implements MRJobInfoFetcher {
   public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundException {
     Preconditions.checkArgument(ProgramType.MAPREDUCE.equals(runId.getProgram().getType()));
 
-    JobClient jobClient;
-    JobStatus[] jobs;
-
-    jobClient = new JobClient(hConf);
-    jobs = jobClient.getAllJobs();
+    JobClient jobClient = new JobClient(hConf);
+    JobStatus[] jobs = jobClient.getAllJobs();
 
     JobStatus thisJob = findJobForRunId(jobs, runId);
 
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
@@ -28,6 +28,7 @@ import co.cask.cdap.api.service.ServiceSpecification;
 import co.cask.cdap.api.service.http.HttpServiceHandlerSpecification;
 import co.cask.cdap.api.service.http.ServiceHttpEndpoint;
 import co.cask.cdap.api.workflow.WorkflowActionSpecification;
+import co.cask.cdap.common.NotFoundException;
 import co.cask.cdap.common.conf.Constants;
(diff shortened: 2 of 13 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
```diff
@@ -71,13 +71,9 @@ public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundExceptio
 
     JobClient jobClient;
```
bdmogal: no need to declare these two separately?

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
```diff
@@ -71,13 +71,9 @@ public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundExceptio
...
-    try {
-      jobClient = new JobClient(hConf);
-      jobs = jobClient.getAllJobs();
-    } catch (Exception e) {
-      LOG.warn("JobClient failed to get all jobs.", e);
-      throw new IOException(e);
-    }
+
+    jobClient = new JobClient(hConf);
+    jobs = jobClient.getAllJobs();
```
chtyim: So if getAllJobs() throws exception, what will happen?

cdap-common/src/main/java/co/cask/cdap/common/HttpExceptionHandler.java
```diff
@@ -34,12 +34,16 @@
   @Override
   public void handle(Throwable t, HttpRequest request, HttpResponder responder) {
     if (t instanceof BadRequestException) {
+      logWarnings(request, t);
```
chtyim: Shouldn't log warning in all these cases are they are error caused by user, meaning someone can have a script that keep performing invalid action to flood the server log.

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
rohitsinha54: getAllJobs can throw IOException when it gets a InterruptedException. We let it propagate up to its caller DistributedMRJobInfoFetcher where we have to catch it as in the case of failure we fail back to metric system. So, we catch it, log and return the info from metric system.

cdap-common/src/main/java/co/cask/cdap/common/HttpExceptionHandler.java
rohitsinha54: Yes you are correct. But isn't it required to log it so that we in a typical case of bad request or not found or something we can look into the logs as what was being requested etc and also the stack trace to debug it ?

chtyim: What I mean is don't log it as warning. If really need to log, it's more like a trace.

rohitsinha54:
These changes failed some tests : http://builds.cask.co/browse/CDAP-DUT2698-1
I will fix them, merge and then other handlers will be changed in the same way.

rohitsinha54:
Fixed unit tests in last commit. New build running here: http://builds.cask.co/browse/CDAP-DUT2698-3

cdap-common/src/main/java/co/cask/cdap/common/HttpExceptionHandler.java
```diff
@@ -52,4 +59,9 @@ public void handle(Throwable t, HttpRequest request, HttpResponder responder) {
       responder.sendStatus(HttpResponseStatus.INTERNAL_SERVER_ERROR);
     }
   }
+
+  private void logWarnings(HttpRequest request, Throwable t) {
+    LOG.trace("Error in handling request={} {} for user={}:", request.getMethod().getName(), request.getUri(),
```
chtyim: The method suggests that it is logging WARNING... but in fact it's logging trace. So, better rename the method.

rohitsinha54:
New build: http://builds.cask.co/browse/CDAP-DUT2698-5

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -443,9 +449,12 @@ public void testBatchStatus() throws Exception {
       "{'appId': 'WordCountApp', 'programType': 'Service', 'programId': 'WordFrequencyService'}," +
       "{'appId': 'WordCountApp', 'programType': 'Mapreduce', 'programId': 'VoidMapReduceJob'}]");
     returnedBody = readResponse(response, LIST_OF_JSONOBJECT_TYPE);
-    Assert.assertEquals("App: WordCountApp not found", returnedBody.get(0).get("error").getAsString());
-    Assert.assertEquals("App: WordCountApp not found", returnedBody.get(1).get("error").getAsString());
-    Assert.assertEquals("App: WordCountApp not found", returnedBody.get(2).get("error").getAsString());
+    Assert.assertEquals("'namespace:testnamespace2/application:WordCountApp' was not found.",
```
bdmogal: all these should also be changed, right?

bdmogal:
:+1:

PR #3788, opened 2015-09-02
CDAP-3561: PluginProperties should be in PluginConfig

gokulavasan:
Moving PluginProperties to ``PluginConfig`` from ``Config``

Build : http://builds.cask.co/browse/CDAP-DUT2682


chtyim:
LGTM

PR #3789, opened 2015-09-02
fixes colors and tag button padding

astewes:
*  Changes color of program-test variable
*  Adjusts size of program tags
*  Adds correct color for text-success and text-danger

program


elfenheart:
:+1: LGTM

PR #3791, opened 2015-09-02
Feature/ui url change to hydrator

ajainarayanan:
- Changes all hydrator stuff to come under /hydrator url
- Fixes adapter.drafts state
- Adds adapters.list state

TODO:
  To have this PR simple we have not changed anything else other than url (/hydrator). We have to rename everything to hydrator when all functionality


elfenheart:
:+1: LGTM

PR #3793, opened 2015-09-02
CDAP-2956 Switching to configurer style API for WorkflowAction

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-DUT2705

JIRA : https://issues.cask.co/browse/CDAP-2956

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/workflow/AbstractWorkflowAction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2014 Cask Data, Inc.
+ * Copyright © 2014-2015 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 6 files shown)


albertshau:
Are there any docs or examples that need to be changed along with this?  If not, lgtm

gokulavasan:
@albertshau The tests are failing in bamboo. So there might be additional changes to fix them. Will remove the on-hold tag and tag you once the PR is ready for review. Thanks!

gokulavasan:
@albertshau Bamboo build passed. Please review. I will open separate PR to change docs and examples.

albertshau:
:+1:

anew:
I do not understand the new API. 

We have a configure(WorkflowActionConfigurer) that calls the old configure() method. If a subcass overrides it, then it has to use configurer.setName()/setDescription() etc. If a subclass does not override it, then when should it call the new setName()/setDescription() of this class? Calling it in configure() will have no effect, because that has to return a spec, which is used to give the name, description etc.  So the new set methods are not useful for anything? 

Also, why do we deprecate getName/Description() and document that setName/Description() should be used instead. This whole change makes little sense to me. I would like to see an example of how it is supposed to be used.

PR #3796, opened 2015-09-02
CDAP-3407: Get cdap version in upgrade tool programmatically

rohitsinha54:


cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -96,8 +96,8 @@
    * Set of Action available in this tool.
    */
   private enum Action {
-    UPGRADE("Upgrades CDAP to 3.2\n" +
-              "  The upgrade tool upgrades the following: \n" +
+    UPGRADE("Upgrades CDAP to " + ProjectInfo.getVersion().toString() +
```
anwar6953: You have to call getVersion() once more, to get the string. Or implement the toString in Version.java.

anwar6953: Sorry, disregard my comment. I was looking at this Version class as reference:
https://github.com/caskdata/cdap/blob/develop/cdap-proto/src/main/java/co/cask/cdap/proto/Version.java

The Version class you're using here has a toString implemented.

anwar6953:
Why close?

rohitsinha54:
Verifying messages on cluster

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -260,6 +260,7 @@ private void doMain(String[] args) throws Exception {
             try {
               startUp();
               performUpgrade();
+              System.out.println("\nUpgrade completed successfully.\n");
```
anwar6953: do we need an extra newline?

rohitsinha54: Ya looks cleaner :)

bdmogal:
:+1:

PR #3797, opened 2015-09-03
CDAP-3560 change isSystem to a scope enum

albertshau:
Changing the isSystem boolean to an enum for artifact scope.
We will start out with system and user scopes, but this allows us
to add new scopes in the future and is consistent with other
concepts like metrics.

Also doing some cleanup so that ArtifactId is used in a few
classes instead of storing the individual id fields.

Also adding missing package-info file for artifacts.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
@@ -17,6 +17,7 @@
 package co.cask.cdap.gateway.handlers;
 
 import co.cask.cdap.api.artifact.ArtifactDescriptor;
+import co.cask.cdap.api.artifact.ArtifactScope;
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.templates.plugins.PluginClass;
 import co.cask.cdap.app.program.ManifestFields;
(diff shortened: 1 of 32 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2707

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
```diff
@@ -135,13 +137,19 @@ public void refreshSystemArtifacts(HttpRequest request, HttpResponder responder)
...
+    throws NamespaceNotFoundException, BadRequestException {
 
     Id.Namespace namespace = validateAndGetNamespace(namespaceId);
 
     try {
-      responder.sendJson(HttpResponseStatus.OK, artifactRepository.getArtifacts(namespace, includeSystem));
+      if (scope == null) {
+        responder.sendJson(HttpResponseStatus.OK, artifactRepository.getArtifacts(namespace, true));
+      } else {
+        ArtifactScope artifactScope = validateScope(scope);
```
gokulavasan: Can't we call validateAndGetNamespace to combine these two lines of code?

gokulavasan:
One comment. Rest LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ArtifactHttpHandler.java
albertshau: good point will do

gokulavasan:
LGTM :+1:

PR #3798, opened 2015-09-03
Allow RPM builds on OSX and other non-Linux systems

wolf31o2:
This tells `fpm` to send the `--target noarch-unknown-linux` option to `rpmbuild` which allows packages to be built on OSX and installed on Linux.


dereklwood:
needs to be added to the top-level ``pom.xml`` too, otherwise this will only affect the base ``cdap`` package).  Then, lgtm.

PR #3799, opened 2015-09-03
Fixes to Doc Build and Deployment

awholegunch:
These changes fix:
1. Errors in the deployment script, which created paths with two consecutive slashes in them ('//')
2. Makes major changes to the build scripts for docs, including:

- Changes destination of doc build from "build" to "target", matching maven conventions;
- Removes old functions no longer being used;
- Moves code between scripts so that common file has only code actually shared between build scripts;
- Removes colored log output on non-OS X machines;
- Changes the ``build.sh`` API to have an optimized ``all`` that shares a single Javadocs invocation between the Web and Github versions; and
- Cleans up errors and inconsistencies in internal naming and coding.

With these changes, a run on Bamboo can produce artifacts, and display them, without needing a staging server task. The other tasks in the Bamboo plan have been simplified, as only the CDAP repository is required .

Demonstration Runs:

- [Docs-only run, no Javadocs, ZIP and HTML artifacts](http://builds.cask.co/browse/CDAP-DOB1-2/artifact)
- [Docs-all run, Javadocs, ZIPs and HTML artifacts](http://builds.cask.co/browse/CDAP-DOB1-3/artifact)


cdap-docs/_common/common-build.sh
```diff
@@ -14,19 +14,12 @@
...
-# Build script for docs
-# Builds the docs (all except javadocs and PDFs) from the .rst source files using Sphinx
-# Builds the javadocs and copies them into place
-# Zips everything up so it can be staged
-# REST PDF is built as a separate target and checked in, as it is only used in SDK and not website
-# Target for building the SDK
-# Targets for both a limited and complete set of javadocs
-# Targets not included in usage are intended for internal usage by script
+# Common code for Build script for docs
+# Not called directly; included in either the main build.sh or the individual manual's build.sh
```
wolf31o2: This isn't visible in the PR, but you shouldn't have a shebang header in a shell file that isn't meant to be executed.

cdap-docs/_common/common-build.sh
```diff
@@ -75,32 +69,24 @@ else
...
-SDK_JAVADOCS="${PROJECT_PATH}/target/site/${APIDOCS}"
+API_JAVADOCS="${PROJECT_PATH}/target/site/${APIDOCS}"
 
 CHECK_INCLUDES="false"
-TEST_INCLUDES_LOCAL="local"
-TEST_INCLUDES_REMOTE="remote"
-if [ "x${3}" == "x" ]; then
-  TEST_INCLUDES="${TEST_INCLUDES_REMOTE}"
+
+if [[ "${OSTYPE}" == "darwin"* ]]; then
```
wolf31o2: This should be controlled by a flag, not the operating system where the code is being run. Don't assume that there may never be Darwin-based build hosts (we currently have one).

cdap-docs/_common/common-build.sh
```diff
@@ -127,9 +113,9 @@ function usage() {
   cd $PROJECT_PATH
   PROJECT_PATH=`pwd`
   echo "Build script for '${PROJECT_CAPS}' docs"
-  echo "Usage: ${SCRIPT} < option > [source test_includes]"
+  echo "Usage: ${SCRIPT} <option> [source]"
```
wolf31o2: How about "action" since it's not really an option (or optional) and it really describes the action you want the script to take.

cdap-docs/_common/common-build.sh
```diff
@@ -140,25 +126,39 @@ function usage() {
...
   echo
 }
 
+function echo_red_bold() {
+  if [[ "${OSTYPE}" == "darwin"* ]]; then
+    echo -e "${RED}${BOLD}${1}${NC}${2}"
+  else
+    echo "${1}${2}"
+  fi
+}
```
wolf31o2: The conditional here is redundant. The output would be the same with either echo. Also, isn't enabling backslash escapes non-safe if you don't know the input? You would need to check and sanitize the input, or ensure that no input has backslashes in it.

cdap-docs/_common/common-build.sh
```diff
@@ -166,40 +166,29 @@ function build_docs_google() {
...
 
-function build_javadocs_sdk() {
-  build_javadocs_api
-  copy_javadocs_sdk
-}
-
-function copy_javadocs_sdk() {
-  cd ${BUILD_PATH}/${HTML}
+function copy_javadocs_api() {
+  cd ${TARGET_PATH}/${HTML}
```
wolf31o2: Don't change working directory. It's unnecessary.

cdap-docs/_common/common-build.sh
```diff
@@ -166,40 +166,29 @@ function build_docs_google() {
...
   rm -rf ${JAVADOCS}
-  cp -r ${SDK_JAVADOCS} .
+  cp -r ${API_JAVADOCS} .
```
wolf31o2: `cp -r ${API_JAVADOCS} ${TARGET_PATH}/${HTML}`

cdap-docs/_common/common-build.sh
```diff
@@ -216,18 +205,18 @@ function build_license_pdfs() {
 }
 
 function copy_license_pdfs() {
-  cd ${BUILD_PATH}/${HTML}/${LICENSES}
+  cd ${TARGET_PATH}/${HTML}/${LICENSES}
   cp ${SCRIPT_PATH}/${LICENSES_PDF}/* .
```
wolf31o2: Same as above. No need to `cd` only to `cp`.

cdap-docs/_common/common-build.sh
```diff
@@ -236,12 +225,6 @@ function make_zip() {
   zip -qr ${ZIP_DIR_NAME}.zip ${PROJECT_VERSION}/* --exclude .DS_Store
```
wolf31o2: Why are we using a license-encumbered format for this artifact? This is an additional dependency that not all systems will have, just for building docs. It just *happens* that our build hosts all have a `zip` binary. If we're going to keep this, as is, then we need to add a capability to our build agents on `zip` and have the builds depend on agents with this capability. Hidden, undocumented dependencies are the devil.

cdap-docs/_common/common-build.sh
```diff
@@ -257,21 +240,31 @@ function build_web() {
...
 
+function build_extras() {
+  # Over-ride this function in guides where Javadocs or licenses are being built or copied.
+  # Currently performed in reference-manual
+  echo "No extras being built or copied."
+}
+
 function set_mvn_environment() {
+  cd ${PROJECT_PATH}
   if [ "$(uname)" == "Darwin" ]; then
```
wolf31o2: Be consistent. Use `${OSTYPE}` here and don't shell out to `uname` for this.

cdap-docs/_common/common-build.sh
```diff
@@ -279,12 +272,21 @@ function check_includes() {
 }
 
 function download_includes() {
+  local includes_dir=${1}
```
wolf31o2: Why add this line? It's unused in this function.

cdap-docs/_common/common-build.sh
```diff
@@ -279,12 +272,21 @@ function check_includes() {
...
   # $1 passed is the directory to which the downloaded files are to be written.
   # For an example of over-riding this function, see developer/build.sh
   echo "No includes to be downloaded."
 }
 
-function test_includes () {
+function test_includes_directory() {
+  local includes_dir=${1}
+  if [ ! -d "${includes_dir}" ]; then
```
wolf31o2: What if it's a file? This will pass and your mkdir below will fail. If this is in a transitive directory structure created by the run, it should be safe for you to `rm -rf ${includes_dir} && mkdir ${includes_dir}`

cdap-docs/_common/common-build.sh
```diff
@@ -511,13 +490,10 @@ function run_command() {
...
-    build-standalone )  build_standalone;;
     copy-javadocs )     copy_javadocs;;
     copy-license-pdfs ) copy_license_pdfs;;
-    javadocs )          build_javadocs_sdk;;
+    javadocs-api )      build_javadocs_api;;
     javadocs-full )     build_javadocs_full;;
-    depends )           build_dependencies;;
-    sdk )               build_sdk;;
     version )           display_version;;
     * )                 usage;;
```
wolf31o2: This whole section could be simpler if you had a 1:1 command:function naming... perhaps with `s/-/_/` applied...

fex.
```
check-includes|docs|copy-javadocs|copy-license-pdfs) ${1/-/_};;
*) usage;;
```

cdap-docs/_common/common-build.sh
```diff
@@ -257,21 +240,31 @@ function build_web() {
...
+    # TODo: hard-coded Java version 1.7
     export JAVA_HOME=$(/usr/libexec/java_home -v 1.7)
   fi
+  # check BUILD.rst for changes
+  BUILD_RST_PATH="${PROJECT_PATH}/${BUILD_RST}"
+  test_an_include ${BUILD_RST_HASH} ${BUILD_RST_PATH}
 }
 
 function check_includes() {
   if [ "${CHECK_INCLUDES}" == "${TRUE}" ]; then
```
wolf31o2: Why compare string to string here? Why even have TRUE as a constant? Why not just evaluate CHECK_INCLUDES as a boolean?

cdap-docs/_common/common-build.sh
```diff
@@ -279,12 +272,21 @@ function check_includes() {
...
   # For an example of over-riding this function, see developer/build.sh
   echo "No includes to be downloaded."
 }
 
-function test_includes () {
+function test_includes_directory() {
+  local includes_dir=${1}
+  if [ ! -d "${includes_dir}" ]; then
+    echo "Creating Includes Directory: ${includes_dir}"
+    mkdir ${includes_dir}
```
wolf31o2: Also, you're guarding the test against spaces in the path, by putting it in quotes, but not doing the same here. If you know there won't be spaces in the path, drop the quotes in the test. Otherwise, add quotes here.

cdap-docs/build.sh
```diff
@@ -30,6 +30,34 @@ ARG_1="${1}"
 ARG_2="${2}"
 ARG_3="${3}"
 
+function usage() {
+  cd ${PROJECT_PATH}
+  PROJECT_PATH=`pwd`
```
wolf31o2: Pretty sure this is wrong. If it isn't, you should use a different variable name if this can have different values at different times.

cdap-docs/build.sh
```diff
@@ -30,6 +30,34 @@ ARG_1="${1}"
 ARG_2="${2}"
 ARG_3="${3}"
```
wolf31o2: Positional parameters are positional parameters. There's no need to wrap them in quotes, even if they includes spaces and are quoted in the input.

cdap-docs/build.sh
```diff
@@ -30,6 +30,34 @@ ARG_1="${1}"
...
 
+function usage() {
+  cd ${PROJECT_PATH}
+  PROJECT_PATH=`pwd`
+  echo "Build script for '${PROJECT_CAPS}' docs"
```
wolf31o2: Do we need two constants for the project, when this is part of CDAP?

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
-  echo "    clean          Clean up (previous builds)"
-  echo ""
-  echo "  with"
-  echo "    source         Path to ${PROJECT} source, if not ${PROJECT_PATH}"
-  echo "    test_includes  local, remote or neither (default: remote); must specify source if used"
-  echo ""
-}
-
 function run_command() {
   case "${1}" in
```
wolf31o2: No need to quote. Best practice is to be conservative with quotes and only use them where necessary.

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
-  cp -r ${1}/${BUILD}/${HTML} ${BUILD}/${HTML}/${1}
+  clean_all_targets
+  run_command javadocs 
+  run_command docs-first-pass ${ARG_2} 
+  run_command docs-github-part ${ARG_2} 
+  _stash_github_zip 
+  run_command docs-web-part ${ARG_2} 
+  _restore_github_zip
+  echo "--------------------------------------------------------"
+  _bell "Completed \"build all\""
```
wolf31o2: Single quotes on the outside to avoid escaping quotes on the inside.

cdap-docs/build.sh
```diff
@@ -30,6 +30,34 @@ ARG_1="${1}"
...
+  echo "    clean     Clean up any previous build's target directories"
+  echo "    javadocs  Build Javadocs"
+  echo "    licenses  Clean build of License Dependency PDFs"
+  echo "    sdk       Build CDAP SDK"
+  echo "    version   Print the version information"
+  echo ""
+  echo "  with"
+  echo "    source    Path to ${PROJECT} source, if not '${PROJECT_PATH}'"
+  echo "              Path is relative to '${SCRIPT_PATH}/../..'"
+  echo ""
```
wolf31o2: `echo` gives the same as `echo ""`

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
-    sphinx-build -D googleanalytics_id=${1} -D googleanalytics_enabled=1 -b html -d build/doctrees build/source build/html
+  if [ "${doc_type}" != "${DOCS}" ]; then
+    clean_all_targets
+  fi
+  run_command docs-first-pass ${source_path}
+  if [ "${doc_type}" == "${DOCS}" ]; then
+    build_docs_outer_level ${source_path} 
+    copy_docs_inner_level
+  fi
+  local warnings="$?"
```
wolf31o2: No quotes.

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
+  if [ "${warnings}" != "0" ]; then
```
wolf31o2: Compare int to int. No need to convert to string.

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
-  echo "Building Web Docs."
-  echo "--------------------------------------------------------"
-  echo ""
-  run_command docs-web-part ${ARG_2} ${ARG_3}
   echo ""
+}
+
+function build_javadocs_api() {
+  set_mvn_environment
+  MAVEN_OPTS="-Xmx1024m -XX:MaxPermSize=128m" mvn clean install -P examples,templates,release -DskipTests -Dgpg.skip=true && mvn clean site -DskipTests -P templates -DisOffline=false
```
wolf31o2: What are we doing here, especially given we're running `clean` in both Maven invocations?

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
+  echo "Completed Building First Pass of Docs"
   echo "========================================================"
   echo ""
-  exit 0
-}
-
-function build_javadocs() {
-  # Uses function from common
-  build_javadocs_api
+  return $?
```
wolf31o2: This will return 0, always. The previous command is `echo`... maybe you meant to save the return value of a previous command and return with that?

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
 }
 
-function build_docs_javadocs() {
-  build_docs_inner_level "build"
+function build_docs_github_part() {
+  echo "========================================================"
+  echo "Building GitHub Docs"
+  echo "--------------------------------------------------------"
+  echo ""
+  _build_docs "build-github" ${GOOGLE_ANALYTICS_GITHUB} ${GITHUB} ${FALSE}
```
wolf31o2: No quotes.

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
-function build_docs() {
-  _build_docs "docs" ${GOOGLE_ANALYTICS_WEB} ${WEB} ${TRUE}
-  return $?
+function _stash_github_zip() {
+  echo "========================================================"
+  echo "Stashing GitHub Zip"
+  echo "========================================================"
+  echo ""
+  cd ${SCRIPT_PATH}
+  TARGET_TEMP=$(mktemp -d /tmp/cdap-docs.XXXX)
```
wolf31o2: Why are you writing outside of the build area? This breaks any kind of sandboxing of the build. Use a relative path for this.

cdap-docs/build.sh
```diff
@@ -38,227 +66,197 @@ function set_project_path() {
...
-function build_docs_github() {
-  _build_docs "build-github" ${GOOGLE_ANALYTICS_GITHUB} ${GITHUB} ${FALSE}
-  return $?
+function _restore_github_zip() {
+  echo "========================================================"
+  echo "Restoring GitHub Zip"
+  echo "========================================================"
+  echo ""
+  mv ${TARGET_TEMP}/*.zip ${SCRIPT_PATH}/${TARGET}
+  rm -rf ${TARGET_TEMP}
```
wolf31o2: Rather than doing cleanup inside functions where a previous command could fail, it would make sense to instead trap signals and have an exit handler which performs cleanup.

```
trap trap_cleanup SIGHUP SIGQUIT SIGINT SIGTERM SIGKILL
function trap_cleanup() { cleanup; exit 1; }
function cleanup() {
  # perform cleanups here
}
```

cdap-docs/build.sh
```diff
@@ -275,53 +273,145 @@ function _build_docs() {
...
+  rewrite ${source_404} "src=\"_static"  "src=\"/cdap/${project_dir}/en/_static"
+  rewrite ${source_404} "src=\"_images"  "src=\"/cdap/${project_dir}/en/_images"
+  rewrite ${source_404} "/href=\"http/!s|href=\"|href=\"/cdap/${project_dir}/en/|g"
+  rewrite ${source_404} "action=\"search.html"  "action=\"/cdap/${project_dir}/en/search.html"
+  echo ""
 }
 
-function build_zip() {
+function _build_zip() {
   cd $SCRIPT_PATH
```
wolf31o2: Missing braces

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh
awholegunch: Fixed. Color is now set by environment variable rather than OS, and no longer uses the '-e' flag in echo statements.

cdap-docs/_common/common-build.sh
awholegunch: Why? Because that's what I was asked to do. People requested a zip file of the documentation.

cdap-docs/_common/common-build.sh
awholegunch: Because this function is a template to be over-shadowed in the inner manuals that use this. I was thinking I could then drop the comment, as this documents what $1 is.

cdap-docs/_common/common-build.sh
awholegunch: Removed function, as it isn't needed. The directory is always created right before this step, so it is a redundant check.

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh
awholegunch: Deleted the function.

cdap-docs/build.sh
awholegunch: Fixed.

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh
awholegunch: Fixed: removed.

cdap-docs/build.sh
awholegunch: Fixed.

cdap-docs/build.sh
awholegunch: Similar versions of this code are used in Coopr and Tigon, hence the usage.

cdap-docs/build.sh
awholegunch: Fixed.

cdap-docs/build.sh
awholegunch: I thought for sure I ran into a problem with using ``echo`` versus ``echo ""``, but...

cdap-docs/build.sh
awholegunch: Fixed.

cdap-docs/build.sh
awholegunch: Fixed.

cdap-docs/build.sh
awholegunch: set_mvn_environment:
- Sets the directory that the command is invoked from
- On OS X (darwin), uses a specific version of Java
- Checks to see if the BUILD.rst file has changed: in that case, the command may need updating, as that is the source of this command.

cdap-docs/build.sh
awholegunch: Yes. I added the echo statements and forgot to capture the value. Thanks.

cdap-docs/build.sh
awholegunch: Fixed.

cdap-docs/build.sh
awholegunch: Fixed.

cdap-docs/build.sh
awholegunch: I was writing outside the build area because the build area was getting blown away between invocations. But I think that with the revisions, this may no longer be required (the stashing between invocations.) I will test and eliminate this somehow.

awholegunch:
Now on to testing...

cdap-docs/_common/common-build.sh
awholegunch: What would you suggest as a replacement? ``tar.gz``? That might not be acceptable on Windows.

cdap-docs/build.sh
awholegunch: That didn't always work: 
- ``'Building "${i}", target "${1}"...'`` didn't work, while 
- ``"Building \"${i}\", target \"${1}\"..."`` does...

cdap-docs/build.sh
awholegunch: This will probably just go away...

cdap-docs/_common/common-build.sh
wolf31o2: Ahh, so we're distributing the zip, not just using it, internally. Got it. That's a good enough reason. ZIP as a format is far more widespread and accessible.

cdap-docs/_common/common-build.sh
wolf31o2: Sweet. :smile:

cdap-docs/build.sh
wolf31o2: Fair enough. We'll keep it as-is until we revamp for the new build after 3.2 is released, then we'll port that work to the other projects.

cdap-docs/build.sh
wolf31o2: Correct. When there's variables, you need to use double quotes for them to be interpreted. Simple strings like these can be single-quoted... or is that not working after being passed to the function?

cdap-docs/build.sh
wolf31o2: Sorry, I meant the actual two `mvn` calls. Both have clean in them. Are we simply trying to fill the Maven cache with the first? Anything built will otherwise be removed by the clean in the next `mvn` command, so I am not sure why we're doing them both this way.

cdap-docs/build.sh
wolf31o2: Ahh, yeah. The other solution would be to build under build and save the artifacts under target. Then, you can wipe out build and still keep previous artifacts. This is how `mvn` does things, unless you run clean, which will wipe target, also.

awholegunch: I will need a tmp area, as the stashing between invocations is still required. But I will modify the code so this happens within the build area.

awholegunch:
I have updated this, fixed all the issues mentioned above, and in particular have removed use of ``/tmp``, and so everything happens inside ``target`` directories. Removed changing of the current directory wherever possible.

Tested both locally and on Bamboo.

cdap-docs/_common/common-build.sh
```diff
@@ -14,19 +12,16 @@
...
-# Builds the javadocs and copies them into place
-# Zips everything up so it can be staged
-# REST PDF is built as a separate target and checked in, as it is only used in SDK and not website
-# Target for building the SDK
-# Targets for both a limited and complete set of javadocs
-# Targets not included in usage are intended for internal usage by script
+# Common code for Build script for docs
+# Not called directly; included in either the main build.sh or the individual manual's build.sh
+
+# Optional Parameters (passed via Bamboo env variable or exported in shell):
```
wolf31o2: Remove Bamboo

cdap-docs/_common/common-build.sh
```diff
@@ -124,82 +112,62 @@ BUILD_RST_HASH="f54ae74bb72f9ad894766b6c0bd2d2df"
...
-  echo "    check-includes Check if included files have changed from source"
-  echo "    depends        Build Site listing dependencies"
-  echo "    sdk            Build SDK"
+  echo "    license-pdfs         Clean build of License Dependency PDFs"
+  echo "    check-includes       Check if included files have changed from source"
+  echo "    display-version      Print the version information"
   echo "  with"
-  echo "    source         Path to $PROJECT source for javadocs, if not $PROJECT_PATH"
-  echo "    test_includes  local, remote or neither (default: remote); must specify source if used"
+  echo "    source               Path to $PROJECT source for javadocs, if not '$PROJECT_PATH'"
```
wolf31o2: Add braces

cdap-docs/_common/common-build.sh
```diff
@@ -257,21 +218,31 @@ function build_web() {
...
   fi
+  # check BUILD.rst for changes
+  BUILD_RST_PATH="${PROJECT_PATH}/${BUILD_RST}"
+  test_an_include ${BUILD_RST_HASH} ${BUILD_RST_PATH}
 }
 
 function check_includes() {
-  if [ "${CHECK_INCLUDES}" == "${TRUE}" ]; then
-    echo "Downloading and checking includes."
+  if [ !"${CHECK_INCLUDES}" ]; then
```
wolf31o2: It looks like this changed the meaning of the check. Can you verify that this is correct?

wolf31o2:
Couple minor comments and one question... otherwise, :+1: LGTM

cdap-docs/_common/common-build.sh
awholegunch: Fixed (two places.)

cdap-docs/_common/common-build.sh

cdap-docs/_common/common-build.sh
awholegunch: Good catch. I went back to the old check, and checked all usages of this variable.

awholegunch:
Running another test on Bamboo to check latest version...

awholegunch:
When running additional tests, I found that messages from the Javadoc portion of the build were not being retained and replayed at the end, and that the return code was not being passed to Bamboo successfully. This has been fixed in the last two commits.

awholegunch:
The build is now successfully failing!

cdap-docs/build.sh
```diff
@@ -38,321 +72,379 @@ function set_project_path() {
...
+    run_command javadocs
+  fi
+  if [ "${doc_type}" == "${GITHUB}" -o "${doc_type}" == "${GITHUB_ONLY}" ]; then
+    run_command docs-github-part ${source_path}
+  elif [ "${doc_type}" == "${WEB}" -o "${doc_type}" == "${WEB_ONLY}" ]; then
+    run_command docs-web-part ${source_path}
+  fi
+  display_version
+  display_messages_file
+  warnings=$?
```
wolf31o2: Return code from `display_messages_file`?

wolf31o2:
One minor comment... otherwise, :+1: LGTM

cdap-docs/build.sh
awholegunch: Yes...``display_messages_file`` determines the return code based on the contents of the message file. Once all the messages have been displayed at what will become the end of the log, it then returns the code and Bamboo can decide what to do with the results. This way, when you go to the Bamboo page, you should be able to see at a glance that there are error messages to be reviewed, and which manuals to poke into.

PR #3800, opened 2015-09-03
Fix integration test setup to ensure that dataset service is up, but …

bdmogal:
…also that the default namespace exists. Also improve exception handling in the checkServicesWithRetry method, to check for both the root cause and the exception itself to be an expected exception.

Build: http://builds.cask.co/browse/CDAP-DUT2711

### DIFF
--- a/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
+++ b/cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
@@ -29,8 +29,6 @@ import co.cask.cdap.client.StreamClient;
 import co.cask.cdap.client.config.ClientConfig;
 import co.cask.cdap.client.config.ConnectionConfig;
 import co.cask.cdap.client.util.RESTClient;
-import co.cask.cdap.common.NotFoundException;
-import co.cask.cdap.common.ProgramNotFoundException;
 import co.cask.cdap.common.UnauthorizedException;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
@@ -40,8 +38,6 @@ import co.cask.cdap.proto.ConfigEntry;
 import co.cask.cdap.proto.DatasetSpecificationSummary;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.NamespaceMeta;
-import co.cask.cdap.proto.ProgramRecord;
-import co.cask.cdap.proto.ProgramType;
 import co.cask.cdap.proto.StreamDetail;
 import co.cask.cdap.security.authentication.client.AccessToken;
 import co.cask.cdap.security.authentication.client.AuthenticationClient;
@@ -53,7 +49,6 @@ import com.google.common.base.Predicate;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
-import org.apache.twill.filesystem.LocalLocationFactory;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
@@ -66,18 +61,16 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.util.List;
-import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import javax.annotation.Nullable;
-import javax.ws.rs.ServiceUnavailableException;
 
 /**
- *
+ * Abstract class for writing Integration tests for CDAP. Provides utility methods to use in Integration tests.
+ * Users should extend this class in their test classes.
  */
-public class IntegrationTestBase {
+public abstract class IntegrationTestBase {
   private static final Logger LOG = LoggerFactory.getLogger(IntegrationTestBase.class);
   private static final long SERVICE_CHECK_TIMEOUT = TimeUnit.MINUTES.toSeconds(10);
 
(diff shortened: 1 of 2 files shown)


cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -161,10 +160,14 @@ private void checkServicesWithRetry(Callable<Boolean> callable,
         }
         TimeUnit.SECONDS.sleep(1);
       } catch (Throwable e) {
-        if (!(e.getCause() instanceof InterruptedException || e.getCause() instanceof IOException)) {
-          // we want to throw UnauthorizedException while suppress and retry on Interrupted or IOException
+        // we want to throw UnauthorizedException while suppress and retry on Interrupted or IOException
```
sagarkapare: Maybe instead of catching Throwable, use -
```java
} catch (InterruptedException | IOException e) {
 // suppress InterruptedException and IOException
}
```

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -101,8 +95,12 @@ public Boolean call() throws Exception {
         if (!getMonitorClient().allSystemServicesOk()) {
           return false;
         }
-        // check that the dataset service is up
-        getNamespaceClient().list();
+        // check that the dataset service is up, and also that the default namespace exists
+        try {
+          getNamespaceClient().get(Id.Namespace.DEFAULT);
```
gsps1: is this needed? if dataset service is not up, other services would not be up right ?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -110,7 +108,8 @@ public Boolean call() throws Exception {
     try {
       checkServicesWithRetry(monitorCallable, "CDAP Services are not available");
     } catch (Throwable e) {
```
sagarkapare: just catch the UnauthorizedException here, since we are handling that only.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
```diff
@@ -119,7 +118,7 @@ public Boolean call() throws Exception {
           throw Throwables.propagate(ex);
```
sagarkapare: Existing code, but can we remove the catch block here?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
bdmogal: We have seen that all services return OK, but dataset service is not up. That causes failures in integration tests. The failures are only in the first few tests in a build, then the service comes up, then tests start passing.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
bdmogal: can't, right? the ``UnauthorizedException`` is the ``rootCause``, not the exception itself?

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
bdmogal: hmmm, ``fetchAccessToken`` throws ``IOException``, and the logic is to only throw ``ServiceUnavailableException`` out of this method. Maybe we can just catch ``IOException``?

sagarkapare: I see. Then yes you are right, we can just catch the IOException. 
```java
catch (IOException e) {
   throw new ServiceUnavailableException("Failed to access the token.", e);
}
```

bdmogal: changed it to catch ``IOException``. However, still unsure if we should throw a ``ServiceUnavailableException`` in this case. Will think about it.

bdmogal: I don't think we should throw a ``ServiceUnavailableException``. That exception is for a service to indicate to clients that the service is unavailable, with an HTTP-like response code. Not sure if it makes sense to throw that from something that's not a service like ``IntegrationTestBase``. In any case, ``ServiceUnavailableException`` is a ``RuntimeException``, so the effect is similar to ``Throwables.propagate(e)``.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestBase.java
gsps1: hmm, then we need to have a JIRA for that issue and figure out the issue. maybe you can add a todo here with that JIRA.

bdmogal: Sure, but either way, this logic will be needed to ensure that the Default namespace is available before starting tests. The MonitorClient#allServicesOk will not guarantee that.

gsps1:
just add a jira about this issue - https://github.com/caskdata/cdap/pull/3800#discussion_r38704277 - otherwise lgtm

sagarkapare:
LGTM

bdmogal:
I am going to merge this, since this test only affects ``IntegrationTestBaseTest``, which passes locally. Its  mainly affects integration tests, which I have tested to pass.

PR #3811, opened 2015-09-03
(CDAP-3579) Reduce locking scope of the AbstractProgramRuntimeService.run() method

chtyim:
- Locking is only required when modifying the runtimeInfos map, no need to lock the entire time inside the run() method
- If we lock the whole run() method, it’s possible that the program finish very fast such that while run() is holding the lock, the ProgramController listener.completed() method is triggered (from another thread), which also tries to acquire the lock to modify the runtimeInfos map, causing a deadlock.


anew:
LGTM and thx for the test case.

chtyim:
Tested locally for the affected test case. Merging to unblock other branches.

PR #3813, opened 2015-09-03
CDAP-3580 deserialize scope enum in a case insensitive way

albertshau:
Adding a type adapter factory for deserializing enums in a case
insensitive way and serializing enums in lowercase. This makes it
a little easier to use the deploy app api since artifact scope
no longer needs to be all caps.

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/io/CaseInsensitiveEnumTypeAdapterFactory.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/io/CaseInsensitiveEnumTypeAdapterFactory.java
@@ -36,12 +36,13 @@ import java.util.Map;
  */
 public class CaseInsensitiveEnumTypeAdapterFactory implements TypeAdapterFactory {
   public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+    @SuppressWarnings("unchecked")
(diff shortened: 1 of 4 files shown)


cdap-common/src/main/java/co/cask/cdap/common/io/CaseInsensitiveEnumTypeAdapterFactory.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Map;
+
+/**
+ * Type adapter that serializes enums in lowercase and deserializes enums in a case insensitive way.
+ * Taken straight from the example in TypeAdapterFactory javadocs except deserializes in a case insenstivie way
+ * instead of deserializing lowercase only.
+ */
+public class CaseInsensitiveEnumTypeAdapterFactory implements TypeAdapterFactory {
+  public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+    Class<T> rawType = (Class<T>) type.getRawType();
```
chtyim: Add a `@SuppressWarnings("unchecked")` to suppress java compiler warnings.

chtyim:
Just one comment. Rest LGTM.

albertshau:
http://builds.cask.co/browse/CDAP-DUT2725

PR #3819, opened 2015-09-03
Return 503(Service unavailable) status code when the dataset service is not discoverable.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-3470

When dataset service is not yet running,  calls made to the Http handlers that depends on the dataset service fail. While failing we log the stack trace which causes master log to get flooded with the error messages. 

Following changes are made as a part of the PR:
1. Added checked exception `ServiceNotRunningException` which is thrown by the `DatasetServiceClient` when the dataset service is not discoverable. The reason for adding this as a checked exception is that the client of the exception can catch it and have retry mechanism if required.

2. In `HttpExceptionHandler`, check if the causal chain of the input exception contains `ServiceNotRunningException`. If it does then return the `503` error code without logging anything.

3. Changed the implementation of the `NamespaceHttpHandler.getAllNamespace` as a prototype to throw exception which would be caught by the `HttpExceptionHandler`. Will fix the other handlers in separate PRs.


sagarkapare:
@chtyim @anew @poornachandra Please help review. Thanks!

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetServiceClient.java
```diff
@@ -122,17 +123,7 @@ public DatasetMeta getInstance(String instanceName) throws DatasetManagementExce
     return GSON.fromJson(response.getResponseBodyAsString(), SUMMARY_LIST_TYPE);
   }
 
-  public Collection<DatasetModuleMeta> getAllModules() throws DatasetManagementException {
```
poornachandra: If service has get all modules, the we should retain the client call too.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DatasetServiceUndiscoverableException.java
```diff
@@ -0,0 +1,30 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data2.dataset2;
+
+/**
+ * Exception thrown when the Dataset service is not discoverable.
+ */
+public class DatasetServiceUndiscoverableException extends DatasetManagementException {
```
poornachandra: This is no longer required, I think.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/upgrade/MetricsDataMigrator.java
```diff
@@ -417,7 +418,7 @@ private MetricsTable getOrCreateMetricsTable(String tableName, DatasetProperties
     try {
       table = DatasetsUtil.getOrCreateDataset(dsFramework, metricsDatasetInstanceId,
                                               MetricsTable.class.getName(), empty, null, null);
-    } catch (DatasetManagementException e) {
+    } catch (DatasetManagementException | ServiceNotRunningException e) {
       String msg = String.format("Cannot access or create table %s.", tableName);
       LOG.error(msg);
```
poornachandra: This can be a warn or lesser.

bdmogal:
Should we use http://docs.oracle.com/javaee/7/api/javax/ws/rs/ServiceUnavailableException.html

sagarkapare:
This is modeled after the existing exception classes in the CDAP. 
Also `ServiceUnavailableException` mentioned in the above link is unchecked exception, while I thought its better to have our `ServiceUnavailableException` as checked one. So that compiler warns user to explicitly handle it.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/store/upgrade/MetricsDataMigrator.java
sagarkapare: Fixed. Made as warn.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DatasetServiceUndiscoverableException.java
sagarkapare: Removed.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetServiceClient.java
sagarkapare: Added it back.

sagarkapare:
@poornachandra @bdmogal Addressed/answered comments. Please take a look again. Thanks!

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT2728-3

poornachandra:
LGTM, please squash commits and merge.

sagarkapare:
Merged the squashed PR #3840 . Closing this.

PR #3821, opened 2015-09-03
CDAP-3390: View HTTP API, client, and CLI

jwang47:
https://issues.cask.co/browse/CDAP-3390
http://builds.cask.co/browse/CDAP-DUT2709

Tested manually using the CLI against a standalone CDAP.

Depends on https://github.com/caskdata/cdap/pull/3835

### DIFF
--- a/cdap-api-common/src/main/java/co/cask/cdap/api/data/format/FormatSpecification.java
+++ b/cdap-api-common/src/main/java/co/cask/cdap/api/data/format/FormatSpecification.java
@@ -22,6 +22,7 @@ import co.cask.cdap.api.data.schema.Schema;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
+import javax.annotation.Nullable;
 
 /**
  * Specification for a {@link RecordFormat}, including the class, schema, and settings to use for the format.
@@ -37,16 +38,21 @@ public final class FormatSpecification {
     this(null, null, Collections.<String, String>emptyMap());
   }
 
-  public FormatSpecification(String name, Schema schema, Map<String, String> settings) {
+  public FormatSpecification(String name, @Nullable Schema schema, @Nullable Map<String, String> settings) {
     this.name = name;
     this.schema = schema;
     this.settings = settings == null ? Collections.<String, String>emptyMap() : settings;
+  }
+
+  public FormatSpecification(String name, @Nullable Schema schema) {
+    this(name, schema, Collections.<String, String>emptyMap());
   }
 
   public String getName() {
     return name;
   }
 
+  @Nullable
   public Schema getSchema() {
     return schema;
   }
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/view/InMemoryViewStore.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/view/InMemoryViewStore.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data.view;
+
+import co.cask.cdap.common.NotFoundException;
+import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.ViewDetail;
+import co.cask.cdap.proto.ViewSpecification;
+import com.google.common.collect.HashBasedTable;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Table;
+
+import java.util.List;
+
+/**
+ * In-memory implementation of {@link ViewStore}.
+ */
+public final class InMemoryViewStore implements ViewStore {
+
+  private final Table<Id.Stream.View, Id.Stream, ViewSpecification> views;
+
+  public InMemoryViewStore() {
+    views = HashBasedTable.create();
+  }
+
+  @Override
+  public boolean createOrUpdate(Id.Stream.View viewId, ViewSpecification config) {
+    boolean created = !views.containsRow(viewId);
+    views.put(viewId, viewId.getStream(), config);
+    return created;
+  }
+
+  @Override
+  public boolean exists(Id.Stream.View viewId) {
+    return views.contains(viewId, viewId.getStream());
+  }
+
+  @Override
+  public void delete(Id.Stream.View viewId) throws NotFoundException {
+    ViewSpecification removed = views.remove(viewId, viewId.getStream());
+    if (removed == null) {
+      throw new NotFoundException(viewId);
+    }
+  }
+
+  @Override
+  public List<Id.Stream.View> list(Id.Stream streamId) {
+    return ImmutableList.<Id.Stream.View>builder().addAll(views.column(streamId).keySet()).build();
+  }
+
+  @Override
+  public ViewDetail get(Id.Stream.View viewId) throws NotFoundException {
+    if (!views.containsRow(viewId)) {
+      throw new NotFoundException(viewId);
+    }
+
+    return new ViewDetail(viewId.getId(), views.get(viewId, viewId.getStream()));
+  }
+}
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/view/ViewStore.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/view/ViewStore.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.data.view;
+
+import co.cask.cdap.common.NotFoundException;
+import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.ViewDetail;
+import co.cask.cdap.proto.ViewSpecification;
+
+import java.util.List;
+
+/**
+ * Interface for storing stream views.
+ */
+public interface ViewStore {
+
+  /**
+   * Creates a view. Updates the view if it already exists.
+   * @param viewId the view
+   * @param config the view config
+   * @return true if a new view was created
+   */
+  boolean createOrUpdate(Id.Stream.View viewId, ViewSpecification config);
+
+  /**
+   * @param viewId the view
+   * @return true if the view exists
+   */
+  boolean exists(Id.Stream.View viewId);
+
+  /**
+   * Deletes a view.
+   *
+   * @param viewId the view
+   */
+  void delete(Id.Stream.View viewId) throws NotFoundException;
+
+  /**
+   * @param streamId the stream
+   * @return list of view IDs for a stream
+   */
+  List<Id.Stream.View> list(Id.Stream streamId);
+
+  /**
+   * @param viewId the view
+   * @return the details of a view
+   */
+  ViewDetail get(Id.Stream.View viewId) throws NotFoundException;
+}
--- a/cdap-explore-client/src/main/java/co/cask/cdap/explore/client/ExploreFacade.java
+++ b/cdap-explore-client/src/main/java/co/cask/cdap/explore/client/ExploreFacade.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2014 Cask Data, Inc.
+ * Copyright © 2015 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
--- a/cdap-explore-client/src/main/java/co/cask/cdap/explore/client/MockExploreClient.java
+++ b/cdap-explore-client/src/main/java/co/cask/cdap/explore/client/MockExploreClient.java
@@ -14,19 +14,17 @@
  * the License.
  */
 
-package co.cask.cdap.explore.jdbc;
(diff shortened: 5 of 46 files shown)


cdap-api-common/src/main/java/co/cask/cdap/api/data/format/FormatSpecification.java
```diff
@@ -37,12 +38,16 @@ private FormatSpecification() {
     this(null, null, Collections.<String, String>emptyMap());
   }
 
-  public FormatSpecification(String name, Schema schema, Map<String, String> settings) {
+  public FormatSpecification(String name, Schema schema, @Nullable Map<String, String> settings) {
```
chtyim: Why null? Better not null and have require empty map.

jwang47: It was nullable before, but not annotated. I'd prefer to leave it as is for this PR.

cdap-cli/src/main/java/co/cask/cdap/cli/command/view/CreateOrUpdateViewCommand.java
```diff
@@ -0,0 +1,123 @@
...
+      .append(Joiner.on(", ").join(Formats.ALL))
+      .append(". <")
+      .append(ArgumentName.SCHEMA)
+      .append("> is a sql-like schema \"column_name data_type, ...\" or Avro-like JSON schema and <")
+      .append(ArgumentName.SETTINGS)
+      .append("> is specified in the format \"key1=v1 key2=v2\".")
+      .toString();
+  }
+
+  private Schema getSchema(Arguments arguments) throws IOException {
```
chtyim: `@Nullable`.

cdap-cli/src/main/java/co/cask/cdap/cli/command/view/CreateOrUpdateViewCommand.java
```diff
@@ -0,0 +1,123 @@
...
+    Id.View viewId = Id.View.from(cliConfig.getCurrentNamespace(), arguments.get(ArgumentName.VIEW.toString()));
+    Id.Stream streamId = Id.Stream.from(cliConfig.getCurrentNamespace(), arguments.get(ArgumentName.STREAM.toString()));
+    String formatName = arguments.get(ArgumentName.FORMAT.toString());
+    Schema schema = getSchema(arguments);
+
+    Map<String, String> settings = Collections.emptyMap();
+    if (arguments.hasArgument(ArgumentName.SETTINGS.toString())) {
+      settings = ArgumentParser.parseMap(arguments.get(ArgumentName.SETTINGS.toString()));
+    }
+    FormatSpecification formatSpecification = new FormatSpecification(formatName, schema, settings);
```
chtyim: So schema can be null? Does `FormatSpecification` allow that?

cdap-cli/src/main/java/co/cask/cdap/cli/command/view/CreateOrUpdateViewCommand.java
```diff
@@ -0,0 +1,123 @@
...
+      // if it's a json object, try to parse it as json
+      String schemaStr = arguments.get(ArgumentName.SCHEMA.toString());
+      schema = isJsonObject(schemaStr) ? Schema.parseJson(schemaStr) : Schema.parseSQL(schemaStr);
+    }
+    return schema;
+  }
+
+  private boolean isJsonObject(String str) {
+    try {
+      GSON.fromJson(str, JsonObject.class);
```
chtyim: So you are actually parsing it twice? Can you just do `Schema.parseJson` and I believe it will throw exception if it is not json?

cdap-cli/src/main/java/co/cask/cdap/cli/command/view/CreateOrUpdateViewCommand.java
jwang47: Appears to be so. Most of this class uses the same logic as https://github.com/caskdata/cdap/blob/develop/cdap-cli/src/main/java/co/cask/cdap/cli/command/SetStreamFormatCommand.java (didn't bother to move the code to a common place since we're going to remove it anyway)

chtyim: In that case, better annotate the `FormatSpecification` correctly.

cdap-cli/src/main/java/co/cask/cdap/cli/command/view/DescribeViewCommand.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Inject
+  public DescribeViewCommand(ViewClient client, CLIConfig cliConfig) {
+    super(cliConfig);
+    this.client = client;
+  }
+
+  @Override
+  public void perform(Arguments arguments, PrintStream output) throws Exception {
+   ViewDetail detail = client.get(
```
chtyim: Wrong alignment.

cdap-cli/src/main/java/co/cask/cdap/cli/completer/element/ViewIdCompleter.java
```diff
@@ -0,0 +1,50 @@
...
+  @Inject
+  public ViewIdCompleter(final ViewClient viewClient, final CLIConfig cliConfig) {
+    super(new Supplier<Collection<String>>() {
+      @Override
+      public Collection<String> get() {
+        try {
+          return viewClient.list(cliConfig.getCurrentNamespace());
+        } catch (IOException e) {
+          return Lists.newArrayList();
+        } catch (UnauthorizedException e) {
```
chtyim: Can be simplified to
```java
catch (IOException | UnauthorizedException e) {
  return new ArrayList<>();
}
```

cdap-data-fabric/src/main/java/co/cask/cdap/data/view/ViewStore.java
```diff
@@ -0,0 +1,63 @@
...
+
+import co.cask.cdap.common.NotFoundException;
+import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.ViewConfig;
+import co.cask.cdap.proto.ViewDetail;
+
+import java.util.List;
+
+/**
+ * Interface for storing Explore views.
```
chtyim: It's not explore views. Just views.

cdap-proto/src/main/java/co/cask/cdap/proto/ViewConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import co.cask.cdap.api.data.format.FormatSpecification;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Represents the configuration of a view, used when creating or updating a view.
+ */
+public class ViewConfig {
```
chtyim: Call it `ViewSpecification` to be consistent with other specifications.

chtyim: oh. It's in proto. Maybe `ViewMeta` then. We try to use "Config" to refer to configuration for an entity.

cdap-data-fabric/src/main/java/co/cask/cdap/data/view/ViewStore.java
```diff
@@ -0,0 +1,63 @@
...
+   *
+   * @param viewId the view
+   */
+  void delete(Id.View viewId) throws NotFoundException;
+
+  /**
+   * @param namespaceId the namespace
+   * @return list of view IDs in a namespace
+   */
+  List<String> list(Id.Namespace namespaceId);
```
chtyim: Better be `List<Id.View>`.

cdap-proto/src/main/java/co/cask/cdap/proto/ViewDetail.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a view in an HTTP response.
+ */
+public class ViewDetail {
+
+  private final String id;
```
chtyim: It seems this and ViewConfig are almost the same. Why is it so?

cdap-explore-client/src/main/java/co/cask/cdap/explore/client/ExploreFacade.java
```diff
@@ -53,6 +54,30 @@ public ExploreFacade(ExploreClient exploreClient, CConfiguration cConf) {
   }
 
   /**
+   * Creates a table for a view.
+   */
+  public void enableExploreView(Id.View view, ViewConfig config) throws ExploreException, SQLException {
```
chtyim: Why need this in explore? Should it be like create table?

chtyim: I mean, explore module doesn't need to be aware of View.

cdap-data-fabric/src/main/java/co/cask/cdap/data/view/InMemoryViewStore.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * In-memory implementation of {@link ViewStore}.
+ */
+public final class InMemoryViewStore implements ViewStore {
+
+  private final Map<Id.View, ViewConfig> views;
```
chtyim: These two can be combined into a guava `Table`?

cdap-proto/src/main/java/co/cask/cdap/proto/ViewDetail.java
jwang47: This is used in listing views, so that it includes the `id` of the view.

cdap-explore-client/src/main/java/co/cask/cdap/explore/client/ExploreFacade.java
jwang47: This follows the existing pattern for datasets and streams unfortunately.

cdap-data-fabric/src/main/java/co/cask/cdap/data/view/InMemoryViewStore.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Override
+  public boolean exists(Id.View viewId) {
+    return views.containsKey(viewId);
+  }
+
+  @Override
+  public void delete(Id.View viewId) throws NotFoundException {
+    if (!views.containsKey(viewId)) {
```
chtyim: no need to check for contains first. Usually map/map like structure in guava will return the value remove or `null` if nothing is removed (since we don't store `null` value in the map, check for `null` after remove should be good enough).

cdap-data-fabric/src/main/java/co/cask/cdap/data/view/InMemoryViewStore.java
```diff
@@ -0,0 +1,91 @@
...
+
+    views.remove(viewId);
+    viewsByNamespace.remove(viewId.getNamespace(), viewId);
+  }
+
+  @Override
+  public List<String> list(Id.Namespace namespaceId) {
+    ImmutableList.Builder<String> builder = ImmutableList.builder();
+    builder.addAll(Collections2.transform(viewsByNamespace.get(namespaceId), new Function<Id.View, String>() {
+      @Nullable
```
chtyim: Remove

jwang47:
Created https://issues.cask.co/browse/CDAP-3586, going to do that refactoring in a separate PR.

cdap-api-common/src/main/java/co/cask/cdap/api/data/format/FormatSpecification.java
```diff
@@ -37,16 +38,21 @@ private FormatSpecification() {
...
   }
 
-  public FormatSpecification(String name, Schema schema, Map<String, String> settings) {
+  public FormatSpecification(String name, @Nullable Schema schema, @Nullable Map<String, String> settings) {
     this.name = name;
     this.schema = schema;
     this.settings = settings == null ? Collections.<String, String>emptyMap() : settings;
   }
 
+  public FormatSpecification(String name, Schema schema) {
```
chtyim: So does this `Schema` also be `@Nullable`? Seems awkward that one constructor is nullable while the other is not.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/view/MDSStreamViewStore.java
```diff
@@ -0,0 +1,186 @@
...
+        public List<StreamViewEntry> apply(MetadataStoreDataset mds) throws Exception {
+          return Objects.firstNonNull(
+            mds.<StreamViewEntry>list(getKey(streamId), StreamViewEntry.class),
+            ImmutableList.<StreamViewEntry>of());
+        }
+      });
+
+    ImmutableList.Builder<Id.Stream.View> builder = ImmutableList.builder();
+    builder.addAll(Collections2.transform(entries, new Function<StreamViewEntry, Id.Stream.View>() {
+      @Nullable
```
chtyim: It can never be null, right?

cdap-proto/src/main/java/co/cask/cdap/proto/StreamViewSpecification.java
```diff
@@ -0,0 +1,57 @@
...
+package co.cask.cdap.proto;
+
+import co.cask.cdap.api.data.format.FormatSpecification;
+
+import java.util.Objects;
+
+/**
+ * Represents the configuration of a stream view, used when creating or updating a view.
+ */
+public class StreamViewSpecification {
```
chtyim: I think we can just call this `ViewSpecification`.

cdap-proto/src/main/java/co/cask/cdap/proto/StreamViewSpecification.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private final FormatSpecification format;
```
chtyim: What's the point of just wrapping around a `FormatSpecification`? Shouldn't there other information like view name and config in the spec? Also, should we consider flattening `FormatSpecification` into this class?

jwang47: Not sure we can both add view name to `ViewSpecification` and use `ViewSpecification` instead of `FormatSpecification`, since we still assume there's a "default" `FormatSpecification` per stream. The config is in `FormatSpecification`.

cdap-explore-client/src/main/java/co/cask/cdap/explore/client/MockExploreClient.java
```diff
@@ -48,8 +46,12 @@
 
   public MockExploreClient(Map<String, List<ColumnDesc>> statementsToMetadata,
                            Map<String, List<QueryResult>> statementsToResults) {
-    this.statementsToMetadata = Maps.newHashMap(statementsToMetadata);
-    this.statementsToResults = Maps.newHashMap(statementsToResults);
+    this.statementsToMetadata = statementsToMetadata == null ? null : Maps.newHashMap(statementsToMetadata);
```
chtyim: Better not to allow null in parameter and have empty map being passed.

chtyim:
Just one comment left. Rest LGTM

PR #3822, opened 2015-09-03
[CDAP-2864] add @Nonnull to Table's get methods

anew:
also change all places where we check whether result is null


chtyim:
LGTM.

anew:
waiting for build to suceed before merging: http://builds.cask.co/browse/CDAP-DUT2733

PR #3823, opened 2015-09-03
CDAP-3479- Integrate error dataset transformation - ETL Realtime

gsps1:
we want to user to be able to write to an error dataset from validator transform for invalid records. 
this PR contains implementation for ETL Realtime. 

JIRA: https://issues.cask.co/browse/CDAP-3479
Build: http://builds.cask.co/browse/CDAP-DUT2735-1

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch-app/src/main/java/co/cask/cdap/app/etl/batch/ETLMapReduce.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch-app/src/main/java/co/cask/cdap/app/etl/batch/ETLMapReduce.java
@@ -17,13 +17,13 @@
 package co.cask.cdap.app.etl.batch;
 
 import co.cask.cdap.api.ProgramLifecycle;
+import co.cask.cdap.api.dataset.DatasetProperties;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.mapreduce.AbstractMapReduce;
 import co.cask.cdap.api.mapreduce.MapReduceContext;
 import co.cask.cdap.api.metrics.Metrics;
 import co.cask.cdap.app.etl.batch.config.ETLBatchConfig;
 import co.cask.cdap.template.etl.api.Transform;
-import co.cask.cdap.template.etl.api.Transformation;
 import co.cask.cdap.template.etl.api.batch.BatchSink;
 import co.cask.cdap.template.etl.api.batch.BatchSinkContext;
 import co.cask.cdap.template.etl.api.batch.BatchSource;
@@ -37,7 +37,9 @@ import co.cask.cdap.template.etl.common.Pipeline;
 import co.cask.cdap.template.etl.common.PipelineRegisterer;
 import co.cask.cdap.template.etl.common.PluginID;
 import co.cask.cdap.template.etl.common.StageMetrics;
+import co.cask.cdap.template.etl.common.TransformDetail;
 import co.cask.cdap.template.etl.common.TransformExecutor;
+import co.cask.cdap.template.etl.common.TransformInfo;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
 import com.google.common.reflect.TypeToken;
@@ -50,6 +52,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.lang.reflect.Type;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -78,7 +81,8 @@ public class ETLMapReduce extends AbstractMapReduce {
     setDescription("MapReduce driver for Batch ETL Adapters");
 
     PipelineRegisterer pipelineRegisterer = new PipelineRegisterer(getConfigurer());
-    Pipeline pipelineIds = pipelineRegisterer.registerPlugins(config);
+    //TODO : CDAP-3480 - passing null now, will implement error dataset using Fileset for ETLMapReduce
+    Pipeline pipelineIds = pipelineRegisterer.registerPlugins(config, null, DatasetProperties.EMPTY);
 
     if (config.getResources() != null) {
       setMapperResources(config.getResources());
@@ -147,9 +151,7 @@ public class ETLMapReduce extends AbstractMapReduce {
   public static class ETLMapper extends Mapper implements ProgramLifecycle<MapReduceContext> {
     private static final Logger LOG = LoggerFactory.getLogger(ETLMapper.class);
     private static final Gson GSON = new Gson();
-    private static final Type STRING_LIST_TYPE = new TypeToken<List<String>>() { }.getType();
-
-    private List<Transform> transforms;
+    private static final Type TRANSFORMDETAILS_LIST_TYPE = new TypeToken<List<TransformInfo>>() { }.getType();
 
     private TransformExecutor<KeyValue, KeyValue> transformExecutor;
     private Metrics mapperMetrics;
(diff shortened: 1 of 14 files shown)


cdap-app-templates/cdap-etl/cdap-etl-realtime-app/pom.xml
```diff
@@ -52,6 +52,11 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-etl-common</artifactId>
```
gsps1: we want to be able to use ``StructuredRecordStringConverter`` in ``ETLWorker``, so adding this dependency

cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactVersion.java
```diff
@@ -112,7 +112,7 @@ public String getSuffix() {
   }
 
   public boolean isSnapshot() {
-    return suffix != null && "snapshot".equals(suffix.toLowerCase());
+    return suffix != null && suffix.toLowerCase().contains("snapshot");
```
gsps1: @albertshau let me know if we want to keep this.

albertshau: i think we should, though it should probably be startsWith

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/DefaultEmitter.java
```diff
@@ -56,6 +58,11 @@ public void emitError(InvalidEntry<T> value) {
     return entryList.iterator();
   }
 
+  public Iterator<InvalidEntry<T>> getErrorIterator() {
+    // we have to return iterator of a copy of list, as we need reset the errorList between stages.
+    return new ArrayList<>(errorList).iterator();
```
albertshau: does this mean we're not immediately consuming them?  the normal iterator() does not return a copy.

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/ETLStage.java
```diff
@@ -49,4 +55,8 @@ public String toString() {
       .add("properties", properties)
       .toString();
   }
+
```
albertshau: Nullable

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/ETLStage.java
```diff
@@ -27,10 +27,16 @@
 public final class ETLStage {
   private final String name;
   private final Map<String, String> properties;
+  private final String datasetName;
```
albertshau: this is a bit misleading. Perhaps name it errorDatasetName?

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/TransformDetails.java
```diff
@@ -0,0 +1,40 @@
...
+
+package co.cask.cdap.template.etl.common;
+
+import javax.annotation.Nullable;
+
+/**
+ * Class to encapsulate id of the transform and the error dataset used in that transform stage.
+ */
+public class TransformDetails {
+  String transformId;
```
albertshau: private final

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/TransformDetails.java
```diff
@@ -0,0 +1,40 @@
...
+  public TransformDetails(String transformId, @Nullable String errorDatasetName) {
+    this.transformId = transformId;
+    this.errorDatasetName = errorDatasetName;
+  }
+
+  public String getTransformId() {
+    return transformId;
+  }
+
+  public String getErrorDatasetName() {
```
albertshau: annotate as Nullable

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/PipelineRegisterer.java
```diff
@@ -94,9 +106,18 @@ public Pipeline registerPlugins(ETLConfig config) {
         throw new IllegalArgumentException(String.format("No Plugin of type '%s' named '%s' was found",
                                                          Constants.Transform.PLUGINTYPE, transformConfig.getName()));
       }
+      // if the transformation is configured to write filtered records to error dataset, we create that dataset.
+      if (transformConfig.getDatasetName() != null) {
+        // TODO : can remove this after implementing CDAP-3480
+        if (errorDatasetType != null) {
+          configurer.createDataset(transformConfig.getDatasetName(), errorDatasetType, DatasetProperties.builder()
```
albertshau: perhaps we should also pass in the DatasetProperties?

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/DefaultEmitter.java
gsps1: the normal iterator just updates a single list between each stage of the transform and after all the transform stages are complete, it just has to return the iterator of the list. whereas for our use case, we are interested in error records from each stage, so we have to do a copy. ``runOneIteration`` runs transform on all stages before returning.

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/TransformDetails.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package co.cask.cdap.template.etl.common;
+
+import javax.annotation.Nullable;
+
+/**
+ * Class to encapsulate id of the transform and the error dataset used in that transform stage.
+ */
+public class TransformDetails {
```
albertshau: nit: lets rename to TransformDetail

albertshau: was confused for a little bit because this looks so similar to TransformationDetails.  Lets rename this to TransformInfo or something like that, and TransformationDetails to TransformDetail

cdap-app-templates/cdap-etl/cdap-etl-batch-app/src/main/java/co/cask/cdap/app/etl/batch/ETLMapReduce.java
```diff
@@ -161,43 +162,40 @@ public void initialize(MapReduceContext context) throws Exception {
...
 
-      transformExecutor = new TransformExecutor<>(pipeline, stageMetrics);
+      transformExecutor = new TransformExecutor<>(pipeline);
     }
 
-    private void addTransforms(List<Transformation> pipeline,
-                               List<StageMetrics> stageMetrics,
-                               List<String> transformIds,
+    private void addTransforms(List<TransformationDetails> pipeline,
+                               List<TransformDetails> transformIds,
```
albertshau: rename variable since its not longer ids

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/TransformExecutor.java
```diff
@@ -36,35 +36,45 @@
...
-    Transformation transform = transforms.get(0);
+    TransformationDetails transformationDetails = transforms.get(0);
     DefaultEmitter currentEmitter = emitters.get(0);
     currentEmitter.reset();
+    Transformation transform = transformationDetails.getTransformation();
     transform.transform(input, currentEmitter);
 
-    DefaultEmitter previousEmitter = currentEmitter;
+    // if there are errors in the iterator, we add it to the errorIterators list
+    if (currentEmitter.getErrorIterator().hasNext()) {
```
albertshau: rather than tracking all these iterators, I think it would be cleaner to have the TransformExecutor take in an InvalidEntryWriter.  The InvalidEntryWriter writes any invalid entries it sees in between transform stages.  Then you don't need to introduce the TransformResponse object and the logic should be simpler.

cdap-app-templates/cdap-etl/cdap-etl-realtime-app/src/main/java/co/cask/cdap/app/etl/realtime/ETLWorker.java
```diff
@@ -227,8 +241,31 @@ public void run(DatasetContext context) throws Exception {
...
+          }
+          final List<TransformResponse.TransformError> errorRecords = transformResponse.getErrorRecords();
+
+          for (final TransformResponse.TransformError response : errorRecords) {
+            final String datasetName = tranformIdToDatasetName.get(response.getTransformId());
+            getContext().execute(new TxRunnable() {
+              @Override
+              public void run(DatasetContext context) throws Exception {
+                Table errorTable = context.getDataset(datasetName);
+                String rowkeyPrefix = System.currentTimeMillis() + SEPARATOR + appName + SEPARATOR +
```
albertshau: row key can be composed a little better to save bytes.  Timestamp is a long, followed by a string for app name, separator, transformid, then UUID is 128 bits.  Bytes has a method to interact with UUID.

cdap-app-templates/cdap-etl/cdap-etl-realtime-app/src/main/java/co/cask/cdap/app/etl/realtime/ETLWorker.java
```diff
@@ -269,6 +306,16 @@ public void run(DatasetContext context) throws Exception {
     }
   }
 
+  private Put constructErrorPut(byte[] rowKey, InvalidEntry entry) throws IOException {
+    Put errorPut = new Put(rowKey);
+    errorPut.add(Constants.ErrorDataset.COLUMN_ERRCODE, entry.getErrorCode());
+    errorPut.add(Constants.ErrorDataset.COLUMN_ERRMSG, entry.getErrorMsg());
+    StructuredRecord record =  (StructuredRecord) entry.getInvalidRecord();
```
albertshau: nit: remove extract space before (StructuredRecord)

albertshau: technically, there is no guarantee that the InvalidRecord is a StructuredRecord, even though this is true for all our transforms. Lets check that it really is a StructuredRecord, then just toString() it if it's not.

albertshau:
took a pass through.  Mostly minor comments except for one about changing where we write to the error datasets

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/TransformExecutor.java
albertshau: ok ignore that previous comment since we want to write everything at the same time in one transaction.  We can still make this better and avoid all the copying that is done by passing in a List<TransformError> to DefaultEmitter and just have the emitter add to the list.

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/PipelineRegisterer.java
gsps1: isnt the schema same between both of them? as we are writing InvalidEntry in both cases?

cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactVersion.java
albertshau: actually lets remove this and do it in a separate PR.  It will exacerbate a bug we found today, such that master restarts will break plugins.

gsps1:
@albertshau have addressed your comments, please take a look again. Thanks

albertshau:
a few comments

albertshau:
lgtm

albertshau:
squashed commits and re-running build.

PR #3824, opened 2015-09-04
CDAP-3582 allow filtering apps by multiple artifact names

albertshau:
Extending the get apps rest api to allow specifying a comma
separated list of artifact names. The endpoint will return all
apps that use an artifact whose name is in that set of names.

Also performing some refactoring along the way, moving logic from
the http handler into the service. This allows us to remove the
usage of Store in the http handler.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
@@ -221,21 +221,6 @@ public interface Store {
   Collection<ApplicationSpecification> getAllApplications(Id.Namespace id);
 
   /**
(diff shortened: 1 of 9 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2739

cdap-client/src/main/java/co/cask/cdap/client/ArtifactClient.java
```diff
@@ -216,7 +216,7 @@ public ArtifactInfo getArtifactInfo(Id.Artifact artifactId, boolean isSystem)
     return getApplicationClasses(namespace, true);
   }
 
-  /**
+  /**i
```
gokulavasan: i

gokulavasan:
LGTM :+1:

PR #3825, opened 2015-09-04
CDAP-2869 throw invalid argument exception if the given configuration string

gokulavasan:
has json syntax error

Build : http://builds.cask.co/browse/CDAP-DUT2740

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
@@ -42,6 +42,7 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
 import org.apache.twill.filesystem.Location;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
```diff
@@ -162,20 +163,22 @@ private String getSpecJson(Application app, final String bundleVersion, final St
...
-    TypeToken typeToken = TypeToken.of(app.getClass());
-    TypeToken<?> configToken = typeToken.resolveType(Application.class.getTypeParameters()[0]);
-    if (Strings.isNullOrEmpty(configString)) {
-      appConfig = (Config) configToken.getRawType().newInstance();
-    } else {
-      //TODO: CDAP-2869 Handle JsonSyntax Exception better and throw a more meaningful error. This will be done when we
-      //use PluginInstantiator methods instead of GSON deserialization
-      appConfig = GSON.fromJson(configString, configToken.getType());
-    }
+      Config appConfig;
```
albertshau: its not wrong, but wondering why these parts were moved into the try block

albertshau:
lgtm

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
gokulavasan: Oh wondering how it was working before. Because we were creating the PluginInstantiator in ``try`` but we are constructing only the Configurer. But the plugin registration is being done only when we call the application's configure method. But PluginInstantiator resource would have been closed by then. So moving that logic inside the try block.

gokulavasan: So it looks like we were just leaving files which are not deleted since we were closing the PluginInstantiator earlier. So functionality was not affected but this change will cleanup the dir.

gokulavasan:

PR #3827, opened 2015-09-04
CDAP-3509 Error checking to make sure duplicate streams, datasets

gokulavasan:
plugins are not added

Build : http://builds.cask.co/browse/CDAP-DUT2736

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
+++ b/cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
@@ -23,9 +23,12 @@ import co.cask.cdap.api.dataset.Dataset;
 import co.cask.cdap.api.dataset.DatasetProperties;
 import co.cask.cdap.api.dataset.module.DatasetModule;
 import co.cask.cdap.internal.dataset.DatasetCreationSpec;
+import com.google.common.base.Preconditions;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Implementation of {@link DatasetConfigurer} for adding datasets and streams in Configurers.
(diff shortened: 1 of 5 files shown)


gokulavasan:
On hold. Need to add more test cases

gokulavasan:
@albertshau Please review

cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
```diff
@@ -50,21 +53,43 @@
   }
 
   public void addStreams(Map<String, StreamSpecification> streams) {
+    Set<String> duplicateStreams = findDuplicates(streams.keySet(), this.streams.keySet());
```
sreevatsanraman: Why is this needed? Will we not be able to deploy two `ETLBatch` pipelines reading from same stream?

cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
```diff
@@ -50,21 +53,43 @@
...
+    checkArgument(duplicateStreams.isEmpty(),
+                  "Streams %s have been added already. Remove the duplicates.", duplicateStreams);
     this.streams.putAll(streams);
   }
 
   public void addDatasetSpecs(Map<String, DatasetCreationSpec> datasetSpecs) {
+    Set<String> duplicateDatasetSpecs = findDuplicates(datasetSpecs.keySet(), this.datasetSpecs.keySet());
+    checkArgument(duplicateDatasetSpecs.isEmpty(),
```
sreevatsanraman: This makes sense only if we are writing to same dataset from two different pipelines.

albertshau:
We need to be able to give better error messages. This approach will tell you there are duplicates, but will not indicate where they come from or how many.

One thing we could do is instead of having `Map<String, XSpecification>` in DefaultAppConfigurer, we could have `Map<String, XConfigurer>`.  Then when we build the app spec in createSpecification, we have the datasets created in each program, and can throw an error that tells people datasetA was added in WorkerY and MapReduceZ with different properties.

cdap-unit-test/src/test/java/co/cask/cdap/test/app/AppWithDuplicateStreams.java
```diff
@@ -0,0 +1,50 @@
...
+
+import co.cask.cdap.api.app.AbstractApplication;
+import co.cask.cdap.api.dataset.lib.KeyValueTable;
+import co.cask.cdap.api.worker.AbstractWorker;
+
+/**
+ * App with streams, datasets added multiple times that should result in an error.
+ */
+public class AppWithDuplicateStreams extends AbstractApplication {
+
```
gsps1: this app doesn't seem to have any duplicate stream

gokulavasan: forgot to push a commit. should be able to find duplicate streams now.

gokulavasan:
@albertshau That is right. If we follow the approach you had suggested, we still may not be able to tell how many duplicates are there (at least in the simpler approach). Also, if streams/datasets added twice in the same program we still won't be to give a better error message?

sreevatsanraman:
Closing this Re-open against release/3.1

PR #3828, opened 2015-09-04
CDAP-3560 changing isSystem to scope in artifact client

albertshau:
Missed a couple places where isSystem was still being used.
Also modifying the test to make sure scope is used.


gokulavasan:
LGTM :+1:

PR #3832, opened 2015-09-04
Support for adding OutputFormatProvider as output of MR

anwar6953:
This allows adding an OutputFormatProvider as an output onto the MapReduceContext.
Previously, only Datasets were allowed.

http://builds.cask.co/browse/CDAP-DUT2743-4

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
@@ -132,7 +132,7 @@ public interface MapReduceContext extends RuntimeContext, DatasetContext, Servic
   void addOutput(String datasetName);
 
   /**
-   * Overrides the output configuration of this MapReduce job to also allow writing to the specified dataset.
+   * Updates the output configuration of this MapReduce job to also allow writing to the specified dataset.
    * Currently, the dataset specified in must be an {@link OutputFormatProvider}.
    * You may want to use this method instead of {@link #addOutput(String)} if your output dataset uses runtime
    * arguments set in your own program logic.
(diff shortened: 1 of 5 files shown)


cdap-api/src/main/java/co/cask/cdap/api/mapreduce/MapReduceContext.java
```diff
@@ -144,6 +144,16 @@
   void addOutput(String datasetName, Map<String, String> arguments);
 
   /**
+   * Updates the output configuration of this MapReduce job to also allow writing, based upon the given
```
albertshau: wording a little awkward, what about "allow writing using the given OutputFormatProvider"

cdap-unit-test/src/test/java/co/cask/cdap/partitioned/AppWithMapReduceConsumingPartitions.java
```diff
@@ -134,6 +157,26 @@ public void beforeSubmit(MapReduceContext context) throws Exception {
...
 
+      Map<String, String> outputArgs = new HashMap<>();
+      PartitionKey partitionKey = PartitionKey.builder().addLongField("time", context.getLogicalStartTime()).build();
+      PartitionedFileSetArguments.setOutputPartitionKey(outputArgs, partitionKey);
+
+      // We know that PartitionedFileSet is an OutputFormatProvider, so we set our own instance of an
+      // OutputFormatProvider (that delegates to the PFS's implementation) as an output to the MapReduce job
+      // to test MapReduceContext#addOutput(String, OutputFormatProvider)
+      final PartitionedFileSet outputLines = context.getDataset("outputLines", outputArgs);
+      context.addOutput("outputLines", new OutputFormatProvider() {
```
albertshau: PartitionedFileSet is already an OutputFormatProvider, could probably just cast it?

albertshau:

cdap-unit-test/src/test/java/co/cask/cdap/partitioned/AppWithMapReduceConsumingPartitions.java
anwar6953: I wasn't sure if the call (without casting) was ambiguous, since the object is both a Dataset and OutputFormatProvider.

If I put the cast, intellij says that its redundant (and so someone might easily remove it in the future)?

albertshau:
:+1:

PR #3835, opened 2015-09-04
CDAP-3390: Made ExploreFacade more flexible in {enable,disable}ExploreStream

jwang47:
http://builds.cask.co/browse/CDAP-DUT2749


cdap-common/src/main/java/co/cask/cdap/internal/explore/ExploreTableNaming.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public String getTableName(Id.Stream streamId) {
+    return cleanTableName(String.format("stream_%s", streamId.getId()));
+  }
+
+  public String getTableName(Id.DatasetInstance datasetID) {
+    return cleanTableName(String.format("dataset_%s", datasetID.getId()));
+  }
+
+  public String cleanTableName(String name) {
```
chtyim: Why public?

cdap-common/src/main/java/co/cask/cdap/internal/explore/ExploreTableNaming.java
```diff
@@ -0,0 +1,40 @@
...
+
+import co.cask.cdap.proto.Id;
+
+/**
+ * Specifies how to name tables for Explore.
+ */
+public final class ExploreTableNaming {
+
+  public String getTableName(Id.Stream streamId) {
+    return cleanTableName(String.format("stream_%s", streamId.getId()));
```
chtyim: Shouldn't it be `return "stream_" + cleanTableName(streadId.getId());`? Meaning only clean the id?

cdap-explore-client/src/main/java/co/cask/cdap/explore/client/AbstractExploreClient.java
```diff
@@ -118,6 +119,47 @@ public QueryHandle getHandle() throws ExploreException, SQLException {
   }
 
   @Override
+  public ListenableFuture<Void> enableExploreStream(final Id.Stream stream, final String tableName) {
```
chtyim: Do we need this? Can we force caller always provide the FormatSpecification?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/pipeline/StreamCreator.java
```diff
@@ -52,7 +56,8 @@ public void createStreams(Set<String> streamNames) throws Exception {
       if (!streamAdmin.exists(streamId)) {
         streamAdmin.create(streamId);
         if (enableExplore) {
-          exploreFacade.enableExploreStream(streamId);
+          StreamConfig config = streamAdmin.getConfig(streamId);
+          exploreFacade.enableExploreStream(streamId, NAMING.getTableName(streamId), config.getFormat());
```
chtyim: No need to call this since it's already done by the `StreamAdmin.create()`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -486,15 +488,15 @@ private void mutateStates(long groupId, int instances, Set<StreamConsumerState>
     }
   }
 
-  private void alterExploreStream(Id.Stream stream, boolean enable) {
+  private void alterExploreStream(Id.Stream stream, boolean enable, FormatSpecification format) {
```
chtyim: `@Nullable` for the format since you pass in null when enable == false.

chtyim:
Just two comments, reset LGTM.

PR #3836, opened 2015-09-04
(CDAP-3014) Prevent SparkSubmit calling System.exit on failure

chtyim:


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
```diff
@@ -174,6 +176,8 @@ public TwillRunnableSpecification configure() {
 
   @Override
   public void initialize(TwillContext context) {
+    System.setSecurityManager(new RunnableSecurityManager());
```
bdmogal: so we don't need to cache and reset it? Or are we going to use ``RunnableSecurityManager`` for the entire duration of the Runnable starting from ``initialize``, but it only applies to spark programs?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
```diff
@@ -473,5 +477,48 @@ public Program create(String path) throws IOException {
...
+  /**
+   * A {@link SecurityManager} used by the runnable container. It specifically disabling
+   * Spark classes to call {@link System#exit(int)}, {@link Runtime#halt(int)}
+   * and {@link System#setSecurityManager(SecurityManager)}.
+   * This is to workaround modification done in HDP Spark (CDAP-3014).
+   */
+  private static final class RunnableSecurityManager extends SecurityManager {
+    @Override
+    public void checkPermission(Permission perm) {
+      if ("setSecurityManager".equals(perm.getName()) && isFromSpark()) {
```
bdmogal: if its not from Spark, should we delegate to ``super.checkPermission``? same for the methods below? or not needed?

chtyim:
Test in HDP 2.2 cluster and CDH 5.3 cluster

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
chtyim: Tried that. It will throw.

bdmogal: oh ok. so it needs to be suppressed completely. so do we become future proof if other systems also decide to call ``System.exit()``? :)

poornachandra:
Since the interception through SecurityManger is only required for Spark, can this change be moved to SparkTwillRunnable?

chtyim:
Better not.
1. Workflow also need it.
2. If set in some Spark specific class (say SparkRuntimeService), it will get set multiple times when a fork is executed
3. The SecurityManager can be evolved to be generic (e.g. also disable user program to call System.exit).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
chtyim: No resetting. The SecurityManager is used through the execution of the container.

bdmogal:
:+1:

bdmogal:
Second commit :+1:

chtyim:
Squashed the commits.

PR #3837, opened 2015-09-04
Change the word "file" to "product" in the license description.

awholegunch:
This is a one-word change to the license file included in the documentation. This license is the standard text used in the Apache License, which was intended to be placed at the top of a source code *file*; and so it uses the word *file*.

But our usage here is slightly different; we are referring to the CDAP *product*, and hence the one-word change.

### DIFF
--- a/cdap-docs/reference-manual/source/licenses/index.rst
+++ b/cdap-docs/reference-manual/source/licenses/index.rst
@@ -161,7 +161,7 @@ The Cask Data Application Platform (CDAP) product is copyright and licensed as f
    Copyright © |copyright|
 
    Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


sreevatsanraman:
lgtm

anew:
I am not sure whether it is ok to change the text of the Apache License, I believe it must remain unchanged if you call it "Apache License" (which we do). Please double check.

awholegunch:
I think this is fine, because if you look at the entire license page (http://www.apache.org/licenses/LICENSE-2.0.txt), at the end, it describes how this is not the license, but a "boilerplate' notice that you add to your work. You are suppose to edit the notice, so we are actually changing the notice, not the license. I perhaps was not specific enough in that the change was not to the license itself, but to the notice of the license. We tend to lump the two together.

anew:
OK, thanks for the explanation, JJ.

PR #3838, opened 2015-09-05
[CDAP-3527] First drop for adding Busineess Metadata Dataset

henry-cask:
This PR is created to help share internally about the new datasets for Business Metadata.

Added new system dataset for Business Metadata called BusinessMetadataDataset that internally backed by IndexedTable to support inverted index search for value and key:value columns.


cdap-proto/src/main/java/co/cask/cdap/proto/BusinessMetadataRecord.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * Represent Busineess Metadata for CDAP.
+ */
+public class BusinessMetadataRecord {
+  private final String targetType;
+  private final Id.NamespacedId  targetId;
```
poornachandra: The type of `targetId` is supposed to be the parent class for all target types, right? In that case let's just use `Id`, rather than `NamespacedId`. Using `NamespacedId` here is a bit misleading while reading the code. What do you say?

cdap-proto/src/main/java/co/cask/cdap/proto/BusinessMetadataRecord.java
```diff
@@ -0,0 +1,97 @@
...
+  private final String key;
+  private final String value;
+  private final String schema;
+
+  public BusinessMetadataRecord(String targetType, Id.NamespacedId targetId, String key, String value) {
+    this(targetType, targetId, key, value, null);
+  }
+
+  public BusinessMetadataRecord(String targetType, Id.NamespacedId  targetId, String key, String value,
```
poornachandra: It might be better to have `targetType` determined in this class based on `targetId`, rather than getting it as a parameter so that there is only one place to determine `targetType`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -0,0 +1,151 @@
...
+    } catch (Exception e) {
+      throw Throwables.propagate(e);
+    }
+  }
+
+  // Helper method to generate key.
+  private MDSKey getInstanceKey(String targetType, Id.NamespacedId targetId, String key) {
+    MDSKey.Builder builder = new MDSKey.Builder();
+    builder.add(targetType);
+    builder.add(targetId.toString());
```
poornachandra: I don't think `toString` methods of `Id` are reliable, we should probably do our own serialization/de-serialization here.

poornachandra:
@henry-cask A few minor comments, please review them.

cdap-proto/src/main/java/co/cask/cdap/proto/BusinessMetadataRecord.java
bdmogal: In fact, according to https://wiki.cask.co/display/CE/Metadata+and+Data+Discovery#MetadataandDataDiscovery-RESTAPIs, we're returning/consuming a simple map of string to strings from the REST APIs. Could we just do away with this class? Especially since ``targetId`` is part of the key as well?

Although I do understand that we may want to keep this class so we have the schema in each record as well.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDatasetTest.java
```diff
@@ -0,0 +1,84 @@
...
+  public static DatasetFrameworkTestUtil dsFrameworkUtil = new DatasetFrameworkTestUtil();
+
+  private static final Id.DatasetInstance datasetInstance =
+    Id.DatasetInstance.from(DatasetFrameworkTestUtil.NAMESPACE_ID, "meta");
+
+  BusinessMetadataDataset dataset;
+
+  @Before
+  public void before() throws Exception {
+    // TODO Add code
```
bdmogal: nit: remove these comments before merging.

cdap-proto/src/main/java/co/cask/cdap/proto/BusinessMetadataRecord.java
bdmogal: @poornachandra if we use just ``Id``, a lot more classes that do not apply can be passed here. E.g. ``Id.QueryHandler``, ``Id.SystemService``, ``Id.Namespace``. I think we should restrict it to ``Id.NamespacedId``, so only apps, programs, datasets and streams can be passed here.

cdap-proto/src/main/java/co/cask/cdap/proto/BusinessMetadataRecord.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+package co.cask.cdap.proto;
+
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * Represent Busineess Metadata for CDAP.
+ */
+public class BusinessMetadataRecord {
```
bdmogal: Since none of the REST APIs return/consume this object, should we move this out of ``cdap-proto``, into ``co.cask.cdap.data2.metadata.dataset``?

henry-cask: Sounds good moving it now

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDatasetTest.java
henry-cask: Cleaning it up ...

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: Could we add as another PR for this or it is urgent to do it now?

henry-cask:
@bdmogal , @poornachandra could you take a look again?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataRecord.java
```diff
@@ -0,0 +1,99 @@
...
+ * Represent Busineess Metadata for CDAP.
+ */
+public class BusinessMetadataRecord {
+  private final String targetType;
+  private final Id.NamespacedId  targetId;
+  private final String key;
+  private final String value;
+  private final String schema;
+
+  public BusinessMetadataRecord(String targetType, Id.NamespacedId targetId, String key, String value) {
```
bdmogal: I think you missed the comment about ``targetType``. It can be derived from ``targetId``. We shouldn't have to pass it from outside

henry-cask: Ah ok, will remove it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: File ticket to track this: https://issues.cask.co/browse/CDAP-3607

henry-cask:
Build: 
http://builds.cask.co/browse/CDAP-DUT2721-8

bdmogal:
LGTM. Please squash commits after build, and merge.

henry-cask:
Will do

henry-cask:
Build passes, merging ...

henry-cask:
Close this one prefer to https://github.com/caskdata/cdap/pull/3847

PR #3840, opened 2015-09-05
CDAP-3470 Return 503(Service unavailable) status code when the dataset service is not discoverable.

sagarkapare:
Squashed commits. Original PR #3819 already reviewed and build passed http://builds.cask.co/browse/CDAP-DUT2728-4


gsps1:
LGTM

PR #3841, opened 2015-09-05
CDAP-2556: Add support for HBase1.1 in CDAP

rohitsinha54:


cdap-hbase-compat-1.1/src/main/java/co/cask/cdap/data2/increment/hbase11/IncrementSummingScanner.java
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  @Override
+  public boolean next(List<Cell> cells, ScannerContext scannerContext) throws IOException {
+    return nextInternal(cells, scannerContext);
+  }
+
+  private boolean nextInternal(List<Cell> cells, ScannerContext scannerContext) throws IOException {
+    if (LOG.isTraceEnabled()) {
+      LOG.trace("nextInternal called with limit=" + scannerContext);
```
poornachandra: Since the message says "with limit", no need to print the whole `scannerContext`, right?

cdap-hbase-compat-1.1/src/main/java/co/cask/cdap/data2/increment/hbase11/IncrementSummingScanner.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+    Cell previousIncrement = null;
+    long runningSum = 0;
+    int addedCnt = 0;
+    baseScanner.startNext();
+    Cell cell;
+    while ((cell = baseScanner.peekNextCell(getBatch())) != null && (getBatch() <= 0 || addedCnt < getBatch())) {
```
poornachandra: We need to use limit from `scannerContext` everwhere rather than `getBatch` when `scannerContext` is available. 

Actually, I think the newly introduced method `getBatch` is not required at all. Any reason for adding it?

cdap-hbase-compat-1.1/src/main/java/co/cask/cdap/data2/util/hbase/HBase11TableUtil.java
```diff
@@ -0,0 +1,356 @@
...
+        return BloomType.NONE;
+      default:
+        throw new IllegalArgumentException("Unsupported bloom filter type: " + type);
+    }
+  }
+
+  @Override
+  public Class<? extends Coprocessor> getTransactionDataJanitorClassForVersion() {
+    CodeSource codeSource = HBase11TableUtil.class.getProtectionDomain().getCodeSource();
+    LOG.error("HBase11TableUtil is getting loaded from {}", codeSource == null ? "<null>" : codeSource.getLocation());
```
poornachandra: All these log messages need to be removed.

cdap-hbase-compat-1.1/src/test/java/co/cask/cdap/data2/increment/hbase11/IncrementHandlerTest.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        return hasMore;
+      } finally {
+        rs.close();
+      }
+    }
+
+    @Override
+    public boolean flush() throws IOException {
+      HRegion.FlushResult result = region.flushcache(true, false);
```
poornachandra: It would be a good idea to add a comment here saying why `flushCache` is `(true, false)`.

poornachandra:
@rohitsinha54 Few comments, please address them.

cdap-hbase-compat-1.1/src/main/java/co/cask/cdap/data2/increment/hbase11/IncrementSummingScanner.java
rohitsinha54: As discussed in person. The only thing which ScannerContext exposes is toString which will print the limit with progress. So keeping this.

cdap-hbase-compat-1.1/src/main/java/co/cask/cdap/data2/increment/hbase11/IncrementSummingScanner.java
```diff
@@ -0,0 +1,370 @@
...
+  private boolean nextInternal(List<Cell> cells, ScannerContext scannerContext) throws IOException {
+    if (LOG.isTraceEnabled()) {
+      LOG.trace("nextInternal called with limit=" + scannerContext);
+    }
+    Cell previousIncrement = null;
+    long runningSum = 0;
+    int addedCnt = 0;
+    baseScanner.startNext();
+    Cell cell;
+    while ((cell = baseScanner.peekNextCell(scannerContext)) != null && (getBatch() <= 0 || addedCnt < getBatch())) {
```
poornachandra: All get batch size requests should come from scanner context. In this case use internal scanner to get the batch size.

rohitsinha54:
This PR needs some changes to handle the API change which affect IncrementSummingScanner according to the review comments. I will fix them today and re-open this PR.

rohitsinha54:
Addressed the comments and also the issue cause by newly introduced ScannerContext.

The scan in HBase has change in 1.1 : https://blogs.apache.org/hbase/entry/scan_improvements_in_hbase_1

It introduces a ScannerContext which takes three kind of limits: batchLimit, sizeLimit and timeLimits. There is no way to get these limits from the ScannerContext as the getter methods are package private.

Now in our IncrementSummingScanner we implement RegionScanner and delegate the next calls to the RegionScanner with limits (now ScannerContext). We cannot depend on the RegionScanner limit tracking for batch size (limit on number of columns) as it does not differentiate between an increment and non-increment cells.
We need to do the limit tracking for cells to identify different columns and progress towards limit in IncrementSummingScanner which differentiate between a increment and non-increment cell and to do this we need to know the batch limit given in the ScannerContext. 

The 	"Add ScannerContextReader for HBase1.1 support" commit introduces a ScannerContextReader class in the same package as ScannerContext in HBase and delegates the call to get the limits to ScannerContext. This allow us to get the limits and use in it IncrementSummingScanner.

Also added a test which explicitly test for limit and the columns retrieved.

rohitsinha54:
New build running here: http://builds.cask.co/browse/CDAP-DUT2754-5

cdap-hbase-compat-1.1/src/main/java/co/cask/cdap/data2/increment/hbase11/IncrementSummingScanner.java
```diff
@@ -0,0 +1,379 @@
...
+    baseScanner.startNext();
+    Cell cell;
+    // We need to to access the batch limit (limit on number of columns) in the ScannerContext.
+    // The ScannerContext does not exposes its method to get the limits so we have a Reader under the same package
+    // which can supply us these limits given a ScannerContext. Also, we cannot depend on the limit tracking of the
+    // internal scanner as it does not differentiate between an increment cell and a non-increment cell.
+    // So, we read all the cells from the internal scanner in batches of limit and do the tracking for columns here
+    // and stop when we have reached the limit.
+    ScannerContextReader scannerContextReader = new ScannerContextReader(scannerContext);
+    int limit = scannerContextReader.getBatchLimit();
```
poornachandra: It would be a good idea to move the above lines into a private method called `getBatchLimit()` or something. Then the six lines of comments can also move to the private method.

chtyim:
LGTM

rohitsinha54:
Waiting for a final build here: http://builds.cask.co/browse/CDAP-DUT2754-6
Will merge as soon as build passes.

PR #3845, opened 2015-09-08
CDAP-3593 move sinks to new addOutput() apis in cdap mapreduce

albertshau:
This changes the sink contexts to use addOutput() instead of
setOutput() in preparation of allowing multiple output sinks.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2764

anwar6953:
LGTM

PR #3847, opened 2015-09-08
CDAP-3527 Add first drop of Business Metadata dataset.

henry-cask:
This PR is created to help share internally about the new datasets for Business Metadata.

Added new system dataset for Business Metadata called BusinessMetadataDataset that internally backed by IndexedTable to support inverted index search for value and key:value columns.


bdmogal:
LGTM

PR #3848, opened 2015-09-08
Feature/cdap 2757 dynamic partitioning of mr output

anwar6953:
Ability to dynamically write to multiple partitions of a PartitionedFileSet dataset, as the output of a MapReduce job.

This is implemented by having a DynamicParitioningOutputFormat which writes records to a relative path in the temporary directory of the MapReduce job, based upon the PartitionKey returned by a DynamicPartitioner class, which is implemented by the user. Then, in the commitJob of the OutputFormatCommitter, these paths are moved to the PartitionedFileSet's baseLocation, and Partitions are registered in the dataset for each of these.

https://issues.cask.co/browse/CDAP-2757
http://builds.cask.co/browse/CDAP-DUT2731-11

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
@@ -35,7 +35,7 @@ public abstract class DynamicPartitioner<K, V> {
    *    on that instance.
    *  </p>
    *  @param mapReduceTaskContext the mapReduceTaskContext for the task that this DynamicPartitioner is running in.
-   *
+   *  Note that the hadoop context is not available on this MapReduceTaskContext.
    */
   public void initialize(MapReduceTaskContext<K, V> mapReduceTaskContext) {
     // do nothing by default
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2014-2015 Cask Data, Inc.
+ * Copyright © 2015 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright © 2014-2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.dataset.lib;
+
+import co.cask.cdap.api.mapreduce.MapReduceTaskContext;
+
+/**
+ * Responsible for dynamically determining a @{link PartitionKey}.
+ * For each K, V pair, the getPartitionKey(K, V) method is called to determine a PartitionKey.
+ *
+ * @param <K> Type of key
+ * @param <V> Type of value
+ */
+public abstract class DynamicPartitioner<K, V> {
+
+
+  /**
+   *  Initializes a DynamicPartitioner.
+   *  <p>
+   *    This method will be called only once per {@link DynamicPartitioner} instance. It is the first method call
+   *    on that instance.
+   *  </p>
+   *  @param mapReduceTaskContext the mapReduceTaskContext for the task that this DynamicPartitioner is running in.
+   *
+   */
+  public void initialize(MapReduceTaskContext<K, V> mapReduceTaskContext) {
+    // do nothing by default
+  }
+
+  /**
+   *  Destroys a DynamicPartitioner.
+   *  <p>
+   *    This method will be called only once per {@link DynamicPartitioner} instance. It is the last method call
+   *    on that instance.
+   *  </p>
+   */
+  public void destroy() {
+    // do nothing by default
+  }
+
+  /**
+   * Determine the PartitionKey for the key-value pair to be written to.
+   *
+   * @param key the key to be written
+   * @param value the value to be written
+   * @return the {@link PartitionKey} for the key-value pair to be written to.
+   */
+  public abstract PartitionKey getPartitionKey(K key, V value);
+}
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceTaskContextProvider.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceTaskContextProvider.java
@@ -55,7 +55,7 @@ public final class MapReduceTaskContextProvider {
   private BasicMapReduceTaskContext context;
   private AbstractMapReduceTaskContextBuilder contextBuilder;
 
-  public MapReduceTaskContextProvider(TaskAttemptContext context, MapReduceMetrics.TaskType type) {
+  public MapReduceTaskContextProvider(TaskAttemptContext context, @Nullable MapReduceMetrics.TaskType type) {
     this.taskContext = context;
     this.type = type;
     this.contextConfig = new MapReduceContextConfig(context.getConfiguration());
@@ -71,7 +71,7 @@ public final class MapReduceTaskContextProvider {
    * inside cannot load program classes. It is used for the cases where only the application specification is needed,
    * but no need to load any class from it.
    */
-  public synchronized BasicMapReduceTaskContext get() {
+  public synchronized <K, V> BasicMapReduceTaskContext<K, V> get() {
     if (context == null) {
       CConfiguration cConf = contextConfig.getConf();
       context = getBuilder(cConf)
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
@@ -19,6 +19,7 @@ package co.cask.cdap.internal.app.runtime.batch.dataset.partitioned;
 import co.cask.cdap.api.dataset.lib.PartitionKey;
 import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
 import co.cask.cdap.api.dataset.lib.Partitioning;
+import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.internal.app.runtime.batch.BasicMapReduceTaskContext;
 import co.cask.cdap.internal.app.runtime.batch.MapReduceTaskContextProvider;
 import com.google.common.collect.Lists;
@@ -69,11 +70,12 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
   public void commitJob(JobContext context) throws IOException {
     BasicMapReduceTaskContext mrTaskContext = new MapReduceTaskContextProvider(taskContext, null).get();
     String outputDatasetName =
-      context.getConfiguration().get(DynamicPartitioningOutputFormat.HCONF_ATTR_OUTPUT_DATASET);
+      context.getConfiguration().get(Constants.Dataset.Partitioned.HCONF_ATTR_OUTPUT_DATASET);
     PartitionedFileSet outputDataset = mrTaskContext.getDataset(outputDatasetName);
     Partitioning partitioning = outputDataset.getPartitioning();
 
     Set<PartitionKey> partitionsToAdd = new HashSet<>();
+    Set<String> relativePaths = new HashSet<>();
     // Go over all files in the temporary directory and keep track of partitions to add for them
     FileStatus[] allCommittedTaskPaths = getAllCommittedTaskPaths(context);
     for (FileStatus committedTaskPath : allCommittedTaskPaths) {
@@ -103,6 +105,7 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
         }
         PartitionKey partitionKey = getPartitionKey(partitioning, relativeDir);
         partitionsToAdd.add(partitionKey);
+        relativePaths.add(relativeDir);
       }
     }
 
@@ -118,6 +121,7 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
     for (PartitionKey partitionKey : partitionsToAdd) {
       outputDataset.getPartitionOutput(partitionKey).addPartition();
     }
+
     // close the TaskContext, which flushes dataset operations
     try {
       mrTaskContext.flushOperations();
@@ -129,7 +133,15 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
 
     // delete the job-specific _temporary folder and create a _done file in the o/p folder
     cleanupJob(context);
-    // this commitJob() omits the marking with '_SUCCESS' file, because the output directory is shared across jobs.
+
+    // mark all the final output paths with a _SUCCESS file, if configured to do so (default = true)
+    if (context.getConfiguration().getBoolean(SUCCESSFUL_JOB_OUTPUT_DIR_MARKER, true)) {
+      for (String relativePath : relativePaths) {
+        Path pathToMark = new Path(finalOutput, relativePath);
+        Path markerPath = new Path(pathToMark, SUCCEEDED_FILE_NAME);
+        fs.create(markerPath).close();
+      }
+    }
   }
 
   private PartitionKey getPartitionKey(Partitioning partitioning, String relativePath) {
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
@@ -107,14 +107,6 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
       }
     }
 
-    try {
-      mrTaskContext.flushOperations();
-    } catch (Exception e) {
-      throw new IOException(e);
-    } finally {
-      mrTaskContext.close();
-    }
-
     // We need to copy to the parent of the FileOutputFormat's outputDir, since we added a _temporary_jobId suffix to
     // the original outputDir.
     Path finalOutput = FileOutputFormat.getOutputPath(context);
@@ -126,6 +118,14 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
     // create all the necessary partitions
     for (PartitionKey partitionKey : partitionsToAdd) {
       outputDataset.getPartitionOutput(partitionKey).addPartition();
+    }
+    // close the TaskContext, which flushes dataset operations
+    try {
+      mrTaskContext.flushOperations();
+    } catch (Exception e) {
+      throw new IOException(e);
+    } finally {
+      mrTaskContext.close();
     }
 
     // delete the job-specific _temporary folder and create a _done file in the o/p folder
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
@@ -22,7 +22,6 @@ import co.cask.cdap.api.dataset.lib.Partitioning;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.internal.app.runtime.batch.BasicMapReduceTaskContext;
 import co.cask.cdap.internal.app.runtime.batch.MapReduceTaskContextProvider;
-import com.google.common.collect.Lists;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.LocatedFileStatus;
@@ -38,6 +37,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -139,13 +139,19 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
       for (String relativePath : relativePaths) {
         Path pathToMark = new Path(finalOutput, relativePath);
         Path markerPath = new Path(pathToMark, SUCCEEDED_FILE_NAME);
-        fs.create(markerPath).close();
+        fs.createNewFile(markerPath);
       }
     }
   }
 
   private PartitionKey getPartitionKey(Partitioning partitioning, String relativePath) {
-    List<String> pathParts = Lists.newArrayList(relativePath.split(Path.SEPARATOR));
+    List<String> pathParts = Arrays.asList(relativePath.split(Path.SEPARATOR));
+
+    if (pathParts.size() != partitioning.getFields().size()) {
+      throw new IllegalArgumentException(
+        String.format("relativePath '%s' does not have same number of components as partitioning '%s",
+                      relativePath, partitioning));
+    }
 
     PartitionKey.Builder builder = PartitionKey.builder();
     int i = 0;
@@ -223,14 +229,6 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
     return fs.listStatus(jobAttemptPath, new CommittedTaskFilter());
   }
 
-  class CommittedTaskFilter implements PathFilter {
-    @Override
-    public boolean accept(Path path) {
-      return !PENDING_DIR_NAME.equals(path.getName());
-    }
-  }
-
-
   /**
    * given two paths as input:
    *    base: /my/base/path
@@ -239,5 +237,12 @@ public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
    */
   private String getRelative(Path base, Path file) {
     return base.toUri().relativize(file.toUri()).getPath();
+  }
+
+  private static class CommittedTaskFilter implements PathFilter {
+    @Override
+    public boolean accept(Path path) {
+      return !PENDING_DIR_NAME.equals(path.getName());
+    }
   }
 }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.internal.app.runtime.batch.dataset.partitioned;
+
+import co.cask.cdap.api.dataset.lib.Partition;
+import co.cask.cdap.api.dataset.lib.PartitionKey;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
+import co.cask.cdap.api.dataset.lib.Partitioning;
+import co.cask.cdap.internal.app.runtime.batch.BasicMapReduceTaskContext;
+import co.cask.cdap.internal.app.runtime.batch.MapReduceTaskContextProvider;
+import com.google.common.collect.Lists;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.LocatedFileStatus;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+import org.apache.hadoop.fs.RemoteIterator;
+import org.apache.hadoop.mapred.FileAlreadyExistsException;
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter;
+import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An OutputCommitter which creates partitions in a configured PartitionedFileSet dataset for all of the partitions
+ * that were written to by a DynamicPartitioningOutputFormat
+ * It enables this by having each job write to a job-specific temporary path within that output directory.
+ * Then, upon commitJob, it moves the files to the final, parent directory if the final output directories do not
+ * already exist.
+ */
+public class DynamicPartitioningOutputCommitter extends FileOutputCommitter {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DynamicPartitioningOutputCommitter.class);
+
+  private final TaskAttemptContext taskContext;
+  private final Path jobSpecificOutputPath;
+
+  // Note that the outputPath passed in is treated as a temporary directory.
+  // The commitJob method moves the files from within this directory to an parent (final) directory.
+  // The cleanupJob method removes this directory.
+  public DynamicPartitioningOutputCommitter(Path outputPath, TaskAttemptContext context) throws IOException {
+    super(outputPath, context);
+    this.taskContext = context;
+    this.jobSpecificOutputPath = outputPath;
+  }
+
+  @Override
+  public void commitJob(JobContext context) throws IOException {
+    BasicMapReduceTaskContext mrTaskContext = new MapReduceTaskContextProvider(taskContext, null).get();
+    String outputDatasetName =
+      context.getConfiguration().get(DynamicPartitioningOutputFormat.HCONF_ATTR_OUTPUT_DATASET);
+    PartitionedFileSet outputDataset = mrTaskContext.getDataset(outputDatasetName);
+    Partitioning partitioning = outputDataset.getPartitioning();
+
+    Set<PartitionKey> partitionsToAdd = new HashSet<>();
+    // Go over all files in the temporary directory and keep track of partitions to add for them
+    FileStatus[] allCommittedTaskPaths = getAllCommittedTaskPaths(context);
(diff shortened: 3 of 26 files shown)


chtyim:
You need some better commit message than `wip`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/file/FileSetDataset.java
```diff
@@ -139,7 +139,8 @@ static Location determineBaseLocation(DatasetContext datasetContext, CConfigurat
 
   private Location determineOutputLocation() {
     String outputPath = FileSetArguments.getOutputPath(runtimeArguments);
-    return outputPath == null ? null : createLocation(outputPath);
+    // TODO: think about all cases of this change (for instance, FileSet as output of MR)
```
anew: yeah this seems risky. Would it cause every map/reduce that does not configure an output path and does not use dynamic partitioner to write to the base location?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -566,7 +560,16 @@ public String getInputFormatClassName() {
 
   @Override
   public String getOutputFormatClassName() {
+    if (isExternal) {
+      throw new UnsupportedOperationException(
+        "Output is not supported for external partitioned file set '" + spec.getName() + "'");
+    }
+    PartitionKey outputKey = PartitionedFileSetArguments.getOutputPartitionKey(runtimeArguments, getPartitioning());
+    if (outputKey == null) {
+      return "co.cask.cdap.internal.app.runtime.batch.dataset.partitioned.DynamicPartitioningOutputFormat";
```
anew: here, too, overloading the lack of output partition with dynamic partitioning? Can it happen that the user did not specify either? Would he still get reasonable behavior? (that is, failure with an understandable message)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -575,16 +578,28 @@ public String getOutputFormatClassName() {
       throw new UnsupportedOperationException(
         "Output is not supported for external partitioned file set '" + spec.getName() + "'");
     }
+
+    // copy the output partition key to the output arguments of the embedded file set
+    // this will be needed by the output format to register the new partition.
+    Map<String, String> outputArgs = Maps.newHashMap(files.getOutputFormatConfiguration());
+
+
```
anew: extra newline?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -485,15 +486,8 @@ private String metadataKeyFromColumnKey(byte[] columnKey) {
...
-  // package visible for PartitionedFileSetDefinition
-  static String getOutputPath(Partitioning partitioning, PartitionKey key) {
-    StringBuilder builder = new StringBuilder();
-    String sep = "";
-    for (String fieldName : partitioning.getFields().keySet()) {
-      builder.append(sep).append(key.getField(fieldName).toString());
-      sep = "/";
-    }
-    return builder.toString();
+  public static String getOutputPath(PartitionKey key) {
```
anew: so, previously this would fail if the partition key is incomplete. (not with a good message, though). Now it simply uses the given key to construct a path. Also, previously, the order of the fields in the path was given by the partitioning (fixed and always he same). Now it depends on the key (which is not guaranteed to be in the same order).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * This class extends the FileOutputFormat and allows writing dynamically to multiple partitions of a PartitionedFileSet
+ * Dataset.
+ *
+ * @param <K> Type of key
+ * @param <V> Type of value
+ */
+public class DynamicPartitioningOutputFormat<K, V> extends FileOutputFormat<K, V> {
+//      TODO: Hive doesn't allow addPartition w/o the partition location existing. verify this?
```
anew: I am pretty sure that this is true.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,173 @@
...
+   * @return a composite record writer
+   * @throws IOException
+   */
+  @Override
+  public RecordWriter<K, V> getRecordWriter(final TaskAttemptContext job) throws IOException {
+
+    final String outputName = FileOutputFormat.getOutputName(job);
+
+    Class<? extends DynamicPartitioner> className = job.getConfiguration()
+      .getClass(PartitionedFileSetArguments.DYNAMIC_PARTITIONER_CLASS_NAME, null, DynamicPartitioner.class);
```
anew: can that be null?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceTaskContextProvider.java
```diff
@@ -55,7 +55,7 @@
   private BasicMapReduceTaskContext context;
   private AbstractMapReduceTaskContextBuilder contextBuilder;
 
-  public MapReduceTaskContextProvider(TaskAttemptContext context, MapReduceMetrics.TaskType type) {
+  public MapReduceTaskContextProvider(TaskAttemptContext context, @Nullable MapReduceMetrics.TaskType type) {
```
anew: what is type ```used``` for? I don't see it...

anew: Is it for metrics? So if we use multi-partition output, the metrics collector will have null as its type?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: Yes, the getOutputFormatConfiguration method will throw a DatasetException if neither is set in the runtime arguments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
anwar6953: I put a check in checkOutputSpecs method for this.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: removed.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: reverted this change.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
anwar6953: ok, I made it so that the files are moved to the file location before the partitions are added.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceTaskContextProvider.java
anwar6953: I updated it to determine the TaskType based upon TaskAttemptContext's taskType.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,189 @@
...
+
+    dynamicPartitioner.initialize(mrTaskContext);
+
+    return new RecordWriter<K, V>() {
+
+      // a cache storing the record writers for different output files.
+      TreeMap<String, RecordWriter<K, V>> recordWriters = new TreeMap<>();
+
+      public void write(K key, V value) throws IOException, InterruptedException {
+        PartitionKey partitionKey = dynamicPartitioner.getPartitionKey(key, value);
```
anew: perhaps it would be better to use the partition key as the key in the cache. Otherwise, we create the output path, and the the final path for every single record that is written. That is a lot of overhead in creating the strings.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,189 @@
...
+
+      @Override
+      public void close(TaskAttemptContext context) throws IOException, InterruptedException {
+        try {
+          for (RecordWriter<K, V> recordWriter : this.recordWriters.values()) {
+            recordWriter.close(context);
+          }
+          this.recordWriters.clear();
+
+          mrTaskContext.flushOperations();
```
anew: why is this needed? Do we perform any writes to datasets (not through the record writers)?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,189 @@
...
+    return new TaskAttemptContextImpl(job.getConfiguration(), context.getTaskAttemptID());
+  }
+
+  /**
+   * @return A RecordWriter object over the given file
+   * @throws IOException
+   */
+  protected RecordWriter<K, V> getBaseRecordWriter(TaskAttemptContext job) throws IOException, InterruptedException {
+    if (fileOutputFormat == null) {
+      // TODO: use the PFS's OutputFormat (encode in conf)
```
anew: there is still a todo here

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,189 @@
...
+
+  // suffixes a Path with a job-specific string
+  private static Path createJobSpecificPath(Path path, JobContext jobContext) {
+    String outputPathSuffix = "_temporary_" + jobContext.getJobID().getId();
+    return new Path(path, outputPathSuffix);
+  }
+
+  @Override
+  public synchronized FileOutputCommitter getOutputCommitter(TaskAttemptContext context) throws IOException {
+    if (committer == null) {
```
anew: Does this have to be through an instance variable? That is, will this ever be called multiple times? And if so, does it matter whether it returns the same instance?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
anwar6953: We create the mrTaskContext and pass it to the dynamicPartitioner on line 86, so we close the mrTaskContext here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,189 @@
...
+    // get delegation token for outDir's file system
+    TokenCache.obtainTokensForNamenodes(job.getCredentials(),
+                                        new Path[]{outDir}, job.getConfiguration());
+
+    // we permit multiple jobs writing to the same output directory. We handle this by each one writing to distinct
+    // paths within that directory. See createJobSpecificPath method and usages of it.
+
+    // additionally check that output dataset and dynamic partitioner class name has been set in conf
+    if (job.getConfiguration().get(HCONF_ATTR_OUTPUT_DATASET) == null) {
+      throw new InvalidJobConfException("Output dataset not set.");
```
anew: A better error message would be "The job configuration does not contain required property " + HCONF_ATTR_OUTPUT_DATASET

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    Class<? extends DynamicPartitioner> className = job.getConfiguration()
+      .getClass(PartitionedFileSetArguments.DYNAMIC_PARTITIONER_CLASS_NAME, null, DynamicPartitioner.class);
+    if (className == null) {
+      throw new InvalidJobConfException("Dynamic Partitioner class name not set.");
```
anew: similar here,

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/file/FileSetDataset.java
anew: can you answer this question?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
```diff
@@ -0,0 +1,237 @@
...
+        PartitionKey partitionKey = getPartitionKey(partitioning, relativeDir);
+        partitionsToAdd.add(partitionKey);
+      }
+    }
+
+    // We need to copy to the parent of the FileOutputFormat's outputDir, since we added a _temporary_jobId suffix to
+    // the original outputDir.
+    Path finalOutput = FileOutputFormat.getOutputPath(context);
+    FileSystem fs = finalOutput.getFileSystem(context.getConfiguration());
+    for (FileStatus stat: getAllCommittedTaskPaths(context)) {
```
anew: missing blank before : (I think our convention requires it)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
```diff
@@ -0,0 +1,237 @@
...
+      throw new IOException(e);
+    } finally {
+      mrTaskContext.close();
+    }
+
+    // delete the job-specific _temporary folder and create a _done file in the o/p folder
+    cleanupJob(context);
+    // True if the job requires output.dir marked on successful job.
+    // Note that by default it is set to true.
+    if (context.getConfiguration().getBoolean(SUCCESSFUL_JOB_OUTPUT_DIR_MARKER, true)) {
```
anew: this would leave a done marker in the file set's base location?
would it make more sense to add a marker in each partition's relative location?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
anwar6953: checkstyle didn't complain, but yea - that is how we commonly do it.

anew:
@anwar6953 I did another pass, please see my comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
anwar6953: I entirely removed this (writing of _SUCCESS file) instead.

anew: why? For partitions that are written without dynamic partitioning, the _SUCCESS file will be there. Should dynamic partitions be different?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
```diff
@@ -575,16 +588,27 @@ public String getOutputFormatClassName() {
...
+      PartitionedFileSetArguments.setOutputPartitionKey(outputArgs, outputKey);
+    } else {
+      String dynamicPartitionerClassName = PartitionedFileSetArguments.getDynamicPartitioner(runtimeArguments);
+      if (dynamicPartitionerClassName == null) {
+        throw new DataSetException(
+          "Either a Partition key or a DynamicPartitioner class must be given as a runtime argument.");
+      }
+      // also delegate to the files.getOutputFormatClassName()? (currently defaulting to TextOutputFormat)
+      PartitionedFileSetArguments.setDynamicPartitioner(outputArgs, dynamicPartitionerClassName);
+      outputArgs.put("output.format.class.name", files.getOutputFormatClassName());
```
anew: why are we not using the constants defined elsewhere?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
anwar6953: good point; added.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java
anwar6953: The constants were defined in app-fabric (where the DynamicPartitioningOutputFormat is).
I moved the constants to Constants.Dataset.Partitioned for usage from both places.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,201 @@
...
+   * @throws IOException
+   */
+  @Override
+  public RecordWriter<K, V> getRecordWriter(final TaskAttemptContext job) throws IOException {
+    final String outputName = FileOutputFormat.getOutputName(job);
+
+    Class<? extends DynamicPartitioner> className = job.getConfiguration()
+      .getClass(PartitionedFileSetArguments.DYNAMIC_PARTITIONER_CLASS_NAME, null, DynamicPartitioner.class);
+
+    @SuppressWarnings("unchecked")
```
chtyim: Why needed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,201 @@
...
+
+      @Override
+      public void close(TaskAttemptContext context) throws IOException, InterruptedException {
+        try {
+          for (RecordWriter<K, V> recordWriter : this.recordWriters.values()) {
+            recordWriter.close(context);
+          }
+          this.recordWriters.clear();
+
+          dynamicPartitioner.destroy();
```
chtyim: destroy should be called in finally.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
anwar6953: The call to create() on the Instantiator returns a DynamicPartitioner, without the K, V type parameters.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,201 @@
...
+          this.recordWriters.put(partitionKey, rw);
+        }
+        rw.write(key, value);
+      }
+
+      @Override
+      public void close(TaskAttemptContext context) throws IOException, InterruptedException {
+        try {
+          for (RecordWriter<K, V> recordWriter : this.recordWriters.values()) {
+            recordWriter.close(context);
```
chtyim: If any of the readerWriter.close() throws exception, then the rest will not get closed. Will this be leaking resources?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,201 @@
...
+          }
+          this.recordWriters.clear();
```
chtyim: Why needed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * @return A RecordWriter object over the given file
+   * @throws IOException
+   */
+  protected RecordWriter<K, V> getBaseRecordWriter(TaskAttemptContext job) throws IOException, InterruptedException {
+    if (fileOutputFormat == null) {
+      Class<? extends FileOutputFormat> delegateOutputFormat = job.getConfiguration()
+        .getClass(Constants.Dataset.Partitioned.HCONF_ATTR_OUTPUT_FORMAT_CLASS_NAME, null, FileOutputFormat.class);
+
+      @SuppressWarnings("unchecked")
```
chtyim: Why needed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
chtyim: But I don't see anything casting in this line.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
```diff
@@ -0,0 +1,201 @@
...
+   * Create a composite record writer that can write key/value data to different output files.
+   *
+   * @return a composite record writer
+   * @throws IOException
+   */
+  @Override
+  public RecordWriter<K, V> getRecordWriter(final TaskAttemptContext job) throws IOException {
+    final String outputName = FileOutputFormat.getOutputName(job);
+
+    Class<? extends DynamicPartitioner> className = job.getConfiguration()
```
chtyim: This is not class name, it's a class. Better call it `partitionerClass`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
anwar6953: I put a static method in MultipleOutputs to close all RecordWriters, and utilized it here now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
anwar6953: Otherwise, there's a warning - unchecked assignment.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceTaskContextProvider.java
chtyim: So in what case that a Task is neither a Mapper nor a Reducer?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
```diff
@@ -0,0 +1,243 @@
...
+   * @return the list of these Paths/FileStatuses.
+   * @throws IOException
+   */
+  private FileStatus[] getAllCommittedTaskPaths(JobContext context) throws IOException {
+    Path jobAttemptPath = getJobAttemptPath(context);
+    FileSystem fs = jobAttemptPath.getFileSystem(context.getConfiguration());
+    return fs.listStatus(jobAttemptPath, new CommittedTaskFilter());
+  }
+
+  class CommittedTaskFilter implements PathFilter {
```
chtyim: Why not a private static class?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+
+  /**
+   * given two paths as input:
+   *    base: /my/base/path
+   *    file: /my/base/path/some/other/file
+   * return "some/other/file"
+   */
+  private String getRelative(Path base, Path file) {
```
chtyim: Group methods together.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
```diff
@@ -0,0 +1,243 @@
...
+      for (String relativePath : relativePaths) {
+        Path pathToMark = new Path(finalOutput, relativePath);
+        Path markerPath = new Path(pathToMark, SUCCEEDED_FILE_NAME);
+        fs.create(markerPath).close();
+      }
+    }
+  }
+
+  private PartitionKey getPartitionKey(Partitioning partitioning, String relativePath) {
+    List<String> pathParts = Lists.newArrayList(relativePath.split(Path.SEPARATOR));
```
chtyim: Use `Arrays.asList` to avoid unnecessary copying.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
```diff
@@ -0,0 +1,243 @@
...
+
+    PartitionKey.Builder builder = PartitionKey.builder();
+    int i = 0;
+    for (Map.Entry<String, Partitioning.FieldType> entry : partitioning.getFields().entrySet()) {
+      String keyName = entry.getKey();
+      Comparable keyValue = entry.getValue().parse(pathParts.get(i));
```
chtyim: Is it safe to have `get(i)`? What's the contract between the number of entries in the map vs number of entries in the list?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
```diff
@@ -0,0 +1,243 @@
...
+
+    // delete the job-specific _temporary folder and create a _done file in the o/p folder
+    cleanupJob(context);
+
+    // mark all the final output paths with a _SUCCESS file, if configured to do so (default = true)
+    if (context.getConfiguration().getBoolean(SUCCESSFUL_JOB_OUTPUT_DIR_MARKER, true)) {
+      for (String relativePath : relativePaths) {
+        Path pathToMark = new Path(finalOutput, relativePath);
+        Path markerPath = new Path(pathToMark, SUCCEEDED_FILE_NAME);
+        fs.create(markerPath).close();
```
chtyim: It is always a new file? If so, better call `fs.createNewFile(markerPath);`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
anwar6953: I put a check above it to assert on size equality:
`pathParts.size() == partitioning.getFields().size()`

chtyim: But other than than the checking, does the logic who generates the map and the list guarantees that they will have the same size (under normal situation)?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceTaskContextProvider.java
anwar6953: In the FileOutputFormat#commitJob, its not from a mapper nor a reducer, but we still instantiate the MapReduceTaskContext, to get access to dataset.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputCommitter.java
anwar6953: Yea, the list is generated from the relative path of the files written (which is based upon the PartitionKey given by the user).
And the map comes from the partitioning of the dataset.
So as long as the PartitionKey that user returns is same size as the dataset's Partitioning, it'll be ok (except for the case that PartitionKey has '/' in one of its components).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceTaskContextProvider.java
chtyim: Then shall we have a TaskType of COMMITTER?

chtyim: otherwise there will be no metric or no context for the metric?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/dataset/partitioned/DynamicPartitioningOutputFormat.java
anwar6953: the warning was for unchecked assignment, not unchecked casting

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceTaskContextProvider.java
anwar6953: There will still be a metric context, it will simply omit the taskId and taskType as part of its tags.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright © 2014-2015 Cask Data, Inc.
```
bdmogal: 2015

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   *  Destroys a DynamicPartitioner.
+   *  <p>
+   *    This method will be called only once per {@link DynamicPartitioner} instance. It is the last method call
+   *    on that instance.
+   *  </p>
+   */
+  public void destroy() {
```
bdmogal: Hmmm, we have a ``Destroyable`` interface in ``cdap-etl-api`` that defines a ``destroy()`` method. Should we move it to ``cdap-api`` and re-use instead of re-defining this method here? Maybe not this PR, but a jira?

anwar6953: There're a lot of classes that have a destroy method like this, so opened a JIRA for future consideration.
https://issues.cask.co/browse/CDAP-3643

bdmogal: :+1:

anew:
All my concerns have been addressed. LGTM as soon as @chtyim is ok with it.

cdap-examples/DataCleansing/src/test/java/co/cask/cdap/examples/datacleansing/DataCleansingMapReduceTest.java
```diff
@@ -168,21 +185,23 @@ private void compareData(Long time, String dsName, Set<String> expectedRecords)
...
-    for (Location location : partitionLocation.list()) {
-      if (location.getName().startsWith("part-")) {
-        try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(location.getInputStream()))) {
-          String line;
-          while ((line = bufferedReader.readLine()) != null) {
-            cleanData.add(line);
+    for (PartitionDetail partition : partitions) {
+      Location partitionLocation = partition.getLocation();
+      for (Location location : partitionLocation.list()) {
+        if (location.getName().startsWith("part-")) {
```
bdmogal: This is existing code, but is it safe to make the assumption that all files start with ``part-``? This is controlled by the configuration [mapreduce.output.basename](https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputFormat.java#L50). Its default value is ``part`` and you get its current value by using [FileOutputFormat.getOutputName()](https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/output/FileOutputFormat.java#L293).

anwar6953: Since this is a test case for a CDAP application we wrote (DataCleansing), and in that CDAP app's MapReduce, we are in control of the job's configuration, I think it's safe to make this assumption.

It wouldn't be okay to make the assumption if this was code for a generic user-written MapReduce, where the user could override the mapreduce.output.basename key, right?

bdmogal: oh oops. didn't notice that this was an example test :-) nvm.

cdap-api/src/main/java/co/cask/cdap/api/dataset/lib/DynamicPartitioner.java
chtyim: Added my comment in the issue.

chtyim:
LGTM

PR #3849, opened 2015-09-08
[CDAP-3517] [CDAP-3520]

bdmogal:
Added a BusinessMetadataStore as a transactional wrapper around BusinessMetadataDataset. Also replaced MockMetadataAdmin with a real MetadataAdmin that talks to the BusinessMetadataStore.

Build: 
http://builds.cask.co/browse/CDAP-DUT2768

Jiras: 
[CDAP-3517](https://issues.cask.co/browse/CDAP-3517)
[CDAP-3520](https://issues.cask.co/browse/CDAP-3520)


bdmogal:
This PR is missing thorough unit tests. However, I will add that in a later PR after we figure out where to run Metadata REST APIs (They have been added in the Dataset Service for now).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -147,4 +241,11 @@ private String getTargetType(Id.NamespacedId namespacedId) {
     }
     return namespacedId.getClass().getSimpleName();
   }
+
+  private String getMetadataKey(byte [] rowKey) {
```
henry-cask: Small nit: extra space between byte and []

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -85,12 +91,100 @@ public BusinessMetadataRecord getBusinessMetadata(Id.NamespacedId targetId, Stri
...
+   * @param targetId the specified {@link Id.NamespacedId}
+   * @return a Map representing the metadata for the specified {@link Id.NamespacedId}
+   */
+  public Map<String, String> getBusinessMetadata(Id.NamespacedId targetId) {
+    String targetType = getTargetType(targetId);
+    MDSKey mdsKey = new MDSKey.Builder().add(targetType).add(targetId.toString()).build();
+    byte[] startKey = mdsKey.getKey();
+    byte[] stopKey = Bytes.stopKeyForPrefix(startKey);
+
+    Map<String, String> metadata = new LinkedHashMap<>();
```
poornachandra: Any reason to use `LinkedHashMap`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -85,12 +91,100 @@ public BusinessMetadataRecord getBusinessMetadata(Id.NamespacedId targetId, Stri
...
+   * Removes the specified keys from the business metadata of the specified {@link Id.NamespacedId}.
+   *
+   * @param targetId the {@link Id.NamespacedId} for which the specified metadata keys are to be removed.
+   * @param keys the keys to remove from the metadata of the specified {@link Id.NamespacedId}
+   */
+  public void removeMetadata(Id.NamespacedId targetId, final String ... keys) {
+    removeMetadata(targetId, new Predicate<String>() {
+      @Override
+      public boolean apply(String input) {
+        return Arrays.asList(keys).contains(input);
```
poornachandra: I think it is better to convert to list outside the predicate, otherwise the array to list conversion may get called for each `apply()` call. Also can we use set instead of list, since contains on a list is `O(n)` operation.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -85,12 +91,100 @@ public BusinessMetadataRecord getBusinessMetadata(Id.NamespacedId targetId, Stri
...
+    byte[] prefix = mdsKey.getKey();
+    byte[] stopKey = Bytes.stopKeyForPrefix(prefix);
+
+    Scanner scan = indexedTable.scan(prefix, stopKey);
+    try {
+      Row next;
+      while ((next = scan.next()) != null) {
+        String keyValue = next.getString(KEYVALUE_COLUMN);
+        String value = next.getString(VALUE_COLUMN);
+        if (keyValue == null || value == null) {
```
poornachandra: Should we use `&&` instead of `||` to be safe here?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -85,12 +91,100 @@ public BusinessMetadataRecord getBusinessMetadata(Id.NamespacedId targetId, Stri
...
+          continue;
+        }
+        if (filter.apply(getMetadataKey(next.getRow()))) {
+          indexedTable.delete(new Delete(next.getRow()).add(KEYVALUE_COLUMN, VALUE_COLUMN));
```
poornachandra: We can delete the whole row here, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -147,4 +241,11 @@ private String getTargetType(Id.NamespacedId namespacedId) {
     }
     return namespacedId.getClass().getSimpleName();
   }
+
+  private String getMetadataKey(byte [] rowKey) {
+    MDSKey.Splitter keySplitter = new MDSKey(rowKey).split();
+    keySplitter.skipString();
```
poornachandra: It would be good to have a visual representation of the key components in comments.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+      }
+    });
+  }
+
+  private Iterable<String> getTags(Map<String, String> businessMetadata) {
+    if (!businessMetadata.containsKey("tags")) {
+      return new ArrayList<>();
+    }
+    String tags = businessMetadata.get("tags");
+    return Splitter.on(",").omitEmptyStrings().split(tags);
```
henry-cask: Should we add "trimResults" to be safe ?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+      @Override
+      public BusinessMdsIterable get() {
+        try {
+          Object dataset = DatasetsUtil.getOrCreateDataset(dsFramework, BUSINESS_METADATA_INSTANCE_ID,
+                                                            BusinessMetadataDataset.class.getSimpleName(),
+                                                            DatasetProperties.EMPTY, null, null);
+          if (dataset instanceof BusinessMetadataDataset) {
+            return new BusinessMdsIterable((BusinessMetadataDataset) dataset);
+          }
+          return new BusinessMdsIterable(new BusinessMetadataDataset((IndexedTable) dataset));
```
poornachandra: This line is not required, I think.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+   *
+   * @param entityId the {@link Id.NamespacedId} for which metadata is to be updated
+   * @param metadata metadata (represented as a map) to update
+   */
+  void addMetadata(final Id.NamespacedId entityId, final Map<String, String> metadata) {
+    txnl.executeUnchecked(new TransactionExecutor.Function<BusinessMdsIterable, Void>() {
+      @Override
+      public Void apply(BusinessMdsIterable input) throws Exception {
+        for (Map.Entry<String, String> entry : metadata.entrySet()) {
+          input.businessMds.createBusinessMetadata(entityId, entry.getKey(), entry.getValue());
```
poornachandra: What do we do here if user specifies `tags` as one of the key in `metadata` map?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * Adds tags for the specified {@link Id.NamespacedId}.
+   */
+  void addTags(final Id.NamespacedId entityId, final String ... tagsToAdd) {
+    txnl.executeUnchecked(new TransactionExecutor.Function<BusinessMdsIterable, Void>() {
+      @Override
+      public Void apply(BusinessMdsIterable input) throws Exception {
+        Map<String, String> businessMetadata = input.businessMds.getBusinessMetadata(entityId);
```
poornachandra: We can fetch only `tags` here, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+        Iterables.removeAll(existingTags, Arrays.asList(tagsToRemove));
+        input.businessMds.createBusinessMetadata(entityId, "tags", Joiner.on(",").join(existingTags));
+        return null;
+      }
+    });
+  }
+
+  private Iterable<String> getTags(Map<String, String> businessMetadata) {
+    if (!businessMetadata.containsKey("tags")) {
+      return new ArrayList<>();
```
poornachandra: Can we return `Collections.emptyList()` over here?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * Removes the specified tags from the {@link Id.NamespacedId}
+   */
+  void removeTags(final Id.NamespacedId entityId, final String ... tagsToRemove) {
+    txnl.executeUnchecked(new TransactionExecutor.Function<BusinessMdsIterable, Void>() {
+      @Override
+      public Void apply(BusinessMdsIterable input) throws Exception {
+        Map<String, String> businessMetadata = input.businessMds.getBusinessMetadata(entityId);
```
poornachandra: Again we can probably only get tags here, right?

henry-cask:
Thanks. Did a pass and add some comments.

poornachandra:
@bdmogal Minor comments, please merge after addressing them.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
bdmogal: I think we can just use ``HashMap`` here. This was going by what ``MetadataStoreDataset`` used.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
bdmogal: I actually prefer it :)

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
bdmogal: So the way I was thinking of it was, this abstraction does not care. There will be a validator layer in ``DefaultNamespaceAdmin``, that ensures that you do not add tags through this API. is that ok?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -85,12 +91,100 @@ public BusinessMetadataRecord getBusinessMetadata(Id.NamespacedId targetId, Stri
...
     return new BusinessMetadataRecord(targetId, key, Bytes.toString(value));
   }
 
+  /**
+   * Retrieves the business metadata for the specified {@link Id.NamespacedId}.
+   *
+   * @param targetId the specified {@link Id.NamespacedId}
+   * @return a Map representing the metadata for the specified {@link Id.NamespacedId}
+   */
+  public Map<String, String> getBusinessMetadata(Id.NamespacedId targetId) {
```
henry-cask: What is the String value here? JSON String version of the data?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
bdmogal: I will think about it during https://issues.cask.co/browse/CDAP-3571

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: We could change it to Map<String, BusinessMetadataRecord> so caller do no t need to  convert the result.

chtyim: In general, it's better to use some data structure/class rather than returning `Map<String, String>` because it'd be easier to define additional constraints and methods in future.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
chtyim: Using LinkedHashMap maintains the scanner ordering. Not sure if that's the guarantees we want to provide in this case.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -85,12 +91,100 @@ public BusinessMetadataRecord getBusinessMetadata(Id.NamespacedId targetId, Stri
...
+  /**
+   * Removes all business metadata for the specified {@link Id.NamespacedId}.
+   *
+   * @param targetId the {@link Id.NamespacedId} for which metadata is to be removed.
+   */
+  public void removeMetadata(Id.NamespacedId targetId) {
+    removeMetadata(targetId, new Predicate<String>() {
+      @Override
+      public boolean apply(String input) {
+        return true;
```
chtyim: Use `Predicates.alwaysTrue()`

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
chtyim: Also, use a set instead of list.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
chtyim: no space is better.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+ */
+public class BusinessMetadataStore {
+
+  private static final Id.DatasetInstance BUSINESS_METADATA_INSTANCE_ID =
+    Id.DatasetInstance.from(Id.Namespace.SYSTEM, "business.metadata");
+  private final Transactional<BusinessMdsIterable, BusinessMetadataDataset> txnl;
+
+  @Inject
+  public BusinessMetadataStore(TransactionExecutorFactory txExecutorFactory, final DatasetFramework dsFramework) {
+    this.txnl = Transactional.of(txExecutorFactory, new Supplier<BusinessMdsIterable>() {
```
chtyim: We are trying to get rid of `Transactional` as it's confusing and also doesn't provide much convenient in most cases. Try to use `TransactionExecutorFactory` instead. See

https://github.com/caskdata/cdap/blob/develop/cdap-data-fabric/src/main/java/co/cask/cdap/data/view/MDSViewStore.java

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+   * Adds tags for the specified {@link Id.NamespacedId}.
+   */
+  void addTags(final Id.NamespacedId entityId, final String ... tagsToAdd) {
+    txnl.executeUnchecked(new TransactionExecutor.Function<BusinessMdsIterable, Void>() {
+      @Override
+      public Void apply(BusinessMdsIterable input) throws Exception {
+        Map<String, String> businessMetadata = input.businessMds.getBusinessMetadata(entityId);
+        Iterable<String> existingTags = getTags(businessMetadata);
+        Iterable<String> newTags = Iterables.concat(existingTags, Arrays.asList(tagsToAdd));
+        input.businessMds.createBusinessMetadata(entityId, "tags", Joiner.on(",").join(newTags));
```
chtyim: What happen if "tag" has comma? Do we check for it? Or would it be better using some safer encoding, e.g. Json?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
bdmogal: So metadata is a map of string key to string value. The only complex structure in the value is that for the 'tags' key, we have a comma-separated string. We do not allow to store any other data structures at this point, which is why I kept it as ``Map<String, String>``

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+      }
+    });
+  }
+
+  void removeMetadata(final Id.NamespacedId entityId, final String ... keys) {
+    txnl.executeUnchecked(new TransactionExecutor.Function<BusinessMdsIterable, Void>() {
+      @Override
+      public Void apply(BusinessMdsIterable input) throws Exception {
+        input.businessMds.removeMetadata(entityId, keys);
+        return null;
```
chtyim: This is a good example of better not to use `Transactional`. The `TransactionExecutor` support executing a `Procedure` that doesn't need to return anything.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -0,0 +1,191 @@
...
+
+  private Iterable<String> getTags(Map<String, String> businessMetadata) {
+    if (!businessMetadata.containsKey("tags")) {
+      return new ArrayList<>();
+    }
+    String tags = businessMetadata.get("tags");
+    return Splitter.on(",").omitEmptyStrings().split(tags);
+  }
+
+  private static final class BusinessMdsIterable implements Iterable<BusinessMetadataDataset> {
```
chtyim: This ugly class is not needed if not using `Transactional`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
chtyim: Internally it can be represented as `Map<String, String>`. What I was commenting on is the API. If you return `Map<String, String>`, it makes usage of this class in general harder, and more difficult to refine the contract in future.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
bdmogal: We are going to add a validator for any key/value/tags as part of https://issues.cask.co/browse/CDAP-3571.

As part of this, we are going to enforce a character set [a-zA-Z0-9_-].

bdmogal:
@chtyim, I will create a new PR to address https://github.com/caskdata/cdap/pull/3849#discussion_r38989906 and https://github.com/caskdata/cdap/pull/3849#discussion_r38990715

bdmogal:
Build passed. Squashed commits. merging.

PR #3850, opened 2015-09-08
CDAP-3591 ability to write to multiple sinks in etl realtime

albertshau:
Adding the ability to write to multiple sinks in the etl realtime
application. The exact same data will be written to all sinks.
If transactional sinks are used, a failure to write to one sink
will cause that write to other sinks to fail. If non-transactional
sinks are used, a write to one sink can succeed while a write to
another sink can fail.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch-app/src/main/java/co/cask/cdap/app/etl/batch/ETLMapReduce.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch-app/src/main/java/co/cask/cdap/app/etl/batch/ETLMapReduce.java
@@ -62,10 +62,11 @@ import java.util.Map;
 public class ETLMapReduce extends AbstractMapReduce {
   public static final String NAME = ETLMapReduce.class.getSimpleName();
   private static final Logger LOG = LoggerFactory.getLogger(ETLMapReduce.class);
+  private static final Type STRING_LIST_TYPE = new TypeToken<List<String>>() { }.getType();
   private static final Gson GSON = new Gson();
 
   private BatchSource batchSource;
-  private BatchSink batchSink;
+  private List<BatchSink> batchSinks;
   private Metrics mrMetrics;
 
   // this is only visible at configure time, not at runtime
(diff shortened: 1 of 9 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT2769

cdap-app-templates/cdap-etl/cdap-etl-batch-app/src/main/java/co/cask/cdap/app/etl/batch/ETLMapReduce.java
```diff
@@ -176,6 +184,9 @@ public void initialize(MapReduceContext context) throws Exception {
 
       addTransforms(pipeline, transformInfos, context);
 
+      List<String> sinkPluginIds = GSON.fromJson(properties.get(Constants.Sink.PLUGINIDS), STRING_LIST_TYPE);
+      // multi-output for mapreduce coming in a later PR
+      String sinkPluginId = sinkPluginIds.get(0);
```
bdmogal: oh, so this is only temporary?

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/PipelineRegisterer.java
```diff
@@ -106,22 +114,41 @@ public Pipeline registerPlugins(ETLConfig config, Class errorDatasetType, Datase
...
+    List<PipelineConfigurable> sinks = new ArrayList<>();
+    for (ETLStage sinkConfig : sinkConfigs) {
+      String sinkPluginId = PluginID.from(Constants.Sink.PLUGINTYPE, sinkConfig.getName(), pluginNum).getID();
+      sinkPluginIds.add(sinkPluginId);
+
+      // try to instantiate the sink
+      PipelineConfigurable sink = configurer.usePlugin(Constants.Sink.PLUGINTYPE, sinkConfig.getName(),
+        sinkPluginId, getPluginProperties(sinkConfig));
+      if (sink == null) {
+        throw new IllegalArgumentException(String.format("No Plugin of type '%s' named '%s' was found.",
```
bdmogal: is there anyway users could recover from this error? should we tell them how in the error message?

bdmogal:
Couple minor questions, LGTM otherwise.

cdap-app-templates/cdap-etl/cdap-etl-batch-app/src/main/java/co/cask/cdap/app/etl/batch/ETLMapReduce.java
albertshau: yep later it will instantiate all sinks and write to all of them

cdap-app-templates/cdap-etl/cdap-etl-core/src/main/java/co/cask/cdap/template/etl/common/PipelineRegisterer.java
albertshau: i will add more detail, something about checking that the type and name are correct, then checking that the plugin artifact exists

PR #3851, opened 2015-09-09
Feature/ui styling hydrator left panel

astewes:
*  Restructures side panels to accommodate new Hydrator layout
*  Reorganizes group-side-panel directive
*  Adds unique header for Hydrator


ajainarayanan:
TODO:
 - Need to separate navbars to two different directives (currently being worked on another PR)
 - Need to extract hydrator specific styles out of directives and have them under hydrator feature (currently the group-side-panel directive has no styles & we might not be able to use them anywhere if we don't have generic styles)
 - This is the first draft of changes for hydrator.
    - Currently filter in the left side panel is not functional (functionality yet to be added)
    - Not all the groups can be opened at one point of time in the left side panel.
    - The bottom panel is not functional (functionality yet to be added).
    - Some design changes are yet to be implemented ( for instance: metadata edit view)
    - Some general cleanups/organize code.(least priority to be done at the end)

ajainarayanan:
:+1: LGTM :fireworks: we can re-iterate in subsequent PRs

PR #3852, opened 2015-09-09
Fix ExploreStatementTest failure

jwang47:


cdap-explore-client/src/main/java/co/cask/cdap/explore/client/MockExploreClient.java
```diff
@@ -46,8 +47,8 @@
 
   public MockExploreClient(Map<String, List<ColumnDesc>> statementsToMetadata,
                            Map<String, List<QueryResult>> statementsToResults) {
-    this.statementsToMetadata = ImmutableMap.copyOf(statementsToMetadata);
-    this.statementsToResults = ImmutableMap.copyOf(statementsToResults);
+    this.statementsToMetadata = Maps.newHashMap(statementsToMetadata);
```
bdmogal: so this makes it mutable now?

albertshau: has to be mutable because remove() is called on it elsewhere in the mock client

albertshau:
:+1:

albertshau:
since this is a mock client, i think we can just merge if the failed test is passing locally

PR #3853, opened 2015-09-09
Feature/ui adapter api integration

ajainarayanan:
- Fixes fetching sources, sinks & transforms
- Fixes fetching properties for plugins
- Fixes saving different artifact types
- Fixes all references of old adapter templates to new artifact names

#### TODO:
 Please don't merge this yet as backend artifact names are changing. We can take that change from backend before merging this.


elfenheart:
:+1: LGTM

PR #3854, opened 2015-09-09
Fix MD5 hash for BUILD.rst.

awholegunch:
Fix the build_standalone command to properly reflect the latest.
Correct an error in a path that introduced an empty path component ("//")
Added a blank echo statement to make the logs easier to read.
Corrected a logic error in the the test_includes function that wasn't handling blank target or empty MD5 hashes correctly.
Replaced echo statements with printf commands to fix the echoing of variables that contain carriage returns.

Build...http://builds.cask.co/browse/CDAP-DOB2-1/log
Deployed...http://builds.cask.co/artifact/CDAP-DOB2/shared/build-1/Docs-HTML/3.2.0-SNAPSHOT/en/index.html


cdap-docs/build.sh
```diff
@@ -408,7 +408,7 @@ function build_license_dependency_pdfs() {
 
 function build_standalone() {
   set_mvn_environment
-  MAVEN_OPTS="-Xmx1024m -XX:MaxPermSize=128m" mvn clean package -pl cdap-standalone,cdap-app-templates/cdap-etl,cdap-examples -am -amd -DskipTests -P examples,templates,dist,release,unit-tests
+  MAVEN_OPTS="-Xmx1024m -XX:MaxPermSize=128m" mvn package -pl cdap-standalone,cdap-app-templates/cdap-etl,cdap-app-templates/cdap-data-quality,cdap-examples -am -amd -DskipTests -P examples,templates,dist,release,unit-tests
```
bdmogal: why remove ``clean``?

cdap-docs/_common/common-build.sh
```diff
@@ -108,7 +108,7 @@ GITHUB="github"
 
 # BUILD.rst
 BUILD_RST="BUILD.rst"
-BUILD_RST_HASH="f54ae74bb72f9ad894766b6c0bd2d2df"
+BUILD_RST_HASH="79a46db1e9fb3d428c384bdea0c3d596"
```
jwang47: Why do we have this?

cdap-docs/build.sh
awholegunch: Good catch. I didn't notice that this had changed. Looking at the Git Blame, I can see that this changed in BUILD.rst, yet I can't see who did it. (I don't see a commit where it went from having clean to not.) I will add it back.

cdap-docs/_common/common-build.sh
awholegunch: So when the BUILD.rst changed, I notice it, as the build.sh scripts use the same values.

cdap-docs/_common/common-build.sh
```diff
@@ -258,32 +259,35 @@ function test_an_include() {
...
-  
-  if [[ "${OSTYPE}" == "darwin"* ]]; then
-    new_md5_hash=`md5 -q ${target}`
-  else
-    new_md5_hash=`md5sum ${target} | awk '{print $1}'`
-  fi
+  if [[ "x${target}" == "x" ]]; then
+    m="No target is set for test_an_include"
+  else  
+    local file_name=`basename ${target}`
```
wolf31o2: You use ticks in some places and dollar parentheses in others. I suggest standardizing on the latter.

cdap-docs/_common/common-build.sh
```diff
@@ -258,32 +259,35 @@ function test_an_include() {
...
-  local file_name=`basename ${target}`
-  
-  if [[ "${OSTYPE}" == "darwin"* ]]; then
-    new_md5_hash=`md5 -q ${target}`
-  else
-    new_md5_hash=`md5sum ${target} | awk '{print $1}'`
-  fi
+  if [[ "x${target}" == "x" ]]; then
+    m="No target is set for test_an_include"
+  else
```
wolf31o2: Reverse your conditional. The main code shouldn't ever be in "else" as "else" should be the fallback case.

cdap-docs/_common/common-build.sh
awholegunch: Ahh, back-ticks, right? Fixed.

wolf31o2: I would have gone the other way, but the main thing was being consistent. Back-ticks are fine.

wolf31o2:
:+1: LGTM

PR #3857, opened 2015-09-09
[CDAP-3495] Persisted the runtime arguments with which a program was …

bdmogal:
…started in its RunRecord's properties.

The runtime arguments are stored in the properties of the RunRecord with the key 'runtimeArgs'.
Also removed some unused code including the existing, unused code for persisting runtime arguments.

This will eventually be used to surface the runtime info including runtime arguments, workflow token, etc in a lineage diagram. 

Jira: [CDAP-3495](https://issues.cask.co/browse/CDAP-3475)
Build: http://builds.cask.co/browse/CDAP-DUT2774


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
```diff
@@ -139,7 +138,10 @@ public void recordProgramStart(Id.Program program, String pid, long startTs, Str
       .add(pid)
       .build();
 
-    write(key, new RunRecordMeta(pid, startTs, null, ProgramRunStatus.RUNNING, adapter, null, twillRunId));
+    Type runtimeArgsType = new TypeToken<Map<String, String>>() { }.getType();
```
albertshau: this can be moved to be a constant

albertshau:
minor comment, otherwise lgtm

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/AppMetadataStore.java
bdmogal: Its only used in this method. Would it be helpful to turn it into a constant?

albertshau: its more so that we don't write one each time the method is called

bdmogal: Fixed.

bdmogal:
Squashed commits. Build green. Merging.

PR #3858, opened 2015-09-09
Remove README.md files from examples

awholegunch:
Organizes the modules in the deploy_pom.xml alphabetically (though this pom.xml seems to be ignored currently)
Replaces a reference to README.md with README.rst in the rat-check
Removes obsolete README.md files and consolidates their information either in a Java file or the main README.rst.

Fix for https://issues.cask.co/browse/CDAP-3609.


cdap-examples/CubeService/src/main/java/co/cask/cdap/examples/cube/CubeService.java
```diff
@@ -22,7 +22,7 @@
 import co.cask.cdap.api.dataset.lib.cube.Cube;
 
 /**
- * Simple Cube App with Cube Service
+ * Simple Cube App with a Cube Service to access cube data
```
bdmogal: nit: cube is very repetitive here

bdmogal:
minor nit, else LGTM

PR #3859, opened 2015-09-09
Creates a 'Snapshot' dataset that will keep only the most recent run …

ataylor4:
…in HDFS. 

http://builds.cask.co/browse/CDAP-DUT2771-1

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/FileSetSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/FileSetSink.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.template.etl.batch.sink;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.dataset.lib.FileSet;
+import co.cask.cdap.template.etl.api.batch.BatchSink;
+import co.cask.cdap.template.etl.api.batch.BatchSinkContext;
+
+import java.util.HashMap;
+
+/**
+ * File Batch Sink class that stores data in a file
+ * @param <KEY_OUT> the type of key the sink outputs
+ * @param <VAL_OUT> the type of value the sink outputs
+ */
+public abstract class FileSetSink<KEY_OUT, VAL_OUT> extends BatchSink<StructuredRecord, KEY_OUT, VAL_OUT> {
+  public static final String NAME_DESC = "Name of the Fileset Dataset to which the records " +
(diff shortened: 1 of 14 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/FileSetSinkConfig.java
```diff
@@ -0,0 +1,40 @@
...
+ * {@link PluginConfig} for {@link FileSetSink}.
+ */
+public abstract class FileSetSinkConfig extends PluginConfig {
+  @Name(Properties.SnapshotFileSet.NAME)
+  @Description(FileSetSink.NAME_DESC)
+  protected String name;
+
+  @Name(Properties.SnapshotFileSet.PATH)
+  @Description(FileSetSink.PATH_DESC)
+  protected String path;
```
albertshau: think it would be better for this to be the basePath like we have for the tpfs sinks. Somebody might not want the directory to be in the cdap directory.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/SnapshotFileSetAvroSink.java
```diff
@@ -0,0 +1,111 @@
...
+  public SnapshotFileSetAvroSink(SnapshotAvroSinkConfig config) {
+    super(config);
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+    String tpfsName = config.name;
+    String basePath = config.name;
```
albertshau: this should be configurable

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/SnapshotFileSetParquetSink.java
```diff
@@ -0,0 +1,116 @@
...
+  public SnapshotFileSetParquetSink(SnapshotParquetSinkConfig config) {
+    super(config);
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+    String tpfsName = config.name;
+    String basePath = config.name;
```
albertshau: this should be configurable

albertshau:
please also add corresponding docs to cdap-docs/application-templates/source/etl/templates/sinks/batch/

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/batch/sink/FileSetSink.java
```diff
@@ -0,0 +1,47 @@
...
+import co.cask.cdap.template.etl.api.batch.BatchSinkContext;
+
+import java.util.HashMap;
+
+/**
+ * File Batch Sink class that stores data in a file
+ * @param <KEY_OUT> the type of key the sink outputs
+ * @param <VAL_OUT> the type of value the sink outputs
+ */
+public abstract class FileSetSink<KEY_OUT, VAL_OUT> extends BatchSink<StructuredRecord, KEY_OUT, VAL_OUT> {
```
sreevatsanraman: can we call this class FileBatchSink to be consistent with the source. Also why is this abstract We should be using it as a Sink if needed right?

albertshau: abstract because you always need an output format and a corresponding way of translating record to the object the output format needs

albertshau:
lgtm, though please add docs at cdap-docs/application-templates/source/etl/templates/sinks/batch/ in a later PR

PR #3861, opened 2015-09-09
Move MockAppConfigurer into test source.

chtyim:
- It shouldn’t be in main.
- Make cdap-integration-test depends on app-fabric test-jar


cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
```diff
@@ -125,7 +126,7 @@ public ApplicationManager deployApplication(Id.Namespace namespace, Class<? exte
       if (configObject != null) {
         appConfig = GSON.toJson(configObject);
       } else {
-        configObject = (Config) configToken.getRawType().newInstance();
```
chtyim: IntelliJ complains about this.

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
```diff
@@ -108,6 +108,7 @@ public ApplicationManager deployApplication(Id.Namespace namespace,
   }
 
   @Override
+  @SuppressWarnings("unchecked")
```
bdmogal: is this needed even after the changes below?

bdmogal:
LGTM

cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
chtyim: yes, because there is still unsafe casting.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2779-1

PR #3862, opened 2015-09-09
Moved RecordWritableConverter to cdap-etl-common

ataylor4:
To allow visibility in cdap plugins repo.

http://builds.cask.co/browse/CDAP-DUT2780-1


albertshau:
:+1:

PR #3865, opened 2015-09-09
CDAP-3523 Implement REST APIs for searching Metadata.

henry-cask:
First drop for REST API to support search on metadata based on query text and type.

Modify MetadataHttpHandler and MetadataHttpAdmin service to support GET /search?query="key=value"&type="Program"

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataAdmin.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataAdmin.java
@@ -20,6 +20,8 @@ import co.cask.cdap.common.ApplicationNotFoundException;
 import co.cask.cdap.common.NamespaceNotFoundException;
 import co.cask.cdap.common.NotFoundException;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.MetadataSearchResultRecord;
+import co.cask.cdap.proto.MetadataSearchTargetType;
 
 import java.util.Map;
 
@@ -80,4 +82,16 @@ public interface MetadataAdmin {
    * @throws ApplicationNotFoundException if the application is not found
    */
   void removeTags(Id.NamespacedId appId, String ... tags) throws NotFoundException;
+
+  /**
+   * Execute search for metadata for particular type of CDAP object.
+   *
+   * @param searchQuery The query need to be executed for the search.
+   * @param type The particular type of CDAP object that the metadata need to be searched.
+   *
+   * @return a {@link Iterable} records for metadata search.
+   * @throws NotFoundException if there is not record found for particular query text.
+   */
+  Iterable<MetadataSearchResultRecord> searchMetadataOnType(String searchQuery, MetadataSearchTargetType type)
+    throws NotFoundException;
 }
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataHttpHandler.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataHttpHandler.java
@@ -19,13 +19,18 @@ package co.cask.cdap.data2.metadata.service;
 import co.cask.cdap.common.BadRequestException;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.MetadataSearchResultRecord;
+import co.cask.cdap.proto.MetadataSearchTargetType;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.http.AbstractHttpHandler;
 import co.cask.http.HttpResponder;
+
 import com.google.common.base.Charsets;
+import com.google.common.base.Strings;
 import com.google.common.collect.Sets;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+
 import org.jboss.netty.buffer.ChannelBuffer;
 import org.jboss.netty.buffer.ChannelBufferInputStream;
 import org.jboss.netty.handler.codec.http.HttpRequest;
@@ -35,7 +40,6 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.lang.reflect.Type;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import javax.ws.rs.DELETE;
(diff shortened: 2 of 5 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataAdmin.java
```diff
@@ -80,4 +81,15 @@
...
+  /**
+   * Execute search for metadata for particular type of CDAP object.
+   *
+   * @param searchQuery The query need to be executed for the search.
+   * @param type The particular type of CDAP object that the metadata need to be searched.
+   *
+   * @return a {@link Iterable} records for metadata search.
+   * @throws NotFoundException if there is not record found for particular query text.
+   */
+  Iterable<MetadataSearchResultRecord> searchMetadataOnType(String searchQuery, String type) throws NotFoundException;
```
bdmogal: Should we turn the type into an enum? So it cannot accept any random strings?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataHttpHandler.java
```diff
@@ -343,4 +347,21 @@ public void removeStreamTags(HttpRequest request, HttpResponder responder,
...
+  @Path("/namespaces/{namespace-id}/search")
+  public void searchMetadata(HttpRequest request, HttpResponder responder,
+                             @PathParam("namespace-id") String namespaceId,
+                             @QueryParam("query") String searchQuery,
+                             @QueryParam("target") String target) throws Exception {
+    if (Strings.isNullOrEmpty(target)) {
+      throw new BadRequestException(String.format("Invalid type for search '%s'", searchQuery));
+    }
+
+    Iterable<MetadataSearchResultRecord> results = metadataAdmin.searchMetadataOnType(searchQuery, target);
```
bdmogal: until we finish implementation of searchMetadataOnType, should we send some mock response from this API? Just in case UI wants to integrate early.

cdap-proto/src/main/java/co/cask/cdap/proto/MetadataSearchResultRecord.java
```diff
@@ -0,0 +1,83 @@
...
+package co.cask.cdap.proto;
+
+import com.google.common.base.Objects;
+
+/**
+ * Represent the Metadata search result record.
+ */
+public class MetadataSearchResultRecord {
+  private final Id.NamespacedId id;
+  private final String type;
```
bdmogal: since the type is provided in the request, should it be part of the response?

henry-cask: Additional information should not hurt in my opinion. Just in case later we allow any type search.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataAdmin.java
henry-cask: Currently, all the types are checked against simple names of the Id.NamespacedId classes, for example Id.Program, or Id.Stream so we will have validator in the service for the string type to make sure it is valid.

One of the reasons I do not want to introduce another enum for types of CDAP component supported because then we need to maintain 2 different source of truth whenever new types of CDAP component is added for Metadata support.

Thoughts?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataHttpHandler.java
henry-cask: I am in the middle on working on real one so hopefully will get something by tomorrow.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/MetadataAdmin.java
poornachandra: I think we need to have another method to search metadata across all target types.

cdap-proto/src/main/java/co/cask/cdap/proto/MetadataSearchResultRecord.java
poornachandra: I think it is better to add an enum for type. String is brittle.

cdap-proto/src/main/java/co/cask/cdap/proto/MetadataSearchResultRecord.java
```diff
@@ -0,0 +1,83 @@
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    MetadataSearchResultRecord that = (MetadataSearchResultRecord) o;
+
+    return Objects.equal(this.id, that.id) &&
```
poornachandra: We are moving to Java 7 Objects for equals and hashcode.

poornachandra:
Also, we will probably need another endpoint to search for tags. Let's discuss this tomorrow.

cdap-proto/src/main/java/co/cask/cdap/proto/MetadataSearchResultRecord.java
henry-cask: I will add Enum for the proto record but translate to class name in the back dataset processing.

cdap-proto/src/main/java/co/cask/cdap/proto/MetadataSearchResultRecord.java
henry-cask: Hmm this is already Java 7 equals mode, isnt it?

henry-cask:
Updated PR to reflect comments.

cdap-proto/src/main/java/co/cask/cdap/proto/MetadataSearchResultRecord.java
chtyim: This one is not. The Java7 one calls `Objects.equals` vs the guava one is `Objects.equal`.

henry-cask: LOL ok, looks like wrong auto import =)

henry-cask:
Build:
http://builds.cask.co/browse/CDAP-DUT2783-3

henry-cask:
This will need to be rebased with PR #3868

PR #3870, opened 2015-09-10
CDAP-3470 Refactored NamespaceHttpHandler to throw exceptions and als…

sagarkapare:
…o made ServiceUnAvailableException unchecked.

JIRA: https://issues.cask.co/browse/CDAP-3470

1. Updated code in the `NamespaceHttpHandler` to throw exceptions which would be caught by `HttpExceptionHandler` in order to return the appropriate ERROR code.
2. Made the `ServiceUnavailableException` unchecked.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
@@ -17,8 +17,7 @@
 package co.cask.cdap.gateway.handlers;
 
 import co.cask.cdap.common.AlreadyExistsException;
-import co.cask.cdap.common.NamespaceCannotBeDeletedException;
-import co.cask.cdap.common.NotFoundException;
+import co.cask.cdap.common.BadRequestException;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.namespace.NamespaceAdmin;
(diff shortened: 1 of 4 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
```diff
@@ -67,64 +66,42 @@ public void getAllNamespaces(HttpRequest request, HttpResponder responder) throw
...
-    } catch (NotFoundException e) {
-      responder.sendString(HttpResponseStatus.NOT_FOUND, String.format("Namespace %s not found", namespaceId));
-    } catch (Exception e) {
-      LOG.error("Failed to read namespace metadata request body.", e);
-      responder.sendString(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage());
-    }
+                                        @PathParam("namespace-id") String namespaceId) throws Exception {
+    NamespaceMeta meta = parseBody(request, NamespaceMeta.class);
+    namespaceAdmin.updateProperties(Id.Namespace.from(namespaceId), meta);
+    responder.sendString(HttpResponseStatus.OK, "Properties updated");
```
bdmogal: nit: Not your PR, but should we send a more meaningful response here: ``Properties for namespace %s updated``?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
```diff
@@ -147,15 +124,13 @@ public void create(HttpRequest request, HttpResponder responder, @PathParam("nam
                            String.format("Namespace '%s' created successfully.", namespaceId));
     } catch (AlreadyExistsException e) {
```
bdmogal: So this is left as is, because we want to keep the API idempotent, right? :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
```diff
@@ -147,15 +124,13 @@ public void create(HttpRequest request, HttpResponder responder, @PathParam("nam
...
     }
   }
 
   @DELETE
   @Path("/unrecoverable/namespaces/{namespace-id}")
-  public void delete(HttpRequest request, HttpResponder responder, @PathParam("namespace-id") String namespace) {
+  public void delete(HttpRequest request, HttpResponder responder, @PathParam("namespace-id") String namespace)
+    throws Exception {
     if (!cConf.getBoolean(Constants.Dangerous.UNRECOVERABLE_RESET, Constants.Dangerous.DEFAULT_UNRECOVERABLE_RESET)) {
       responder.sendString(HttpResponseStatus.FORBIDDEN,
```
bdmogal: not this PR, but maybe ``HttpExceptionHandler`` should handle ``FORBIDDEN`` as well?

cdap-common/src/main/java/co/cask/cdap/common/namespace/AbstractNamespaceClient.java
```diff
@@ -135,7 +135,9 @@ public void deleteDatasets(Id.Namespace namespaceId) throws Exception {
     if (response.getResponseCode() == HttpURLConnection.HTTP_NOT_FOUND) {
       throw new NamespaceNotFoundException(namespaceId);
     } else if (HttpURLConnection.HTTP_FORBIDDEN == response.getResponseCode()) {
-      throw new NamespaceCannotBeDeletedException(namespaceId, response.getResponseBodyAsString());
+      String msg = String.format("Datasets in the namespace '%s' cannot be deleted. ", namespaceId)
```
bdmogal: Maybe add a ``Reason: `` after ``deleted.``. Also, you can reuse the existing ``String.format`` instead of concatenating the ``response.getResponseBodyAsString()``

bdmogal:
@sagarkapare only minor comments. Rest LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
sagarkapare: Yes. Will do that as a separate PR with JIRA: https://issues.cask.co/browse/CDAP-3638

cdap-common/src/main/java/co/cask/cdap/common/namespace/AbstractNamespaceClient.java
sagarkapare: Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/NamespaceHttpHandler.java
sagarkapare: Updated the response.

sagarkapare:
New build is running here - http://builds.cask.co/browse/CDAP-DUT2792-2

bdmogal:
LGTM. Please squash commits before merging.

sagarkapare:
Build passed. Squashed and merging.

PR #3872, opened 2015-09-10
Adds license to side-service.js & modifies pom.xml to exclude node/**…

ajainarayanan:
… under cdap-ui folder

### DIFF
--- a/cdap-ui/pom.xml
+++ b/cdap-ui/pom.xml
@@ -97,6 +97,36 @@
         </executions>
       </plugin>
     </plugins>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.rat</groupId>
+          <artifactId>apache-rat-plugin</artifactId>
+          <version>0.10</version>
+          <executions>
+            <execution>
+              <id>rat-check</id>
+              <phase>validate</phase>
+              <goals>
+                <goal>check</goal>
+              </goals>
+              <configuration>
+                <excludes>
+                  <exclude>**/*.editorconfig</exclude>
+                  <exclude>**/*.jshintignore</exclude>
+                  <exclude>**/*.jshintrc</exclude>
+                  <exclude>**/*.svg</exclude>
+                  <exclude>**/bower_components/**</exclude>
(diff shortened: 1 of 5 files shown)


elfenheart:
JS files :+1: LGTM

pom.xml
```diff
@@ -1297,6 +1297,7 @@
                   <exclude>**/*.svg</exclude>
                   <exclude>**/bower_components/**</exclude>
                   <exclude>**/node_modules/**</exclude>
+                  <exclude>**/node/**</exclude>
```
chtyim: shouldn't be at top level pom. Better have it localized in the cdap-ui module (in fact not just `**/node/**`, but also the bower and node_modules one).

cdap-ui/pom.xml
```diff
@@ -97,6 +97,36 @@
...
+              <configuration>
+                <excludes>
+                  <exclude>**/*.editorconfig</exclude>
+                  <exclude>**/*.jshintignore</exclude>
+                  <exclude>**/*.jshintrc</exclude>
+                  <exclude>**/*.svg</exclude>
+                  <exclude>**/bower_components/**</exclude>
+                  <exclude>**/node_modules/**</exclude>
+                  <exclude>**/node/**</exclude>
+                  <exclude>**/*.json</exclude>
```
awholegunch: Add ``README.*`` to the excludes?

chtyim: Do we have README inside cdap-ui module?

elfenheart: we do have README in cdap-ui folder

bdmogal:
Yes, there is a README. But like other READMEs, I added a license header at the end of it.

cdap-ui/pom.xml
bdmogal: Yes, there is a README. But like other READMEs, I added a license header at the end of it.

bdmogal:
tested locally, I'm going to merge this, since its causing feature branches to fail RAT checks.

anwar6953:
LGTM

PR #3877, opened 2015-09-10
CDAP-3621-throw-IllegalStateException-instead-of-NPE

anwar6953:
If Hadoop Job History Server has permission issues reading some files, the JobClient#getJob(JobId) method returns null.
So, instead of throwing NullPointerException, throw a NotFoundException, which will be caught by DistributedMRJobInfoFetcher#getMRJobInfo and fall back to the localMRJobInfoFetcher.

https://issues.cask.co/browse/CDAP-3621
http://builds.cask.co/browse/CDAP-DUT2797-2

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
@@ -82,6 +82,10 @@ public class MRJobClient implements MRJobInfoFetcher {
     JobStatus thisJob = findJobForRunId(jobs, runId);
 
     RunningJob runningJob = jobClient.getJob(thisJob.getJobID());
+    if (runningJob == null) {
+      throw new IllegalStateException(String.format("JobClient returned null for RunId: '%s', JobId: '%s'",
+                                                    runId, thisJob.getJobID()));
+    }
     Counters counters = runningJob.getCounters();
 
     TaskReport[] mapTaskReports = jobClient.getMapTaskReports(thisJob.getJobID());
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/DistributedMRJobInfoFetcher.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/DistributedMRJobInfoFetcher.java
@@ -16,15 +16,11 @@
 
 package co.cask.cdap.app.mapreduce;
 
-import co.cask.cdap.common.NotFoundException;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.MRJobInfo;
 import com.google.inject.Inject;
 import org.slf4j.Logger;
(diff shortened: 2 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
```diff
@@ -82,6 +82,10 @@ public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundExceptio
     JobStatus thisJob = findJobForRunId(jobs, runId);
 
     RunningJob runningJob = jobClient.getJob(thisJob.getJobID());
+    if (runningJob == null) {
+      LOG.warn("JobClient returned null for JobId: '{}'", thisJob.getJobID());
+      throw new NotFoundException(runId);
```
sagarkapare: This will be thrown when the runid is available but the jobClient could not return the corresponding `RunningJob`. Should it be `IllegalStateException`?

anwar6953: I think the NFE is also appropriate, since the JHS couldn't retrieve the job for the particular runId, but IllegalStateException or IOException also seem to fit.

sagarkapare: But if we `throw new NotFoundException(runId)`, the message that user will get is "runid not found", which would be incorrect since the runid exists in the system. Maybe throw an exception with more accurate message.

anwar6953: ok, updated to throw IllegalStateException with appropriate message. Also changed caller to catch Exception and use backup, instead of only doing it upon IOE and NFE.

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
```diff
@@ -82,6 +82,9 @@ public MRJobInfo getMRJobInfo(Id.Run runId) throws IOException, NotFoundExceptio
     JobStatus thisJob = findJobForRunId(jobs, runId);
 
     RunningJob runningJob = jobClient.getJob(thisJob.getJobID());
+    if (runningJob == null) {
+      throw new IllegalStateException(String.format("JobClient returned null for JobId: %s", thisJob.getJobID()));
```
sagarkapare: Its better to put the runId in the message.

sagarkapare:
Minor comment otherwise LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/app/mapreduce/MRJobClient.java
anwar6953: Ok, I'll put both.

sagarkapare:
LGTM :+1:

PR #3880, opened 2015-09-10
CDAP-3519 Adding more tests to lineage computation

poornachandra:


cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/lineage/LineageServiceTest.java
```diff
@@ -38,73 +39,358 @@
...
+    Lineage oneLevelLineage = lineageService.computeLineage(dataset1, 500, 20000, 1);
+
+    Assert.assertEquals(ImmutableSet.of(program1, program2), oneLevelLineage.getPrograms());
+    Assert.assertEquals(ImmutableSet.of(dataset1, dataset2, dataset3), oneLevelLineage.getData());
 
     Assert.assertEquals(
       ImmutableSet.of(
-        new Relation(datasetInstance2, program2, AccessType.WRITE, toSet(twillRunId(run2)), toSet(flowlet2)),
-        new Relation(datasetInstance3, program2, AccessType.READ, toSet(twillRunId(run2)), toSet(flowlet2))
+        new Relation(dataset2, program1, AccessType.WRITE, toSet(twillRunId(run1)), toSet(flowlet1)),
```
bdmogal: why do we use twill run id here? or is it just the method name? In production it would be the cdap run id right?

Also, should we add a detailed javadoc to the ``Relation`` class? I think it would be helpful to explain what some of the members in that class stand for.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/lineage/LineageServiceTest.java
```diff
@@ -38,73 +39,358 @@
...
+    //       |     |           |
+    //       |     |           |
+    // D0 -->|     ---------------> P4 -> D7
+
+    LineageStore lineageStore = new LineageStore(getTxExecFactory(), dsFrameworkUtil.getFramework(),
+                                                 Id.DatasetInstance.from("default", "testBranchLoopLineage"));
+    LineageService lineageService = new LineageService(lineageStore);
+
+    // Define datasets
+    Id.DatasetInstance dataset0 = Id.DatasetInstance.from("default", "dataset0");
```
bdmogal: can these definitions be made ``private final`` class members?

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/lineage/LineageServiceTest.java
```diff
@@ -38,73 +39,358 @@
...
+    Id.DatasetInstance dataset1 = Id.DatasetInstance.from("default", "dataset1");
+    Id.DatasetInstance dataset2 = Id.DatasetInstance.from("default", "dataset2");
+    Id.DatasetInstance dataset3 = Id.DatasetInstance.from("default", "dataset3");
+    Id.DatasetInstance dataset4 = Id.DatasetInstance.from("default", "dataset4");
+    Id.DatasetInstance dataset5 = Id.DatasetInstance.from("default", "dataset5");
+    Id.DatasetInstance dataset6 = Id.DatasetInstance.from("default", "dataset6");
+    Id.DatasetInstance dataset7 = Id.DatasetInstance.from("default", "dataset7");
+
+    // Define programs and runs
+    Id.Program program1 = Id.Program.from("default", "app1", ProgramType.FLOW, "flow1");
```
bdmogal: ditto. Also, for flows, you can use ``Id.Flow`` for brevity

bdmogal:
@poornachandra minor comments. Also, I don't think we have a case in these tests where the same run both reads and writes to the same dataset? Unless I missed it, can we add that as well?

anew:
one additional test case is a direct cycle: D1 is read by P1 which writes back to D1 in the same run. And compare that to the situation where the cycle is formed by two different runs.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/lineage/LineageServiceTest.java
poornachandra: CDAP uses twill RunId in production too.

poornachandra: Also, I'm making changes to Relation class in a different PR. I'll add javadocs over there.

poornachandra:
@bdmogal @anew Thanks for the review. I have addressed the comments, please take another look.

poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT2789-4

bdmogal:
LGTM

poornachandra:
Squashing commits

poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT2789-5

poornachandra:
Build is green, merging.

PR #3881, opened 2015-09-11
Code refactoring related to stream and explore

chtyim:
- Enabling of explore table on stream is always done by StreamAdmin. Shouldn’t be done in StreamCreator
- Consistently using injection for ExploreNaming instead of direct instantiation
- Fix warnings on Hive ExploreService classes


jwang47:
LGTM

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2801-1

PR #3888, opened 2015-09-11
CDAP-3629 removing adapter client

albertshau:
Removes AdapterClient and related classes. This means adapters
are no longer visible to the unit test and integration test
frameworks.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2808

bdmogal:
My only concern is some non-Beta marked APIs (especially in TestFramework) have been removed. However, given that the main adapter framework itself has been removed, I see no choice. LGTM.

PR #3891, opened 2015-09-11
[CDAP-3613] Split the metadata of an entity into properties and tags …

bdmogal:
…in the Business Metadata Dataset.

Jira: [CDAP-3613](https://issues.cask.co/browse/CDAP-3613)
Build: http://builds.cask.co/browse/CDAP-DUT2811

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
@@ -24,15 +24,18 @@ import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Scanner;
 import co.cask.cdap.data2.dataset2.lib.table.MDSKey;
 import co.cask.cdap.proto.Id;
-
 import co.cask.cdap.proto.MetadataSearchTargetType;
 import co.cask.cdap.proto.ProgramType;
+import com.google.common.base.Joiner;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
-import com.google.common.base.Throwables;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -40,12 +43,34 @@ import java.util.Set;
 import javax.annotation.Nullable;
 
 /**
- * Implementation of Business Metadata on top of {@link IndexedTable}.
+ * Dataset that manages Business Metadata using an {@link IndexedTable}.
  */
 public class BusinessMetadataDataset extends AbstractDataset {
+  private static final String TAGS_KEY = "tags";
+  private static final String TAGS_SEPARATOR = ",";
+
   // column keys
-  public static final String KEYVALUE_COLUMN = "kv";
-  public static final String VALUE_COLUMN = "v";
+  static final String KEYVALUE_COLUMN = "kv";
+  static final String VALUE_COLUMN = "v";
+
+  /**
+   * Identifies the type of metadata - property or tag
+   */
+  private static enum MetadataType {
+    PROPERTY("p"),
+    TAG("t");
+
+    private final String serializedForm;
+
+    private MetadataType(String serializedForm) {
+      this.serializedForm = serializedForm;
+    }
+
+    @Override
+    public String toString() {
+      return serializedForm;
+    }
+  }
 
   public static final String KEYVALUE_SEPARATOR = ":";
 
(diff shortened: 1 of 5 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -129,23 +198,49 @@ public BusinessMetadataRecord getBusinessMetadata(Id.NamespacedId targetId, Stri
   }
 
   /**
+   * Retrieves all the properties for the specified {@link Id.NamespacedId}.
+   *
+   * @param targetId the {@link Id.NamespacedId} for which properties are to be retrieved
+   * @return the properties of the specified {@link Id.NamespacedId}
+   */
+  public Map<String, String> getProperties(Id.NamespacedId targetId) {
```
bdmogal: We should standardize between the return type of ``getProperties(Id.NamespacedId targetId)`` and ``getProperty(Id.NamespacedId targetId, String key)``.

They should either respectively return:
1.  ``Map<String, String>`` and ``Map.Entry<String, String>``; or
2. ``List<BusinessMetadataRecord>`` and ``BusinessMetadataRecord``.

@poornachandra @henry-cask thoughts?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -26,24 +26,50 @@
...
-  public static final String KEYVALUE_COLUMN = "kv";
-  public static final String VALUE_COLUMN = "v";
+  static final String KEYVALUE_COLUMN = "kv";
+  static final String VALUE_COLUMN = "v";
+
+  /**
+   * Identifies the type of metadata - property or tag
+   */
+  private static enum MetadataType {
+    Property("p"),
```
poornachandra: All caps for enum names, right?

henry-cask:
There are some improvements of code that not related to this PR, could you help me pointed out particular changes needed to be reviewed?

bdmogal:
The main change in this PR is, we have a ``MetadataType`` (property or tag) that is used to split metadata in the business metadata dataset. This new type appears in the rowkey just before the metadata key. 

Other changes: ``MetadataAdmin`` methods return a Set rather than Iterable, more extensive tests for ``BusinessMetadataDataset``

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDatasetTest.java
```diff
@@ -43,7 +44,13 @@
   private static final Id.DatasetInstance datasetInstance =
     Id.DatasetInstance.from(DatasetFrameworkTestUtil.NAMESPACE_ID, "meta");
 
-  BusinessMetadataDataset dataset;
+  private BusinessMetadataDataset dataset;
+
+  private final Id.Application app1 = Id.Application.from("ns1", "app1");
+  // Have to use Id.Program here because Id.Flow does not implement equals. It needs to be updated.
```
anwar6953: doesn't Id.Flow inherit the equals method from Id.Program?

bdmogal: This comment isn't entirely correct. The real problem is that we do an ``equals`` comparison on lines 175 and 211. However, the search API that the result is from, actually returns an ``Id.Program``, not specifically an ``Id.Flow``. That's so the actual object is an ``Id.Program``, whereas the expected object If I used ``Id.Flow`` here would be ``Id.Flow``. That's the reason why the comparison fails. So unless the dataset APIs are modified to return the individual program Id classes (not worth the trouble IMO right now), we will have to use ``Id.Program`` here. I'll update the comment.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -341,22 +480,26 @@ private String getTargetType(Id.NamespacedId namespacedId) {
 
   private String getMetadataKey(String type, byte[] rowKey) {
     MDSKey.Splitter keySplitter = new MDSKey(rowKey).split();
-    // The rowkey is [targetType][targetId][key], so skip the first few strings.
+    // The rowkey is [targetType][targetId][metadata-type][key], so skip the first few strings.
```
henry-cask: Need to update this comment and others to make sure it reflect the right MDSKey format.

bdmogal: This is updated right?

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDatasetTest.java
anwar6953: Would it be a good idea to make Id.Program#equals check if the other object is an instance of Id.Program, instead of equals that class? And then the type can be checked also?
Currently, the equals doesn't check type, which is not good.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: I always like abstraction of class over String if possible. This will allow us to add more stuff when searching for metadata.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDatasetTest.java
bdmogal: probably, but I'd refrain from doing it right now. We should report a jira to refactor that entire class - to use consistent methods of object creation, use java.util.Objects for comparison etc.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
bdmogal: Found one more and updated.

henry-cask:
Did a pass and add some comments.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: But we could address it as separate PR

henry-cask:
This LGTM

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
bdmogal: I agree its more flexible. However, we cannot return ``BusinessMetadataRecord`` as is from the REST API. So was wondering that we will have to do a pass over the ``List<BusinessMetadataRecord>`` passed from here, and convert it into a representation suitable for REST somewhere downstream.

Since this is an internal API, I'm going to keep it as is, maybe in future we can return BusinessMetadataRecord, and have a custom serializer to omit the fields that are not required from the REST API.

bdmogal:
Build has been passing. Squashed commits. Merging.

PR #3892, opened 2015-09-11
CDAP-3189 Avoid adding cdap-default.xml multiple times to the configuration.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-3189

In order to pass the cdap configurations (cdap-default.xml and cdap-site.xml) to the mapper and reducer tasks, they are serialized into the hConf under property name "hconf.cconf".

While deserializing, we first create the new `CConfiguration` object which includes the cdap-default.xml and cdap-site.xml. Then to this object we add the value of "hconf.cconf" property as a new resource. This results in the `CConfiguration` object to contain duplicate configurations. 

As a result we see lot of warning messages when the MapReduce program is run:
```java
2015-09-11 09:23:00,407 - WARN  [LocalJobRunner Map Task Executor #5:c.c.c.c.c.Configuration@1748] - java.io.ByteArrayInputStream@6158ab02:an attempt to override final parameter: app.template.plugin.dir;  Ignoring.
2015-09-11 09:23:00,407 - WARN  [LocalJobRunner Map Task Executor #4:c.c.c.c.c.Configuration@1748] - java.io.ByteArrayInputStream@11b48062:an attempt to override final parameter: app.template.plugin.dir;  Ignoring.
2015-09-11 09:23:00,409 - WARN  [LocalJobRunner Map Task Executor #2:c.c.c.c.c.Configuration@1748] - java.io.ByteArrayInputStream@5f8ce5e:an attempt to override final parameter: stream.instance.file.prefix;  Ignoring.
2015-09-11 09:23:00,409 - WARN  [LocalJobRunner Map Task Executor #5:c.c.c.c.c.Configuration@1748] - java.io.ByteArrayInputStream@6158ab02:an attempt to override final parameter: stream.instance.file.prefix;  Ignoring.
``` 

In this PR, while deserializing, we create an empty `CConfiguration` object which will have no resource added to it and then to this object we add the value of "hconf.cconf" property as a new resource.

 Tested on standalone and I no longer see those warning messages for `app.template.plugin.dir` and `stream.instance.file.prefix`

Will run the integration tests on cluster as well.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
@@ -265,7 +265,7 @@ public final class MapReduceContextConfig {
   }
 
   public CConfiguration getConf() {
-    CConfiguration conf = CConfiguration.create();
+    CConfiguration conf = CConfiguration.createEmpty();
     conf.addResource(new ByteArrayInputStream(hConf.get(HCONF_ATTR_CCONF).getBytes()));
     return conf;
   }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
@@ -17,6 +17,7 @@
 package co.cask.cdap.internal.app.runtime.batch;
 
 import co.cask.cdap.api.artifact.Plugin;
+import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.data.batch.SimpleSplit;
 import co.cask.cdap.api.data.batch.Split;
 import co.cask.cdap.api.workflow.WorkflowToken;
@@ -265,9 +266,7 @@ public final class MapReduceContextConfig {
   }
 
   public CConfiguration getConf() {
-    CConfiguration conf = CConfiguration.createEmpty();
-    conf.addResource(new ByteArrayInputStream(hConf.get(HCONF_ATTR_CCONF).getBytes()));
-    return conf;
+    return CConfiguration.create(new ByteArrayInputStream(Bytes.toBytes(hConf.get(HCONF_ATTR_CCONF))));
   }
 
   private void setTx(Transaction tx) {
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
@@ -193,7 +193,7 @@ public abstract class AbstractProgramTwillRunnable<T extends ProgramRunner> impl
 
       UserGroupInformation.setConfiguration(hConf);
 
-      cConf = CConfiguration.create(new File(configs.get("cConf")).toURI().toURL());
+      cConf = CConfiguration.create(new File(configs.get("cConf")));
 
       // Alter the template directory to only the name part in the container directory.
       // It works in pair with the ProgramRunner.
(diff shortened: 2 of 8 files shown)


sreevatsanraman:
Add the reference to the build from the feature branch. Great info in the PR, can this be updated to the JIRA as well.

LGTM. Please merge if the tests on cluster pass.

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT2813-1

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
```diff
@@ -265,7 +265,7 @@ private void setConf(CConfiguration conf) {
   }
 
   public CConfiguration getConf() {
-    CConfiguration conf = CConfiguration.create();
+    CConfiguration conf = CConfiguration.createEmpty();
```
chtyim: There are many places we do similar thing by calling `cConf.clear()` first (search for usage of clear(), you will see a quite a number). It's better to unify to do it in one way. I propose to do it like this:

```java
CConfiguration cConf = CConfiguration.create(resource);

...
// Two new factory method to create with the give resources only
public static CConfiguration create(URI resource, URI...moreResources);
public static CConfiguration create(InputStream resource);
```

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
```diff
@@ -265,7 +265,7 @@ private void setConf(CConfiguration conf) {
...
     conf.addResource(new ByteArrayInputStream(hConf.get(HCONF_ATTR_CCONF).getBytes()));
```
chtyim: The existing code is actually wrong. Should never call `String.getBytes()` without providing encoding. Better use `Bytes.toBytes(hConf.get(HCONF_ATTR_CCONF))`

gsps1: Configuration has a method ``addResource(String)``, can we just use that instead ?

chtyim: You can creates as many static create(X) method :)

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
sagarkapare: Added static methods. Using URL as a parameter for the create to make it in consistent with the `addResource` method parameters. Please let me know if this is ok.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceContextConfig.java
sagarkapare: Fixed.

sagarkapare:
@chtyim Addressed comments. Please take a look again.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
```diff
@@ -193,9 +193,7 @@ public void initialize(TwillContext context) {
 
       UserGroupInformation.setConfiguration(hConf);
 
-      cConf = CConfiguration.create();
-      cConf.clear();
-      cConf.addResource(new File(configs.get("cConf")).toURI().toURL());
+      cConf = CConfiguration.create(new File(configs.get("cConf")).toURI().toURL());
```
chtyim: Maybe having a create method that takes File.

chtyim: more precisely, `create(File file, File...moreFiles)`.

chtyim:
Just one comment left.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractProgramTwillRunnable.java
sagarkapare: done.

sagarkapare:

chtyim:
LGTM

sagarkapare:
The build passed here - http://builds.cask.co/browse/CDAP-DUT2813-3. Also tested on the cluster with the help of integration tests. No log statements  `an attempt to override final parameter: app.template.plugin.dir;  Ignoring.` and `an attempt to override final parameter: stream.instance.file.prefix;  Ignoring.` observed in the master log files and MapReduce completed. Merging.

PR #3893, opened 2015-09-11
[CDAP-3523] [CDAP-3522] Implement REST APIs for searching Metadata

henry-cask:
CDAP-3523 Add new endpoint of REST APIs for searching metadata using value and key=value to metadata handler.

CDAP-3522 Add IndexedTable based search to the Business Metadata dataset.

Add more descriptive name for internal IndexedTable for business metadata dataset.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/DefaultMetadataAdmin.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/DefaultMetadataAdmin.java
@@ -19,13 +19,20 @@ package co.cask.cdap.metadata;
 import co.cask.cdap.common.NamespaceNotFoundException;
 import co.cask.cdap.common.NotFoundException;
 import co.cask.cdap.common.namespace.AbstractNamespaceClient;
-import co.cask.cdap.common.namespace.NamespaceAdmin;
+import co.cask.cdap.data2.metadata.dataset.BusinessMetadataRecord;
 import co.cask.cdap.data2.metadata.service.BusinessMetadataStore;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.MetadataSearchResultRecord;
+import co.cask.cdap.proto.MetadataSearchTargetType;
+
 import com.google.common.base.Throwables;
 import com.google.inject.Inject;
 
+import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nullable;
 
 /**
  * Implementation of {@link MetadataAdmin} that interacts directly with {@link BusinessMetadataStore}
@@ -76,6 +83,27 @@ public class DefaultMetadataAdmin implements MetadataAdmin {
   public void removeTags(Id.NamespacedId entityId, String... tags) throws NotFoundException {
     ensureEntityExists(entityId);
     businessMds.removeTags(entityId, tags);
+  }
+
+  @Override
+  public Set<MetadataSearchResultRecord> searchMetadata(String searchQuery,
+                                                        @Nullable MetadataSearchTargetType type)
+    throws NotFoundException {
+    Iterable<BusinessMetadataRecord> results;
+    if (type == null) {
+      results = businessMds.searchMetadata(searchQuery);
+    } else {
+      results = businessMds.searchMetadataOnType(searchQuery, type);
+    }
+
+    Set<MetadataSearchResultRecord> searchResultRecords = new LinkedHashSet<>();
+    for (BusinessMetadataRecord bmr : results) {
+      MetadataSearchResultRecord msr =
+        new MetadataSearchResultRecord(bmr.getTargetId(),
+                                       (type == null ? MetadataSearchTargetType.ALL : type));
+      searchResultRecords.add(msr);
+    }
+    return searchResultRecords;
   }
 
   /**
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
@@ -19,14 +19,18 @@ package co.cask.cdap.metadata;
 import co.cask.cdap.common.BadRequestException;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.MetadataSearchResultRecord;
+import co.cask.cdap.proto.MetadataSearchTargetType;
 import co.cask.cdap.proto.ProgramType;
 import co.cask.http.AbstractHttpHandler;
 import co.cask.http.HttpResponder;
+
 import com.google.common.base.Charsets;
 import com.google.common.collect.Sets;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import com.google.inject.Inject;
+
 import org.jboss.netty.buffer.ChannelBuffer;
 import org.jboss.netty.buffer.ChannelBufferInputStream;
 import org.jboss.netty.handler.codec.http.HttpRequest;
@@ -38,11 +42,13 @@ import java.io.Reader;
 import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
 
 /**
  * HttpHandler for Metadata
@@ -188,8 +194,8 @@ public class MetadataHttpHandler extends AbstractHttpHandler {
   @POST
   @Path("/namespaces/{namespace-id}/apps/{app-id}/tags")
   public void addAppTags(HttpRequest request, HttpResponder responder,
-                            @PathParam("namespace-id") String namespaceId,
-                            @PathParam("app-id") String appId) throws Exception {
+                         @PathParam("namespace-id") String namespaceId,
+                         @PathParam("app-id") String appId) throws Exception {
     Id.Application app = Id.Application.from(namespaceId, appId);
     metadataAdmin.addTags(app, readArray(request));
     responder.sendString(HttpResponseStatus.OK,
@@ -343,5 +349,18 @@ public class MetadataHttpHandler extends AbstractHttpHandler {
   @Path("/metadata/history")
   public void recordRun(HttpRequest request, HttpResponder responder) {
     responder.sendString(HttpResponseStatus.OK, "Metadata recorded successfully");
+  }
+
+  // *** Search endpoints ***
+
+  @GET
+  @Path("/namespaces/{namespace-id}/metadata/search")
+  public void searchMetadata(HttpRequest request, HttpResponder responder,
+                             @PathParam("namespace-id") String namespaceId,
+                             @QueryParam("query") String searchQuery,
+                             @QueryParam("target") MetadataSearchTargetType target) throws Exception {
+    Set<MetadataSearchResultRecord> results = metadataAdmin.searchMetadata(searchQuery, target);
+
+    responder.sendJson(HttpResponseStatus.OK, results);
   }
 }
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
@@ -25,14 +25,16 @@ import co.cask.cdap.api.dataset.table.Scanner;
 import co.cask.cdap.data2.dataset2.lib.table.MDSKey;
 import co.cask.cdap.proto.Id;
 
+import co.cask.cdap.proto.MetadataSearchTargetType;
 import co.cask.cdap.proto.ProgramType;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.common.base.Throwables;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 
 import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import javax.annotation.Nullable;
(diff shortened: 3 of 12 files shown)


henry-cask:
Build:
http://builds.cask.co/browse/CDAP-DUT2812-1

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
```diff
@@ -344,4 +349,17 @@ public void removeStreamTags(HttpRequest request, HttpResponder responder,
...
+
+  @GET
+  @Path("/namespaces/{namespace-id}/metadata/search")
+  public void searchMetadata(HttpRequest request, HttpResponder responder,
+                             @PathParam("namespace-id") String namespaceId,
+                             @QueryParam("query") String searchQuery,
+                             @QueryParam("target") MetadataSearchTargetType target) throws Exception {
+    Iterable<MetadataSearchResultRecord> results = metadataAdmin.searchMetadata(searchQuery, target);
+
+    responder.sendJson(HttpResponseStatus.OK, Sets.newHashSet(results));
```
bdmogal: Should we just return a ``Set<MetadataSearchResultRecord>`` from ``metadataAdmin.searchMetadata(searchQuery, target)``? I made the change for the other methods in https://github.com/caskdata/cdap/pull/3891

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -167,6 +169,34 @@ public Void apply(BusinessMdsIterable input) throws Exception {
...
+   */
+  public Iterable<BusinessMetadataRecord> searchMetadataOnType(final String searchQuery,
+                                                        final MetadataSearchTargetType type) {
+    return txnl.executeUnchecked(new TransactionExecutor.Function<BusinessMdsIterable,
+      Iterable<BusinessMetadataRecord>> () {
+      @Override
+      public Iterable<BusinessMetadataRecord> apply(BusinessMdsIterable input) throws Exception {
+        // Currently we support two types of search formats: value and key=value.
+        // Check for exiatance of '=' char to make sure we did search in the right indexed column.
+        if (searchQuery.contains("=")) {
```
bdmogal: will ``=`` conflict with query params in REST API? If yes, should we use ``:`` instead?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -187,13 +189,56 @@ public void removeMetadata(Id.NamespacedId targetId, Predicate<String> filter) {
...
+   * @return The {@link Iterable} of {@link BusinessMetadataRecord} that fit the key value pair.
+   */
+  public List<BusinessMetadataRecord> findBusinessMetadataOnKeyValue(String keyValue, MetadataSearchTargetType type) {
+    return executeSearchOnColumns(BusinessMetadataDataset.KEYVALUE_COLUMN, keyValue, type);
+  }
+
+  // Helper method to execute IndexedTable search on target index column.
+  List<BusinessMetadataRecord> executeSearchOnColumns(String column, String searchValue,
+                                                      MetadataSearchTargetType type) {
+    Scanner scanner = indexedTable.readByIndex(Bytes.toBytes(column), Bytes.toBytes(searchValue));
```
bdmogal: Should ``KEYVALUE_COLUMN`` and ``VALUE_COLUMN`` be bytes, instead of doing this conversion every single time?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
chtyim: Why a Set anyway? The underlying call from the Dataset actually returns a List.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -187,13 +189,56 @@ public void removeMetadata(Id.NamespacedId targetId, Predicate<String> filter) {
...
+    List<BusinessMetadataRecord> results = new LinkedList<>();
+    try {
```
chtyim: move try one line up. It's a better practice to have try right after the resource creation. You can either have the List moved to the top or have the return statement moves inside the try.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
chtyim: There shouldn't be conflict since query params need to be URL encoded.

```
path?p=key%3Dvalue
```

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -187,13 +189,56 @@ public void removeMetadata(Id.NamespacedId targetId, Predicate<String> filter) {
...
+   * @param keyValue The metadata value to be found.
+   * @param type The target type of objects to search from.
+   * @return The {@link Iterable} of {@link BusinessMetadataRecord} that fit the key value pair.
+   */
+  public List<BusinessMetadataRecord> findBusinessMetadataOnKeyValue(String keyValue, MetadataSearchTargetType type) {
+    return executeSearchOnColumns(BusinessMetadataDataset.KEYVALUE_COLUMN, keyValue, type);
+  }
+
+  // Helper method to execute IndexedTable search on target index column.
+  List<BusinessMetadataRecord> executeSearchOnColumns(String column, String searchValue,
```
bdmogal: this can be private

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
henry-cask: Set since the returned data is the distinct objects inverted index by the value or key=value.
The dataset return list of actual records coming which could have duplicate target objects.

For example. If you search for value of "blah" and someone had 2 key-value pairs like key1=blah and key2=blah then the Dataset will return 2 records.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
henry-cask: Yeah, it will be decoded when come to server.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: Yes we could just use bytes instead.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDatasetModule.java
```diff
@@ -31,6 +31,6 @@ public void register(DatasetDefinitionRegistry registry) {
     DatasetDefinition<IndexedTable, ? extends DatasetAdmin> indexedTableDef =
       registry.get(IndexedTable.class.getName());
     registry.add(new BusinessMetadataDefinition("businessMetadataDataset", indexedTableDef));
-    registry.add(new BusinessMetadataDefinition(BusinessMetadataDataset.class.getSimpleName(), indexedTableDef));
+    registry.add(new BusinessMetadataDefinition(BusinessMetadataDataset.class.getName(), indexedTableDef));
```
bdmogal: why this change? is this a standard we use?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
bdmogal: :+1:

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -167,6 +169,34 @@ public Void apply(BusinessMdsIterable input) throws Exception {
...
+   */
+  public Iterable<BusinessMetadataRecord> searchMetadata(final String searchQuery) {
+    return searchMetadataOnType(searchQuery, MetadataSearchTargetType.ALL);
+  }
+
+  /**
+   * Search to the underlying Business Metadata Dataset for a target type.
+   */
+  public Iterable<BusinessMetadataRecord> searchMetadataOnType(final String searchQuery,
+                                                        final MetadataSearchTargetType type) {
```
bdmogal: nit: indentation

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
bdmogal: Not a big deal, but we had used ``Set`` because the results from all these APIs are collections with unique elements.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDatasetModule.java
henry-cask: According to @poornachandra yes. We try to use getName instead.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
henry-cask: Got it. Thanks!

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: Actually lets keep it as String for now. It is used in the Row.getString and registration of BusinessMetadataDefinition. 
We could change it later.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: Will do.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
henry-cask: Sure. Set is fine too

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
bdmogal: hmmm, perhaps then create another constant - ``VALUE_COLUMN_BYTES`` and ``KEYVALUE_COLUMN_BYTES``, so we don't need conversion multiple times on every call?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
henry-cask: I set it as package private for now just in case we want to access it to do unit tests.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
henry-cask: But the design doc said we use key:value format so let me change it to that for now.

henry-cask:
Succes build: 
http://builds.cask.co/browse/CDAP-DUT2812-2

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/DefaultMetadataAdmin.java
```diff
@@ -78,6 +85,33 @@ public void removeTags(Id.NamespacedId entityId, String... tags) throws NotFound
...
+    Set<MetadataSearchResultRecord> searchResultRecords = new LinkedHashSet<>();
+    Iterator<BusinessMetadataRecord> iterator = results.iterator();
+    while (iterator.hasNext()) {
+      BusinessMetadataRecord bmr = iterator.next();
+      MetadataSearchResultRecord msr =
+        new MetadataSearchResultRecord(bmr.getTargetId(),
+                                       (type == null ? MetadataSearchTargetType.ALL : type));
+
+      // Check if already added
+      if (!searchResultRecords.contains(msr)) {
```
bdmogal: since this is a set, this check is not required

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/DefaultMetadataAdmin.java
```diff
@@ -78,6 +85,33 @@ public void removeTags(Id.NamespacedId entityId, String... tags) throws NotFound
...
+    throws NotFoundException {
+    Iterable<BusinessMetadataRecord> results;
+    if (type == null) {
+      results = businessMds.searchMetadata(searchQuery);
+    } else {
+      results = businessMds.searchMetadataOnType(searchQuery, type);
+    }
+
+    Set<MetadataSearchResultRecord> searchResultRecords = new LinkedHashSet<>();
+    Iterator<BusinessMetadataRecord> iterator = results.iterator();
```
bdmogal: you should be able to use
```
for (BusinessMetadataRecord result : results) 
```
here?

bdmogal:
@henry-cask just two minor comments. Not necessary to run build again. If you think they should be fixed, fix, squash and merge. LGTM

henry-cask:
Thanks! Rebase, squash, and merging ...

PR #3894, opened 2015-09-11
CDAP-3219 Allow setting HBASE_VERSION

wolf31o2:
Fixes CDAP-3219 by allowing a user to set HBASE_VERSION and CDAP respecting the setting.


dereklwood:
:+1:

PR #3895, opened 2015-09-11
CDAP-3639: Moving ArtifactDescriptor to app-fabric

gokulavasan:
PluginSelector should have access only to ArtifactIds. Moving ArtifactDescriptor to cdap-app-fabric.

JIRA : https://issues.cask.co/browse/CDAP-3639

Build : http://builds.cask.co/browse/CDAP-DUT2817


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -235,7 +237,23 @@ public ArtifactDetail getArtifact(Id.Artifact artifactId) throws IOException, Ar
...
+    SortedMap<ArtifactId, PluginClass> artifactIds = Maps.newTreeMap();
+    for (Map.Entry<ArtifactDescriptor, PluginClass> pluginClassEntry : pluginClasses.entrySet()) {
+      artifactIds.put(pluginClassEntry.getKey().getArtifactId(), pluginClassEntry.getValue());
+    }
+    Map.Entry<ArtifactId, PluginClass> chosenArtifact = selector.select(artifactIds);
+    if (chosenArtifact == null) {
+      throw new PluginNotExistsException(artifactId, pluginType, pluginName);
+    }
+
+    for (Map.Entry<ArtifactDescriptor, PluginClass> pluginClassEntry : pluginClasses.entrySet()) {
```
albertshau: we can't just call get on the map? ArtifactDescriptor doesn't use location in its equality and comparison implementations right?

albertshau: er nevermind, was confused, keys are different

albertshau: i guess we could do a get with a null location... not sure how we could simplify this

albertshau:
would be nice to avoid all the copying, though not sure how we would do that. lgtm

gokulavasan:

PR #3897, opened 2015-09-11
CDAP-3626 Disallow writing to named output, when there is only one ou…

anwar6953:
Disallow writing to named output, when there is only one output configured.
This restriction prevents an issue when having multiple instances of a RecordWriter (and hence multiple close() calls to those RecordWriters) is problematic -- for instance, in the case of AvroKeyOutputFormat.

https://issues.cask.co/browse/CDAP-3628
http://builds.cask.co/browse/CDAP-DUT2818-2


anwar6953:
This simply enforces the restriction documented in commit:
https://github.com/caskdata/cdap/commit/621ff2bcf7943d477638149ac808ff5d5349ce63
(disallows writing using the write(String,K,V) method when there is only a single output).

albertshau:
:+1:

anwar6953:
@albertshau Updated the PR to not use MultipleOutputsMainOutputWrapper when there is only one OutputFormat, since we can set it directly on the job.

albertshau:
:+1:

PR #3898, opened 2015-09-11
Fix Documentation Build

awholegunch:
Add missing SnapshotAvro and SnapshotParquet documents to the Sphinx index.

Adds a test to prevent confusing error messages when running the build in a manual sub-directory.


wolf31o2:
:+1: LGTM

PR #3899, opened 2015-09-11
Clean up scripts, simplify coding, use more bash internals

wolf31o2:
This is mostly a style/syntax update. There's not much refactoring done and little functionality change. It's mostly the following:

- Adding braces to variables
- Simplify setting variables
- Use Bash variable expansion capabilities
- Reduce shell-outs (cat/grep)
- Functions `return` instead of `exit`
- Use `$( )` rather than backticks

### DIFF
--- a/cdap-common/bin/service
+++ b/cdap-common/bin/service
@@ -30,7 +30,7 @@ SVCNAME=${0##*/}
 
 if [[ "${SVCNAME}" == "service" ]]; then
   echo "Create a soft link to this script as svc-<component>"
-  echo "E.g. ln -s service svc-<component>"
+  echo "Example: ln -s service svc-<component>"
   exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


cdap-common/bin/service
```diff
@@ -15,30 +15,31 @@
...
-APP=`echo $SVCNAME | cut -d'-' -f2-`; export APP
+APP=${SVCNAME##svc-}
+export APP
 
 # Determine symlink target (full path)
-TARGET=`readlink $0`
+TARGET=$(readlink ${0}) # TODO: readlink isn't always available
 if [[ $? -ne 0 ]]; then
   echo "Create a soft link to this script as svc-<component>"
   echo "E.g. ln -s service svc-<component>"
```
awholegunch: Common: change 'E.g.' to 'Example:' 
````
echo "Example: ln -s service svc-<component>"

sreevatsanraman:
lgtm

PR #3900, opened 2015-09-11
[CDAP-3179] upgrade tool to suport hbase upgrade

anew:
This adds a new option to the upgrade tool to force upgrading the coprocessors of all HBase tables after an upgrade of the underlying HBase or Hadoop distro. 

Because of the way, the dataset framework implements upgrade(), we cannot pass arguments to the upgrade method. This should be changed in the future. For now, I worked around this by setting a system property that indicates to the HBase admin class that the coprocessors must upgraded regardless of whether a change in the table or the CDAP version is detected.  

Also changes the way we generate the coprocessor jar file name to include the current CDAP version - that makes it easier to detect discrepancies when debugging problems. 

Also removes the upgrade step for the usage registry  - it was only needed between 3.0 and 3.1


cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -107,14 +109,16 @@
...
-              "  5. Application Specifications\n" +
+              "  2. Application Specifications\n" +
+              "  3. Removes all Adapters\n" +
+              "  4. Application Specifications\n" +
               "      - Adds artifacts for existing applications\n" +
               "      - Updates application metadata to include newly added artifact\n" +
               "      - Deletes all ApplicationTemplates\n" +
               "  Note: Once you run the upgrade tool you cannot rollback to the previous version."),
+    UPGRADE_HBASE("After an HBase upgrade, updates the coprocessor jars of all user and \n" +
+                    "system HBase tables to a version that is compatible with the new HBase \n " +
```
awholegunch: Extra trailing space:
```
HBase \n" +

File: cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java [hunk 2]
```diff
@@ -107,14 +109,16 @@
...
               "  1. User and System Datasets (upgrades the coprocessor jars)\n" +
-              "  2. UsageRegistry Dataset Type\n" +
-              "  3. Application Specifications\n" +
-              "  4. Removes all Adapters\n" +
-              "  5. Application Specifications\n" +
+              "  2. Application Specifications\n" +
+              "  3. Removes all Adapters\n" +
+              "  4. Application Specifications\n" +
               "      - Adds artifacts for existing applications\n" +
               "      - Updates application metadata to include newly added artifact\n" +
```
awholegunch: Should this be "to include newly added artifacts"? (Sounds like it should be plural)

anew: Note that I did not change this. But actually I think it is correct.. every application's meta data will have a reference to one artifact.

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
anew: fixed

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
awholegunch: No, I realized you didn't change it, but I figured these are always a good chance to review things...the change depends on how precise we'd like to be. Given that there is one artifact referenced in each application metadata, it could be: 
"Updates each application's metadata to include the newly added artifact", to make clear the relationship...

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
```diff
@@ -365,6 +387,19 @@ private void performUpgrade() throws Exception {
     applicationLifecycleService.upgrade(true);
   }
 
+  private void performHBaseUpgrade() throws Exception {
+
+    System.setProperty(AbstractHBaseDataSetAdmin.SYSTEM_PROPERTY_FORCE_HBASE_UPGRADE, Boolean.TRUE.toString());
```
sreevatsanraman: This code is shared between HbaseUpgrade and regular upgrade. Maybe we can call the `performHBaseUpgrade()` from `performUpgrade()`. Otherwise the logic is in two places.

sreevatsanraman:
One minor comment. We need to add corresponding docs for this as well change the upgrade procedure when it involves underlying Hbase upgrade.

cdap-master/src/main/java/co/cask/cdap/data/tools/UpgradeTool.java
anew: we can't do that because performHBaseUpgrade sets a system property. But we can factor it out into a method. I did not think that was necessary because it is really small but will do.

anew:
I addressed the comments, and I adapted the section in the docs that documents CDH upgrade.

anew:
I addressed the last comment. Keeping this open for @awholegunch to take a look. Please let me know when I can merge.

sreevatsanraman:
LGTM :+1:

anew:
Merging this now. @awholegunch if you have docs concerns, I will fix them in a new PR

PR #3901, opened 2015-09-11
CDAP-3465 Workflow statistics

awholegunch:
This documents the three new handlers added that provide Workflow Statistics.

Fix for https://issues.cask.co/browse/CDAP-3465

Pages of interest:

- [Index](http://builds.cask.co/artifact/CDAP-DDBD570/shared/build-1/Docs-HTML/3.2.0-SNAPSHOT/en/reference-manual/http-restful-api/index.html)
- [Workflow RESTful API](http://builds.cask.co/artifact/CDAP-DDBD570/shared/build-1/Docs-HTML/3.2.0-SNAPSHOT/en/reference-manual/http-restful-api/workflow.html)


cdap-docs/reference-manual/source/http-restful-api/workflow.rst
```diff
@@ -133,4 +135,307 @@ Examples
...
+   * - Parameter
+     - Description
+   * - ``<namespace>``
+     - Namespace ID
+   * - ``<app-id>``
+     - Name of the application
+   * - ``<workflow-id>``
+     - Name of the workflow
+   * - ``<start-time>``
+     - Start time of runs; default is ``now`` (optional)
```
sreevatsanraman: We should specify the unit - seconds? Milli-seconds?

sreevatsanraman: Looking at the code we can also pass timestamp and if we do it should be in seconds
https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/WorkflowStatsSLAHttpHandler.java#L91-L100

sreevatsanraman: other options are `now-<n>s` `now-<n>m` `now-<n>h`

awholegunch: Added units and a note explaining how to use them.

cdap-docs/reference-manual/source/http-restful-api/workflow.rst
```diff
@@ -11,10 +11,12 @@ Workflow HTTP RESTful API
 
 .. highlight:: console
 
+Workflow Tokens
+===============
+
```
anwar6953: nit: why a space here, but not beneath "Workflow Statistics" on line 140?

sreevatsanraman:
LGTM

cdap-docs/reference-manual/source/http-restful-api/workflow.rst
awholegunch: Short answer: It doesn't actually matter, as the formatted HTML is controlled by the CSS of the theme we specify and use with Sphinx. 

Long answer: JJ (me) is inconsistent with his use of spacing, and should (and will) try to do a better job in the future. I will add something to our rst style guide about this. I think an appropriate approach would be before every major title (those are the first ones, usually indicated with "=" underlining), that there be two spaces before. (An exception could possibly be made if there already something that has the same affect, such as an anchor or highlight.)  And one space before the next title level (usually indicated with "-" underlining). After the underlining, text could be tight, which would keep the file size (as rendered in an editor) shorter and quicker to navigate. I will adjust this file to match.

PR #3902, opened 2015-09-12
moves adapter modal functionality to node config tab

astewes:
*  Moves adapter modal functionality to Hydrator Node Config tab


ajainarayanan:
:+1: LGTM

PR #3903, opened 2015-09-12
Route /ping to metadata service until we figure out a better way of m…

bdmogal:
…onitoring the service.

Build: http://builds.cask.co/browse/CDAP-DUT2814-3


bdmogal:
Going to close this to re-think the approach. Will get back to it later.

PR #3906, opened 2015-09-12
CDAP-3637 removing ApplicationTemplate and Adapter from api

albertshau:
This change removes ApplicationTemplate and Adapter related
classes from the cdap api. It also cleans up a lot of the code
that became unused due to the removal of the classes. This results
in a LOT of code removed.


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/DynamicMapReduceContext.java
```diff
@@ -82,12 +81,10 @@ public DynamicMapReduceContext(Program program,
                                  TransactionSystemClient txClient,
                                  DatasetFramework dsFramework,
                                  LocationFactory locationFactory,
-                                 @Nullable AdapterDefinition adapterSpec,
-                                 @Nullable PluginInstantiator pluginInstantiator,
                                  @Nullable PluginInstantiator artifactPluginInstantiator) {
     super(program, type, runId, taskId, runtimeArguments, Collections.<String>emptySet(), spec,
           logicalStartTime, programNameInWorkflow, workflowToken, discoveryServiceClient, metricsCollectionService,
-          dsFramework, locationFactory, adapterSpec, pluginInstantiator, artifactPluginInstantiator);
+          dsFramework, locationFactory, artifactPluginInstantiator);
```
gokulavasan: can be renamed to pluginInstantiator?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceClassLoader.java
```diff
@@ -101,11 +98,8 @@ public MapReduceClassLoader(ClassLoader programClassLoader) {
   public MapReduceClassLoader(ClassLoader programClassLoader,
                               Configuration hConf,
                               Map<String, Plugin> plugins,
-                              @Nullable AdapterDefinition adapterSpec,
-                              @Nullable PluginInstantiator pluginInstantiator,
                               @Nullable PluginInstantiator artifactPluginInstantiator) {
-    this(new Parameters(programClassLoader, hConf, plugins, adapterSpec,
-                        pluginInstantiator, artifactPluginInstantiator));
+    this(new Parameters(programClassLoader, hConf, plugins, artifactPluginInstantiator));
```
gokulavasan: same here

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/DynamicMapReduceContext.java
gokulavasan: or was it to reduce conflicts?

gokulavasan: i guess we can make that change in another PR later

gokulavasan:
LGTM :+1:

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/DynamicMapReduceContext.java
albertshau: renamed in some places, but missed this one.  Will rename it

PR #3908, opened 2015-09-12
CDAP-3521 Recording lineage information for a dataset accessed by a flowlet

poornachandra:
Only recording lineage for datasets from flowlets in this PR to keep changes minimal. I'll add recording for streams and other program types in an upcoming PR.

Also, the following TODOs will be fixed in an upcoming PR:
1. Avoid duplicate writes of lineage for a dataset instance.
2. Add complete metadata and tags when recording lineage.

The PR looks big due to refactoring of tests and new classes for serializing lineage over HTTP.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/serialize/RelationProto.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/serialize/RelationProto.java
@@ -14,52 +14,27 @@
  * the License.
  */
 
-package co.cask.cdap.data2.metadata.lineage;
-
-import co.cask.cdap.proto.Id;
-import com.google.common.collect.ImmutableSet;
-import org.apache.twill.api.RunId;
+package co.cask.cdap.metadata.serialize;
 
 import java.util.Objects;
 import java.util.Set;
 
 /**
- * Represents a Dataset access by a Program.
+ * Class to serialize {@link co.cask.cdap.data2.metadata.lineage.Relation}.
  */
-public class Relation {
-  private final Id.NamespacedId data;
-  private final Id.Program program;
-  private final AccessType access;
-  private final Set<RunId> runs;
-  private final Set<Id.NamespacedId> components;
+public class RelationProto {
+  private final String data;
+  private final String program;
+  private final String access;
+  private final Set<String> runs;
+  private final Set<String> components;
 
-  public Relation(Id.DatasetInstance data, Id.Program program, AccessType access, Set<RunId> runs,
-                  Set<? extends Id.NamespacedId> components) {
+  public RelationProto(String data, String program, String access, Set<String> runs, Set<String> components) {
     this.data = data;
     this.program = program;
     this.access = access;
-    this.runs = ImmutableSet.copyOf(runs);
-    this.components = ImmutableSet.copyOf(components);
-  }
-
-  public Id.NamespacedId getData() {
-    return data;
-  }
-
-  public Id.Program getProgram() {
-    return program;
-  }
-
-  public AccessType getAccess() {
-    return access;
-  }
-
-  public Set<RunId> getRuns() {
-    return runs;
-  }
-
-  public Set<Id.NamespacedId> getComponents() {
-    return components;
+    this.runs = runs;
+    this.components = components;
   }
 
   @Override
@@ -70,7 +45,7 @@ public class Relation {
     if (o == null || getClass() != o.getClass()) {
       return false;
     }
-    Relation relation = (Relation) o;
+    RelationProto relation = (RelationProto) o;
     return Objects.equals(data, relation.data) &&
       Objects.equals(program, relation.program) &&
       Objects.equals(access, relation.access) &&
@@ -85,10 +60,10 @@ public class Relation {
 
   @Override
   public String toString() {
-    return "Relation{" +
-      "data=" + data +
-      ", program=" + program +
-      ", access=" + access +
+    return "RelationProto{" +
+      "data='" + data + '\'' +
+      ", program='" + program + '\'' +
+      ", access='" + access + '\'' +
       ", runs=" + runs +
       ", components=" + components +
       '}';
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
@@ -23,15 +23,21 @@ import co.cask.cdap.data2.dataset2.DatasetDefinitionRegistryFactory;
 import co.cask.cdap.data2.dataset2.DatasetFramework;
 import co.cask.cdap.data2.dataset2.DefaultDatasetDefinitionRegistry;
 import co.cask.cdap.data2.dataset2.InMemoryDatasetFramework;
+import co.cask.cdap.data2.metadata.writer.BasicLineageWriter;
+import co.cask.cdap.data2.metadata.writer.LineageWriter;
+import co.cask.cdap.data2.metadata.writer.LineageWriterDatasetFramework;
 import com.google.inject.Module;
 import com.google.inject.PrivateModule;
 import com.google.inject.Scopes;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
+import com.google.inject.name.Names;
 
 /**
  * DataSets framework bindings
  */
 public class DataSetsModules extends RuntimeModule {
+
+  public static final String BASIC_DATASET_FRAMEWORK = "basicDatasetFramework";
 
   @Override
   public Module getInMemoryModules() {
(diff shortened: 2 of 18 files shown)


poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT2810-8

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/BasicLineageWriter.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public class BasicLineageWriter implements LineageWriter {
+  private static final Logger LOG = LoggerFactory.getLogger(BasicLineageWriter.class);
+  private static final Gson GSON = new Gson();
+
+  private final BusinessMetadataStore businessMetadataStore;
+  private final LineageStore lineageStore;
+
+  @Inject
+  public BasicLineageWriter(BusinessMetadataStore businessMetadataStore, LineageStore lineageStore) {
```
bdmogal: package private

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/BasicLineageWriter.java
```diff
@@ -0,0 +1,66 @@
...
+                        Id.NamespacedId component) {
+    // TODO: avoid duplicate writes for a dataset instance
+    String metadata = gatherMetadata(run, datasetInstance);
+    LOG.debug("Writing access for run {}, dataset {}, accessType {}, component {}, metadata = {}",
+              run, datasetInstance, accessType, component, metadata);
+    lineageStore.addAccess(run, datasetInstance, accessType, metadata, component);
+  }
+
+  private String gatherMetadata(Id.Run run, Id.DatasetInstance datasetInstance) {
+    // TODO: add app metadata and tags
```
bdmogal: this is available, right? just that you need to create an object to represent all this info? Makes sense to do in another PR

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/LineageWriterDatasetFramework.java
```diff
@@ -0,0 +1,206 @@
...
+ */
+public class LineageWriterDatasetFramework implements DatasetFramework, ProgramContextAware {
+  private final DatasetFramework delegate;
+  private final LineageWriter lineageWriter;
+
+  private final AtomicReference<Id.Run> runRef = new AtomicReference<>();
+  private final AtomicReference<Id.NamespacedId> componentIdRef = new AtomicReference<>();
+
+  @Inject
+  public LineageWriterDatasetFramework(
```
bdmogal: package private

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/serialize/RelationProto.java
```diff
@@ -0,0 +1,71 @@
...
+
+package co.cask.cdap.metadata.serialize;
+
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * Class to serialize {@link co.cask.cdap.data2.metadata.lineage.Relation}.
+ */
+public class RelationProto {
```
bdmogal: should these classes move to ``cdap-proto``?

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
```diff
@@ -23,16 +23,22 @@
...
 import com.google.inject.Scopes;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
+import com.google.inject.name.Names;
 
 /**
  * DataSets framework bindings
  */
 public class DataSetsModules extends RuntimeModule {
 
+  public static final String BASIC_DATASET_FRAMEWORK = "basicDatasetFramework";
```
bdmogal: should we call this ``delegateDatasetFramework`` to indicate that the ``LineageDatasetFramework`` will delegate to these classes?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/lineage/Lineage.java
```diff
@@ -27,27 +26,15 @@
  */
 public class Lineage {
   private final Set<Relation> relations;
-  private final Set<Id.Program> programs;
```
bdmogal: how come these two were removed? Just the fact that they can be derived from the set of relations? I think they provided a nice utility, instead of users having to compute this themselves?

bdmogal:
@poornachandra the approach makes sense to me and is easy to understand. Only major comments  about moving the proto classes to ``cdap-proto``, and the removal of programs and data from the ``Lineage`` class.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/serialize/RelationProto.java
poornachandra: Since these classes are only used for Lineage now, let's leave it here. It should be easy to move them when required, right?

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
poornachandra: `basicDatasetFramework` is used by classes other than `LineageDatasetFramework`, for them delegate does not add any meaning I think.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/lineage/Lineage.java
poornachandra: Right now they are not being used anywhere. Having them only increased the size of code, hence removed them. It should be easy to add them back if required later.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/BasicLineageWriter.java
poornachandra: Yes.

poornachandra:
@bdmogal Thanks for the review. Addressed your comments. Can you please take another look?

Also, can I make the package-private changes in the upcoming PR?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/serialize/RelationProto.java
bdmogal: ``LineageProto`` is the output of the lineage REST API, though, right? Its ok to do it in another PR, but we should do it in 3.2?

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
bdmogal: Ok, maybe I missed it. Will the ``DatasetFramework`` named as ``basicDatasetFramework`` be used by anything other than the ``LineageDatasetFramework``?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/lineage/Lineage.java
bdmogal: Sounds good.

bdmogal:
Ok to make changes in another PR since multiple PRs depend on each other here. LGTM.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
poornachandra: Yes, it will be used by classes that `LineageDatasetFramework` depends on, such as `BusinessMetadataStore` and `LineageStore` to prevent circular dependency.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/serialize/RelationProto.java
poornachandra: If all classes used as output of REST API have to go to cdap-proto then I'll move them.

bdmogal: yes, that's the convention. It's ok to create a move-only PR after this one though.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
bdmogal: Ok clearly I missed that. **basic** is reasonable.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/serialize/RelationProto.java
poornachandra: Will do.

poornachandra:
Build passed, merging.

PR #3910, opened 2015-09-12
CDAP-3647 fix bug to allow deletion of system artifacts

albertshau:
This fixes a bug where system artifacts could never be deleted,
which is problematic if one is somehow deployed incorrectly.


gokulavasan:
LGTM :+1:

albertshau:
http://builds.cask.co/browse/CDAP-DUT2831

PR #3912, opened 2015-09-12
Adding record for metadata changes. Also moved metadata proto classes…

bdmogal:
… under their own namespace


bdmogal:
Just import modifications, and added two unused classes to share between @poornachandra and me. Ensured that checkstyle passes - http://builds.cask.co/browse/CDAP-DRC2662-3
Not going to run a full build.

cdap-proto/src/main/java/co/cask/cdap/proto/metadata/MetadataChangeRecord.java
```diff
@@ -0,0 +1,89 @@
...
+  @Nullable
+  public String getUpdater() {
+    return updater;
+  }
+
+  /**
+   * Represents the changes between the previous and the new record
+   */
+  public static final class MetadataDiffRecord {
+    private final MetadataRecord additions;
```
poornachandra: Will an update be a deletion and an addition?

bdmogal: Yes. I think we should leave it at delete + addition, primarily because tags do not have updates, only properties do.

poornachandra: Sounds good.

poornachandra:
LGTM

bdmogal:
Squashed commits. Checkstyle passes. Also ran metadata tests locally to be certain. Merging.

PR #3916, opened 2015-09-13
CDAP-3521 Record lineage for all program types

poornachandra:


poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT2838-4

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
```diff
@@ -70,10 +70,6 @@
   private static final String OTHER_NAME = "test1";
   private static final Gson GSON = new Gson();
 
-  private HttpResponse createNamespace(String id) throws Exception {
-    return doPut(String.format("%s/namespaces/%s", Constants.Gateway.API_VERSION_3, id));
-  }
-
   private HttpResponse createNamespace(String metadata, String id) throws Exception {
```
bdmogal: nit: perhaps this should move too, to be consistent

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/LineageTest.java
```diff
@@ -77,13 +87,120 @@ public void testLineage() throws Exception {
...
+    } finally {
+      try {
+        deleteApp(app, 200);
+      } catch (Throwable e) {
+        LOG.error("Got exception while deleting app", e);
+      }
+    }
+  }
+
+  private RunId runAndWait(Id.Program program) throws Exception {
```
bdmogal: all the methods from here on seem pretty useful, should we move them to ``AppFabricTestBase``?

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/LineageTest.java
```diff
@@ -77,13 +87,120 @@ public void testLineage() throws Exception {
...
+                                          }
+                                        }));
+        return Iterables.size(runRecords.get());
+      }
+    }, 60, TimeUnit.SECONDS, 10, TimeUnit.MILLISECONDS);
+    Assert.assertEquals(1, Iterables.size(runRecords.get()));
+    return RunIds.fromString(Iterables.getFirst(runRecords.get(), null).getPid());
+  }
+
+  private int status(org.apache.http.HttpResponse response) {
```
bdmogal: Hmmm, I think I had used common-http's Http client in MetadataTestBase because at the time the org.apache one didn't allow to send request body in a delete call. Can you report a jira for me to standardize on Http client in these classes? Seems weird to have to  use two different ones.

cdap-app-fabric/src/test/java/co/cask/cdap/AllProgramsApp.java
```diff
@@ -245,17 +268,27 @@ public void run(DatasetContext context) throws Exception {
...
       @Override
       protected void configure() {
         useDatasets(DATASET_NAME);
       }
+
+      @Path(ENDPOINT)
+      @GET
+      public void handler(HttpServiceRequest request, HttpServiceResponder responder) {
+        LOG.info("Endpoint {} called in service {}", ENDPOINT, NAME);
+        KeyValueTable table = getContext().getDataset(DATASET_NAME);
```
bdmogal: should we add a different dataset here and assert that the service accesses a different dataset than the other programs? Would that make the test even more compelling?

bdmogal:
@poornachandra just a few minor comments, and a question about the unit test. Other than that, LGTM.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/NamespaceHttpHandlerTest.java
poornachandra: okay

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/LineageTest.java
poornachandra: okay

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/LineageTest.java
poornachandra: JIRA https://issues.cask.co/browse/CDAP-3651. I would prefer co.cask.common HTTP client in both. But that might involve a lot of changes.

bdmogal: Reported https://issues.cask.co/browse/CDAP-3652

cdap-app-fabric/src/test/java/co/cask/cdap/AllProgramsApp.java
poornachandra: We would then have to use a different dataset for every program. I don't think having a single dataset should matter much, because the access is recorded while the dataset is being retrieved. If an incorrect dataset gets recorded, then the same incorrect dataset would get returned and all tests would fail. What do you say?

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/LineageTest.java
bdmogal: oops. conflict. Will mark mine as duplicate.

cdap-app-fabric/src/test/java/co/cask/cdap/AllProgramsApp.java
bdmogal: I think that's fair.

bdmogal:
LGTM

poornachandra:
Squashing commits

poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT2838-5

poornachandra:
Build is green.

PR #3917, opened 2015-09-13
Added a way to serialize/deserialize Id.NamespacedId objects by seria…

bdmogal:
…lizing a 'type' key for use during deserialization. Currently only supports subclasses of Id.NamespacedId that are supported in the metadata system.

This will be used at multiple places in the metadata system.

Build: http://builds.cask.co/browse/CDAP-DUT2839


poornachandra:
LGTM

bdmogal:
Build Green. Merging.

PR #3920, opened 2015-09-13
[CDAP-3518] Added a MetadataChangePublisher to publish metadata chang…

bdmogal:
…es to external systems. Also added a Kafka based implementation for this interface that publishes to a Kafka topic.

Jira: [CDAP-3518](https://issues.cask.co/browse/CDAP-3518)
Build: http://builds.cask.co/browse/CDAP-DUT2834

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.metadata;
+
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.common.conf.CConfiguration;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.codec.NamespacedIdCodec;
+import co.cask.cdap.proto.metadata.MetadataChangeRecord;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.inject.Inject;
+import org.apache.twill.kafka.client.Compression;
+import org.apache.twill.kafka.client.KafkaClient;
+import org.apache.twill.kafka.client.KafkaPublisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.ByteBuffer;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * A {@link MetadataChangePublisher} that publishes metadata changes to an Apache Kafka topic determined by the
+ * configuration parameter <code>metadata.updates.kafka.topic</code>. External systems can get notifications of
+ * metadata changes by subscribing to this topic.
+ */
+public class KafkaMetadataChangePublisher implements MetadataChangePublisher {
+  private static final Logger LOG = LoggerFactory.getLogger(KafkaMetadataChangePublisher.class);
+  private static final Gson GSON = new GsonBuilder()
+    .registerTypeAdapter(Id.NamespacedId.class, new NamespacedIdCodec())
+    .create();
+  private static final String PUBLISHER_CACHE_KEY = "kafkaPublisher";
+  private final KafkaPublisher.Ack ack;
+  // This cache will always only contain a single entry, since we will always use the same key.
+  // However, using it will take care of synchronization issues while creating a publisher.
+  private final LoadingCache<String, KafkaPublisher> publishers;
+  private final String topic;
+
+  @Inject
+  KafkaMetadataChangePublisher(CConfiguration cConf, final KafkaClient kafkaClient) {
+    this.ack = KafkaPublisher.Ack.FIRE_AND_FORGET;
+    this.topic = cConf.get(Constants.Metadata.METADATA_UPDATES_KAFKA_TOPIC);
+    this.publishers = CacheBuilder.newBuilder().build(new CacheLoader<String, KafkaPublisher>() {
+      @Override
+      @SuppressWarnings("NullableProblems")
+      public KafkaPublisher load(String key) throws Exception {
+        return kafkaClient.getPublisher(ack, Compression.SNAPPY);
+      }
+    });
+  }
+
+  @Override
+  public void publish(MetadataChangeRecord changeRecord) {
+    KafkaPublisher publisher;
+    try {
+      publisher = publishers.get(PUBLISHER_CACHE_KEY);
+    } catch (ExecutionException e) {
+      LOG.warn("Unable to get kafka publisher, will not be able to publish metadata.");
+      return;
+    }
+    byte[] changesToPublish = Bytes.toBytes(GSON.toJson(changeRecord));
+    KafkaPublisher.Preparer preparer = publisher.prepare(topic);
+    preparer.add(ByteBuffer.wrap(changesToPublish), changeRecord.getPrevious().getTargetId().hashCode());
+    preparer.send();
+  }
+}
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataServiceModule.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataServiceModule.java
@@ -17,25 +17,81 @@
 package co.cask.cdap.metadata;
 
 import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.runtime.RuntimeModule;
 import co.cask.cdap.gateway.handlers.CommonHandlers;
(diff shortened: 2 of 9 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
```diff
@@ -0,0 +1,89 @@
...
+  @Nullable
+  private KafkaPublisher getOrCreatePublisher() {
+    KafkaPublisher publisher = publisherRef.get();
+    if (publisher != null) {
+      return publisher;
+    }
+    try {
+      publisher = kafkaClient.getPublisher(ack, Compression.SNAPPY);
+    } catch (IllegalStateException e) {
+      // can happen if there are no kafka brokers because the kafka server is down.
```
poornachandra: It would be good idea to log the exception at debug level.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Override
+  public void publish(MetadataChangeRecord changeRecord) {
+    KafkaPublisher publisher = getOrCreatePublisher();
+    if (publisher == null) {
+      LOG.warn("Unable to get kafka publisher, will not be able to publish metadata.");
+      return;
+    }
+    byte[] changesToPublish = Bytes.toBytes(GSON.toJson(changeRecord));
+    String topic = cConf.get(Constants.Metadata.METADATA_UPDATES_KAFKA_TOPIC);
```
poornachandra: This can move as a final field, since this won't change after program start.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataServiceModule.java
```diff
@@ -37,5 +51,18 @@ protected void configure() {
     handlerBinder.addBinding().to(LineageHandler.class);
 
     bind(MetadataAdmin.class).to(DefaultMetadataAdmin.class);
+    if (changePublishingRequired) {
```
poornachandra: I think this should be restructured a bit. There are two different cases here -
   1. Users do not want to publish metadata (I think you are planning on handling that in a later PR).
   2. No metadata publishing for in-memory and standalone mode.

For point 1, we need to have a config parameter that will be used to control the publishing.

For point 2, we can directly bind `MetadataChangePublisher` to `NoOpMetadataChangePublisher` in `getSingleNodeModules()` and `getInMemoryModules()`, after extending  `MetadataServiceModule` from `RuntimeModule`

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/KafkaMetadataChangePublisherTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+        @Override
+        public void finished() {
+          stopLatch.countDown();
+        }
+      }
+    );
+    Assert.assertTrue(latch.await(15, TimeUnit.SECONDS));
+    cancellable.cancel();
+    Assert.assertTrue(stopLatch.await(1, TimeUnit.SECONDS));
```
poornachandra: It might be a good idea to increase this time to 15 seconds too.

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/KafkaMetadataChangePublisherTest.java
```diff
@@ -0,0 +1,231 @@
...
+    Assert.assertTrue(expected.containsAll(actual));
+    Assert.assertTrue(actual.containsAll(expected));
```
poornachandra: The above two assertions can be replaced by `Assert.assertEquals(expected, actual)`, right?

poornachandra:
@bdmogal Minor comments, please fix and merge.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
```diff
@@ -0,0 +1,89 @@
...
+    KafkaPublisher.Preparer preparer = publisher.prepare(topic);
+    preparer.add(ByteBuffer.wrap(changesToPublish), changeRecord.hashCode());
```
poornachandra: All changes to an entity should go to the same partition, hence we need to use entityId's hashcode here.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override
+  public void publish(MetadataChangeRecord changeRecord) {
+    KafkaPublisher publisher = getOrCreatePublisher();
+    if (publisher == null) {
+      LOG.warn("Unable to get kafka publisher, will not be able to publish metadata.");
+      return;
+    }
+    byte[] changesToPublish = Bytes.toBytes(GSON.toJson(changeRecord));
+    KafkaPublisher.Preparer preparer = publisher.prepare(topic);
+    preparer.add(ByteBuffer.wrap(changesToPublish), changeRecord.getPrevious().getTargetId().hashCode());
```
poornachandra: This assumes that `changeRecord.getPrevious()` can never be null, is that true?

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/KafkaMetadataChangePublisherTest.java
```diff
@@ -0,0 +1,230 @@
...
+      publisher.publish(metadataChangeRecord);
+    }
+    assertPublishedMetadata(metadataChangeRecords);
+  }
+
+  private List<MetadataChangeRecord> generateMetadataChanges() {
+    long currentTime = System.currentTimeMillis();
+    ImmutableList.Builder<MetadataChangeRecord> changesBuilder = ImmutableList.builder();
+    Id.DatasetInstance dataset = Id.DatasetInstance.from("ns1", "ds1");
+    // Initial state: contains one property and one tag
```
poornachandra: btw - I think initial state is not getting published here.

poornachandra:
@bdmogal Two more comments to address.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Nullable
+  private KafkaPublisher getOrCreatePublisher() {
+    KafkaPublisher publisher = publisherRef.get();
+    if (publisher != null) {
+      return publisher;
+    }
+    try {
+      publisher = kafkaClient.getPublisher(ack, Compression.SNAPPY);
```
chtyim: So there could be race that two publishers are being created. Is it acceptable?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataServiceModule.java
```diff
@@ -17,25 +17,70 @@
...
+  public Module getStandaloneModules() {
+    return new InMemoryMetadataModule();
+  }
+
+  @Override
+  public Module getDistributedModules() {
+    return new DistributedMetadataModule();
+  }
+
+  private static final class CommonMetadataModule extends AbstractModule {
```
chtyim: consider using PrivateModule and only expose the `MetadataAdmin`.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataServiceModule.java
```diff
@@ -17,25 +17,70 @@
...
+      handlerBinder.addBinding().to(LineageHandler.class);
+
+      bind(MetadataAdmin.class).to(DefaultMetadataAdmin.class);
+    }
+  }
+
+  private static final class DistributedMetadataModule extends AbstractModule {
+    @Override
+    protected void configure() {
+      install(new CommonMetadataModule());
```
chtyim: Better not to use install. You can extend from the CommonModule and implements a abstract method defined in CommonModule.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
bdmogal: @poornachandra yes, ``changeRecord.getPrevious()`` can never be null. Its contained properties or tags may be empty, but the ``MetadataChangeRecord`` object that it returns will never be run. I have added a verification in the latest commit to illustrate this. Please take another look.

poornachandra:
@bdmogal Changes for my comments look good.

bdmogal:
@chtyim addressed comments. Please take a look again.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/KafkaMetadataChangePublisher.java
```diff
@@ -0,0 +1,84 @@
...
+ * A {@link MetadataChangePublisher} that publishes metadata changes to an Apache Kafka topic determined by the
+ * configuration parameter <code>metadata.updates.kafka.topic</code>. External systems can get notifications of
+ * metadata changes by subscribing to this topic.
+ */
+public class KafkaMetadataChangePublisher implements MetadataChangePublisher {
+  private static final Logger LOG = LoggerFactory.getLogger(KafkaMetadataChangePublisher.class);
+  private static final Gson GSON = new GsonBuilder()
+    .registerTypeAdapter(Id.NamespacedId.class, new NamespacedIdCodec())
+    .create();
+  private static final String CONSTANT_CACHE_KEY = "ignore";
```
poornachandra: This could be called - `PUBLISHER_CACHE_KEY = "kafkaPublisher"`, or something meaningful.

poornachandra:
@bdmogal One minor comment, otherwise LGTM.

poornachandra:
LGTM

bdmogal:
Build green - http://builds.cask.co/browse/CDAP-DUT2834-7.

Squashed commits. Merging.

PR #3921, opened 2015-09-13
Feature/ui hydrator static layout

astewes:
*  Changes Hydrator layout from absolute positioning to static + floats
*  Adds dynamic height to bottom panel
*  Adds set height to groups in left side panel
*  Grid and bottom panel now auto-adjust to occupy extra space when left panel is collapsed


ajainarayanan:
:+1: LGTM

PR #3922, opened 2015-09-14
Added APIs to get all metadata and remove all metadata for an entity.…

bdmogal:
… Also cleaned up tests where necessary.

Build: http://builds.cask.co/browse/CDAP-DUT2844

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
@@ -26,6 +26,7 @@ import co.cask.cdap.data2.dataset2.tx.Transactional;
 import co.cask.cdap.data2.metadata.dataset.BusinessMetadataDataset;
 import co.cask.cdap.data2.metadata.dataset.BusinessMetadataRecord;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.metadata.MetadataRecord;
 import co.cask.cdap.proto.metadata.MetadataSearchTargetType;
 import co.cask.tephra.TransactionExecutor;
 import co.cask.tephra.TransactionExecutorFactory;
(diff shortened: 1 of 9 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -97,6 +99,21 @@ public Void apply(BusinessMdsIterable input) throws Exception {
   }
 
   /**
+   * @return a {@link MetadataRecord} representing all the metadata (including properties and tags) for the specified
+   * {@link Id.NamespacedId}.
+   */
+  public Set<MetadataRecord> getMetadata(final Id.NamespacedId entityId) {
```
poornachandra: `BusinessMetadataStore.getMetadata(final Id.NamespacedId entityId)` still needs to return just single `MetadataRecord` object.

`DefaultMetadataAdmin..getMetadata(final Id.NamespacedId entityId)` can then make a call to metadata stores for all scopes and return a set of `MetadataRecord`, right?

bdmogal: Depends on whether we should do both in a single transaction. Is your point that it would be too expensive to club together into one? If so, I agree, I'll make the change you suggested.

bdmogal: Ignore the previous comment. The set is to contain both user and system level ``MetadataRecord``. ``BusinessMetadataStore`` should only return user scope metadata. ``DefaultMetadataAdmin`` needs possibly make calls to multiple stores in future to get both user and system metadata.

anwar6953:
LGTM

bdmogal:
Build is green. Ran all metadata tests and checkstyle after minor last commit about changing a return type. Squashed commits. Merging.

PR #3923, opened 2015-09-14
Fix Doc Build: example includes

awholegunch:
Removes duplicate messages in the warnings in the doc build log.
Fixes the included files, both for line numbers and MD5 hashes.

Built successfully: http://builds.cask.co/browse/CDAP-DDBD571-1


sreevatsanraman:
LGTM

PR #3924, opened 2015-09-14
CDAP-3521 Handle duplicate access record updates

poornachandra:


poornachandra:
Build - http://builds.cask.co/browse/CDAP-DUT2843-3

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/BasicLineageWriter.java
```diff
@@ -78,4 +89,43 @@ private String gatherMetadata(Id.Run run, Id.NamespacedId id) {
     Map<String, String> dataMetadata = businessMetadataStore.getProperties(id);
     return GSON.toJson(dataMetadata);
   }
+
+  private boolean alreadyRegistered(Id.Run run, Id.NamespacedId data, AccessType accessType,
+                                    @Nullable Id.NamespacedId component) {
+    return registered.putIfAbsent(new CheckKey(run, data, accessType, component), true) != null;
+  }
+
+  private static final class CheckKey {
```
bdmogal: Is this more like ``DataAccessIdentifier``? ``CheckKey`` is a little vague?

bdmogal:
@poornachandra code LGTM. Should we add a test?

poornachandra:
Filed a JIRA for adding test - https://issues.cask.co/browse/CDAP-3655

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/BasicLineageWriter.java
poornachandra: At this stage, since changing name will require another build. And this is a private class, I'll leave it as-is for now. We can change this when we touch this class again.

bdmogal:
LGTM

poornachandra:
Build is green - http://builds.cask.co/browse/CDAP-DUT2843-4

PR #3927, opened 2015-09-14
CDAP-3605 Intercept entitiy deletion to also remove metadata.

henry-cask:
Intercept the entity deletion of App, Program, Dataset, and Stream so it also remove the corresponding business metadata (property and tags).

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
@@ -45,6 +45,7 @@ import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.io.Locations;
 import co.cask.cdap.common.namespace.NamespacedLocationFactory;
 import co.cask.cdap.config.PreferencesStore;
+import co.cask.cdap.data2.metadata.service.BusinessMetadataStore;
 import co.cask.cdap.data2.registry.UsageRegistry;
 import co.cask.cdap.data2.transaction.queue.QueueAdmin;
 import co.cask.cdap.data2.transaction.stream.StreamConsumerFactory;
@@ -85,6 +86,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -126,6 +128,7 @@ public class ApplicationLifecycleService extends AbstractIdleService {
   private final MetricStore metricStore;
   private final ArtifactRepository artifactRepository;
   private final ManagerFactory<AppDeploymentInfo, ApplicationWithPrograms> managerFactory;
+  private final BusinessMetadataStore businessMds;
 
   @Inject
   public ApplicationLifecycleService(ProgramRuntimeService runtimeService, Store store, CConfiguration configuration,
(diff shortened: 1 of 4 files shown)


henry-cask:
Build:
http://builds.cask.co/browse/CDAP-DUT2851-2

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -688,6 +693,8 @@ void deleteApp(Id.Application appId, ApplicationSpecification spec) throws Excep
 
     store.removeApplication(appId);
 
+    deleteAppBusinessMetadata(appId);
```
sreevatsanraman: this should be before we remove app from MDS.

sreevatsanraman:
No test case for this PR. Don't merge without adding one.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -695,6 +702,65 @@ void deleteApp(Id.Application appId, ApplicationSpecification spec) throws Excep
...
+    for (ProgramType programType : ProgramType.values()) {
+      deleteProgramBusinessMetadataForApp(programType, appId, appSpec);
+    }
+  }
+
+  /**
+   *  Delete the business metadata for a program type in an application.
+   *  TODO: Need to remove this we support prefix search of metadata type.
+   */
+  private void deleteProgramBusinessMetadataForApp(ProgramType programType, Id.Application appId,
```
sreevatsanraman: This code is brittle. Everytime we add a new program type, we need to remember to add the program type here to delete. Why not delete the Business metadata based on a prefix key for app.

sreevatsanraman:
This PR doesn't take into account when we remove programs on re-deploying an app. Look at `DeletedProgramHandlerStage`

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
sreevatsanraman: nvm i see a comment about this. Let's do this in 3.2

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -695,6 +702,65 @@ void deleteApp(Id.Application appId, ApplicationSpecification spec) throws Excep
...
+        LOG.info("Unable to remove business metadata for unsupported program type: " + programType.name());
+        break;
+    }
+
+    if (programNames == null) {
+      LOG.debug("No program metadata removed for application {}", appId.toString());
+      return;
+    }
+
+    for (String programName : programNames) {
```
anew: Doesn't this also require notifications of the change? How are the notifications sent? Perhaps all meta data notifications should go through the metadata service?

henry-cask:
Break lots of unit tests. Need to revisit this.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
henry-cask: Are we doing similar things for programs related changes? This is not actually change the app metastore but separate dataset that store metadata for programs.

PR #3929, opened 2015-09-14
CDAP-3594, CDAP-3604 : Removing Location from AppSpec, Localizing Plugin JARs in Distributed Mode

gokulavasan:
JIRAs : 
https://issues.cask.co/browse/CDAP-3594
https://issues.cask.co/browse/CDAP-3604

Build : 
http://builds.cask.co/browse/CDAP-DUT2793

Summary of changes:

i) Removed the use of location of plugin JARs in AppSpec (instead following the specific dir structure and propagating the artifact file names [since they have UUID component and thus not predictable] in ProgramRuntimeService).

ii) The directory structure of plugins ( /data/namespaces/<ns>/artifacts/<artifactname>/<artifactjar>  ) is localized in distributed mode. In standalone/ide, plugins are not copied and the local data directory is used. 

iii) PluginInstantiaor that uses direct location to create ClassLoader is required in couple of scenarios:

A) When the inspection of artifacts are done during deployment time (since the files won't be available in the data/namespaces/.... directory yet)
B) During application configuration in distributed mode, the files are not localized in the master node and thus we need the actual location to create instances


gokulavasan:
@albertshau Please review. Thanks!

cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactId.java
```diff
@@ -49,11 +49,7 @@ public ArtifactScope getScope() {
 
   @Override
   public String toString() {
-    return "ArtifactId{" +
-      "name='" + name + '\'' +
-      ", version=" + version +
-      ", scope='" + scope + '\'' +
-      '}';
+    return String.format("%s-%s-%s", name, version, scope);
```
albertshau: scope:name-version to be more consistent with how we've displayed it elsewhere

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -71,12 +83,47 @@ public RuntimeInfo run(Program program, ProgramOptions options) {
...
   /**
+   * Return the copy of the {@link ProgramOptions} including locations of plugin artifacts in it.
+   * @param options the {@link ProgramOptions} in which the locations of plugin artifacts needs to be included
+   * @param namespace namespace of the Program
+   * @param appSpec program's Application Specification
+   * @return the copy of the program options with locations of plugin artifacts included in them
+   */
+  private ProgramOptions addPluginLocations(ProgramOptions options, Id.Namespace namespace,
+                                            ApplicationSpecification appSpec) {
+    if (appSpec == null) {
```
albertshau: can this ever happen?  If so lets annotate with nullable and add a comment for when it can happen

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -71,12 +83,47 @@ public RuntimeInfo run(Program program, ProgramOptions options) {
...
+        Plugin plugin = pluginEntry.getValue();
+        Map.Entry<ArtifactDescriptor, PluginClass> artifactEntry = artifactRepository.getPlugin(
+          appArtifactId, plugin.getPluginClass().getType(), plugin.getPluginClass().getName(), plugin.getArtifactId());
+        artifactLocations.put(plugin.getArtifactId().toString(), artifactEntry.getKey().getLocation().getName());
+      }
+      builder.put(ProgramOptionConstants.PLUGIN_FILENAMES, GSON.toJson(artifactLocations));
+      return new SimpleProgramOptions(options.getName(), new BasicArguments(builder.build()),
+                                      options.getUserArguments(), options.isDebug());
+    } catch (PluginNotExistsException | IOException e) {
+      return options;
```
albertshau: seems like this should throw an exception.  Otherwise the user will see some weird classloading error at runtime and have no clue what happened.

albertshau: though I'm also confused why we need to add the filenames as an option.  Shouldn't we be able to tell the exact path based on the plugin artifact's id?  In fact it is a bug to use the actual filename here since it opens this up to race conditions.  What if the filename changes between now and whenever you pull the files down from hdfs?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java
```diff
@@ -114,6 +142,12 @@ public ClassLoader getArtifactClassLoader(ArtifactDescriptor artifactDescriptor)
     return (Class<T>) getArtifactClassLoader(artifactDescriptor).loadClass(pluginClass.getClassName());
   }
 
+  @SuppressWarnings("unchecked")
+  public <T> Class<T> loadClass(Location location,
```
albertshau: there should only be one way to load a class.  Do we still need the one that takes an ArtifactDescriptor?  Seems like we could use Location everywhere instead

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java
```diff
@@ -63,17 +65,26 @@
...
   private final File tmpDir;
+  private final File pluginDir;
   private final ClassLoader parentClassLoader;
 
   public PluginInstantiator(CConfiguration cConf, ClassLoader parentClassLoader) {
     this.instantiatorFactory = new InstantiatorFactory(false);
-
     File tmpDir = new File(cConf.get(Constants.CFG_LOCAL_DATA_DIR),
       cConf.get(Constants.AppFabric.TEMP_DIR)).getAbsoluteFile();
+    File prefix = cConf.get(Constants.AppFabric.MRTASK_PLUGIN_DIR) != null ?
```
albertshau: MRTASK_PLUGIN_DIR is a bad name since it's not always in mapreduce.  What about something like LOCAL_PLUGIN_DIR

albertshau: also, this setting should always have some value so we don't need the null check. we should be sure to set it in cdap-default if we are not.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
gokulavasan: this was needed since the ``AbstractProgramRuntimeServiceTest`` was generating null for appSpec.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceClassLoader.java
```diff
@@ -232,26 +234,21 @@ private static PluginInstantiator createArtifactPluginInstantiator(MapReduceCont
...
-        LocationFactory hdfsLocationFactory = new HDFSLocationFactory(hConf);
-
-        // Need appropriate LocationFactory since we only the Location URI from Plugin
-        locationFactory = (MapReduceTaskContextProvider.isLocal(hConf)) ? localLocationFactory : hdfsLocationFactory;
         List<ClassLoader> pluginClassLoaders = Lists.newArrayList();
-        for (Plugin plugin : plugins.values()) {
-          ArtifactDescriptor artifactDescriptor = new ArtifactDescriptor(plugin.getArtifactId(),
-            locationFactory.create(
-              plugin.getLocationURI()));
+        for (Map.Entry<String, Plugin> pluginEntry : plugins.entrySet()) {
```
albertshau: the old `for (Plugin plugin : plugins.values())` is more succinct since you aren't using the key

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/MapReduceClassLoader.java
```diff
@@ -96,10 +96,11 @@ public MapReduceClassLoader(ClassLoader programClassLoader) {
    * followed by plugin Export-Package ClassLoader and with the system ClassLoader last.
    */
   public MapReduceClassLoader(ClassLoader programClassLoader,
-                              Configuration hConf,
+                              String namespace,
```
albertshau: better to pass around Id.Namespace

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java
gokulavasan: This is used only for Mapper/Reducer tasks. Since in those tasks, the base directory for the PluginInstantiator will be the JAR name. That is, instead of data/namespaces/default/artifacts..... it will be artifacts.jar/data/namespaces/default/.... Hence we need to append the base directory name only for Mapper/Reducer. Hence the constant name.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java
```diff
@@ -211,8 +266,10 @@ public void close() throws IOException {
     @Override
     public ClassLoader load(ArtifactDescriptor key) throws Exception {
       File unpackedDir = DirUtils.createTempDir(tmpDir);
-      BundleJarUtil.unpackProgramJar(key.getLocation(), unpackedDir);
-
+      String artifactFile = String.format("%s/artifacts/%s/%s", key.getArtifact().getNamespace().getId(),
```
albertshau: ok wasn't clear at first that this is meant to be the same path as the actual file for local mode.  This is brittle and will break in a confusing way if somebody changes the implementation of the write method in ArtifactStore.  They should share the same code.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java
gokulavasan: We still use ``Id.Artifact`` in ``ArtifactDescriptor`` to construct the path to the artifact jar.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java
gokulavasan: true, the code needs to be shared among ArtifactStore, PluginInstantiator and MapReduceRuntimeService.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java
albertshau: that's mapreduce specific logic though.  In that case it's better to pass in the directory rather than look up a conf setting.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -223,6 +223,18 @@ public ArtifactDetail getArtifact(Id.Artifact artifactId) throws IOException, Ar
     return artifactStore.getPluginClasses(artifactId, pluginType, pluginName);
   }
 
+  public Map.Entry<ArtifactDescriptor, PluginClass> getPlugin(Id.Artifact artifactId, String pluginType,
```
albertshau: if you just need the location of the plugin's artifact, use getArtifact() rather than adding this method.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -71,12 +84,53 @@ public RuntimeInfo run(Program program, ProgramOptions options) {
...
+        artifactEntry = artifactRepository.getPlugin(
+          appArtifactId, plugin.getPluginClass().getType(), plugin.getPluginClass().getName(), plugin.getArtifactId());
+      } catch (IOException e) {
+        throw Throwables.propagate(e);
+      } catch (PluginNotExistsException e) {
+        throw new IllegalArgumentException(String.format("Plugin of type %s, name %s could not be found",
+                                                         plugin.getPluginClass().getType(),
+                                                         plugin.getPluginClass().getName()), e);
+      }
+      artifactLocations.put(plugin.getArtifactId().toString(), artifactEntry.getKey().getLocation().getName());
```
chtyim: Move this into the try block. I like statement closer to the context. Also, the declaration of the artifactEntry can be moved inside the try block as well.

sreevatsanraman:
Closing this re-open against release/3.2

PR #3918, opened 2015-09-13
CDAP-3521 CDAP-3632 Recording stream lineage information

poornachandra:

PR #3930, opened 2015-09-14
[CDAP-3613] Publish metadata when it is updated

bdmogal:
Build: http://builds.cask.co/browse/CDAP-DUT2850


cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -1558,4 +1558,13 @@
     <description>Kafka topic to which metadata update notifications are published</description>
   </property>
 
+  <property>
+    <name>metadata.updates.publish.enabled</name>
+    <value>true</value>
+    <description>
+      Determines if metadata updates will be published to an external system (like Apache Kafka). External systems can
+      subscribe to recieve notifications of metadata updates.
```
anew: receive

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -1558,4 +1558,13 @@
     <description>Kafka topic to which metadata update notifications are published</description>
   </property>
 
+  <property>
+    <name>metadata.updates.publish.enabled</name>
+    <value>true</value>
+    <description>
+      Determines if metadata updates will be published to an external system (like Apache Kafka). External systems can
```
anew: like Kafka? I thought is it always and only Kafka?

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
```diff
@@ -41,6 +44,10 @@
 
   @Override
   public Module getInMemoryModules() {
+    return getInMemoryModules(false);
+  }
+
+  public Module getInMemoryModules(final boolean kafkaRequired) {
```
anew: seems like the boolean should be named publishRequired. Whether Kafka is required is a consequence of that.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
```diff
@@ -79,9 +92,10 @@ protected void configure() {
         expose(LineageWriter.class);
         bind(DatasetFramework.class).to(LineageWriterDatasetFramework.class);
         expose(DatasetFramework.class);
+        bind(MetadataChangePublisher.class).to(NoOpMetadataChangePublisher.class);
```
anew: why no notification support in Standalone?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -230,6 +336,12 @@ public Void apply(BusinessMdsIterable input) throws Exception {
     });
   }
 
+  private void publish(MetadataRecord previous, MetadataRecord additions, MetadataRecord deletions) {
```
anew: this may be called with null, if publish is not enabled. You need to protect against that. Also annotate with @Nullable

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -48,11 +56,17 @@
 
   private static final Id.DatasetInstance BUSINESS_METADATA_INSTANCE_ID =
     Id.DatasetInstance.from(Id.Namespace.SYSTEM, "business.metadata");
+  private static final Map<String, String> EMPTY_PROPERTIES = ImmutableMap.of();
+  private static final Set<String> EMPTY_TAGS = ImmutableSet.of();
+
   private final Transactional<BusinessMdsIterable, BusinessMetadataDataset> txnl;
+  private final boolean publishEnabled;
+  private final MetadataChangePublisher changePublisher;
```
anew: There is an implicit assumption here that publishEnabled is false iff changePublisher is NoOpChangePublisher. If that does not hold, this class might crash all over the place.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
```diff
@@ -57,6 +64,12 @@ protected void configure() {
         expose(LineageWriter.class);
         bind(DatasetFramework.class).to(LineageWriterDatasetFramework.class);
         expose(DatasetFramework.class);
+        if (kafkaRequired) {
```
anew: why is this in the datasets module? The publishing is a function of the metadata service, not the data set framework.

anew:
I did one pass and I have a few comments.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
bdmogal: Its required here because the ``LineageWriterDatasetFramework`` needs it (it binds a ``BasicLineageWriter``, which needs ``BusinessMetadataStore``, which needs this).

However, we discussed at https://github.com/caskdata/cdap/pull/3918#discussion_r39352536 that all of these need to be refactored into their own module.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
bdmogal: yes, my bad. This logic got a bit jumbled up because I initially had the publish method inside the transaction, but then pulled it out. This is probably also the reason why I can't get a test to verify disabling of publishing to pass. Should call it a day when the body demands it :).

Will fix and also add a test.

cdap-common/src/main/resources/cdap-default.xml
bdmogal: I had it so, because ``MetadataChangePublisher`` is an interface and kafka is one implementation.  I will update the comment so its not confusing.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
bdmogal: Do we have a kafka client configured in standalone? I will check and update.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
bdmogal: agreed. will fix.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataServiceModule.java
```diff
@@ -33,65 +30,19 @@
...
-  }
+  protected void configure() {
+    Multibinder<HttpHandler> handlerBinder = Multibinder.newSetBinder(
+      binder(), HttpHandler.class, Names.named(Constants.Metadata.METADATA_HANDLERS_NAME));
 
-  private static final class NoOpMetadataChangePublisher implements MetadataChangePublisher {
+    CommonHandlers.add(handlerBinder);
+    handlerBinder.addBinding().to(MetadataHttpHandler.class);
+    handlerBinder.addBinding().to(LineageHandler.class);
+    expose(Key.get(new TypeLiteral<Set<HttpHandler>>() { }, Names.named(Constants.Metadata.METADATA_HANDLERS_NAME)));
```
chtyim: Why need to expose handlers?

cdap-common/src/main/resources/cdap-default.xml
chtyim: if it is enabled, where can the user specify Kafka configurations for publishing to (e.g. broker list/zk quorum, topic, partitions.. etc).

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -48,11 +56,17 @@
...
 
   @Inject
-  public BusinessMetadataStore(TransactionExecutorFactory txExecutorFactory,
-                               @Named(DataSetsModules.BASIC_DATASET_FRAMEWORK) final DatasetFramework dsFramework) {
+  BusinessMetadataStore(TransactionExecutorFactory txExecutorFactory,
+                        @Named(DataSetsModules.BASIC_DATASET_FRAMEWORK) final DatasetFramework dsFramework,
```
chtyim: Why need to use the base one? I thought the Lineage tracking DatasetFramework is able to work even run context is not set (meaning it'll just delegate to the underlying DS framework).

cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
```diff
@@ -882,5 +882,6 @@
     public static final String EXEC_THREADS = "metadata.service.exec.threads";
     public static final String METADATA_HANDLERS_NAME = "metadata.handlers";
     public static final String METADATA_UPDATES_KAFKA_TOPIC = "metadata.updates.kafka.topic";
+    public static final String METADATA_UPDATES_PUBLISH_ENABLED = "metadata.updates.publish.enabled";
```
chtyim: Be consistent with the naming. You have `EXEC_THREADS` above, but then you also have `METADATA_` prefixed named. I think the variable name doesn't need to have the `METADATA_` prefix since it's already defined inside `class Metadata`.

chtyim: Also, for the `EXEC_THREADS`, it's better name as `SERVICE_EXEC_THREADS` so that when used in code, it's fluently read as `Constants.Metadata.SERVICE_EXEC_THREADS` which matches the key name in cConf pretty well.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
chtyim: Which means, the basic dataset framework should be a private binding.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataServiceModule.java
bdmogal: Otherwise the MetadataHttpHandlerTest fails that it cannot find this binding.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
bdmogal: I am not sure, in this diff, I just removed the public modifier and added another parameter. @poornachandra, can you please confirm?

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetsModules.java
bdmogal: Reported https://issues.cask.co/browse/CDAP-3661

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
henry-cask: I think this needs to be remove so we could inject LineageDatasetFramework that relies on the  BASIC_DATASET_FRAMEWORK annotation as delegate.

bdmogal:
@anew @chtyim @poornachandra Reported https://issues.cask.co/browse/CDAP-3662 to publish to external kafka

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
poornachandra: LineageDatasetFramework depends on BusinessMetadataStore and LineageStore, which in turn depend on DatasetFramework. This leads to circular dependency. Hence BusinessMetadataStore and LineageStore need to use basic DatasetFramework.

poornachandra: JIRA for refactoring Guice bindings https://issues.cask.co/browse/CDAP-3661?focusedCommentId=18490&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18490

bdmogal:
Build is green: https://builds.cask.co/browse/CDAP-DUT2850-6/.

@anew , could you please take another look?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -230,6 +406,15 @@ public Void apply(BusinessMdsIterable input) throws Exception {
     });
   }
 
+  /**
+   * Note: This is
```
anew: Incomplete sentence?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
```diff
@@ -67,16 +81,54 @@ public BusinessMdsIterable get() {
...
+          input.businessMds.setProperty(entityId, entry.getKey(), entry.getValue());
+        }
+        return null;
+      }
+    });
+    publish(previousRef.get(), new MetadataRecord(entityId, propAdditions.build(), EMPTY_TAGS),
+            new MetadataRecord(entityId, propDeletions.build(), EMPTY_TAGS));
+  }
+
+  private void setPropertiesNoPublish(final Id.NamespacedId entityId, final Map<String, String> properties) {
```
anew: why the duplication of code for nnnNoPublish()? If we already inject a NoOpPublisher, that means you need no separate handling of the non-publishing case at all. It is actually better for test coverage if the code path is the same in both cases, just that the publisher won't to anything in one case.

anew:
I still have a major concern about duplicate code.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/BusinessMetadataStore.java
bdmogal: The main reason for doing that was to avoid the scans to get current state and calculate the diff, when publishing is disabled. Let me think if there is a better way of doing it.

anew: got it

anew:
LGTM

bdmogal:
Squashed commits. Tested on cluster. Merging.

PR #3931, opened 2015-09-14
Feature/package maintainer scripts update

dereklwood:
Fixes CDAP-3646

- [x] adds ``-L`` to all find commands, so that it will accommodate certain internal tools that symlink package-owned files out to alternate locations
- [x] fixes/formats the file header comments


wolf31o2:
:+1: LGTM

PR #3932, opened 2015-09-14
CDAP-2966 In case of internal server error while handling the request send more information in the response body.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-2966
Build: http://builds.cask.co/browse/CDAP-DUT2853-1

In `DatasetServiceClient`, we call http handlers for the desired functionality and check the response code. If the response code is not `OK`, we throw `DatasetManagementException` with response body as a part of exception message. However `HttpExceptionHandler` while sending the `INTERNAL_SERVER_ERROR` response, sends an empty message in the response string. This causes the appropriate message to be not propagated till the UI.


sagarkapare:
Updated the JIRA with the new screenshot of the error after fix.

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-DUT2853-1. @anew Please take a look.

sreevatsanraman:
LGTM :+1:

PR #3936, opened 2015-09-14
CDAP-3658: Fix Id.Program hierarchy

jwang47:
https://issues.cask.co/browse/CDAP-3658
http://builds.cask.co/browse/CDAP-DUT2858

### DIFF
--- a/cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
+++ b/cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
@@ -364,12 +364,12 @@ public abstract class Id {
       if (this == o) {
         return true;
       }
-      if (o == null || getClass() != o.getClass()) {
+      if (o == null || !Program.class.isAssignableFrom(o.getClass())) {
         return false;
       }
 
       Program program = (Program) o;
-      return application.equals(program.application) && id.equals(program.id);
+      return application.equals(program.application) && type.equals(program.type) && id.equals(program.id);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
```diff
@@ -1131,7 +1131,7 @@ public int hashCode() {
     public String toString() {
       return Objects.toStringHelper(this)
        .add("namespace", namespace)
-       .add("module", moduleId)
+        .add("module", moduleId)
```
anwar6953: why space

cdap-proto/src/test/java/co/cask/cdap/proto/IdTest.java
```diff
@@ -0,0 +1,39 @@
...
+    Id.Flow sameFlow = Id.Flow.from("ns1", "app1", "flow1");
+    Assert.assertEquals(sameFlow, flow);
+    Assert.assertEquals(sameFlow.toString(), flow.toString());
+
+    Id.Program program = Id.Program.from("ns1", "app1", ProgramType.FLOW, "flow1");
+    Assert.assertEquals(flow, program);
+    Assert.assertEquals(flow.toString(), program.toString());
+  }
+
+}
```
anwar6953: need newline

anwar6953:
Can you also add a comparison of 'type' field here:
https://github.com/caskdata/cdap/blob/feature/3658/cdap-proto/src/main/java/co/cask/cdap/proto/Id.java#L372

anwar6953:
LGTM

PR #3938, opened 2015-09-14
Use `cdap-site.xml` settings to kinit only when necessary

wolf31o2:
- Create `get_conf` to pull a property from a Hadoop-formatted XML file
- Use get_conf to only run `hive` when necessary
- Remove use of `/etc/default/cdap-master` environment for Kerberos
- Move all Kerberos-related functionality to service script


sreevatsanraman:
LGTM

PR #3939, opened 2015-09-15
exclude jetty from the etl batch app to avoid jetty conflicts.

albertshau:


albertshau:
http://builds.cask.co/browse/CDAP-DUT2862

gsps1:
LGTM

PR #3933, opened 2015-09-14
CDAP-3605 Intercept entitiy deletion to also remove metadata

henry-cask:
Intercept the entity deletion of App, Program, Dataset, and Stream so it also remove the corresponding business metadata (property and tags).

Add Dataset service and transaction manager to some tests.

PR #3944, opened 2015-09-15
CDAP-3605 Entity deletion should remove its metadata.

henry-cask:
Intercept the entity deletion of App, Program, Dataset, and Stream so it also remove the corresponding business metadata (property and tags).

Move BusinessMetadataStore to interface and add new implementations for in memory and distributed mode.

For UpgradeTool I bind the BusinessMetadataStore to in-memory since it seems like it will have dataset service ready at the time of the upgrade.

Use Modules.override to change the binding for the DatasetsModule to make BusinessMetadataStore bind to in-memory for some tests that do not need business metadata store testing and does not have DatasetService and TransactionManager running.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
@@ -45,6 +45,7 @@ import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.io.Locations;
 import co.cask.cdap.common.namespace.NamespacedLocationFactory;
 import co.cask.cdap.config.PreferencesStore;
+import co.cask.cdap.data2.metadata.service.BusinessMetadataStore;
 import co.cask.cdap.data2.registry.UsageRegistry;
 import co.cask.cdap.data2.transaction.queue.QueueAdmin;
 import co.cask.cdap.data2.transaction.stream.StreamConsumerFactory;
@@ -68,6 +69,7 @@ import co.cask.cdap.proto.artifact.ArtifactSummary;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
@@ -85,6 +87,8 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -126,6 +130,7 @@ public class ApplicationLifecycleService extends AbstractIdleService {
   private final MetricStore metricStore;
   private final ArtifactRepository artifactRepository;
   private final ManagerFactory<AppDeploymentInfo, ApplicationWithPrograms> managerFactory;
+  private final BusinessMetadataStore businessMds;
 
   @Inject
   public ApplicationLifecycleService(ProgramRuntimeService runtimeService, Store store, CConfiguration configuration,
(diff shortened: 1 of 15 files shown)


henry-cask:
Build:
http://builds.cask.co/browse/CDAP-RBT418-1

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -695,6 +705,45 @@ void deleteApp(Id.Application appId, ApplicationSpecification spec) throws Excep
...
+    }
+    if (appSpec.getSpark() != null) {
+      programTypeToNames.put(ProgramType.SPARK, appSpec.getSpark().keySet());
+    }
+    if (appSpec.getWorkers() != null) {
+      programTypeToNames.put(ProgramType.WORKER, appSpec.getWorkers().keySet());
+    }
+
+    for (ProgramType programType : programTypeToNames.keySet()) {
+      Set<String> programNames = programTypeToNames.get(programType);
```
poornachandra: We can use the following construct here to simplify the code, right?
```
for (Map.Entry<ProgramType, Set<String> entry : programTypeToNames.entrySet()) {
  ...
}
```

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/InMemoryBusinessMetadataStore.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Override
+  public void addTags(Id.NamespacedId entityId, String... tagsToAdd) {
+    // NO-OP
+  }
+
+  @Override
+  public MetadataRecord getMetadata(Id.NamespacedId entityId) {
+    return null;
```
poornachandra: The interface says none of the methods can return null, it would be good to return empty objects so that the contract remains the same.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/InMemoryBusinessMetadataStore.java
```diff
@@ -0,0 +1,95 @@
...
+import co.cask.cdap.proto.metadata.MetadataRecord;
+import co.cask.cdap.proto.metadata.MetadataSearchTargetType;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Implementation of {@link BusinessMetadataStore} used in memory mode.
+ */
+public class InMemoryBusinessMetadataStore implements BusinessMetadataStore {
```
poornachandra: To make it clear that this metadata store does not do anything, it would be good to call it `NoOpBusinessMetadataStore`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/service/DistributedBusinessMetadataStore.java
```diff
@@ -0,0 +1,460 @@
...
+    Id.DatasetInstance.from(Id.Namespace.SYSTEM, "business.metadata");
+  private static final Map<String, String> EMPTY_PROPERTIES = ImmutableMap.of();
+  private static final Set<String> EMPTY_TAGS = ImmutableSet.of();
+
+  private final Transactional<BusinessMdsIterable, BusinessMetadataDataset> txnl;
+  private final CConfiguration cConf;
+  private final MetadataChangePublisher changePublisher;
+
+  @Inject
+  DistributedBusinessMetadataStore(TransactionExecutorFactory txExecutorFactory,
```
poornachandra: I agree with @bdmogal's comment about name change.

poornachandra:
Also, we need to address some of @bdmogal's comments from PR https://github.com/caskdata/cdap/pull/3933.

poornachandra:
@henry-cask I took a pass, the code looks good overall. Please address comments.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
henry-cask: I did it before but Albert said it is better to explicitly define which programs to access (you can check the one from #3933 (https://github.com/caskdata/cdap/commit/1f269be43f024da3ef4483765980cf293b5db9ea)

henry-cask:
Updated the PR based on comments and run another build:
http://builds.cask.co/browse/CDAP-RBT418-2

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
bdmogal: @henry-cask , I think @poornachandra meant to say that using a loop on entryset instead of keyset might reduce another get that you have to do on line 739.

henry-cask: Ah ok, sorry misunderstood. Yeah we could make it that way.

poornachandra:
The changes look good. Please resolve conflicts, that will lead to undoing of over-riding guice bindings in non-metadata test classes.

henry-cask:
Build pass: http://builds.cask.co/browse/CDAP-RBT418-JOB1-4

Will rebase and merge

PR #3948, opened 2015-09-15
Fix/ui save reset button

ajainarayanan:
- Adds the ability to communicate between bottom panel controller and individual tabs
- Reverses 'save' & 'reset' buttons' functionality in the bottom panel.
- Adds a confirmation before user clicks on other nodes when the current node's configuration needs to be saved.

TODO:
  - Will be adding in label 
  - Handle the case when a node is deleted to blank out the bottom panel.

savereset

### DIFF
--- a/cdap-ui/app/features/adapters/services/node-config-service.js
+++ b/cdap-ui/app/features/adapters/services/node-config-service.js
@@ -1,0 +1,25 @@
+angular.module(PKG.name + '.feature.adapters')
+  .service('MyNodeConfigService', function() {
+    this.pluginChangeListeners = [];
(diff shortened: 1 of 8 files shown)


cdap-ui/app/features/adapters/services/node-config-service.js
```diff
@@ -0,0 +1,19 @@
+angular.module(PKG.name + '.feature.adapters')
```
elfenheart: license header

elfenheart:
just need to add license header to node-config-service.js, otherwise :+1: LGTM

PR #3950, opened 2015-09-15
Fix Documentation Build and Namespaces description

awholegunch:
- Replace "physical" with "logical" for namespaces in three files.
- Set the parent branch explicitly for the documentation build to get it to build with new "docs" prefix.

Fixes: https://issues.cask.co/browse/CDAP-3664 (namespaces)

Building at: http://builds.cask.co/browse/CDAP-DDBD572-1


sreevatsanraman:
LGTM

PR #3951, opened 2015-09-15
Feature/revert cdap 3119

chtyim:


bdmogal:
Code LGTM. If you have a cluster with this code, should we run integration test?

chtyim:
Tested on CDH cluster. We can test for it in integration test clusters as well. I don't expect it will fail as it's simply 3.1 release code.

PR #3952, opened 2015-09-15
Move kinit code to cdap_kinit function

wolf31o2:
This allows the functionality to be reused in other scripts and functions without otherwise modifying the environment.


jwang47:
LGTM

PR #3953, opened 2015-09-15
Fix/ui hydrator detail view

elfenheart:
TODO:
- Status view for Realtime [DONE]
- Where to show Pipeline Status (waiting on Brady) [DONE]

hydratordetail


cdap-ui/app/features/adapters/controllers/detail-ctrl.js
```diff
@@ -15,12 +15,214 @@
...
+
+
+    $scope.nodes = [];
+
+    function initializeDAG() {
+      try{
+        rAdapterDetail.config = JSON.parse(rAdapterDetail.configuration);
+      } catch(e) {
+        console.log('ERROR in configuration from backend: ', e);
+      }
```
ajainarayanan: Should we have a return statement here? Beyond this catch block nothing is valid right?

cdap-ui/app/features/adapters/controllers/detail-ctrl.js
```diff
@@ -15,12 +15,214 @@
...
+              .$promise
+              .then(function () {
+                $scope.appStatus = 'SUSPENDED';
+              });
+          } else {
+            myWorkersApi.doAction(angular.extend(params, { action: 'stop' }, {}))
+              .$promise
+              .then(function () {
+                $scope.appStatus = 'STOPPED';
+              });
```
ajainarayanan: Should we show the status from backend? Do we get one? Right now we are updating the status in UI as soon as we get a 200 OK. But inherently there could be something wrong with the operation in the backend right?

When the backend returns 200 OK it doesn't mean the action is complete (does it?)

cdap-ui/app/features/adapters/services/adapter-detail-factory.js
```diff
@@ -0,0 +1,69 @@
...
+          if (program.type === 'Workflow') {
+            publicObj.params.workflowId = program.id;
+          } else if (program.type === 'Mapreduce') {
+            publicObj.logsParams.mapreduceId = program.id;
+          }
+        });
+
+      } else {
+        publicObj.api = myWorkersApi;
+        publicObj.logsApi = myWorkersApi;
```
ajainarayanan: :+1: Awesome!

cdap-ui/app/features/adapters/services/adapter-detail-factory.js
```diff
@@ -0,0 +1,69 @@
...
+        namespace: $state.params.namespace,
+        appId: app.name,
+        max: 50
+      };
+
+      var config;
+      try {
+        config = JSON.parse(app.configuration);
+      } catch (e) {
+        console.log('ERROR cannot parse configuration');
```
ajainarayanan: Same here. We should return in the catch block as subsequent statements become invalid beyond this point if there is a JSON parse error.

cdap-ui/app/features/adapters/templates/tabs/datasets.html
```diff
@@ -0,0 +1,17 @@
+<div class="table-responsive">
+  <table class="table table-bordered">
```
ajainarayanan: We should have a view where is there are no associated datasets (before even running once)

cdap-ui/app/features/adapters/templates/tabs/metrics.html
```diff
@@ -0,0 +1,34 @@
+<div ng-controller="AdaptersDetailMetricsController">
+  <div class="table-responsive">
```
ajainarayanan: Here as well. When there are no metrics this tab should not be showing an empty table with just header.

cdap-ui/app/features/adapters/controllers/detail-ctrl.js
elfenheart: When backend returns 200 OK, it does mean the action is complete. It won't return anything if the backend is still stopping or resuming the schedule. And if something goes wrong, it will throw an error

ajainarayanan:
:+1: LGTM

PR #3955, opened 2015-09-15
CDAP-3563 removing cdap cluster disk usage metric

albertshau:
The metric was introduced a long time ago before filesets were
added. The metric is incorrect, and it requires read permissions
on hbase directory which we might not always have. Removing it
until there is a better way to calculate it.


albertshau:
http://builds.cask.co/browse/CDAP-RBT420

chtyim:
LGTM

PR #3956, opened 2015-09-15
Misc lineage fixes

poornachandra:
* Renaming lineage REST enpoint /metadata/accesses to /metadata
 * Adding test to verify duplicate access record filtering

It would be easier to review them as separate commits.

### DIFF
--- a/cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/lineage/LineageDatasetTest.java
+++ b/cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/lineage/LineageDatasetTest.java
@@ -71,8 +71,8 @@ public class LineageDatasetTest {
     Assert.assertEquals(expected, relations.iterator().next());
 
     // Assert metadata
-    System.out.println(lineageDataset.getAccesses(run));
-    Assert.assertEquals(metadataRecords, lineageDataset.getAccesses(run));
+    System.out.println(lineageDataset.getRunMetadata(run));
+    Assert.assertEquals(metadataRecords, lineageDataset.getRunMetadata(run));
   }
 
   @Test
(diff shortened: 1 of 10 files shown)


poornachandra:
Build - http://builds.cask.co/browse/CDAP-RBT421-1

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/lineage/LineageDatasetTest.java
```diff
@@ -71,8 +71,8 @@ public void testOneRelation() throws Exception {
     Assert.assertEquals(expected, relations.iterator().next());
 
     // Assert metadata
-    System.out.println(lineageDataset.getAccesses(run));
-    Assert.assertEquals(metadataRecords, lineageDataset.getAccesses(run));
+    System.out.println(lineageDataset.getRunMetadata(run));
```
bdmogal: remove

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/writer/BasicLineageWriterTest.java
```diff
@@ -0,0 +1,116 @@
...
+        new MetadataRecord(stream, ImmutableMap.<String, String>of(), ImmutableSet.of("stag1", "stag2")),
+        new MetadataRecord(program.getApplication(), ImmutableMap.<String, String>of(), ImmutableSet.<String>of()),
+        new MetadataRecord(program, ImmutableMap.<String, String>of(), ImmutableSet.<String>of())),
+      lineageStore.getRunMetadata(run1));
+
+    // Add another tag to stream
+    businessMetadataStore.addTags(stream, "stag3");
+    // Write access for run1 again
+    lineageWriter.addAccess(run1, stream, AccessType.READ);
+    // The write should be no-op, and metadata for run1 should not be updated
```
bdmogal: This is so we don't keep updating everytime the same access is made by the same runId, right? Eventually we may have some sort of internal notification that metadata has been updated, so we update the access the next time it happens?

But thats for future. This :+1: right now.

bdmogal:
:+1: . Merge after a clean unit test build.

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/writer/BasicLineageWriterTest.java
poornachandra: I think we want to record the tags the first time a Dataset is accessed, because that's when the Dataset code is loaded.

poornachandra:
Will merge after build - http://builds.cask.co/browse/CDAP-RBT421-3

bdmogal:
Build green. Merging.

PR #3957, opened 2015-09-15
CDAP-3406 Update Kerberos docs

wolf31o2:
This updates the Kerberos section of our installation documents. It removes some sections which are best left up to the individual distributions to document and only documents the changes required for CDAP, specifically.

### DIFF
--- a/cdap-docs/admin-manual/source/installation/installation.rst
+++ b/cdap-docs/admin-manual/source/installation/installation.rst
@@ -607,7 +607,7 @@ When all the packages and dependencies have been installed, and the configuratio
 parameters set, you can start the services on each of the CDAP boxes by running the
 command::
 
-  $ for i in /etc/init.d/cdap* ; do sudo service $i restart ; done
+  $ for i in `ls /etc/init.d/ | grep cdap` ; do sudo service $i restart ; done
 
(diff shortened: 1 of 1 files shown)


awholegunch:
LGTM, but I will ask Chris to take a final pass to see if I've missed or messed anything.

wolf31o2:
:+1: Edits LGTM

cdap-docs/admin-manual/source/installation/installation.rst
```diff
@@ -653,7 +607,7 @@ When all the packages and dependencies have been installed, and the configuratio
 parameters set, you can start the services on each of the CDAP boxes by running the
 command::
 
-  $ for i in `ls /etc/init.d/ | grep cdap` ; do sudo service $i restart ; done
+  $ for i in /etc/init.d/cdap* ; do sudo service $i restart ; done
```
dereklwood: this wont work... it would need to be ``sudo $i restart`` since  ``service`` wont recognize ``/etc/init.d/cdap-master`` as a service name.  i think the original is better, to preserve the use of ``service``

wolf31o2: ACK. You're correct. My bad. I didn't mean to include this, anyway.

PR #3958, opened 2015-09-15
CDAP-3604 Localize plugin Jars for both standalone and distributed runs

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-3604

Build : http://builds.cask.co/browse/CDAP-RBT422

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
@@ -15,32 +15,48 @@
  */
 package co.cask.cdap.app.runtime;
 
+import co.cask.cdap.api.app.ApplicationSpecification;
+import co.cask.cdap.api.plugin.Plugin;
 import co.cask.cdap.app.program.Program;
+import co.cask.cdap.common.ArtifactNotFoundException;
 import co.cask.cdap.common.app.RunIds;
+import co.cask.cdap.common.conf.CConfiguration;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.io.Locations;
+import co.cask.cdap.common.utils.DirUtils;
 import co.cask.cdap.internal.app.runtime.AbstractListener;
 import co.cask.cdap.internal.app.runtime.BasicArguments;
 import co.cask.cdap.internal.app.runtime.ProgramOptionConstants;
 import co.cask.cdap.internal.app.runtime.ProgramRunnerFactory;
 import co.cask.cdap.internal.app.runtime.SimpleProgramOptions;
+import co.cask.cdap.internal.app.runtime.artifact.ArtifactDetail;
+import co.cask.cdap.internal.app.runtime.artifact.ArtifactRepository;
+import co.cask.cdap.internal.app.runtime.artifact.Artifacts;
 import co.cask.cdap.internal.app.runtime.service.SimpleRuntimeInfo;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.ProgramType;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
+import com.google.common.base.Throwables;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.google.common.collect.Table;
+import com.google.common.io.Files;
 import com.google.common.util.concurrent.AbstractIdleService;
 import org.apache.twill.api.RunId;
 import org.apache.twill.common.Threads;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
+import java.io.IOException;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -55,25 +71,119 @@ public abstract class AbstractProgramRuntimeService extends AbstractIdleService 
   private static final EnumSet<ProgramController.State> COMPLETED_STATES = EnumSet.of(ProgramController.State.COMPLETED,
                                                                                       ProgramController.State.KILLED,
                                                                                       ProgramController.State.ERROR);
-
+  private final CConfiguration cConf;
   private final ReadWriteLock runtimeInfosLock;
   private final Table<ProgramType, RunId, RuntimeInfo> runtimeInfos;
   private final ProgramRunnerFactory programRunnerFactory;
+  private final ArtifactRepository artifactRepository;
 
-  protected AbstractProgramRuntimeService(ProgramRunnerFactory programRunnerFactory) {
+  protected AbstractProgramRuntimeService(CConfiguration cConf, ProgramRunnerFactory programRunnerFactory,
+                                          ArtifactRepository artifactRepository) {
+    this.cConf = cConf;
     this.runtimeInfosLock = new ReentrantReadWriteLock();
     this.runtimeInfos = HashBasedTable.create();
     this.programRunnerFactory = programRunnerFactory;
+    this.artifactRepository = artifactRepository;
   }
 
   @Override
   public RuntimeInfo run(Program program, ProgramOptions options) {
     ProgramRunner runner = programRunnerFactory.create(ProgramRunnerFactory.Type.valueOf(program.getType().name()));
     Preconditions.checkNotNull(runner, "Fail to get ProgramRunner for type " + program.getType());
-    ProgramOptions optionsWithRunId = addRunId(options, RunIds.generate());
-    final RuntimeInfo runtimeInfo = createRuntimeInfo(runner.run(program, optionsWithRunId), program);
-    programStarted(runtimeInfo);
+    RunId runId = RunIds.generate();
+    ProgramOptions optionsWithRunId = addRunId(options, runId);
+    File tempDir = createTempDirectory(program.getId(), runId);
+    Runnable cleanUpTask = createCleanupTask(tempDir);
+    ProgramOptions optionsWithPlugins = null;
+    try {
+      optionsWithPlugins = createPluginSnapshot(optionsWithRunId, program.getId(), tempDir,
+                                                program.getApplicationSpecification());
+    } catch (IOException e) {
+      cleanUpTask.run();
+      LOG.error("Exception while trying to createPluginSnapshot", e);
+      Throwables.propagate(e);
+    }
+
+    final RuntimeInfo runtimeInfo = createRuntimeInfo(runner.run(program, optionsWithPlugins), program);
+    programStarted(runtimeInfo, cleanUpTask);
     return runtimeInfo;
+  }
+
+  private Runnable createCleanupTask(final File... resources) {
+    return new Runnable() {
+      @Override
+      public void run() {
+        for (File file : resources) {
+          if (file == null) {
+            continue;
+          }
+
+          try {
+            if (file.isDirectory()) {
+              DirUtils.deleteDirectoryContents(file);
+            } else {
+              file.delete();
+            }
+          } catch (Throwable t) {
+            LOG.warn("Exception when cleaning up resource {}", file, t);
+          }
+        }
+      }
+    };
+  }
+
+  /**
+   * Creates a local temporary directory for this program run.
+   */
+  private File createTempDirectory(Id.Program programId, RunId runId) {
+    File tempDir = new File(cConf.get(Constants.CFG_LOCAL_DATA_DIR),
+                            cConf.get(Constants.AppFabric.TEMP_DIR)).getAbsoluteFile();
+    File dir = new File(tempDir, String.format("%s.%s.%s.%s.%s",
+                                               programId.getType().name().toLowerCase(),
+                                               programId.getNamespaceId(), programId.getApplicationId(),
+                                               programId.getId(), runId.getId()));
+    dir.mkdirs();
+    return dir;
+  }
+
+  /**
+   * Return the copy of the {@link ProgramOptions} including locations of plugin artifacts in it.
+   * @param options the {@link ProgramOptions} in which the locations of plugin artifacts needs to be included
+   * @param programId Id of the Program
+   * @param tempDir Temporary Directory to create the plugin artifact snapshot
+   * @param appSpec program's Application Specification
+   * @return the copy of the program options with locations of plugin artifacts included in them
+   */
+  private ProgramOptions createPluginSnapshot(ProgramOptions options, Id.Program programId, File tempDir,
+                                              @Nullable ApplicationSpecification appSpec) throws IOException {
+    // appSpec is null in an unit test
+    if (appSpec == null) {
+      return options;
+    }
+
+    Set<String> files = Sets.newHashSet();
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+    builder.putAll(options.getArguments().asMap());
+    for (Map.Entry<String, Plugin> pluginEntry : appSpec.getPlugins().entrySet()) {
+      Plugin plugin = pluginEntry.getValue();
+      File destFile = new File(tempDir, Artifacts.getFileName(plugin.getArtifactId()));
+      // Skip if the file has already been copied.
+      if (!files.add(destFile.getName())) {
+        continue;
+      }
+
+      try {
+        ArtifactDetail detail = artifactRepository.getArtifact(Id.Artifact.from(programId.getNamespace(),
+                                                                                plugin.getArtifactId()));
+        Files.copy(Locations.newInputSupplier(detail.getDescriptor().getLocation()), destFile);
+      } catch (ArtifactNotFoundException e) {
+        throw new IllegalArgumentException(String.format("Artifact %s could not be found", plugin.getArtifactId()), e);
+      }
+    }
+    LOG.debug("Plugin artifacts of {} copied to {}", programId, tempDir.getAbsolutePath());
+    builder.put(ProgramOptionConstants.PLUGIN_DIR, tempDir.getAbsolutePath());
+    return new SimpleProgramOptions(options.getName(), new BasicArguments(builder.build()),
+                                    options.getUserArguments(), options.isDebug());
   }
 
   /**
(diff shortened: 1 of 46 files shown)


cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactId.java
```diff
@@ -49,11 +49,7 @@ public ArtifactScope getScope() {
 
   @Override
   public String toString() {
-    return "ArtifactId{" +
-      "name='" + name + '\'' +
-      ", version=" + version +
-      ", scope='" + scope + '\'' +
-      '}';
+    return String.format("%s-%s-%s", scope, name, version);
```
albertshau: scope:name-version for consistency with other places

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -71,11 +90,91 @@ public RuntimeInfo run(Program program, ProgramOptions options) {
...
+      return options;
+    }
+
+    Id.Artifact appArtifactId = Id.Artifact.from(namespace, appSpec.getArtifactId());
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+    builder.putAll(options.getArguments().asMap());
+    for (Map.Entry<String, Plugin> pluginEntry : appSpec.getPlugins().entrySet()) {
+      Plugin plugin = pluginEntry.getValue();
+      try {
+        Map.Entry<ArtifactDescriptor, PluginClass> artifactEntry = artifactRepository.getPlugin(
```
albertshau: use `artifactRepository.getArtifact(pluginArtifactId)` instead of adding this method

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -71,11 +90,91 @@ public RuntimeInfo run(Program program, ProgramOptions options) {
...
+        Files.copy(Locations.newInputSupplier(artifactEntry.getKey().getLocation()), destFile);
+      } catch (IOException e) {
+        throw Throwables.propagate(e);
+      } catch (PluginNotExistsException e) {
+        throw new IllegalArgumentException(String.format("Plugin of type %s, name %s could not be found",
+                                                         plugin.getPluginClass().getType(),
+                                                         plugin.getPluginClass().getName()), e);
+      }
+    }
+    LOG.info("Local files are here : {}", tempDir.getAbsolutePath());
```
albertshau: debug

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -71,11 +90,91 @@ public RuntimeInfo run(Program program, ProgramOptions options) {
...
+    Id.Artifact appArtifactId = Id.Artifact.from(namespace, appSpec.getArtifactId());
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+    builder.putAll(options.getArguments().asMap());
+    for (Map.Entry<String, Plugin> pluginEntry : appSpec.getPlugins().entrySet()) {
+      Plugin plugin = pluginEntry.getValue();
+      try {
+        Map.Entry<ArtifactDescriptor, PluginClass> artifactEntry = artifactRepository.getPlugin(
+          appArtifactId, plugin.getPluginClass().getType(), plugin.getPluginClass().getName(), plugin.getArtifactId());
+        File destFile = new File(tempDir, String.format("%s.jar", plugin.getArtifactId().toString()));
+        Files.copy(Locations.newInputSupplier(artifactEntry.getKey().getLocation()), destFile);
```
albertshau: can have multiple plugins that use the same artifact, won't need to copy every time

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -71,11 +90,91 @@ public RuntimeInfo run(Program program, ProgramOptions options) {
...
+
+    Id.Artifact appArtifactId = Id.Artifact.from(namespace, appSpec.getArtifactId());
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+    builder.putAll(options.getArguments().asMap());
+    for (Map.Entry<String, Plugin> pluginEntry : appSpec.getPlugins().entrySet()) {
+      Plugin plugin = pluginEntry.getValue();
+      try {
+        Map.Entry<ArtifactDescriptor, PluginClass> artifactEntry = artifactRepository.getPlugin(
+          appArtifactId, plugin.getPluginClass().getType(), plugin.getPluginClass().getName(), plugin.getArtifactId());
+        File destFile = new File(tempDir, String.format("%s.jar", plugin.getArtifactId().toString()));
```
albertshau: better to construct the filename yourself rather than relying on the fact that toString() gives a valid filename

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/adapter/PluginInstantiator.java
```diff
@@ -98,38 +112,86 @@ public ClassLoader getArtifactClassLoader(ArtifactDescriptor artifactDescriptor)
...
+   */
+  @SuppressWarnings("unchecked")
+  public <T> Class<T> loadClass(Location location,
                                 PluginClass pluginClass) throws IOException, ClassNotFoundException {
-    return (Class<T>) getArtifactClassLoader(artifactDescriptor).loadClass(pluginClass.getClassName());
+    return (Class<T>) getArtifactClassLoader(location).loadClass(pluginClass.getClassName());
   }
 
   /**
    * Creates a new instance of the given plugin class.
```
albertshau: can you add some details to the javadocs about the pluginDir and expectations this class has for where the jars are? Without that its very hard for somebody who isn't already familiar with the class to figure out what's going on.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepository.java
```diff
@@ -225,12 +225,34 @@ public ArtifactDetail getArtifact(Id.Artifact artifactId) throws IOException, Ar
...
    * Returns a {@link Map.Entry} representing the plugin information for the plugin being requested.
+   * @param artifactId the id of the artifact to get plugins for
+   * @param pluginType plugin type name
+   * @param pluginName plugin name
+   * @param pluginArtifactId {@link ArtifactId} plugin's artifact id
+   * @return the entry found
+   * @throws IOException if there was an exception reading plugin metadata from the artifact store
+   * @throws PluginNotExistsException if no plugins of the given type and name are available to the given artifact
+   */
+  public Map.Entry<ArtifactDescriptor, PluginClass> getPlugin(Id.Artifact artifactId, String pluginType,
```
albertshau: shouldn't need this method, use getArtifact on the plugin artifact id.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -210,8 +220,30 @@ public TwillController launch(TwillApplication twillApplication, Iterable<String
...
 
+
+  private ProgramOptions addArtifactPluginFiles(ProgramOptions options, Map<String, Plugin> plugins,
+                                      Map<String, LocalizeResource> localizeResources) {
+    File localDir = new File(options.getArguments().getOption(ProgramOptionConstants.PLUGIN_BASEPATH));
+    Set<String> files = Sets.newHashSet();
+    for (Plugin plugin : plugins.values()) {
+      File localFile = new File(localDir, String.format("%s.jar", plugin.getArtifactId().toString()));
+      String localizedName = String.format("%s/%s", Constants.Plugin.DIRECTORY, localFile.getName());
+      if (files.contains(localizedName)) {
```
albertshau: nit: can simplify this, Set.add() will tell you if the element existed already

albertshau:
good work, some small comments

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -71,11 +90,91 @@ public RuntimeInfo run(Program program, ProgramOptions options) {
...
+  /**
+   * Creates a local temporary directory for this program run.
+   */
+  private File createTempDirectory(Id.Program programId) {
+    File tempDir = new File(cConf.get(Constants.CFG_LOCAL_DATA_DIR),
+                            cConf.get(Constants.AppFabric.TEMP_DIR)).getAbsoluteFile();
+    File dir = new File(tempDir, String.format("%s.%s.%s.%s.%s",
+                                               programId.getType().name().toLowerCase(),
+                                               programId.getNamespaceId(), programId.getApplicationId(),
+                                               programId.getId(), UUID.randomUUID().toString()));
```
chtyim: Why not use the runId instead of random uuid?

cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactId.java
gokulavasan: i had it as ':' before. But because of this issue (https://issues.apache.org/jira/browse/HADOOP-3257) we can't localize files with ':' in the name.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
chtyim: BTW, the MapReduceRuntimeService is using the same directory (if using runid). I think it's better for MapReduceRuntimeService to create a sub-dir under the tempDir. Something like `data/temp/runner/mapreduce.appId....`

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
chtyim: Better message it as `"Plugin artifacts of {} copied to {}", programId, tempDir`.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -177,14 +276,14 @@ protected void updateRuntimeInfo(ProgramType type, RunId runId, RuntimeInfo runt
     lock.lock();
     try {
       if (!runtimeInfos.contains(type, runId)) {
-        runtimeInfos.put(type, runId, programStarted(runtimeInfo));
+        runtimeInfos.put(type, runId, programStarted(runtimeInfo, null));
```
chtyim: It's better not to use null (and target methods are not `@Nullable` anyway). Use a no-op Runnable.

cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactId.java
gokulavasan: May be I will add a method to generate the file name of the Jar in ArtifactId and leave the toString as is.

chtyim: I think it's better not to localize based on `toString`. You can always construct the file name explicitly (like "name-version-scope.jar").

chtyim: The generation of file name shouldn't be part of `ArtifactId`. This class should be pure data. You can have helper method in app-fabric to unify generation of file name from artifactId.

gokulavasan: Forgot that this class is in fact in cdap-api. Will move the method into app-fabric.

gokulavasan:
@albertshau @chtyim Addressed all comments. Please take a look. Thanks!

gokulavasan:
Build passed : http://builds.cask.co/browse/CDAP-RBT422-5

cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactId.java
```diff
@@ -49,11 +49,7 @@ public ArtifactScope getScope() {
 
   @Override
   public String toString() {
-    return "ArtifactId{" +
-      "name='" + name + '\'' +
-      ", version=" + version +
-      ", scope='" + scope + '\'' +
-      '}';
+    return String.format("%s:%s-%s", scope, name, version);
```
chtyim: Do you still need this change? I prefer the old toString

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactFile.java
```diff
@@ -0,0 +1,30 @@
...
+
+package co.cask.cdap.internal.app.runtime.artifact;
+
+import co.cask.cdap.api.artifact.ArtifactId;
+import co.cask.cdap.proto.Id;
+
+/**
+ * Helper methods related to handling of {@link Id.Artifact}s.
+ */
+public class ArtifactFile {
```
chtyim: final class and private constructor for util class. Also, better not to have such specific util class. Maybe you can have a `Artifacts` util class which has a collection of helper methods for Artifact related operations.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -55,27 +71,115 @@
...
+      if (!files.add(destFile.getName())) {
+        continue;
+      }
+
+      try {
+        ArtifactDetail detail = artifactRepository.getArtifact(Id.Artifact.from(programId.getNamespace(),
+                                                                                plugin.getArtifactId()));
+        Files.copy(Locations.newInputSupplier(detail.getDescriptor().getLocation()), destFile);
+      } catch (IOException e) {
+        throw Throwables.propagate(e);
```
chtyim: Better have this method declare `throws IOException` so that the caller need to catch and handle it (e.g. need to run the cleanup task)

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -55,27 +71,115 @@
...
+      } catch (ArtifactNotFoundException e) {
+        throw new IllegalArgumentException(String.format("Artifact %s could not be found", plugin.getArtifactId()), e);
+      }
+    }
+    LOG.debug("Plugin artifacts of {} copied to {}", tempDir.getAbsolutePath());
```
chtyim: Argument not matching. You have two `{}`, but there is only one argument. I believe the first one should be the programId (meaning the programId need to be passed to this method).

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/DefaultPluginConfigurer.java
```diff
@@ -133,29 +142,24 @@ public void addPlugins(Map<String, Plugin> plugins) {
...
 
-  /**
-   * Register the given plugin in this configurer.
-   */
-  private void registerPlugin(String pluginId, ArtifactDescriptor artifactDescriptor, PluginClass pluginClass,
-                              PluginProperties properties) {
-    plugins.put(pluginId, new Plugin(artifactDescriptor.getArtifactId(), artifactDescriptor.getLocation().toURI(),
-                                     pluginClass, properties));
+    ArtifactId artifactId = pluginEntry.getKey().getArtifactId();
+    File destFile = new File(pluginInstantiator.getBaseDir(), ArtifactFile.getFileName(artifactId));
```
chtyim: Maybe better to move this logic into `PluginInstantiator` than exposing the baseDir. You can have `PluginInstantiator.addArtifact(Location artifactLocation)`.

chtyim: And have the `addArtifact` method `throws IOException`.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -141,8 +150,8 @@ protected EventHandler createEventHandler(CConfiguration cConf) {
   }
 
   @Override
-  public final ProgramController run(final Program program, final ProgramOptions options) {
-    final String schedulerQueueName = options.getArguments().getOption(Constants.AppFabric.APP_SCHEDULER_QUEUE);
+  public final ProgramController run(final Program program, final ProgramOptions oldOptions) {
```
chtyim: Why rename to oldOptions? I don't see a newOptions.

chtyim: oh.. I see it below now.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -210,8 +221,31 @@ public TwillController launch(TwillApplication twillApplication, Iterable<String
     }
   }
 
+
+  private ProgramOptions addArtifactPluginFiles(ProgramOptions options, Map<String, Plugin> plugins,
+                                      Map<String, LocalizeResource> localizeResources) {
```
chtyim: misalignment

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -210,8 +221,31 @@ public TwillController launch(TwillApplication twillApplication, Iterable<String
...
+    File localDir = new File(options.getArguments().getOption(ProgramOptionConstants.PLUGIN_BASEPATH));
```
chtyim: Better call it pluginDir or pluginBase.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/distributed/AbstractDistributedProgramRunner.java
```diff
@@ -210,8 +221,31 @@ public TwillController launch(TwillApplication twillApplication, Iterable<String
...
+    for (Plugin plugin : plugins.values()) {
+      File localFile = new File(localDir, ArtifactFile.getFileName(plugin.getArtifactId()));
+      // Skip file if it has already been added
+      if (!files.add(localFile.getName())) {
+        continue;
+      }
+
+      String localizedName = String.format("%s/%s", Constants.Plugin.DIRECTORY, localFile.getName());
+      localizeResources.put(localizedName, new LocalizeResource(localFile, false));
+      LOG.info("Localizing Resource {} here {}", localFile.toURI().toString(), localizedName);
```
chtyim: no need to toURI.toString. Just log the `localFile`.

albertshau:
one comment about javadoc, otherwise lgtm

gokulavasan:
@albertshau Thanks for the review. Will merge after the build passes.

gokulavasan:
@albertshau Can you review the couple of last commits? Fix was required to make MR Plugin archive work in distributed mode.

albertshau:
:+1:

gokulavasan:

PR #3960, opened 2015-09-15
Improve network handling for CDAP VM

wolf31o2:
This prevents the persistent network generator rules from running, removes any previously created rules, removes any DHCP and UDEV cached data, and adds a two second delay after interface is up before starting DHCP (copied from chef/bento's networking scripts).


sreevatsanraman:
LGTM

PR #3961, opened 2015-09-16
CDAP-3662: Metadata notifications should always only be published to external Kafka

jwang47:
https://issues.cask.co/browse/CDAP-3662
http://builds.cask.co/browse/CDAP-RBT424

- [x] Get KafkaMetadataChangePublisherTest to pass locally
- [x] Test in cluster with external Kafka

### DIFF
--- a/cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
+++ b/cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
@@ -883,5 +883,6 @@ public final class Constants {
     public static final String HANDLERS_NAME = "metadata.handlers";
     public static final String UPDATES_KAFKA_TOPIC = "metadata.updates.kafka.topic";
     public static final String UPDATES_PUBLISH_ENABLED = "metadata.updates.publish.enabled";
+    public static final String UPDATES_KAFKA_BROKER_LIST = "metadata.updates.kafka.broker.list";
   }
 }
--- a/cdap-common/src/main/resources/cdap-default.xml
+++ b/cdap-common/src/main/resources/cdap-default.xml
@@ -1553,14 +1553,20 @@
   </property>
 
   <property>
+    <name>metadata.updates.kafka.broker.list</name>
+    <value>127.0.0.1:${kafka.bind.port}</value>
+    <description>Kafka broker list to which metadata update notifications are published</description>
+  </property>
+
+  <property>
     <name>metadata.updates.kafka.topic</name>
-    <value>metadata-updates</value>
+    <value>cdap-metadata-updates</value>
     <description>Kafka topic to which metadata update notifications are published</description>
   </property>
 
   <property>
     <name>metadata.updates.publish.enabled</name>
-    <value>true</value>
+    <value>false</value>
     <description>
       Determines if metadata updates will be published to Apache Kafka. External systems can subscribe to the
       Kafka topic determined by ${metadata.updates.kafka.topic} to receive notifications of metadata updates.
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/publisher/KafkaMetadataChangePublisher.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/publisher/KafkaMetadataChangePublisher.java
@@ -22,20 +22,24 @@ import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.proto.Id;
 import co.cask.cdap.proto.codec.NamespacedIdCodec;
 import co.cask.cdap.proto.metadata.MetadataChangeRecord;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.inject.Inject;
+import kafka.javaapi.producer.Producer;
+import kafka.producer.KeyedMessage;
+import kafka.producer.ProducerConfig;
+import org.apache.twill.internal.kafka.client.ByteBufferEncoder;
+import org.apache.twill.internal.kafka.client.IntegerEncoder;
+import org.apache.twill.internal.kafka.client.IntegerPartitioner;
 import org.apache.twill.kafka.client.Compression;
-import org.apache.twill.kafka.client.KafkaClient;
 import org.apache.twill.kafka.client.KafkaPublisher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.nio.ByteBuffer;
-import java.util.concurrent.ExecutionException;
+import java.util.Properties;
 
 /**
  * A {@link MetadataChangePublisher} that publishes metadata changes to an Apache Kafka topic determined by the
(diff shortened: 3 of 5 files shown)


cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -1553,6 +1553,11 @@
   </property>
 
   <property>
+    <name>metadata.kafka.broker.list</name>
+    <value>127.0.0.1:${kafka.bind.port}</value>
+    <description>Kafka broker list to which metadata update notifications are published</description>
+  </property>
+  <property>
```
bdmogal: nit: add a newline before this

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/publisher/KafkaMetadataChangePublisher.java
```diff
@@ -47,38 +51,43 @@
...
 
   @Override
   public void publish(MetadataChangeRecord changeRecord) {
-    KafkaPublisher publisher;
-    try {
-      publisher = publishers.get(PUBLISHER_CACHE_KEY);
-    } catch (ExecutionException e) {
-      LOG.warn("Unable to get kafka publisher, will not be able to publish metadata.");
+    byte[] changesToPublish = Bytes.toBytes(GSON.toJson(changeRecord));
+    if (changesToPublish == null) {
```
bdmogal: I don't believe ``Bytes.toBytes`` can ever be ``null``. Also, ``changeRecord`` is never null either.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/publisher/KafkaMetadataChangePublisher.java
```diff
@@ -47,38 +51,43 @@
   private static final Gson GSON = new GsonBuilder()
     .registerTypeAdapter(Id.NamespacedId.class, new NamespacedIdCodec())
     .create();
-  private static final String PUBLISHER_CACHE_KEY = "kafkaPublisher";
+  private static final Compression DEFAULT_COMPRESSION = Compression.SNAPPY;
```
bdmogal: nit: can use ``Compression.SNAPPY`` directly

bdmogal:
@jwang47 some minor comments. However, please get the Kafka code reviewed from @poornachandra  as well.

poornachandra:
@jwang47 What happens when publishing is enabled, but `metadata.kafka.broker.list` is not set? We need to define the behavior for this case.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/publisher/KafkaMetadataChangePublisher.java
```diff
@@ -47,38 +51,38 @@
...
-      publisher = publishers.get(PUBLISHER_CACHE_KEY);
-    } catch (ExecutionException e) {
-      LOG.warn("Unable to get kafka publisher, will not be able to publish metadata.");
-      return;
-    }
     byte[] changesToPublish = Bytes.toBytes(GSON.toJson(changeRecord));
-    KafkaPublisher.Preparer preparer = publisher.prepare(topic);
-    preparer.add(ByteBuffer.wrap(changesToPublish), changeRecord.getPrevious().getTargetId().hashCode());
-    preparer.send();
+    Object partitionKey = changeRecord.getPrevious().getTargetId().hashCode();
```
poornachandra: Since we are doing a hashcode later, this can just be `Object partitionKey = changeRecord.getPrevious().getTargetId()`, right?

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -1553,6 +1553,12 @@
...
 
   <property>
+    <name>metadata.kafka.broker.list</name>
+    <value>127.0.0.1:${kafka.bind.port}</value>
+    <description>Kafka broker list to which metadata update notifications are published</description>
+  </property>
+
+  <property>
     <name>metadata.updates.kafka.topic</name>
     <value>metadata-updates</value>
```
poornachandra: Since we are publishing to external kafka now, it would be better to prefix the topic name with `cdap-`.

poornachandra:
@jwang47 Done with my review, the changes look good. The only concern is about behavior when kafka broker list is not set, or when external kafka server is down. We need to define the behavior and document it.

jwang47:
@poornachandra If `metadata.kafka.broker.list` is not set, then the default value from cdap-default.xml will be used. For when the external kafka server is down, I'll make a change to log an error.

poornachandra:
Also, I think since we are now dependent on an external system to be up for publishing, we should make the publishing of notifications false by default. The user can set the kafka broker list and enable the publish setting to be true.

jwang47:
Sounds good.

anew:
Agreed that it should be off by default

cdap-common/src/main/java/co/cask/cdap/common/conf/Constants.java
```diff
@@ -883,5 +883,6 @@
     public static final String HANDLERS_NAME = "metadata.handlers";
     public static final String UPDATES_KAFKA_TOPIC = "metadata.updates.kafka.topic";
     public static final String UPDATES_PUBLISH_ENABLED = "metadata.updates.publish.enabled";
+    public static final String KAFKA_BROKER_LIST = "metadata.updates.kafka.broker.list";
```
bdmogal: nit: this should be ``UPDATES_KAFKA_BROKER_LIST`` to be consistent

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -1553,14 +1553,20 @@
   </property>
 
   <property>
+    <name>metadata.updates.kafka.broker.list</name>
+    <value></value>
```
bdmogal: nit: is it a good idea to leave this empty? My intellij complains that xml cannot have tags with empty values. Having a default value will not hurt, no?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/publisher/KafkaMetadataChangePublisher.java
```diff
@@ -47,38 +51,46 @@
...
-  // However, using it will take care of synchronization issues while creating a publisher.
-  private final LoadingCache<String, KafkaPublisher> publishers;
+  private final Supplier<Producer<Integer, ByteBuffer>> producer;
   private final String topic;
+  private final String brokerList;
 
   @Inject
-  KafkaMetadataChangePublisher(CConfiguration cConf, final KafkaClient kafkaClient) {
+  KafkaMetadataChangePublisher(CConfiguration cConf) {
+    LOG.info("REMOVE: created kafka md publisher");
```
bdmogal: remove before merge

cdap-data-fabric/src/test/java/co/cask/cdap/data2/metadata/publisher/MetadataKafkaTestBase.java
```diff
@@ -82,12 +82,16 @@
...
+    Preconditions.checkState(kafkaPort > 0, "Failed to get random port.");
     int zkServerPort = Networks.getRandomPort();
     zkServer = InMemoryZKServer.builder().setDataDir(TMP_FOLDER.newFolder()).setPort(zkServerPort).build();
     zkServer.startAndWait();
-    kafkaServer = new EmbeddedKafkaServer(generateKafkaConfig());
+    kafkaServer = new EmbeddedKafkaServer(generateKafkaConfig(kafkaPort));
     kafkaServer.startAndWait();
     cConf = CConfiguration.create();
+    cConf.set(Constants.Metadata.UPDATES_PUBLISH_ENABLED, "true");
+    cConf.set(Constants.Metadata.KAFKA_BROKER_LIST, "127.0.0.1:" + kafkaPort);
```
bdmogal: can use ``InetAddress.getLoopbackAddress().getHostAddress()`` instead of hard coding

bdmogal:
LGTM. Merge after a successful build and testing on cluster.

PR #3962, opened 2015-09-16
CDAP-3657: Improve Id serialization

jwang47:
http://builds.cask.co/browse/CDAP-RBT423

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
@@ -436,20 +436,20 @@ public interface Store {
   /**
    * Updates the {@link WorkflowToken} for a specified run of a workflow.
    *
-   * @param workflowId {@link Id.Workflow} of the workflow whose {@link WorkflowToken} is to be updated
+   * @param workflowId {@link Id.Program} of the workflow whose {@link WorkflowToken} is to be updated
    * @param workflowRunId Run Id of the workflow for which the {@link WorkflowToken} is to be updated
    * @param token the {@link WorkflowToken} to update
    */
-  void updateWorkflowToken(Id.Workflow workflowId, String workflowRunId, WorkflowToken token);
+  void updateWorkflowToken(Id.Program workflowId, String workflowRunId, WorkflowToken token);
 
   /**
    * Retrieves the {@link WorkflowToken} for a specified run of a workflow.
    *
-   * @param workflowId {@link Id.Workflow} of the workflow whose {@link WorkflowToken} is to be retrieved
+   * @param workflowId {@link Id.Program} of the workflow whose {@link WorkflowToken} is to be retrieved
    * @param workflowRunId Run Id of the workflow for which the {@link WorkflowToken} is to be retrieved
    * @return the {@link WorkflowToken} for the specified workflow run
    */
-  WorkflowToken getWorkflowToken(Id.Workflow workflowId, String workflowRunId);
+  WorkflowToken getWorkflowToken(Id.Program workflowId, String workflowRunId);
 
   /**
    * Used by {@link co.cask.cdap.gateway.handlers.WorkflowStatsSLAHttpHandler} to get the statistics of all completed
@@ -461,7 +461,7 @@ public interface Store {
    * @param percentiles List of percentiles that the user wants to see
    * @return the statistics for a given workflow
    */
-  WorkflowStatistics getWorkflowStatistics(Id.Workflow workflowId, long startTime,
+  WorkflowStatistics getWorkflowStatistics(Id.Program workflowId, long startTime,
                                            long endTime, List<Double> percentiles);
 
   /**
@@ -471,7 +471,7 @@ public interface Store {
    * @param runId RunId of the workflow run
    * @return A workflow run record corresponding to the runId
    */
-  WorkflowDataset.WorkflowRunRecord getWorkflowRun(Id.Workflow workflowId, String runId);
+  WorkflowDataset.WorkflowRunRecord getWorkflowRun(Id.Program workflowId, String runId);
 
   /**
    * Get a list of workflow runs that are spaced apart by time interval in both directions from the run id provided.
@@ -482,8 +482,8 @@ public interface Store {
    * @param timeInterval The timeInterval with which the user wants to space out the runs
    * @return Map of runId of Workflow to DetailedStatistics of the run
    */
-  Collection<WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(Id.Workflow workflow, String runId,
-                                                                      int limit, long timeInterval);
+  Collection<WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(Id.Program workflow, String runId,
+                                                                       int limit, long timeInterval);
 
   /**
    * @return programs that were running between given start and end time.
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
@@ -67,13 +67,13 @@ public class DefaultStreamWriter implements StreamWriter {
   /**
    * The owners of this {@link StreamWriter}.
    */
-  private final List<Id> owners;
+  private final List<? extends Id.NamespacedId> owners;
   private final Id.Run run;
   private final LineageWriter lineageWriter;
 
   @Inject
   public DefaultStreamWriter(@Assisted("run") Id.Run run,
-                             @Assisted("owners") List<Id> owners,
+                             @Assisted("owners") List<? extends Id.NamespacedId> owners,
                              UsageRegistry usageRegistry,
                              LineageWriter lineageWriter,
                              DiscoveryServiceClient discoveryServiceClient) {
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/ConsumerSupplier.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/ConsumerSupplier.java
@@ -29,7 +29,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.List;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
@@ -47,24 +46,27 @@ final class ConsumerSupplier<T> implements Supplier<T>, Closeable {
   private final int numGroups;
   private final UsageRegistry usageRegistry;
   private final Id.Namespace namespace;
-  private final List<Id> owners;
+  private final Iterable<? extends Id.NamespacedId> owners;
   private ConsumerConfig consumerConfig;
   private Closeable consumer;
 
-  static <T> ConsumerSupplier<T> create(Id.Namespace namespace, List<Id> owners, UsageRegistry usageRegistry,
+  static <T> ConsumerSupplier<T> create(Id.Namespace namespace, Iterable<? extends Id.NamespacedId> owners,
+                                        UsageRegistry usageRegistry,
                                         DataFabricFacade dataFabricFacade,
                                         QueueName queueName, ConsumerConfig consumerConfig) {
     return create(namespace, owners, usageRegistry, dataFabricFacade, queueName, consumerConfig, -1);
   }
 
-  static <T> ConsumerSupplier<T> create(Id.Namespace namespace, List<Id> owners, UsageRegistry usageRegistry,
+  static <T> ConsumerSupplier<T> create(Id.Namespace namespace, Iterable<? extends Id.NamespacedId> owners,
+                                        UsageRegistry usageRegistry,
                                         DataFabricFacade dataFabricFacade, QueueName queueName,
                                         ConsumerConfig consumerConfig, int numGroups) {
     return new ConsumerSupplier<>(namespace, owners, usageRegistry, dataFabricFacade,
                                    queueName, consumerConfig, numGroups);
   }
 
-  private ConsumerSupplier(Id.Namespace namespace, List<Id> owners, UsageRegistry usageRegistry,
+  private ConsumerSupplier(Id.Namespace namespace, Iterable<? extends Id.NamespacedId> owners,
+                           UsageRegistry usageRegistry,
                            DataFabricFacade dataFabricFacade, QueueName queueName,
                            ConsumerConfig consumerConfig, int numGroups) {
     this.namespace = namespace;
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
@@ -125,8 +125,10 @@ public class ProgramLifecycleHttpHandlerTest extends AppFabricTestBase {
     HttpResponse response = deploy(WordCountApp.class, Constants.Gateway.API_VERSION_3_TOKEN, TEST_NAMESPACE1);
     Assert.assertEquals(200, response.getStatusLine().getStatusCode());
 
-    Id.Flow wordcountFlow1 = Id.Flow.from(TEST_NAMESPACE1, WORDCOUNT_APP_NAME, WORDCOUNT_FLOW_NAME);
-    Id.Flow wordcountFlow2 = Id.Flow.from(TEST_NAMESPACE2, WORDCOUNT_APP_NAME, WORDCOUNT_FLOW_NAME);
+    Id.Program wordcountFlow1 = Id.Program.from(TEST_NAMESPACE1, WORDCOUNT_APP_NAME,
+                                                ProgramType.FLOW, WORDCOUNT_FLOW_NAME);
+    Id.Program wordcountFlow2 = Id.Program.from(TEST_NAMESPACE2, WORDCOUNT_APP_NAME,
+                                                ProgramType.FLOW, WORDCOUNT_FLOW_NAME);
 
     // flow is stopped initially
     Assert.assertEquals(STOPPED, getProgramStatus(wordcountFlow1));
@@ -392,8 +394,11 @@ public class ProgramLifecycleHttpHandlerTest extends AppFabricTestBase {
     Assert.assertEquals(new NotFoundException(Id.Program.from("testnamespace1", "WordCountApp", ProgramType.FLOW,
                                                               "NotExist")).getMessage(),
                         returnedBody.get(0).get("error").getAsString());
-    Assert.assertEquals("'namespace:testnamespace1/application:WordCountApp/program:flows:NotExist' was not found.",
-                        returnedBody.get(0).get("error").getAsString());
+    // namespace:testnamespace1/application:WordCountApp/program:flows:NotExist
+    Assert.assertEquals(
+      String.format("'%s' was not found.",
+                    Id.Program.from(TEST_NAMESPACE1, "WordCountApp", ProgramType.FLOW, "NotExist").toString()),
+      returnedBody.get(0).get("error").getAsString());
     // The programType should be consistent. Second object should have proper status
     Assert.assertEquals("Flow", returnedBody.get(1).get("programType").getAsString());
     Assert.assertEquals(STOPPED, returnedBody.get(1).get("status").getAsString());
@@ -781,9 +786,11 @@ public class ProgramLifecycleHttpHandlerTest extends AppFabricTestBase {
     HttpResponse response = deploy(AppWithServices.class, Constants.Gateway.API_VERSION_3_TOKEN, TEST_NAMESPACE2);
     Assert.assertEquals(200, response.getStatusLine().getStatusCode());
 
-    Id.Service service1 = Id.Service.from(Id.Namespace.from(TEST_NAMESPACE1), APP_WITH_SERVICES_APP_ID,
+    Id.Program service1 = Id.Program.from(Id.Namespace.from(TEST_NAMESPACE1), APP_WITH_SERVICES_APP_ID,
+                                          ProgramType.SERVICE,
                                           APP_WITH_SERVICES_SERVICE_NAME);
-    Id.Service service2 = Id.Service.from(Id.Namespace.from(TEST_NAMESPACE2), APP_WITH_SERVICES_APP_ID,
+    Id.Program service2 = Id.Program.from(Id.Namespace.from(TEST_NAMESPACE2), APP_WITH_SERVICES_APP_ID,
+                                          ProgramType.SERVICE,
                                           APP_WITH_SERVICES_SERVICE_NAME);
 
     // start service in wrong namespace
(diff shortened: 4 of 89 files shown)


bdmogal:
@jwang47 not reviewed the PR, but we should test upgrade thoroughly before merging this. Will this impact serialization of any ``Id`` objects in CDAP meta tables - ``app.meta``, ``usage.registry``, ``dataset.instance``, ``dataset.type``, etc?

poornachandra:
It might be a good idea to add a comment on top of `toString()` methods saying that they are part of external contract, and cannot be changed without checking for backwards compatibility.

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
```diff
@@ -59,33 +53,19 @@ private static boolean isValidDatasetId(String datasetId) {
...
-  public String toString() {
-    return getIdRep();
+  public String getIdType() {
+    return getIdType(getClass());
   }
 
-  @Nullable
-  protected abstract Id getParent();
-
-  public abstract String getId();
```
poornachandra: Any reason for the removal of this method? I should be able to get an Id from an Id object, right? This breaks compilation of LineageRecord class.

jwang47: This method doesn't make sense for all `Id`s, so I removed it. However, I can simply make `toString()` return e.g. `program:namespace1.app1.flow.flow1` and `getId()` return `namespace1.app1.flow.flow1`

jwang47:
@bdmogal The only places we're using `Id.toString()` is in `DatasetServiceClient`/`DatasetInstanceHandler` (internal API) and lineage, so there won't be any upgrade step needed.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
```diff
@@ -482,7 +482,7 @@ WorkflowStatistics getWorkflowStatistics(Id.Workflow workflowId, long startTime,
    * @param timeInterval The timeInterval with which the user wants to space out the runs
    * @return Map of runId of Workflow to DetailedStatistics of the run
    */
-  Collection<WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(Id.Workflow workflow, String runId,
+  Collection<WorkflowDataset.WorkflowRunRecord> retrieveSpacedRecords(Id.Program workflow, String runId,
```
bdmogal: why do we need these changes?

jwang47: Only `Id.Program` will be used from now on, rather than both `Id.Program` and `Id.Workflow`, etc. `Id.Workflow` will no longer be instantiable and will only contain helper methods.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/ConsumerSupplier.java
```diff
@@ -47,24 +46,27 @@
   private final int numGroups;
   private final UsageRegistry usageRegistry;
   private final Id.Namespace namespace;
-  private final List<Id> owners;
+  private final Iterable<? extends Id.NamespacedId> owners;
```
bdmogal: why does the list change to Iterable here?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -392,8 +392,10 @@ public void testBatchStatus() throws Exception {
     Assert.assertEquals(new NotFoundException(Id.Program.from("testnamespace1", "WordCountApp", ProgramType.FLOW,
                                                               "NotExist")).getMessage(),
                         returnedBody.get(0).get("error").getAsString());
-    Assert.assertEquals("'namespace:testnamespace1/application:WordCountApp/program:flows:NotExist' was not found.",
-                        returnedBody.get(0).get("error").getAsString());
+    // namespace:testnamespace1/application:WordCountApp/program:flows:NotExist
+    Assert.assertEquals(
+      String.format("'%s' was not found.", Id.Flow.from(TEST_NAMESPACE1, "WordCountApp", "NotExist").toString()),
```
bdmogal: can just use new ``new NotFoundException(Id.Flow.from(TEST_NAMESPACE1, "WordCountApp", "NotExist")).getMessage()`` here, so the message becomes internal to the Exception class?

bdmogal: Please change other instances in this file that have a similar pattern

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetServiceClient.java
```diff
@@ -84,16 +89,18 @@ public EndpointStrategy get() {
   }
 
   @Nullable
-  public DatasetMeta getInstance(String instanceName, @Nullable Iterable<? extends Id> owners)
-    throws DatasetManagementException {
+  public DatasetMeta getInstance(String instanceName,
+                                 @Nullable Iterable<? extends Id> owners)
```
bdmogal: this can also be ``Iterable<? extends Id.NamespacedId>``?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -199,7 +200,8 @@ public void deleteAllInstances(Id.Namespace namespaceId) throws DatasetManagemen
   public <T extends Dataset> T getDataset(
     Id.DatasetInstance datasetInstanceId, Map<String, String> arguments,
     @Nullable ClassLoader classLoader,
-    @Nullable Iterable<? extends Id> owners) throws DatasetManagementException, IOException {
+    @Nullable Iterable<? extends Id> owners)
```
bdmogal: ``Iterable<? extends Id.NamespacedId>``

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -199,7 +200,8 @@ public void deleteAllInstances(Id.Namespace namespaceId) throws DatasetManagemen
...
+    throws DatasetManagementException, IOException, ServiceUnavailableException {
```
bdmogal: ``ServiceUnavailableException`` is unchecked. It just got removed in another PR. You may have conflicts.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -218,7 +220,8 @@ public void deleteAllInstances(Id.Namespace namespaceId) throws DatasetManagemen
     Id.DatasetInstance datasetInstanceId, @Nullable Map<String, String> arguments,
     ClassLoader classLoader,
     DatasetClassLoaderProvider classLoaderProvider,
-    @Nullable Iterable<? extends Id> owners) throws DatasetManagementException, IOException {
+    @Nullable Iterable<? extends Id> owners) throws DatasetManagementException, IOException,
+    ServiceUnavailableException {
```
bdmogal: please remove ``ServiceUnavailableException`` if it exists after rebasing with release/3.2

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -218,7 +220,8 @@ public void deleteAllInstances(Id.Namespace namespaceId) throws DatasetManagemen
     Id.DatasetInstance datasetInstanceId, @Nullable Map<String, String> arguments,
     ClassLoader classLoader,
     DatasetClassLoaderProvider classLoaderProvider,
-    @Nullable Iterable<? extends Id> owners) throws DatasetManagementException, IOException {
+    @Nullable Iterable<? extends Id> owners) throws DatasetManagementException, IOException,
```
bdmogal: ``Iterable<? extends Id.NamespacedId>``

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
```diff
@@ -214,26 +214,11 @@ public void executeDataOp(HttpRequest request, HttpResponder responder, @PathPar
...
-        String ownerType = parts[0];
-        String ownerId = parts[1];
-        if (ownerType.equals(Id.getType(Id.Program.class))) {
-          return Id.Program.fromStrings(ownerId.split("/"));
-        } else {
-          return null;
-        }
-      }
-    });
+  private List<Id.NamespacedId> decodeOwners(String owners) throws UnsupportedEncodingException {
```
bdmogal: Is owners ``@Nullable``? If so, mark it. If not, remove the check below.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DatasetFramework.java
```diff
@@ -239,7 +239,7 @@ void updateInstance(Id.DatasetInstance datasetInstanceId, DatasetProperties prop
   @Nullable
   <T extends Dataset> T getDataset(Id.DatasetInstance datasetInstanceId, @Nullable Map<String, String> arguments,
                                    @Nullable ClassLoader classLoader, @Nullable Iterable<? extends Id> owners)
-    throws DatasetManagementException, IOException;
+    throws DatasetManagementException, IOException, ServiceUnavailableException;
```
bdmogal: remove ServiceUnavailableException

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DatasetFramework.java
```diff
@@ -278,7 +278,7 @@ void updateInstance(Id.DatasetInstance datasetInstanceId, DatasetProperties prop
                                    @Nullable ClassLoader classLoader,
                                    DatasetClassLoaderProvider classLoaderProvider,
                                    @Nullable Iterable<? extends Id> owners)
-    throws DatasetManagementException, IOException;
+    throws DatasetManagementException, IOException, ServiceUnavailableException;
```
bdmogal: ditto

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/LineageWriterDatasetFramework.java
```diff
@@ -159,7 +160,7 @@ public void deleteAllInstances(Id.Namespace namespaceId) throws DatasetManagemen
   public <T extends Dataset> T getDataset(Id.DatasetInstance datasetInstanceId,
                                           @Nullable Map<String, String> arguments, @Nullable ClassLoader classLoader,
                                           @Nullable Iterable<? extends Id> owners)
-    throws DatasetManagementException, IOException {
+    throws DatasetManagementException, IOException, ServiceUnavailableException {
```
bdmogal: ditto

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/LineageWriterDatasetFramework.java
```diff
@@ -181,7 +182,7 @@ public void deleteAllInstances(Id.Namespace namespaceId) throws DatasetManagemen
                                           @Nullable Map<String, String> arguments, @Nullable ClassLoader classLoader,
                                           DatasetClassLoaderProvider classLoaderProvider,
                                           @Nullable Iterable<? extends Id> owners)
-    throws DatasetManagementException, IOException {
+    throws DatasetManagementException, IOException, ServiceUnavailableException {
```
bdmogal: ditto

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
```diff
@@ -329,7 +279,42 @@ public Program(Application application, ProgramType type, final String id) {
...
+    public static Program from(Id.Namespace namespaceId, String appId, ProgramType type, String pgmId) {
+      return new Program(new Application(namespaceId, appId), type, pgmId);
+    }
+
+    public static Program from(String namespaceId, String appId, ProgramType type, String pgmId) {
+      return new Program(new Application(new Namespace(namespaceId), appId), type, pgmId);
+    }
+
+    public static Program fromString(String string) {
+      String[] parts = string.split(":", 2);
```
bdmogal: always scared of these string to object decodings. Where is this used?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
poornachandra: Asserting on message is always brittle. It might be enough to just assert on the exception class itself, right?

cdap-proto/src/test/java/co/cask/cdap/proto/codec/NamespacedIdCodecTest.java
```diff
@@ -40,12 +40,12 @@
   private final Id.Namespace ns = Id.Namespace.from("ns");
   private final Id.Application app = Id.Application.from(ns, "app");
   private final Id.Program program = Id.Program.from(app, ProgramType.CUSTOM_ACTION, "action");
-  private final Id.Flow flow = Id.Flow.from(app, "flow");
```
bdmogal: This pattern is consistent across this PR. So can't users ever do ``Id.Flow flow = Id.Flow.from(app, "flow")`` now? Does the type always have to be declared as Program? I think that's because ``Id.Flow`` and others don't extend ``Id.Program``. What was the reason for that change?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/flow/ConsumerSupplier.java
jwang47: `List` is not needed, only `Iterable`.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
bdmogal: Should we make it simple and then just remove ``Id.Workflow`` and other classes? It seems weird to me (probably because I've now gotten used to it :-) ), that ``Id.Workflow``, etc does not extend ``Id.Program``. If they are supposed to be helper classes, lets move them outside ``Id`` itself, into a helper ``Programs`` class. But even that is an overkill for saving on just a single ``ProgramType`` parameter.

cdap-proto/src/test/java/co/cask/cdap/proto/codec/NamespacedIdCodecTest.java
jwang47: Currently it's possible to have an `Id.Program` and `Id.Flow` that represents the same thing, but  it's better to only have `Id.Program` since it's less confusing.

cdap-app-fabric/src/main/java/co/cask/cdap/app/store/Store.java
jwang47: True, we can simply remove them.

bdmogal:
@jwang47 some comments. Major concern is the helper classes, which IMO are not adding any significant value, now that they do not extend ``Id.Program``.

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
jwang47: Oops. It can be removed now.

bdmogal:
@jwang47, sorry, I thought about this a little more. Aren't our proto classes equivalent to public APIs? Is there a way we can do this while keeping the classes?

cdap-app-fabric/src/main/java/co/cask/cdap/app/stream/DefaultStreamWriter.java
```diff
@@ -67,13 +67,13 @@
...
    */
-  private final List<Id> owners;
+  private final List<? extends Id.NamespacedId> owners;
   private final Id.Run run;
   private final LineageWriter lineageWriter;
 
   @Inject
   public DefaultStreamWriter(@Assisted("run") Id.Run run,
-                             @Assisted("owners") List<Id> owners,
+                             @Assisted("owners") List<? extends Id.NamespacedId> owners,
```
anew: Isn't this Id.Program?

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -392,8 +394,11 @@ public void testBatchStatus() throws Exception {
     Assert.assertEquals(new NotFoundException(Id.Program.from("testnamespace1", "WordCountApp", ProgramType.FLOW,
                                                               "NotExist")).getMessage(),
                         returnedBody.get(0).get("error").getAsString());
-    Assert.assertEquals("'namespace:testnamespace1/application:WordCountApp/program:flows:NotExist' was not found.",
-                        returnedBody.get(0).get("error").getAsString());
+    // namespace:testnamespace1/application:WordCountApp/program:flows:NotExist
+    Assert.assertEquals(
```
anew: Better define the Id as a variable and use it twice

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/ProgramLifecycleHttpHandlerTest.java
```diff
@@ -392,8 +394,11 @@ public void testBatchStatus() throws Exception {
...
+      String.format("'%s' was not found.",
+                    Id.Program.from(TEST_NAMESPACE1, "WordCountApp", ProgramType.FLOW, "NotExist").toString()),
+      returnedBody.get(0).get("error").getAsString());
```
anew: indentation is off

anew:
This is a huge change, so late before release? It is not clear to me what bug it fixes?

anew:
Actually, I could see convenience methods Id.<ProgramType>.from(...) that return an Id.Program. That would also have kept the change a lot smaller.

anew:
If you make this huge change, why not rename the getId() method to getName() at the same time? A id is a composition of names. It is quite confusing to say that programId.getId() returns the name of the program.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetServiceClient.java
```diff
@@ -84,16 +89,18 @@ public EndpointStrategy get() {
...
+                                 @Nullable Iterable<? extends Id.NamespacedId> owners)
+    throws DatasetManagementException, ServiceUnavailableException {
 
     String query = "";
     if (owners != null) {
-      Set<String> ownerParams = Sets.newHashSet();
-      for (Id owner : owners) {
-        ownerParams.add("owner=" + owner.getIdType() + "::" + owner.getIdRep());
+      try {
+        String ownersJson = GSON.toJson(owners, new TypeToken<List<Id.NamespacedId>>() { }.getType());
```
anew: is everything in that JSON a valid URL character?

chtyim: It got URL encoded in the next line.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
```diff
@@ -97,8 +99,7 @@ public void list(HttpRequest request, HttpResponder responder,
    *
    * @param namespaceId namespace of the dataset instance
    * @param name name of the dataset instance
-   * @param owners a list of owners of the dataset instance, in the form @{code <type>::<id>}
-   *               (e.g. "program::namespace:default/application:PurchaseHistory/program:flow:PurchaseFlow")
+   * @param owners a UTF-8-encoded JSON list of Id owners of the dataset instance
```
anew: so we would have something like ?owners=[ns1.app1.flow1,ns1.app2.svc3] in the url? Normally, to specify multiple values for a parameter in an URL, I believe, you would use repetition, as in ?owner=ns1.app1.flow1@owner=ns1.app2.svc3

anew: Is this a public REST API? If so, we need to keep the old syntax and deprecate it

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/lineage/LineageDataset.java
```diff
@@ -415,9 +415,9 @@ private void addProgram(MDSKey.Builder keyBuilder, Id.Program program) {
   }
 
   private void addComponent(MDSKey.Builder keyBuilder, Id component) {
-    if (component instanceof Id.Flow.Flowlet) {
+    if (component instanceof Id.Flowlet) {
       keyBuilder.add(FLOWLET_MARKER)
-        .add(component.getId());
+        .add(((Id.Flowlet) component).getId());
```
anew: why is the cast needed?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/writer/LineageWriterDatasetFramework.java
```diff
@@ -180,8 +181,8 @@ public void deleteAllInstances(Id.Namespace namespaceId) throws DatasetManagemen
   public <T extends Dataset> T getDataset(Id.DatasetInstance datasetInstanceId,
                                           @Nullable Map<String, String> arguments, @Nullable ClassLoader classLoader,
                                           DatasetClassLoaderProvider classLoaderProvider,
-                                          @Nullable Iterable<? extends Id> owners)
-    throws DatasetManagementException, IOException {
+                                          @Nullable Iterable<? extends Id.NamespacedId> owners)
+    throws DatasetManagementException, IOException, ServiceUnavailableException {
```
anew: is this an unrelated change?

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
anew: getId is a getter and should simply return the value of the id field. Actually it would be better to rename id to name and then you'd have getName() return "flow1", getFullName() return "namespace1.app1.flow.flow1", and toString() returns  "program:namespace1.app1.flow.flow1"

cdap-proto/src/test/java/co/cask/cdap/proto/codec/NamespacedIdCodecTest.java
anew: and less convenient. Also less type-safe. Id.Program could be abstract to remove the concern that you can have an Id.Program and an Id.Flow representing the same thing.

anew:
I have done one pass. Please see my comments.

jwang47:
@anew @bdmogal Good points, I will rework this PR so that it doesn't modify any public API classes.

poornachandra:
@jwang47 Lets hold-off on this PR for release 3.2. We need some more discussion on this. We'll work on this for release 3.3.

bdmogal:
@jwang47 , i think for 3.2, we should probably just fix the problem mentioned in the jira (about making sure that all id classes have the same serialization pattern) Lets do refactoring later.

PR #3964, opened 2015-09-16
add missing route to streams lineage in RouterPathLookup

gsps1:
we were missing the route to streams lineage.

### DIFF
--- a/cdap-gateway/src/main/java/co/cask/cdap/gateway/router/RouterPathLookup.java
+++ b/cdap-gateway/src/main/java/co/cask/cdap/gateway/router/RouterPathLookup.java
@@ -88,6 +88,7 @@ public final class RouterPathLookup extends AbstractHttpHandler {
       matches(uriParts, "v3", "namespaces", null, "streams", null, "metadata", "tags") ||
       matches(uriParts, "v3", "namespaces", null, "metadata", "search") ||
       matches(uriParts, "v3", "namespaces", null, "datasets", null, "lineage") ||
+      matches(uriParts, "v3", "namespaces", null, "streams", null, "lineage") ||
       matches(uriParts, "v3", "namespaces", null, "apps", null, null, null, "runs", null, "metadata")) {
       return Constants.Service.METADATA_SERVICE;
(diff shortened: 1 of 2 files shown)


cdap-gateway/src/main/java/co/cask/cdap/gateway/router/RouterPathLookup.java
```diff
@@ -88,6 +88,7 @@ private String getV3RoutingService(String [] uriParts, AllowedMethod requestMeth
       matches(uriParts, "v3", "namespaces", null, "streams", null, "metadata", "tags") ||
       matches(uriParts, "v3", "namespaces", null, "metadata", "search") ||
       matches(uriParts, "v3", "namespaces", null, "datasets", null, "lineage") ||
+      matches(uriParts, "v3", "namespaces", null, "streams", null, "lineage") ||
```
bdmogal: add a verification for this in ``RouterPathTest.testMetadataRouting()`` as well.

gsps1: done

bdmogal:
LGTM. Merge after successful build.

gsps1:
Green, merging.

cdap-gateway/src/test/java/co/cask/cdap/gateway/router/RouterPathTest.java
```diff
@@ -355,6 +355,7 @@ public void testMetadataPath() {
     assertMetadataRouting("/v3/namespaces/default/metadata/search");
     // lineage
     assertMetadataRouting("/v3/namespaces/default/////datasets/ds1/lineage");
+    assertMetadataRouting("/v3/namespaces/default/streams/st1/lineage");
```
poornachandra: Does this test succeed? Looks like it is missing a few path elements.

gsps1: it does pass, which path elements are missing ?

PR #3966, opened 2015-09-16
CDAP-3618: Fix to not lose suspended schedules on master restart

rohitsinha54:
Issue: https://issues.cask.co/browse/CDAP-3618

- This fixes the bug in DatasetBasedTimeScheduleStore which while retrieving triggers back from the store used to retrieve triggers only in normal state and skip all others. We cannot skip schedules in suspended state (triggers in paused state) as this will cause them not be added to the scheduler while its being started back again. 
- The executePersist should always persist with its state. 
- After we read the schedules (triggers) from the persistence store we need to check its state which we have saved and if it was paused we need to pause it back. Since the scheduler has not actually been started at this point its fine to call pause like this as schedule cannot get fired before paused is called on it.
- Added test case to check schedule state of paused schedules across scheduler restarts.
- Also refactored the SchedulerTest (DatasetBasedTimeScheduleStoreTest)

Build: http://builds.cask.co/browse/CDAP-RBT428-2

- [x] Test on standalone with multiple restart of master
- [x] Test on singlenode with multiple restart of master
- [x] Test on HA cluster by killing random master and also all masters: HA Cluster creation failed. Will test tomorrow morning.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
@@ -81,13 +81,13 @@ public class DatasetBasedTimeScheduleStore extends RAMJobStore {
   @Override
   public void storeJob(JobDetail newJob, boolean replaceExisting) throws ObjectAlreadyExistsException {
     super.storeJob(newJob, replaceExisting);
-    executePersist(newJob, null, null);
+    persistJobAndTrigger(newJob, null);
   }
 
   @Override
   public void storeTrigger(OperableTrigger newTrigger, boolean replaceExisting) throws JobPersistenceException {
     super.storeTrigger(newTrigger, replaceExisting);
-    executePersist(null, newTrigger, super.getTriggerState(newTrigger.getKey()));
+    persistJobAndTrigger(null, newTrigger);
   }
 
 
@@ -104,19 +104,19 @@ public class DatasetBasedTimeScheduleStore extends RAMJobStore {
     super.storeJob(newJob, true);
     super.storeTrigger(newTrigger, true);
 
-    executePersist(newJob, newTrigger, super.getTriggerState(newTrigger.getKey()));
+    persistJobAndTrigger(newJob, newTrigger);
   }
 
   @Override
   public void pauseTrigger(TriggerKey triggerKey) {
     super.pauseTrigger(triggerKey);
-    executePersist(triggerKey, Trigger.TriggerState.PAUSED);
+    persistChangeOfState(triggerKey, Trigger.TriggerState.PAUSED);
   }
 
   @Override
   public void resumeTrigger(TriggerKey triggerKey) {
     super.resumeTrigger(triggerKey);
-    executePersist(triggerKey, Trigger.TriggerState.NORMAL);
+    persistChangeOfState(triggerKey, Trigger.TriggerState.NORMAL);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
```diff
@@ -104,7 +104,7 @@ public void storeJobAndTrigger(JobDetail newJob, OperableTrigger newTrigger)
     super.storeJob(newJob, true);
     super.storeTrigger(newTrigger, true);
 
-    executePersist(newJob, newTrigger);
+    executePersist(newJob, newTrigger, super.getTriggerState(newTrigger.getKey()));
```
anew: seems that this method can just call storeJob() and then storeTrigger()?

rohitsinha54: I think this comment is for storeJobsAndTriggers() and github is collapsing the lines.
Yes that is what I am planning to do. Thats how super.storeJobsAndTriggers() does too with some validation initially will use the same logic.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
```diff
@@ -170,16 +170,18 @@ public void apply() throws Exception {
     }
   }
 
-  private void executePersist(final TriggerKey triggerKey, final Trigger.TriggerState state) {
+  private void executePersist(final TriggerKey triggerKey, final Trigger.TriggerState newTriggerState) {
```
anew: Since we are already refactoring. This executePersist() methods saves a changes of state, whereas the other saves a new or changed definition of a job and trigger. Why not name them accordingly: persistChangeOfState() and persistJobAndTrigger()

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
```diff
@@ -81,13 +81,13 @@ public void initialize(ClassLoadHelper loadHelper, SchedulerSignaler schedSignal
...
     super.storeJob(newJob, replaceExisting);
-    executePersist(newJob, null);
+    executePersist(newJob, null, null);
   }
 
   @Override
   public void storeTrigger(OperableTrigger newTrigger, boolean replaceExisting) throws JobPersistenceException {
     super.storeTrigger(newTrigger, replaceExisting);
-    executePersist(null, newTrigger);
+    executePersist(null, newTrigger, super.getTriggerState(newTrigger.getKey()));
```
anew: Actually, looking at how this executePersist() is called, you can move the super.getTriggerState() into executePersist(). Because is it always called if the trigger is not null.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
```diff
@@ -170,16 +170,18 @@ public void apply() throws Exception {
...
       factory.createExecutor(ImmutableList.of((TransactionAware) table))
                             .execute(new TransactionExecutor.Subroutine() {
                               @Override
                               public void apply() throws Exception {
                                 if (triggerKey != null) {
-                                  TriggerStatus trigger = readTrigger(triggerKey);
-                                  if (trigger != null) {
-                                    persistTrigger(table, trigger.trigger, state);
+                                  TriggerStatus storedTriggerStatus = readTrigger(triggerKey);
+                                  if (storedTriggerStatus != null) {
```
anew: why do we only persist if there was an existing status? When would this be called such that there is no existing status?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
```diff
@@ -170,16 +170,18 @@ public void apply() throws Exception {
...
+                                    // its okay to persist the same trigger back again since during pause/resume
```
anew: this comment is confusing. You are not storing the same trigger again, You are storing with a new state.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
rohitsinha54: This executePersist for triggers is only called from pause and resume. So the TriggerStatus(our wrapper on trigger with its status) has to be present while we perform pause or resume to it. It will be persisted initially when the trigger is added.

anew: So, if this is called with a null triggerKey, that would be an internal error? Shouldn't we throw an exception then?

And if it is called for a non-existent trigger, also? Or at least log a warning, because we do not expect this to happen.

anew:
@rohitsinha54 Overall the change looks good, but I had a few suggestions to make code simpler/easier to understand.

rohitsinha54:
New Build: http://builds.cask.co/browse/CDAP-RBT428-3

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
anew: I don't see this change in the latest commit.

anew:
Seems like one of my earlier comments was not addressed. Other than that LGTM.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
rohitsinha54: Oh I thought you made this comment on some other line (Line: 95 storeJobsAndTriggers method which is not implemented yet and github is wrapping lines). Something similar to what you mentioned here needs to be done to implement that too.
But looks like your comment was on this line itself. Sorry for this misunderstanding. I will change it.

rohitsinha54:
@anew : While testing on cluster I noticed a bug which got introduced during addressing comments. I have fixed it in this comment: https://github.com/caskdata/cdap/commit/29adf69b4886dd0de9d0db79be55d857ea2a9249

Please have a look. Thanks.

anew:
LGTM

PR #3967, opened 2015-09-16
CDAP-3326 row key should not be required for table source

albertshau:
Fixes a bug where the row key was required for table source.
Row key is only required for table sink when we write. For a
source it is ok if we never read anything from the rowkey.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/TableSourceConfig.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/TableSourceConfig.java
@@ -28,7 +28,7 @@ import javax.annotation.Nullable;
 /**
  * {@link PluginConfig} for {@link TableSource}, {@link TableSink} and {@link RealtimeTableSink}
  */
-public class TableConfig extends PluginConfig {
+public class TableSourceConfig extends PluginConfig {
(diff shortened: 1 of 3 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-RBT429

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/TableSourceConfig.java
```diff
@@ -41,6 +41,7 @@
 
   @Name(Properties.Table.PROPERTY_SCHEMA_ROW_FIELD)
   @Description("The name of the record field that should be used as the row key when writing to the table.")
```
gokulavasan: I think the description should change to 'Field name that for the Table's row key' ? If this property is not provided, will we create the structured record only from the columns of the row?

gokulavasan:
One question/comment.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/template/etl/common/TableSourceConfig.java
albertshau: ah yes you are right, will change.

albertshau:
only change since green build was a doc change. merging.

PR #3969, opened 2015-09-16
[CDAP-3675] Metadata change publishing should only be required in Dat…

bdmogal:
…asetOpExecutorServerTwillRunnable. Bind a NoOpPublisher by default.

Jira: [CDAP-3675](https://issues.cask.co/browse/CDAP-3675)
Build: http://builds.cask.co/browse/CDAP-RBT426


poornachandra:
LGTM

bdmogal:
Unit test Build passed. All but two integration tests pass (1 failure was flaky, the other a known failure because of an older spark version). Merging.

PR #3970, opened 2015-09-16
CDAP-3472 Correct error messsage when trying to stop the MapReduce pr…

sagarkapare:
…ogram started by Workflow.

JIRA: https://issues.cask.co/browse/CDAP-3472

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -1361,8 +1361,9 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
         if (runRecord != null && runRecord.getProperties().containsKey("workflowrunid")
           && runRecord.getStatus().equals(ProgramRunStatus.RUNNING)) {
           String workflowRunId = runRecord.getProperties().get("workflowrunid");
-          throw new BadRequestException(String.format("Cannot stop the program '%s' started by the Workflow run '%s'. " +
-                                                        "Please stop the Workflow.", programRunId, workflowRunId));
+          throw new BadRequestException(String.format("Cannot stop the program '%s' started by the Workflow " +
+                                                        "run '%s'. Please stop the Workflow.", programRunId,
+                                                      workflowRunId));
         }
         throw new NotFoundException(programRunId);
       }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -1357,13 +1357,14 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
       } else if (runId != null) {
         // Check if the program is running and is started by the Workflow
         RunRecordMeta runRecord = store.getRun(identifier, runId);
-        if (runRecord.getProperties().containsKey("workflowrunid")
+        if (runRecord != null && runRecord.getProperties().containsKey("workflowrunid")
           && runRecord.getStatus().equals(ProgramRunStatus.RUNNING)) {
           throw new BadRequestException("MapReduce program run started by Workflow cannot be stopped. " +
                                           "Please stop the Workflow.");
         }
+        throw new NotFoundException(new Id.Run(identifier, runId));
       }
-      throw new NotFoundException(new Id.Run(identifier, runId));
+      throw new BadRequestException(String.format("Program '%s' is not running.", identifier));
     }
 
     try {
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -1355,14 +1355,16 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
       } else if (!store.programExists(identifier)) {
         throw new ProgramNotFoundException(identifier);
       } else if (runId != null) {
+        Id.Run programRunId = new Id.Run(identifier, runId);
         // Check if the program is running and is started by the Workflow
         RunRecordMeta runRecord = store.getRun(identifier, runId);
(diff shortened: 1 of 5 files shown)


sagarkapare:
Release build is running here - http://builds.cask.co/browse/CDAP-RBT430

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1354,11 +1354,16 @@ private AppFabricServiceStatus stop(Id.Program identifier, @Nullable String runI
...
+      } else if (runId != null) {
+        // Check if the program is running and is started by the Workflow
+        RunRecordMeta runRecord = store.getRun(identifier, runId);
+        if (runRecord.getProperties().containsKey("workflowrunid")
+          && runRecord.getStatus().equals(ProgramRunStatus.RUNNING)) {
+          throw new BadRequestException("MapReduce program run started by Workflow cannot be stopped. " +
+                                          "Please stop the Workflow.");
+        }
       }
+      throw new NotFoundException(new Id.Run(identifier, runId));
```
anwar6953: runId might be null in this case, right?

anwar6953: Also, you're not properly responded with BadRequestException("Program not running") like before. Isn't that still a possibility?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1354,11 +1354,16 @@ private AppFabricServiceStatus stop(Id.Program identifier, @Nullable String runI
...
-      } else if (runId == null) {
-        throw new BadRequestException("Program not running");
-      } else {
-        throw new NotFoundException(new Id.Run(identifier, runId));
+      } else if (runId != null) {
+        // Check if the program is running and is started by the Workflow
+        RunRecordMeta runRecord = store.getRun(identifier, runId);
+        if (runRecord.getProperties().containsKey("workflowrunid")
+          && runRecord.getStatus().equals(ProgramRunStatus.RUNNING)) {
+          throw new BadRequestException("MapReduce program run started by Workflow cannot be stopped. " +
```
bdmogal: should we add the mapreduce and workflow ids in the message?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1354,11 +1354,16 @@ private AppFabricServiceStatus stop(Id.Program identifier, @Nullable String runI
...
       } else if (!store.programExists(identifier)) {
         throw new ProgramNotFoundException(identifier);
-      } else if (runId == null) {
-        throw new BadRequestException("Program not running");
-      } else {
-        throw new NotFoundException(new Id.Run(identifier, runId));
+      } else if (runId != null) {
+        // Check if the program is running and is started by the Workflow
+        RunRecordMeta runRecord = store.getRun(identifier, runId);
+        if (runRecord.getProperties().containsKey("workflowrunid")
```
bdmogal: what is the expected behavior if this condition is false?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
bdmogal: yeah something seems to be missing here. perhaps this should move one line above. and there should be a ``BadRequestException(String.format("Program %s not running")`` on this line. I'm not entirely sure, though.

sagarkapare: Fixed the code to throw BadRequestException("Program not running") when the run id is null.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: It should actually throw NotFoundException when the condition is false. Fixed.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
sagarkapare: I think currently its not possible to get the program name from runid.

sagarkapare:
@bdmogal Addressed comments. Please take a look again.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
anwar6953: We have access to a variable named 'identifier' which contains this information.

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -1354,11 +1354,17 @@ private AppFabricServiceStatus stop(Id.Program identifier, @Nullable String runI
...
         throw new ProgramNotFoundException(identifier);
-      } else if (runId == null) {
-        throw new BadRequestException("Program not running");
-      } else {
+      } else if (runId != null) {
+        // Check if the program is running and is started by the Workflow
+        RunRecordMeta runRecord = store.getRun(identifier, runId);
+        if (runRecord != null && runRecord.getProperties().containsKey("workflowrunid")
+          && runRecord.getStatus().equals(ProgramRunStatus.RUNNING)) {
+          throw new BadRequestException("MapReduce program run started by Workflow cannot be stopped. " +
```
bdmogal: lets add atleast the mapreduce name and run id, along with the workflow runid as well.

sagarkapare: Fixed error message.

bdmogal:
LGTM. Merge after build succeeds.

sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-RBT430-3

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-RBT430-6. Merging.

PR #3971, opened 2015-09-16
CDAP-3603 make invalid transform configs fail app creation

albertshau:
Moved logic in transforms that validate config settings from
initialize into configurePipeline so that invalid configs fail
early at application creation instead of at runtime.


albertshau:
http://builds.cask.co/browse/CDAP-RBT431

gokulavasan:
LGTM :+1:

PR #3972, opened 2015-09-16
CDAP-3442 fix deploy failures when a config is given to plain app

albertshau:
Fixes a bug where application deployment would fail if a config
was given to an Application class that extends AbstractApplication
without any special config object.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
@@ -173,7 +173,7 @@ public final class InMemoryConfigurator implements Configurator {
         appConfig = ((Class<? extends Config>) configType).newInstance();
       } else {
         try {
-          appConfig = GSON.fromJson(configString, Artifacts.getConfigType(app.getClass()));
+          appConfig = GSON.fromJson(configString, configType);
         } catch (JsonSyntaxException e) {
           throw new IllegalArgumentException("Invalid JSON configuration was provided. Please check the syntax.", e);
         }
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
@@ -34,6 +34,7 @@ import co.cask.cdap.internal.app.runtime.artifact.ArtifactClassLoaderFactory;
 import co.cask.cdap.internal.app.runtime.artifact.ArtifactRepository;
 import co.cask.cdap.internal.app.runtime.artifact.CloseableClassLoader;
 import co.cask.cdap.internal.io.ReflectionSchemaGenerator;
+import co.cask.cdap.internal.lang.Reflections;
 import co.cask.cdap.proto.Id;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
@@ -50,6 +51,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Type;
 import java.util.jar.Manifest;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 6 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-RBT432

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
```diff
@@ -171,7 +171,11 @@ private String getSpecJson(Application app, final String bundleVersion, final St
         appConfig = (Config) configToken.getRawType().newInstance();
       } else {
         try {
-          appConfig = GSON.fromJson(configString, configToken.getType());
```
chtyim: Can you explain a bit? When is `configToken.getType()` returns if the application has extends without specifying the parameterized type (I guess that's what the fix is about)?

gokulavasan:
@albertshau ``Config`` doesn't contain PluginProperties anymore. Is this issue still present?

albertshau:
it's still an issue without PluginProperties.  What happens is configToken.getType() returns some generic T type so gson deserializes it as a TreeMap.  It then tries to assign that to appConfig and we get a ClassCastException since it's trying to cast TreeMap to Config.

chtyim:
It seems wrong that the configType.getType() returns generic type. How do we create the configToken? The deployment pipeline should be extracting it from the parameterized type and verify it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
```diff
@@ -187,4 +184,15 @@ private String getSpecJson(Application app, final String bundleVersion, final St
...
   }
 
+  private Type getConfigType(Class<? extends Application> appClass) {
+    TypeToken<?> configType = TypeToken.of(appClass).resolveType(Application.class.getTypeParameters()[0]);
+    if (Reflections.isResolved(configType.getType())) {
+      return configType.getType();
+    }
+
+    // It has to be Config
+    Preconditions.checkArgument(Config.class == configType.getRawType(), "Application config type not supported. " +
```
gokulavasan: Should this check be done here or when we add an app artifact?

albertshau: thats a good point, this should be checked when deploying an artifact

albertshau:
@gokulavasan @chtyim i've addressed comments, could you take another look?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
```diff
@@ -169,13 +167,13 @@ private String getSpecJson(Application app, final String bundleVersion, final St
...
-      TypeToken<?> configToken = typeToken.resolveType(Application.class.getTypeParameters()[0]);
+      Type configType = Artifacts.getConfigType(app.getClass());
       if (Strings.isNullOrEmpty(configString)) {
-        appConfig = (Config) configToken.getRawType().newInstance();
+        //noinspection unchecked
+        appConfig = ((Class<? extends Config>) configType).newInstance();
       } else {
         try {
-          appConfig = GSON.fromJson(configString, configToken.getType());
+          appConfig = GSON.fromJson(configString, Artifacts.getConfigType(app.getClass()));
```
chtyim: Why need to get the config type again? Isn't it the `configType` already?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/Artifacts.java
```diff
@@ -28,6 +35,25 @@ public static String getFileName(ArtifactId artifactId) {
...
+   */
+  public static Type getConfigType(Class<? extends Application> appClass) {
+    TypeToken<?> configType = TypeToken.of(appClass).resolveType(Application.class.getTypeParameters()[0]);
+    if (Reflections.isResolved(configType.getType())) {
+      return configType.getType();
+    }
+
+    // It has to be Config
+    Preconditions.checkArgument(Config.class == configType.getRawType(), "Application config type not supported. " +
+      "Type must extend Config and cannot be parameterized.");
```
chtyim: Maybe good to have `configType` as part of the error message.

chtyim:
Just two minor comments. Rest LGTM.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/deploy/InMemoryConfigurator.java
albertshau: yeah this should not be here

PR #3973, opened 2015-09-16
CDAP-3400 Do not log error when the Notification service is not yet started.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-3400
Build: http://builds.cask.co/browse/CDAP-RBT433-1

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/LocalStreamService.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/LocalStreamService.java
@@ -18,6 +18,7 @@ package co.cask.cdap.data.stream.service;
 
 import co.cask.cdap.api.data.stream.StreamSpecification;
 import co.cask.cdap.api.metrics.MetricStore;
+import co.cask.cdap.common.ServiceUnavailableException;
 import co.cask.cdap.common.conf.Constants;
 import co.cask.cdap.common.stream.notification.StreamSizeNotification;
 import co.cask.cdap.data.stream.StreamCoordinatorClient;
@@ -200,6 +201,9 @@ public class LocalStreamService extends AbstractStreamService {
       try {
         notificationService.publish(streamFeed, new StreamSizeNotification(System.currentTimeMillis(), absoluteSize))
           .get();
+      } catch (ServiceUnavailableException e) {
+        // No need to propagate the exception. Attempt to publish was made when the NotificationService was not up.
+        // Next run iteration should have the problem fixed.
       } catch (NotificationFeedException e) {
         LOG.warn("Error with notification feed {}", streamFeed, e);
       } catch (Throwable t) {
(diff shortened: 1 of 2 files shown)


sagarkapare:
@chtyim Please help review the PR.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/LocalStreamService.java
```diff
@@ -200,6 +201,9 @@ private void publishNotification(long absoluteSize) {
       try {
         notificationService.publish(streamFeed, new StreamSizeNotification(System.currentTimeMillis(), absoluteSize))
           .get();
+      } catch (ServiceUnavailableException e) {
```
anwar6953: Would it be good to propagate the exception if the next iteration doesn't have it fixed (I mean like after N iterations, the issue is still there)?

chtyim: Should be fine. No need to propagate.

cdap-notifications/src/main/java/co/cask/cdap/notifications/service/inmemory/InMemoryNotificationService.java
```diff
@@ -64,6 +65,10 @@ protected void shutDown() throws Exception {
   public <N> ListenableFuture<N> publish(final Id.NotificationFeed feed, final N notification,
                                          final Type notificationType)
     throws NotificationException {
+    if (executorService == null) {
```
chtyim: It should never happen because if `executorService` is null, meaning this `InMemoryNotificationService`, which is a guava service, is not started. Technically it should be an IllegalStateException.

sagarkapare: In standalone, stream service is started before the app fabric starts and it tries to publish even before the `InMemoryNotificationService.startUp` gets called, which is why the `executorService` is null here. Once the `InMemoryNotificationService.startUp` happens the `executorService` gets initialized and next calls to publish succeeds. This is why I thought of throwing `ServiceUnavailable` here. 
However `IllegalStateException` also makes sense here. Let me know if need to change.

chtyim: Better throw IllegalStateException then.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/LocalStreamService.java
chtyim: The notification system actually don't care about not able to publish a feed (which is part of the design) since it requires feed consumer to resort to polling mechanism if no feed arrives in certain time bound.

I think it's ok to just have a `catch (Throwable t)` and log it as debug.

cdap-notifications/src/main/java/co/cask/cdap/notifications/service/inmemory/InMemoryNotificationService.java
sagarkapare: Sounds good. Updated.

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/service/LocalStreamService.java
sagarkapare: Fixed.

sagarkapare:
@chtyim Please take a look again. Thanks!

chtyim:
LGTM

sagarkapare:
Build passed. Merging.

PR #3979, opened 2015-09-16
 removes icons + chevrons from item headings

astewes:
Removes the following from Hydrator left panel item headings:
*  Plugin type icons
* Chevon-left and chevron-right

left
.2


ajainarayanan:
:+1: LGTM

PR #3981, opened 2015-09-16
CDAP-3690 Fail fast on missing xmllint

wolf31o2:
...and update package dependencies to pull it in on RHEL/UBUNTU


sagarkapare:
LGTM :+1:

PR #3983, opened 2015-09-16
CDAP-3691 fix snapshot filesets to work with multiple sinks

albertshau:
Adding avro and parquet schema settings as arguments passed to
the filesets so that they can be used for the correct output when
multiple sinks are configured.


albertshau:
http://builds.cask.co/browse/CDAP-RBT435

anwar6953:
LGTM

PR #3984, opened 2015-09-16
fix program history to get appId and programId

elfenheart:


elfenheart:
need to reopen against release/3.2

PR #3986, opened 2015-09-16
CDAP-3676: Metadata HTTP RESTful API should support 'now'

jwang47:
https://issues.cask.co/browse/CDAP-3676
http://builds.cask.co/browse/CDAP-RBT436

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/LineageHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/LineageHandler.java
@@ -18,6 +18,7 @@ package co.cask.cdap.metadata;
 
 import co.cask.cdap.common.BadRequestException;
 import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.utils.TimeMathParser;
 import co.cask.cdap.data2.metadata.lineage.Lineage;
 import co.cask.cdap.data2.metadata.lineage.LineageStore;
 import co.cask.cdap.metadata.serialize.LineageRecord;
@@ -36,6 +37,7 @@ import org.jboss.netty.handler.codec.http.HttpResponseStatus;
 
 import java.lang.reflect.Type;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
@@ -66,15 +68,17 @@ public class LineageHandler extends AbstractHttpHandler {
   public void datasetLineage(HttpRequest request, HttpResponder responder,
                              @PathParam("namespace-id") String namespaceId,
                              @PathParam("dataset-id") String datasetId,
-                             @QueryParam("start") @DefaultValue("-1") long start,
-                             @QueryParam("end") @DefaultValue("-1") long end,
+                             @QueryParam("start") String startStr,
+                             @QueryParam("end") String endStr,
                              @QueryParam("levels") @DefaultValue("10") int levels) throws Exception {
 
-    checkArguments(start, end, levels);
+    checkLevels(levels);
+    TimeRange range = parseRange(startStr, endStr);
 
     Id.DatasetInstance datasetInstance = Id.DatasetInstance.from(namespaceId, datasetId);
-    Lineage lineage = lineageGenerator.computeLineage(datasetInstance, start, end, levels);
-    responder.sendJson(HttpResponseStatus.OK, new LineageRecord(start, end, lineage.getRelations()),
+    Lineage lineage = lineageGenerator.computeLineage(datasetInstance, range.getStart(), range.getEnd(), levels);
+    responder.sendJson(HttpResponseStatus.OK,
+                       new LineageRecord(range.getStart(), range.getEnd(), lineage.getRelations()),
                        LineageRecord.class, GSON);
   }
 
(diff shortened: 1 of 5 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/metadata/LineageHandler.java
```diff
@@ -109,18 +114,56 @@ public void getAccessesForRun(HttpRequest request, HttpResponder responder,
     responder.sendJson(HttpResponseStatus.OK, lineageStore.getRunMetadata(run), SET_METADATA_RECORD_TYPE, GSON);
   }
 
-  private void checkArguments(long start, long end, int levels) throws BadRequestException {
+  private TimeRange checkArguments(String startStr, String endStr, int levels) throws BadRequestException {
```
bdmogal: it seems a little weird to have a ``checkArguments`` method that returns a time range. Perhaps we could move the ``levels`` checking out of this method, and have it only be a ``validateAndParseRange()``?

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/LineageTest.java
```diff
@@ -128,14 +128,26 @@ public void testFlowLineage() throws Exception {
...
       // Fetch stream lineage
       httpResponse = fetchLineage(stream, now - oneHourMillis, now + oneHourMillis, 10);
       Assert.assertEquals(200, httpResponse.getResponseCode());
       lineage = GSON.fromJson(httpResponse.getResponseBodyAsString(), LineageRecord.class);
-
       // same as dataset's lineage
       Assert.assertEquals(expected, lineage);
 
+      // Fetch stream lineage with time strings
+      httpResponse = fetchLineage(stream, "now-1h", "now+1h", 10);
```
bdmogal: should we add a verification with bad strings?

bdmogal:
@jwang47 only minor comments. Rest LGTM.

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/LineageTest.java
jwang47: done

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/LineageHandler.java
```diff
@@ -109,18 +114,60 @@ public void getAccessesForRun(HttpRequest request, HttpResponder responder,
...
   }
 
-  private void checkArguments(long start, long end, int levels) throws BadRequestException {
+  private void checkLevels(int levels) throws BadRequestException {
+    if (levels < 1) {
+      throw new BadRequestException(String.format("Invalid levels (%d), should be greater than 0.", levels));
+    }
+  }
+
+  // TODO: This is a fairly common operation useful in various handlers
```
bdmogal: hmmm, sorry I may have missed this earlier, but then should this move to TimeMathParser itself?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/LineageHandler.java
```diff
@@ -109,18 +114,60 @@ public void getAccessesForRun(HttpRequest request, HttpResponder responder,
...
+    }
+    if (endStr == null) {
+      throw new BadRequestException("End time is required.");
+    }
+
+    long now = System.currentTimeMillis();
+    long start;
+    long end;
+    try {
+      start = TimeMathParser.parseTime(now, startStr, TimeUnit.MILLISECONDS);
```
poornachandra: We will need to take in time from user in seconds, and convert it into milliseconds for lineage APIs.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/LineageHandler.java
jwang47: `TimeMathParser` is more general use and throws `IllegalArgumentException`. We'll need some common code for parsing input from HTTP requests and throw `BadRequestException`.

poornachandra: +1 to Alvin's comments.

bdmogal: Ok. Can we report a Jira and add to the TODO then?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/serialize/LineageRecord.java
```diff
@@ -69,6 +69,10 @@ public LineageRecord(long start, long end, Set<Relation> lineageRelations) {
     addRelations(lineageRelations);
   }
 
+  public Set<RelationRecord> getRelations() {
+    return relations;
```
poornachandra: It would be a good idea to return an un-modifiable set here.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/LineageHandler.java
```diff
@@ -66,15 +68,17 @@
...
-    checkArguments(start, end, levels);
+    checkLevels(levels);
+    TimeRange range = parseRange(startStr, endStr);
 
     Id.DatasetInstance datasetInstance = Id.DatasetInstance.from(namespaceId, datasetId);
-    Lineage lineage = lineageGenerator.computeLineage(datasetInstance, start, end, levels);
-    responder.sendJson(HttpResponseStatus.OK, new LineageRecord(start, end, lineage.getRelations()),
+    Lineage lineage = lineageGenerator.computeLineage(datasetInstance, range.getStart(), range.getEnd(), levels);
+    responder.sendJson(HttpResponseStatus.OK,
+                       new LineageRecord(range.getStart(), range.getEnd(), lineage.getRelations()),
```
poornachandra: When we change to take in seconds from user, we need to change what we return also into seconds for consistency.

poornachandra:
@jwang47 Minor comments, please address them.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/LineageHandler.java
jwang47: Done https://issues.cask.co/browse/CDAP-3715

PR #3991, opened 2015-09-17
CDAP-3502 Choose the default behavior of ReflectionSchemaGenerator's …

gokulavasan:
…null property of field schema

JIRA: https://issues.cask.co/browse/CDAP-3502

Build :


gokulavasan:
Will remove against release/3.2

PR #3994, opened 2015-09-17
CDAP-3635: Move metadata handlers into DatasetOpExecutorService

jwang47:
https://issues.cask.co/browse/CDAP-3635
http://builds.cask.co/browse/CDAP-RBT439

- [ ] Test on cluster


bdmogal:
@jwang47 ``RouterPathTest`` will need to be updated too?

jwang47:
Updated `RouterPathTest`

chtyim:
Why closed?

jwang47:
@chtyim It's not possible move metadata handlers into cdap-data-fabric since the metadata handlers depend on LineageGenerator which depends on cdap-app-fabric (Store).

PR #3995, opened 2015-09-17
Fix/ui styling node config tab

astewes:
*  Fixes various styling issues in the node configuration tab
*  Removes styling that is no longer used


ajainarayanan:
There are issues in Firefox and safari. 
 - Anchor points for the nodes
 - Scrolling is tough (different) in both these browsers. Have to move my cursor to a different place to scroll further down the page.

ajainarayanan:
Talked to @astewes The firefox issues can be fixed in a separate PR(it is not the original scope of this PR). :+1:  LGTM

PR #4003, opened 2015-09-17
CDAP-3239 Check if the rowfield type is simple or nullable simple

gokulavasan:
Also modifying couple of test cases to check explore that explore works when the row key field is of type nullable simple and when it is of simple type

JIRA : https://issues.cask.co/browse/CDAP-3239

Build : http://builds.cask.co/browse/CDAP-RBT445-3


albertshau:
:+1:

gokulavasan:

PR #4004, opened 2015-09-17
using NamespacedId codec to serialize the response from metadata search

gsps1:


bdmogal:
LGTM, but can we add a unit test, so we can catch this even earlier?

gsps1:
sure, have added a test case, and added one minor bugfix for the case when target=ALL in query param for search.

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
```diff
@@ -141,6 +143,28 @@ public void testTags() throws IOException {
     tags = getTags(mystream);
     Assert.assertTrue(tags.containsAll(streamTags));
     Assert.assertTrue(streamTags.containsAll(tags));
+    // test search for stream
+    Set<MetadataSearchResultRecord> searchTags = searchTags(Id.Namespace.DEFAULT.getId(), "stT", "STREAM");
```
bdmogal: is it possible to use ``MetadataSearchType`` here?

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
```diff
@@ -141,6 +143,28 @@ public void testTags() throws IOException {
...
+    Set<MetadataSearchResultRecord> expected = ImmutableSet.of(
+      new MetadataSearchResultRecord(mystream, MetadataSearchTargetType.STREAM)
+    );
+    Assert.assertEquals(expected.size(), searchTags.size());
```
bdmogal: can't we just ``Assert.assertEquals(expected, searchTags)``

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
```diff
@@ -141,6 +143,28 @@ public void testTags() throws IOException {
...
+    Assert.assertTrue(searchTags.containsAll(expected));
+
+    // test prefix search, should match stream and service programs
+    searchTags = searchTags(Id.Namespace.DEFAULT.getId(), "s", "ALL");
+    expected = ImmutableSet.of(
+      new MetadataSearchResultRecord(mystream, MetadataSearchTargetType.STREAM),
+      new MetadataSearchResultRecord(pingService, MetadataSearchTargetType.PROGRAM)
+    );
+    Assert.assertEquals(expected.size(), searchTags.size());
```
bdmogal: ditto

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
```diff
@@ -141,6 +143,28 @@ public void testTags() throws IOException {
...
+    Assert.assertTrue(searchTags.containsAll(expected));
+
+    // search without any target param
+    searchTags = searchTags(Id.Namespace.DEFAULT.getId(), "s", null);
+    Assert.assertEquals(expected.size(), searchTags.size());
```
bdmogal: ditto

bdmogal:
Should we add tests for searching properties as well?

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
gsps1: that should be enough.

gsps1:
@bdmogal have added coverage for properties and addressed your comment.

henry-cask:
LGTM

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
```diff
@@ -85,6 +87,26 @@ public void testProperties() throws IOException {
...
+    searchProperties = searchMetadata(Id.Namespace.DEFAULT.getId(), "sKey:s", "ALL");
+    expected = ImmutableSet.of(
+      new MetadataSearchResultRecord(pingService, MetadataSearchTargetType.PROGRAM)
+    );
+    Assert.assertEquals(expected, searchProperties);
+
+    // search without any target param
+    searchProperties = searchMetadata(Id.Namespace.DEFAULT.getId(), "sKey:s", null);
+    Assert.assertEquals(expected, searchProperties);
+
```
bdmogal: in both these search verifications, can you add a test that returns empty? a good case would be to search for an existing property/tag, but on the wrong target type.

bdmogal:
@gsps1 one final verification request. Other than that, LGTM.

bdmogal:
Thanks for adding tests!

gsps1:
my build failed due to a flaky test, but https://github.com/caskdata/cdap/pull/4010 - which is based off my branch passed. so merging this.

PR #4006, opened 2015-09-17
CDAP-3635: Make metadata service monitorable as a system service

jwang47:
https://issues.cask.co/browse/CDAP-3635
http://builds.cask.co/browse/CDAP-RBT446

- [x] Test on cluster


bdmogal:
@jwang47, this does not work. I had tried this. On a cluster, it always shows requested and provisioned as 0. We need to store some info in the ``service.store`` for this to work correctly. There is some code in ``MasterServiceMain`` to do that.

jwang47:
@bdmogal Hopefully fixed. This will now use "dataset.executor" as the YARN container that it looks for, and "metadata.service" as the discoverable name.

jwang47:
@bdmogal Works in cluster.

gokulavasan:
LGTM :+1:

PR #4008, opened 2015-09-17
remove all uses of WorkflowAction.configure()

anew:


anew:
Build is running here: http://builds.cask.co/browse/CDAP-RBT448-1

gokulavasan:
Do we want to leave at least one test which uses the old configure API so that we know it is working? We can convert that too when we remove the old API?

gokulavasan:
One comment. LGTM otherwise.

anew:
I added a custom action that implements the legacy configure(). It turns out that even before my change, that did not work, because it does not set the class name in the action spec. It works if you extend AbstractWorkflowAction, but not if you implement WorkflowAction directly. I will try to fix this, too.

anew:
@gokulavasan I fixed all the issues I could find locally on my laptop. New build is running here: http://builds.cask.co/browse/CDAP-RBT448. Please review again. Thanks

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/WorkflowNodeCreator.java
```diff
@@ -56,11 +57,9 @@ static WorkflowNode createWorkflowCustomActionNode(WorkflowAction action) {
     Preconditions.checkArgument(action != null, "WorkflowAction is null.");
     WorkflowActionSpecification spec;
     if (action instanceof AbstractWorkflowAction) {
-      DefaultWorkflowActionConfigurer configurer = new DefaultWorkflowActionConfigurer(action);
-      ((AbstractWorkflowAction) action).configure(configurer);
-      spec = configurer.createSpecification();
+      spec = DefaultWorkflowActionConfigurer.configureAction((AbstractWorkflowAction) action);
```
gokulavasan: i guess we don't need this casting any more?

gokulavasan: do we still need that casting?

cdap-app-fabric/src/test/java/co/cask/cdap/AppWithWorkflow.java
```diff
@@ -89,11 +90,9 @@ public DummyAction(String name) {
...
 
     @Override
-    public WorkflowActionSpecification configure() {
-      return WorkflowActionSpecification.Builder.with()
-        .setName(name)
-        .setDescription(getDescription())
-        .build();
+    public void configure(WorkflowActionConfigurer configurer) {
+      super.configure(configurer);
+      setName(name);
```
gokulavasan: any reason to exclude the setDescription method?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/workflow/WorkflowNodeCreator.java
anew: We do need it. The configureAction() requires an AbstractWorkflowAction.

cdap-app-fabric/src/test/java/co/cask/cdap/AppWithWorkflow.java
anew: because getDescription is deprecated, and defaults to the same as name.

gokulavasan:
Minor nit. LGTM :+1:

PR #4009, opened 2015-09-17
Only localize plugin artifact jar if there is plugin being used for an application.

chtyim:
Tested in cluster.


gokulavasan:
LGTM :+1:

chtyim:
Doesn't affect local unit-test. Merging.

PR #4010, opened 2015-09-17
Add support for both prefix scan search via * and keyword specific search

henry-cask:
Currently the metadata search using prefix scan search by default.

So search for value will return for value, value1, value2, and value21.

With the change search for value will return value, search for value2* will return value2 and value21.

Merge changes from #4004 for testing.

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
@@ -97,18 +97,25 @@ public class MetadataHttpHandlerTest extends MetadataTestBase {
     Assert.assertEquals(expected, searchProperties);
 
     // test prefix search for service
-    searchProperties = searchMetadata(Id.Namespace.DEFAULT.getId(), "sKey:s", "ALL");
+    searchProperties = searchMetadata(Id.Namespace.DEFAULT.getId(), "sKey:s*", "ALL");
     expected = ImmutableSet.of(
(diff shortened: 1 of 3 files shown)


henry-cask:
Build:

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataHttpHandlerTest.java
```diff
@@ -97,18 +97,18 @@ public void testProperties() throws IOException {
...
-    searchProperties = searchMetadata(Id.Namespace.DEFAULT.getId(), "sKey:s", "ALL");
+    searchProperties = searchMetadata(Id.Namespace.DEFAULT.getId(), "sKey:s*", "ALL");
     expected = ImmutableSet.of(
       new MetadataSearchResultRecord(pingService, MetadataSearchTargetType.PROGRAM)
     );
     Assert.assertEquals(expected, searchProperties);
 
     // search without any target param
-    searchProperties = searchMetadata(Id.Namespace.DEFAULT.getId(), "sKey:s", null);
+    searchProperties = searchMetadata(Id.Namespace.DEFAULT.getId(), "sKey:s*", null);
```
bdmogal: what will happen without the ``*``? Can you add a verification for that? It should return an empty set also, right?

henry-cask: Yeah, I will add test for it.

bdmogal:
LGTM.

henry-cask:
Thanks @bdmogal , will squash

bdmogal:
Build passed. Merging.

PR #4011, opened 2015-09-17
CDAP-3680: Views shouldn't be created over non-existing streams

jwang47:
https://issues.cask.co/browse/CDAP-3680
http://builds.cask.co/browse/CDAP-RBT450

- [x] Run int tests against cluster (since some guice binding changes are needed)


cdap-app-fabric/src/main/java/co/cask/cdap/app/guice/AppFabricServiceRuntimeModule.java
```diff
@@ -277,13 +275,15 @@ protected void configure() {
 
       install(
         new FactoryModuleBuilder()
-          .implement(new TypeLiteral<Manager<AppDeploymentInfo, ApplicationWithPrograms>>() { },
-            new TypeLiteral<LocalApplicationManager<AppDeploymentInfo, ApplicationWithPrograms>>() { })
-          .build(new TypeLiteral<ManagerFactory<AppDeploymentInfo, ApplicationWithPrograms>>() { })
+          .implement(new TypeLiteral<Manager<AppDeploymentInfo, ApplicationWithPrograms>>() {
+          },
+                     new TypeLiteral<LocalApplicationManager<AppDeploymentInfo, ApplicationWithPrograms>>() {
```
chtyim: Why the formatting change?

cdap-data-fabric/src/main/java/co/cask/cdap/data/stream/StreamViewHttpHandler.java
```diff
@@ -99,39 +101,35 @@ public void delete(HttpRequest request, HttpResponder responder,
...
-                   @PathParam("stream") String stream) throws NotFoundException {
+                   @PathParam("stream") String stream) throws Exception {
 
-    Id.Stream streamId;
     try {
-      streamId = Id.Stream.from(namespace, stream);
+      Id.Stream streamId = Id.Stream.from(namespace, stream);
+      Collection<String> list = Collections2.transform(
+        admin.listViews(streamId), new Function<Id.Stream.View, String>() {
+          @Nullable
```
chtyim: Remove.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/transaction/stream/FileStreamAdmin.java
```diff
@@ -353,6 +359,30 @@ public void drop(Id.Stream streamId) throws Exception {
   }
 
   @Override
+  public boolean createOrUpdateView(Id.Stream.View viewId, ViewSpecification spec) throws Exception {
+    Preconditions.checkArgument(exists(viewId.getStream()), "Stream '%s' does not exist.", viewId.getStreamId());
```
chtyim: All these actions should be done through the `StreamCoordinatorClient`, which provides locking mechanism for stream admin operations. You might want to add a new method to `StreamCoordinatorClient` so that it can perform arbitrary operation while holding the lock. Something like:

```java
interface StreamCoordinatorClient {

  <T> T exclusiveAction(Id.Stream streamId, Callable<T> action) throws Exception;

  void exclusiveAction(Id.Stream streamId, Runnable action) throws Exception;
}
```

chtyim:
LGTM

PR #4012, opened 2015-09-17
Change default log back to INFO

sreevatsanraman:
* Change default log back container to info so that the INFO messages for customer apps are collected by default.


dereklwood:
:+1: lgtm

sreevatsanraman:
Merging without a build since this is a config change.

PR #4013, opened 2015-09-17
Temporarily decreasing Log Saver check-pointing frequency

poornachandra:


sreevatsanraman:
LGTM

poornachandra:
Going to merge without build since it is only a configuration change.

PR #4014, opened 2015-09-17
removes search from Hydrator navbar

astewes:
* Removes search HTML and associated styling from Hydrator navbar

screen shot 2015-09-17 at 2 52 34 pm


ajainarayanan:
:+1: LGTM

PR #4015, opened 2015-09-17
fix create app command so that the config sent in the request

albertshau:
comes from a config field in the file contents. This lets us
extend the object if we have more arguments in the future, and it
also makes it so that the same file can be used to create an app
through curl and the CLI

### DIFF
--- a/cdap-cli/src/main/java/co/cask/cdap/cli/command/app/CreateAppCommand.java
+++ b/cdap-cli/src/main/java/co/cask/cdap/cli/command/app/CreateAppCommand.java
@@ -31,16 +31,19 @@ import co.cask.cdap.proto.artifact.ArtifactSummary;
 import co.cask.common.cli.Arguments;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
+import com.google.gson.reflect.TypeToken;
 import com.google.inject.Inject;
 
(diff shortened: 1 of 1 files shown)


cdap-cli/src/main/java/co/cask/cdap/cli/command/app/CreateAppCommand.java
```diff
@@ -67,7 +67,8 @@ public void perform(Arguments arguments, PrintStream output) throws Exception {
     if (configPath != null) {
       File configFile = resolver.resolvePathToFile(configPath);
       try (FileReader reader = new FileReader(configFile)) {
-        config = GSON.fromJson(reader, JsonObject.class);
+        ConfigFile configContents = GSON.fromJson(reader, ConfigFile.class);
```
jwang47: Can we use whatever object the HTTP handler uses here instead?

albertshau: thats a good point, will do that

albertshau:
@jwang47 addressed comment, could you take another look?

jwang47:
LGTM

albertshau:
http://builds.cask.co/browse/CDAP-RBT460

PR #4017, opened 2015-09-17
[ CDAP-3634] Add support for case insensitive search for metadata

henry-cask:
Support metadata search with case insensitive. So search for value "value" and "VaLue" should yields same result.

To do this, we are introducing new column on the metadata index table column to store lower case value and key:value combo.


henry-cask:
Build:
http://builds.cask.co/browse/CDAP-RBT451-1

bdmogal:
@henry-cask nit: the commit message has the wrong jira. The jira for this is CDAP-3634. Please fix before merging.

bdmogal:
@henry-cask this LGTM. Do you want to add a verification in the ``MetadataHttpHandlerTest`` as well, just for the sake of completion?

henry-cask:
Ah yes, thanks for catching it.

henry-cask:
Build passes: http://builds.cask.co/browse/CDAP-RBT451-1
Merging ...

PR #4018, opened 2015-09-17
reduces top offset of dag node placement

astewes:
*  Adjusts top offset of DAG node placement on Hydrator canvas

dag


ajainarayanan:
:+1: LGTM

PR #4019, opened 2015-09-18
Fix/ui push messages to console tab

ajainarayanan:
- Adds error badges during plugin add
- Adds console tab functionality (show warnings, info and error messages).
- Dynamically show (all the time) how many required fields left to be filled for any node.
- On validate/publish make the warning badges as errors.
- Fix validator to be edited in the new node config tab
- Fix plugins that has schema as its property.
- Fix editing plugins that doesn't have any configuration in node proxy.
- Fix the ability for the user to switch between different tabs when editing a plugin.
- Fix the need for user to save a node that has just been opened.
- Add reference tab contents.
- Add the validate button functionality.

### DIFF
--- a/cdap-ui/app/features/adapters/controllers/create/plugin-edit-form-ctrl.js
+++ b/cdap-ui/app/features/adapters/controllers/create/plugin-edit-form-ctrl.js
@@ -186,6 +186,13 @@ angular.module(PKG.name + '.feature.adapters')
             // TODO: Hacky. Need to fix this for am-fade-top animation for modals.
             $timeout(function() {
               $scope.pluginCopy = angular.copy($scope.plugin);
+              $scope.$watch('pluginCopy.properties', function() {
+                  var strProp1 = JSON.stringify($scope.pluginCopy.properties);
+                  var strProp2 = JSON.stringify($scope.plugin.properties);
+                  if (strProp1 !== strProp2) {
+                    $scope.data['isModelTouched'] = true;
+                  }
+              }, true);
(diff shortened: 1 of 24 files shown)


cdap-ui/app/features/adapters/controllers/create/plugin-edit-form-ctrl.js
```diff
@@ -216,7 +215,12 @@ angular.module(PKG.name + '.feature.adapters')
     }
 
     this.reset = function () {
-      $scope.pluginCopy = angular.copy($scope.plugin);
+      $scope.pluginCopy.properties = angular.copy($scope.plugin.properties);
+      $scope.pluginCopy.outputSchema = angular.copy($scope.plugin.outputSchema);
+      $scope.pluginCopy.inputSchema = angular.copy($scope.plugin.inputSchema);
+      $scope.pluginCopy.errorDatasetName = angular.copy($scope.plugin.errorDatasetName);
+      // $scope.pluginCopy.validationFields = angular.copy($scope.plugin.validationFields);
```
elfenheart: remove?

cdap-ui/app/features/adapters/templates/create/toppanel.html
```diff
@@ -15,12 +15,13 @@
 -->
 
 <div class="side-panel top text-center clearfix">
-  <div class="adapter-metadata pull-left"
+  <!-- <div class="adapter-metadata pull-left"
```
elfenheart: remove?

elfenheart:
:+1: LGTM

PR #4021, opened 2015-09-18
CDAP-3714 remove requirement that main class is an application

albertshau:
We need to remove this requirement because there are 3rd party
jars that include this attribute in their manifest.


albertshau:
http://builds.cask.co/browse/CDAP-RBT452

gokulavasan:
LGTM :+1:

PR #4022, opened 2015-09-18
(CDAP-3716) Refactor code that causes deadlock

chtyim:
- Remove unnecessary synchronize method
- Always update the in memory runtimeInfos from the controller callback thread
- Remove bunch of unused code


chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT455

chtyim:
Verified on cluster for the latest changes.

cdap-app-fabric/src/main/java/co/cask/cdap/app/runtime/AbstractProgramRuntimeService.java
```diff
@@ -287,14 +281,20 @@ protected void updateRuntimeInfo(ProgramType type, RunId runId, RuntimeInfo runt
     lock.lock();
```
gokulavasan: Do we need to place a writeLock here? What write operation are we performing?

chtyim: right, a read lock should be enough.

gokulavasan:
LGTM :+1:

PR #4023, opened 2015-09-18
Fix/ui hydrator cron

elfenheart:
TODO: 
- Styling for slanted tabs.... [DONE]

screen shot 2015-09-18 at 11 42 11 am


ajainarayanan:
On clicking save the save and reset button remains clickable and I was able to reset even after saving.

1. Disable save and reset buttons after save
2. User should not be able to reset once saved.

ajainarayanan:
:+1: LGTM

PR #4024, opened 2015-09-18
CDAP-3263 Removing row.case.insensitive field and adding a field case…

gokulavasan:
… choice in DBSource plugin

JIRA : https://issues.cask.co/browse/CDAP-3263

Build : http://builds.cask.co/browse/CDAP-RBT458

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
@@ -308,7 +308,7 @@ public class BatchETLDBTest extends BaseETLBatchTest {
     mrManager.start();
     mrManager.waitForFinish(5, TimeUnit.MINUTES);
     List<RunRecord> runRecords = mrManager.getHistory();
-    Assert.assertNotEquals(ProgramRunStatus.FAILED, runRecords.get(0).getStatus());
+    Assert.assertEquals(ProgramRunStatus.COMPLETED, runRecords.get(0).getStatus());
 
     // No records should be written
     DataSetManager<Table> outputManager = getDataset("outputTable1");
@@ -359,7 +359,7 @@ public class BatchETLDBTest extends BaseETLBatchTest {
     mrManager.start();
     mrManager.waitForFinish(5, TimeUnit.MINUTES);
     List<RunRecord> runRecords = mrManager.getHistory();
-    Assert.assertNotEquals(ProgramRunStatus.FAILED, runRecords.get(0).getStatus());
+    Assert.assertEquals(ProgramRunStatus.COMPLETED, runRecords.get(0).getStatus());
   }
 
   private void createInputData() throws Exception {
--- a/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
@@ -26,6 +26,8 @@ import co.cask.cdap.etl.batch.config.ETLBatchConfig;
 import co.cask.cdap.etl.common.ETLStage;
 import co.cask.cdap.etl.common.Properties;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.ProgramRunStatus;
+import co.cask.cdap.proto.RunRecord;
 import co.cask.cdap.proto.artifact.AppRequest;
 import co.cask.cdap.test.ApplicationManager;
 import co.cask.cdap.test.DataSetManager;
@@ -66,7 +68,7 @@ import javax.sql.rowset.serial.SerialBlob;
 /**
  * Test for ETL using databases
  */
-public class BatchETLDBAdapterTest extends BaseETLBatchTest {
+public class BatchETLDBTest extends BaseETLBatchTest {
   private static final long currentTs = System.currentTimeMillis();
   private static final String clobData = "this is a long string with line separators \n that can be used as \n a clob";
   private static HSQLDBServer hsqlDBServer;
@@ -180,7 +182,7 @@ public class BatchETLDBAdapterTest extends BaseETLBatchTest {
   @Test
   @Category(SlowTests.class)
   @SuppressWarnings("ConstantConditions")
-  public void testDBSourceWithCaseInsensitiveRowKey() throws Exception {
+  public void testDBSource() throws Exception {
     String importQuery = "SELECT ID, NAME, SCORE, GRADUATED, TINY, SMALL, BIG, FLOAT_COL, REAL_COL, NUMERIC_COL, " +
       "DECIMAL_COL, BIT_COL, DATE_COL, TIME_COL, TIMESTAMP_COL, BINARY_COL, BLOB_COL, CLOB_COL FROM my_table " +
       "WHERE ID < 3";
@@ -196,9 +198,7 @@ public class BatchETLDBAdapterTest extends BaseETLBatchTest {
     ETLStage sink = new ETLStage("Table", ImmutableMap.of(
       "name", "outputTable",
       Properties.Table.PROPERTY_SCHEMA, schema.toString(),
-      Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "id",
-      // Test case-insensitive row key matching
-      Properties.Table.CASE_SENSITIVE_ROW_FIELD, "false"));
+      Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "ID"));
 
     ETLBatchConfig etlConfig = new ETLBatchConfig("* * * * *", source, sink, Lists.<ETLStage>newArrayList());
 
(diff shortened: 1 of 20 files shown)


cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
```diff
@@ -300,12 +307,24 @@ public void testDBSourceWithCaseSensitiveRowKey() throws Exception {
     MapReduceManager mrManager = appManager.getMapReduceManager(ETLMapReduce.NAME);
     mrManager.start();
     mrManager.waitForFinish(5, TimeUnit.MINUTES);
+    List<RunRecord> runRecords = mrManager.getHistory();
+    Assert.assertNotEquals(ProgramRunStatus.FAILED, runRecords.get(0).getStatus());
```
bdmogal: can we assert succeeded here?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
```diff
@@ -339,6 +358,8 @@ public void testDBSink() throws Exception {
...
```
bdmogal: ditto

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/source/DBSource.java
```diff
@@ -60,6 +64,11 @@
     "import from the specified table. Examples: SELECT COUNT(*) from <my_table> where <my_column> 1, " +
     "SELECT COUNT(my_column) from my_table. NOTE: Please include the same WHERE clauses in this query as the ones " +
     "used in the import query to reflect an accurate number of records to import.";
+  private static final String COLUMN_CASE_DESCRIPTION = "Sets the case of the column names returned from the query." +
+    "Possible options are UPPER, lower. By default or for any other input, the column names are not modified" +
```
bdmogal: I can see why you did this, but is it confusing in a description to have "UPPER, lower"? Does that give the notion that those are the only two supported values (whereas the code accepts any case for these two)

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/StructuredRecordUtils.java
```diff
@@ -0,0 +1,78 @@
...
+   * @return {@link StructuredRecord} which contains field names confirming to the {@link FieldCase} passed in
+   * @throws Exception if there is a conflict in the field names while converting the case
+   */
+  public static StructuredRecord convertCase(StructuredRecord input, FieldCase fieldCase) throws Exception {
+    if (fieldCase.equals(FieldCase.NONE)) {
+      return input;
+    }
+
+    Schema oldSchema = input.getSchema();
+    Map<String, String> fieldNameMap = Maps.newHashMap();
```
bdmogal: ``new HashMap<>();``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/StructuredRecordUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    for (Map.Entry<String, String> nameMap : fieldNameMap.entrySet()) {
+      recordBuilder.set(nameMap.getValue(), input.get(nameMap.getKey()));
+    }
+    return recordBuilder.build();
+  }
+
+  private StructuredRecordUtils() {
+  }
+
+  private static String nameChange(String oldName, FieldCase fieldCase) {
```
bdmogal: ``changeName()``, to make it a verb?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/etl/common/StructuredRecordUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+  @Test
+  public void testLowerAndUpperCaseTransformation() throws Exception {
+    StructuredRecord record = StructuredRecord.builder(
+      Schema.recordOf("dbrecord",
+                      Schema.Field.of("Name", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+                      Schema.Field.of("Age", Schema.of(Schema.Type.INT)))).set("Name", "Abcd").set("Age", 10).build();
+    StructuredRecord upperCaseRecord = StructuredRecordUtils.convertCase(
+      record, FieldCase.toFieldCase("upPer"));
+    Assert.assertEquals("Abcd", upperCaseRecord.get("NAME"));
+    Assert.assertEquals(10, upperCaseRecord.get("AGE"));
```
bdmogal: should we assert that the structured record does not have the old fields?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/etl/common/StructuredRecordUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    StructuredRecord lowerCaseRecord = StructuredRecordUtils.convertCase(
+      record, FieldCase.toFieldCase("lowEr"));
+    Assert.assertEquals("Abcd", lowerCaseRecord.get("name"));
+    Assert.assertEquals(10, lowerCaseRecord.get("age"));
```
bdmogal: ditto

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/StructuredRecordUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    Schema oldSchema = input.getSchema();
+    Map<String, String> fieldNameMap = Maps.newHashMap();
+    List<Schema.Field> newFields = Lists.newArrayList();
+    for (Schema.Field field : oldSchema.getFields()) {
+      String newName = nameChange(field.getName(), fieldCase);
+      if (fieldNameMap.containsValue(newName)) {
+        // field name used already. indication of field names conflict. can't do anything.
+        throw new Exception(String.format(
```
bdmogal: Do we have an ``InvalidConfigurationException`` or something like that? If the exception is checked (expected to be caught by callers), then better to throw a specific exception? If we don't care about clients catching it, then should we just throw ``IllegalStateException``?

bdmogal:
@gokulavasan only minor comments.

cdap-ui/templates/common/Database.json
```diff
@@ -59,7 +59,15 @@
           "countQuery" : {
              "widget": "textbox",
              "label": "Count SQL"
-          }
+          },
+	  "columnNameCase" : {
+	     "widget": "select",
+	     "label" : "Column Case",
+	     "properties" : {
+		"values" : ["UPPER", "lower", "No change"],
```
bdmogal: nit: indentation

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/source/DBSource.java
gokulavasan: Oh hmm. the code accepts any case of upper, lower. Should we just check for one specific case?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/StructuredRecordUtils.java
gokulavasan: Are we moving away from Guava classes? Or just want to reduce dependencies in new code?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/StructuredRecordUtils.java
gokulavasan: I looked for a good type of exception to throw. There is ``InvalidPluginConfigurationException`` but that is only in cdap-app-fabric. May we can move it elsewhere.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/StructuredRecordUtils.java
bdmogal: yeah, with java 7, the ``new HashMap<>()`` is concise, so we're moving away

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/source/DBSource.java
bdmogal: no, I just meant in the description does it throw you off (that these are the only two values supported). I saw that the code accepts any case, which is good.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/StructuredRecordUtils.java
bdmogal: yeah, not sure if that will increase the size of this PR, but that sounds like a good exception. Is that exception supposed to be used elsewhere?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/source/DBSource.java
gokulavasan: These are the two options that users will see on the UI tool tip. Can you suggest an alternate description?

bdmogal: upper, lower?

(Only if you think the current description could throw users off). I don't feel too strongly about it.

gokulavasan:
@bdmogal Addressed PR comments. Please take a look again.

bdmogal:
LGTM

gokulavasan:
Build is :green_apple:. Merging. Thanks for the review @bdmogal.

PR #4026, opened 2015-09-18
CDAP-3700 Reduce LogSaver connections to HBase

gsps1:
we now Use TransactionExecutor instead of Transactional to avoid having many connections open to  HBase while writing in an transaction.

we have also added few improvements to reduce the number of transactions by sending the partition and the checkpoint in a map and having them written in one transaction, rather than creating a new transaction for each partition and checkpoint.

we also skip writing to table if the checkpoints have not changed by comparing with a local map of the past checkpoint.

JIRA: https://issues.cask.co/browse/CDAP-3700

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/logging/save/CheckpointManager.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/logging/save/CheckpointManager.java
@@ -19,14 +19,18 @@ package co.cask.cdap.logging.save;
 import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Table;
-import co.cask.cdap.data2.dataset2.tx.DatasetContext;
-import co.cask.cdap.data2.dataset2.tx.Transactional;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.data2.transaction.Transactions;
+import co.cask.tephra.TransactionAware;
 import co.cask.tephra.TransactionExecutor;
 import co.cask.tephra.TransactionExecutorFactory;
-import com.google.common.base.Supplier;
-import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * Manages reading/writing of checkpoint information for a topic and partition.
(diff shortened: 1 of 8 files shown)


gsps1:
build : http://builds.cask.co/browse/CDAP-RBT459-1

gsps1:
build GREEN : http://builds.cask.co/browse/CDAP-RBT459-3

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/CheckpointManager.java
```diff
@@ -37,50 +41,84 @@
...
+
+  public void saveCheckpoint(final Map<Integer, Checkpoint> checkpointMap) throws Exception {
+    LOG.trace("Saving checkpoints for partitions {}", checkpointMap);
+
+    // if the checkpoints have not changed, we skip writing to table and return.
+    if (lastCheckpoint.equals(checkpointMap)) {
+      return;
+    }
+
+    execute(new TransactionExecutor.Function<Table, Void>() {
```
bdmogal: You can use ``TransactionExecutor.Procedure`` if you want to return void.

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/CheckpointManager.java
```diff
@@ -37,50 +41,84 @@
...
+    return execute(new TransactionExecutor.Function<Table, Map<Integer, Checkpoint>>() {
       @Override
-      public Void apply(DatasetContext<Table> ctx) throws Exception {
-        Table table = ctx.get();
-        byte[] key = Bytes.add(rowKeyPrefix, Bytes.toBytes(partition));
-        table.put(key, OFFSET_COLNAME, Bytes.toBytes(checkpoint.getNextOffset()));
-        table.put(key, MAX_TIME_COLNAME, Bytes.toBytes(checkpoint.getMaxEventTime()));
-        return null;
+      public Map<Integer, Checkpoint> apply(Table table) throws Exception {
+        Map<Integer, Checkpoint> checkpointMap1 = new HashMap<Integer, Checkpoint>();
```
bdmogal: better var name?

bdmogal: Also, throughout, can we remove ``Map`` from the var names? The type already indicates that it is a map.

cdap-watchdog/src/main/java/co/cask/cdap/logging/write/FileMetaDataManager.java
```diff
@@ -102,12 +93,12 @@ private void writeMetaData(final String logPartition,
     LOG.debug("Writing meta data for logging context {} as startTimeMs {} and location {}",
               logPartition, startTimeMs, location.toURI());
 
-    mds.execute(new TransactionExecutor.Function<DatasetContext<Table>, Void>() {
+    execute(new TransactionExecutor.Function<Table, Void>() {
```
bdmogal: use procedure to return void.

bdmogal:
minor comments. Fix and merge. I don't think you need another build for the changes. Please run tests and checkstyle locally after addressing comments. LGTM

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/CheckpointManager.java
```diff
@@ -37,50 +41,84 @@
...
+                                                   "Exception while trying to cast it to TransactionAware. " +
+                                                   "Please check why the table is not TransactionAware", table));
+      }
+    } catch (Exception e) {
+      throw new RuntimeException(String.format("Error accessing %s table", tableUtil.getMetaTableName()), e);
+    }
+  }
+
+  public void saveCheckpoint(final Map<Integer, Checkpoint> checkpointMap) throws Exception {
+    LOG.trace("Saving checkpoints for partitions {}", checkpointMap);
```
poornachandra: It would be a good idea to move this log message after the condition below, so that we log message only if there is a write.

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/CheckpointManager.java
```diff
@@ -37,50 +41,84 @@
...
       @Override
-      public Checkpoint apply(DatasetContext<Table> ctx) throws Exception {
-        Row result =
-          ctx.get().get(Bytes.add(rowKeyPrefix, Bytes.toBytes(partition)));
+      public Checkpoint apply(Table table) throws Exception {
+        Row result = table.get(Bytes.add(rowKeyPrefix, Bytes.toBytes(partition)));
         return new Checkpoint(result.getLong(OFFSET_COLNAME, -1), result.getLong(MAX_TIME_COLNAME, -1));
       }
     });
-    LOG.trace("Read checkpoint {} for partition {}", checkpoint, partition);
```
poornachandra: Lets' retain this log message.

poornachandra:
@gsps1 Two minor comments, please fix and merge.

cdap-watchdog/src/main/java/co/cask/cdap/logging/save/CheckpointManager.java
gsps1: yeah, good point

bdmogal:
LGTM. Squash commits and merge.

PR #4035, opened 2015-09-18
(CDAP-3695) Make sure every notification in StreamSizeSchedulerTest i…

bdmogal:
…s published 1 second later than the previous. One hypothesis for the flaky nature could be that multiple notifications get published at the same time, thereby leading to some notifications getting lost?

Not completely sure this fixes the flakyness completely yet, want some feedback -

The test tries to ensure that the stream size increased from the last time, so a notification must have gotten published, so a workflow must have started. Then it asserts that the workflow runs increased from the previous runs. It does this multiple times, and we've seen failures for all these assertions from time to time. The failures are because the test can't find that the runs are increased in 5 seconds after the notification is published. In this PR, I'm trying to ensure that each notification gets published one second after the last, to reduce its flakyness. However, if that's the reason for failure, could there be a problem in the notification system where it misses notifications?

Jira: [CDAP-3695](https://issues.cask.co/browse/CDAP-3695)
Build: http://builds.cask.co/browse/CDAP-RBT463-1


bdmogal:
Still get the same failure even with this fix. Will close and investigate further.

PR #4037, opened 2015-09-18
[CDAP-3722] Add filtering search metadata on active namespace.

henry-cask:
Right now, when search done on value or key:value the returned entities are not filtered by active namespace.

We now add namespace as prefix on the value and key:value index columns in the IndexedTable for BusinessMetadataDataset.

Updated tests accordingly.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataAdmin.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataAdmin.java
@@ -116,6 +116,7 @@ public interface MetadataAdmin {
   /**
    * Execute search for metadata for particular type of CDAP object.
    *
+   * @param namespaceId The namespace id to be filter the search by.
    * @param searchQuery The query need to be executed for the search.
    * @param type The particular type of CDAP object that the metadata need to be searched. If null all possible types
    *             will be searched.
@@ -123,6 +124,6 @@ public interface MetadataAdmin {
    * @return a {@link Set} records for metadata search.
    * @throws NotFoundException if there is not record found for particular query text.
    */
-  Set<MetadataSearchResultRecord> searchMetadata(String searchQuery, @Nullable MetadataSearchTargetType type)
-    throws NotFoundException;
+  Set<MetadataSearchResultRecord> searchMetadata(String namespaceId, String searchQuery,
+                                                 @Nullable MetadataSearchTargetType type) throws NotFoundException;
 }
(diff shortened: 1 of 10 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataAdmin.java
```diff
@@ -123,6 +124,6 @@ void addProperties(Id.NamespacedId entityId, Map<String, String> properties)
    * @return a {@link Set} records for metadata search.
    * @throws NotFoundException if there is not record found for particular query text.
    */
-  Set<MetadataSearchResultRecord> searchMetadata(String searchQuery, @Nullable MetadataSearchTargetType type)
-    throws NotFoundException;
+  Set<MetadataSearchResultRecord> searchMetadata(String namespaceId, String searchQuery,
```
bdmogal: Better to take in ``Id.Namespace``?

henry-cask:
Build: http://builds.cask.co/browse/CDAP-RBT464-1

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataAdmin.java
henry-cask: No because Id.NamespacedId does not have the actual namespace id. Need to cast it to right child class.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/metadata/dataset/BusinessMetadataDataset.java
```diff
@@ -461,13 +465,27 @@ private void addNamespaceIdToKey(MDSKey.Builder builder, Id.NamespacedId namespa
     throw new IllegalArgumentException("Illegal Type " + type + " of metadata source.");
   }
 
+  private String getNamespaceId(MDSKey key) {
```
bdmogal: This should be ``getNamespacedId``, since namespaceId means something else.

bdmogal: oh oops no I may have got that wrong. Can you instead change the comment two lines below to split the ``[targetId]`` into its components? Or another comment mentioning namespaceId is the first part of ``[targetId]``

henry-cask: Will do

bdmogal:
Can you also add verifications in ``MetadataHttpHandlerTest``?

henry-cask:
Rerun build after rebase:
http://builds.cask.co/browse/CDAP-RBT464-2

bdmogal:
:+1:

henry-cask:
Build passes, merging ...

PR #4041, opened 2015-09-18
Set worker instances and worker resources from the ETLWorker Config

gokulavasan:
Build : http://builds.cask.co/browse/CDAP-RBT465


cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/etl/realtime/ETLRealtimeApplication.java
```diff
@@ -36,7 +36,7 @@ public void configure() {
     }
 
     setDescription("Realtime Extract-Transform-Load (ETL) Template");
-    addWorker(new ETLWorker(config));
+    addWorker(new ETLWorker(config, instances));
```
albertshau: the config has instances, don't need to pass it as a separate arg. Instead, we should move the verification logic a few lines up into the worker

albertshau:
one minor comment, otherwise lgtm

gokulavasan:

PR #4044, opened 2015-09-18
Fix/hydrator firefox

astewes:
*  Fixes node endpoint display issue (Firefox only)
*  Moves left panel toggle button from group-side-panel to adapters
*  Replaces floats with fixed left panel
*  Adjusts positioning of left panel groups when not opened
*  Slightly shrinks the canvas height

groups


ajainarayanan:
:+1: LGTM

PR #4049, opened 2015-09-18
changes layout of hydrator header

astewes:
*  Moves navbar links to center
*  Moves return to CDAP link to right

screen shot 2015-09-18 at 3 06 35 pm


elfenheart:
:+1: LGTM

PR #4051, opened 2015-09-18
increases width of metadata name input

astewes:
*  Increases length of metadata name input in edit mode
*  Left aligns metadata name container so fix left padding issue
*  Hides pencil from display in edit mode

meta


elfenheart:
:+1: LGTM

PR #4054, opened 2015-09-18
CDAP-3727 fix bug with order of system artifact loading

albertshau:
The previous implementation tried to enforce ordering by sorting
on the artifacts, with parents ordered larger than children. The
problem was the comparison make sense when 2 artifacts were being
compared, but it wasn't a natural ordering. That is, there could
be a scenario where A < B, B < C, but A > C.

Instead, since we enforce only 1 level of dependencies, just
adding all parents first, then adding the children.

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepositoryTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepositoryTest.java
@@ -141,10 +141,10 @@ public class ArtifactRepositoryTest {
       createManifest(ManifestFields.EXPORT_PACKAGE, PluginTestRunnable.class.getPackage().getName()));
 
     // write plugins jar
-    Id.Artifact pluginArtifactId = Id.Artifact.from(Id.Namespace.SYSTEM, "myPlugin", "1.0.0");
+    Id.Artifact pluginArtifactId = Id.Artifact.from(Id.Namespace.SYSTEM, "APlugin", "1.0.0");
 
     Manifest manifest = createManifest(ManifestFields.EXPORT_PACKAGE, TestPlugin.class.getPackage().getName());
-    createPluginJar(TestPlugin.class, new File(systemArtifactsDir, "myPlugin-1.0.0.jar"), manifest);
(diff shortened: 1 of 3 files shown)


cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -187,7 +187,8 @@
 
   <property>
     <name>app.artifact.dir</name>
-    <value>/opt/cdap/master/artifacts</value>
+    <!--value>/opt/cdap/master/artifacts</value-->
```
gokulavasan: Is this change intentional? If yes, should we retain the comment?

albertshau: ugh no it's not, changing it back

albertshau:
http://builds.cask.co/browse/CDAP-RBT472

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepositoryTest.java
```diff
@@ -141,10 +141,10 @@ public void testAddSystemArtifacts() throws Exception {
       createManifest(ManifestFields.EXPORT_PACKAGE, PluginTestRunnable.class.getPackage().getName()));
 
     // write plugins jar
-    Id.Artifact pluginArtifactId = Id.Artifact.from(Id.Namespace.SYSTEM, "myPlugin", "1.0.0");
+    Id.Artifact pluginArtifactId = Id.Artifact.from(Id.Namespace.SYSTEM, "APlugin", "1.0.0");
```
gokulavasan: Do the renaming in this test test the fix? If yes, how?

gokulavasan:
One question. Rest LGTM :+1:

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactRepositoryTest.java
albertshau: o this was leftovers from debugging.  I will modify the test to actually test it... need to add another plugin jar to this test case.

PR #4055, opened 2015-09-18
increases size of pipeline icon

astewes:
*  Hydrator icon font size is now slightly larger

screen shot 2015-09-18 at 4 09 04 pm


ajainarayanan:
:+1: LGTM

PR #4057, opened 2015-09-18
(CDAP-3695) Increased timeout for run record verifications in StreamS…

bdmogal:
…izeSchedulerTest, since it seems like sometimes expecting a completed run record to be available 5 seconds after a workflow run (or two) starts is unreasonable


anwar6953:
LGTM

bdmogal:
I've tested this locally over 10 times. In the morning I could reproduce the failures fairly regularly. Not happening anymore.

bdmogal:
Tested this locally multiple times. Since this is only increasing timeout in unit tests, and has so much benefit, I'm going to merge without build. Checkstyle passes as well.

PR #4059, opened 2015-09-18
Make sure the fork executor terminates before the Workflow completed

chtyim:
- This is needed since the individual program run records is written
  from the thread inside the executor.


sagarkapare:
LGTM :+1:

chtyim:
Building in http://builds.cask.co/browse/CDAP-RBT470-1

chtyim:
Verified on cluster.

bdmogal:
Build is green. Merging.

sagarkapare:
Verified for the positive(normal) test case as well.

PR #4060, opened 2015-09-19
(CDAP-3723) Should stop the service after usage of it is done

chtyim:
- It was working before because we didn’t block on the stop, so if we get
  lucky, the DS service is still running.


poornachandra:
LGTM

chtyim:
Tested the affected test cases locally. Merge to unblock other builds.

PR #4066, opened 2015-09-19
[CDAP-3709] Standalone fails after network connection drops

anew:
Fixed by making all services bind to 127.0.0.1 in standalone. 
Tested multiple times on my laptop. 
Tested listing apps and programs, running apps, getting metrics, getting logs executing queries.
Also tested that I works fine after network drops. 
Also tested that everything works fine after the network comes back.


nitinmotgi:
Tested this on flight! LGTM :100:

anwar6953:
@nitinmotgi 
Based upon the discussion I heard from Andreas and Terence, the issue isn't when you run Standalone without Wifi. Instead, its when you start Standalone with wifi, and then disconnect (or vice-versa).
But it requires a change in the wifi connectivity state to be an issue.

chtyim:
LGTM

anew:
Build passed: http://builds.cask.co/browse/CDAP-RBT474-1
Merging.

nitinmotgi:
@anwar6953 yes, but it's also the same issue.

PR #4067, opened 2015-09-19
Fixes schema propagation + editing outputschema on plugins saveable

ajainarayanan:
- Fixes schema propagation.
- Disables save and reset when the user opens the node configuration and makes no change.


elfenheart:
:+1: LGTM

PR #4069, opened 2015-09-19
bugfix for skipping workflow stats when workflow node has failed

gsps1:
when one of the workflow node fails in the workflow, we would set workflow status to be completed so we do not restart the workflow.

We want the workflow stats to be recorded only for successfully completed workflow runs, so if there is a workflow node run whose status is not completed, we skip writing stats.

https://issues.cask.co/browse/CDAP-3460

build : http://builds.cask.co/browse/CDAP-RBT475-1

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
@@ -280,9 +280,9 @@ public class DefaultStore implements Store {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


bdmogal:
LGTM. Merge when build passes. No way we can test this scenario in a unit test, right?

gsps1:
yeah. in singlenode, if mapreduce fails, workflow status is failed. there is a inconsistency between singlenode and distributed if you look at the status.

bdmogal:
Build passed - http://builds.cask.co/browse/CDAP-RBT475-1. Merging.

PR #4071, opened 2015-09-19
Docs/3.2 fix default tool

awholegunch:
Added comparing two cdap-default.xml files to the tool.


awholegunch:
"On-hold" because I added a change this morning to figure out the properties that are the same in two files but with different values. And some additional cleanup of that code would be helpful.

bdmogal:
LGTM

PR #4073, opened 2015-09-19
Throw an error if an application tries to add streams, datasets,

gokulavasan:
plugins with same name in same/different programs.

Punting the more detailed error message to next release (JIRA : https://issues.cask.co/browse/CDAP-3729)

Build : http://builds.cask.co/browse/CDAP-RBT477

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
+++ b/cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
@@ -25,7 +25,9 @@ import co.cask.cdap.api.dataset.module.DatasetModule;
 import co.cask.cdap.internal.dataset.DatasetCreationSpec;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Implementation of {@link DatasetConfigurer} for adding datasets and streams in Configurers.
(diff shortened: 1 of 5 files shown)


gokulavasan:
@albertshau Please review

cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
```diff
@@ -50,21 +52,43 @@
...
   public void addDatasetSpecs(Map<String, DatasetCreationSpec> datasetSpecs) {
+    Set<String> duplicateDatasetSpecs = findDuplicates(datasetSpecs.keySet(), this.datasetSpecs.keySet());
+    checkArgument(duplicateDatasetSpecs.isEmpty(),
+                  "Dataset Instances %s have already been added. Remove the duplicates.", duplicateDatasetSpecs);
     this.datasetSpecs.putAll(datasetSpecs);
   }
 
   public void addDatasetModules(Map<String, String> datasetModules) {
+    for (Map.Entry<String, String> newEntry : datasetModules.entrySet()) {
+      if (datasetModules.containsKey(newEntry.getKey())) {
```
sagarkapare: should it be `this.datasetModules`?

gokulavasan: Yes, good catch! Thanks! Updated the PR and also enhanced the unit test to cover this case.

gokulavasan:
@sagarkapare Please take a look again. Thanks!

sagarkapare:
LGTM :+1:

gokulavasan:

PR #4076, opened 2015-09-19
Fixes converting numbers while saving pipeline + Kafka.json configuration fixes

ajainarayanan:
- Converts all numbers or any other types to be string for plugin properties before publishing it to backend.
- Fixes Kafka.json file (offset, partition and schema) to have proper default values.


elfenheart:
:+1: LGTM

PR #4077, opened 2015-09-19
fixes padding issue on select dropdown

astewes:
*  Adds right padding to select dropdown to prevent overflowing text

text


ajainarayanan:
:+1: LGTM

PR #4083, opened 2015-09-20
Remove Elasticsearch UI configuration template from CDAP since we do …

bdmogal:
…not package elasticsearch plugin with CDAP. It has already been checked in into the cdap-plugins repo.


gokulavasan:
LGTM :+1:

PR #4085, opened 2015-09-20
restructure plugin doc to first describe plugins and capabilities,

albertshau:
then walk through an example. Also moving some of the deployment
docs from custom etl plugins doc to the general plugins doc and
fixing errors.


awholegunch:
I've made a few edits, fixed a couple of formatting problems with literals you were having (by using an escaped space, '\ '). LGTM, but please check my edits.

albertshau:
edits lgtm

PR #4087, opened 2015-09-20
removes add button from list template

astewes:
*  Removes Add+ button from empty pipeline list template
*  Adds a small top margin to the empty well to be consistent with table margin in list view

screen shot 2015-09-20 at 10 08 05 am
I


ajainarayanan:
:+1: LGTM

PR #4088, opened 2015-09-20
Fix/ui increase node label width

astewes:
*  Increases width of node container
*  Adds padding/container to empty node config tab body
*  Decreases padding/font-size of plugin items (left panel)
*  Increases myEllipsis of plugin item names from 10 to 15

dag


ajainarayanan:
:+1: LGTM

PR #4093, opened 2015-09-21
CDAP-3748 fix load artifact command to parse parents correctly

albertshau:
The load artifact command was incorrectly parsing parents as
system artifacts. System artifacts are supposed by be prefixed
by 'system:' in the parent list to distinguish between parents
that are in the system scope versus in the user scope.

Refactoring ArtifactConfig so that it is purely for the config
file instead of a container for the artifact id, file, and config.
Introducing ArtifactConfigReader that will read files for a
particular namespace instead of assuming its in the system
namespace.


albertshau:
http://builds.cask.co/browse/CDAP-RBT482

gokulavasan:
LGTM :+1:

PR #4098, opened 2015-09-21
(CDAP-3645) Copy the latest TypeResolver and related classes from Guava

chtyim:
- This should be compilable with both Java 7 and 8


gokulavasan:
LGTM :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2890-1

PR #4099, opened 2015-09-21
Move stream CLI commands to INGEST category

jwang47:


awholegunch:
I'd like to change some of these help messages, but there's no point as they are also in the code and unless we change them there, they will get wiped out the next update. Was there a PR to update the source?

jwang47:
This PR just moves stream commands to the ingest category and copy pastes the output from `GenerateCLIDocsTable`. We can modify the help messages in a separate PR.

awholegunch:
ok, thanks.

awholegunch:
LGTM.

PR #4107, opened 2015-09-21
adds directive to calc height of item body

astewes:
*  Adds directive to calculate height of parent element so that plugin items can occupy the full amount of space before starting to scroll

calc


astewes:
@ajainarayanan changes LGTM

PR #4109, opened 2015-09-21

jwang47:
- [x] Tested locally in standalone with MySQL server

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
@@ -35,7 +35,9 @@ import co.cask.cdap.etl.common.DBUtils;
 import co.cask.cdap.etl.common.ETLDBOutputFormat;
 import co.cask.cdap.etl.common.JDBCDriverShim;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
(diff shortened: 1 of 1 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
```diff
@@ -140,8 +143,9 @@ private void setResultSetMetadata() throws Exception {
            ResultSet rs = statement.executeQuery(String.format("SELECT %s FROM %s WHERE 1 = 0",
                                                                dbSinkConfig.columns, dbSinkConfig.tableName))
       ) {
+        List<String> columns = ImmutableList.copyOf(Splitter.on(",").split(dbSinkConfig.columns));
         ResultSetMetaData resultSetMetadata = rs.getMetaData();
-        int columnCount = resultSetMetadata.getColumnCount();
+        int columnCount = columns.size();
```
bdmogal: does this work? I mean this will mean that the for loop on line 151 will only pick up the first ``<columnCount>`` columns from resultSetMetadata, right? What if a column in the middle was removed?

jwang47: @gokulavasan mentioned the test case that raised this issue after I opened the PR. After testing, I discovered this issue and it's now fixed.

jwang47:
Oops. Reopened here https://github.com/caskdata/cdap/pull/4114

PR #4110, opened 2015-09-21
[CDAP-3782] [CDAP-3780] [CDAP-3781] Fix/ui plugin templates fixes

elfenheart:
- Adding description to plugin template
- Change icons in left sidebar to be plugin base type
- fix fetch list of plugins based on templates and type
- renaming a plugin template name will not create a new one

plugintemplates

### DIFF
--- a/cdap-ui/app/features/admin/controllers/namespace/templates-ctrl.js
+++ b/cdap-ui/app/features/admin/controllers/namespace/templates-ctrl.js
@@ -39,6 +39,8 @@ angular.module(PKG.name + '.feature.admin')
       GLOBALS.pluginTypes[vm.GLOBALS.etlRealtime].transform
     ];
 
+    vm.pluginDescription = '';
+
     var plugin;
 
     function initialize() {
@@ -139,6 +141,8 @@ angular.module(PKG.name + '.feature.admin')
             lock: template.lock
           };
 
+          vm.pluginDescription = template.description;
+
           initialize();
 
         });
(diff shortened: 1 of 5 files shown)


cdap-ui/app/features/admin/controllers/namespace/templates-ctrl.js
```diff
@@ -199,6 +211,20 @@ angular.module(PKG.name + '.feature.admin')
             return;
           }
 
+          if (vm.isEdit && oldTemplateName !== vm.pluginConfig.pluginTemplate) {
+            if (config) {
+              $alert({
+                type: 'danger',
+                content: 'Template name already exist! Please choose another name'
```
anwar6953: "exist" should be "exists"

cdap-ui/app/features/admin/templates/namespace/templates.html
```diff
@@ -33,8 +33,15 @@ <h2 ng-if="TemplatesController.isEdit">Edit Template</h2>
...
     <label class="label-control">Choose Plugin Type</label>
     <select ng-model="TemplatesController.pluginType"
-            data-ng-options="o for o in TemplatesController.pluginoptions"
-            class="form-control">
+            data-ng-options="o for o in TemplatesController.pluginoptionsBatch"
+            class="form-control"
+            ng-if="TemplatesController.templateType === TemplatesController.GLOBALS.etlBatch">
+    </select>
+
+    <select ng-model="TemplatesController.pluginType"
```
ajainarayanan: Note: Add the ng-if to a parent div to avoid momentarily showing the dropdown.

cdap-ui/app/features/admin/controllers/namespace/templates-ctrl.js
ajainarayanan: Note: If we could move this constants.js then that would be even better.

ajainarayanan:
:+1: lgtm

PR #4111, opened 2015-09-21
adds 2px bottom margin to input rows

astewes:
*  Adds 2px bottom margin to rows in keyvalue-container and dsv-container
*  Adds auto height to schema container textarea

rows


elfenheart:
:+1: LGTM

PR #4112, opened 2015-09-21
add a known issue about CDAP-3732

albertshau:


cdap-docs/reference-manual/source/release-notes.rst
```diff
@@ -23,6 +23,19 @@ Cask Data Application Platform Release Notes
...
+=============================================================
+
+.. _known-issues-320:
+
+Known Issues
+------------
+
+- `CDAP-3732 <https://issues.cask.co/browse/CDAP-3732>`__ -
+  The first time a flow is started in a namespace, there will be an error message in the logs
+  about a failure to enable explore on the system.queue table. The call to enable exploration
```
awholegunch: Would it be better to say "A call to enable exploration"...as that is what we are suggesting, no?

awholegunch:
One comment. otherwise, LGTM

PR #4114, opened 2015-09-21

jwang47:
- [x] Tested locally in standalone with MySQL server

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
@@ -21,6 +21,7 @@ import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.data.batch.OutputFormatProvider;
 import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.plugin.PluginConfig;
 import co.cask.cdap.api.plugin.PluginProperties;
@@ -34,8 +35,11 @@ import co.cask.cdap.etl.common.DBRecord;
 import co.cask.cdap.etl.common.DBUtils;
 import co.cask.cdap.etl.common.ETLDBOutputFormat;
 import co.cask.cdap.etl.common.JDBCDriverShim;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
 import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
 import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.mapred.lib.db.DBConfiguration;
 import org.slf4j.Logger;
@@ -48,7 +52,9 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
@@ -64,6 +70,7 @@ public class DBSink extends BatchSink<StructuredRecord, DBRecord, NullWritable> 
   private Class<? extends Driver> driverClass;
   private JDBCDriverShim driverShim;
   private int [] columnTypes;
+  private List<String> columns;
 
   public DBSink(DBSinkConfig dbSinkConfig) {
     this.dbSinkConfig = dbSinkConfig;
@@ -111,7 +118,20 @@ public class DBSink extends BatchSink<StructuredRecord, DBRecord, NullWritable> 
 
   @Override
   public void transform(StructuredRecord input, Emitter<KeyValue<DBRecord, NullWritable>> emitter) throws Exception {
-    emitter.emit(new KeyValue<DBRecord, NullWritable>(new DBRecord(input, columnTypes), null));
+    // Create StructuredRecord that only has the columns in this.columns
+    List<Schema.Field> outputFields = new ArrayList<>();
+    for (String column : getColumns()) {
+      Schema.Field field = input.getSchema().getField(column);
+      Preconditions.checkNotNull(field, "Missing schema field for column '%s'", column);
+      outputFields.add(field);
+    }
+    StructuredRecord.Builder output = StructuredRecord.builder(
+      Schema.recordOf(input.getSchema().getRecordName(), outputFields));
+    for (String column : getColumns()) {
+      output.set(column, input.get(column));
+    }
+
+    emitter.emit(new KeyValue<DBRecord, NullWritable>(new DBRecord(output.build(), columnTypes), null));
   }
 
   @Override
@@ -125,6 +145,11 @@ public class DBSink extends BatchSink<StructuredRecord, DBRecord, NullWritable> 
     DBUtils.cleanup(driverClass);
   }
 
+  @VisibleForTesting
+  List<String> getColumns() {
+    return columns;
+  }
+
   private void setResultSetMetadata() throws Exception {
     ensureJDBCDriverIsAvailable();
     Connection connection;
@@ -133,6 +158,8 @@ public class DBSink extends BatchSink<StructuredRecord, DBRecord, NullWritable> 
     } else {
       connection = DriverManager.getConnection(dbSinkConfig.connectionString, dbSinkConfig.user, dbSinkConfig.password);
     }
+
+    Map<String, Integer> columnToType = new HashMap<>();
     try {
       try (Statement statement = connection.createStatement();
            // Run a query against the DB table that returns 0 records, but returns valid ResultSetMetadata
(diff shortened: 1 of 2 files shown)


cdap-api-common/src/main/java/co/cask/cdap/api/data/format/StructuredRecord.java
```diff
@@ -58,6 +58,15 @@ public Schema getSchema() {
   }
 
   /**
+   * Get the fields in the record.
+   *
+   * @return the fields in the record.
+   */
+  public Map<String, Object> getFields() {
```
gokulavasan: Do we need to add this method? Can't we get the schema and then get fields from it?

jwang47: I added this to make it easier to write the unit test.

bdmogal: This is an internal detail. should probably not expose it outside, especially just for tests.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
```diff
@@ -141,11 +169,19 @@ private void setResultSetMetadata() throws Exception {
...
         // JDBC driver column indices start with 1
-        for (int i = 0; i < columnCount; i++) {
-          columnTypes[i] = resultSetMetadata.getColumnType(i + 1);
+        for (int i = 0; i < rs.getMetaData().getColumnCount(); i++) {
+          String name = resultSetMetadata.getColumnName(i + 1);
+          int type = resultSetMetadata.getColumnType(i + 1);
+          columnToType.put(name, type);
+        }
+
+        columns = ImmutableList.copyOf(Splitter.on(",").split(dbSinkConfig.columns));
```
bdmogal: can this (line 180-185) move outside the try-catch? Want to do as less as possible while the ``Connection is active.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
```diff
@@ -141,13 +170,20 @@ private void setResultSetMetadata() throws Exception {
...
-        for (int i = 0; i < columnCount; i++) {
-          columnTypes[i] = resultSetMetadata.getColumnType(i + 1);
+        for (int i = 0; i < rs.getMetaData().getColumnCount(); i++) {
+          String name = resultSetMetadata.getColumnName(i + 1);
+          int type = resultSetMetadata.getColumnType(i + 1);
+          columnToType.put(name, type);
         }
       }
+
+      columns = ImmutableList.copyOf(Splitter.on(",").split(dbSinkConfig.columns));
```
bdmogal: can it move outside this try-catch too?

I mean after the connection is closed?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
```diff
@@ -111,7 +118,20 @@ public void initialize(BatchRuntimeContext context) throws Exception {
 
   @Override
   public void transform(StructuredRecord input, Emitter<KeyValue<DBRecord, NullWritable>> emitter) throws Exception {
-    emitter.emit(new KeyValue<DBRecord, NullWritable>(new DBRecord(input, columnTypes), null));
+    // Create StructuredRecord that only has the columns in this.columns
+    List<Schema.Field> outputFields = Lists.newArrayList();
```
bdmogal: ``new ArrayList<>()``

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
```diff
@@ -111,7 +118,20 @@ public void initialize(BatchRuntimeContext context) throws Exception {
...
+    for (String column : getColumns()) {
+      Schema.Field field = input.getSchema().getField(column);
+      Preconditions.checkNotNull(field, "Missing schema field for column '%s'", column);
+      outputFields.add(field);
+    }
+    StructuredRecord.Builder output = StructuredRecord.builder(
+      Schema.recordOf(input.getSchema().getRecordName(), outputFields));
+    for (String column : getColumns()) {
```
bdmogal: nit: can use ``columns`` directly here instead of a method call?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
```diff
@@ -111,7 +118,20 @@ public void initialize(BatchRuntimeContext context) throws Exception {
...
-    emitter.emit(new KeyValue<DBRecord, NullWritable>(new DBRecord(input, columnTypes), null));
+    // Create StructuredRecord that only has the columns in this.columns
+    List<Schema.Field> outputFields = Lists.newArrayList();
+    for (String column : getColumns()) {
+      Schema.Field field = input.getSchema().getField(column);
+      Preconditions.checkNotNull(field, "Missing schema field for column '%s'", column);
+      outputFields.add(field);
+    }
+    StructuredRecord.Builder output = StructuredRecord.builder(
+      Schema.recordOf(input.getSchema().getRecordName(), outputFields));
```
bdmogal: ``input.getSchema()`` will give the overall schema right? You only want a subset of it, consisting of ``columns`` only?

bdmogal: nvm. I misread this.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
```diff
@@ -125,6 +145,14 @@ public void destroy() {
     DBUtils.cleanup(driverClass);
   }
 
+  public int[] getColumnTypes() {
```
bdmogal: do we need this to be exposed? Its only used in this class.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/etl/batch/sink/DBSinkTest.java
```diff
@@ -0,0 +1,162 @@
...
+    MockEmitter<KeyValue<DBRecord, NullWritable>> emitter = new MockEmitter<>();
+    sink.transform(input, emitter);
+
+    Assert.assertEquals(1, emitter.getEmitted().size());
+    Assert.assertEquals(
+      getRecordFields(output),
+      getRecordFields(emitter.getEmitted().get(0).getKey().getRecord()));
+  }
+
+  @Test(expected = NullPointerException.class)
```
bdmogal: nit: Where will this be thrown? If you are expecting a generic exception like ``NullPointerException`` (which could be thrown at multiple places, and not a specific exception, then it would be better to enclose the exact lines that throw ``NullPointerException`` in a try catch and ignore the exception.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/etl/batch/sink/DBSinkTest.java
```diff
@@ -0,0 +1,162 @@
...
+        Schema.Field.of("body", Schema.of(Schema.Type.STRING)),
+        Schema.Field.of("ts", Schema.of(Schema.Type.LONG))))
+      .set("body", "sdfsdf")
+      .set("ts", 123)
+      .build();
+
+    MockEmitter<KeyValue<DBRecord, NullWritable>> emitter = new MockEmitter<>();
+    sink.transform(input, emitter);
+
+    Assert.assertEquals(1, emitter.getEmitted().size());
```
bdmogal: are these assertions needed if there is an exception? there shouldn't be anything in the emitter in this case, right?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/batch/sink/DBSink.java
jwang47: This is useful for testing when we want to use a specific list of columns without having to go through unrelated code paths.

bdmogal: ok, then lets make it package-private and ``@VisibleForTesting``? for both those methods?

bdmogal: also a comment perhaps, so people don't remove it later?

jwang47: Seems like overkill, especially since breaking this would cause `DBSinkTest` to fail, which only takes 1s to run.

bdmogal:
Code LGTM.

jwang47:
There's an issue with how this handles Oracle SQL tables, because when we fetch the column info, the column names are always in uppercase, so our schema column names also need to be in uppercase. We want to provide some way to handle this, similar to how it's done in DBSource.

For now, we'll merge to develop first and fix the uppercase issue. Once it's fixed, we'll cherry-pick the commits to release/3.2.

jwang47:
Reopened against develop: https://github.com/caskdata/cdap/pull/4149

PR #4115, opened 2015-09-21
Fix CLI examples.

awholegunch:
Fixes a problem in the table generation. This, however, will probably not be easy to fix in the Java code that generates this.


jwang47:
LGTM

PR #4117, opened 2015-09-21
adds item count to opened group

astewes:
*  Adds badge with total item count to opened group in left side panel

badge


elfenheart:
When we are filtering the list, do we want to update the count to become the filtered list or should it remain as the original count?

ajainarayanan:
:+1: LGTM apart from @elfenheart's comment.

elfenheart:
updated with the filtered length

astewes:
LGTM

PR #4122, opened 2015-09-21
fixes font size of pipeline type

astewes:
*  Adds font size and makes pipeline type consistent with how it appears in studio view

type


elfenheart:
:+1: LGTM

PR #4125, opened 2015-09-21
removing elastic search plugins since they were moved to the

albertshau:
cdap-plugins repo


gokulavasan:
LGTM :+1:

awholegunch:
LGTM. The only concerns I have are: we don't have anything that links to the "new" location; some of the info in the docs, such as details on how it works, the need to be running the server before the app is created—though maybe that really meant uploaded—haven't been captured in the other repo.

albertshau:
o we should update the other repo if it is missing documentation that is present here

awholegunch:
https://issues.cask.co/browse/CDAP-3800 has been created to do that later.

albertshau:
thanks, will merge this then

albertshau:
after build http://builds.cask.co/browse/CDAP-RBT486

albertshau:
build green, merging

PR #4126, opened 2015-09-21
Cleanup Gson usage to make sure custom types are not cached permanent…

chtyim:
…ly in static Gson variable.


bdmogal:
@chtyim this should be against release/3.2?

chtyim:
Oops.

PR #4141, opened 2015-09-22
Fixes UI to show console error when un-saved node config exists while publishing a pipeline

ajainarayanan:
- Fixes studio to throw an error while publishing/validating a pipeline if there are un-saved node configurations.


astewes:
LGTM

PR #4142, opened 2015-09-22
Hides schema for sinks when it doesn't have a schema property

ajainarayanan:
- Removes output schema for sinks if it doesn't have a schema property.

FIXME: 
  Need to confirm about implicit-schema. If there is implicit-schema should we display the output schema or hide it?


elfenheart:
:+1: LGTM

PR #4143, opened 2015-09-22
CDAP-3793 and CDAP-3798 : Add a timeout for ZKClient, BrokerService w…

gokulavasan:
…hen initializing worker and error out if we are unable to connect; Sleep for a second, if we get exception in connection in order to print huge overflow of LOG messages since there is no polling wait in the Worker and thus the poll method is executed continuously

Also, modifying the Kafka.json to allow offsets upto -2. And remove description from front end JSON to allow description from backend to surface.

JIRAs:
https://issues.cask.co/browse/CDAP-3798
https://issues.cask.co/browse/CDAP-3793

Build : http://builds.cask.co/browse/CDAP-RBT488


gokulavasan:
@albertshau @chtyim Please review. Thanks!

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/realtime/kafka/Kafka08SimpleApiConsumer.java
```diff
@@ -184,10 +187,18 @@ public void initialize(RealtimeContext context) throws Exception {
...
-      zkClient.startAndWait();
-
       brokerService = new ZKBrokerService(zkClient);
-      brokerService.startAndWait();
+
+      try {
+        Futures.get(zkClient.start(), 3, TimeUnit.SECONDS, TimeoutException.class);
+        Futures.get(brokerService.start(), 3, TimeUnit.SECONDS, TimeoutException.class);
+      } catch (TimeoutException e) {
+        Futures.getUnchecked(brokerService.stop());
```
chtyim: This is the same as calling stopAndWait()

nitinmotgi:
Gokul can you provide a very simple description of what is offset. People who use UI and people who use rest API or CLI are different.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/realtime/source/KafkaSource.java
```diff
@@ -110,14 +112,14 @@ public void initialize(RealtimeContext context) throws Exception {
...
-  public SourceState poll(Emitter<StructuredRecord> writer, SourceState currentState) {
+  public SourceState poll(Emitter<StructuredRecord> writer, SourceState currentState) throws InterruptedException {
     try {
       // Lets set the internal offset store
       kafkaConsumer.saveState(currentState);
-
       kafkaConsumer.pollMessages(writer);
     } catch (Throwable t) {
       LOG.error("Error encountered during poll to get message for Kafka source.", t);
+      TimeUnit.SECONDS.sleep(EXCEPTION_SLEEP_IN_SEC);
```
albertshau: in addition to this can we keep a flag that indicates whether to log an error or not.  Then the first time we catch and error, we log saying there is some problem, but don't log again until we are able to successfully poll

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/realtime/kafka/Kafka08SimpleApiConsumer.java
```diff
@@ -402,6 +413,7 @@ private InetSocketAddress findLeader(String brokers, String topic, int partition
       } catch (Exception e) {
         LOG.error("Failed to communicate with broker {}:{} for leader lookup for topic-partition {}-{}",
                   broker.getKey(), broker.getValue(), topic, partition, e);
+        Throwables.propagate(e);
```
albertshau: what is the effect of this? Also, think its better for this method to throw the exception rather than propagate it as a runtime exception.

gokulavasan: When I checked the cluster where https://issues.cask.co/browse/CDAP-3798 was observed, I noticed that all the log messages were from this logging. Without propagating this exception, we might not able to detect it and sleep.

gokulavasan:
@albertshau Addressed PR comments. Please take a look.

albertshau:
:+1:

gokulavasan:
@awholegunch Can you rename the ``kafka.default.offset`` field to ``kafka.initial.offset`` in the docs?

ajainarayanan:
:+1: on json file changes

gokulavasan:
Tested on cluster. Can be merged once @awholegunch has his fixes in.

bdmogal:
I am going to merge this. Let's do doc changes in a separate PR.

Thanks @gokulavasan !

PR #4146, opened 2015-09-22
Fix ETL image

awholegunch:
Updates the ETL image used at http://docs.cask.co/cdap/3.2.0-SNAPSHOT/en/included-applications/etl/index.html

Includes both  PNG and SVG versions.


albertshau:
:+1:

PR #4149, opened 2015-09-22

jwang47:


jwang47:
Build passed here: http://builds.cask.co/browse/CDAP-RBT483

PR #4150, opened 2015-09-22
Fix/ui metadata tag

elfenheart:
metadata


bdmogal:
@elfenheart is it necessary to click a button to submit? Should we just accept enter?

elfenheart:
@bdmogal ability to hit enter added

ajainarayanan:
:+1: LGTM

PR #4151, opened 2015-09-22
Instead of returning stack trace just return the rootcause message.

sagarkapare:
In case of exception while accessing the service handler, instead of returning entire stack trace, just return the root cause message.


chtyim:
LGTM

sagarkapare:
Release build is running here - http://builds.cask.co/browse/CDAP-RBT490-1

sagarkapare:
Build passed here - http://builds.cask.co/browse/CDAP-RBT490-3. Merging.

PR #4152, opened 2015-09-22
Move the dependency of geronimo-jms_1.1_spec as quick fix in 3.2

henry-cask:
This will change it as a direct dependency in etl-realtime app and have Export-Package of "javax.jms.*".

Need to revert this hack and introduce a proper plugin class loading model like we used to have in 3.1.
See: https://issues.cask.co/browse/CDAP-3810


chtyim:
LGTM

henry-cask:
CC @chtyim

henry-cask:
http://builds.cask.co/browse/CDAP-RBT491-1

henry-cask:
Build passes, merging ...

PR #4153, opened 2015-09-22
UIFix/ui styling admin templates table

astewes:
*  Adds proper styling to Admin > namespace > templates table
*  Removes Processing column from program status directive
*  Switches order of Save/Reset buttons in Hydrator > Pipeline Config tab

pipeline


cdap-ui/app/features/admin/admin.less
```diff
@@ -227,4 +227,78 @@ body.theme-cdap {
...
+        .box-shadow(none);
+        .placeholder-color(@color: #666, @font-weight: 400);
+        &:focus {
+          border-color: @table-border-color;
+          .box-shadow(none);
+        }
+      }
+    }
+  }
 }
```
elfenheart: add a newline

elfenheart:
minor comment, otherwise :+1: LGTM

PR #4159, opened 2015-09-23
Fixes number widget in hydrator to have an internal 'numbered' model …

ajainarayanan:
…while cloning/importing


elfenheart:
:+1: LGTM

PR #4160, opened 2015-09-23
documenting workaround for CDAP-3817

dereklwood:
This documents the workaround for CDAP-3817.  We instruct the user to create the ``tx.snapshot`` directory prior to starting cdap, to avoid the potential for it getting created with improper permissions.


awholegunch:
LGTM, but please check my changes.

dereklwood:
:+1: lgtm

PR #4164, opened 2015-09-23
License dependencies

awholegunch:
Building at: http://builds.cask.co/browse/CDAP-DDB16-1


awholegunch:
On-hold because there appear to be some discrepancies in the lists. Some items are being dropped that probably shouldn't. Attempts to ssh to recent clusters failed. More research is needed.

awholegunch:
I fixed the problem with the enterprise (distributed) dependencies. Removed the on-hold label.
Building at: http://builds.cask.co/browse/CDAP-DDB16-2

bdmogal:
As long as this has been verified on more than one clusters and standalone, :+1:

PR #4172, opened 2015-09-23
[CDAP-3830] Fix the metadata search on tags

henry-cask:
Since the tags actually will be stored as comma delimited strings, when searching
for one of the tags doing both prefix and keyword match will not work well
if an entity has more than one tags.

Assume a program has tags: tag1,tag2,tag3 then search on tags:tag2 will not yield
any result nor search on tags:tag2*.

Need to add special handling for searching tags for key:value format.


henry-cask:
Build:
http://builds.cask.co/browse/CDAP-RBT492-1

PR #4183, opened 2015-09-24
Removing threshold for persisting kafka offset for metrics

gsps1:
https://issues.cask.co/browse/CDAP-3832


gsps1:
build running : http://builds.cask.co/browse/CDAP-DUT2905-1

albertshau:
what are the performance implications of removing the threshold? Are we sure we want to remove it? Is there be some way of persisting for each threshold plus persisting before shutting down?

gsps1:
i don't think this will affect the performance much, as anyway we are writing the metrics to table whenever next() is called, now we also write the offset whenever next() is called. 

since we are writing to the same key for storing the offset, it should be a memory-write in hbase as well. 

good point mentioning about persisting before shutdown, as i noticed we do persist currently in shutdown, but that wasn't being called due to this issue, https://issues.cask.co/browse/CDAP-3861

gsps1:
closing : will reopen against 3.2 branch

PR #4186, opened 2015-09-25
CDAP-3832 - persisting kafka offset only after metrics are processed

gsps1:
Reopening : https://github.com/caskdata/cdap/pull/4183

### DIFF
--- a/cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/PersistedMessageCallback.java
+++ b/cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/PersistedMessageCallback.java
@@ -60,6 +60,10 @@ public final class PersistedMessageCallback implements KafkaConsumer.MessageCall
   @Override
   public void onReceived(Iterator<FetchedMessage> messages) {
     delegate.onReceived(new OffsetTrackingIterator(messages));
+    if (messageCount.get() >= persistThreshold) {
+      messageCount.set(0);
+      persistOffsets();
+    }
   }
 
   @Override
@@ -68,11 +72,11 @@ public final class PersistedMessageCallback implements KafkaConsumer.MessageCall
       delegate.finished();
     } finally {
       // Save the offset
-      persist();
+      persistOffsets();
     }
   }
 
-  private void persist() {
+  private void persistOffsets() {
     try {
       metaTable.save(ImmutableMap.copyOf(offsets));
     } catch (Exception e) {
@@ -102,12 +106,11 @@ public final class PersistedMessageCallback implements KafkaConsumer.MessageCall
 
     @Override
     public FetchedMessage next() {
-      recordLastOffset();
       FetchedMessage message = delegate.next();
       lastTopicPartition = message.getTopicPartition();
       lastOffset = message.getNextOffset();
-
       messageCount.incrementAndGet();
+      recordOffset();
       return message;
     }
 
@@ -116,13 +119,9 @@ public final class PersistedMessageCallback implements KafkaConsumer.MessageCall
       delegate.remove();
     }
 
-    private void recordLastOffset() {
+    private void recordOffset() {
       if (lastOffset >= 0) {
         offsets.put(lastTopicPartition, lastOffset);
-      }
-      if (messageCount.get() >= persistThreshold) {
-        messageCount.set(0);
-        persist();
       }
     }
   }


gsps1:
http://builds.cask.co/browse/CDAP-DUT2905-2

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/PersistedMessageCallback.java
```diff
@@ -120,10 +112,7 @@ private void recordLastOffset() {
       if (lastOffset >= 0) {
         offsets.put(lastTopicPartition, lastOffset);
       }
-      if (messageCount.get() >= persistThreshold) {
```
sreevatsanraman: Will this cause a performance impact? Have we tested Scale app?

poornachandra: @gsps1 Can you confirm that the offset persisting and writing metrics to HBase happen in the same transaction?

chtyim: Writing to metrics table is non-transactional, right?

poornachandra: Yes, I think so too. Then we need to find out if persisting offsets is transactional or not.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/PersistedMessageCallback.java
```diff
@@ -120,10 +112,7 @@ private void recordLastOffset() {
...
-        messageCount.set(0);
-        persist();
-      }
+      persist();
```
chtyim: I think we shouldn't do it in the Iterator. Writes to metrics table are actually buffered for the whole Iterator (which is controlled by the fetch size when fetching from Kafka). We should only persist the offset after actually writing to metrics table, otherwise it's possible that we lost metrics, due to consumption of this Iterator is just to create Metrics record in memory.

chtyim: Meaning, you should do it in the `onReceived` method instead.

```java
  public void onReceived(Iterator<FetchedMessage> messages) {
    delegate.onReceived(new OffsetTrackingIterator(messages));
    persist();
  }
```

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/PersistedMessageCallback.java
gsps1: Its non-transactional, KafkaMetaTable is also a MetricsTable which is non-transactional.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/PersistedMessageCallback.java
gsps1: you are right.

gsps1:
@chtyim addressed your comment, please review. Thanks

chtyim:
LGTM

gsps1:
on-hold. @poornachandra wanted to discuss about performance before merging this PR.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/PersistedMessageCallback.java
```diff
@@ -60,6 +60,10 @@ public PersistedMessageCallback(KafkaConsumer.MessageCallback delegate,
   @Override
   public void onReceived(Iterator<FetchedMessage> messages) {
     delegate.onReceived(new OffsetTrackingIterator(messages));
+    if (messageCount.get() >= persistThreshold) {
+      messageCount.set(0);
+      persist();
```
poornachandra: It would be a good idea to call this method `persistOffsets`.

cdap-watchdog/src/main/java/co/cask/cdap/metrics/process/PersistedMessageCallback.java
```diff
@@ -117,13 +119,10 @@ public void remove() {
     }
 
     private void recordLastOffset() {
+      messageCount.incrementAndGet();
```
poornachandra: Moving `messageCount.incrementAndGet()` to its original place would make code clearer.

poornachandra:
Two minor comments, please fix and merge.

gsps1:
thanks for the review @poornachandra , addressed the comments, merging.

PR #4189, opened 2015-09-25
CDAP-3860 fixed bugs with available plugins for an artifact

albertshau:
Fixed a bug where plugins in an artifact were not getting included
in the list of plugins available to that artifact. In other words,
only plugins from child artifacts were being included.

Also fixed a bug where listing the plugins for a non-existent
artifact would return an empty list instead of a 404.


albertshau:
http://builds.cask.co/browse/CDAP-RBT495

awholegunch:
I adjusted the case of the exception messages. Please check.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -374,30 +377,41 @@ public ArtifactData apply(DatasetContext<Table> context) throws Exception {
...
         public SortedMap<ArtifactDescriptor, List<PluginClass>> apply(DatasetContext<Table> context) throws Exception {
-          SortedMap<ArtifactDescriptor, List<PluginClass>> result = Maps.newTreeMap();
+          Table table = context.get();
+          SortedMap<ArtifactDescriptor, List<PluginClass>> result = getPluginsInArtifact(table, parentArtifactId);
+          if (result == null) {
+            return null;
+          }
 
-          Scanner scanner = context.get().scan(scanPlugins(parentArtifactId));
+          Scanner scanner = table.scan(scanPlugins(parentArtifactId));
```
bdmogal: ``scanner`` should be closed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/artifact/ArtifactStore.java
```diff
@@ -408,17 +422,24 @@ public ArtifactData apply(DatasetContext<Table> context) throws Exception {
...
         public SortedMap<ArtifactDescriptor, List<PluginClass>> apply(DatasetContext<Table> context) throws Exception {
-          SortedMap<ArtifactDescriptor, List<PluginClass>> result = Maps.newTreeMap();
+          Table table = context.get();
+          SortedMap<ArtifactDescriptor, List<PluginClass>> result = getPluginsInArtifact(table, parentArtifactId);
+          if (result == null) {
+            return null;
+          }
 
-          Scanner scanner = context.get().scan(scanPlugins(parentArtifactId, type));
+          Scanner scanner = table.scan(scanPlugins(parentArtifactId, type));
```
bdmogal: ditto

gokulavasan:
LGTM :+1:

PR #4190, opened 2015-09-25
(CDAP-2988) Added tests for Spark using getDataset() directly.

bdmogal:
Also removed some unused code.

Build: http://builds.cask.co/browse/CDAP-DUT2910-1

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
@@ -19,8 +19,10 @@ package co.cask.cdap.internal.app.runtime.spark;
 import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.common.RuntimeArguments;
 import co.cask.cdap.api.common.Scope;
+import co.cask.cdap.api.dataset.lib.CloseableIterator;
 import co.cask.cdap.api.dataset.lib.FileSet;
 import co.cask.cdap.api.dataset.lib.FileSetArguments;
+import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.dataset.lib.KeyValueTable;
 import co.cask.cdap.api.dataset.lib.ObjectStore;
 import co.cask.cdap.api.dataset.lib.PartitionDetail;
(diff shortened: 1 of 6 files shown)


cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
```diff
@@ -396,6 +400,122 @@ private void testSparkWithCustomFileSet(Class<?> appClass, Class<?> programClass
...
+
+    KeyValueTable logStatsTable = datasetInstantiator.getDataset("logStats");
+    validateGetDatasetOutput(logStatsTable);
+  }
+
+  private void validateGetDatasetOutput(KeyValueTable logStatsTable) {
+    CloseableIterator<KeyValue<byte[], byte[]>> scan = logStatsTable.scan(null, null);
+    try {
+      // must have 4 records
+      Assert.assertTrue(scan.hasNext());
```
rohitsinha54: Can we do something like this: 

```
      int counter = 0;
      while (counter < 4) {
        Assert.assertTrue("Does not have" + counter + "value", scan.hasNext());
        validateKeyValue(scan.next());
        counter++;
      }
```

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/spark/SparkProgramRunnerTest.java
```diff
@@ -396,6 +400,122 @@ private void testSparkWithCustomFileSet(Class<?> appClass, Class<?> programClass
...
+    SparkAppUsingGetDataset.LogStats fredStats2 = new SparkAppUsingGetDataset.LogStats(1, 50);
+    SparkAppUsingGetDataset.LogStats bradStats1 = new SparkAppUsingGetDataset.LogStats(1, 50);
+    SparkAppUsingGetDataset.LogStats bradStats2 = new SparkAppUsingGetDataset.LogStats(1, 50);
+
+    SparkAppUsingGetDataset.LogKey logKey =
+      GSON.fromJson(Bytes.toString(keyValue.getKey()), SparkAppUsingGetDataset.LogKey.class);
+    SparkAppUsingGetDataset.LogStats logStats =
+      GSON.fromJson(Bytes.toString(keyValue.getValue()), SparkAppUsingGetDataset.LogStats.class);
+
+    if (fredKey1.equals(logKey)) {
```
rohitsinha54: I think this if else loop can be simplified a bit if you put the above in a Map and then do contains key with logKey in then get if found to compare value else your assert fail.

rohitsinha54:
Two comments besides that LGTM for merge after build passes.

rohitsinha54:
LGTM

bdmogal:
Build green. Squashed commits. Merging.

PR #4192, opened 2015-09-25
CDAP-3807: Fix case sensitivity handling in DBSink

jwang47:
https://issues.cask.co/browse/CDAP-3807
http://builds.cask.co/browse/CDAP-DUT2912

- [x] Test with Oracle DB


gokulavasan:
LGTM :+1:

PR #4196, opened 2015-09-29
Refactor Metadata Tags Code

elfenheart:


ajainarayanan:
The changes look good. But found this jittery stuff when playing around with the tags. I think we had this from the beginning. Can we have the 'x' icon appear over the text so that the width doesn't exceed causing elements to move left and right when user hovers over the mouse.

tags

ajainarayanan:
Otherwise :+1: LGTM

elfenheart:
The problem with that is that when later we change the behavior of the tags to become links, we won't be able to click the tags as link..we can only remove them

ajainarayanan:
:+1: LGTM

PR #4197, opened 2015-09-29
Documentation Fixes

awholegunch:
Corrects an error in the upgrade instructions, by adding 'cdap-hbase-compat-1.1'.
Corrects an error in the build script that prevents deployment of the ZIP files of the documentation by correcting the case of two global variables.
Updates the MD5 hash for DBSink. Update the documentation for the DBSink plugin to match changes.

Being built at: http://builds.cask.co/browse/CDAP-DDB22-2


awholegunch:
On-hold because DBSink.java has changed and I need to fix the hash.

gokulavasan:
LGTM :+1:

PR #4198, opened 2015-09-29
Fix for output schema for sinks

ajainarayanan:
- Removes output schema for sinks that doesn't have a 'schema' property.
- Adds red asterisk to output schema for sinks that has a required schema property.
- Adds output schema for sink if the node configuration has an implicit schema attached to it.


elfenheart:
:+1: LGTM

PR #4199, opened 2015-09-29
consolidates adapter styling

astewes:
*  Removes etlapps.less
*  Cleans up styling blocks that are no longer used


cdap-ui/app/features/adapters/adapters.less
```diff
@@ -14,42 +14,523 @@
...
+    my-dag {
+      width: 100%;
+      // Dag-Height 60% on smaller screen
+      height: 60vh;
+      @media (min-height: 800px) {
+        height: 40vh;
+      }
+    }
+
+    .tab-pane .panel-default {
```
ajainarayanan: are we using panel-default in adapter-create state?

cdap-ui/app/features/adapters/adapters.less
```diff
@@ -14,42 +14,523 @@
...
+          }
+        }
+      }
+    }
+
+    .col-xs-12.text-right {
+      > .btn.btn-default { margin-bottom: 20px; }
+    }
+
+    .panel.panel-default {
```
ajainarayanan: are we using panel-default in adapter-create state?

cdap-ui/app/features/adapters/adapters.less
```diff
@@ -14,42 +14,523 @@
...
+      @media (min-height: 800px) {
+        height: 40vh;
+      }
+    }
+
+    .tab-pane .panel-default {
+      .panel-heading > span.panel-title { line-height: 30px; }
+    }
+
+    .as-sortable-drag {
```
ajainarayanan: I believe we no longer use as-sortable directive. I think we can remove it.

cdap-ui/app/features/adapters/adapters.less
```diff
@@ -14,42 +14,523 @@
...
       }
     }
-  }
-  &.state-adapters {
 
-    fieldset.view-plugin-json {
-      textarea {
-        resize: none;
-        height: 60vh;
+    .modal .modal-header {
```
ajainarayanan: Do we need modal classes here? Didn't we extract out modal styles to themes.less? I thought we decided to extract out modal styles out to a separate theme file, correct me if I am wrong.

cdap-ui/app/features/adapters/adapters.less
```diff
@@ -14,42 +14,523 @@
...
+          display: block;
+          &:first-child {
+            font-size: 15px;
+          }
+          &:last-child {
+            font-size: 18px;
+          }
+        }
+      }
+      ul {
```
ajainarayanan: Do we need ul and li styles for adapter list view?

cdap-ui/app/features/adapters/adapters.less
```diff
@@ -14,42 +14,523 @@
...
+          border: 1px dashed black;
+          color: black;
+          background-color: white;
+          text-align: center;
+          font-size: 18px;
+        }
+      }
+    }
+
+    .plugins {
```
ajainarayanan: plugins css class too. Are we still using it in hydrator view?

cdap-ui/app/features/adapters/adapters.less
elfenheart: I think the <ul> and <li> are for the total count on the status bar. @astewes correct me if I'm wrong?

astewes: Correct, it's for the status bar.

cdap-ui/app/features/adapters/adapters.less
astewes: There was only one style block that needed to be kept, and since it's tied to validators, I've moved it to validators.less.

cdap-ui/app/features/adapters/adapters.less
astewes: Modal classes are (still) there for Export, which is the only modal being used in Hydrator at this point. I plan to remove it in the PR that consolidates modal styling.

ajainarayanan:
:+1: LGTM

PR #4200, opened 2015-09-29
Metadata and Lineage clients

anwar6953:
Implement `MetadataClient` and `LineageClient`.
Update metadata/lineage test cases to utilize these.
Update `MetadataHttpHandler` to respond with 400 (instead of 500), if null metadata is given in request.

The methods exposed in `MetadataClient` are:
`searchMetadata`
For each of the following program entity types: `Application`, `DatasetInstance`, `Stream`, `Program`:
`getMetadata`
`addTags`
`addProperties`
`removeProperties`
`removeTags`
`removeMetadata`

https://issues.cask.co/browse/CDAP-3889
http://builds.cask.co/browse/CDAP-DUT2918-4

### DIFF
--- a/cdap-app-fabric/pom.xml
+++ b/cdap-app-fabric/pom.xml
@@ -92,6 +92,13 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-client</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>co.cask.cdap</groupId>
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
@@ -37,6 +37,7 @@ import org.jboss.netty.buffer.ChannelBufferInputStream;
 import org.jboss.netty.handler.codec.http.HttpRequest;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
 
+import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.lang.reflect.Type;
@@ -521,31 +522,41 @@ public class MetadataHttpHandler extends AbstractHttpHandler {
   }
 
   private Map<String, String> readMetadata(HttpRequest request) throws BadRequestException {
-    try {
-      ChannelBuffer content = request.getContent();
-      Preconditions.checkArgument(content.readable());
-      try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
-        Map<String, String> metadata = GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
-        Preconditions.checkNotNull(metadata);
-        return metadata;
-      }
-    } catch (Throwable t) {
-      throw new BadRequestException("Unable to read business metadata from the request.", t);
+    ChannelBuffer content = request.getContent();
+    if (!content.readable()) {
+      throw new BadRequestException("Unable to read business metadata from the request.");
     }
+
+    Map<String, String> metadata;
+    try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
+      metadata = GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
+    } catch (IOException e) {
+      throw new BadRequestException("Unable to read business metadata from the request.", e);
+    }
+
+    if (metadata == null) {
+      throw new BadRequestException("Null metadata was read from the request");
+    }
+    return metadata;
   }
 
   private String[] readArray(HttpRequest request) throws BadRequestException {
-    try {
-      ChannelBuffer content = request.getContent();
-      Preconditions.checkArgument(content.readable());
-      try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
-        List<String> toReturn = GSON.fromJson(reader, LIST_STRING_TYPE);
-        Preconditions.checkNotNull(toReturn);
-        return toReturn.toArray(new String[toReturn.size()]);
-      }
-    } catch (Throwable t) {
-      throw new BadRequestException("Unable to read a list of keys from the request.", t);
+    ChannelBuffer content = request.getContent();
+    if (!content.readable()) {
+      throw new BadRequestException("Unable to read a list of keys from the request.");
     }
+
+    List<String> toReturn;
+    try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
+      toReturn = GSON.fromJson(reader, LIST_STRING_TYPE);
+    } catch (IOException e) {
+      throw new BadRequestException("Unable to read a list of keys from the request.", e);
+    }
+
+    if (toReturn == null) {
+      throw new BadRequestException("Null value was read from the request.");
+    }
+    return toReturn.toArray(new String[toReturn.size()]);
   }
 
   // *** Search endpoints ***
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
@@ -27,6 +27,7 @@ import co.cask.cdap.proto.metadata.MetadataSearchTargetType;
 import co.cask.http.AbstractHttpHandler;
 import co.cask.http.HttpResponder;
 import com.google.common.base.Charsets;
+import com.google.common.base.Preconditions;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
@@ -521,30 +522,30 @@ public class MetadataHttpHandler extends AbstractHttpHandler {
   }
 
   private Map<String, String> readMetadata(HttpRequest request) throws BadRequestException, IOException {
-    ChannelBuffer content = request.getContent();
-    if (!content.readable()) {
-      throw new BadRequestException("Unable to read business metadata from request.");
-    }
-    try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
-      Map<String, String> metadata = GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
-      if (metadata == null) {
-        throw new BadRequestException("Null metadata was read from request.");
+    try {
+      ChannelBuffer content = request.getContent();
+      Preconditions.checkArgument(content.readable());
+      try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
+        Map<String, String> metadata = GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
+        Preconditions.checkNotNull(metadata);
+        return metadata;
       }
-      return metadata;
+    } catch (Throwable t) {
+      throw new BadRequestException("Unable to read business metadata from the request.", t);
     }
   }
 
   private String[] readArray(HttpRequest request) throws BadRequestException, IOException {
-    ChannelBuffer content = request.getContent();
-    if (!content.readable()) {
-      throw new BadRequestException("Unable to read a list of keys from the request.");
-    }
-    try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
-      List<String> toReturn = GSON.fromJson(reader, LIST_STRING_TYPE);
-      if (toReturn == null) {
-        throw new BadRequestException("Null value was read from request.");
+    try {
+      ChannelBuffer content = request.getContent();
+      Preconditions.checkArgument(content.readable());
+      try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
+        List<String> toReturn = GSON.fromJson(reader, LIST_STRING_TYPE);
+        Preconditions.checkNotNull(toReturn);
+        return toReturn.toArray(new String[toReturn.size()]);
       }
-      return toReturn.toArray(new String[toReturn.size()]);
+    } catch (Throwable t) {
+      throw new BadRequestException("Unable to read a list of keys from the request.", t);
     }
   }
 
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
@@ -37,7 +37,6 @@ import org.jboss.netty.buffer.ChannelBufferInputStream;
 import org.jboss.netty.handler.codec.http.HttpRequest;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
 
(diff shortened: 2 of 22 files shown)


cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
```diff
@@ -526,7 +526,11 @@ public void removeStreamTags(HttpRequest request, HttpResponder responder,
       throw new BadRequestException("Unable to read business metadata from request.");
     }
     try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
-      return GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
+      Map<String, String> metadata = GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
```
gsps1: not related to your change, but i think it makes sense to try ``fromJson`` for invalid JSON, instead of throwing the exception, currently we return Internal_Server_Error on an invalid json format, while we should return Bad_Request.

cdap-client/src/main/java/co/cask/cdap/client/LineageClient.java
```diff
@@ -0,0 +1,143 @@
...
+  private final ClientConfig config;
+
+  @Inject
+  public LineageClient(ClientConfig config, RESTClient restClient) {
+    this.config = config;
+    this.restClient = restClient;
+  }
+
+  public LineageClient(ClientConfig config) {
+    this.config = config;
```
gsps1: ``this(config, new RESTClient(config))``

cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java
```diff
@@ -0,0 +1,473 @@
...
+    }
+  }
+
+  /**
+   * Removes properties from an application.
+   *
+   * @param appId app to remove properties from
+   * @param propertyToRemove property to be removed, or {@code null} to remove all properties for the entity
+   */
+  public void removeProperties(Id.Application appId, @Nullable String propertyToRemove)
```
gsps1: we are just remove only property right ? shouldn't the name of this method be ``removeProperty`` ?

cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java
```diff
@@ -0,0 +1,473 @@
...
+                                programId.getApplicationId(), programId.getType().getCategoryName(), programId.getId());
+    removeProperties(programId, path, propertyToRemove);
+  }
+
+  private void removeProperties(Id.NamespacedId namespacedId, String path, @Nullable String propertyToRemove)
+    throws IOException, UnauthorizedException, NotFoundException, BadRequestException {
+    if (propertyToRemove != null) {
+      path = path + "/" + propertyToRemove;
+    }
+    URL tagURL = config.resolveNamespacedURLV3(namespacedId.getNamespace(), path);
```
gsps1: ``propertiesURL``

cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java
```diff
@@ -0,0 +1,473 @@
...
+    String path = String.format("apps/%s/metadata", appId.getId());
+    return getMetadata(appId, path);
+  }
+
+  /**
+   * @param datasetInstance the dataset for which to retrieve metadata
+   * @return The metadata for the dataset.
+   */
+  public Set<MetadataRecord> getMetadata(Id.DatasetInstance datasetInstance)
+    throws IOException, UnauthorizedException, NotFoundException, BadRequestException {
```
gsps1: there is a lot of similar methods with parameter type change, can we enumerate them in an Enum and use a switch case to generate appropriate URL's ? what do you think?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
```diff
@@ -521,23 +522,30 @@ public void removeStreamTags(HttpRequest request, HttpResponder responder,
   }
 
   private Map<String, String> readMetadata(HttpRequest request) throws BadRequestException, IOException {
```
sagarkapare: This method no longer throws `IOException`, right?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
```diff
@@ -521,23 +522,30 @@ public void removeStreamTags(HttpRequest request, HttpResponder responder,
...
-    ChannelBuffer content = request.getContent();
-    if (!content.readable()) {
-      throw new BadRequestException("Unable to read business metadata from request.");
-    }
-    try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
-      return GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
+    try {
+      ChannelBuffer content = request.getContent();
+      Preconditions.checkArgument(content.readable());
```
sagarkapare: I think the previous code to return the BadRequestException here makes more sense and its more explicit.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
anwar6953: right!

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
```diff
@@ -521,23 +522,30 @@ public void removeStreamTags(HttpRequest request, HttpResponder responder,
...
+      try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
+        Map<String, String> metadata = GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
+        Preconditions.checkNotNull(metadata);
+        return metadata;
+      }
+    } catch (Throwable t) {
```
sagarkapare: No need to catch Throwable here. BadRequestException and IOException can be propagated to the HttpExceptionHandler directly.

cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java
anwar6953: if null is passed in, it removes all properties for that entity.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
anwar6953: but the IOException is from `GSON.fromJson()`, which is actually a BadRequestException.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
```diff
@@ -521,23 +522,30 @@ public void removeStreamTags(HttpRequest request, HttpResponder responder,
...
-    ChannelBuffer content = request.getContent();
-    if (!content.readable()) {
-      throw new BadRequestException("Unable to read a list of keys from the request.");
-    }
-    try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
-      List<String> toReturn = GSON.fromJson(reader, LIST_STRING_TYPE);
-      return toReturn.toArray(new String[toReturn.size()]);
+    try {
+      ChannelBuffer content = request.getContent();
+      Preconditions.checkArgument(content.readable());
```
sagarkapare: same here. 
```java
if (!content.readable()) {
      throw new BadRequestException("Unable to read a list of keys from the request.");
}
```
would be better, I think

anwar6953: I changed it from that because there are 3 assertions in this method that we need to do and throw BadRequestException if any of those 3 assertions fail:
1) content.isReadable()
2) GSON.fromJson - potential failures, if client doesn't use proper json format
3) GSON.fromJson - returning null

So, instead of individually wrapping each of these things with something like:
```
try {
  GSON.fromJson( ...
catch (JSONException... ) {
  throw new BadRequestException
}
```
I thought its better to wrap the whole method in a try-catch, and throw BadRequestException if any Exception happens while parsing the request body.

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
sagarkapare: yes, but in that case we can do 
```java
Map<String, String> metadata;
try {
   metadata = GSON.fromJson(reader, MAP_STRING_STRING_TYPE);
} catch (IOException e) {
   throw new BadRequestExcepton("Invalid json provided", e);
}
```

cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java
gsps1: that sounds a bit dangerous, we should have separate endpoint for remove all properties right ?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
anwar6953: My concern was that in this small method, we were checking and throwing BadRequestException in three different places.

cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java
anwar6953: I refactored the string formatting to `constructPath` methods, which simplifies most of the methods, though I'm not sure how to reduce the number of methods further.

cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java
anwar6953: It is a separate endpoint, just that both endpoints are serviced by this method in the client.
It should also be separate methods in the client, you mean?

bdmogal: @anwar6953, since this is a public client, it would be better to add methods like ``removeProperties(Id.Application appId)`` that just delegate to ``removeProperties(appId, null)``, so the API is simpler for users.

anwar6953: Added additional methods for `removeTag` and `removeProperty` methods.
Changed the `removeTags` and `removeProperties` methods to remove the `@Nullable` annotation from the second parameter.

cdap-app-fabric/pom.xml
```diff
@@ -92,6 +92,13 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-client</artifactId>
```
bdmogal: this seems a little weird. why is this needed?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
bdmogal: Throwing BadRequestException multiple times is ok. I think it is better than throwing an IllegalArgumentException or a NullPointerException, catching it and then throwing a BadRequestException?

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
```diff
@@ -520,24 +520,31 @@ public void removeStreamTags(HttpRequest request, HttpResponder responder,
...
+      try (Reader reader = new InputStreamReader(new ChannelBufferInputStream(content), Charsets.UTF_8)) {
+        List<String> toReturn = GSON.fromJson(reader, LIST_STRING_TYPE);
+        Preconditions.checkNotNull(toReturn);
+        return toReturn.toArray(new String[toReturn.size()]);
+      }
+    } catch (Throwable t) {
+      throw new BadRequestException("Unable to read a list of keys from the request.", t);
```
bdmogal: same comment as above

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
anwar6953: done.

cdap-app-fabric/pom.xml
anwar6953: We use the Metadata/Lineage clients in MetadataHttpHandlerTestBase.java

cdap-app-fabric/src/main/java/co/cask/cdap/metadata/MetadataHttpHandler.java
anwar6953: done.

cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java
jwang47: You could maintain a map from Class<Id> to Function<Id, String>, but I don't think it's very helpful in this case where there aren't too many cases. In Java 8 it would be slightly nicer, even in this case. E.g:

```
paths = ImmutableMap.builder()
  .put(Id.DatasetInstance.class, (id) -> String.format("datasets/%s", id.getId()))
  .put(Id.Stream.class, (id) -> String.format("streams/%s", streamId.getId()))
  .put(..)
  .build();

String constructPath(Id id) {
  if (!paths.containsKey(id)) { throw ??; }
  return paths.get(id).apply(id);
}
```

So it's probably fine as is.

jwang47: Oh, actually I just saw all of the `removeProperty` methods. I think it might be better to use the map of Class to Function here (may need some casting):

```
paths = ImmutableMap<Class<? extends Id>, Function<? extends Id, String>.builder()
  .put(Id.DatasetInstance.class, new Function<? extends Id, String>() {
    public String apply(DatasetInstanceId id) {
      return String.format("datasets/%s", id.getId());
    }
  })
  .put(Id.Stream.class, new Function<? extends Id, String>() {
    public String apply(DatasetInstanceId id) {
      return String.format("streams/%s", id.getId());
    }
  })
  .put(..)
  .build();

String constructPath(Id id) {
  if (!paths.containsKey(id)) { throw ??; }
  return paths.get(id).apply(id);
}

Comment (anwar6953) [inline cdap-client/src/main/java/co/cask/cdap/client/MetadataClient.java]: Hey @jwang47, that doesn't reduce the number of removeProperty methods, right?
It just replaces the constructPath methods into a map containing several functions and one constructPath method, right?

How is that better than having the functions be their own methods, and having the java compiler match the Id type to method (by the method's signature) rather than have a map of all the functions and do the mapping ourself?

For instance, in the PR, I have this:
```java
  private String constructPath(Id.Application appId) {
    return String.format("apps/%s", appId.getId());
  }

  private String constructPath(Id.Program programId) {
    return String.format("apps/%s/%s/%s",
                         programId.getApplicationId(), programId.getType().getCategoryName(), programId.getId());
  }

  private String constructPath(Id.Run runId) {
    Id.Program programId = runId.getProgram();
    return String.format("apps/%s/%s/%s/runs/%s",
                         programId.getApplicationId(), programId.getType().getCategoryName(), programId.getId(),
                         runId.getId());
  }

  private String constructPath(Id.DatasetInstance datasetInstance) {
    return String.format("datasets/%s", datasetInstance.getId());
  }

  private String constructPath(Id.Stream streamId) {
    return String.format("streams/%s", streamId.getId());
  }
```

jwang47:
LGTM

PR #4201, opened 2015-09-30
improves styling of select in ff, safari

astewes:
*  Replaces native browser styling of select dropdowns with custom styling

ops


ajainarayanan:
:+1: LGTM. Had one issue with select dropdowns in firefox (looks different compared to other browsers). Apart from that :+1:

ajainarayanan:
Talked to @astewes . Its a general issue with firefox and safari for the dropdowns to be styled. Hence merging this PR as other changes looks good.

PR #4203, opened 2015-09-30
Fix Documentation for 3.2.0

awholegunch:
Add a conditional text block to the database sink reference page as these changes do not take affect until 3.2.1.

The PR that was merged yesterday fixed the documentation build, but it inadvertently added a change that doesn't take affect until 3.2.1. To allow the documentation to be built as "3.2.0", yet contain the changes required for 3.2.1, this adds a conditional test for including the affected paragraph. Once the version changes to greater than 3.2.0, the paragraph will be included.

If other documentation manuals need fixes such as this, this code could be moved to the common conf.py and accessed from all.

Building at: http://builds.cask.co/browse/CDAP-DDB23-1

### DIFF
--- a/cdap-docs/included-applications/source/conf.py
+++ b/cdap-docs/included-applications/source/conf.py
@@ -19,3 +19,13 @@ html_context = {"html_short_title_toc":html_short_title_toc}
 
 # Remove this guide from the mapping as it will fail as it has been deleted by clean
 intersphinx_mapping.pop("includedapps", None)
+
+# Add a custom config value that can be used for conditional content
+def setup(app):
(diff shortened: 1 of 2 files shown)


gokulavasan:
Changes LGTM. But what if we have releases 3.2.2, 3.2.3 etc which require different docs from 3.2.1?

awholegunch:
If that happens, we'll change this. We can add more conditionals and tests as required. I can easily imagine having to re-write this in the future, as it doesn't strike me as robust. But I think this will be easier to solve it when the need actually arises rather trying to anticipate the need now. For instance, once release 3.2.1 comes out, this code can be dropped.

PR #4206, opened 2015-09-30
Replace default descriptions.

awholegunch:
Replaces the default descriptions used in the ETLBatch and ETLRealtime applications with updated text.

Replaces usage of "Application Template" and its ilk.

Build and Test running at: http://builds.cask.co/browse/CDAP-RBT496
Passed build and test.


awholegunch:
On-hold because I built with the wrong plan. Fixing.

gokulavasan:
LGTM :+1:

PR #4207, opened 2015-09-30
CDAP-3900 Return empty list in case of logs are requested for the deleted program runs.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-3900
Build: http://builds.cask.co/browse/CDAP-DUT2925-1

It is possible that the log files for the requested range are already deleted, in case of old program runs. For such requests both the start and toTimeMs will fall outside the range sortedFiles. In that case return empty list.


poornachandra:
LGTM

sagarkapare:
Build passed. Merging.

PR #4208, opened 2015-09-30
Fixes to continue polling when frequence or duration is changed

ajainarayanan:
In Ops dashboard, on change in frequency or duration the polling should automatically change without forcing the user to stop and restart the polling again.


elfenheart:
:+1: LGTM

PR #4210, opened 2015-09-30
CDAP-3901: Remove namespace existence checks in DatasetInstanceService

jwang47:
https://issues.cask.co/browse/CDAP-3901
http://builds.cask.co/browse/CDAP-DUT2922

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceService.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceService.java
@@ -99,8 +99,6 @@ public class DatasetInstanceService {
    * @throws IOException if there is a problem in making an HTTP request to check if the namespace exists.
    */
   public DatasetMeta get(Id.DatasetInstance instance, List<? extends Id> owners) throws Exception {
-    // Throws NamespaceNotFoundException if the namespace does not exist
-    ensureNamespaceExists(instance.getNamespace());
     DatasetSpecification spec = instanceManager.get(instance);
     if (spec == null) {
       throw new NotFoundException(instance);


cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandler.java
```diff
@@ -209,11 +232,19 @@ public void executeAdmin(HttpRequest request, HttpResponder responder, @PathPara
...
+
+    Id.Namespace namespace = Id.Namespace.from(namespaceId);
+    ensureNamespaceExists(namespace);
+
     // todo: execute data operation
     responder.sendStatus(HttpResponseStatus.NOT_IMPLEMENTED);
   }
 
+  private void ensureNamespaceExists(Id.Namespace namespace) throws Exception {
+    namespaceAdmin.get(namespace);
```
bdmogal: you will need special handling for ``Id.Namespace.SYSTEM``, like there was in ``DatasetInstanceService``

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceService.java
```diff
@@ -241,9 +228,6 @@ public void drop(Id.DatasetInstance instance) throws Exception {
    * @throws IOException if there was a problem in checking if the namespace exists over HTTP
    */
   public DatasetAdminOpResponse executeAdmin(Id.DatasetInstance instance, String method) throws Exception {
-    // Throws NamespaceNotFoundException if the namespace does not exist
-    ensureNamespaceExists(instance.getNamespace());
```
bdmogal: why were the namespace checks moved to the handler?

bdmogal:
@jwang47 , ``DatasetService`` used ``AbstractNamespaceClient`` instead of ``NamespaceAdmin``, because ``NamespaceAdmin`` uses the ``Store`` to access ``app.meta`` directly, and ``DatasetService`` does not have access to the ``Store``, since it is an app-fabric class. As a result, the binding to ``Store`` needed in ``DefaultNamespaceAdmin`` may not be available in certain scenarios. Please be sure to test in all modes (test, standalone and distributed) before making the switch to ``NamespaceAdmin``

jwang47:
@bdmogal revised to simply take out ns check in DatasetInstanceService.get()

poornachandra:
@bdmogal Is there anyway to figure out if a namespace exists from Dataset Service itself, without making a remote call?

jwang47:
The namespace check is indirectly done via DefaultStore.getNamespace(). From DatasetInstanceService, we could either (1) pass an instance of the app.meta dataset to DefaultStore instead of using `DatasetsUtil.getOrCreateDataset()` with a `RemoteDatasetFramework` or (2) pass a `DatasetFramework` implementation to `DatasetsUtil.getOrCreateDataset()` that doesn't make remote calls.

jwang47:
Actually, I don't really understand why we need to make remote calls for datasets, at least for acquiring a dataset instance. Perhaps we can eliminate some remote calls for datasets.

jwang47:
Will instead opt to make a DatasetFramework implementation that directly uses `DatasetInstanceService` instead using `RemoteDatasetFramework` which executes operations by making HTTP requests via `DatasetServiceClient`.

PR #4215, opened 2015-10-01
Fix messages.

awholegunch:
Fixes a number of simple errors in the UI messages. Adds periods, removes spaces, restructures one of the messages...

### DIFF
--- a/cdap-ui/app/features/adapters/controllers/create/toppanel-ctrl.js
+++ b/cdap-ui/app/features/adapters/controllers/create/toppanel-ctrl.js
@@ -34,7 +34,7 @@ angular.module(PKG.name + '.feature.adapters')
       } else {
         $alert({
           type: 'danger',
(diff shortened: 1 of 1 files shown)


elfenheart:
:+1: LGTM

awholegunch:
I guess we should run a build, just to make sure....

awholegunch:
http://builds.cask.co/browse/CDAP-RBT497-1

cdap-ui/app/features/adapters/controllers/create/toppanel-ctrl.js
```diff
@@ -110,7 +110,7 @@ angular.module(PKG.name + '.feature.adapters')
                       });
                     }.bind(this),
                     function error() {
-                      console.log('ERROR: ' + 'exporting adapter failed');
+                      console.log('ERROR: Exporting adapter ' + MyAppDAGService.metadata.name + ' failed.');
```
gokulavasan: There is no adapter in 3.2.

PR #4219, opened 2015-10-02
Implement nextRuntimes(Id.Workflow) in ScheduleClient. Also made the …

anwar6953:
…workflow http handler method properly respond with 404 (instead of empty list) if a nonexistent workflow is given.

https://issues.cask.co/browse/CDAP-3616
http://builds.cask.co/browse/CDAP-DUT2928-1

### DIFF
--- a/cdap-client-tests/src/test/java/co/cask/cdap/client/ScheduleClientTestRun.java
+++ b/cdap-client-tests/src/test/java/co/cask/cdap/client/ScheduleClientTestRun.java
@@ -20,9 +20,11 @@ import co.cask.cdap.api.schedule.ScheduleSpecification;
 import co.cask.cdap.client.app.FakeApp;
 import co.cask.cdap.client.app.FakeWorkflow;
 import co.cask.cdap.client.common.ClientTestBase;
+import co.cask.cdap.common.NotFoundException;
 import co.cask.cdap.internal.schedule.StreamSizeSchedule;
 import co.cask.cdap.internal.schedule.TimeSchedule;
 import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.ScheduledRuntime;
 import co.cask.cdap.test.XSlowTests;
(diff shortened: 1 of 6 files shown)


cdap-client-tests/src/test/java/co/cask/cdap/client/ScheduleClientTestRun.java
```diff
@@ -91,5 +94,17 @@ public void testAll() throws Exception {
     scheduleClient.suspend(schedule);
     status = scheduleClient.getStatus(schedule);
     Assert.assertEquals("SUSPENDED", status);
+
+    List<ScheduledRuntime> scheduledRuntimes = scheduleClient.nextRuntimes(workflow);
+    Assert.assertEquals(1, scheduledRuntimes.size());
+    // simply assert that its scheduled for some time in the future (or scheduled for now, but hasn't quite
+    // executed yet
+    Assert.assertTrue(scheduledRuntimes.get(0).getTime() >= System.currentTimeMillis() + TimeUnit.MINUTES.toMillis(1));
```
sagarkapare: I think, this would fail if the test case is executed on 31st December between 11:59 PM and midnight :)

sagarkapare:
major comment, otherwise LGTM :+1:

cdap-client-tests/src/test/java/co/cask/cdap/client/ScheduleClientTestRun.java
anwar6953: Hmm, could you explain your reasoning?
I thought I made it handle that case.

anwar6953: Ah right, I did ` + 1 minute`, when I meant to do ` - 1 minute`.
Updated.

sagarkapare:
:+1:

PR #4226, opened 2015-10-02
replaces image with animated icon

astewes:
*  Replaces animated gif with animated font icon

fist
.3


elfenheart:
@astewes so the fist is not bumping anymore?

ajainarayanan:
It looks super cool to me! :+1: Have to make smoother in chrome . Right now the animation looks smooth in FF and safari. If for chrome this PR is good to go!

PR #4227, opened 2015-10-03
Implement IntegrationTestManager#getQueryClient, addDatasetInstance. …

anwar6953:
Implement methods of IntegrationTestManager:
getQueryClient, addDatasetInstance, addDatasetModule.

Update RESTClient to throw an exception if service is continuously unavailable. Previously, if the retries for service unavailable run out, it would return the response instead of throwing an exception.
Also increased the timeout between retries from 100ms to 1 second.

https://issues.cask.co/browse/CDAP-2641
http://builds.cask.co/browse/CDAP-DUT2932-3


cdap-integration-test/src/main/java/co/cask/cdap/test/IntegrationTestManager.java
```diff
@@ -259,21 +273,36 @@ public void clear() throws Exception {
...
+                            datasetModule.getName(),
+                            createModuleJarFile(datasetModule));
+  }
+
+  private File createModuleJarFile(Class<?> cls) throws IOException {
+    String version = String.format("1.0.%d-SNAPSHOT", System.currentTimeMillis());
+    File moduleJarFile = new File(tmpFolder, String.format("%s-%s.jar", cls.getSimpleName(), version));
+    Location deploymentJar = AppJarHelper.createDeploymentJar(locationFactory, cls);
+    Files.copy(Locations.newInputSupplier(deploymentJar), moduleJarFile);
+    return moduleJarFile;
```
jwang47: Is this tested?

anwar6953: Yea, the test case deploys the module before deploying the app.
The app requires the module is there.

jwang47:
LGTM

PR #4228, opened 2015-10-03
Add project versioning to READMEs

awholegunch:
This adds project version and filtering to the CDAP SDK /examples README.

To the main SDK README, it adds some additional instances of using the versioning to set correct paths to the website based on the version. This will help people end up at the correct version of the examples and the documentation for the version of the SDK that they are using.

It adds additional and updated descriptions of the examples to reflect recent changes, and includes a description for all but the CubeService (which is not a complete example; it will be dealt with in another PR.)

Building at: http://builds.cask.co/browse/CDAP-RBT498-1

### DIFF
--- a/cdap-examples/README.rst
+++ b/cdap-examples/README.rst
@@ -49,9 +49,9 @@ HelloWorld
 ----------
 - This is a simple HelloWorld example that uses a stream, a dataset, a flow, and a
   service.
-- A dataset, a KeyValueTable.
-- A flow, with a single flowlet that reads from the stream and stores each name in the KeyValueTable.
-- A service, that reads the name from the KeyValueTable and responds with "Hello [Name]!"
+- A dataset created using a KeyValueTable.
+- A flow with a single flowlet that reads from the stream and stores each name in the KeyValueTable.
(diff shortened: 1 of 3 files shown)


cdap-examples/README.rst
```diff
@@ -29,85 +29,106 @@ CountRandom
...
 ----------
-- This is a simple HelloWorld example that uses one Stream, one Dataset, one Flow and one
-  Service.
-- A Dataset, a KeyValueTable.
-- A Flow, with a single Flowlet that reads from the Stream and stores each name in the KeyValueTable.
-- A Service, that reads the name from the KeyValueTable and responds with "Hello [Name]!"
+- This is a simple HelloWorld example that uses a stream, a dataset, a flow, and a
+  service.
+- A dataset, a KeyValueTable.
+- A flow, with a single flowlet that reads from the stream and stores each name in the KeyValueTable.
```
anwar6953: Do we need a comma after 'flow'?
And after 'service' in the next line?

awholegunch: No, we don't. Good idea.

awholegunch:
@anwar6953 I changed the text; please take another look.

anwar6953:
LGTM

PR #4230, opened 2015-10-05
(CDAP-3627) Enhanced http service handler file upload support

chtyim:
- Introduced HttpContentConsumer class for handling upload content in chunks
  - Also NonTransactionalHttpContentConsumer for non-transactional upload chunks handling
    - For faster performance
- Introduced Transactional interface to unify function execution inside transaction


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
```diff
@@ -0,0 +1,195 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Multimap;
+import com.ning.http.client.HttpContent;
+import org.jboss.netty.buffer.ChannelBuffer;
+
+import java.nio.ByteBuffer;
+import javax.annotation.Nullable;
+
+/**
+ * An adapter class to adapter calls to {@link BodyConsumer} to {@link HttpContentConsumer}.
```
anwar6953: Does this javadoc seem more appropriate?
An adapter class to **delegate** calls **from** {@link BodyConsumer} to {@link HttpContentConsumer}.

chtyim: Fixed

cdap-unit-test/src/test/java/co/cask/cdap/service/FileUploadApp.java
```diff
@@ -0,0 +1,245 @@
...
+          chunk.reset();
+          channel.write(chunk);
+        }
+
+        @Override
+        public void onFinish(HttpServiceResponder responder) throws Exception {
+          channel.close();
+          String uploadedMd5 = BaseEncoding.base64().encode(messageDigest.digest());
+          if (!md5.equals(uploadedMd5)) {
+            throw new IllegalArgumentException("MD5 not match. Expected '" + md5 + "', received '" + uploadedMd5 + "'");
```
anwar6953: just wondering - if this exception is thrown, does the `onError` callback below get triggered?

chtyim: yes

cdap-unit-test/src/test/java/co/cask/cdap/service/FileUploadServiceTestRun.java
```diff
@@ -0,0 +1,120 @@
...
+        iterSize++;
+      }
+      // The iterator should have size >= 2, since we uses different chunk size for two different upload
+      Assert.assertTrue(iterSize >= 2);
+
+      // The sum of all chunks sizes should be the same as the
+      // content size * 2 (since we have one failure, one success upload)
+      Assert.assertEquals(content.length * 2, sum);
+
+    } finally {
```
anwar6953: stopping programs isn't necessary in user test cases anymore.
All programs will get stopped in: TestBase#afterTest

chtyim: I know, but I prefer better resource management (if new code are added, it won't run inside unintended resource contention that easily).

cdap-unit-test/src/test/java/co/cask/cdap/service/FileUploadApp.java
anwar6953: :+1:

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2913-2

cdap-api/src/main/java/co/cask/cdap/api/service/http/NonTransactionalHttpContentConsumer.java
```diff
@@ -0,0 +1,51 @@
...
+import co.cask.cdap.api.dataset.Dataset;
+
+import java.nio.ByteBuffer;
+
+/**
+ * A {@link HttpContentConsumer} that supports non-transactional processing of HTTP request body.
+ * One can extends from this class and implements the {@link #onReceived(ByteBuffer, Transactional)} method
+ * to achieve better performance when receiving data from the client.
+ */
+public abstract class NonTransactionalHttpContentConsumer extends HttpContentConsumer {
```
sreevatsanraman: It is a bit confusing to call this class `NonTransactionalHttpContentConsumer` and expecting people to implement the `onReceived(ByteBuffer, Transactional)` method

chtyim: Why? And what's the better naming suggestion?

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Instance of this class is for consuming HTTP request body incrementally. Methods defined in
+ * {@link HttpServiceHandler} can return an instance of this class to consume request body in small chunks
+ * to avoid running out of memory for requests with large body.
+ *
+ * Example:
+ *
+ * <p>
+ *   <pre><code>
+ *      public class MyHttpHanlder extends AbstractHttpServiceHandler {
```
anew: typo: MyHttpHandler?

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+ * </p>
+ */
+public abstract class HttpContentConsumer {
+
+  /**
+   * This method will get invoked when a new chunk of the request body is available to be consumed.
+   * It is guaranteed that no concurrent calls to this method will be made.
+   * <p>
+   * Also note that each invocation
+   * to this method is executed inside a single transaction, hence access to transactional {@link Dataset}
```
anew: datasets?

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+   * is possible through the {@link HttpServiceContext} provided to the
+   * {@link HttpServiceHandler#initialize(HttpServiceContext)} method.
+   * </p>
+   *
+   * If transaction is not always needed, you can extend from the {@link NonTransactionalHttpContentConsumer} class and
```
anew: a transaction

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+   * {@link HttpServiceResponder} to send respond in order to complete the HTTP call. This method is
+   * always executed inside a single transaction.
+   *
+   * @param responder a {@link HttpServiceResponder} for sending respond
+   * @throws Exception if there is any error
+   */
+  public abstract void onFinish(HttpServiceResponder responder) throws Exception;
+
+  /**
+   * This method will get invoked when there is error while processing the request body chunks. If {@link Exception}
```
anew: an error

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+   * are being thrown from either the {@link #onReceived(ByteBuffer)} or {@link #onFinish(HttpServiceResponder)}
+   * methods, this method will get called. This method is always executed inside a single transaction.
+   *
+   * @param responder a {@link HttpServiceResponder} for sending respond
```
anew: response

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+   * @throws Exception if there is any error when processing the received chunk
+   */
+  public abstract void onReceived(ByteBuffer chunk) throws Exception;
+
+  /**
+   * This method will get invoked when reached the end of the request body. It must use the given
+   * {@link HttpServiceResponder} to send respond in order to complete the HTTP call. This method is
+   * always executed inside a single transaction.
+   *
+   * @param responder a {@link HttpServiceResponder} for sending respond
```
anew: response

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * This method will get invoked when there is error while processing the request body chunks. If {@link Exception}
+   * are being thrown from either the {@link #onReceived(ByteBuffer)} or {@link #onFinish(HttpServiceResponder)}
+   * methods, this method will get called. This method is always executed inside a single transaction.
+   *
+   * @param responder a {@link HttpServiceResponder} for sending respond
+   * @param failureCause the reason of the failure
+   */
+  public abstract void onError(HttpServiceResponder responder, Throwable failureCause);
+}
```
anew: what happens if some other error happens? Like channel closed etc. Does this method get called? If not, is there a way to clean up resources used by the body consumer? For example, suppose we write to file. If an error occurs, we want to - at least - close the file handle, most likely also delete the incomplete file.

cdap-api/src/main/java/co/cask/cdap/api/service/http/NonTransactionalHttpContentConsumer.java
```diff
@@ -0,0 +1,51 @@
...
+
+import co.cask.cdap.api.Transactional;
+import co.cask.cdap.api.TxRunnable;
+import co.cask.cdap.api.dataset.Dataset;
+
+import java.nio.ByteBuffer;
+
+/**
+ * A {@link HttpContentConsumer} that supports non-transactional processing of HTTP request body.
+ * One can extends from this class and implements the {@link #onReceived(ByteBuffer, Transactional)} method
```
anew: extend ... implement

anew: This is confusing. So the NonTransactionalHttpContentConsumer needs to implement onReceived(chunk, Transactional)?

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: This method will still gets called

cdap-api/src/main/java/co/cask/cdap/api/service/http/NonTransactionalHttpContentConsumer.java
chtyim: Yes. The onReceive(ByeBuffer) is marked final, hence cannot be implemented.

chtyim: The idea is to have the method receives a Transactional instance, so that it can still optionally execute function inside a transaction if needed, but the call to the onReceive method is not inside a transaction.

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: It was relying on the netty-http library to call `BodyConsumer.handleError` when there is error so that the `BodyConsumerAdapter` (down below) will do the forwarding. However, it seems like there is bug on the netty-http library side that when body consumer is involve, nothing will get triggered if the client close connect before completing the streaming.

Will fix it separate from the netty-http library to guarantee right calls are being made.

chtyim: Related bug: https://issues.cask.co/browse/NETTY-9
Related fix: https://github.com/caskdata/netty-http/pull/24

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: fixed

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: fixed

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: fixed

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: fixed

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: fixed

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: fixed

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
```diff
@@ -0,0 +1,195 @@
...
+          txContext.start();
+          try {
+            ClassLoader oldClassLoader = ClassLoaders.setContextClassLoader(contextClassLoader);
+            try {
+              runnable.run(datasetContext);
+            } finally {
+              ClassLoaders.setContextClassLoader(oldClassLoader);
+            }
+            txContext.finish();
+          } catch (Throwable t) {
```
anew: I think this is wrong. The txContext can throw TransactionConflictException or other exceptions (for example if it can't talk to transaction service). In that case it is not a user handler exception. I would think that in case we need a catch around the runnable.run(), which wraps any exception into a TxFailureException. For finish, we should not catch exceptions, because it only throws TransactionFailureException (or its subclass TransactionConflictException). Not sure why we need WrappedException?

cdap-api/src/main/java/co/cask/cdap/api/Transactional.java
```diff
@@ -0,0 +1,36 @@
...
+public interface Transactional {
+
+  /**
+   * Execute a set of operations via a {@link TxRunnable} that are committed as a single transaction.
+   * The {@link TxRunnable} can gain access to {@link Dataset} through the {@link DatasetContext} provided
+   * to it.
+   *
+   * @param runnable the runnable to be executed in the transaction
+   */
+  void execute(TxRunnable runnable);
```
anew: why does this not throw exceptions? I would think it will throw TransactionFailure/ConflictExceptions.
Also, should this have an option to retry N times in case of conflict?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
```diff
@@ -0,0 +1,195 @@
...
+      }
+    };
+  }
+
+  /**
+   * A {@link RuntimeException} for wrapping any {@link Throwable}. It is needed because {@link Transactional}
+   * only throws {@link RuntimeException} so that in {@link #onError(Throwable, DelayedHttpServiceResponder)},
+   * we can unwrap and get the actual cause.
+   */
+  private static final class WrappedException extends RuntimeException {
```
anew: This is like ExecutionException? Why not keep the TransactionFailureException hierarchy?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
```diff
@@ -0,0 +1,195 @@
...
+    this.contextClassLoader = new CombineClassLoader(null, ImmutableList.of(delegate.getClass().getClassLoader(),
+                                                                            getClass().getClassLoader()));
+  }
+
+  @Override
+  public void chunk(final ChannelBuffer request, HttpResponder responder) {
+    try {
+      // Based on the type of the content consumer, call onReceive with transaction or not.
+      if (delegate instanceof NonTransactionalHttpContentConsumer) {
+        final ClassLoader oldClassLoader = ClassLoaders.setContextClassLoader(contextClassLoader);
```
anew: I am getting confused about all the class loader switching. If I understand correctly: oldClassLoader is the system class loader, whereas contextClassLoader is the program class loader? (perhaps name them that way?) So this gets called from system code, with system class loader. We switch to program class loader before calling onReceived. But that gets a Transactional, which switches back to system class loader before delegating to the actual Transactional. Which then switches back to the program class loader. Why is all this needed?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
```diff
@@ -0,0 +1,195 @@
...
+        try {
+          ((NonTransactionalHttpContentConsumer) delegate).onReceived(request.toByteBuffer(), new Transactional() {
+            @Override
+            public void execute(TxRunnable runnable) {
+              // Need to reset the context classloader to the system one before calling the
```
anew: why do we need to do that?

cdap-api/src/main/java/co/cask/cdap/api/Transactional.java
chtyim: If throws exception, it will break current API (in WorkerContext). We can have this interface defined, but not to have WorkerContext to extend from it, but it would be rather confusing too.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
```diff
@@ -0,0 +1,195 @@
...
+    final Throwable failureCause = (cause instanceof WrappedException) ? cause.getCause() : cause;
+    try {
+      transactional.execute(new TxRunnable() {
+        @Override
+        public void run(DatasetContext context) throws Exception {
+          delegate.onError(responder, failureCause);
+        }
+      });
+    } catch (Throwable t) {
+      responder.setTransactionFailureResponse(t);
```
anew: So: if onReceived fails, we wrap the exception into a WrappedException and call this method. This again calls Transactional.execute(), so if that throws an exception, it is a WrappedException, and we respond with that? Seems not right.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
chtyim: First, `contextClassLoader` is not the program class loader, but rather a combine class loader that prefer using program class loader to load class first (we do this for all program execution context, otherwise for program that uses HBase directly, it cannot find the right resource for the HBaseConfiguration). Anyway, the `contextClassLoader` is the class loader needs to be set as the context class loader whenever we are going to invoke user method.

The `oldClassLoader` is the existing context class loader.

We need the switching because we are going to call the `NonTransactionalHttpContentConsumer.onReceive` method, which is a user method.

The `Transactional

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/service/http/HttpHandlerGeneratorTest.java
```diff
@@ -120,6 +140,44 @@ public void echo(HttpServiceRequest request, HttpServiceResponder responder) {
...
+          channel.write(chunk);
+        }
+
+        @Override
+        public void onFinish(HttpServiceResponder responder) throws Exception {
+          channel.close();
+          responder.sendStatus(200);
+        }
+
+        @Override
```
anew: I guess this never happens in this test?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
chtyim: See my explanation above.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
chtyim: This implementation is adapted from `DefaultTransactionExecutor.executeOnce` method (https://github.com/caskdata/tephra/blob/develop/tephra-core/src/main/java/co/cask/tephra/DefaultTransactionExecutor.java#L131), which was used in the generated Service handler code as well prior to this change.

I just found a long standing tech debt related to this (https://issues.cask.co/browse/CDAP-617). Need to revisit the approach again and unify all code involved usage of `TransactionContext` directly.

For the wrapping, it is simply because the `Transactional.execute` don't have throws Exception in the signature (see above) so that there is a exception to unwrap it back when calling `onError`. Due to the refactoring of usage of `TransactionContext` is needed, the wrapping might goes away.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
chtyim: Yeah, should get the failure cause as in line 145.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/runtime/service/http/HttpHandlerGeneratorTest.java
chtyim: In this test, no. There is another test that covers failure upload case. This test is more for testing the byte code generation correctness.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
chtyim: Will revisit to see if we need the wrapping or not.

cdap-api/src/main/java/co/cask/cdap/api/service/http/NonTransactionalHttpContentConsumer.java
sreevatsanraman: `TxSuppliedHttpContentConsumer` since users have to supply the `Transactional` object in the method.

chtyim: The user doesn't supply the Transactional. It is provided by the CDAP system.

anew:
I like this a lot better than the previous one. LGTM except for a minor comment. 

One other thought: If the user's runnable throws an exception, then the error returned in the HTTP response will be a TxFailureException. That is either a conflict (and should we give the user a chance to deal with that by retrying?) or a wrapped exception (as the cause of the TxFailureExn). In that case, I feel that the HTTP response should have the cause instead of the TxFailureExn. Not sure how to best achieve that, but from user's stand point, that would make most sense

chtyim:
For the retrying on conflict, would like to handle it with https://issues.cask.co/browse/CDAP-617

When user runnable throws an exception, the response will be depended on when it happens. If it is from `onReceive` or `onFinish`, then `onError` will get call, and it's up to the user to interpret the exception and respond appropriately. However, if it from `onError`, then the system will respond with 500, which the user doesn't really have control on what the error message is.

chtyim:
Squashed commits. Waiting for the last build to complete.

http://builds.cask.co/browse/CDAP-DUT2913-4

anew:
That sounds good to me.

PR #4232, opened 2015-10-05
Fix/ui hydrator sidepanel default

elfenheart:
Default state is all 3 groups opened with 1/3 of the height. Click on any group, it will expand the whole height.

sidepanel


astewes:
Nice work! LGTM.

ajainarayanan:
:+1: :8ball: LGTM

PR #4233, opened 2015-10-05
optimizes images; adds sprites where appropriate

astewes:
*  Reduces size of images in styles/img (run through image optimizer)
*  Removes images that are no longer used
*  Moves welcome panel images to a CSS sprite


elfenheart:
:+1: LGTM

ajainarayanan:
@astewes Can't we add all the images to one sprite and use the same sprite everywhere? Just welcome page images alone looks too small. We could avoid making individual calls for fetching the images (even though they are small)

ajainarayanan:
:+1: LGTM

PR #4234, opened 2015-10-05
(CDAP-2470) Re-factored DBSource/DBSink to move common code to a DBMa…

bdmogal:
…nager class. Also added a precondition during deployment to check that the specified table exists in the database.

Moved DB classes under the co.cask.cdap.etl.common.db package.
Some other minor refactoring of tests.

Jira: [CDAP-2470](https://issues.cask.co/browse/CDAP-2470) [CDAP-3369](https://issues.cask.co/browse/CDAP-3369)
Build: http://builds.cask.co/browse/CDAP-DUT2915


bdmogal:
Pending long running test, will do that after an initial review.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
```diff
@@ -142,15 +147,15 @@ private static void createTestTables(Connection conn) throws SQLException {
                      "BLOB_COL BLOB(100), " +
                      "CLOB_COL CLOB(100)" +
                      ")");
-      stmt.execute("CREATE TABLE my_dest_table AS (" +
-                     "SELECT * FROM my_table) WITH DATA");
+      stmt.execute("CREATE TABLE \"my_dest_table\" AS (" +
```
gokulavasan: Why do we need the escape characters?

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
```diff
@@ -345,29 +354,43 @@ public void testUserNamePasswordCombinations() throws Exception {
...
       Properties.DB.IMPORT_QUERY, importQuery,
       Properties.DB.COUNT_QUERY, countQuery);
 
     Map<String, String> baseSinkProps = ImmutableMap.of(
       Properties.DB.CONNECTION_STRING, hsqlDBServer.getConnectionUrl(),
+      Properties.DB.TABLE_NAME, "my_table",
       Properties.DB.JDBC_PLUGIN_NAME, "hypersql",
-      Properties.DB.TABLE_NAME, "table",
-      Properties.DB.COLUMNS, "*");
+      Properties.DB.COLUMNS, "*",
```
gokulavasan: Hmm, do we document the option to use '*' ?

gokulavasan:
Few comments. Rest LGTM :+1:

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
bdmogal: HSQL by default (if you do not quote it), converts the table name to upper case. Quoting it retains the case. We need to think of whether we want to add the ``FieldCase``-kind of checks for table name too, but I'll leave that for later.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
bdmogal: Good call. Currently, we allow users to pass in ``*`` (even that is not documented in the field description though). However, maybe we can make the property nullable and have ``*`` as the default value. Will make the change.

bdmogal: So this was supported until 3.2.0, but I think its broken in 3.2.1. Reported https://issues.cask.co/browse/CDAP-3942. Will handle it separately, because it would have to be fixed in release/3.2.

bdmogal:
@gokulavasan please review the last commit.

cdap-app-templates/cdap-etl/cdap-etl-batch/src/test/java/co/cask/cdap/etl/batch/BatchETLDBTest.java
bdmogal: https://issues.cask.co/browse/CDAP-3953

gokulavasan:
Last commit looks good!

bdmogal:
Long running test currently running, will merge once it has run for a day or so.

bdmogal:
Build green. SDK has been running on ubuntu for over 30 hours. Merging.

PR #4235, opened 2015-10-05
moves active class to list item from anchor

astewes:
*  Moves actives class assignment via ng-class from list item to anchor
*  Adjusts styling of nav elements accordingly

hyd


elfenheart:
:+1: LGTM

PR #4236, opened 2015-10-05
[CDAP-3841] add color indicator to last finished run status

elfenheart:
for 3.3

screen shot 2015-10-05 at 3 51 06 pm

PR #4237, opened 2015-10-05
[CDAP-3841] Fix/ui hydrator list last finished for 3.2

elfenheart:
same as #4236 

screen shot 2015-10-05 at 3 51 06 pm


astewes:
LGTM

PR #4238, opened 2015-10-05
CDAP-3345, dataset TTL inconsistency fix.

gsps1:
use seconds in Dataset Specification, conversion to milliseconds only happens in HBaseTableAdmin implementation, all REST API's and  dataset specs uses seconds.

UpgradeTool has been updated to update DatasetSpecification Entries (from milli-seconds to seconds) in Dataset Instance MDS Table.

JIRA: https://issues.cask.co/browse/CDAP-3345

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Tables.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Tables.java
@@ -43,7 +43,7 @@ public final class Tables {
    * Adds {@link Table} data set to be created at application deploy if not exists.
    * @param configurer application configurer
    * @param datasetName data set name
-   * @param ttl time to live for data written into a table, in ms. {@link #NO_TTL} means unlimited
+   * @param ttl time to live for data written into a table, in seconds. {@link #NO_TTL} means unlimited
    */
   public static void createTable(ApplicationConfigurer configurer, String datasetName, int ttl) {
     createTable(configurer, datasetName, ConflictDetection.ROW, ttl, DatasetProperties.EMPTY);
@@ -64,7 +64,7 @@ public final class Tables {
    * @param configurer application configurer
    * @param datasetName data set name
    * @param level level on which to detect conflicts in changes made by different transactions
-   * @param ttl time to live for data written into a table, in ms. {@link #NO_TTL} means unlimited
+   * @param ttl time to live for data written into a table, in seconds. {@link #NO_TTL} means unlimited
    */
   public static void createTable(ApplicationConfigurer configurer, String datasetName,
                                  ConflictDetection level, int ttl) {
@@ -76,7 +76,7 @@ public final class Tables {
    * @param configurer application configurer
    * @param datasetName data set name
    * @param level level on which to detect conflicts in changes made by different transactions
-   * @param ttl time to live for data written into a table, in ms. {@link #NO_TTL} means unlimited
+   * @param ttl time to live for data written into a table, in seconds. {@link #NO_TTL} means unlimited
    * @param props any additional data set properties
    */
   public static void createTable(ApplicationConfigurer configurer, String datasetName,
@@ -89,7 +89,7 @@ public final class Tables {
   /**
    * Creates properties for {@link Table} or {@link Table} data set instance.
    * @param level level on which to detect conflicts in changes made by different transactions
-   * @param ttl time to live for data written into a table, in ms. {@link #NO_TTL} means unlimited
+   * @param ttl time to live for data written into a table, in seconds. {@link #NO_TTL} means unlimited
    * @return {@link DatasetProperties} for the data set
    */
   public static DatasetProperties tableProperties(ConflictDetection level, int ttl, DatasetProperties props) {
--- a/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTableTest.java
+++ b/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTableTest.java
@@ -134,7 +134,7 @@ public class HBaseTableTest extends BufferingTableTest<BufferingTable> {
   public void testTTL() throws Exception {
     // for the purpose of this test it is fine not to configure ttl when creating table: we want to see if it
     // applies on reading
-    int ttl = 1000;
+    int ttl = 1;
     String ttlTable = "ttl";
     String noTtlTable = "nottl";
     DatasetProperties props = DatasetProperties.builder().add(Table.PROPERTY_TTL, String.valueOf(ttl)).build();
@@ -150,7 +150,7 @@ public class HBaseTableTest extends BufferingTableTest<BufferingTable> {
     table.put(b("row1"), b("col1"), b("val1"));
     table.commitTx();
 
-    TimeUnit.SECONDS.sleep(2);
+    TimeUnit.MILLISECONDS.sleep(1010);
 
     tx = txSystemClient.startShort();
     table.startTx(tx);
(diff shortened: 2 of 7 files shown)


cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Tables.java
```diff
@@ -89,7 +89,7 @@ public static void createTable(ApplicationConfigurer configurer, String datasetN
   /**
    * Creates properties for {@link Table} or {@link Table} data set instance.
    * @param level level on which to detect conflicts in changes made by different transactions
-   * @param ttl time to live for data written into a table, in ms. {@link #NO_TTL} means unlimited
+   * @param ttl time to live for data written into a table, in seconds. {@link #NO_TTL} means unlimited
```
chtyim: So this actually is a non-backward compatible API change. Any old code being deploy again on 3.3 will experience 1000 times longer TTL.

awholegunch:
I have revised the Javadocs. Please check and confirm the change.

cdap-api/src/main/java/co/cask/cdap/api/dataset/table/Tables.java
gsps1: this is more like a documentation bug, it should be seconds instead of ms. also i am not sure how this Class is used, i don't see any usages of this.. i see this is part of cdap-api, but not sure how it can use used..

anew: and actually I am surprised that this is an int. With 2B as the max int and ~86K seconds in a day, that is roughly 60 years. Probably enough. But when this was in milliseconds, that is less than a month.

cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTableTest.java
```diff
@@ -136,7 +136,7 @@ protected boolean isReadlessIncrementSupported() {
   public void testTTL() throws Exception {
     // for the purpose of this test it is fine not to configure ttl when creating table: we want to see if it
     // applies on reading
-    int ttl = 1000;
+    int ttl = 1;
```
anew: can we reduce the sleep in this test to 1.01 seconds?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTableAdmin.java
```diff
@@ -83,6 +84,8 @@ public void create() throws IOException {
     String ttlProp = spec.getProperties().get(Table.PROPERTY_TTL);
     if (ttlProp != null) {
       long ttl = Long.parseLong(ttlProp);
+      // convert ttl from seconds to milli-seconds
```
anew: this can be inside the if { }

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/table/hbase/HBaseTableAdmin.java
```diff
@@ -132,14 +135,19 @@ protected boolean upgradeTable(HTableDescriptor tableDescriptor) {
       tableUtil.setBloomFilter(columnDescriptor, HBaseTableUtil.BloomType.ROW);
       needUpgrade = true;
     }
+    // ttl not null, convert to millis
```
anew: this is only needed in the else { }

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetSpecificationUpgrader.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  @VisibleForTesting
+  DatasetSpecification updateTTLInSpecification(DatasetSpecification specification) {
+    Map<String, String> properties = updatedProperties(specification.getProperties());
+    List<DatasetSpecification> updatedSpecs = new ArrayList<>();
+    for (DatasetSpecification datasetSpecification : specification.getSpecifications().values()) {
+      updatedSpecs.add(updateTTLInSpecification(datasetSpecification));
+    }
+    return DatasetSpecification.builder(specification.getName(),
```
anew: I believe this is wrong. Because spec.getName will be enclosingName.name. You need to use the key from the specs map that has the unqualified name. Better double-check.

cdap-master/src/test/java/co/cask/cdap/data/tools/DatasetSpecificationUpgradeTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    Assert.assertEquals(expected, datasetSpecificationUpgrader.updateTTLInSpecification(specification));
+
+    DatasetSpecification nestedSpec = DatasetSpecification.builder("dataset2", "Table").
+      properties(ImmutableMap.of(Table.PROPERTY_TTL, "7200000")).datasets(specification).build();
+
+    // TTL should be converted to seconds for both parent and nested specs
+    DatasetSpecification expected2 = DatasetSpecification.builder("dataset2", "Table").
+      properties(ImmutableMap.of(Table.PROPERTY_TTL, "7200")).
+      datasets(DatasetSpecification.builder("dataset2.dataset1", "Table").
```
anew: wrong. See my comment above and also https://issues.cask.co/browse/CDAP-1721

anew:
Overall good, but please see my comments as the rewriting of the spec during upgrade is incorrect.

cdap-master/src/test/java/co/cask/cdap/data/tools/DatasetSpecificationUpgradeTest.java
gsps1: i see, thanks for the clarification.

cdap-master/src/main/java/co/cask/cdap/data/tools/DatasetSpecificationUpgrader.java
```diff
@@ -0,0 +1,137 @@
...
+   * Since the spec (as of CDAP version 3.3) is in seconds, the instance MDS entries must be updated.
+   * This is to be called only if the current CDAP version is < 3.3.
+   * </p>
+   * @throws Exception
+   */
+  public void upgrade() throws Exception {
+    TableId datasetSpecId = TableId.from(Id.Namespace.SYSTEM.getId(), DatasetMetaTableUtil.INSTANCE_TABLE_NAME);
+    HBaseAdmin hBaseAdmin = new HBaseAdmin(conf);
+    if (!tableUtil.tableExists(hBaseAdmin, datasetSpecId)) {
+      LOG.info("Dataset instance table does not exist: {}. Should not happen", datasetSpecId);
```
anwar6953: log this as a WARN or ERROR, instead of simply as an INFO, since it shouldn't happen.

sagarkapare: Or it should throw an exception right?

gsps1:
@anew, have addressed the comments, can you please review again. Thanks!

anew:
Thanks for addressing all the comments, the code looks good now. 

One concern I have (and that I did not realize in the first round of review) is that this upgrade is not idempotent. All our upgrades must be idempotent, because users may run the upgrade tool repeatedly (if it fails the first time). Not sure how to deal with that. Perhaps as part of the upgrade, you can also add a property that specifies the time unit of TTL. If the tool is run again, you can skip any table that already has the new property - because it is a new property, such a table must have been upgraded already.

Or if you have other ideas how to make this idempotent, feel free.

gsps1:
thanks for the comment @anew, I now add a property to the dataset spec, so once the row has been upgraded, we will skip the next time if that field is present. The upgrade procedure just upgrades the dataset spec in dataset.instances table. 

As for other tables, since we are using seconds as resolution and calling the upgrade sets the TTL as we expect to be (in seconds), and that operation is idempotent. 

please review again, thanks!

gsps1:
@anew have fixed the issues, please review. thanks

gsps1:
http://builds.cask.co/browse/CDAP-DUT3042 - build Green

anew:
LGTM

gsps1:
rebased commits, merging.

PR #4239, opened 2015-10-06
fixes ns dropdown on smaller devices

astewes:
*  Fixes positioning issue affecting namespace dropdown (breadcrumbs directive) on device widths ranging from 992px - 1200px

dd

### DIFF
--- a/cdap-ui/app/directives/breadcrumbs/breadcrumbs.less
+++ b/cdap-ui/app/directives/breadcrumbs/breadcrumbs.less
@@ -135,52 +135,6 @@ body.theme-cdap {
         }
       }
 
-      // ul.dropdown-menu {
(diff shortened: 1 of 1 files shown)


cdap-ui/app/directives/breadcrumbs/breadcrumbs.less
```diff
@@ -18,195 +18,245 @@
...
-            display: inline-block;
-            position: relative;
+          &:last-child {
+            margin-right: 7px;
           }
         }
-        &:nth-last-child(2) {
+      }
+
+      // ul.dropdown-menu {
```
anwar6953: can we entirely remove this? Or why leave it commented?

elfenheart:
:+1: LGTM

PR #4240, opened 2015-10-06
[CDAP-3838] adding last finished info on hydrator detail status tab

elfenheart:
screen shot 2015-10-06 at 11 14 00 am

screen shot 2015-10-06 at 11 14 12 am

### DIFF
--- a/cdap-ui/app/features/hydrator/controllers/tabs/status-ctrl.js
+++ b/cdap-ui/app/features/hydrator/controllers/tabs/status-ctrl.js
@@ -24,15 +24,22 @@ angular.module(PKG.name + '.feature.hydrator')
 
     $scope.programType = HydratorDetail.programType;
 
-    HydratorDetail.api.runs(params)
+    HydratorDetail.api.pollRuns(params)
       .$promise
       .then(function (runs) {
         $scope.stats.numRuns = runs.length;
         $scope.stats.lastRunTime = runs.length > 0 && runs[0].end ? runs[0].end - runs[0].start : 'N/A';
+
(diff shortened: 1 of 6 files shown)


cdap-ui/app/features/hydrator/controllers/tabs/status-ctrl.js
```diff
@@ -24,15 +24,22 @@ angular.module(PKG.name + '.feature.hydrator')
...
 
-    HydratorDetail.api.runs(params)
+    HydratorDetail.api.pollRuns(params)
       .$promise
       .then(function (runs) {
         $scope.stats.numRuns = runs.length;
         $scope.stats.lastRunTime = runs.length > 0 && runs[0].end ? runs[0].end - runs[0].start : 'N/A';
+
+        for (var i = 0; i < runs.length; i++) {
+          if (runs[i].status !== 'RUNNING') {
```
anwar6953: A potential status is 'STARTING', which is not finished, but this code would assume it is finished, right?

ajainarayanan:
@anwar6953 I think we want to identify the last "Killed" or "Completed" or "Stopped" run. So we are navigating through the list of runs to find the latest run that is having either of these states (or not one of 'Starting' or 'Stopping' or 'Running').

ajainarayanan:
Same as the other PR (for 3.2) If for one comment :+1: LGTM. (if @anwar6953 is convinced :smile: )

anwar6953:
I see. I thought there was only one state it would check against - 'STOPPED'.
Since there's also KILLED and COMPLETED, I guess its significantly better.

LGTM.

PR #4241, opened 2015-10-06
[CDAP-3838] adding last finished info on hydrator detail status tab for 3.2

elfenheart:
same as #4240 

screen shot 2015-10-06 at 11 14 00 am

screen shot 2015-10-06 at 11 14 12 am

### DIFF
--- a/cdap-ui/app/features/adapters/bottompanel.less
+++ b/cdap-ui/app/features/adapters/bottompanel.less
@@ -350,10 +350,6 @@ body.theme-cdap.state-adapters {
         font-weight: 500;
         font-size: 24px;
       }
-
-      h4.last-finished-time {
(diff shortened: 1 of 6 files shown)


cdap-ui/app/features/adapters/bottompanel.less
```diff
@@ -350,6 +350,10 @@ body.theme-cdap.state-adapters {
         font-weight: 500;
         font-size: 24px;
       }
+
+      h4.last-finished-time {
```
ajainarayanan: Any specific reason why the finished time alone has different font-weight? Can we use ```<small></small>``` inside the ```h-tag```

ajainarayanan:
If for a minor comment :+1: LGTM

PR #4242, opened 2015-10-06
CDAP-3696 Disabling the flaky test till its fixed.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-3696

This test case tests that the failure of one MapReduce program in the Workflow fork should cause the MapReduce program running in other branch to get marked as killed by interrupting it. However the test case is flaky and the outcome depends on the timing of when the interrupt is generated. Disabling the test till we have the robust way of verification.

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
@@ -59,6 +59,7 @@ import com.google.gson.reflect.TypeToken;
 import org.apache.http.HttpResponse;
 import org.apache.http.util.EntityUtils;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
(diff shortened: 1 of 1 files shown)


gsps1:
lgtm, please add a comment in the JIRA about why you are disabling it now and how you are planning to make it robust.

cdap-app-fabric/src/test/java/co/cask/cdap/internal/app/services/http/handlers/WorkflowHttpHandlerTest.java
```diff
@@ -1293,6 +1294,7 @@ public void testWorkflowTokenPut() throws Exception {
     Assert.assertEquals(1, sparkProgramRuns.size());
   }
 
+  @Ignore
```
bdmogal: is there a jira to fix the test?

sagarkapare: Yes. CDAP-3696

sagarkapare:
Added comment in the JIRA(CDAP-3696). Merging it since it is only disabling the test case change.

PR #4246, opened 2015-10-06
[CDAP-3849] add red highlight for empty required fields

elfenheart:
Same as #4245

screen shot 2015-10-06 at 3 03 52 pm


ajainarayanan:
Talked to @nitinmotgi about this. The highlighting is for validating individual fields of the plugin. We could have highlighting required fields on validate but not when the user is opening up node config of a plugin for the first time. We need to decide on how and what validations we are going to perform and do it as part of a major release (right now this is not for 3.2.1). Hence closing this PR to be reopened later.

PR #4247, opened 2015-10-07
CDAP-3677: implement storeJobsandTriggers

rohitsinha54:
Issue: https://issues.cask.co/browse/CDAP-3677
Build: http://builds.cask.co/browse/CDAP-DUT2947-1

- implementing overriding method storeJobsAndTriggers in DatasetBasedTimeScheduleStore which will support storing multiple jobs and triggers as one operation. This does not had any implementation before.
- This will help to solve the issue in which while deploying multiple triggers of a job the failure of one trigger will fail the deployment of all as the sanity check is done before hand. 
- The bulk deployment will be done during the refactoring of the Scheduler: https://issues.cask.co/browse/CDAP-3598
- It will also allow us to add all the triggers with a job in one go rather than iterating over all schedules and adding them one by one.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
@@ -95,7 +95,12 @@ public class DatasetBasedTimeScheduleStore extends RAMJobStore {
   public void storeJobsAndTriggers(Map<JobDetail, Set<? extends Trigger>> triggersAndJobs,
                                    boolean replace) throws JobPersistenceException {
     super.storeJobsAndTriggers(triggersAndJobs, replace);
-    // TODO (sree): Add persisting job and triggers
+    for (Map.Entry<JobDetail, Set<? extends Trigger>> e: triggersAndJobs.entrySet()) {
+      storeJob(e.getKey(), replace);
+      for (Trigger trigger: e.getValue()) {
+        storeTrigger((OperableTrigger) trigger, replace);
+      }
+    }
   }
 
   @Override
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
@@ -303,7 +303,7 @@ public class DatasetBasedTimeScheduleStore extends RAMJobStore {
                 LOG.debug("Schedule: trigger with key {} added", trigger.trigger.getKey());
               } else {
                 LOG.debug("Schedule: trigger with key {} and state {} skipped", trigger.trigger.getKey(),
-                          trigger.state.toString());
+                          trigger.state);
               }
             }
           } else {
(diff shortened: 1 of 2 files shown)


sagarkapare:
@rohitsinha54 Can you please add explanation as to why this change is required?

rohitsinha54:
@sagarkapare : Added in description.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
```diff
@@ -95,7 +95,12 @@ public void storeTrigger(OperableTrigger newTrigger, boolean replaceExisting) th
   public void storeJobsAndTriggers(Map<JobDetail, Set<? extends Trigger>> triggersAndJobs,
                                    boolean replace) throws JobPersistenceException {
     super.storeJobsAndTriggers(triggersAndJobs, replace);
```
sagarkapare: do we need this, since `storeJob` and `storeTrigger` already calling super?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
```diff
@@ -95,7 +95,12 @@ public void storeTrigger(OperableTrigger newTrigger, boolean replaceExisting) th
...
-    // TODO (sree): Add persisting job and triggers
+    for (Map.Entry<JobDetail, Set<? extends Trigger>> e: triggersAndJobs.entrySet()) {
+      storeJob(e.getKey(), replace);
+      for (Trigger trigger: e.getValue()) {
+        storeTrigger((OperableTrigger) trigger, replace);
+      }
+    }
```
sagarkapare: Is it better to have `storeJobTrigger(JobDetail newJob, OperableTrigger newTrigger, boolean replace)` method? This can be used then by the above code as well as by following `storeJobAndTrigger(JobDetail newJob, OperableTrigger newTrigger)`

sagarkapare:
Bulk operation will still use the `persistJobAndTrigger`, which only persists the individual job and trigger at a time. Are we planning to add version of this method to persist multiple jobs and triggers in one transaction?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
rohitsinha54: Yes. Its a different method call on super not at all related to storeJob and storeTrigger.

rohitsinha54:
Yes. The idea is to able to store multiple triggers for a job in one transaction so the failure of one can fail the complete operation on a higher level too. This storeJobsAndTriggers will be used at that point with one job and multiple triggers in the set.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
rohitsinha54: That will cause the storeJob to be called again and again for every storeTrigger which is unnecessary work as the job is same for all its associated trigger and since storeJobAndTrigger just makes call to other methods which we are doing directly here I don't think it is a code duplication issue.

rohitsinha54:
New Build: http://builds.cask.co/browse/CDAP-DUT2947-2

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/store/DatasetBasedTimeScheduleStore.java
```diff
@@ -298,7 +303,7 @@ public void apply() throws Exception {
                 LOG.debug("Schedule: trigger with key {} added", trigger.trigger.getKey());
               } else {
                 LOG.debug("Schedule: trigger with key {} and state {} skipped", trigger.trigger.getKey(),
-                                                                                trigger.state.toString());
+                          trigger.state.toString());
```
chtyim: No need to call `toString()`.

sagarkapare:
LGTM :+1:

PR #4248, opened 2015-10-07
CDAP-3926 Make the regex match until end of line, and restrict versio…

anwar6953:
CDAP-3926 Make the regex match until end of line, and restrict version suffix from having numerical values. This allows the artifact name to have numbers in it, without ambiguity about what components are the version and what is the name.

https://issues.cask.co/browse/CDAP-3926
http://builds.cask.co/browse/CDAP-RBT499-1

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactVersion.java
+++ b/cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactVersion.java
@@ -34,7 +34,7 @@ import javax.annotation.Nullable;
 @Beta
 public final class ArtifactVersion implements Comparable<ArtifactVersion> {
 
-  private static final Pattern PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:[.\\-](.*))?");
+  private static final Pattern PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:[.\\-](.*))?$");
 
   private final String version;
   private final Integer major;
@@ -68,7 +68,7 @@ public final class ArtifactVersion implements Comparable<ArtifactVersion> {
     String suffix = null;
     if (str != null) {
       Matcher matcher = PATTERN.matcher(str);
-      boolean matches = matchSuffix ? (matcher.find() && matcher.hitEnd()) : matcher.matches();
+      boolean matches = matchSuffix ? (matcher.find()) : matcher.matches();
 
       if (matches) {
         tmpVersion = matcher.group(0);
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/api/artifact/ArtifactVersionTest.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/api/artifact/ArtifactVersionTest.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.cdap.api.artifact;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Tests the parsing of artifact version.
+ */
+public class ArtifactVersionTest {
+
+  @Test
+  public void testParseVersion() {
+    // Versions without suffix
+    ArtifactVersion version = new ArtifactVersion("1");
+    Assert.assertEquals(Integer.valueOf(1), version.getMajor());
+
+    version = new ArtifactVersion("1.2");
+    Assert.assertEquals(Integer.valueOf(1), version.getMajor());
+    Assert.assertEquals(Integer.valueOf(2), version.getMinor());
+
+    version = new ArtifactVersion("1.2.3");
+    Assert.assertEquals(Integer.valueOf(1), version.getMajor());
+    Assert.assertEquals(Integer.valueOf(2), version.getMinor());
+    Assert.assertEquals(Integer.valueOf(3), version.getFix());
+
+    // Versions with suffix
+    version = new ArtifactVersion("1-suffix");
+    Assert.assertEquals(Integer.valueOf(1), version.getMajor());
+    Assert.assertEquals("suffix", version.getSuffix());
+
+    version = new ArtifactVersion("1.2-suffix");
+    Assert.assertEquals(Integer.valueOf(1), version.getMajor());
+    Assert.assertEquals(Integer.valueOf(2), version.getMinor());
+    Assert.assertEquals("suffix", version.getSuffix());
+
+    version = new ArtifactVersion("1.2.3-suffix");
+    Assert.assertEquals(Integer.valueOf(1), version.getMajor());
+    Assert.assertEquals(Integer.valueOf(2), version.getMinor());
+    Assert.assertEquals(Integer.valueOf(3), version.getFix());
+    Assert.assertEquals("suffix", version.getSuffix());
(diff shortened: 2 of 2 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/api/artifact/ArtifactVersionTest.java
```diff
@@ -0,0 +1,82 @@
...
+    assertVersionEquals("3.2.0-suffix", 3, 2, 0, "suffix", artifactVersion);
+
+    versionString = "1.0-SNAPSHOT";
+    artifactVersion = new ArtifactVersion(versionString, false);
+    assertVersionEquals("1.0-SNAPSHOT", 1, 0, null, "SNAPSHOT", artifactVersion);
+  }
+
+  @Test
+  public void testInvalidVersions() {
+    // invalid simply because there are no numbers in the string
```
gokulavasan: Can we add an test to assert invalidity when we have numbers in suffix?

anwar6953: Line 65 is an example of something that is an invalid version, but would be valid if numbers were allowed in the suffix.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/api/artifact/ArtifactVersionTest.java
```diff
@@ -0,0 +1,82 @@
...
+public class ArtifactVersionTest {
+
+  @Test
+  public void testValidVersions() {
+    // since a version's suffix can not have numbers in it, simply the number 1 is the version
+    String versionString = "cdap-etl-batch-3.5-SNAPSHOT1";
+    ArtifactVersion artifactVersion = new ArtifactVersion(versionString, true);
+    assertVersionEquals("1", 1, null, null, null, artifactVersion);
+
+    versionString = "s3_analytics-1.0-SNAPSHOT";
```
gokulavasan: May we add a test with no suffix? For exmaple, cdap-etl-batch-3.2.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/api/artifact/ArtifactVersionTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Tests the parsing of artifact version.
+ */
+public class ArtifactVersionTest {
+
+  @Test
+  public void testValidVersions() {
+    // since a version's suffix can not have numbers in it, simply the number 1 is the version
+    String versionString = "cdap-etl-batch-3.5-SNAPSHOT1";
```
gokulavasan: Shouldn't 1 be the version only if we have a hyphen in front of '1' ?

anwar6953: good point. this should be an invalid version. We need a as the prefix in the regex matcher, then?

cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactVersion.java
```diff
@@ -30,11 +30,12 @@
  * parse version string in format of [major].[minor].[fix](-|.)[suffix]
  * }
  * </pre>
+ * Note that the suffix can not have numerical values in it.
  */
 @Beta
 public final class ArtifactVersion implements Comparable<ArtifactVersion> {
```
chtyim: How is this class different from "https://github.com/caskdata/cdap/blob/develop/cdap-app-fabric/src/main/java/co/cask/cdap/templates/PluginVersion.java" ?

anwar6953: Seems like its just a renaming, with PluginVersion left in place (but deprecated) for backwards compat.

chtyim: Then why not all tests from PluginVersionTest are copied to ArtifactVersionTest?

anwar6953: That's a good question - I'm unsure why that wasn't done.
I'll copy them over.

cdap-api/src/main/java/co/cask/cdap/api/artifact/ArtifactVersion.java
```diff
@@ -30,11 +30,12 @@
...
 
-  private static final Pattern PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:[.\\-](.*))?");
+  private static final Pattern PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:[.\\-]([^\\d]*))?$");
```
chtyim: Why need to exclude number from suffix? I think the right logic should be using the old pattern, but then keep calling `matcher.find()` while it returns true until `matcher.hitEnd()` return true. Something like this:

```java
boolean matches = false;
if (matchSuffix) {
  while (!matches && matcher.find()) {
    matches = matcher.hitEnd();
  }
} else {
  matches = matcher.matches();
}
```

chtyim: I verified it locally, then should fix the issue.

chtyim: Verified by:

```java
Assert.assertEquals(new PluginVersion("1.0-SNAPSHOT"), new PluginVersion("s3_analytics-1.0-SNAPSHOT", true));
```

anwar6953: My understanding is that artifact name can also have numbers and dashes.
Because of this, if we allow suffix to have numbers, then isn't it ambiguous in this case:
"ali-1-1-foo-1"
Is the artifact name "ali", version is "1-1-foo-1"
or is the artifact name "ali-1" and the version is "1-foo-1"
or is the artifact name "ali-1-1-foo" and the version is "1"?

If we restrict suffix from having numbers, then its clear that the name is "ali-1-1-foo", and the version is "1"

chtyim: First, you cannot restrict suffix to only have number, since snapshot can have timestamp in it.

For the example you gave, version is "1". because version string is having "." separation.

anwar6953: The version could also be "1-foo-1", for instance.
Major version is "1", and suffix is "foo-1".

chtyim: We can improve it a bit by requiring a "-" before the version (for suffix parsing, `matchSuffix == true` case).

chtyim: Right. Since regex is by default greedy parsing, it will take the whole string. "1-foo-1" as the version string. Major = 1, suffix = foo-1.

As long as it's deterministic, it's good enough.

chtyim: We should be able to document the naming convention of artifact file name to make it clear what the parsing logic is. At the end, the user has control on what the file name is, and should be relatively easy to make a name so that CDAP can grab a version string correctly. Also with the current logic, many common artifact name can be parsed correctly by default.

anwar6953: Another issue:
"s3_analytics-1.0-SNAPSHOT"
If we just change the first '_' to a '-', everything after the first letter ('s') is taken as the version string, because the regex matches everything starting with the '3'.

anwar6953: Even in this case:
"cdap-etl-3.5-batch-1.0-SNAPSHOT"
"3.5-batch-1.0-SNAPSHOT" is being matched as the full version string.
3 is major version, 5 is minor version, and remainder is suffix.

You brought up a good point that suffix needs to allow numbers (because it might have timestamp in it), though.

anwar6953:
Update the pattern to have '$' at the end, to match the last occurrence.

chtyim:
LGTM.

PR #4250, opened 2015-10-07
(CDAP-3627) Merge the HttpContentConsumer and NonTransactionHttpContentConsumer

chtyim:
- In most use cases, transaction is not needed in the onReceive method


cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
```diff
@@ -71,20 +72,15 @@
    * This method will get invoked when a new chunk of the request body is available to be consumed.
    * It is guaranteed that no concurrent calls to this method will be made.
    * <p>
-   * Also note that each invocation
-   * to this method is executed inside a single transaction, hence access to transactional {@link Dataset Datasets}
-   * is possible through the {@link HttpServiceContext} provided to the
-   * {@link HttpServiceHandler#initialize(HttpServiceContext)} method.
+   * Access to transactional {@link Dataset} should be done through the
```
anew: Datasets

anew:
One nitty comment, otherwise LGTM

cdap-api/src/main/java/co/cask/cdap/api/service/http/HttpContentConsumer.java
chtyim: fixed

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2913-5

Squashing and merging.

PR #4252, opened 2015-10-07
Fixes the condition to show output schema for sinks in hydrator

ajainarayanan:


cdap-ui/app/features/hydrator/templates/create/popovers/plugin-edit-config-form.html
```diff
@@ -61,8 +61,7 @@
             <li class="text-danger" ng-repeat="error in errors">{{ error }}</li>
           </ul>
         </div>
-
-        <h4 class="sink-schema" ng-if="pluginCopy.properties.schema || pluginCopy.implicitSchema || pluginCopy._backendProperties['schema'].required">
+        <h4 class="sink-schema" ng-if="pluginCopy.properties.schema !== undefined || pluginCopy.implicitSchema || pluginCopy._backendProperties['schema'].required">
```
elfenheart: I feel like pluginCopy._backendProperties['schema'].required is wrong. The schema property might be there but not required, and we still have to surface the widget, right?

elfenheart:
:+1: LGTM

PR #4255, opened 2015-10-07
(CDAP-3932) Added app.program.extra.classpath option

chtyim:
- The option carries extra classpath to be used for all programs
- The paths specified in the classpath are included when
  in the program class loader.


cdap-common/src/main/java/co/cask/cdap/common/lang/DirectoryClassLoader.java
```diff
@@ -92,6 +97,19 @@ public Manifest getManifest() {
       for (String libDir : libDirs) {
         addJarURLs(new File(dir, libDir), urls);
       }
+
+      if (extraClassPath != null) {
+        for (String classPath : Splitter.on(File.pathSeparatorChar).omitEmptyStrings().split(extraClassPath)) {
```
jwang47: Maybe `classPathEntry` would be a better name here.

jwang47:
One minor comment, otherwise LGTM.

cdap-common/src/main/java/co/cask/cdap/common/lang/DirectoryClassLoader.java
chtyim: Renamed to `path`.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-RBT501-1

PR #4257, opened 2015-10-08
(CDAP-3342) Updated hadoop native binaries for Windows. These do not …

bdmogal:
…have

dependencies on debug binaries. They were built from hadoop source
(branch/2.3.0) for 64-bit Windows machines.


bdmogal:
@awholegunch tested these on a Windows 64 bit environment where CDAP would previously not start up.

awholegunch:
LGTM. Tested under Windows 8 under Parallels on the Mac. Deployed CountRandom and ran successfully.

PR #4258, opened 2015-10-08
CDAP-3950 namespace is not needed in request body of multi status

albertshau:


awholegunch:
LGTM. I thought for sure I had already fixed this...

PR #4261, opened 2015-10-08
CDAP-3902: Remove remote call for registering dataset usage in DatasetInstanceService

jwang47:
https://issues.cask.co/browse/CDAP-3902
http://builds.cask.co/browse/CDAP-DUT2942

- [x] Integration tests


cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/SystemUsageRegistry.java
```diff
@@ -0,0 +1,37 @@
...
+import co.cask.cdap.data2.dataset2.DatasetFramework;
+import co.cask.tephra.TransactionExecutorFactory;
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
+
+/**
+ * {@link UsageRegistry} implementation that uses Constants.Bindings.SYSTEM_DATASET_FRAMEWORK named implementation of
+ * {@link DatasetFramework}.
+ */
+public class SystemUsageRegistry extends UsageRegistry {
```
poornachandra: This class can be replaced by a provider binding, right?

poornachandra:
@jwang47 I did a pass, and changes mostly look good. Since this involves a lot of refactoring, I'll do another pass later.

jwang47:
Got integration tests passing.

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/DataSetServiceModules.java
```diff
@@ -69,6 +71,10 @@ protected void configure() {
         install(new FactoryModuleBuilder()
                   .implement(DatasetDefinitionRegistry.class, DefaultDatasetDefinitionRegistry.class)
                   .build(DatasetDefinitionRegistryFactory.class));
+
+        bind(UsageRegistry.class).toProvider(SystemUsageRegistryProvider.class);
```
poornachandra: I think you'll still have to bind this to a specific key. Otherwise, all instances of UsageRegistry where this module is included will turn into SystemUsageRegistry.

jwang47:
Since we don't need all the functionality of LocalDatasetFramework, we can simply create a class that creates a dataset given the datasetMeta and datasetTypeMeta and use this in both RemoteDatasetFramework.getDataset() and UsageRegistry.

PR #4262, opened 2015-10-08
CDAP-3657: Improve Id serialization

jwang47:
- Create new set of id classes (ElementId) with flat representation
- Use ElementId for Id.toString

http://builds.cask.co/browse/CDAP-DUT2949

- [x] Integration tests


jwang47:
Got integration tests passing.

cdap-proto/src/main/java/co/cask/cdap/proto/codec/IdTypeAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonSerializationContext;
+
+import java.lang.reflect.Type;
+
+/**
+ * Class for serialize/deserialize Id object to/from json through {@link com.google.gson.Gson Gson}.
+ * Uses {@link ElementId} as the actual object to be serialized and deserialized.
+ */
+public final class IdTypeAdapter extends AbstractSpecificationCodec<Id> {
```
chtyim: It's not really a specification. Why need to extend from `AbstractSpecificationCodec`?

jwang47: Oops, this shouldn't extend that. Will fix.

cdap-proto/src/main/java/co/cask/cdap/proto/element/ElementType.java
```diff
@@ -0,0 +1,132 @@
...
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableMap;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.util.Map;
+
+/**
+ *
```
chtyim: Some javadoc?

cdap-proto/src/main/java/co/cask/cdap/proto/element/ElementType.java
```diff
@@ -0,0 +1,132 @@
...
+  static {
+    ImmutableMap.Builder<Class<? extends ElementId>, ElementType> builder = ImmutableMap.builder();
+    for (ElementType type : ElementType.values()) {
+      builder.put(type.getIdClass(), type);
+    }
+    byIdClass = builder.build();
+  }
+
+  private static final Map<Class<? extends Id>, ElementType> byOldIdClass;
+  static {
```
chtyim: You can combine the two static blocks into one.

cdap-proto/src/main/java/co/cask/cdap/proto/id/FlowletId.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    if (!super.equals(o)) {
```
anew: if you call this, you don't need the preceding lines

cdap-proto/src/main/java/co/cask/cdap/proto/id/ElementId.java
```diff
@@ -0,0 +1,104 @@
...
+import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.element.ElementType;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+
+import java.util.Iterator;
+import java.util.Objects;
+
+/**
+ * Uniquely identifies a particular instance of an element.
```
anew: Can you add some instructions on how to add a new element type? (it requires that ElementType enum is updated).

cdap-proto/src/main/java/co/cask/cdap/proto/id/ElementId.java
```diff
@@ -0,0 +1,104 @@
...
+    return elementType;
+  }
+
+  public static <T extends Id> T fromStringOld(String string, Class<T> oldIdClass) {
+    ElementType type = ElementType.valueOfOldIdClass(oldIdClass);
+    ElementId id = fromString(string, type.getIdClass());
+    return (T) id.toId();
+  }
+
+  public static <T extends ElementId> T fromString(String string, Class<T> idClass) {
```
anew: probably every subclass should implement fromString(String)

anew:
Changes look good to me. See the few comments I have. 
Also I would prefer ti use period instead of colon as the toString separator.

PR #4263, opened 2015-10-08
make status non clickable on hydrator top panel

elfenheart:


astewes:
Nit-picky, but I believe there's a mixin available for box-shadow (ie, .box-shadow(none);) to make it work with more browsers.

astewes:
:+1: LGTM

PR #4265, opened 2015-10-08
[CDAP-3883] Fix ResourceCoordinatorTest testAssignment test is flaky

henry-cask:
From the stack of error, seemed like the call timed out on polling the result.
Proposed fix is to add more tries using StopWatch with quick sleep per polling timeout.

### DIFF
--- a/cdap-common/src/test/java/co/cask/cdap/common/zookeeper/coordination/ResourceCoordinatorTest.java
+++ b/cdap-common/src/test/java/co/cask/cdap/common/zookeeper/coordination/ResourceCoordinatorTest.java
@@ -103,17 +103,17 @@ public class ResourceCoordinatorTest {
           Cancellable cancelSubscribe1 = subscribe(client, discoverable1, assignmentQueue, finishSemaphore);
 
           // Assert that it received the changes.
-          Collection<PartitionReplica> assigned = assignmentQueue.poll(5, TimeUnit.SECONDS);
+          Collection<PartitionReplica> assigned = assignmentQueue.poll(30, TimeUnit.SECONDS);
           Assert.assertNotNull(assigned);
           Assert.assertEquals(5, assigned.size());
 
           // Unregister from discovery, the handler should receive a change with empty collection
           cancelDiscoverable1.cancel();
-          Assert.assertTrue(assignmentQueue.poll(5, TimeUnit.SECONDS).isEmpty());
+          Assert.assertTrue(assignmentQueue.poll(30, TimeUnit.SECONDS).isEmpty());
 
           // Register to discovery again, would receive changes.
           cancelDiscoverable1 = discoveryService.register(ResolvingDiscoverable.of(discoverable1));
-          assigned = assignmentQueue.poll(5, TimeUnit.SECONDS);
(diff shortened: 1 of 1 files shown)


cdap-common/src/test/java/co/cask/cdap/common/zookeeper/coordination/ResourceCoordinatorTest.java
```diff
@@ -236,4 +245,22 @@ public InetSocketAddress getSocketAddress() {
       }
     };
   }
+
+  private Collection<PartitionReplica> waitForChange(BlockingQueue<Collection<PartitionReplica>> assignmentQueue,
```
chtyim: Why need this? The `BlockingQueue.poll` should be doing it already.

henry-cask: From one of the "fail" test run looks like call to the ``poll(5, TimeUnit.SECONDS);`` returns null which likely because that specified waiting time elapses before an element is available. 
Adding this will retry multiple poll calls.

chtyim: It returns null because there is nothing in the queue even after 5 seconds. We don't need the retry logic, but simply having a longer timeout if it's just timing issue.

BlockingQueue.poll will return immediately if there is something in the queue.

henry-cask: Or if timeout passed. We could always extend the timing issue but I thought this would be add more flexibility on retry.

chtyim: How is it more flexible?

henry-cask: Rather than just extending the timeout, we add multiple possible pollings, based on timeout, during larger period of time, 30s for now.
So, in best case in will return less than 5 sec but could wait as long as 30s before an element occur.

chtyim: So it's the same as just calling `poll(30, TimeUnit.SECONDS)` then.

henry-cask: Ah yes, I should probably just add the timeout. Not sure what was I thinking.
Will update the PR

henry-cask:
Build: 
http://builds.cask.co/browse/CDAP-RBT502-2

henry-cask:
@chtyim any objection/ comment on the proposed updated fix?

chtyim:
The changes LGTM. I am just not sure if it is the test timing issue or actually there are race condition bugs inside the resource coordinator framework

henry-cask:
Good point. I am not sure about that. I would suggest we increase the timeout for now to see if it resolve the flakiness. If it is still broken this would mean that some bigger problem.

I need to resubmit against 3.2 release branch for this PR.

PR #4267, opened 2015-10-08
CDAP-3939 improve error handling with artifact and app deployment

albertshau:
Improving the failure handling for the API call that creates an
artifact and application in the same call. If the artifact was
added successfully, but the application failed to deploy,
remove the newly added artifact so that deployment can be
re-tried. If the artifact already exists, give a more detailed
error message explaining how to deploy the app.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
@@ -313,6 +313,7 @@ public class ApplicationLifecycleService extends AbstractIdleService {
       } catch (IOException e2) {
         // if the delete fails, nothing we can do, just log it and continue on
         LOG.warn("Failed to delete artifact {} after deployment of artifact and application failed.", artifactId, e2);
+        e.addSuppressed(e2);
       }
       throw e;
     }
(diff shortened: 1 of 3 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-RBT503

sreevatsanraman:
LGTM

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/services/ApplicationLifecycleService.java
```diff
@@ -303,10 +303,21 @@ public ApplicationWithPrograms deployAppAndArtifact(Id.Namespace namespace, @Nul
...
-    return deployApp(namespace, appName, configStr, programTerminator, artifactDetail);
+    try {
+      return deployApp(namespace, appName, configStr, programTerminator, artifactDetail);
+    } catch (Exception e) {
+      // if we added the artifact, but failed to deploy the application, delete the artifact to bring us back
+      // to the state we were in before this call.
+      try {
+        artifactRepository.deleteArtifact(artifactId);
+      } catch (IOException e2) {
+        // if the delete fails, nothing we can do, just log it and continue on
```
anwar6953: I think it would be good to `e.addSuppressed(e2)` in this case.

PR #4268, opened 2015-10-08
CDAP-3961 Update dependencies to include cdap-hbase-compat-1.1

wolf31o2:
Otherwise, users will experience problems if they do not explicitly install all of the compat packages.


dereklwood:
There needs to be an associated JIRA with this, otherwise :+1:

wolf31o2:
https://issues.cask.co/browse/CDAP-3961

rohitsinha54:
LGTM

PR #4270, opened 2015-10-08
Feature/ui styling modal consolidate

astewes:
*  Consolidates modal styling into theme.less
*  Adds consistent markup across all modal instances (ie, \<h3>\ used for modal title
*  Removes "center" class from modal instances that are not cask-confirmable or cask-promptable
*  Removes all references to cdap-modal, along with any other modal helper classes that are no longer needed
*  Reduces specificity of certain modal styles when it is not needed


cdap-ui/app/features/hydrator/templates/partial/confirm.html
```diff
@@ -1,20 +1,23 @@
...
   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
   License for the specific language governing permissions and limitations under
   the License.
---> 
+-->
 
-<div class="text-center">
+<div class="modal-header">
+  <h3 class="modal-titl">Confirmation</h3>
```
elfenheart: typo on *title*

elfenheart:
I think instead of adding disabled class, you can use either ```:disabled``` or ```[disabled]``` to target the styling

ajainarayanan:
:+1: LGTM :fireworks:

PR #4272, opened 2015-10-09
Remove unnecessary logic for building dataset jar

chtyim:
- Just use the application jar for dataset module deployment.

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
@@ -23,6 +23,9 @@ import co.cask.cdap.api.dataset.DatasetProperties;
 import co.cask.cdap.api.dataset.DatasetSpecification;
 import co.cask.cdap.api.dataset.module.DatasetDefinitionRegistry;
 import co.cask.cdap.api.dataset.module.DatasetModule;
+import co.cask.cdap.common.conf.CConfiguration;
+import co.cask.cdap.common.conf.Constants;
+import co.cask.cdap.common.io.Locations;
 import co.cask.cdap.common.lang.ClassLoaders;
 import co.cask.cdap.data2.datafabric.dataset.type.ConstantClassLoaderProvider;
 import co.cask.cdap.data2.datafabric.dataset.type.DatasetClassLoaderProvider;
@@ -46,13 +49,14 @@ import com.google.common.collect.Sets;
 import com.google.common.io.ByteStreams;
 import com.google.inject.Inject;
 import org.apache.twill.discovery.DiscoveryServiceClient;
-import org.apache.twill.filesystem.LocalLocationFactory;
 import org.apache.twill.filesystem.Location;
 import org.apache.twill.internal.ApplicationBundler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2961

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -86,6 +91,12 @@ public DatasetServiceClient load(Id.Namespace namespace) throws Exception {
 
   @Override
   public void addModule(Id.DatasetModule moduleId, DatasetModule module) throws DatasetManagementException {
+    addModule(moduleId, module, null);
```
albertshau: when do we call this one without the location?

gokulavasan:
LGTM :+1:

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -241,20 +252,42 @@ public void deleteNamespace(Id.Namespace namespaceId) throws DatasetManagementEx
...
 
-  private void addModule(Id.DatasetModule moduleId, Class<?> typeClass) throws DatasetManagementException {
-    try {
-      File tempFile = File.createTempFile(typeClass.getName(), ".jar");
+  private void addModule(Id.DatasetModule moduleId, Class<?> typeClass,
+                         @Nullable Location jarLocation) throws DatasetManagementException {
+
+    Location deploymentJar = jarLocation;
+    // Only trace dependency and create new jar if jar location is not provided.
+    if (deploymentJar == null) {
```
albertshau: may be a little cleaner to have new jar creation and cleanup logic in the other method to avoid null checks.

albertshau:
minor comments, :+1:

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
chtyim: Refactored the logic.

albertshau:
:+1:

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
chtyim: This is called when trying to add system dataset module or when the JAR is not available (it will not be the case in current code base). Added javadoc in `DatasetFramework` to explain that.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/RemoteDatasetFramework.java
```diff
@@ -241,26 +258,13 @@ public void deleteNamespace(Id.Namespace namespaceId) throws DatasetManagementEx
...
-      throw new DatasetManagementException(msg, e);
-    }
-  }
-
-  private static Location createDeploymentJar(Class<?> clz, Location destination) throws IOException {
-    Location tempBundle = destination.getTempFile(".jar");
+  private Location createDeploymentJar(Class<?> clz) throws IOException {
+    File tempDir = new File(cConf.get(Constants.CFG_LOCAL_DATA_DIR),
+                            cConf.get(Constants.AppFabric.TEMP_DIR)).getAbsoluteFile();
+    tempDir.mkdirs();
```
henry-cask: This always annoy me with with java.io APIs. Should we check for return value for ``mkdirs`` call?

chtyim: for the usage here, we don't need to, since the following line is to create a temp file inside the tempDir, which will fail and throwing IOException (which we want), if the tempDir is not a directory.

chtyim:
Build passed: http://builds.cask.co/browse/CDAP-DUT2961-7

chtyim:
Squashed. Merging.

PR #4278, opened 2015-10-10
Feature/hdinsight install

dereklwood:
Script to install cdap on an hdinsight edgenode.

   - [x] clones cdap repo & uses the packer script to setup cookbooks
   - [x] sources our init scripts to get zookeeper quorum from hbase-site.xml
      - [x] a little bit of an ugly catch 22 here since our init scripts require xmllint, so it installs cdap-cli pkg first
   - [x] installs ntp (since it is missing) and installs/configures/starts CDAP

### DIFF
--- a/cdap-distributions/src/hdinsight/cdap-site.xml.template
+++ b/cdap-distributions/src/hdinsight/cdap-site.xml.template
@@ -1,0 +1,13 @@
+{
+  "cdap": {
+    "cdap_site": {
+      "zookeeper.quorum": "ZK_QUORUM",
+      "kafka.log.dir": "/var/cdap/kafka-logs"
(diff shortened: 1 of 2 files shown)


wolf31o2:
:+1: LGTM with one question/feature request... Should we specify a branch somewhere, in case packer scripts change or for testing

dereklwood:
Yeah I haven't included any hooks for release mgmt/lifecycle yet, as I want to see what our options are on the azure side.  It'll probably be a ``--branch release/3.1`` in the git command, and a cdap version in the template.  We will also need to figure out where to host this as an artifact.  I figured this was all worth a discussion and a subsequent PR.

PR #4280, opened 2015-10-10
Increase the startup memory of the SDK

awholegunch:
Increase the startup memory for the SDK to allow the WikipediaPipeline Example (and others) to run.

Fixes: https://issues.cask.co/browse/CDAP-3949


anew:
We saw the same issue when demoing at Strata. Thanks for fixing this.
LGTM.

awholegunch:
Thanks.

PR #4281, opened 2015-10-12
Fix/ui hydrator usability

ajainarayanan:
- Fixes displaying group names provided in plugin's node proxy configuration 
- Adds a border for groups in node configuration
- Removes 'Save' & 'Reset' buttons from node configuration tabs.
- Removes hardcoding 'schema' property in UI. 
   - Replaces schema with 'outputschema' as explicit label in node proxy to mark a field as output schema.
   - The reasoning behind this is to have any name for a property and assign it as output schema. The end goal is to have multiple output schema. Right now we assume only 1 output schema.


astewes:
:+1: LGTM

anwar6953:
LGTM

PR #4283, opened 2015-10-12
Update CDAP SDK docs for Windows

awholegunch:
Quick Build passed: http://builds.cask.co/browse/CDAP-DOB13-1

[Updated CDAP SDK docs](http://builds.cask.co/artifact/CDAP-DOB13/shared/build-1/Docs-HTML/3.2.0/en/developers-manual/getting-started/standalone/index.html#standalone-index)

Fix for https://issues.cask.co/browse/CDAP-3952


bdmogal:
LGTM

nitinmotgi:
@awholegunch I would also add a comment that says we generally recommend VM or Docker for Windows environment as there are very different versions.

PR #4284, opened 2015-10-12
CDAP-3870 Adding logging and metrics support for transforms 

gsps1:
Metrics context is similar to how we record filtered metrics currently. 
For Logging context,  Along with the class name we also add the stage number 
Example : co.cask.cdap.etl.transform.ValidatorTransform.stage-2

In order to show the ValidatorTransform along with the stage, i have modified the LogHandler default pattern to be the set context (``c``) instead of class name (``C``).

JIRA : https://issues.cask.co/browse/CDAP-3870


gsps1:
build running : http://builds.cask.co/browse/CDAP-DUT2966-1

gsps1:
should open against 3.2, closing this.

PR #4285, opened 2015-10-12
CDAP-3870 Adding logging and metrics support for transforms 

gsps1:
Metrics context is similar to how we record filtered metrics currently. 
For Logging context,  Along with the class name we also add the stage number 
Example : co.cask.cdap.etl.transform.ValidatorTransform.stage-2

In order to show the ValidatorTransform along with the stage, i have modified the LogHandler default pattern to be the set context (``c``) instead of class name (``C``).

JIRA : https://issues.cask.co/browse/CDAP-3870

build running : http://builds.cask.co/browse/CDAP-DUT2966-1

Reopened : https://github.com/caskdata/cdap/pull/4284


cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/TransformContext.java
```diff
@@ -55,6 +55,12 @@
   Metrics getMetrics();
 
   /**
+   * Get the stage number of the transform, this is useful for setting the context for logging in transform.
+   * @return {@link Integer} stage number
+   */
+  Integer getStageId();
```
chtyim: Why returns `Integer` but not `int`? Can it be null?

cdap-watchdog/src/main/java/co/cask/cdap/logging/LoggingConfiguration.java
```diff
@@ -51,7 +51,7 @@
   // Table used to store log metadata
   public static final String LOG_META_DATA_TABLE = "log.meta";
   // Defaults
-  public static final String DEFAULT_LOG_PATTERN = "%d{ISO8601} - %-5p [%t:%C{1}@%L] - %m%n";
+  public static final String DEFAULT_LOG_PATTERN = "%d{ISO8601} - %-5p [%t:%c{40}@%L] - %m%n";
```
albertshau: what does this change?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptFilterTransform.java
```diff
@@ -98,6 +102,10 @@ private void init() {
     ScriptEngineManager manager = new ScriptEngineManager();
     engine = manager.getEngineByName("JavaScript");
 
+    // initialize metrics and logging
+    engine.put("metrics", metrics);
+    engine.put("LOG", logger);
```
chtyim: Why `metrics` is all small, but `LOG` is all caps?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/test/java/co/cask/cdap/etl/transform/ScriptTransformTest.java
```diff
@@ -197,6 +197,8 @@ public void testComplex() throws Exception {
       "  var e = input.inner1.list[0].e;\n" +
       "  var val = power(pi.val, 3) + power(e.val, 2);\n" +
       "  print(pi); print(e);\n" +
+      "  metrics.count(\"script.transform.count\", 1);\n" +
```
albertshau: can we check that this was emitted somehow? I think there are some tests that use a mock metrics.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/TransformContext.java
gsps1: it cannot, have changed it.

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/TransformContext.java
```diff
@@ -55,6 +55,12 @@
   Metrics getMetrics();
 
   /**
+   * Get the stage number of the transform, this is useful for setting the context for logging in transform.
+   * @return stage number
+   */
+  int getStageId();
```
albertshau: what about returning the PluginID.  Would have more info, including the stage id.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptFilterTransform.java
```diff
@@ -98,6 +102,10 @@ private void init() {
     ScriptEngineManager manager = new ScriptEngineManager();
     engine = manager.getEngineByName("JavaScript");
 
+    // initialize metrics and logging
+    engine.put("metrics", metrics);
```
albertshau: should describe how to use these in the docs, otherwise nobody will know they exist.

cdap-watchdog/src/main/java/co/cask/cdap/logging/LoggingConfiguration.java
gsps1: with the pattern ``[%t:%C{1}@%L]`` we will always print the className as the context, so when we emit the log from javascript, the context would be ``sun.reflect.NativeMethodAccessorImpl``, instead we want to have the transform context, that's why we have to change the pattern to be 
[%t:%c{40}@%L] , the 40 is to show the stage number as well as the class name {Eg: validatorTransform}

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptFilterTransform.java
gsps1: i tried to be consistent with our usage in java code, but can be changed if this is not appropriate.

cdap-watchdog/src/main/java/co/cask/cdap/logging/LoggingConfiguration.java
albertshau: won't this change all our logs? Is there some way to do this only for the logger in the transform?

gsps1: this is used by the LogHandler endpoint, so this will only be reflected in the UI and REST calls and not the logs we write to the files.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptFilterTransform.java
chtyim: In java, because `LOG` is a global static while `metrics` is a class field. Since this is Javascript, we can have a convention saying all global variables available to the script is in all caps. Better yet, we can namespace the CDAP variables by a `CDAP` namespace:

```java
  engine.put("CDAP", new ScriptGlobal());
  engine.eval("CDAP.logger.info('Hello, World')");

  public static final class ScriptGlobal {
    public final Logger logger;
    public final Metrics metrics;
   ...
  }
```

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/TransformContext.java
albertshau: ignore this, plugins already know their type and name

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptFilterTransform.java
gsps1: ok, i am just wondering how we should set validators.. should we namespace them with CDAP as well ? currently its set as separate global variables and referenced liked, ``Eg: coreValidator.maxLength(input.body, 10)``

chtyim: Is that specific for one transform plugin?

gsps1: yes

chtyim: In general, there are two ways to do this. One is through a global context object (like `CDAP` I mentioned above). The other is through a context object passed to the function. Personally I like the function argument better since using global is always less preferred.

gsps1:
@chtyim @albertshau have addressed your comments. Also added the relevant UI changes (updating json) and docs changes to this PR. @awholegunch can you please review the UI changes. Thanks

cdap-watchdog/src/main/java/co/cask/cdap/logging/LoggingConfiguration.java
albertshau: so what will the old logs look like now? Seems wrong to change UI and REST output just for this transform.

albertshau:
a couple minor comments, plus a question about the log format change.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ValidatorTransform.java
```diff
@@ -56,21 +57,31 @@
...
+    "         metrics.count(\"invalid.date\", 1);" +
     "      } else if (!coreValidator.isUrl(input.url)) { " +
     "         isValid = false; errMsg = \"invalid url\"; errCode = 7;" +
+    "         metrics.count(\"invalid.url\", 1);" +
     "      } else if (!coreValidator.isInRange(input.content_length, 0, 1024 * 1024)) {" +
     "         isValid = false; errMsg = \"content length >1MB\"; errCode = 10;" +
+    "         metrics.count(\"invalid.body.size\", 1);" +
+    "      }" +
+    "      if (!isValid) {" +
+    "       logger.warn(\"Validation failed for record {}\", input);" +
```
awholegunch: Does this work in Javascript? Should it not be something such as:
```
logger.warn(\"Validation failed for record \" + input.id);" +

cdap-watchdog/src/main/java/co/cask/cdap/logging/LoggingConfiguration.java
albertshau: ok understand now, this looks fine

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ValidatorTransform.java
gsps1: it will work, as its calling a java object

albertshau:
:+1:

gsps1:
Rebased, will merge after build - http://builds.cask.co/browse/CDAP-DUT2966-3

gsps1:
Green, Merging

PR #4286, opened 2015-10-12
replaces calc-height directive with css calc

astewes:
*  Removes calculate-height directive and associated attributes
*  Adds use of calc() css to calculate container height when only one left panel group is opened

panels


ajainarayanan:
:+1: #awesome!

PR #4287, opened 2015-10-12
fixes duration cell rendering issue

astewes:
*  Modifies ng-if logic within duration cell to prevent emdash from appearing underneath text

buttons


astewes:
@ajainarayanan your change :+1: LGTM

PR #4290, opened 2015-10-12
fixes styling of alert rows

astewes:
*  Adds correct background to alert rows (via .alert-danger class)
*  Adds additional styling to make my-error more consistent with the UI

alert
I


ajainarayanan:
:+1: LGTM

PR #4292, opened 2015-10-13
(CDAP-3951) Have S3BatchConfig extend FileBatchConfig, so users can r…

bdmogal:
…e-use fileSystemProperties to send file system configurations to Hadoop, like in FileBatchSource.

- Remove the unused fileSystem property from FileBatchSource. Other minor cleanup.
- Introduce fileSystemProperties for S3 Batch Sinks.

(CDAP-3815) Fix FileBatchFilter to apply the provided regex filter on file names instead of on the complete path. Also update S3 MapReduce test to test this behavior.

Jiras: 
[CDAP-3951](https://issues.cask.co/browse/CDAP-3951)
[CDAP-3815](https://issues.cask.co/browse/CDAP-3815)

Build: http://builds.cask.co/browse/CDAP-RBT506-1

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/BatchFileFilter.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/BatchFileFilter.java
@@ -80,7 +80,8 @@ public class BatchFileFilter extends Configured implements PathFilter {
 
     //filter by file name using regex from configuration
     if (!useTimeFilter) {
-      Matcher matcher = regex.matcher(filePathName);
+      String fileName = path.getName();
+      Matcher matcher = regex.matcher(fileName);
       return matcher.matches();
     }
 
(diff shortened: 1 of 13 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/common/BatchFileFilter.java
```diff
@@ -80,7 +80,8 @@ public boolean accept(Path path) {
 
     //filter by file name using regex from configuration
     if (!useTimeFilter) {
-      Matcher matcher = regex.matcher(filePathName);
```
anwar6953: The regex is meant only to match against the last component of the file Path, or the entire file path?

bdmogal: Its a regex filter on the ``fileName``. The directory is passed separately. @sreevatsanraman that's how its supposed to work, right?

anwar6953:
Just one concern. Otherwise, LGTM

sreevatsanraman:
@bdmogal yes. That's the right behavior.

bdmogal:
Thanks!

bdmogal:
Build passed. Merging.

PR #4296, opened 2015-10-13
CDAP-2751 Added runid for the custom actions.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-2751

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
@@ -35,6 +35,7 @@ import co.cask.cdap.app.program.Program;
 import co.cask.cdap.app.runtime.Arguments;
 import co.cask.cdap.app.runtime.ProgramOptions;
 import co.cask.cdap.app.store.Store;
+import co.cask.cdap.common.ServiceUnavailableException;
 import co.cask.cdap.common.app.RunIds;
 import co.cask.cdap.common.conf.CConfiguration;
 import co.cask.cdap.common.conf.Constants;
@@ -67,6 +68,7 @@ import com.google.common.base.Supplier;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.reflect.TypeToken;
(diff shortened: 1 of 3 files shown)


sagarkapare:
Build is running here - http://builds.cask.co/browse/CDAP-DUT2967-2

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -388,13 +391,46 @@ private void executeFork(final ApplicationSpecification appSpec, WorkflowForkNod
...
+                                  actionNode.getNodeId(), startTimeInSeconds, null);
+
+    ProgramRunStatus runStatus = ProgramRunStatus.COMPLETED;
+
+    try {
+      executeAction(actionNode, instantiator, classLoader, token);
+    } catch (Throwable t) {
+      runStatus = ProgramRunStatus.FAILED;
+      if (Iterables.size(Iterables.filter(Throwables.getCausalChain(t), InterruptedException.class)) > 0) {
+        runStatus = ProgramRunStatus.KILLED;
```
chtyim: Why is it so? Seems not a very good way to tell if a custom action is killed or failed based on the exception type.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -388,13 +391,46 @@ private void executeFork(final ApplicationSpecification appSpec, WorkflowForkNod
...
+      }
+      Throwables.propagateIfPossible(t, Exception.class);
```
chtyim: I don't think you need this and the next line. Simply do a `throw t` should work (since you have `throws Exception` in the method signature).

poornachandra:
@sagarkapare Since custom actions now have runId, and they access Datasets/Streams(?), we should record lineage data access from custom actions too.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
sagarkapare: For custom actions we do not have program runner/listeners. 
I thought of using the `runningThread` so that
```java
catch (Throwable t) {
   if (runningThread == null) {
      // mark custom action as KILLED    
   }
  else {
     // mark custom action as FAILED
  }
}
```
However above code only handles the explicit `stop` of the Workflow. If custom action is running as a part of the FORK branch and if program on another branch is failed, the custom action need to be marked as KILLED. So I used checks for the `InterruptedException`.

chtyim: That sounds pretty hacky. Can we just have a uniform way of executing actions so that they can be handled with the same code path?

sagarkapare:
Closing this PR. Will re-open after implementing the execution of the custom actions as other programs.

PR #4299, opened 2015-10-13
(CDAP-3887) Added a configurable idle connection timeout in the Router. 

bdmogal:
Open connections are closed once this connection timeout is hit.

Jira: [CDAP-3887](https://issues.cask.co/browse/CDAP-3887)
Build: http://builds.cask.co/browse/CDAP-RBT504

### DIFF
--- a/cdap-common/src/main/resources/cdap-default.xml
+++ b/cdap-common/src/main/resources/cdap-default.xml
@@ -1193,6 +1193,44 @@
     </description>
   </property>
 
+  <property>
+    <name>router.connection.backlog</name>
+    <value>20000</value>
+    <description>Determines the connection backlog in the Router Server.</description>
+  </property>
+
+  <property>
+    <name>router.server.boss.threads</name>
+    <value>1</value>
+    <description>Deterermines the number of boss threads in the Router Server.</description>
+  </property>
+
+  <property>
+    <name>router.server.worker.threads</name>
+    <value>10</value>
+    <description>Determines the number of worker threads in the Router Server.</description>
+  </property>
+
+  <property>
+    <name>router.client.boss.threads</name>
+    <value>1</value>
+    <description>Determines the number of boss threads in the Router Client.</description>
+  </property>
+
+  <property>
+    <name>router.client.worker.threads</name>
+    <value>10</value>
+    <description>Determines the number of worker threads in in the Router Client.</description>
+  </property>
+
+  <property>
+    <name>router.connection.timeout.secs</name>
+    <value>15</value>
+    <description>
+      Determines the number of seconds after which idle router connections are closed.
+    </description>
+  </property>
+
 
   <!-- Security Configuration -->
 
--- a/cdap-common/src/main/resources/cdap-default.xml
+++ b/cdap-common/src/main/resources/cdap-default.xml
@@ -1094,7 +1094,7 @@
     <name>notification.transport.system</name>
     <value>kafka</value>
     <description>
-      Transport system used by the Notification system - can be kafka/stream
+      Transport system used by the Notification system: can be kafka/stream
     </description>
   </property>
 
@@ -1147,11 +1147,49 @@
   </property>
 
   <property>
+    <name>router.client.boss.threads</name>
+    <value>1</value>
+    <description>The number of boss threads in the router client</description>
+  </property>
+
+  <property>
+    <name>router.client.worker.threads</name>
+    <value>10</value>
+    <description>The number of worker threads in the router client</description>
+  </property>
+
+  <property>
+    <name>router.connection.backlog</name>
+    <value>20000</value>
+    <description>The connection backlog in the router server</description>
+  </property>
+
+  <property>
+    <name>router.connection.timeout.secs</name>
+    <value>15</value>
+    <description>
+      The number of seconds after which idle router connections are closed
+    </description>
+  </property>
+
+  <property>
+    <name>router.server.boss.threads</name>
+    <value>1</value>
+    <description>The number of boss threads in the router server</description>
+  </property>
+
+  <property>
     <name>router.server.port</name>
     <value>${router.bind.port}</value>
     <description>
       Router port to which CDAP UI connects
     </description>
+  </property>
+
+  <property>
+    <name>router.server.worker.threads</name>
+    <value>10</value>
+    <description>The number of worker threads in the Router server</description>
   </property>
 
   <property>
@@ -1190,44 +1228,6 @@
     <value>false</value>
     <description>
       Determines if the webapp listening service should be started at router
-    </description>
-  </property>
-
-  <property>
-    <name>router.connection.backlog</name>
-    <value>20000</value>
-    <description>Determines the connection backlog in the Router Server.</description>
-  </property>
-
-  <property>
-    <name>router.server.boss.threads</name>
-    <value>1</value>
-    <description>Deterermines the number of boss threads in the Router Server.</description>
-  </property>
-
-  <property>
-    <name>router.server.worker.threads</name>
-    <value>10</value>
-    <description>Determines the number of worker threads in the Router Server.</description>
-  </property>
-
-  <property>
-    <name>router.client.boss.threads</name>
-    <value>1</value>
-    <description>Determines the number of boss threads in the Router Client.</description>
-  </property>
-
-  <property>
-    <name>router.client.worker.threads</name>
-    <value>10</value>
-    <description>Determines the number of worker threads in in the Router Client.</description>
-  </property>
-
-  <property>
-    <name>router.connection.timeout.secs</name>
-    <value>15</value>
-    <description>
-      Determines the number of seconds after which idle router connections are closed.
     </description>
   </property>
 
--- a/cdap-gateway/src/main/java/co/cask/cdap/gateway/router/NettyRouter.java
+++ b/cdap-gateway/src/main/java/co/cask/cdap/gateway/router/NettyRouter.java
@@ -221,7 +221,7 @@ public class NettyRouter extends AbstractIdleService {
           }
           pipeline.addLast("tracker", connectionTracker);
           // idle state handler
-          pipeline.addLast("idle-state-handler",
+          pipeline.addLast("idle-event-generator",
                            new IdleStateHandler(timer, connectionTimeout, connectionTimeout, connectionTimeout));
           pipeline.addLast("router-idle-state-handler", new RouterIdleStateHandler());
           pipeline.addLast("http-response-encoder", new HttpResponseEncoder());
@@ -283,7 +283,7 @@ public class NettyRouter extends AbstractIdleService {
         ChannelPipeline pipeline = Channels.pipeline();
         pipeline.addLast("tracker", connectionTracker);
         // idle state handler
-        pipeline.addLast("idle-state-handler",
+        pipeline.addLast("idle-event-generator",
                          new IdleStateHandler(timer, connectionTimeout, connectionTimeout, connectionTimeout));
         pipeline.addLast("router-idle-state-handler", new RouterIdleStateHandler());
         pipeline.addLast("request-encoder", new HttpRequestEncoder());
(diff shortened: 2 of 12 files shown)


cdap-gateway/src/main/java/co/cask/cdap/gateway/router/NettyRouter.java
```diff
@@ -236,6 +220,10 @@ public ChannelPipeline getPipeline() throws Exception {
             pipeline.addLast("ssl", sslHandlerFactory.create());
           }
           pipeline.addLast("tracker", connectionTracker);
+          // idle state handler
+          pipeline.addLast("idle-state-handler",
```
poornachandra: Calling this `idle-event-generator` might make the intent for `IdleStateHandler` clearer.

cdap-common/src/main/resources/cdap-default.xml
```diff
@@ -1193,6 +1193,44 @@
...
+
+  <property>
+    <name>router.client.worker.threads</name>
+    <value>10</value>
+    <description>Determines the number of worker threads in in the Router Client.</description>
+  </property>
+
+  <property>
+    <name>router.connection.timeout.secs</name>
+    <value>15</value>
```
poornachandra: I think 15 seconds of default timeout is a bit low.

cdap-gateway/src/main/java/co/cask/cdap/gateway/router/handlers/RouterIdleStateHandler.java
```diff
@@ -0,0 +1,50 @@
...
+  private final AtomicInteger numConnectionsClosed = new AtomicInteger(0);
+
+  @Override
+  public void channelIdle(ChannelHandlerContext ctx, IdleStateEvent e) throws Exception {
+    LOG.debug("Received idle event '{}' for channel '{}'.", e.getState(), ctx.getName());
+    if (IdleState.ALL_IDLE == e.getState()) {
+      // No data has been sent or received for a while. Close channel.
+      ctx.getChannel().close().addListener(ChannelFutureListener.CLOSE);
+      numConnectionsClosed.incrementAndGet();
+      LOG.trace("No data has been sent or received for channel '{}' for more than the configured idle timeout. " +
```
poornachandra: It might be a good idea to log source ip:port and dest ip:port for the connection closed.

cdap-gateway/src/main/java/co/cask/cdap/gateway/router/handlers/RouterIdleStateHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private final AtomicInteger numConnectionsClosed = new AtomicInteger(0);
+
+  @Override
+  public void channelIdle(ChannelHandlerContext ctx, IdleStateEvent e) throws Exception {
+    LOG.debug("Received idle event '{}' for channel '{}'.", e.getState(), ctx.getName());
+    if (IdleState.ALL_IDLE == e.getState()) {
+      // No data has been sent or received for a while. Close channel.
+      ctx.getChannel().close().addListener(ChannelFutureListener.CLOSE);
+      numConnectionsClosed.incrementAndGet();
```
poornachandra: If we are not exposing `numConnectionsClosed` as a metric, then I see no point in counting it.

cdap-gateway/src/test/java/co/cask/cdap/gateway/router/NettyRouterTestBase.java
```diff
@@ -388,6 +364,29 @@ public void testConnectionClose() throws Exception {
     Assert.assertEquals(times, defaultServer1.getNumRequests() + defaultServer2.getNumRequests());
   }
 
+  @Test
+  public void testConnectionNoIdleTimeout() throws Exception {
+    // 4700ms is just under the configured idle timeout of 5 seconds
+    URL url = new URL(resolveURI(Constants.Router.GATEWAY_DISCOVERY_NAME, "/v1/timeout/4700"));
```
poornachandra: If you have access to `cConf` over is here, then I think it is a good idea to subtract a few hundred millis from `Constants.Router.CONNECTION_TIMEOUT_SECS` rather than hard-coding.

cdap-gateway/src/test/java/co/cask/cdap/gateway/router/NettyRouterTestBase.java
```diff
@@ -388,6 +364,29 @@ public void testConnectionClose() throws Exception {
...
+    HttpURLConnection urlConnection = openURL(url);
+    urlConnection.getResponseCode();
+  }
+
+  @Test(expected = SocketException.class)
+  public void testConnectionIdleTimeout() throws Exception {
+    // 5300ms is just over the configured idle timeout of 5 seconds
+    URL url = new URL(resolveURI(Constants.Router.GATEWAY_DISCOVERY_NAME, "/v1/timeout/5300"));
```
poornachandra: Same here.

cdap-gateway/src/test/java/co/cask/cdap/gateway/router/NettyRouterTestBase.java
```diff
@@ -388,6 +364,29 @@ public void testConnectionClose() throws Exception {
...
+    HttpURLConnection urlConnection = openURL(url);
+    urlConnection.getResponseCode();
+  }
+
+  // because the connection is closed after a timeout, the NettyResponseFuture.get() throws an ExecutionException
+  @Test(expected = ExecutionException.class)
+  @Category(SlowTests.class)
+  public void testUploadWithTimeout() throws Exception {
+    testUpload(5300, TimeUnit.MILLISECONDS);
```
poornachandra: Same here.

cdap-common/src/main/resources/cdap-default.xml
wolf31o2: So any keepalives need to be less than this to keep the connection open? I think that this is reasonable, as the defaults in Apache HTTPd are as "high" as 15 seconds for 1.3/2.0 and as low as 5 seconds with 2.2+... https://en.wikipedia.org/wiki/HTTP_persistent_connection#HTTP_1.1

cdap-gateway/src/main/java/co/cask/cdap/gateway/router/handlers/RouterIdleStateHandler.java
```diff
@@ -0,0 +1,50 @@
...
+import org.jboss.netty.handler.timeout.IdleState;
+import org.jboss.netty.handler.timeout.IdleStateAwareChannelHandler;
+import org.jboss.netty.handler.timeout.IdleStateEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Handles states when the router has been idle for a configured time interval.
```
awholegunch: Should this be "Handles state when..." (singular)?

bdmogal: There are multiple states - ``READER_IDLE``, ``WRITER_IDLE``, ``ALL_IDLE``. Currently we only special case ``ALL_IDLE``, but we may handle others in future, so I think this is ok.

poornachandra:
A few more questions -
1. What happens on the client side when a connection gets closed due to timeout? Is there a way to make the HTTP client to automatically retry in this case?
2. What happens on the Router, when a server-side connection is idle and is about to get timed out, then a new client-side connection comes that needs to to talk to the the about to be closed connection?

cdap-common/src/main/resources/cdap-default.xml
poornachandra: Considered from the resource recovery angle, 15 seconds or less makes sense.

cdap-gateway/src/main/java/co/cask/cdap/gateway/router/handlers/RouterIdleStateHandler.java
```diff
@@ -0,0 +1,47 @@
...
+public class RouterIdleStateHandler extends IdleStateAwareChannelHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(RouterIdleStateHandler.class);
+
+  @Override
+  public void channelIdle(ChannelHandlerContext ctx, IdleStateEvent e) throws Exception {
+    LOG.debug("Received idle event '{}' for channel '{}'.", e.getState(), ctx.getName());
+    if (IdleState.ALL_IDLE == e.getState()) {
+      // No data has been sent or received for a while. Close channel.
+      Channel channel = ctx.getChannel();
+      channel.close().addListener(ChannelFutureListener.CLOSE);
```
poornachandra: `ChannelFutureListener.CLOSE` also closes a channel, so I don't think we need it here.

bdmogal:
@poornachandra:
> 1. What happens on the client side when a connection gets closed due to timeout? Is there a way to make the HTTP client to automatically retry in this case?

- Opened an HttpURLConnection. Verified the response. Did not wait to hit the idle connection timeout. Made another request to the same endpoint. The same connection gets re-used.
- Opened an HttpURLConnection. Verified the response. Waited to hit the idle connection timeout. Made another request to the same endpoint. A new connection gets created.

cdap-gateway/src/main/java/co/cask/cdap/gateway/router/NettyRouter.java
```diff
@@ -154,22 +140,20 @@ public NettyRouter(CConfiguration cConf, SConfiguration sConf, @Named(Constants.
...
       if (webAppEnabled) {
-        this.serviceToPortMap.put(Constants.Router.WEBAPP_DISCOVERY_NAME,
-                                  Integer.parseInt(cConf.get(Constants.Router.WEBAPP_PORT,
-                                                             Constants.Router.DEFAULT_WEBAPP_PORT)));
+        this.serviceToPortMap.put(Constants.Router.WEBAPP_DISCOVERY_NAME, cConf.getInt(Constants.Router.WEBAPP_PORT));
       }
       this.sslHandlerFactory = null;
     }
+    this.connectionTimeout = cConf.getInt(Constants.Router.CONNECTION_TIMEOUT_SECS);
+    this.timer = new HashedWheelTimer();
```
chtyim: You need to stop the timer, otherwise the thread will hang around.

chtyim: Also, good to provide a `ThreadFactory` so that the thread created is named properly.

cdap-common/src/main/resources/cdap-default.xml
chtyim: So this might be problematic for large app jar deployment? An artifact deployment involves opening the jar and inspecting classes inside (the larger the jar, potentially more classes to inspect). It also involves copy the jar from master box to HDFS, which time taken is proportional to file size as well.

chtyim: We may need to consider turning long server side operations into async operations (like execute query) so that it won't hold the connection open without any activities.

poornachandra: Maybe we can add the idle state handler only after the response has been sent? This way any in-progress requests will not be affected by the idle timeout. This means we'll need to remove the idle state handler from pipeline every time we get a new request.

chtyim: I think the original problem that this PR is trying to solve is for inflight request with the firewall blocking the forwarding request from the router to the cluster. This mean timeout also needs to apply for inflight requests.

poornachandra:
> - Opened an HttpURLConnection. Verified the response. Did not wait to hit the idle connection timeout. Made another request to the same endpoint. The same connection gets re-used.
> - Opened an HttpURLConnection. Verified the response. Waited to hit the idle connection timeout. Made another request to the same endpoint. A new connection gets created.

Cool. RFC 7230 also says the same thing - "Connections can be closed at any time, with or without intention. Implementations ought to anticipate the need to recover from asynchronous close events."

cdap-common/src/main/resources/cdap-default.xml
poornachandra: We have separate pipelines for client and server connections, and we can set the idle state handlers independently on them for inflight requests, right?

bdmogal:
This PR requires some re-work. We will de-scope this for 3.2 and get back to this for 3.3.

PR #4300, opened 2015-10-13
fixes instance control styling in Safari

astewes:
*  Fixes instance control (number input) in Service detail view
*  As styling of the number input is only possible with webkit, non-webkit browsers will default to their native styling for this element (see graphics below)

Safari:
safari

Chrome:
chrome

Firefox:
ff


ajainarayanan:
:+1: LGTM

PR #4301, opened 2015-10-13
Upgrade kafka version to 0.8.2.2 to include the fix for KAFKA-1228

anwar6953:
Upgrade kafka version to 0.8.2.2 to include the fix for https://issues.apache.org/jira/browse/KAFKA-1228.

Tested on SDK, this resolves the issue we were facing in CDAP-3963.
Will test on cluster shortly.

https://issues.cask.co/browse/CDAP-3963
http://builds.cask.co/browse/CDAP-RBT507-4


chtyim:
LGTM. Can you test it on the cluster as well? Need to make sure it works with Twill as well, since Twill also use the Partitioner interface. Hopefully the type erasure make it binary compatible.

anwar6953:
Tested against 3.2.0 cluster.

chtyim:
LGTM

PR #4302, opened 2015-10-13
CDAP-3990 allow adding the same dataset instance or module

albertshau:
Fixed a bug where the dataset configurer would error out if the
same dataset instance or module was added multiple times.
Enhancing the logic to check for equality, so that an exception
is only thrown if the same instance or module is added with a
different value.

### DIFF
--- a/cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
+++ b/cdap-api/src/main/java/co/cask/cdap/internal/api/DefaultDatasetConfigurer.java
@@ -87,8 +87,12 @@ public class DefaultDatasetConfigurer implements DatasetConfigurer {
   public void addStream(Stream stream) {
     checkArgument(stream != null, "Stream cannot be null.");
     StreamSpecification spec = stream.configure();
-    checkArgument(!streams.containsKey(spec.getName()),
-                  "Stream %s has already been added. Remove the duplicate addition.", spec.getName());
+
+    StreamSpecification existingSpec = streams.get(spec.getName());
+    if (existingSpec != null && !existingSpec.equals(spec)) {
+      throw new IllegalArgumentException(String.format("Stream '%s' was added multiple times with different specs. " +
+        "Please resolve the conflict so that there is only one spec for the stream.", spec.getName()));
+    }
     streams.put(spec.getName(), spec);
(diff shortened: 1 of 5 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-RBT508

chtyim:
LGTM

anew:
I am not sure about the approach. I think we should support the following:
1. Two different plugins should be able to install different dataset modules, but using  the same name (the two plugins don't know about each other, so they both might use a name like "customModule").
2. Two different plugins should be able to install the same module with different names. (one uses "partitionedFileSetModule", the other uses "pfs"). 

In case 1, we could allow that prefixing the name that the plugin provides with the plugin name. (Module names are not used by apps at runtime, they are only used internally by the dataset framework to track dependecies between modules).

For case 2, we need to make sure that module only gets added once, because the second time it would fail due to its dataset types already existing.

chtyim:
@anew but then how to handle case when pipeline wants two plugins actually sharing the same dataset?

albertshau:
This particular change is not directly related to plugins, though it came up because I tried to add the same custom dataset in both an etl source and sink, and app deployment failed with this error.  It's simply relaxing a check that was too restrictive.

Whether or not a plugin should be isolated from other dataset modules is ultimately up to the application logic. For ETL, there are both cases where it makes sense and where it doesn't.  It doesn't make sense for the use case mentioned above, where multiple plugins actually want to use the same dataset.  But it does make sense it other places where there are unintentional conflicts. 

Because we have situations where we want to share, I don't think we can prefix the datasets and dataset modules.  To handle the case where there are unintentional conflicts, maybe it makes sense to extend the platform capabilities so that an application can react to the presence of an existing dataset or dataset type in different ways.  Today we always continue on with app deployment if the object already exists, but it may make sense for the application to compare what's pre-existing to what it's trying to add and act accordingly.

anew:
I agree that this will need more work in the future. For now, this fix solves the most common use case, where two plugins share the same module/code. LGTM

anew:
@albertshau can you open (a) Jira(s) for the future work that you suggested? Thanks :)

albertshau:
opened CDAP-3994 and CDAP-3995 to address

PR #4303, opened 2015-10-13
CDAP-3822 Don't propagate the failure to create the Spark framework c…

anwar6953:
Opening PR for review of the approach, which I am unsure of.

Don't propagate the failure to create the Spark framework classloader. Instead, try to do without it.

https://issues.cask.co/browse/CDAP-3822


anwar6953:
Closing this PR.
The proper fix is actually to the test case of the kafka-pack repo.
Additional details are in the comments of JIRA: https://issues.cask.co/browse/CDAP-3822

PR #4304, opened 2015-10-14
changes node config tab columb widths

astewes:
*  Shrinks column with for input and output columns (both studio and pipeline detail views)

nodes


ajainarayanan:
:+1: LGTM

PR #4305, opened 2015-10-14
fixes instance control styling in Safari

astewes:
*  Fixes styling of Services instance control number input

Safari:
safari


ajainarayanan:
:+1: LGTM

PR #4307, opened 2015-10-14
[CDAP-961] implementation of dynamic datasets in all programs

anew:


cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
```diff
@@ -63,10 +68,16 @@
...
   private final Map<String, Plugin> plugins;
+  private final Transaction transaction;
 
   private MultipleOutputs multipleOutputs;
   private TaskInputOutputContext<?, ?, KEYOUT, VALUEOUT> context;
 
+  // keeps track of all tx-aware datasets to perform the transaction lifecycle for them. Note that
+  // the transaction is already started, and it will be committed or aborted outside of this task.
+  // TODO: (CDAP-3983) The datasets should be managed by the dataset context. That requires TEPHRA-99.
+  private Set<TransactionAware> txAwares = Sets.newIdentityHashSet();
```
chtyim: final?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
```diff
@@ -192,8 +208,38 @@ public LoggingContext getLoggingContext() {
     return loggingContext;
   }
 
+  //---- following are methods to manage transaction lifecycle for the datasets. This needs to
+  //---- be refactored after [TEPHRA-99] and [CDAP-3893] are resolved.
+
+  private void initializeTransactionAwares() {
```
chtyim: So this method is just for starting TX over static datasets in the dataset cache, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
```diff
@@ -192,8 +208,38 @@ public LoggingContext getLoggingContext() {
...
+   * the dataset is a transaction-aware, it starts the transaction and remembers the dataset.
+   */
+  @Override
+  public <T extends Dataset> T getDataset(String name, Map<String, String> arguments)
+    throws DatasetInstantiationException {
+    T dataset = super.getDataset(name, arguments);
+    if (dataset instanceof TransactionAware) {
+      TransactionAware txAware = (TransactionAware) dataset;
+      txAwares.add(txAware);
+      txAware.startTx(transaction);
```
chtyim: Do you need to check if the `txAware` is already in the `txAwares` set (by the result of the `add()` method) before calling `startTx`? Otherwise `startTx` might get called multiple times on the same instance?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
anew: done

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
```diff
@@ -53,16 +54,15 @@
...
    */
   BodyConsumerAdapter(DelayedHttpServiceResponder responder, HttpContentConsumer delegate,
-                      DatasetContext datasetContext, TransactionContext txContext) {
+                      TransactionalHttpServiceContext serviceContext) {
     this.responder = responder;
     this.delegate = delegate;
-    this.txContext = txContext;
-    this.datasetContext = datasetContext;
-    this.transactional = createTransactional(txContext, datasetContext);
+    this.txContext = serviceContext.newTransactionContext();
```
chtyim: I think the creation should moved into `txExecute` so that a new context is being creating for the transactional execution and get dismissed after use.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
anew: right. And there was a small bug there, too: We do not start a transaction context in the dataset cache (because we have to reuse an existing tx), hence getTransactionAwares() would return nothing.  Fixed that now by adding an extra method that returns all static tx-awares regardless of whether a tx context was created.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/batch/BasicMapReduceTaskContext.java
anew: Yes.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/service/http/BodyConsumerAdapter.java
anew: First I thought you are right about that. However, the delegate can call transactional execute multiple times in each onReceived(). I think it is best to dismiss the txContext only at the end of the request, that is in onFinish() and onError()

anew: Done

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/spark/ExecutionSparkContext.java
```diff
@@ -267,15 +274,13 @@ public ExecutionSparkContext(ApplicationSpecification appSpec,
 
   @Override
   public synchronized <T extends Dataset> T getDataset(String name, Map<String, String> arguments) {
-    Map<String, String> datasetArgs = RuntimeArguments.extractScope(Scope.DATASET, name, getRuntimeArguments());
```
chtyim: Why this is removed? So we don't do argument extraction from program runtime arguments anymore?

chtyim: Oh, figured that it's now belongs to the dataset cache. All good.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
```diff
@@ -528,6 +521,7 @@ private void destroy(WorkflowActionSpecification actionSpec, WorkflowAction acti
       LOG.warn("Exception on WorkflowAction.destroy(): {}", actionSpec, t);
       // Just log, but not propagate
     } finally {
+
```
chtyim: nit: remove extra new line

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/TransactionExecutorModule.java
```diff
@@ -0,0 +1,58 @@
...
+import co.cask.tephra.TransactionAware;
+import co.cask.tephra.TransactionContext;
+import co.cask.tephra.TransactionExecutor;
+import co.cask.tephra.TransactionSystemClient;
+import com.google.common.base.Supplier;
+import com.google.inject.AbstractModule;
+import com.google.inject.Inject;
+
+/**
+ * DataFabricLocalModule defines the Local/HyperSQL bindings for the data fabric.
```
chtyim: Seems not a right description of the class.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/workflow/WorkflowDriver.java
anew: sure

cdap-data-fabric/src/main/java/co/cask/cdap/data/runtime/TransactionExecutorModule.java
anew: oops

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetCache.java
```diff
@@ -0,0 +1,234 @@
...
+
+  @Override
+  public TransactionContext get() {
+    return newTransactionContext();
+  }
+
+  /**
+   * @return the transaction-aware datasets that participate in the current transaction.
+   */
+  public abstract Iterable<TransactionAware> getTransactionAwares();
```
chtyim: So this one only returns "non-extra" dataset (maybe better call those static dataset?)? Maybe better to document it clearly.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetCache.java
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Add an extra transaction aware to the static datasets. This is a transaction aware that
+   * is not instantiated through this factory, but needs to participate in every transaction.
+   */
+  public void addExtraTransactionAware(TransactionAware txAware) {
```
chtyim: So this method should only be called before the txContext has started, right? Otherwise the txAware won't be participant in the current transaction?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetCache.java
```diff
@@ -0,0 +1,234 @@
...
+   */
+  public void removeExtraTransactionAware(TransactionAware txAware) {
+    extraTxAwares.remove(txAware);
+  }
+
+  /**
+   * Close and dismiss all datasets that were obtained through this factory, and destroy the factory.
+   */
+  public void close() {
+    for (TransactionAware txAware : extraTxAwares) {
```
chtyim: Seems not matching with the javadoc. The javadoc said it only close datasets obtained through this cache, meaning it shouldn't includes extra datasets?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetCache.java
anew: Actually it returns all tx-awares that are in the current transaction context. Or nothing if no tx context has been created. I updated the javadocs.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetCache.java
anew: Correct. Or, better to change it to add it to the current tx context. That makes it foolproof

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
```diff
@@ -0,0 +1,206 @@
...
+import javax.annotation.ParametersAreNonnullByDefault;
+
+/**
+ * Implementation of {@link DatasetContext} that allows to dynamically load datasets
+ * into a started {@link TransactionContext}. Datasets acquired from this context are distinct from any
+ * Datasets instantiated outside this class.
+ */
+public class SingleThreadDatasetCache extends DynamicDatasetCache {
+
+  private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(SingleThreadDatasetCache.class);
```
chtyim: Why full name?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetCache.java
anew: Done

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/DynamicDatasetCache.java
anew: True. I first thought it should not, but then again, it seems better for sanity to close them. In CDAP, only flows use this feature for stream/queue consumers/producers, and it does normally remove them all before the context gets closed. But if in some error condition it doesn't, then it make sense to close them here.

anew: Updated the javadoc

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
anew: Probably IntelliJ magic/madness.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/MultiThreadDatasetCache.java
```diff
@@ -0,0 +1,139 @@
...
+                                 final MetricsContext metricsContext,
+                                 @Nullable final Map<String, Map<String, String>> staticDatasets) {
+    super(instantiator, txClient, namespace, runtimeArguments, metricsContext);
+    this.perThreadMap = CacheBuilder.newBuilder()
+      .softValues()
+      .removalListener(new RemovalListener<Long, DynamicDatasetCache>() {
+        @Override
+        @ParametersAreNonnullByDefault
+        public void onRemoval(RemovalNotification<Long, DynamicDatasetCache> notification) {
+          DynamicDatasetCache factory = notification.getValue();
```
chtyim: nit: Better call the variable `cache`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/MultiThreadDatasetCache.java
```diff
@@ -0,0 +1,139 @@
...
+   */
+  public MultiThreadDatasetCache(final SystemDatasetInstantiator instantiator,
+                                 final TransactionSystemClient txClient,
+                                 final Id.Namespace namespace,
+                                 final Map<String, String> runtimeArguments,
+                                 final MetricsContext metricsContext,
+                                 @Nullable final Map<String, Map<String, String>> staticDatasets) {
+    super(instantiator, txClient, namespace, runtimeArguments, metricsContext);
+    this.perThreadMap = CacheBuilder.newBuilder()
+      .softValues()
```
chtyim: Better be `weakValues`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/MultiThreadDatasetCache.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  @Override
+  public void close() {
+    super.close();
+    perThreadMap.invalidateAll();
+  }
+
+  @Override
+  public synchronized <T extends Dataset> T getDataset(DatasetCacheKey key, boolean bypass)
```
chtyim: why `synchronized`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
```diff
@@ -0,0 +1,206 @@
...
+    this.datasetCache = CacheBuilder.newBuilder().removalListener(
+      new RemovalListener<DatasetCacheKey, Dataset>() {
+        @Override
+        @ParametersAreNonnullByDefault
+        public void onRemoval(RemovalNotification<DatasetCacheKey, Dataset> notification) {
+          if (notification.getValue() != null) {
+            try {
+              notification.getValue().close();
+            } catch (Throwable e) {
+              LOG.warn("Error closing dataset '{}'", notification.getKey());
```
chtyim: Good to log the value and the exception as well.

chtyim: or at least log the value class name.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
```diff
@@ -0,0 +1,206 @@
...
+    if (staticDatasets != null) {
+      for (Map.Entry<String, Map<String, String>> entry : staticDatasets.entrySet()) {
+        this.staticDatasets.put(new DatasetCacheKey(entry.getKey(), entry.getValue()),
+                                getDataset(entry.getKey(), entry.getValue()));
+      }
+    }
+  }
+
+  @Override
+  public synchronized <T extends Dataset> T getDataset(DatasetCacheKey key, boolean bypass)
```
chtyim: Since this class is for single thread usage, we shouldn't need `synchronized`.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/MultiThreadDatasetCache.java
anew: yes

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/MultiThreadDatasetCache.java
anew: agreed

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/MultiThreadDatasetCache.java
anew: no need. I think this is a leftover from the old DynamicDatasetCache (the grandfather of this class) and that had synchronized it.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
anew: right., will do.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
```diff
@@ -0,0 +1,206 @@
...
+    throws DatasetInstantiationException {
+
+    Dataset dataset;
+    try {
+      if (bypass) {
+        dataset = datasetLoader.load(key);
+      } else {
+        try {
+          dataset = datasetCache.get(key);
```
chtyim: Should we always check the `staticDatasets` map first and return the one from there if found?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
anew: yeah same as above. removed.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
anew: It would be the same one. In the constructor, we add all static datasets to the cache, so it is guaranteed to be in the cache. Or do think it is faster to get it from the static datasets?

chtyim:
Did a pass, mostly look good, just couple comments.

anew:
@chtyim I addressed your comments, please take another look. 
Build is here: http://builds.cask.co/browse/CDAP-DUT2955-18

cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/SingleThreadDatasetCache.java
chtyim: Got it. Didn't notice that it was added to the cache in the constructor.

chtyim:
LGTM

anew:
@chtyim I have fixed the threading issue, and added a check that the per-thread cleanup works properly. Please take another look.

anew:
New build is here: http://builds.cask.co/browse/CDAP-DUT2955-19

chtyim:
LGTM

PR #4310, opened 2015-10-15
Improve SDK packaging

awholegunch:
We have been including in the SDK's ``examples`` directories files that are not required. For example, in 3.2.0 SDK's *Purchase* example *target* directory:
```
Purchase-3.2.0-javadoc.jar
Purchase-3.2.0-sources.jar
Purchase-3.2.0.jar
apidocs
stage-packaging
stage-packaging-deb
```
Only the first three JARs are required; ``apidocs`` just has GIFs, and the ``stage-packaging`` directories has copies of the ``bin`` scripts included higher up.

This PR fixes that by excluding these directories, and reduces the file size of the ZIP...


chtyim:
LGTM

PR #4313, opened 2015-10-16
CDAP-3833 fix snapshot file sink so that its always readable.

albertshau:
The previous snapshot file sink had behavior where it could not
be read while a write is taking place.
This made it impossible to really know when it was safe to read
it. This was done because the requirement was to always write
to the same hdfs path. This write pattern doesn't make much sense,
so we are changing the behavior so that each write will write
a new snapshot, and the dataset keeps track of which snapshot
is the latest snapshot. Readers are then directed to the actual
files for the latest snapshot instead of reading from a constant
location. The implementation uses a PartitionedFileSet that has
a single time partition, and an additional file at the base
location that stores the most recent partition. Updates to the
file and reads from the file are controlled using file locks
to avoid race conditions.

Also adding a corresponding snapshot source to read from a
snapshot sink.


albertshau:
http://builds.cask.co/browse/CDAP-RBT509

sreevatsanraman:
We need to add the rst files for the SNAPSHOT source. 

The code LGTM.

sreevatsanraman:
LGTM

sreevatsanraman:
We need to open a new one for Parquet sources.

awholegunch:
Is there an issue for this? Can we add a release note to it?

PR #4314, opened 2015-10-16
Fixes polling latest runs in hydrator detailed view for metrics, logs and runs count

ajainarayanan:
- Reduces the poll interval for polling latest runs from 10 seconds to 2 seconds.
- Modifies request to poll for logs and metrics based on latest run and program type.


astewes:
:+1: LGTM

PR #4316, opened 2015-10-16
CDAP-3902: Remove remote call for registering dataset usage in DatasetInstanceService

jwang47:
- Added DatasetProvider and DatasetMetaProvider
- Made UsageRegistry obtain the usage registry dataset instance by using
  DatasetProvider instead of DatasetFramework

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetMetaProvider.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetMetaProvider.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.cdap.data2.datafabric.dataset;
+
+import co.cask.cdap.api.dataset.DatasetSpecification;
+import co.cask.cdap.proto.DatasetTypeMeta;
+import co.cask.cdap.proto.Id;
+
+/**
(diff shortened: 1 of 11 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetMetaProvider.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.cdap.data2.datafabric.dataset;
+
+import co.cask.cdap.api.dataset.DatasetSpecification;
+import co.cask.cdap.proto.DatasetMeta;
```
rohitsinha54: Unused import

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetProvider.java
```diff
@@ -0,0 +1,151 @@
...
+        DatasetModule module = DatasetModules.getDatasetModule(moduleClass);
+        module.register(registry);
+      } catch (Exception e) {
+        LOG.error("Was not able to load dataset module class {} while trying to load type {}",
+                  moduleMeta.getClassName(), implementationInfo, e);
+        throw Throwables.propagate(e);
+      }
+    }
+
+    // contract of DatasetTypeMeta is that the last module returned by getModules() is the one
```
rohitsinha54: Is this true for custom dataset which embed other custom dataset ? Will the dataset type will appear after the embedded one ? or the embedded one  afterwards ?

jwang47: Not sure, I just moved this code from somewhere else.

cdap-master/src/main/java/SimpleTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+/**
+ *
+ */
+public class SimpleTest {
```
poornachandra: This class is needs to be removed.

jwang47:
Reopened https://github.com/caskdata/cdap/pull/4320

PR #4318, opened 2015-10-16
Add more prints in ElementIdTest

jwang47:


bdmogal:
@jwang47 we shouldn't need this PR anymore, right, given that we're adding verifications for ``toString()`` in #4361. Shouldn't need any tests that are ignored now?

PR #4319, opened 2015-10-16
fixes issues with node label dag display

astewes:
* Adds character limit to display of node label name in DAG
*  Removes overflow-x scrolling of node label name

label


ajainarayanan:
:+1: LGTM

PR #4320, opened 2015-10-16
CDAP-3902: Remove unnecessary remote calls when getting a dataset instance

jwang47:
- Added DatasetProvider and DatasetMetaProvider
- Made UsageRegistry obtain the usage registry dataset instance by using
  DatasetProvider instead of DatasetFramework
- Removed namespace existence check in DatasetInstanceService.get()

https://issues.cask.co/browse/CDAP-3902
http://builds.cask.co/browse/CDAP-DUT2980

# Performance test
* Distributed CDAP cluster with 3 slaves
* N threads concurrently doing a single get dataset operation with 1 owner
  * Before changes, the get dataset RPC should cause the dataset service to make 2 extra RPCs: one for getting the usage registry dataset and another for checking namespace existence
  * After changes, the get dataset RPC should cause the dataset service to make 0 extra RPCs

cdap-site.xml:
```
<property>
  <name>dataset.service.exec.threads</name>
  <value>300</value>
</property>
<property>
  <name>dataset.service.worker.threads</name>
  <value>30</value>
</property>
<property>
  <name>app.exec.threads</name>
  <value>20</value>
</property>
```

## Results
Format: `N=<# of threads>: <total time for run 1>, <total time for run 2>, ..`

### After changes (total time taken)
N=300: 3632, 3510, 3312, 3343, 3912
N=500: 4666, 4486, 4942, 4786, 4976
N=1000: 6995, 6712, 6523, 6379, 6492
N=2000: 11092, 10241 (w/ failures)

### Before changes (total time taken)
N=300: 6902, 5285, 4778, 4576, 4826
N=500: 17425 (was able to send ~20 requests with ~2000ms response time, but most requests failed), 17049 (same as first try)

### DIFF
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetMetaProvider.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetMetaProvider.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.cdap.data2.datafabric.dataset;
+
+import co.cask.cdap.proto.DatasetMeta;
+import co.cask.cdap.proto.Id;
+
+/**
+ * Provides metadata about datasets for {@link LocalDatasetProvider}.
+ */
+public interface DatasetMetaProvider {
+  DatasetMeta getMeta(Id.DatasetInstance instance) throws Exception;
+}
--- a/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/LocalDatasetProvider.java
+++ b/cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/LocalDatasetProvider.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.cdap.data2.datafabric.dataset;
+
+import co.cask.cdap.api.dataset.Dataset;
+import co.cask.cdap.api.dataset.DatasetContext;
+import co.cask.cdap.api.dataset.DatasetSpecification;
+import co.cask.cdap.api.dataset.module.DatasetDefinitionRegistry;
+import co.cask.cdap.api.dataset.module.DatasetModule;
+import co.cask.cdap.common.lang.ClassLoaders;
+import co.cask.cdap.data2.datafabric.dataset.service.DatasetInstanceService;
+import co.cask.cdap.data2.datafabric.dataset.type.ConstantClassLoaderProvider;
+import co.cask.cdap.data2.datafabric.dataset.type.DatasetClassLoaderProvider;
+import co.cask.cdap.data2.dataset2.DatasetDefinitionRegistryFactory;
+import co.cask.cdap.data2.dataset2.module.lib.DatasetModules;
+import co.cask.cdap.proto.DatasetMeta;
+import co.cask.cdap.proto.DatasetModuleMeta;
+import co.cask.cdap.proto.DatasetTypeMeta;
+import co.cask.cdap.proto.Id;
+import com.google.common.base.Objects;
+import com.google.common.base.Throwables;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Provides {@link Dataset} instances using {@link DatasetMetaProvider}.
+ * Used by {@link RemoteDatasetFramework} and {@link DatasetInstanceService}.
+ * Use this when you want to control how dataset instances are created. For example, when
+ * you want to obtain a {@link Dataset} instance without having to make remote calls.
+ */
+public class LocalDatasetProvider implements DatasetProvider {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LocalDatasetProvider.class);
+  private final DatasetDefinitionRegistryFactory registryFactory;
+  private final DatasetMetaProvider metaProvider;
+
+  private LocalDatasetProvider(DatasetDefinitionRegistryFactory registryFactory, DatasetMetaProvider metaProvider) {
+    this.registryFactory = registryFactory;
+    this.metaProvider = metaProvider;
+  }
+
+  public LocalDatasetProvider(DatasetDefinitionRegistryFactory registryFactory, RemoteDatasetFramework metaProvider) {
+    this(registryFactory, (DatasetMetaProvider) metaProvider);
+  }
+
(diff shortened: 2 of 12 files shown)


cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/FrameworkDatasetProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import co.cask.cdap.proto.Id;
+import com.google.inject.Inject;
+
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Provides {@link Dataset} instances using {@link DatasetFramework}.
+ */
+@SuppressWarnings("unchecked")
```
poornachandra: It is a good idea to move `SuppressWarnings("unchecked")` to the line where it is needed.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/DatasetMetaProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.cdap.data2.datafabric.dataset;
+
+import co.cask.cdap.proto.DatasetMeta;
+import co.cask.cdap.proto.Id;
+
+/**
+ * Provides metadata about datasets for {@link FrameworkDatasetProvider}.
```
poornachandra: `Provides metadata about datasets for {@link DatasetProvider}`

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/LocalDatasetProvider.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Provides {@link Dataset} instances directly using {@link DatasetMetaProvider}.
+ */
+@SuppressWarnings("unchecked")
```
poornachandra: Same here.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/FrameworkDatasetProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import co.cask.cdap.api.dataset.Dataset;
+import co.cask.cdap.data2.dataset2.DatasetFramework;
+import co.cask.cdap.proto.Id;
+import com.google.inject.Inject;
+
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Provides {@link Dataset} instances using {@link DatasetFramework}.
```
poornachandra: Can you add javadocs on when to use this class?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/LocalDatasetProvider.java
```diff
@@ -0,0 +1,150 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Provides {@link Dataset} instances directly using {@link DatasetMetaProvider}.
```
poornachandra: Can you add javadocs on when to use this class?

cdap-data-fabric/src/test/java/co/cask/cdap/data2/datafabric/dataset/service/DatasetInstanceHandlerTest.java
```diff
@@ -288,8 +288,9 @@ public void testNotFound() throws IOException {
     HttpResponse response = makeInstancesRequest(nonExistent.getId());
     assertNamespaceNotFound(response, nonExistent);
 
-    response = getInstance(datasetInstance);
-    assertNamespaceNotFound(response, nonExistent);
+    // TODO: commented out for now until we add back namespace checks on get dataset
```
poornachandra: Add a JIRA for this

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/FrameworkDatasetProvider.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+@SuppressWarnings("unchecked")
+public class FrameworkDatasetProvider implements DatasetProvider {
```
poornachandra: Where is `FrameworkDatasetProvider` used?

poornachandra:
@jwang47 I did a pass through the PR.

The names `FrameworkDatasetProvider` and `LocalDatasetProvider` are a bit confusing. Please address the comments, I'll do another pass after that.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/FrameworkDatasetProvider.java
jwang47: This is the default implementation of DatasetProvider, used when `UsageRegistry` is constructed via injection.

You can alternatively manually construct a `UsageRegistry` with the `LocalDatasetProvider`, where you need to pass in an implementation of `DatasetMetaProvider`. This alternative is used by `DatasetInstanceService` when getting a dataset instance.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/LocalDatasetProvider.java
```diff
@@ -0,0 +1,154 @@
...
+ * Use this when you want to control how dataset instances are created. For example, when
+ * you want to obtain a {@link Dataset} instance without having to make remote calls.
+ */
+public class LocalDatasetProvider implements DatasetProvider {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LocalDatasetProvider.class);
+  private final DatasetDefinitionRegistryFactory registryFactory;
+  private final DatasetMetaProvider metaProvider;
+
+  public LocalDatasetProvider(DatasetDefinitionRegistryFactory registryFactory, DatasetMetaProvider metaProvider) {
```
poornachandra: Since `LocalDatasetProvider` can only be used by `RemoteDatasetFramework` and `DatasetInstanceService`, let's make it explicit in the constructor by having only the following two constructors - 
```
public LocalDatasetProvider(DatasetDefinitionRegistryFactory registryFactory, RemoteDatasetFramework remoteDatasetFramework)

public LocalDatasetProvider(DatasetDefinitionRegistryFactory registryFactory, DatasetInstanceService datasetInstanceService)

```

cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/UsageRegistry.java
```diff
@@ -52,14 +54,22 @@
...
       @Override
       public UsageDatasetIterable get() {
         try {
-          Object usageDataset = DatasetsUtil.getOrCreateDataset(datasetFramework, USAGE_INSTANCE_ID,
-                                                                UsageDataset.class.getSimpleName(),
-                                                                DatasetProperties.EMPTY, null, null);
+          Object usageDataset = provider.get(USAGE_INSTANCE_ID, null, null);
+          if (usageDataset == null) {
+            // use DatasetFramework only to create
+            DatasetsUtil.createIfNotExists(framework, USAGE_INSTANCE_ID,
```
poornachandra: Won't this create call still be a remote call from `DatasetInstanceService`?

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/FrameworkDatasetProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import co.cask.cdap.proto.Id;
+import com.google.inject.Inject;
+
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Provides {@link Dataset} instances using {@link DatasetFramework}. Default implementation.
+ */
+public class FrameworkDatasetProvider implements DatasetProvider {
```
poornachandra: Since this class has to be used almost everywhere for getting Dataset instances, I think it would better to call it `DefaultDatasetProvider`. What do you say?

poornachandra:
@jwang47 Final few comments, please address them.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/registry/UsageRegistry.java
jwang47: Only when the usage registry dataset doesn't exist yet, but otherwise the first provider.get() should not return null.

cdap-data-fabric/src/main/java/co/cask/cdap/data2/datafabric/dataset/FrameworkDatasetProvider.java
jwang47: Sounds good.

PR #4321, opened 2015-10-16
CDAP-4021 refactor batch status call

albertshau:
Refactors the batch status and instances calls to move their
request and response objects into cdap-proto so that they can
be reused by the ProgramClient. Also adding a method to
ProgramClient to perform the batch status call.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2981

cdap-proto/src/main/java/co/cask/cdap/proto/BatchProgram.java
```diff
@@ -0,0 +1,67 @@
...
+
+package co.cask.cdap.proto;
+
+import com.google.common.base.Joiner;
+
+/**
+ * Array components of the batch status request to POST /namespaces/{namespace}/status.
+ */
+public class BatchProgram {
+  protected final String appId;
```
anwar6953: why not use `Id.Program` class?

anwar6953: oh, because of json serialization, I guess.
And the new,replacement classes that Alvin is working on aren't available yet?

albertshau: yep for serialization

anwar6953:
LGTM

PR #4326, opened 2015-10-19
CDAP-4032: Support dataset lookup from TransformContext

jwang47:
Changes
* Make TransformContext extend DatasetContext
  * Modify BatchTransformContext, RealtimeTransformContext, other internal implementations
* Wrap Transform initialization and record transformation in transactions. This will add extra overhead to transforms, so maybe we want some way for the Transform implementation to specify if it needs to be wrapped in a transaction.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/etl/realtime/ETLWorker.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/etl/realtime/ETLWorker.java
@@ -192,7 +192,7 @@ public class ETLWorker extends AbstractWorker {
   @SuppressWarnings("unchecked")
   private void initializeSinks(WorkerContext context) throws Exception {
     List<SinkInfo> sinkInfos = GSON.fromJson(context.getSpecification().getProperty(Constants.Sink.PLUGINIDS),
-                                                 SINK_INFO_TYPE);
+                                             SINK_INFO_TYPE);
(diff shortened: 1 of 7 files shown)


cdap-app-templates/cdap-etl/cdap-etl-realtime/src/main/java/co/cask/cdap/etl/realtime/ETLWorker.java
```diff
@@ -267,33 +273,38 @@ public void run(DatasetContext context) throws Exception {
...
 
-      // For each object emitted by the source, invoke the transformExecutor and collect all the data
-      // to be persisted in the sink.
-      for (Object sourceData : sourceEmitter) {
-        try {
-          TransformResponse transformResponse = transformExecutor.runOneIteration(sourceData);
-          while (transformResponse.getEmittedRecords().hasNext()) {
-            dataToSink.add(transformResponse.getEmittedRecords().next());
-          }
+      getContext().execute(new TxRunnable() {
```
jwang47: We may want some way to not have to wrap every transform operation in a transaction, but not sure.

chtyim:
It is not simple as that in real time case. The real time Worker, I believe the execution of transform functions are not within transaction, meaning providing access to `DatasetContext` will result in error because non of the operations on the datasets returned by the `DatasetContext` will be executed in transaction.

chtyim:
I think we shouldn't expose `Dataset` directly to transform, but rather having a `Lookup` interface through the ETL app to provide read only lookup that can be backed by one or more dataset(s).

PR #4327, opened 2015-10-19
CDAP-3832 persist the offset only after the metrics are processed (Porting)

gsps1:
porting : https://github.com/caskdata/cdap/pull/4186 to 3.1


chtyim:
LGTM

PR #4328, opened 2015-10-19
CDAP-3999 improve error logging in Script transforms.

anwar6953:
When script fails to run the script on an input, it is helpful to specify what the input record was that it failed on.

Also, it may not necessarily be an invalid filter condition that caused the exception (ScriptFilterTransform.java).

https://issues.cask.co/browse/CDAP-3999
http://builds.cask.co/browse/CDAP-RBT512-1

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptFilterTransform.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptFilterTransform.java
@@ -102,7 +102,7 @@ public class ScriptFilterTransform extends Transform<StructuredRecord, Structure
         metrics.count("filtered", 1);
       }
     } catch (Exception e) {
-      throw new IllegalArgumentException("Invalid filter condition.", e);
+      throw new IllegalArgumentException("Could not transform input: " + input, e);
     }
   }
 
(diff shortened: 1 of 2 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptFilterTransform.java
```diff
@@ -102,7 +102,7 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
         metrics.count("filtered", 1);
       }
     } catch (Exception e) {
-      throw new IllegalArgumentException("Invalid filter condition.", e);
+      throw new IllegalArgumentException("Could not transform input: " + input, e);
```
chtyim: We might need to be careful as logging with user data might not be warned if it contains sensitive data.

awholegunch: Is it possible to have a setting (user preference) such that only if a script is run with that setting does the input show?

chtyim: I think in general, the framework shouldn't log user data as it can easily become a security breach. For example, if I configure the script transform wrongly then suddenly all user sensitive data that is supposed to be readable by me is now readable by whoever can read system log (most likely operation people).

anwar6953:
Closing this PR then, since we don't want to log user data.

PR #4329, opened 2015-10-19
CDAP-4030 make groupid required for etl plugin archetype

albertshau:
Making the groupId property required for the etl plugin archetypes.
This is because the defaults will generate a project where the
plugin is in the co.cask.cdap package, which is not visible when
CDAP inspects the artifact for plugins. So instead of having a
bad default, force the user to set the groupId, which will usually
be their organization.

Also changing docs to reflect this change. We could have also set
a default value for groupId, but when you do that the property
cannot be set in interactive mode of archetype generation.


bdmogal:
:+1:

PR #4330, opened 2015-10-19
fixes boostrap columns in node config tab

astewes:
*  Adds correct grid class names to template file

Source:
source

Transform:
transform

Sink:
screen shot 2015-10-19 at 1 47 30 pm


ajainarayanan:
:+1: LGTM

PR #4332, opened 2015-10-19
CDAP-3982 add method to RuntimeContext to get namespace

albertshau:
this makes it possible for a program to get its namespace at
runtime. This is useful for a variety of use cases. For example,
you could have a worker that makes program lifecycle calls,
which need namespace in the url. Or you could call some program
that makes an external call to some http endpoint, and needs the
namespace to uniquely identify itself.


albertshau:
http://builds.cask.co/browse/CDAP-DUT2990

chtyim:
LGTM.

PR #4333, opened 2015-10-19
corrects behavior of bottom panel in pipeline detail view

astewes:
*  Changes Node Config tab message when pipeline is already published
*  Fixes Schedule tab dropdown cursor behavior on hover 

detail


ajainarayanan:
:+1: LGTM

PR #4335, opened 2015-10-19
(CDAP-4026) Made MapReduceTaskContext available to plugins.

bdmogal:
Jira: [CDAP-4026](https://issues.cask.co/browse/CDAP-4026)
Build: http://builds.cask.co/browse/CDAP-RBT511-1

Usage example: 
```
class MySource extends BatchSource {
@Override
  public void initialize(BatchRuntimeContext context) throws Exception {
    super.initialize(context);
    // You can also cache this if its required in the transform method
    MapReduceTaskContext mrTaskContext = (MapReduceTaskContext) context.getOriginalContext();
    Path[] cacheFiles = mapperContext.getLocalCacheFiles();
    // work on files from dist cache....
    Path filePath = ((CombineFileSplit) mapperContext.getInputSplit()).getPath(0);
    // make decisions based on InputSplit
    // update MR counters 
    mapperContext.getCounter("group", "counter").increment(1);
}
```

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/batch/BatchRuntimeContext.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/batch/BatchRuntimeContext.java
@@ -47,6 +47,13 @@ public interface BatchRuntimeContext<T> extends DatasetContext, TransformContext
    */
   Map<String, String> getRuntimeArguments();
 
-  T getOriginalContext();
+  /**
+   * Returns the processing framework's context, to enable advanced use-cases in plugins.
+   * E.g. this method could give access to the {@link MapReduceTaskContext} for pipelines that use
+   * MapReduce as the processing framework.
+   *
+   * @return the processing framework context
+   */
 
+  T getOriginalContext();
 }
--- a/cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/batch/BatchRuntimeContext.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/batch/BatchRuntimeContext.java
@@ -18,15 +18,18 @@ package co.cask.cdap.etl.api.batch;
 
 import co.cask.cdap.api.annotation.Beta;
 import co.cask.cdap.api.data.DatasetContext;
+import co.cask.cdap.api.mapreduce.MapReduceTaskContext;
 import co.cask.cdap.etl.api.TransformContext;
 
 import java.util.Map;
 
 /**
  * Context passed to Batch Source and Sink.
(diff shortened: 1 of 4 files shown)


cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/batch/BatchRuntimeContext.java
```diff
@@ -44,4 +47,6 @@
    */
   Map<String, String> getRuntimeArguments();
 
+  T getOriginalContext();
```
sagarkapare: Missing Javadoc

cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/co/cask/cdap/etl/api/batch/BatchRuntimeContext.java
```diff
@@ -18,15 +18,18 @@
...
 import java.util.Map;
 
 /**
  * Context passed to Batch Source and Sink.
+ *
+ * @param <T>  the context used to construct the BatchRuntimeContext, eg. {@link MapReduceTaskContext}
  */
 @Beta
-public interface BatchRuntimeContext extends DatasetContext, TransformContext {
+public interface BatchRuntimeContext<T> extends DatasetContext, TransformContext {
```
chtyim: How does the user know what the type `T` is in the `BatchSource/Sink` `initialize` method?

bdmogal: good point. any suggestion? 

The only way that I can think of, is if the initialize method takes a MapReduceRuntimeContext. One way of doing that is making the BatchRuntimeContext parameter to StageLifecycle generic. However, that would mean all current plugins become mapreduce specific (because BatchSource will take the subclass of BatchRuntimeContext as a generic parameter). I wanted to keep it non-MapReduce specific, so used generics here instead. But yes, this does mean that its not very straightforward to Plugin developers.

any better ideas?

chtyim: One way is not to have `<T>` as a class parameter, but just to have it for the `getOriginalContext` method (BTW, that method name is not so great as well).

chtyim: However, from the plugin developer point of view, it is still confusing as there is no way for him to figure out what the actual type is while writing the code. It also requires the developer to check for the type in runtime, cast it and call different method based on different context type, which is arguably hard to use.

bdmogal:
Requires some more thinking after discussion with @chtyim. Will re-open against develop.

PR #4342, opened 2015-10-20
Remove commons-logging dependency

sreevatsanraman:
Commons-logging should not be included in the dependency of CDAP. it should come from the hadoop dependency which will be provided in the cluster.

Verified this change by installing CDAP on a cluster and running programs


chtyim:
LGTM

PR #4346, opened 2015-10-20
rename template to pass RAT check

dereklwood:
renaming json file to *.json so that it is excluded from RAT check


bdmogal:
:+1:

PR #4347, opened 2015-10-20
Show error message (if any) while running once in relatime pipeline or starting a schedule in batch pipeline

ajainarayanan:
Surfacing errors (if any) in published pipeline. This is an intermittent work around (rather than not showing anything). In 3.3 we need to design a way to show error/info/success messages in a different way.

pipelineerror

### DIFF
--- a/cdap-ui/app/features/adapters/controllers/detail-ctrl.js
+++ b/cdap-ui/app/features/adapters/controllers/detail-ctrl.js
@@ -15,7 +15,7 @@
  */
 
 angular.module(PKG.name + '.feature.adapters')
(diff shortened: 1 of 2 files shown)


cdap-ui/app/features/adapters/controllers/detail-ctrl.js
```diff
@@ -196,17 +200,16 @@ angular.module(PKG.name + '.feature.adapters')
...
           break;
         case 'Run Once':
+          $scope.runOnceLoading = true;
           myWorkFlowApi.doAction(angular.extend(params, { action: 'start' }), {})
             .$promise
             .then(function () {
-              $scope.runOnceLoading = true;
               $scope.appStatus = 'RUNNING';
-
-              $timeout(function () {
```
anwar6953: why remove this part?
Now `runOnceLoading` will never go to false?

anwar6953:
LGTM

PR #4349, opened 2015-10-20
Clean up messages and strings. 

awholegunch:
Removes a confusing example, makes case consistent, and other cleanup.


ajainarayanan:
:+1: LGTM on UI changes.

cdap-ui/templates/cdap-etl-realtime/Stream.json
```diff
@@ -22,8 +22,8 @@
         },
         "name": {
           "widget": "stream-selector",
-          "label" : "Name",
-          "description" : "Name of the stream to which we write to. If the stream is present, then data will be written to it. If it's not present then one would be created.",
+          "label" : "Stream Name",
```
bdmogal: When users see this, they're already configuring a Stream source. Do we need to mention Stream again?

bdmogal:
Quick Question. LGTM otherwise.

cdap-ui/templates/cdap-etl-realtime/Stream.json
awholegunch: I think so. There are three fields there, all ending in "Name". For some reason, the "Stream Name" is last. It is also consistent with others.

PR #4350, opened 2015-10-20
Copy ServiceMain class from twill

sreevatsanraman:
Copying the ServiceMain class from Twill for CDAP 3.2.1 release. This should be removed after we upgrade to twill 0.7.0
https://issues.cask.co/browse/CDAP-4052


poornachandra:
LGTM

PR #4361, opened 2015-10-21
CDAP-3657: Minor cleanup for EntityId

jwang47:
http://builds.cask.co/browse/CDAP-DUT2988


cdap-proto/src/test/java/co/cask/cdap/proto/id/EntityIdTest.java
```diff
@@ -115,8 +127,16 @@ public void testInvalidId() {
   @Test
   @Ignore
   public void testPrintToString() {
-    for (ElementId id : ids) {
-      System.out.println(id.toString());
+    for (EntityId id : ids) {
+      System.out.println(id.getEntity() + ": " + id.toString());
```
bdmogal: what are these two tests for? and why are they ``@Ignore``d?

jwang47: These are only for seeing what toString() and GSON.toJson() gives.

bdmogal: seeing meaning manually? we should verify the output, no? You can set the ``PrintStream``s in ``System`` like [so](http://stackoverflow.com/a/1119559) and verify their contents?

jwang47: I suppose it does make sense to verify the toString() output, will add a test case.

bdmogal: Right, my bad, you don't actually need to test the stdout here :)

cdap-proto/src/test/java/co/cask/cdap/proto/id/EntityIdTest.java
```diff
@@ -113,10 +125,36 @@ public void testInvalidId() {
...
+                        Ids.namespace("foo").app("app").flow("flo").run("run1").toString());
+    Assert.assertEquals("flowlet:foo.app.flo.flol",
+                        Ids.namespace("foo").app("app").flow("flo").flowlet("flol").toString());
+    Assert.assertEquals("flowlet_queue:foo.app.flo.flol.q",
+                        Ids.namespace("foo").app("app").flow("flo").flowlet("flol").queue("q").toString());
+  }
+
+  @Test
   @Ignore
   public void testPrintToString() {
```
bdmogal: shouldn't need this test now, right?

cdap-proto/src/test/java/co/cask/cdap/proto/id/EntityIdTest.java
```diff
@@ -113,10 +125,36 @@ public void testInvalidId() {
...
-    for (ElementId id : ids) {
-      System.out.println(id.toString());
+    for (EntityId id : ids) {
+      System.out.println(id.getEntity() + ": " + id.toString());
+    }
+  }
+
+  @Test
+  @Ignore
+  public void testPrintToJson() {
```
bdmogal: can we add a test for ``toJson`` as well instead of this?

cdap-proto/src/main/java/co/cask/cdap/proto/Id.java
```diff
@@ -952,8 +1018,8 @@ public String getId() {
       return String.format("%s-%s", name, version.getVersion());
     }
 
-    public ArtifactId toArtifactId() {
-      return new ArtifactId(name, version,
+    public co.cask.cdap.api.artifact.ArtifactId toArtifactId() {
+      return new co.cask.cdap.api.artifact.ArtifactId(name, version,
                             Namespace.SYSTEM.equals(namespace) ? ArtifactScope.SYSTEM : ArtifactScope.USER);
```
bdmogal: indentation

cdap-proto/src/test/java/co/cask/cdap/proto/id/EntityIdTest.java
jwang47: This isn't a test, it's just a convenient way to see what the toString() gives.

cdap-proto/src/main/java/co/cask/cdap/proto/id/ArtifactId.java
```diff
@@ -25,14 +25,14 @@
 /**
  * Uniquely identifies an artifact.
  */
-// TODO: handle duplication with ArtifactId
-public class NamespacedArtifactId extends ElementId implements NamespacedId, ParentedId<NamespaceId> {
+// TODO: handle duplication with ArtifactId in cdap-api
+public class ArtifactId extends EntityId implements NamespacedId, ParentedId<NamespaceId> {
```
bdmogal: I think it is very inconvenient to have two classes called ``ArtifactId``. The other one is in cdap-api. Should that be called ``Artifact`` instead? Since it is in cdap-api we may have to deprecate it first though, even though it is marked as ``@Beta``. @albertshau: thoughts?

cdap-proto/src/main/java/co/cask/cdap/proto/id/EntityIdCompatible.java
```diff
@@ -19,8 +19,8 @@
 import co.cask.cdap.proto.Id;
 
 /**
- * Compatibility from {@link Id} to {@link ElementId}.
+ * Compatibility from {@link Id} to {@link EntityId}.
  */
-public interface ElementIdCompatible {
-  ElementId toElementId();
+public interface EntityIdCompatible {
+  EntityId toEntityId();
```
bdmogal: need Javadocs for interface method

cdap-proto/src/main/java/co/cask/cdap/proto/id/ParentedId.java
```diff
@@ -17,10 +17,10 @@
...
 
 /**
- * An {@link ElementId} which has a parent {@link ElementId}.
+ * An {@link EntityId} which has a parent {@link EntityId}.
  *
  * @param <T> the type of parent
  */
-public interface ParentedId<T extends ElementId> {
+public interface ParentedId<T extends EntityId> {
   T getParent();
```
bdmogal: Javadocs

cdap-proto/src/main/java/co/cask/cdap/proto/id/IdCompatible.java
```diff
@@ -19,7 +19,7 @@
 import co.cask.cdap.proto.Id;
 
 /**
- * Compatibility from {@link ElementId} to {@link Id}.
+ * Compatibility from {@link EntityId} to {@link Id}.
  */
 public interface IdCompatible {
   Id toId();
```
bdmogal: javadocs

cdap-proto/src/test/java/co/cask/cdap/proto/id/EntityIdTest.java
bdmogal: But then we shouldn't check it in as a ``@Test``, right? Why not just keep it as a comment on the ``toString()``?

cdap-proto/src/test/java/co/cask/cdap/proto/id/EntityIdTest.java
jwang47: Added `testCompatibility`

cdap-proto/src/test/java/co/cask/cdap/proto/id/EntityIdTest.java
jwang47: I don't want to maintain a comment on `toString()`, and I'd like a place where I can see all the `EntityId.toString()`s together. What do you suggest?

cdap-proto/src/main/java/co/cask/cdap/proto/id/ArtifactId.java
jwang47: Also, `ArtifactId` needs to be in cdap-api it looks like. I'll need to do some rework on `ArtifactId`.

jwang47: Renamed back to `NamespacedArtifactId` for now.

bdmogal:
LGTM. can you squash commits before merging?

PR #4362, opened 2015-10-21
Move RecordConverter to cdap-etl-common to make it available for plugins to use

gokulavasan:
Build: http://builds.cask.co/browse/CDAP-DUT3005-1

Moving RecordConverter abstract class to etl-common so that we can use it in hydrator-plugins repo. This is necessary to address comments in this PR: https://github.com/caskdata/hydrator-plugins/pull/28


gokulavasan:
@albertshau Please review

albertshau:
lgtm

gokulavasan:
Build passed. Merging.

PR #4363, opened 2015-10-21
adding comments to call out shared init functions with the CSD

dereklwood:
When the CSD starts up CDAP processes, the CSD control script (found [here](https://github.com/caskdata/cm_csd/blob/develop/src/scripts/cdap-control.sh)) sources the common init scripts and invokes some its functions.  As we plan to refactor these init scripts this can be dangerous to overlook.

This PR just adds comments to the required function headers pointing this out, to help avoid potential CSD breakage


awholegunch:
LGreatTM.

PR #4364, opened 2015-10-21
Make sure that stop is called for HBaseTestBase, ZkClientService and …

bdmogal:
…ZkServer even if there are previous exceptions. Stop was not getting called previously if either tableUtil or testHBase was null in the deleteNamespace method. This fix does not silently mask the NPE, but only makes sure that stop is called irrespective of it. Will need to debug the NPE separately.

Most recent failure for this test: https://builds.cask.co/download/CDAP-DUT3003-JOB1/build_logs/CDAP-DUT3003-JOB1-2.log

Build: http://builds.cask.co/browse/CDAP-DUT3007-1

### DIFF
--- a/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/dataset2/lib/kv/HBaseKVTableTest.java
+++ b/cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/dataset2/lib/kv/HBaseKVTableTest.java
@@ -40,23 +40,21 @@ import org.junit.rules.TemporaryFolder;
 @Category(SlowTests.class)
 public class HBaseKVTableTest extends NoTxKeyValueTableTest {
   @ClassRule
-  public static TemporaryFolder tmpFolder = new TemporaryFolder();
+  public static final TemporaryFolder TMP_FOLDER = new TemporaryFolder();
+  @ClassRule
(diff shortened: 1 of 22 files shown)


chtyim:
We should probably turn those into `@ClassRule` to have junit to handle all cases.

bdmogal:
@chtyim do you mean turn HBaseTestBase classes into ``@ClassRule``?

cdap-data-fabric-tests/src/test/java/co/cask/cdap/data2/dataset2/lib/kv/HBaseKVTableTest.java
```diff
@@ -41,14 +41,13 @@
 public class HBaseKVTableTest extends NoTxKeyValueTableTest {
   @ClassRule
   public static TemporaryFolder tmpFolder = new TemporaryFolder();
+  @ClassRule
+  public static HBaseTestBase testHBase = new HBaseTestFactory().get();
```
chtyim: This should be a static final.

chtyim:
Just one comment.

bdmogal:
@chtyim addressed comment. Build is :green_apple:

chtyim:
LGTM

bdmogal:
Squashed commits. Merging.

PR #4366, opened 2015-10-21
Fixes maven UI build to ignore dist and other un-necessary folders + fixes UI issues

ajainarayanan:
- Fixes maven UI build process, where the maven build would take forever to RAT check UI because the 'dist' folder generated as part of the build process still stays and we are not ignoring it.
- Fixes a UI es6ify issue (ng-sanitize for app status controller)


astewes:
:+1: LGTM

sreevatsanraman:
LGTM

PR #4367, opened 2015-10-21
Fix/ui outputschema null

ajainarayanan:
- Fixes output schema to be properly propagated. Recent change broke the assignment of 'schema' property of any plugin. This PR fixes that.
- Fixes output schema to be optional. The same plugin (for instance MongoDB) could have an output schema for source but not one for sink. This should not fail or error out in UI.

TODO:
 Right now we assume there is going to be only one output schema. When we refactor the plugin edit controller we will account to more than one output schema per plugin.


astewes:
:+1: LGTM

anwar6953:
LGTM

PR #4368, opened 2015-10-21
CDAP-4076: Remove usage of Transactional in Store

jwang47:
https://issues.cask.co/browse/CDAP-4076
http://builds.cask.co/browse/CDAP-DUT3010


anwar6953:
LGTM

PR #4370, opened 2015-10-22
(CDAP-4039) Added better defaults handling in FileBatchConfig. Also m…

bdmogal:
…ade FileBatchConfig parameters public so they can be accessed by derived class objects.

Build: http://builds.cask.co/browse/CDAP-DUT3003-5
Jira: [CDAP-4039](https://issues.cask.co/browse/CDAP-4039)

### DIFF
--- a/cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataTestBase.java
+++ b/cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataTestBase.java
@@ -166,6 +166,7 @@ public abstract class MetadataTestBase extends AppFabricTestBase {
       getProperties(app);
       Assert.fail("Expected not to be able to get properties from invalid entity: " + app);
     } catch (NotFoundException expected) {
+      // expected
     }
   }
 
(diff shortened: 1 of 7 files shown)


cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataTestBase.java
```diff
@@ -166,6 +166,8 @@ protected void getPropertiesFromInvalidEntity(Id.Application app) throws Excepti
       getProperties(app);
       Assert.fail("Expected not to be able to get properties from invalid entity: " + app);
     } catch (NotFoundException expected) {
+      // expected
+      // expected
```
sagarkapare: twice?

sagarkapare:
one minor comment, otherwise LGTM :+1:

cdap-app-fabric/src/test/java/co/cask/cdap/metadata/MetadataTestBase.java
awholegunch: It was really, really expected. :)

awholegunch:
I've made changes, and fixed the merge. Please review. Otherwise, LGTM.

awholegunch:
[Docs Quick Build](http://builds.cask.co/browse/CDAP-DOB22-2)

bdmogal:
Addressed comment, squashed commits. Merging.

PR #4371, opened 2015-10-22
alertOnValium: Fix/ui thwart mad alerts clicking

ajainarayanan:
- Adds a service that can be replaced in place of ```$alert``` to show subsequent alerts only when the previous one is closed.
- Modifies relevant references of ```$alert``` to ```myAlertOnValium``` .

alert


anwar6953:
lgtm

PR #4373, opened 2015-10-22
Feature/hdinsight hive

dereklwood:
enhancements to the hdinsight installer:

   - [x] specify the cdap version and explore.enabled in the script, and then do token substitution in the template.  This is so that all variables are in the released artifact.  Otherwise as soon as we update the git release branch they would take effect.
   - [x] run the cdap[init] recipe, which creates HDFS directories, and is required for hive
   - [x] disable hive for now as it is not working
   - [x] use ``{{TOKEN}}`` for token substitution, consistent with other scripts we have.


awholegunch:
LGTM

PR #4375, opened 2015-10-22
rename conf json

dereklwood:
renaming the config template file to ``cdap-conf.json`` as it is really the chef config and not just a template for cdap-site.xml.


awholegunch:
LGTM

PR #4378, opened 2015-10-22
disabled resizing in modal textareas

astewes:
*  Disables resizing in modal textreas. Modal textareas will scroll vertically when there is more text than rows initially displayed

textarea


ajainarayanan:
:+1: LGTM

PR #4380, opened 2015-10-22
CDAP-4022 add batch start and stop rest apis for programs

albertshau:
Adding restful apis to start and stop multiple programs in the
same request. Also adding corresponding methods to ProgramClient,
as well as a convenience method to wait for a batch of programs
to be in a specified state.

### DIFF
--- a/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
+++ b/cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
@@ -657,8 +657,7 @@ public class ProgramLifecycleHttpHandler extends AbstractAppFabricHttpHandler {
           new Function<ProgramController, BatchProgramResult>() {
             @Override
             public BatchProgramResult apply(ProgramController input) {
-              return new BatchProgramResult(
-                program, AppFabricServiceStatus.OK.getCode().getCode(), AppFabricServiceStatus.OK.getMessage());
+              return new BatchProgramResult(program, AppFabricServiceStatus.OK.getCode().getCode(), null);
             }
           });
         issuedStops.add(issuedStop);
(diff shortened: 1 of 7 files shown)


albertshau:
http://builds.cask.co/browse/CDAP-DUT3017

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -597,19 +602,149 @@ public void getStatuses(HttpRequest request, HttpResponder responder,
...
+    // information in that scenario
+    int i = 0;
+    for (ListenableFuture<BatchProgramResult> issuedStop : issuedStops) {
+      try {
+        output.add(issuedStop.get());
+      } catch (Throwable t) {
+        LOG.warn(t.getMessage(), t);
+        output.add(new BatchProgramResult(programs.get(i),
+                                          AppFabricServiceStatus.INTERNAL_ERROR.getCode().getCode(),
+                                          AppFabricServiceStatus.INTERNAL_ERROR.getMessage()));
```
anwar6953: for the last parameter to BatchProgramResult constructor, shouldn't we pass in t.getMessage(), instead of INTERNAL_ERROR.getMessage())?

cdap-app-fabric/src/main/java/co/cask/cdap/gateway/handlers/ProgramLifecycleHttpHandler.java
```diff
@@ -597,19 +602,149 @@ public void getStatuses(HttpRequest request, HttpResponder responder,
...
+    for (final BatchProgram program : programs) {
+      Id.Program programId = Id.Program.from(namespaceId, program.getAppId(),
+                                             program.getProgramType(), program.getProgramId());
+      try {
+        ListenableFuture<BatchProgramResult> issuedStop = Futures.transform(issueStop(programId, null),
+          new Function<ProgramController, BatchProgramResult>() {
+            @Override
+            public BatchProgramResult apply(ProgramController input) {
+              return new BatchProgramResult(
+                program, AppFabricServiceStatus.OK.getCode().getCode(), AppFabricServiceStatus.OK.getMessage());
```
anwar6953: we should pass null as the last parameter to the constructor, right?
Since the last parameter is the error string.

anwar6953:
LGTM

PR #4381, opened 2015-10-22
Fixes empty output schema to have appropriate message instead of showing both the table and the message

ajainarayanan:
outputschema

### DIFF
--- a/cdap-ui/app/features/hydrator/templates/create/popovers/plugin-edit-output-schema.html
+++ b/cdap-ui/app/features/hydrator/templates/create/popovers/plugin-edit-output-schema.html
@@ -37,12 +37,23 @@
   </h4>
 
   <fieldset class="clearfix" ng-disabled="isDisabled">
-    <my-schema-editor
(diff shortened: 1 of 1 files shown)


cdap-ui/app/features/hydrator/templates/create/popovers/plugin-edit-output-schema.html
```diff
@@ -37,12 +37,23 @@
...
-    <my-schema-editor
-      ng-model="plugin.outputSchema"
-      data-disabled="plugin.implicitSchema || isDisabled"
-      plugin-properties="plugin.properties"
-      config="PluginEditController.schemaProperties">
-    </my-schema-editor>
+    <div ng-if="isDisabled">
+      <my-schema-editor
+        ng-model="plugin.outputSchema"
+        ng-if="pluging.outputSchema"
```
astewes: is this supposed to be plugin instead of pluging?

astewes:
1 minor comment otherwise :+1:

PR #4383, opened 2015-10-23
ability to specify tag so that we can make changes to packer/template…

dereklwood:
- [x] ability to specify a git tag ``CDAP_TAG``, so that any released version of this script will always checkout the same chef-config template and packer scripts.  We should set this for each release.  It's optional so we can test against an unreleased branch.
- [x] added some comments


wolf31o2:
:+1: LGTM

PR #4384, opened 2015-10-23
(CDAP-4089) Temporary workaround by removing offending jar

dereklwood:
- [x] ugly hack to workaround https://issues.cask.co/browse/CDAP-4089, otherwise there is no way to start cdap_kafka.  This should be fixed in the next patch release.


wolf31o2:
Change to `rm -f` so it won't fail on a missing file. Otherwise, :+1: LGTM

PR #4385, opened 2015-10-23
Moving RowRecordTransformer and its unit test to etl-common so that i…

gokulavasan:
…t can be used in hydrator-plugins

Build : http://builds.cask.co/browse/CDAP-DUT3020-1


bdmogal:
:+1:

PR #4386, opened 2015-10-23
fixes horizontal line scroll in firefox

astewes:
*  Adds consistent styling to log viewer that eliminated unnecessary horizontal scroll in each log line

Firefox:
ff

Chrome:
chrome

Safari:
safari


elfenheart:
:+1: LGTM

PR #4387, opened 2015-10-23
Template for CI-based Docker builds

wolf31o2:
This will be used by the upcoming build pipeline to take build artifacts from previous builds and using them, versus only using released artifacts. This is done by replacing the URI with a location on the local disk, skipping the need for a cdap cookbook update to test new SDK versions in Docker/VM.

### DIFF
--- a/cdap-distributions/src/packer/files/cdap-sdk-with-uri.json.template
+++ b/cdap-distributions/src/packer/files/cdap-sdk-with-uri.json.template
@@ -1,7 +1,7 @@
 {
   "cdap": {
-    "version": "VERSION",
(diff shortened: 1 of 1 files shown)


dereklwood:
use ``{{VERSION}}`` and ``{{URL}}`` tokens to be consistent with our other similar scripts, and more importantly avoid introducing replacement ordering issues if we add another token like "FOO_VERSION".  Other than that, :+1: lgtm

wolf31o2:
Since this is a new template, I can do that. Previous versions of this already in use in the Docker and VM builds don't use the `{{TOKEN}}` syntax.

PR #4388, opened 2015-10-23
CDAP-4016: Java 8 support for ScriptTransform

jwang47:
https://issues.cask.co/browse/CDAP-4016
http://builds.cask.co/browse/CDAP-DUT3023

- Ran ScriptTransformTest, ScriptFilterTransformTest, and ValidatorTransformTest with Java 8.

### DIFF
--- a/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptTransform.java
+++ b/cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptTransform.java
@@ -36,6 +36,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
@@ -63,6 +65,9 @@ public class ScriptTransform extends Transform<StructuredRecord, StructuredRecor
   private final Config config;
   private Metrics metrics;
   private Logger logger;
+
+  @Nullable
+  private Method somValuesMethod;
 
   /**
    * Configuration for the script transform.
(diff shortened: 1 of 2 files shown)


cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptTransform.java
```diff
@@ -167,31 +184,46 @@ private StructuredRecord decodeRecord(Map nativeObject, Schema schema) {
     return builder.build();
   }
 
+  private List jsObject2List(Object object) {
+    if (object.getClass().getName().equals("jdk.nashorn.api.scripting.ScriptObjectMirror")) {
```
albertshau: do we need to do this classname comparison for each record or is there some way we can tell at initialize time?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptTransform.java
```diff
@@ -110,6 +115,18 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
...
     metrics = context.getMetrics();
     logger = LoggerFactory.getLogger(ScriptTransform.class.getName() + " - Stage:" + context.getStageId());
+
+    // for Nashorn (Java 8+) support -- get method to convert ScriptObjectMirror to List
+    try {
+      Class<?> somClass = Class.forName("jdk.nashorn.api.scripting.ScriptObjectMirror");
+      somValuesMethod = somClass.getMethod("values");
+    } catch (NoSuchMethodException e) {
+      throw new RuntimeException(
+        "Failed to method ScriptObjectMirror#values() for converting ScriptObjectMirror to List", e);
```
albertshau: Failed to get method ....

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptTransform.java
jwang47: If we assume we only support Java 7 and 8, then a non-null `somValuesMethod` would mean the object will be a ScriptObjectMirror.

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptTransform.java
albertshau: We should also include some pointer for what the user should do. Check that their Java includes nashorn?

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptTransform.java
albertshau: doing a null check would be better than string comparison if we know that's the correct behavior.

it's weird that they represent a json array as a map...

albertshau:
minor comments/questions, otherwise lgtm

cdap-app-templates/cdap-etl/cdap-etl-lib/src/main/java/co/cask/cdap/etl/transform/ScriptTransform.java
jwang47: Will do null check.

It's probably a map because all JavaScript objects are actually maps, including arrays.

PR #4389, opened 2015-10-23
Repository build script for YUM

wolf31o2:
Create a YUM repository tarball artifact.

Includes:
- shell script to create a YUM repository after building the `rpm` profile (and its dependencies)
- expect script used for GPG signing packages

### DIFF
--- a/cdap-distributions/bin/build_yum_repo.sh
+++ b/cdap-distributions/bin/build_yum_repo.sh
@@ -45,13 +45,10 @@ REPO_HOME=`pwd -P`
 TARGET_DIR=${DISTRIBUTIONS_HOME}/target
 STAGE_DIR=${TARGET_DIR}/yumrepo
 
-echo "REPO_HOME: ${REPO_HOME}"
-
-CDAP_HOME=${CDAP_HOME:-${REPO_HOME}}
 S3_BUCKET=${S3_BUCKET:-repository.cask.co}
 S3_REPO_PATH=${S3_REPO_PATH:-centos/6/x86_64/cdap} # No leading or trailing slashes
 VERSION=${VERSION:-$(basename ${TARGET_DIR}/cdap-*.rpm | cut -d- -f2)}
-__version=${VERSION/-SNAPSHOT//}
+__version=${VERSION/-SNAPSHOT/}
 __maj_min=$(echo ${__version} | cut -d. -f1,2)
 
 function die() { echo "ERROR: ${1}" 1>&2; exit 1; }
@@ -140,4 +137,5 @@ export_public_key
 createrepo_in_repo_staging || die "Failed to create repository from staging directory"
 create_repo_tarball || die "Failed to create YUM repository tarball"
 
+echo "Complete: cdap-yumrepo-${__maj_min}.tar.gz created"
 exit 0 # We made it!
--- a/cdap-distributions/bin/build_yum_repo.sh
+++ b/cdap-distributions/bin/build_yum_repo.sh
@@ -1,0 +1,143 @@
+#!/usr/bin/env bash
+#
+# Copyright © 2015 Cask Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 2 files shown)


cdap-distributions/bin/build_yum_repo.sh
```diff
@@ -0,0 +1,143 @@
...
+DISTRIBUTIONS_HOME=`pwd -P`
+cd `dirname ${DISTRIBUTIONS_HOME}` >&-
+REPO_HOME=`pwd -P`
+
+TARGET_DIR=${DISTRIBUTIONS_HOME}/target
+STAGE_DIR=${TARGET_DIR}/yumrepo
+
+echo "REPO_HOME: ${REPO_HOME}"
+
+CDAP_HOME=${CDAP_HOME:-${REPO_HOME}}
```
dereklwood: not used?  or, I suspect line 80 should use ``CDAP_HOME`` instead of ``REPO_HOME``

cdap-distributions/bin/build_yum_repo.sh
```diff
@@ -0,0 +1,143 @@
...
+S3_BUCKET=${S3_BUCKET:-repository.cask.co}
+S3_REPO_PATH=${S3_REPO_PATH:-centos/6/x86_64/cdap} # No leading or trailing slashes
+VERSION=${VERSION:-$(basename ${TARGET_DIR}/cdap-*.rpm | cut -d- -f2)}
+__version=${VERSION/-SNAPSHOT//}
```
dereklwood: ``__version=${VERSION/-SNAPSHOT/}`` (extra trailing slash)

cdap-distributions/bin/build_yum_repo.sh
```diff
@@ -0,0 +1,143 @@
...
+  echo "Copying packages to ${STAGE_DIR}/${__maj_min}/rpms"
+  for __rpm in $(ls -1 ${REPO_HOME}/*/target/*.rpm ${REPO_HOME}/target/*.rpm 2>/dev/null); do
+    echo "- ${__rpm}"
+    cp -f ${__rpm} ${STAGE_DIR}/${__maj_min}/rpms
+  done
+}
+
+function verify_signature_on_package() {
+  local __rpm=${1}
+  rpm --checksig ${__rpm} | grep -e pgp -e gpg # We want grep's return code, not rpm's
```
dereklwood: do we need to check for ``NOT OK`` to make sure its a known key?  thinking that if we only keep a single key on the build agents, then we could verify that it's the *same* key, which would be good since we later clobber any existing pubkey.gpg.

cdap-distributions/bin/build_yum_repo.sh
```diff
@@ -0,0 +1,143 @@
...
+
+# Here we go!
+setup_repo_staging || die "Something went wrong setting up the staging directory"
+add_packages_to_repo_staging || die "Failed copying packages to staging directory"
+sign_packages_in_repo_staging
+export_public_key
+createrepo_in_repo_staging || die "Failed to create repository from staging directory"
+create_repo_tarball || die "Failed to create YUM repository tarball"
+
+exit 0 # We made it!
```
dereklwood: Since this script is already verbose, a success message might be appropriate

cdap-distributions/bin/build_yum_repo.sh
wolf31o2: Nope. Since we only have the signing keys imported, a NOT OK would need to be resigned, anyway. Otherwise, it wouldn't match pubkey.gpg

cdap-distributions/bin/build_yum_repo.sh
wolf31o2: I just forgot to remove this.

cdap-distributions/bin/build_yum_repo.sh
wolf31o2: Either is correct. I'll remove the "extra" slash, though.

PR #4390, opened 2015-10-23
Cleanup Documentation Build

awholegunch:
- Cleanup of documentation build; removes cruft, duplicated code and code that has been either discarded or moved...
- Consolidate the sidebar topics into one section.
- Removes hard-coding of manual names in the page sidebar template.

Simplifies most conf.py files down to three lines, with nothing required to be specified.
Brings the common_conf.py code to match standard, by changing double-quotes to single except when used in a replace.

Building at: [Quick Build](http://builds.cask.co/browse/CDAP-DOB23-8) *(Revised link)*

[**Check out the new pages.**](http://builds.cask.co/artifact/CDAP-DOB23/shared/build-8/Docs-HTML/3.3.0-SNAPSHOT/en/index.html) *(Revised link)*


awholegunch:
Poorna tried this (on http://builds.cask.co/artifact/CDAP-DOB23/shared/build-3/Docs-HTML/3.3.0-SNAPSHOT/en/examples-manual/examples/sport-results.html) and it gave him an error, saying that the reporter needed to be set. I'll see if that can be fixed.

awholegunch:
Turns out that if you are in a group with the permission "ModifyReporter", you need to set that field; otherwise, this works fine and sets the field automatically to you, if you are logged in. @poornachandra You can always get dropped from that group, unless you really need it.

awholegunch:
"On-hold" because a recent build is broken.

awholegunch:
Fixed build, cleaned up duplicated code (multiple locations with the manual names duplicated). Re-built and passed. The main reason why this was never done in the first place was that it turns out to be tricky to do this with the templating engine used by Sphinx.

sreevatsanraman:
LGTM

PR #4391, opened 2015-10-23
CDAP-3945 adding runtime schedule requirements

albertshau:
Adding a Requirements object that describes different requirements
that must be satisfied before a schedule can be run. Starting off
with just a single one which sets a limit on the number of
concurrent runs for that schedule.


albertshau:
http://builds.cask.co/browse/CDAP-DUT3025

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
```diff
@@ -34,12 +38,14 @@ public Schedule(String name, String description, String cronEntry) {
     this.name = name;
     this.description = description;
     this.cronEntry = cronEntry;
+    this.runRequirements = null;
```
gsps1: this constructor is marked ``@Deprecated`` but doesn't say what should be used and when we should remove this usage, etc.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
```diff
@@ -0,0 +1,69 @@
...
+   * @param schedule the schedule to check
+   * @return whether all run requirements are satisfied
+   */
+  public boolean checkSatisfied(Id.Program programId, Schedule schedule) {
+    Integer programThreshold = schedule.getRunRequirements().getConcurrentProgramRunsThreshold();
+    if (programThreshold != null) {
+      try {
+        List<RunRecordMeta> running =
+          store.getRuns(programId, ProgramRunStatus.RUNNING, 0, Long.MAX_VALUE, programThreshold + 1);
+        if (running.size() > programThreshold) {
```
gsps1: should be `if (running.size() >= programThreshold) {`` right?

gsps1: nevermind, makes sense as the api is referred as ``exceeds``

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Checks whether all schedule requirements are met for the schedule to run.
+ */
+public final class RunRequirementsChecker {
+  private static final Logger LOG = LoggerFactory.getLogger(RunRequirementsChecker.class);
+
+  private final Store store;
+
+  RunRequirementsChecker(Store store) {
```
gsps1: public ?

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
gsps1: also maybe better to call ``this(name, desc, cron, runRequirements)``

cdap-api/src/main/java/co/cask/cdap/api/schedule/RunRequirements.java
```diff
@@ -0,0 +1,65 @@
...
+   *         For example, if set to 0, the scheduled run will be skipped if that are any program runs
+   *         in the RUNNING state. Returns null if there is no limit.
+   */
+  @Nullable
+  public Integer getConcurrentProgramRunsThreshold() {
+    return concurrentProgramRunsThreshold;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
gsps1: better to name the variable as ``other`` ?

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
gsps1: also this can cause NPE in ``RunRequirementsChecker`` when we execute ``schedule.getRunRequirements().getConcurrentProgramRunsThreshold()``, or is this constructor  not executed anywhere? seems unclear.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
```diff
@@ -0,0 +1,69 @@
...
+   * @param programId the id of the program to check
+   * @param schedule the schedule to check
+   * @return whether all run requirements are satisfied
+   */
+  public boolean checkSatisfied(Id.Program programId, Schedule schedule) {
+    Integer programThreshold = schedule.getRunRequirements().getConcurrentProgramRunsThreshold();
+    if (programThreshold != null) {
+      try {
+        List<RunRecordMeta> running =
+          store.getRuns(programId, ProgramRunStatus.RUNNING, 0, Long.MAX_VALUE, programThreshold + 1);
```
gsps1: if programThreshold is ``Integer.MAX_VALUE`` this will cause overflow and the value will be negative for limit. while ``Integer.MAX_VALUE`` is a valid input for that API.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/ScheduleTaskRunner.java
```diff
@@ -86,6 +84,9 @@ public ScheduleTaskRunner(Store store, ProgramLifecycleService lifecycleService,
     }
 
     ScheduleSpecification spec = appSpec.getSchedules().get(scheduleName);
+    if (!requirementsChecker.checkSatisfied(programId, spec.getSchedule())) {
+      return Futures.<Void>immediateFuture(null);
```
gsps1: should we throw an exception instead?

gsps1:
@albertshau few comments

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
albertshau: all good points, will fix and also add a null check when getting it.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
albertshau: didn't see a need for other packages to use it

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
albertshau: good point, will check for that

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/ScheduleTaskRunner.java
albertshau: don't think we want to throw an exception here as its not really an error in cdap.  Though perhaps we would want to record somewhere why the run was skipped...

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedules.java
```diff
@@ -64,4 +64,87 @@ private Schedules() {
...
+
+    /**
+     * Skip running the scheduled program if the number of concurrent runs of the program exceeds the specified
+     * threshold. Prior to running the program, the scheduler will check how many runs of the program are currently
+     * in the RUNNING state.
+     * If that number is greater than the specified threshold, the scheduler will not run the program.
+     * This number includes runs of the program that were started outside of this schedule,
+     * either manually or by another schedule. For example, if the threshold is set to 0, the scheduler will
+     * only run the program if there are no other runs of the program currently in the RUNNING state.
+     *
```
anew: So if I understand correctly, what will happen is:
1. Time schedule triggers
2. Scheduler checks how many runs are currently in RUNNING state. If that exceeds the threshold: skip this run
3. Start the new run
4. New run goes into STARTING state
5. New run goes into RUNNING state
 
I think there can be a race condition: During step 2, some runs may be submitted but not in RUNNING state yet. Perhaps STARTING. I think step 2 should at least count STARTING runs in addition to RUNNING ones. Even with that, another run could be between steps 3 and 4 (submitted but not STARTING yet). Or is that not possible?

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedules.java
```diff
@@ -64,4 +64,87 @@ private Schedules() {
...
+     * If no threshold is set, there will be no limit on the number of concurrent runs.
+     *
+     * @param threshold skip the scheduled run if the number of concurrent program runs is above this threshold
+     * @return the Builder
+     */
+    public Builder skipIfConcurrentProgramRunsExceed(int threshold) {
```
anew: Is there also a way to defer the new run until the concurrency is below the threshold? Or does it have to be skipped?

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedules.java
```diff
@@ -64,4 +64,87 @@ private Schedules() {
...
+      if (threshold < 0) {
```
anew: does 0 make sense? Isn't that equivalent to no threshold?

cdap-api/src/main/java/co/cask/cdap/api/schedule/RunRequirements.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * Defines requirements that must be satisfied at runtime in order for the scheduler to launch a run.
+ *
+ * Currently only contains the maximum number of concurrent runs. In the future, other checks may be added,
+ * such as the amount of available memory in the YARN cluster.
+ */
+public class RunRequirements {
```
anew: Is RunConstraints a better name? It describes limitations rather than requirements. But that is just my take...

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
anew: the the class itself does not need to be public?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
```diff
@@ -0,0 +1,72 @@
...
+  public boolean checkSatisfied(Id.Program programId, Schedule schedule) {
+    Integer programThreshold = schedule.getRunRequirements().getConcurrentProgramRunsThreshold();
+    if (programThreshold != null) {
+      try {
+        // limit the number of run records we get to try and be a little more efficient.
+        // make sure we don't overflow by adding 1.
+        int max = programThreshold == Integer.MAX_VALUE ? Integer.MAX_VALUE : programThreshold + 1;
+        List<RunRecordMeta> running =
+          store.getRuns(programId, ProgramRunStatus.RUNNING, 0, Long.MAX_VALUE, max);
+        if (running.size() > programThreshold) {
```
anew: If the threshold is 5, and 5 runs are currently RUNNING, the should the new run start or be skipped? I feel that it should be skipped. But this code would allow it to start. It will only skip it if there are more than 5 RUNNING. 

Actually I would find it more intuitive if the RunRequirement (or RunConstraint) would configure a maxConcurrentRuns. If I set that to 5, then a new run would only be started if there are at most 4 currently running.

anew:
@albertshau I did a pass over this. Please see my comments

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
albertshau: true, will change it

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedules.java
albertshau: it's confusing, but run records have different state transitions than programs.  The ProgramRunStatus enum does not have a STARTING state. When a program has the STARTING status, its run record has the RUNNING status.

It's true there is a race in between 3 and 4 like you mentioned. Somebody could start a workflow manually, or if there could be multiple schedules that trigger at the same time.  We would either have to do some sort of locking, or provide a weaker guarantee at the schedule level instead of the program level.

cdap-api/src/main/java/co/cask/cdap/api/schedule/RunRequirements.java
albertshau: constraints sounds better, will rename

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunRequirementsChecker.java
albertshau: both you and shankar brought this up so I will change it to set a max instead of skipping if it exceeds.

cdap-api/src/main/java/co/cask/cdap/api/schedule/RunConstraints.java
```diff
@@ -0,0 +1,63 @@
...
+ * Defines constraints that must be met at runtime in order for the scheduler to launch a run.
+ *
+ * Currently only contains the maximum number of concurrent runs. In the future, other checks may be added,
+ * such as the amount of available memory in the YARN cluster.
+ */
+public class RunConstraints {
+  public static final RunConstraints NONE = new RunConstraints(null);
+  private final Integer maxConcurrentRuns;
+
+  RunConstraints(Integer maxConcurrentRuns) {
```
anew: parameter should be Nullable and it should have Javadocs

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
```diff
@@ -74,30 +88,17 @@ public boolean equals(final Object o) {
...
-    if (name != null ? !name.equals(schedule.name) : schedule.name != null) {
-      return false;
-    }
+    Schedule that = (Schedule) o;
 
-    return true;
+    return Objects.equals(name, that.name) &&
+      Objects.equals(description, that.description) &&
+      Objects.equals(cronEntry, that.cronEntry) &&
+      Objects.equals(runConstraints, that.runConstraints);
```
anew: would a schedule with RunConstraints.NONE be equal to a schedule without RunConstraints? Not sure where this is used, but above you are treating them the same for backwards-compatibility.

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
```diff
@@ -74,30 +88,17 @@ public boolean equals(final Object o) {
...
   }
 
   @Override
   public int hashCode() {
-    int result = name != null ? name.hashCode() : 0;
-    result = 31 * result + (description != null ? description.hashCode() : 0);
-    result = 31 * result + (cronEntry != null ? cronEntry.hashCode() : 0);
-    return result;
+    return Objects.hash(name, description, cronEntry, runConstraints);
```
anew: similar here.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunConstraintsChecker.java
```diff
@@ -0,0 +1,93 @@
...
+   * Checks if the run requirements for the specified schedule and program are all satisfied.
+   *
+   * @param programId the id of the program to check
+   * @param schedule the schedule to check
+   * @return whether all run requirements are satisfied
+   */
+  public boolean checkSatisfied(Id.Program programId, Schedule schedule) {
+    String scheduleName = schedule.getName();
+    Integer maxRuns = schedule.getRunConstraints().getMaxConcurrentRuns();
+    Predicate<RunRecordMeta> scheduleFilter = getScheduleFilter(scheduleName);
```
anew: this can be inside the if, and same for the scheduleName

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
albertshau: hm thats true. I guess it's simpler to remove NONE and have null mean there are none.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunConstraintsChecker.java
```diff
@@ -0,0 +1,93 @@
...
+    if (maxRuns != null) {
+      try {
+        int numActive = 0;
+        List<RunRecordMeta> runs = store.getRuns(programId, ProgramRunStatus.RUNNING, 0, Long.MAX_VALUE, maxRuns,
+                                                 scheduleFilter);
+        numActive += runs.size();
+        runs = store.getRuns(programId, ProgramRunStatus.SUSPENDED, 0, Long.MAX_VALUE, maxRuns,
```
anew: why SUSPENDED?

anew: Hmm I see, they also count. Perhaps only do this if numActive does not already exceed maxRuns? To save a scan in the common case.

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/ScheduleTaskRunner.java
anew: the checker logs it, right?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunConstraintsChecker.java
```diff
@@ -0,0 +1,93 @@
...
+                             scheduleFilter);
+        numActive += runs.size();
+
+        if (numActive >= maxRuns) {
+          LOG.info("Skipping run of program {} because there are already {} active runs of schedule {}.",
```
anew: Perhaps log how many running and how many suspended?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/store/DefaultStore.java
```diff
@@ -265,7 +271,7 @@ public Void apply(AppMetadataStore mds) throws Exception {
 
   @Override
   public void setStart(Id.Program id, String pid, long startTime) {
-    setStart(id, pid, startTime, null, ImmutableMap.<String, String>of());
+    setStart(id, pid, startTime, null, ImmutableMap.<String, String>of(), ImmutableMap.<String, String>of());
```
anew: Don't we have a constant somewhere for no arguments?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunConstraintsChecker.java
albertshau: the thought was that somebody probably suspends a program because they want to limit resource usage, or some other system needs to be fixed, or some other reason that makes it not a good time to run the program. So it's probably not desired for another run to kick off. Do you think we shouldn't count suspended programs?

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedules.java
```diff
@@ -64,4 +64,88 @@ private Schedules() {
...
+      return this;
+    }
+
+    /**
+     * Create a time schedule based on the specified cron expression.
+     *
+     * @param cronExpression the cron expression specifying when the scheduler should attempt to start a run
+     * @return the time schedule
+     */
+    public Schedule createTimeSchedule(String cronExpression) {
```
anew: should the old static methods be deprecated?

cdap-app-fabric/src/main/java/co/cask/cdap/internal/app/runtime/schedule/RunConstraintsChecker.java
albertshau: oops should refresh the page more often

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedules.java
albertshau: I think they should, but wanted to do it separately since they're used all over the place in tests.

anew:
I did another pass. The API is much better now. 

Just to clarify: 
- we only support limiting the concurrent runs of a particular schedule. If the same program has active runs from another schedule (or manually started), they won't affect the constraint.
- we can only skip a schedule, not defer it until the constraints are met

If this is what was required, then this looks pretty good. Most of my comments are minor or style related.

cdap-api/src/main/java/co/cask/cdap/api/schedule/Schedule.java
anew: or use getRunConstraints() instead of accessing the field directly, here and in hashCode and toString.

albertshau:
thanks for the review.  Both points are correct, I'll be sure to put them in the javadocs.

anew:
LGTM. Please open a Jira for deprecating the old Schedules APIs.

